PR #8, opened 2019-12-17
Create mfwg-delegates.md

sffc:
I created a Markdown file with the delegate acronyms as well as instructions for choosing your acronym.

Although for TC39-TG2, I try to upstream the delegate acronyms, since MFWG is chartered under the Unicode Consortium, I would like to maintain a separate list, with the recommendation that the acronyms be consistent.

@romulocintra @mathiasbynens


romulocintra:
@sffc  LGTM, probably i would change filename to delegates.md

sffc:
> @sffc LGTM, probably i would change filename to delegates.md

Done

PR #14, opened 2020-01-14
Update README.md

srl295:
Document that this is a CLDR WG

may be other improvements here


romulocintra:
LGTM,   maybe is missing : 
- [ ] Licence
- [ ] Code of Conduct 

Should we add them?

PR #18, opened 2020-01-27
Message Format Requirements Template

romulocintra:


romulocintra:
cc @echeran  @nbouvrette

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,20 @@
...
+
+---
+
+**Is your feature request related to a problem? Please describe.**
+A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
+
+**Describe the solution you'd like**
+A clear and concise description of what you want to happen.
+
+**Describe alternatives you've considered**
```
sffc: For the third question, maybe instead do something like: "Why does your solution belong in the standard instead of in user land?"

romulocintra: Cool, also rephrased other questions

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,20 @@
...
+name: Feature request Message Format WG
+about: Suggest an idea or feature for Message Format
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+**Is your feature request related to a problem? Please describe.**
+A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
```
echeran: I think rephrasing "I'm always frustrated when [...]" to be something like "It is difficult to [...]" so that we de-emphasize the individual and try to keep emotions out of (what is probably) an objective technical issue.

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -13,8 +13,8 @@ A clear and concise description of what the problem is. Ex. I'm always frustrate
 **Describe the solution you'd like**
 A clear and concise description of what you want to happen.
 
-**Describe alternatives you've considered**
-A clear and concise description of any alternative solutions or features you've considered.
+**Describe why does your solution should become a standard**
```
sffc: Remove "does" to fix the grammar.

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,20 @@
...
+
+---
+
+**Is your feature request related to a problem? Please describe.**
+A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
+
+**Describe the solution you'd like**
+A clear and concise description of what you want to happen.
+
+**Describe why does your solution should become a standard**
```
echeran: nit: Instead of "become a standard", can we say "shape the standard"?  I don't know if any one pre-existing project or proposal is going to be a full and complete standard, which is what the wording of the question kind of sounds like.  Feel free to disagree or ignore.

PR #24, opened 2020-01-28
Add delegates and update acronyms

sffc:
I'm adding new delegates from yesterday's meeting.  I am also changing all delegates to have 3-letter acronyms, since we are operating separately from TC39.  This means that I added a letter to @zbraniecki and @rxaviers.  Let me know if it look OK.  (This only affects MF-WG, not ECMA-402)


romulocintra:
LGTM!

PR #76, opened 2020-04-14
Draft decision process

DavidFatDavidF:
Strawman of the decision making process to be discussed in the wg monthly meeting on 20th April 2020

### DIFF
--- a/guidelines/decision-process.md
+++ b/guidelines/decision-process.md
@@ -1,7 +1,5 @@
 # Decision making process
 
-## Preamble
-The Chair Group discussion led to a consensus to start with a light weight consensus-based process that would eventually use voting as a last resort method in case of impasse.. The group could eventually be in need to defend its decision making process against trolls, yet at the moment all group members are considered in good faith, so the working group reserves the option to specify disciplinary actions against trolls as part of its guidelines at a later stage..
 
 ## Definitions
 1. *Consensus* is defined as lack of sustained opposition.
--- a/guidelines/decision-process.md
+++ b/guidelines/decision-process.md
@@ -1,1 +1,23 @@
+# Decision making process
 
+## Preamble
+The Chair Group discussion led to a consensus to start with a light weight consensus-based process that would eventually use voting as a last resort method in case of impasse.. The group could eventually be in need to defend its decision making process against trolls, yet at the moment all group members are considered in good faith, so the working group reserves the option to specify disciplinary actions against trolls as part of its guidelines at a later stage..
+
+## Definitions
+1. *Consensus* is defined as lack of sustained opposition.
+2. *Good standing* is a characteristic of group members who fullfill their implicit and explicit obligations and hence are allowed to excercise all of their rights within the group without restriction.
+3. *Blacklist* (*blacklisted*, *blacklisting*) is a taxative enumeration of group mebers who are temporarily excluded from excercising their rights within the group.
+4. *Normative keywords* as defined in [BCP 14](https://tools.ietf.org/html/bcp14)
+
+## Rules
+1. Any current group member is deemed to be in good standing unless blacklisted.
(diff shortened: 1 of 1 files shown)


guidelines/decision-process.md
```diff
@@ -1 +1,23 @@
+# Decision making process
 
+## Preamble
+The Chair Group discussion led to a consensus to start with a light weight consensus-based process that would eventually use voting as a last resort method in case of impasse.. The group could eventually be in need to defend its decision making process against trolls, yet at the moment all group members are considered in good faith, so the working group reserves the option to specify disciplinary actions against trolls as part of its guidelines at a later stage..
```
nbouvrette, line 4: Why the double dots?

> case of impasse..
> later stage..

DavidFatDavidF: First, not married to the double dots.. feel free to convert them to full stops.
Second, it's in the informal preamble/background piece indicating a topic that can have development, more could be said if necessary..

guidelines/decision-process.md
```diff
@@ -1 +1,23 @@
...
+
+## Definitions
+1. *Consensus* is defined as lack of sustained opposition.
+2. *Good standing* is a characteristic of group members who fullfill their implicit and explicit obligations and hence are allowed to excercise all of their rights within the group without restriction.
+3. *Blacklist* (*blacklisted*, *blacklisting*) is a taxative enumeration of group mebers who are temporarily excluded from excercising their rights within the group.
```
romulocintra, line 9:
```suggestion
3. *Blacklist* (*blacklisted*, *blacklisting*) is a taxative enumeration of group members who are temporarily excluded from exercising their rights within the group.
```

guidelines/decision-process.md
romulocintra, line 4: > so the working group reserves the option to specify disciplinary actions against trolls as part of its guidelines at a later stage..

 Not sure about the phrase(if it should belong to another place)

guidelines/decision-process.md
```diff
@@ -1 +1,23 @@
...
+4. *Normative keywords* as defined in [BCP 14](https://tools.ietf.org/html/bcp14)
+
+## Rules
+1. Any current group member is deemed to be in good standing unless blacklisted.
```
romulocintra, line 13: Instead of Rules probably i would call it "Process" what do you think ?

DavidFatDavidF: I think that it's indeed rules rather than process steps..

guidelines/decision-process.md
DavidFatDavidF: This is an informal preamble explanation of why the blacklisting process is TBD atm

The whole preamble can be deleted, it doesn't add anything normative, it's just to explain why the rules are currently lightweight and partially underspecified.. 

It is intended to help the reader of the normative part to understand why things are the way they are ;-)

aphillips: I think it is foolish to leave this undefined--having experienced some world-class trolls and/or ill-behaved persons (you haven't lived until you have received death threats online??). We'd be better off to handle this [the way CLDR and ICU do](http://cldr.unicode.org/index/process): with a reference to the formal rules, located [here](http://www.unicode.org/consortium/tc-procedures.html). In practice, voting and such are almost never used in favor of consensus.

romulocintra: > (you haven't lived until you have received death threats online??). We'd be better off t
😂😂😂😂😂😂😂😂😂


> We'd be better off to handle this [the way CLDR and ICU do](http://cldr.unicode.org/index/process): with a reference to the formal rules, located [here](http://www.unicode.org/consortium/tc-procedures.html). 

Agree!!

DavidFatDavidF:
I removed Preamble.. This was agreed to be our decision making process in the monthly meeting on 20th April 2020.

PR #84, opened 2020-05-14
Why `MessageFormat` needs a successor

mihnita:
[Issue #49](https://github.com/unicode-org/message-format-wg/issues/49)

In preparation for the group meeting.

### DIFF
--- a/doc/why_mf_next.md
+++ b/doc/why_mf_next.md
@@ -22,20 +22,33 @@ Other efforts: [Fluent](https://projectfluent.org/),
 
 ## Core problems with the current `MessageFormat`
 
-1. Does not have any “extension points”
-2. Can't remove anything, even if now we know better
+1. The design is not modular enough
+    * Does not have any “extension points”
+    * Can't deprecate anything, even if now we know better
+2. Some existing problems
 3. Hard to map to the existing localization core structures
 4. Designed to be API only, plain text, UI, “imperative style”
 
-### 1. Does not have any “extension points”
+### 1. The design is not modular enough
+
+The "data model" is hard-coded in the standard and in the syntax.
+This makes it very rigid.
+
+#### 1.1 Does not have any “extension points”
 
 No extension points means that it is hard to add new functionality unless you
 are doing it in ICU itself.
+
 It also means most tools used to process these messages are built rigidly,
 and are unprepared to handle changes
-(think localization tools, liners, friendly UIs, etc.).
+(think localization tools, linters, friendly UIs, etc.).
 
-### 2. Can't remove anything, even if now we know better
+The most basic functionality would be adding a new formatter. Meantime ICU
+added other formatters: time intervals, measurement, lists. But MessageFormat
+did not keep up. And adding support for any of these new formats risks to break
+existing tools.
+
+### 1.2. Can't deprecate anything, even if now we know better
 
 ICU is old, but also very popular (right now it is the core i18n library
 for all major operating systems, and many products).
@@ -50,6 +63,23 @@ teaching us what works (for instance skeletons did not exist when the
 date/time parameters were added).
 
 But the stability requirements prevent any major cleanup.
+
+### 2. Some existing problems
+* ICU added new formatters, but MessageFormat does not support them
+* Combined selectors (select + plural) results in unreadable and error
+prone nesting
+* Select and plurals inside the message are difficult to translate because of 
+grammatical agreement requires words outside select / plural to change.
+See https://en.wikipedia.org/wiki/Agreement_(linguistics)
+* Patterns in the date / time / number placeholders are bad i18n, should use skeletons
+* No official support for gender. It can be done with `select`, but it
+is not the same thing (same as the difference between an `enum` and integer/strings). Developers can use masculine/feminine, masc/fem, male/female, etc.
+* Formatting for “parameters” known at compile time
+* Escaping with apostrophe is error prone. There is no reliable way to tell if
+it has to be doubled or not.
+* The # is used in plural format instead of {...}, but does not work for nesting unless the plural is the innermost selector. But named placeholders don't work
+properly for plurals with offset. So there are 2 ways to do the same thing that work in 98% of cases, but in special situations only one of the ways works. 
+* Does not support inflections, and it would be hard to add without breaking existing tools.
 
 ### 3. Hard to map to the existing localization core structures
 
--- a/doc/why_mf_next.md
+++ b/doc/why_mf_next.md
@@ -4,39 +4,28 @@
 
 The `MessageFormat` API and syntax have been around for a long time.
 
-Its “ancestor”, [java/text/MessageFormat](https://docs.oracle.com/javase/7/docs/api/java/text/MessageFormat.html), was introduced with Java 1.4, February 2002.
+Intro
+* `MessageFormat` is the Unicode API for software localization
+* It is 20 years old, well designed, proven solution
+* Its design is optimized for the software development model of 20y ago and its
+shortcomings result in mixed reception and adoption by the industry.
 
-The [ICU MessageFormat](https://unicode-org.github.io/icu-docs/apidoc/released/icu4j/com/ibm/icu/text/MessageFormat.html) is tagged as stable API since ICU 3.0 (June 2004)
+The current wave of software development uses dynamic languages, modern UI
+frameworks and new forms of user interactions (voice, VR etc.).
 
-The ICU version evolved compared to the JDK one:
-* added support for plurals (ICU 3.8, 2007)
-* added support for select (ICU 4.4, 2010)
-* named arguments (`...{user}...` vs `...{0}...`)
-* better handling of the apostrophe escaping
-* date/time/number skeletons (2018)
-* more
+Considering these new challenges, combined with the lessons learned from using
+`MessageFormat`, we aim to design the next iteration of `MessageFormat`
+suitable for current generation of software, and adoption by Web Standards.
 
-Despite being around for such a long time, it is still not well supported by localization tools.
-
-Other efforts: Fluent, FB
+Other efforts: [Fluent](https://projectfluent.org/),
+[FBT](https://facebook.github.io/fbt/)
 
 ## Core problems with the current `MessageFormat`
 
-I've started with the list of problems in the next section.
-But that ends up being a (biased) reshuffling of the issues and feature requests that we collected in [GitHub](https://github.com/unicode-org/message-format-wg/issues)
-
-So I have tried to distill that to a few root causes
-(think [“5 why”](https://en.wikipedia.org/wiki/Five_whys))
-
-I think that these are the problems we need to avoid repeating,
-otherwise it is just a matter of time until we end up in the same place.
-
-Here it is:
 1. Does not have any “extension points”
-2. Not a formal standard with an “acceptance test suite”
-3. Can't remove anything, even if we know know better
-4. Hard to map to the existing localization core structures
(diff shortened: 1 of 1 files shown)


mihnita:
@Zibi: thank you for offering to help.
I had you in mind since I started writing this :-)

---

If you look at the history of the file, I started with a list of bullets that were the same as the list of issues we already collected in GitHub.
But "filtered" through my opinionated view of things. 
And that would not be very useful.

doc/why_mf_next.md
```diff
@@ -0,0 +1,113 @@
...
+* added support for plurals (ICU 3.8, 2007)
+* added support for select (ICU 4.4, 2010)
+* named arguments (`...{user}...` vs `...{0}...`)
+* better handling of the apostrophe escaping
+* date/time/number skeletons (2018)
+* more
+
+Despite being around for such a long time, it is still not well supported by localization tools.
+
+Other efforts: Fluent, FB
```
rxaviers, line 21: Does FB refer to https://github.com/facebookincubator/fbt/?
```suggestion
Other efforts: Fluent, FBT
```

doc/why_mf_next.md
```diff
@@ -0,0 +1,113 @@
...
+And it was API only. \
+No standard way to store the stings in a serialized format and to carry
+info or directives for translators or localization tools.
+No comments, length limits, protecting non-translatable sections of text, etc.
+
+---
+
+**Mandatory xkcd:**
+
+[<img src='https://imgs.xkcd.com/comics/standards.png'>](https://xkcd.com/927/)
```
romulocintra, line 113: 😂😂😂😂😂😂😂😂😂😂😂

zbraniecki:
Going high-level first.

I like the "5 points" - it simplifies the logic and makes it easy to read "what they're addressing" and "what should I expect out of the outcome".

I don't like personal sentences in WG documents. Sentences like "I've started with..." read more like a personal take than a formal position of the WG.

Here's my summary of your outline:

* Intro
  * MessageFormat is the Unicode API for software localization
  * It's 20y old, well designed, proven solution
  * Its design is optimized for the software development model of 20y ago and its shortcomings result in mixed reception and adoption by the industry.
  * The current wave of software development coming from dynamic languages, modern UI frameworks and new forms of user interactions (voice, VR etc.) combined with the lessons learned from MessageFormat, we aim to design the next iteration of MessageFormat suitable for current generation of software and adoption by Web Standards.
* Core problems
  * Non-extensible syntax
  * Every "feature" encoded in the data model
  * Limited specification and tooling around semantic features
  * Designed for imperative APIs only

I removed pt (2) from your listing because it doesn't feel like a problem that requires fixing via new iteration - we could fix it just by adding test suite. Let me know if you agree.

On the other hand, I see a number of MF shortcomings that are not captured here:

* MF syntax makes it hard to recover from resolution errors, which makes the system less resilient and limits fallbacking.
* MF doesn't provide any meta-data ([semantic comments](https://github.com/projectfluent/fluent/issues/139), localizer comments, string versioning etc.)
* Limited multi-variant messages (I'm not sure how to phrase it, but it seems that we aim to provide more generic ability to provide variants of the same message that depend on some information, while MF is designed to allow for variants that are tied to basically PLURAL since custom formatters are deprecated and never got any adoption)
* Multiline strings are not well supported (not sure if the handling of whitespace in multilines is actually specified well, for example for HTML fragments)
* Modern formatters enable much richer/better integration that MF wasn't designed for, and with mixed formatting control

I trust you on selection of those, so feel free to incorporate or reject any and all of the above.

Finally, I'm wondering if this is the right place to extract the conclusion from the shortcoming listed by you about "hard to add/remove features" into some form of "L10n systems are hard to change, they stay for decades as software changes. Encoding every feature deep into the data model and syntax does not stand the test of time. In conclusion we believe that designing a more generic data model on top of which we can develop features that can be extended and even deprecated as the time goes without affecting low-level foundational tooling is critical for the longevity of the system".
Do you think it's the consensus or am I projecting my personal bias here? :)

doc/why_mf_next.md
Fleker: Would it be better to include links to each project?

mihnita:
> I don't like personal sentences in WG documents. Sentences like "I've started with..." read more like a personal take than a formal position of the WG.

Thanks, I'll do that
(for a long time I've been told that I am "too forceful" and learned to "sprinkle" a lots of "I think", "in my opinion" and so on :-)

I'll go through your list and refactor, but can't take you full suggestion "as is"

There are things that are not addressed because they are not `MessageFormat` issues, or are already included, or can be fixed in the existing form, no need for a new version:

---

> * MF syntax makes it hard to recover from resolution errors, which makes the system less resilient and limits fallbacking.

There is no fallback or resolution, because "Designed to be API only", so it is already captured.

The loading of the string (including resolution / fallback) is the responsibility of some other component. You load the string (using the Android ResourceManager, Java Resources, Windows / MacOS APIs, C/C++ gettext), then pass it to the `MessageFormat` class.

This allows one to use ICU without being force to migrate the whole resource resolution out of the OS (making the resolution results inconsistent in the process).

Add this to the "detailed notes" of the bullet?

---

> * MF doesn't provide any meta-data ([semantic comments](https://github.com/projectfluent/fluent/issues/139), localizer comments, string versioning etc.)
> * Multiline strings are not well supported (not sure if the handling of whitespace in multilines is actually specified well, for example for HTML fragments)

Same for both items, addressed by "Designed to be API only", not a complete solution.

You can use whatever metadata the native string storage mechanism you use.
`MessageFormat` is agnostic in that respect, treats this as "out of scope"
If I have a way to say add metadata to non-ICU messages (comments, examples, length limits) in "strings storage", then I am free to use them for ICU messages too.
Same for multiline.
One can say that "Java properties are not multiline friendly"

Whitespaces and multiline can't be specified well in HTML fragments to begin with, `MessageFormat` or not better or worse in this respect.
If my fragment is used inside a `<pre>` or a div with `white-space` set to `nowrap` / `pre` / `pre-line` / `pre-wrap` then I get different behavior. If my fragment is used in 2 different contexts, it might end up rendered differently.

So `MessageFormat` is again agnostic here. Solving multiline spaces are the problem of the string loading mechanism, or of the rendering part.

And all of the 3 issues above can be solved without a new version of `MessageFormat`, we "only need to add a "MessageFormat store", with resolution + fallback.

But the price is loss of flexibility: now as an iOS / Android / Qt developer I need to either move ALL my strings to a "MF store", with its own resolution / fallback / metadata, or keep non MF strings in a "native store" and MF strings in a "MF store" (with the huge risk that the resolution is different)

---

> * Limited multi-variant messages (I'm not sure how to phrase it, but it seems that we aim to provide more generic ability to provide variants of the same message that depend on some information, while MF is designed to allow for variants that are tied to basically PLURAL since custom formatters are deprecated and never got any adoption)
> * Modern formatters enable much richer/better integration that MF wasn't designed for, and with mixed [formatting control](https://github.com/projectfluent/fluent/wiki/Fluent-and-ICU-MessageFormat#formatting-control)

Agree with both.
But I've considered that to be captured in **"Does not have any “extension points”"**
We can split it in a separate point, or add this clarification in the explanation of the bullet?

---

> Finally, I'm wondering if this is the right place to extract the conclusion from the shortcoming listed by you about "hard to add/remove features" into some form of "L10n systems are hard to change, they stay for decades as software changes. Encoding every feature deep into the data model and syntax does not stand the test of time. In conclusion we believe that designing a more generic data model on top of which we can develop features that can be extended and even deprecated as the time goes without affecting low-level foundational tooling is critical for the longevity of the system".

I agree with this, and I'll see how I take this and use it.
Most likely not a bullet, but probably in the explanation section of *"Hard to map to the existing localization core structures"*

---

> Do you think it's the consensus or am I projecting my personal bias here? :)

As you might have noticed in my answers above, there is some bias (probably yours and mine :-)
Let me "sleep on it" a bit and I'll see what we get.

doc/why_mf_next.md
mihnita: > Would it be better to include links to each project?

Will do.

stasm:
Would it be a good idea to enumarate with more detail the design mistakes which you briefly touch on in point 3? Things like positional and named parameters, escaping, etc. I feel like each would benefit from its own bullet point with an example and the rationale about why we think they didn't work out.

doc/why_mf_next.md
```diff
@@ -0,0 +1,113 @@
...
+* date/time/number skeletons (2018)
+* more
+
+Despite being around for such a long time, it is still not well supported by localization tools.
+
+Other efforts: Fluent, FB
+
+## Core problems with the current `MessageFormat`
+
+I've started with the list of problems in the next section.
```
nbouvrette, line 25: Maybe this is just me, but if the goal of the document to explain the rationale behind the working group's existence, it might be better phrased at the 3rd person point of view?

romulocintra: Agree

mihnita:
> Would it be a good idea to enumarate with more detail the design mistakes which you briefly touch on in point 3?

I'm afraid that would make this document too big.

There is an older Google Doc (public, shared) where I listed what I like / dislike about MessageFormat. I can link to it. Or I can convert it to a markdown file and post it somewhere (where?)
That document is "kind of opinionated", it is not necessarily the position of the working group.

But if others agree to include all that here I am also OK to include some of that doc here.

mihnita:
Changed things quite a bit. I've tried to capture / address all of your points, even if that didn't change the bullets too much.

doc/why_mf_next.md
```diff
@@ -0,0 +1,113 @@
...
+
+Most of it can't be “blamed” directly on a bad decision, it is just time
+teaching us what works (for instance skeletons did not exist when the
+date/time parameters were added).
+
+But the stability requirements prevent any major cleanup.
+
+### 4. Hard to map to the existing localization core structures
+
+The format is not supported by any major localization system that I know of. \
```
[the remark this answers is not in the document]
mihnita: The term was "intentionally vague", since I didn't want to get into debates about TMS vs GMS vs CAT vs whateve-else-abbreviations-people-use.
---

Support means most of the stuff people do with other formats: import, leverage, validate, export.
I don't count "take the file as is, with curly brackets and all, and let translators edit it raw" as "support".

doc/why_mf_next.md
```diff
@@ -0,0 +1,116 @@
...
+3. Hard to map to the existing localization core structures
+4. Designed to be API only, plain text, UI, “imperative style”
+
+### 1. Does not have any “extension points”
+
+No extension points means that it is hard to add new functionality unless you
+are doing it in ICU itself.
+It also means most tools used to process these messages are built rigidly,
+and are unprepared to handle changes
+(think localization tools, liners, friendly UIs, etc.).
```
zbraniecki, line 36: I'd maybe add that, from the experience with Fluent, the fact that almost every feature in MF is a syntax feature, means that on the very lowest level - AST, parser - every extensions basically breaks the existing functionality because the file cannot be parsed and its AST operated on.

Fluent's AST is more generic which means that more of the functionality falls on higher levels of abstraction and in fact, over the last 3 years, all accrued "wants" can be implemented either on the higher level (Semantic Comments, Dynamic References etc.) or require additive syntax relaxations (flatten selectors, rich overlays).

doc/why_mf_next.md
```diff
@@ -0,0 +1,146 @@
+# Why `MessageFormat` needs a successor ([issue #49](https://github.com/unicode-org/message-format-wg/issues/49))
```
romulocintra, line 1: In my opinion, the issue number should be placed along the article in a "more info" ,  "links" , "discussions" section or in the intro while giving context.

PR #87, opened 2020-05-19
Doc showing how to use TCQ

### LINKED ISSUE #56: How to use TCQ - Wiki
Is not clear the workflow nor the correct usage of [TCQ.app](https://tcq.app/) in our meetings.

Create a wiki page or short documrnt describing the TCQ usage with and anything that helps members to fluently use the tool

echeran:
In case using the Preview button in the PR UI doesn't make the images show up, you can view the doc directly in the branch [here](https://github.com/echeran/message-format-wg/blob/tcq-howto/docs/tcq-howto.md).

Closes #56


sffc:
@bterlson

romulocintra:
@echeran this exceed my expectations about documenting the TCQ process. Great Job!!! I already see people using this as official TCQ usage docs 😂

sffc:
There are two ways to transition to the next speaker.

1. The host presses the "Next Speaker" button
2. The speaker presses the "I'm Done Speaking" button

Which one should we recommend?

echeran:
I think the first option is useful when you want to ensure a quicker pace for time reasons.  I think the second option might be useful when you want to encourage people to talk longer.

I think it's easier to default to the 1st option -- letting the host press the button.  In this scenario, the host still has to wait a few moments to know when someone is done speaking, and during that time, the speaker can still opt to press the "I'm done speaking" button to make it extra clear.

sffc:
The UX issue I've had here is that sometimes there's a race condition: the speaker thinks they're supposed to press the button, but so does the host, so then they both press it, which causes the next speaker in the queue to be skipped (they got to the top of the queue for a fraction of a second before being unintentionally dismissed by the host).

@bterlson do you have suggestions?

romulocintra:
@echeran should we merge this ?

echeran:
There is still an open review with some outstanding review-related comments, but I wonder if those may be of the non-blocking kind that we address later after merging what we have now.

@sffc What do you think?

sffc:
@bterlson (the author of TCQ) told me on IRC:

> 9:54 AM [the PR] looks good to me
> 9:54 AM there are two bugs that might be worth calling out?
> 9:54 AM first: the redirect still doesn't work, so if you need to log in, better to do that first then click the link
> 9:54 AM or click the link again
> 9:54 AM second, for hosts, better to empty the queue before moving to the next agenda item

PR #112, opened 2020-08-24
Update glossary-and-resources.md

romulocintra:


echeran:
@DavidFatDavidF - after the link you provided to your Localization Reader, what was the link after that to the more detailed teaching version document?

romulocintra:
> @DavidFatDavidF - after the link you provided to your Localization Reader, what was the link after that to the more detailed teaching version document?

romulocintra:
> Approving now in case you want to merge now without the other link. We can add that later if possible & useful.

Just found second link 👍

PR #117, opened 2020-09-24
Merging the 'doc' folder into 'docs'

mihnita:
Unclear why we had both a "doc" and a "docs" folder.


romulocintra (approved):
Thanks for cleanup this

PR #150, opened 2021-02-15
Update ts_data_models_name_mapping.md

zbraniecki:
Adding coalesced Eemeli+Zibi naming scheme


echeran (approved):
LGTM, r+, etc.

Although FYI, you're now mixing types in a table that currently only has entity names (not even field names, at least not yet). We don't want to get too far into the weeds in Thursday's meeting on that so as to save for the monthly plenary mtg, as we said in today's meeting.

PR #154, opened 2021-02-22
Add consensus decisions

### LINKED ISSUE #103: Do we allow multiple multi-select messages to nest inside one another?
This issue is to start (continue) the discussion.  Previous discussions have occurred in working group meetings and more recently https://github.com/zbraniecki/message-format-2.0-rs/issues/6

One option proposed in the above linked issue shows the style of nested multi-select messages that is currently supported in ICU MessageFormat, Fluent, etc.

Example: `10 friends from 2 countries liked her profile.`:

```
{ PLURAL($friendsNum) ->
    [one] { $friendsNum } friend
   *[other] { $friendsNum} friends
} from { PLURAL($countriesNum) ->
    [one] { $countriesNum} country
   *[other] { $countriesNum} countries
} liked { GENDER($user) ->
    [masculine] his
    [feminine] her
   *[other] their
} profile.
```

The issue also links to this issue representing an alternative that has also been discussed during working group meetings: https://github.com/projectfluent/fluent/issues/4 .  Applying the alternative to the above example might look like 

```
{ PLURAL($friendsNum), PLURAL($countriesNum), GENDER($user) ->
    [one, one, masculine]     { $friendsNum } friend from { $countriesNum} country liked his profile.}
    [one, one, feminine]      { $friendsNum } friend from { $countriesNum} country liked her profile.}
    [one, one, other]         { $friendsNum } friend from { $countriesNum} country  liked their profile.}
    [one, other, masculine]   { $friendsNum } friend from { $countriesNum} countries liked his profile.}
    [one, other, feminine]    { $friendsNum } friend from { $countriesNum} countries liked her profile.}
    [one, other, other]       { $friendsNum } friend from { $countriesNum} countries liked their profile.}
    [other, one, masculine]   { $friendsNum} friends from { $countriesNum} country liked his profile.}
    [other, one, feminine]    { $friendsNum} friends from { $countriesNum} country liked her profile.}
    [other, one, other]       { $friendsNum} friends from { $countriesNum} country liked their profile.}
    [other, other, masculine] { $friendsNum} friends from { $countriesNum} countries liked his profile.}
    [other, other, feminine]  { $friendsNum} friends from { $countriesNum} countries liked her profile.}
    [other, other, other]     { $friendsNum} friends from { $countriesNum} countries liked their profile.}
}
```

Other sources, docs, and slide decks have been made for the working group in favor of these options -- feel free to include those here as part of the discussion.

### LINKED ISSUE #127: Nested Selectors Sub-Group proposal
Our exploration of data model ramifications led our Work Group to a strongly visible realization on challenges and tradeoffs around nested vs top level selectors.

The main arguments for *disallowing* nested selectors seem to be around complexity of such feature, and interoperability with existing localization tooling and vendor ecosystems.
The main arguments for *allowing* nested selectors seem to be around flexibility, support for complex scenarios and future-opportunities.

Mihai and I brainstormed how to move forward with recognition of that status.
Mihai wrote about avenues forward here: https://github.com/unicode-org/message-format-wg/issues/103#issuecomment-674572523

From my understanding (2) is what Mihai sides with, while (3) is where I see myself.

From this brainstorm came the following suggestion:

We (Mihai and Zibi) suggest forming an informal sub-group of MessageFormat 2.0 Work Group devoted to the exploration of a Nested Selection data model on top of the work performed by MF2.0 WG.

Such sub-group would be tasked with continuous monitoring of the work the main Work Group is doing for easy inclusion of the nested selectors.

The group would in result validate two hypothesis:

* Algorithmic conversion from nested to top level is possible and doesn’t bring limitations that would make such conversion lossy
* The output of MF2.0 Work Group can be extended to support nested selectors (in other words, no data model, API or syntax decision implicitly blocked nested selectors to be possible to be added)

In particular (2) is crucial here. The sub group would inform the MF2.0 WG on the consequences of the various decisions on ability to add nested selection. If the sub-group were to come to the main Work Group and notify us on such consequence, MF2.0 WG would still be free to make such a decision, but it would be an informed decision, rather than implicit lock-in.

If successful, the result of the sub-group work would be:

* A set of tools, APIs, data model, and syntax that *extends* MF 2.0 output to support nested selection
* Recommendation on auxiliary features and facets of features that would be valuable for nested selectors (meta information, syntax decisions, format algorithms etc.)
* If the WG will over time decide that nested selection is the right call, the sub-group’s work should make such extension smooth
* Alternatively, as the WG work progresses, the sub-group may come to main WG with a conclusion that nested selectors are no longer necessary as all the needs of nested selectors stakeholders have been met with the feature set of MF 2.0 proposal
* Finally, if we release MF2.0 without nested selection, but later the industry were to eventually decide to add nested selection, it would be MF 2.1, rather than MF 3.0 level of change.

In all scenarios, the nested selector sub-group would *inform* the main Work Group and secure the future extension for MF based on the perceived need, and validate hypotheses around it.

In practice, it would free the main Work Group to rely on the sub group to prevent unexpectedly blocking ourselves from being able to add nested selectors, while allowing the work to result in something that can easily be extended to support nested selection if the claim strengthens over time, or not, if the claim weakens.

How does it sound to you?

### LINKED ISSUE #137: External Selectors - Consensus Summary
In order to get consensus for External selectors we need a summarized text for all pros/cons and agreements we are getting the consensus. Having those ideias from last meeting the responsible for this tasks would prepare a concise paragraph/text/bullet points describing the context and decisions.

eemeli:
Resolves #103, resolves #127, resolves #137

As discussed in recent meetings, we have reached consensus on a number of issues, and this ought to be well documented and made more explicit. This PR adds a new document, `docs/consensus_decisions.md`, which enumerates said decisions, along with some clarifying discussion.

The first two were originally documented in the minutes of the [third meeting of the #103 taskforce](https://github.com/unicode-org/message-format-wg/blob/master/meetings/task-force/%23103-2020-10-26.md) and were approved at the [November meeting](https://github.com/unicode-org/message-format-wg/blob/master/meetings/2020/notes-2020-11-16.md) of the workgroup, while the last four are from the discussion on issue #137 (where "3" and "4" are referred to as "1" and "2"), and were last approved at the [February meeting](https://github.com/unicode-org/message-format-wg/blob/meeting-notes-2021-02-15/meetings/2021/notes-2021-02-15.md) of the workgroup.

Even though technically we've already agreed on all of this, it would be useful to communicate your approval about all this on this issue. (Switch to the "Files changed" tab, click on "Review changes", select "Approve", and then "Submit review")

The intent is to get this merged sometime before the next workgroup meeting, which is on 15 March.

### DIFF
--- a/docs/consensus_decisions.md
+++ b/docs/consensus_decisions.md
@@ -65,5 +65,5 @@ The group believes that the known value of this feature can be sufficiently cove
 
 **Discussion:**
 The cost analysis of the nested selectors feature was performed in the absence of sufficient in-field experience of use in production systems.
-In result, the group's decision to reject the feature is based on the lack of sufficient known value that would require them, which the group recognizes may change in the future.
+In result, the group's decision to not currently incorporate the feature is based on the lack of sufficient known value that would require them, which the group recognizes may change in the future.
 In result, it is the intent of the group to attempt to design MessageFormat 2 in such a way, that wouldn't block future revisions of the standard to be extended with nested selectors feature.
--- a/docs/consensus_decisions.md
+++ b/docs/consensus_decisions.md
@@ -1,0 +1,51 @@
+# Consensus Decisions
+
+During its proceedings, the working group has reached internal consensus on a number of issues.
+This document enumerates those, and provides a reference for later actions.
+
+## 1: Include message references in the data model.
+
+**Discussion:**
+The implementers would find a way to include references anyways, but including it in the data model (standard) can make it subject to best practices.
+It’s still possible for users to do “the wrong thing” (ex: concatenation of strings/messages), but then you would find it more difficult to achieve.
+
+One of the drawbacks of message references is that referenced messages effectively have a public API (names of parameters, variables, variants, etc.) which must be consistent across all callsites.
+This leads us to consensus 2.
+
+## 2: Allow parameters passed with message references to the message being referenced and validate it.
+
+**Discussion:**
+The variables/fields passed should not be completely untyped and unchecked.
+We want a validation mechanism that can allow providing early error feedback to the translators & developers.
+We need to decide on when the validation can & should happen, including the meaning of “build time” and “run time” in regards to validation.
+
+## 3: Allow for selectors to select a case depending on the value of one or more input arguments.
+
+**Discussion:**
+This is a prerequisite for top-level selectors to be able to represent complex messages, without requiring those messages to be split up in an unergonomic manner.
+This is an extension or relaxation of what's allowed in MessageFromat 1.
+
+While message references make it technically possible for the data model to represent multi-argument selectors otherwise, this requires the use of n²-1 artificial "messages", where n is the number of arguments. This is not desirable.
+
+## 4: Only allow for selectors at the top level of a message.
+
+**Discussion:**
+Requiring selectors to only be available at the top level is a good way of helping to maintain the translatability of messages, as well as otherwise guiding MessageFormat 2 users towards good practices.
+
+After an in-depth exploration of the problem space, we have determined that while selectors are a necessary feature of MessageFormat, it is not necessary for them to be available within the body of a message, or directly within a case of a parent selector.
+
+All identified use cases of such constructions may be cleanly represented using a top-level selector that may use more than one input argument to select among a set of messages.
+Furthermore, we may enable complete reversibility of message transformations to and from languages such as MessageFormat 1 and Fluent by using message references.
+
+## 5: Top level selectors together with message references provides the same value as nested selectors at a lower cost.
+
+**Discussion:**
+Nested selectors provides capabilities that may be useful in avoiding variant permutation explosion in edge cases, but the use of them has not been evaluated in production localization systems to date.
+The group believes that the known value of this feature can be sufficiently covered by the combination of message references and top level selection features, which together provide a sufficient feature set at a lower cost to the ecosystem than nested selectors would do.
+
+## 6: The group will attempt to avoid blocking addition of nested selectors in the future revisions of the standard.
+
+**Discussion:**
+The cost analysis of the nested selectors feature was performed in the absence of sufficient in-field experience of use in production systems.
+In result, the group's decision to reject the feature is based on the lack of sufficient known value that would require them, which the group recognizes may change in the future.
+In result, it is the intent of the group to attempt to design MessageFormat 2 in such a way, that wouldn't block future revisions of the standard to be extended with nested selectors feature.
--- a/docs/consensus_decisions.md
+++ b/docs/consensus_decisions.md
@@ -30,7 +30,7 @@ It will unfortunately still be possible, but much more difficult, for users to d
 One of the drawbacks of message references is that referenced messages effectively have a public API (names of parameters, variables, variants, etc.) which must be consistent across all callsites.
 This leads us to consensus 2.
 
-## 2: Allow parameters passed with message references to the message being referenced and validate it.
+## 2: Allow message references to include parameters in a form that enables their validation.
 
 **Discussion:**
 The variables/fields passed should not be completely untyped and unchecked.
--- a/docs/consensus_decisions.md
+++ b/docs/consensus_decisions.md
@@ -3,6 +3,21 @@
 During its proceedings, the working group has reached internal consensus on a number of issues.
 This document enumerates those, and provides a reference for later actions.
 
+### Sources
+
+For more details on the process that lead to these decisions, please refer to the following:
+
+- **Consensus 1 & 2:**
+  Reached during the first three meetings of the "Issue #103 task-force", and codified during its [October 2020 meeting](https://github.com/unicode-org/message-format-wg/blob/HEAD/meetings/task-force/%23103-2020-10-26.md).
+  Accepted at the [November 2020 meeting](https://github.com/unicode-org/message-format-wg/blob/HEAD/meetings/2020/notes-2020-11-16.md) of the working group.
+- **Consensus 3 & 4:**
+  Reached in principle during the [December 2020 meeting](https://github.com/unicode-org/message-format-wg/blob/HEAD/meetings/2020/notes-2020-12-14.md) of the working group.
+  Codified in [issue #137](https://github.com/unicode-org/message-format-wg/issues/137).
+  Discussed and accepted at the [January 2021](https://github.com/unicode-org/message-format-wg/issues/146) and [February 2021](https://github.com/unicode-org/message-format-wg/blob/HEAD/meetings/2021/notes-2021-02-15.md) meetings of the working group.
+- **Consensus 5 & 6:**
+  Codified in [issue #137](https://github.com/unicode-org/message-format-wg/issues/137) during the [January 2021 meeting](https://github.com/unicode-org/message-format-wg/issues/146) of the working group.
+  Discussed and accepted at the  [February 2021 meeting](https://github.com/unicode-org/message-format-wg/blob/HEAD/meetings/2021/notes-2021-02-15.md) of the working group.
+
 ## 1: Include message references in the data model.
 
 **Discussion:**
--- a/docs/consensus_decisions.md
+++ b/docs/consensus_decisions.md
@@ -8,13 +8,16 @@ This document enumerates those, and provides a reference for later actions.
 For more details on the process that lead to these decisions, please refer to the following:
 
 - **Consensus 1 & 2:**
-  Reached during the first three meetings of the "Issue #103 task-force", and codified during its [October 2020 meeting](https://github.com/unicode-org/message-format-wg/blob/HEAD/meetings/task-force/%23103-2020-10-26.md).
+  Identified as prerequisites for maintaining backwards-compatibility with MessageFormat 1 once Consensus 3 & 4 are agreed upon.
+  Reached during the meetings of the [issue #103](https://github.com/unicode-org/message-format-wg/issues/103) task-force, and codified during the [October 2020 task-force meeting](https://github.com/unicode-org/message-format-wg/blob/HEAD/meetings/task-force/%23103-2020-10-26.md).
   Accepted at the [November 2020 meeting](https://github.com/unicode-org/message-format-wg/blob/HEAD/meetings/2020/notes-2020-11-16.md) of the working group.
 - **Consensus 3 & 4:**
+  The core result of the [issue #103](https://github.com/unicode-org/message-format-wg/issues/103) task-force ([minutes](https://github.com/unicode-org/message-format-wg/tree/master/meetings/task-force)).
   Reached in principle during the [December 2020 meeting](https://github.com/unicode-org/message-format-wg/blob/HEAD/meetings/2020/notes-2020-12-14.md) of the working group.
   Codified in [issue #137](https://github.com/unicode-org/message-format-wg/issues/137).
   Discussed and accepted at the [January 2021](https://github.com/unicode-org/message-format-wg/issues/146) and [February 2021](https://github.com/unicode-org/message-format-wg/blob/HEAD/meetings/2021/notes-2021-02-15.md) meetings of the working group.
 - **Consensus 5 & 6:**
+  The solution for [issue #127](https://github.com/unicode-org/message-format-wg/issues/127).
   Codified in [issue #137](https://github.com/unicode-org/message-format-wg/issues/137) during the [January 2021 meeting](https://github.com/unicode-org/message-format-wg/issues/146) of the working group.
   Discussed and accepted at the  [February 2021 meeting](https://github.com/unicode-org/message-format-wg/blob/HEAD/meetings/2021/notes-2021-02-15.md) of the working group.
 
--- a/docs/consensus_decisions.md
+++ b/docs/consensus_decisions.md
@@ -61,7 +61,7 @@ Furthermore, we may enable complete reversibility of message transformations to 
 Nested selectors provides capabilities that may be useful in avoiding variant permutation explosion in edge cases, but the use of them has not been evaluated in production localization systems to date.
 The group believes that the known value of this feature can be sufficiently covered by the combination of message references and top level selection features, which together provide a sufficient feature set at a lower cost to the ecosystem than nested selectors would do.
 
-## 6: The group will attempt to avoid blocking addition of nested selectors in the future revisions of the standard.
+## 6: The group does not intend to block the addition of nested selectors in the future revisions of the standard.
 
 **Discussion:**
 The cost analysis of the nested selectors feature was performed in the absence of sufficient in-field experience of use in production systems.
--- a/docs/consensus_decisions.md
+++ b/docs/consensus_decisions.md
@@ -25,7 +25,7 @@ For more details on the process that lead to these decisions, please refer to th
 
 **Discussion:**
 The implementers would find a way to include references anyways, but including it in the data model (standard) can make it subject to best practices.
-It’s still possible for users to do “the wrong thing” (ex: concatenation of strings/messages), but then you would find it more difficult to achieve.
+It will unfortunately still be possible, but much more difficult, for users to do “the wrong thing” by concatenating strings or messages.
 
 One of the drawbacks of message references is that referenced messages effectively have a public API (names of parameters, variables, variants, etc.) which must be consistent across all callsites.
 This leads us to consensus 2.
@@ -34,14 +34,14 @@ This leads us to consensus 2.
 
 **Discussion:**
 The variables/fields passed should not be completely untyped and unchecked.
-We want a validation mechanism that can allow providing early error feedback to the translators & developers.
-We need to decide on when the validation can & should happen, including the meaning of “build time” and “run time” in regards to validation.
+We want a validation mechanism that can allow providing early error feedback to the translators and developers.
+We need to decide on when the validation can and should happen, including the meaning of “build time” and “run time” in regards to validation.
 
 ## 3: Allow for selectors to select a case depending on the value of one or more input arguments.
 
 **Discussion:**
 This is a prerequisite for top-level selectors to be able to represent complex messages, without requiring those messages to be split up in an unergonomic manner.
-This is an extension or relaxation of what's allowed in MessageFromat 1.
+This is an extension or relaxation of what's allowed in MessageFormat 1.
 
 While message references make it technically possible for the data model to represent multi-argument selectors otherwise, this requires the use of n²-1 artificial "messages", where n is the number of arguments. This is not desirable.
 
--- a/docs/consensus_decisions.md
+++ b/docs/consensus_decisions.md
@@ -61,7 +61,7 @@ Furthermore, we may enable complete reversibility of message transformations to 
 Nested selectors provides capabilities that may be useful in avoiding variant permutation explosion in edge cases, but the use of them has not been evaluated in production localization systems to date.
 The group believes that the known value of this feature can be sufficiently covered by the combination of message references and top level selection features, which together provide a sufficient feature set at a lower cost to the ecosystem than nested selectors would do.
 
-## 6: The group does not intend to block the addition of nested selectors in the future revisions of the standard.
+## 6: The model will be designed to leave the door open for nested selectors being potentially reconsidered in the future.
 
 **Discussion:**
 The cost analysis of the nested selectors feature was performed in the absence of sufficient in-field experience of use in production systems.
--- a/docs/consensus_decisions.md
+++ b/docs/consensus_decisions.md
@@ -66,4 +66,4 @@ The group believes that the known value of this feature can be sufficiently cove
 **Discussion:**
 The cost analysis of the nested selectors feature was performed in the absence of sufficient in-field experience of use in production systems.
 In result, the group's decision to not currently incorporate the feature is based on the lack of sufficient known value that would require them, which the group recognizes may change in the future.
-In result, it is the intent of the group to attempt to design MessageFormat 2 in such a way, that wouldn't block future revisions of the standard to be extended with nested selectors feature.
+In result, it is the intent of the group to design MessageFormat 2 in a way that wouldn't prevent future revisions of the standard to be extended with nested selectors feature.


zbraniecki (approved):

sffc (commented):
Could you add more background so that someone coming in and reading this doc for the first time can understand what this means? Issue number references would also help.

docs/consensus_decisions.md
```diff
@@ -0,0 +1,51 @@
...
+## 1: Include message references in the data model.
+
+**Discussion:**
+The implementers would find a way to include references anyways, but including it in the data model (standard) can make it subject to best practices.
+It’s still possible for users to do “the wrong thing” (ex: concatenation of strings/messages), but then you would find it more difficult to achieve.
+
+One of the drawbacks of message references is that referenced messages effectively have a public API (names of parameters, variables, variants, etc.) which must be consistent across all callsites.
+This leads us to consensus 2.
+
+## 2: Allow parameters passed with message references to the message being referenced and validate it.
```
dminor, line 15: This is a little difficult for me to parse. Could we say "Allow referencing and validation of parameters passed with message references to the message"?

docs/consensus_decisions.md
```diff
@@ -0,0 +1,51 @@
...
+All identified use cases of such constructions may be cleanly represented using a top-level selector that may use more than one input argument to select among a set of messages.
+Furthermore, we may enable complete reversibility of message transformations to and from languages such as MessageFormat 1 and Fluent by using message references.
+
+## 5: Top level selectors together with message references provides the same value as nested selectors at a lower cost.
+
+**Discussion:**
+Nested selectors provides capabilities that may be useful in avoiding variant permutation explosion in edge cases, but the use of them has not been evaluated in production localization systems to date.
+The group believes that the known value of this feature can be sufficiently covered by the combination of message references and top level selection features, which together provide a sufficient feature set at a lower cost to the ecosystem than nested selectors would do.
+
+## 6: The group will attempt to avoid blocking addition of nested selectors in the future revisions of the standard.
```
dminor, line 46: I think something like "The group does not intend to block addition of..." would be a little clearer here.

docs/consensus_decisions.md
```diff
@@ -0,0 +1,51 @@
+# Consensus Decisions
+
+During its proceedings, the working group has reached internal consensus on a number of issues.
+This document enumerates those, and provides a reference for later actions.
+
+## 1: Include message references in the data model.
+
+**Discussion:**
+The implementers would find a way to include references anyways, but including it in the data model (standard) can make it subject to best practices.
+It’s still possible for users to do “the wrong thing” (ex: concatenation of strings/messages), but then you would find it more difficult to achieve.
```
dminor, line 10: I suggest "It will unfortunately still be possible, but much more difficult, for users to do “the wrong thing” (ex: concatenation of strings/messages) if references are part of the data model.

docs/consensus_decisions.md
```diff
@@ -0,0 +1,51 @@
...
+**Discussion:**
+The variables/fields passed should not be completely untyped and unchecked.
+We want a validation mechanism that can allow providing early error feedback to the translators & developers.
+We need to decide on when the validation can & should happen, including the meaning of “build time” and “run time” in regards to validation.
+
+## 3: Allow for selectors to select a case depending on the value of one or more input arguments.
+
+**Discussion:**
+This is a prerequisite for top-level selectors to be able to represent complex messages, without requiring those messages to be split up in an unergonomic manner.
+This is an extension or relaxation of what's allowed in MessageFromat 1.
```
dminor, line 26: nit: MessageFormat, not MessageFromat.

docs/consensus_decisions.md
```diff
@@ -0,0 +1,51 @@
...
+It’s still possible for users to do “the wrong thing” (ex: concatenation of strings/messages), but then you would find it more difficult to achieve.
+
+One of the drawbacks of message references is that referenced messages effectively have a public API (names of parameters, variables, variants, etc.) which must be consistent across all callsites.
+This leads us to consensus 2.
+
+## 2: Allow parameters passed with message references to the message being referenced and validate it.
+
+**Discussion:**
+The variables/fields passed should not be completely untyped and unchecked.
+We want a validation mechanism that can allow providing early error feedback to the translators & developers.
```
dminor, line 19: nit: Please spell out "and" rather than using & here and in the following line.

dminor (commented):
Looking at this again with fresh eyes there were a few parts of this that I found difficult to follow. I have some suggestions that might help with readability without changing the substance of the consensus.

eemeli:
@sffc Sure, I'll add a few links.

@dminor Would it be possible for you to reformulate your proposed changes as suggestions? That'd make their inclusion significantly easier, as well as making the requests clearer. If you edit your comments, it's the first icon that has the ± inside a 📄.

docs/consensus_decisions.md
```diff
@@ -0,0 +1,69 @@
...
+We need to decide on when the validation can & should happen, including the meaning of “build time” and “run time” in regards to validation.
```
dminor, line 38:
```suggestion
We need to decide on when the validation can and should happen, including the meaning of “build time” and “run time” in regards to validation.
```

docs/consensus_decisions.md
eemeli: @zbraniecki, would you be comfortable with this change? I'm not entirely sure whether the change in phrasing changes the intent here.

docs/consensus_decisions.md
eemeli:
```suggestion
It will unfortunately still be possible, but much more difficult, for users to do “the wrong thing” by concatenating strings or messages.
```

The "if references are part of the data model" qualifier isn't actually valid, as explicit references are by no means required for bad usage to be possible. With MF1 I've seen plenty of examples with e.g. messages foo_1, foo_2 and foo_3 expected to be combined together after formatting.

docs/consensus_decisions.md
zbraniecki: I don't think it captures the spirit of my intent as well as the original version, since the original version expresses an intent to avoid an action (`we will try to avoid X`), while the revision expresses lack of intent for an opposite action (`we do not intend to do Y` where `Y = !X`), but at this point, I feel like the amount of energy invested in this point is starting to outweigh the original value of this point :)

If Dan feels this is a better sentence, and Eemeli agrees, I'm ok with that.

docs/consensus_decisions.md
eemeli:
```suggestion
## 2: Allow message references to include parameters in a form that enables their validation.
```

Would that be clearer?

docs/consensus_decisions.md
dminor: Sorry, I didn't mean to stir the pot here. If it changes the intent, even slightly, then we should probably go with the original wording.

docs/consensus_decisions.md
dminor: Yes, that looks good to me.

docs/consensus_decisions.md
eemeli: Let's keep the original then.

dminor (approved):

docs/consensus_decisions.md
```diff
@@ -0,0 +1,69 @@
...
+
+**Discussion:**
+Nested selectors provides capabilities that may be useful in avoiding variant permutation explosion in edge cases, but the use of them has not been evaluated in production localization systems to date.
+The group believes that the known value of this feature can be sufficiently covered by the combination of message references and top level selection features, which together provide a sufficient feature set at a lower cost to the ecosystem than nested selectors would do.
+
+## 6: The group will attempt to avoid blocking addition of nested selectors in the future revisions of the standard.
+
+**Discussion:**
+The cost analysis of the nested selectors feature was performed in the absence of sufficient in-field experience of use in production systems.
+In result, the group's decision to reject the feature is based on the lack of sufficient known value that would require them, which the group recognizes may change in the future.
```
echeran, line 68:
```suggestion
In result, the group's decision to not currently incorporate the feature is based on the lack of sufficient known value that would require them, which the group recognizes may change in the future.
```

echeran (approved):
Apart from the one suggested change, everything looks good to me. 

And thanks for finding links to decisions, that is very handy.

sffc (approved):

stasm (commented):
I'm not going to block this PR, but before I approve it, I'd like to ask @zbraniecki to reconsider the wording of consensus 6.

I'd be much more comfortable with something along the lines of @dminor's suggestion, and I think the reason for that is that it's easier for me to agree with a statement about the present (_…does not intend to block the addition…_) than it is to agree with a statement about the future (_…will attempt to avoid blocking addition…_). In the latter, what am I really signing up for? How can we verify in the future that we have attempted hard enough?

zbraniecki:
> I'm not going to block this PR, but before I approve it, I'd like to ask @zbraniecki to reconsider the wording of consensus 6.

I believe the value I'm seeking out of that statement is far outweighed by the amount of energy and brainpower my excellent colleagues devoted to it.
Therefore, I believe that it would be best to adopt a form that makes the largest possible group of people comfortable and avoids making anyone uncomfortable in an attempt to direct this groups energy toward higher priority areas of our work.

I'm comfortable with @dminor's rewording and it seems to help @stasm. Let's do this and advance :)

docs/consensus_decisions.md
eemeli: Ping @stasm, @zbraniecki: Just to be clear, this is the change that's now intended, yes?

zbraniecki: yes.

sffc: I wasn't involved in this discussion, so I don't know the background, and it sounds quite contentious, but it seems that the group decided not to include X in the first version of the standard (where X = nested selectors) on the condition that they would be considered given real world experience. I think it is going a bit far to declare that a future standards body should "avoid" blocking X, since that phrasing seems to imply that the body in 2021 wants to short-circuit debate in the future. So I prefer the new wording.

I would personally remove the double-negativd from the bullet point and say simply: "The group plans to revaluate nested selectors given real world experience."

aphillips: Like @sffc I wasn't involved in this discussion, and am playing catch-up here. I don't want to derail a meaningful consensus that I might agree with. But...

I'm concerned about the statement in question since at Amazon we use nested plural/select statements often and I have "evaluated it in a production localization system" :-). While we discourage deep nesting (more than about 3 levels) and our tools/formats are harder to use with deep nesting (it hasn't been a priority for us to work on it), we nevertheless currently allow an arbitrary level of select/plural nesting. I have used multiple selects to good effect in cases too numerous for me to ignore. So I agree that we should not attempt to block the addition of nested selectors--in fact, I think I'd insist on the ability to do multiple levels.

The challenges of select/plural nesting when applied to string fragments (vs. forcing "complete thought string" repetition) I've enumerated elsewhere, but I am violently opposed to the original wording of item 6. I freely admit this is a hard problem: we should not shy away from hard problems.

aphillips:
> Therefore, I believe that it would be best to adopt a form that makes the largest possible group of people comfortable and avoids making anyone uncomfortable in an attempt to direct this groups energy toward higher priority areas of our work.

@zbraniecki I agree with this statement but would stress that, while rough consensus and running code form the core of good standards (I'm borrowing IETF's koan here on purpose), technical considerations should also inform our choices. One right outlying opinion ought to outweigh 1000 badly formed consensus opinions. (judging "rightness" being, of course, the problem :-))

docs/consensus_decisions.md
```diff
@@ -0,0 +1,69 @@
...
+We want a validation mechanism that can allow providing early error feedback to the translators and developers.
+We need to decide on when the validation can and should happen, including the meaning of “build time” and “run time” in regards to validation.
+
+## 3: Allow for selectors to select a case depending on the value of one or more input arguments.
+
+**Discussion:**
+This is a prerequisite for top-level selectors to be able to represent complex messages, without requiring those messages to be split up in an unergonomic manner.
+This is an extension or relaxation of what's allowed in MessageFormat 1.
+
+While message references make it technically possible for the data model to represent multi-argument selectors otherwise, this requires the use of n²-1 artificial "messages", where n is the number of arguments. This is not desirable.
```
aphillips, line 46: The artificiality/undesirability of the n^2-1 messages is in the eye of the beholder. While it is the case that nested select/plural structures beyond a couple of levels produces a lot of rather repetitious (nearly identical but often subtly different--particularly in a forgiving language like English) messages, this is not necessarily undesirable. When the localization system can participate actively, the burden on translators can be reduced.

I am not arguing for folks to have 10-level nested structures, please note. The size of the data is a problem and consistency management becomes a chore. But I often see teams writing 2-3 levels of nest. The best part is, we can make tools that understand this stuff and make it easier to author and it is something I can teach to developers. If we didn't provide this, developers would go back to their bad old ways of writing switches in code or doing substring replacement (grammatical consistency forgotten).

I don't have an alternate suggestion for wording just now, but this is something I have my eye on.

aphillips: It also occurs to me that I may not understand the term "nested" in this context. For me, this is a "nested" structure (in this case a select in a plural--I copied this from a code review and I'd have written it with the select on the outside, but still...):

```
{
  "value": {
    "param": "unreadCount",
    "pluralItems": {
      "=0": {
        "param": "msgType",
        "selectItems": {
          "email": "Hello {name}, you have no emails in your inbox.",
          "notification": "Hello {name}, you have no notifications in your inbox.",
          "other": "Hello {name}, you have no new items in your inbox."
        }
      },
      "one": {
        "param": "msgType",
        "selectItems": {
          "email": "Hello {name}, you have {unreadCount} email in your inbox.",
          "notification": "Hello {name}, you have {unreadCount} notification in your inbox.",
          "other": "Hello {name}, you have {unreadCount} item in your inbox."
        }
      },
      "other": {
        "param": "msgType",
        "selectItems": {
          "email": "Hello {name}, you have {unreadCount} emails in your inbox.",
          "notification": "Hello {name}, you have {unreadCount} notifications in your inbox.",
          "other": "Hello {name}, you have {unreadCount} item in your inbox."
        }
      }
    }
  }
}
```

Is this what we mean by nested? Or would nested mean the more classical MessageFormat 1 like:

```
Hello {name}, you have {unreadCount,plural, =0 {{msgType,select, email {{unreadCount,number,integer} emails}} one ...//  etc... won't bother to type the whole thing
```

eemeli: The current aim is not to allow for one selector to be inside another selector in the same message, instead enabling the representation of messages such as your example by having that top-level selector be able to use more than one variable (e.g. both `unreadCount` and `msgType`) to select one of the cases. Here's what your message would look like with one of the data model candidates we're developing:

```
{
  id: 'unread-somethings',
  value: {
    select: [
      { func: 'plural', args: [{ var_path: ['unreadCount'] }] },
      { var_path: ['msgType'] }
    ],
    cases: [
      {
        key: [0, 'email'],
        value: ['Hello ', { var_path: ['name'] }, ' you have no emails in your inbox.']
      },
      {
        key: [0, 'notification'],
        value: ['Hello ', { var_path: ['name'] }, ' you have no notifications in your inbox.']
      },
      {
        key: [0, 'other'],
        value: ['Hello ', { var_path: ['name'] }, ' you have no new items in your inbox.']
      },
      {
        key: ['one', 'email'],
        value: ['Hello ', { var_path: ['name'] }, ' you have ', { var_path: ['unreadCount'] }, ' email in your inbox.']
      },
      {
        key: ['one', 'notification'],
        value: ['Hello ', { var_path: ['name'] }, ' you have ', { var_path: ['unreadCount'] }, ' notification in your inbox.']
      },
      {
        key: ['one', 'other'],
        value: ['Hello ', { var_path: ['name'] }, ' you have ', { var_path: ['unreadCount'] }, ' item in your inbox.']
      },
      {
        key: ['other', 'email'],
        value: ['Hello ', { var_path: ['name'] }, ' you have ', { var_path: ['unreadCount'] }, ' emails in your inbox.']
      },
      {
        key: ['other', 'notification'],
        value: ['Hello ', { var_path: ['name'] }, ' you have ', { var_path: ['unreadCount'] }, ' notifications in your inbox.']
      },
      {
        key: ['other', 'other'],
        value: ['Hello ', { var_path: ['name'] }, ' you have ', { var_path: ['unreadCount'] }, ' items in your inbox.']
      }
    ]
  }
}
```

The n^2 -1 reference in the text is for one of the alternative representations of this message, where we avoid using more than one argument for a selector by extracting variants into separate messages, and one of those is then picked by a parent message. Kind of like this:

```
{
  id: 'unread-emails',
  value: {
    select: [{ func: 'plural', args: [{ var_path: ['unreadCount'] }] }],
    cases: [
      {
        key: [0],
        value: ['Hello ', { var_path: ['name'] }, ' you have no emails in your inbox.']
      },
      {
        key: ['one'],
        value: ['Hello ', { var_path: ['name'] }, ' you have ', { var_path: ['unreadCount'] }, ' email in your inbox.']
      },
      {
        key: ['other'],
        value: ['Hello ', { var_path: ['name'] }, ' you have ', { var_path: ['unreadCount'] }, ' emails in your inbox.']
      }
    ]
  }
},

{
  id: 'unread-notifications',
  value: {
    select: [{ func: 'plural', args: [{ var_path: ['unreadCount'] }] }],
    cases: [
      {
        key: [0],
        value: ['Hello ', { var_path: ['name'] }, ' you have no notifications in your inbox.']
      },
      {
        key: ['one'],
        value: ['Hello ', { var_path: ['name'] }, ' you have ', { var_path: ['unreadCount'] }, ' notification in your inbox.']
      },
      {
        key: ['other'],
        value: ['Hello ', { var_path: ['name'] }, ' you have ', { var_path: ['unreadCount'] }, ' notifications in your inbox.']
      }
    ]
  }
},

{
  id: 'unread-items',
  value: {
    select: [{ func: 'plural', args: [{ var_path: ['unreadCount'] }] }],
    cases: [
      {
        key: [0],
        value: ['Hello ', { var_path: ['name'] }, ' you have no new items in your inbox.']
      },
      {
        key: ['one'],
        value: ['Hello ', { var_path: ['name'] }, ' you have ', { var_path: ['unreadCount'] }, ' item in your inbox.']
      },
      {
        key: ['other'],
        value: ['Hello ', { var_path: ['name'] }, ' you have ', { var_path: ['unreadCount'] }, ' items in your inbox.']
      }
    ]
  }
},

{
  id: 'unread-somethings',
  value: {
    select: [{ var_path: ['msgType'] }],
    cases: [
      { key: ['email'], value: [{msg_path: ['unread-emails']}] },
      { key: ['notification'], value: [{msg_path: ['unread-notifications']}] },
      { key: ['other'], value: [{msg_path: ['unread-items']}] }
    ]
  }
}
```

The "actual" message is still the last one, but with only one argument being used for the top-level selectors, the three others are also required.

docs/consensus_decisions.md
zbraniecki: @sffc i feel like we did not commit to reevaluate it as a group. My intention for this point was to give a clue to some future group that will be evaluating it that we did, and the fact that we didn't decide to include it was not because we don't see their value.
I also wanted to put a bit of a constrain on ourselves for the remainder of this work group existence to try not to lock the spec from being able to be extended with this feature.

@stasm seems to be concerned that it is requesting us to "sign up for something" that he doesn't understand. I wish I had energy to solve it by improving understanding of what I'd like us to commit to, but j frankly don't have it.

@aphillips I'm your ally here. I really am. For months. And unfortunately the strongest thing i was able to almost get a consensus on was the original wording.
If you are violently opposing it, may I ask what wording would you suggest?

zbraniecki: May I also point out that it's really sad that a major stakeholder org like Amazon has not been represented in the task force group evaluating nested selectors.
I think something like this should be explicitly stated as a malfunction of our work group.

sffc: @zbraniecki @dminor @eemeli  Thanks, that helps clarify. How about this then: "The group is not ruling out consideration of nested selectors in the future."

(I don't have anything to say to @aphillips; I'm only iterating on the phrasing discussion)

aphillips: @zbraniecki it's more my/Amazon's fault for not participating actively enough or pushing on the WG when the timing of meetings became a problem than the WGs fault.

I think I may have misread the original text, but I tend to favor @sffc's suggestion "The group is not ruling out consideration of nested selectors in the future."

zbraniecki: > I think I may have misread the original text, but I tend to favor @sffc's suggestion "The group is not ruling out consideration of nested selectors in the future."

I see four possible messages we may want to convey:

1) This group would like the reader from the future to know, that it's decision to not include nested selectors was driven by insufficient justification at the time and should not be interpreted as a rejection of the concept.
2) The group may revisit nested selectors in the future.
3) The group conducted a lot of research and brainstorming on nested selectors and the group saw value in the future, but not sufficient to invest in it at the time.
4) The group would like the reader to know that the group would like the reader, if the reader is involved in the work on MF2.0, to put effort in ensuring that the standard will not end up making extension of nested selectors impossible.

My original statement was primarily focused on (4), with some of (3) and (1).
The @dminor's revision switched it to primarily (1) with a bit of implicit (2) and (4).
@sffc's version switches to primarily (2).

I think my aim to achieve (4) is futile, and I do not wish to invest this group's time in trying to recognize the value of my goal.

docs/consensus_decisions.md
aphillips: @eemeli thanks for this. I had followed the discussion of matrixed selectors elsewhere, which is why I hadn't come out of the woodwork with comments earlier. Your top example and mine are just "misspellings" of each other--functionally equivalent while being syntactically different. As long as I was mentally able to map the two functionally, I wasn't concerned. When I saw and paused to read the proposed consensus, though, I went "huh? must have missed something here..."

> This is a prerequisite for top-level selectors to be able to represent complex messages, without requiring those messages to be split up in an unergonomic manner.

I think the more relevant detail here is: _"This allows selectors to represent complex messages while avoiding linguistically problematic constructs that can occur when selectors operate only part of a message"_ (... such as found in MF1)

>This is an extension or relaxation of what's allowed in MessageFormat 1.

I don't find this sentence helpful. This doesn't relax any constraints of MF1 (since it's a separate syntax altogether). I suppose it could be "an extension". But really it's just different from the philosophy of MF1.

> While message references make it technically possible for the data model to represent multi-argument selectors otherwise, this requires the use of n²-1 artificial "messages", where n is the number of arguments. This is not desirable.

The latter example is a logical outgrowth of message references. But I don't see how the proposed text leads here: this should be reserved for a discussion of references, as it doesn't have anything to do with number 3?

docs/consensus_decisions.md
grhoten: I have gone back and forth in my mind for whether internal selectors should or should not be allowed. I found the current wording acceptable for now.

On one side, traditional translators tend to not simplify logic and prefer duplicating a translation. They seem to appreciate seeing the entire sentence in one line.  My anecdotal evidence is that it lowers their cognitive load during translation. It's also easier to detect message changes because when the selector is in the middle, it may be harder to diff the changes when updating a translation of a message.

On the other side, message authors will need to modify more complicated messages. If they need to modify a word or two, the inability to have internal selectors can be rather burdensome because the same word will need to be modified across several selection statements. I know of one message with at least 10 **optional** arguments that we have. Though it's closer to about 20 arguments if you include the non-optional arguments. It's one of the worst messages that our translators ever have to touch. You could say that the author shouldn't have created such a monstrous message. The rebuttal to such an argument is typically that such a message is required, and that the technology should support such things. If it didn't have internal selectors, then they would probably resort to concatenating the string in code to get what they wanted, and that would thus put the translators in a worse position.

So I'm in conflict, and I'm okay with a "we're not saying no" approach for now. If no internal selectors for now allows for us to work on other important things, then I'm okay with that. Allowing us to detect and select grammatical properties of words or conveying context dependencies between selectors is of the utmost importance for my goals.

sffc: @zbraniecki - my intent was to convey mainly points 1 and 3. The existing explanation below the topic sentence explains this pretty clearly. However, I can see that we want to also convey point 4.

I think maybe my confusion arises from the phrase "avoid blocking".  I interpret that double-negative to mean that the *standards body* will avoid blocking, in the **procedural** sense of the word.  However, it appears that this is the wrong sense of the word "block"; do you intend to mean it in the sense that we should not make *design decisions* that would *prevent* a potential future addition of the feature?

How about this: "The model will be designed the door open for nested selectors to be potentially reconsidered in the future."

eemeli: @grhoten As planned, the technology will support even your "monstrous message" via message references. While I'm not advocating that the following is a good idea, it's at the very least possible based on the consensus decisions included in this PR.

This is a third equivalent expression of the message provided by @aphillips in [another thread on this PR](#discussion_r584185434), this time optimised for lack of repetition of content. The syntax is a representation of one of the data model candidates we're developing.

```
{
  id: 'unread-somethings',
  value: [
    'Hello ',
    { var_path: ['name'] },
    ' you have ',
    { msg_path: ['unread-somethings-count'] },
    ' in your inbox.'
  ]
},

{
  id: 'unread-somethings-count',
  value: {
    select: [{ func: 'plural', args: [{ var_path: ['unreadCount'] }] }],
    cases: [
      {
        key: [0],
        value: [' no ', { msg_path: ['something', { var_path: ['msgType'] }, 'plural'] }]
      },
      {
        key: ['one'],
        value: [
          { var_path: ['unreadCount'] },
          ' ',
          { msg_path: ['something', { var_path: ['msgType'] }, 'single'] }
        ]
      },
      {
        key: ['other'],
        value: [
          { var_path: ['unreadCount'] },
          ' ',
          { msg_path: ['something', { var_path: ['msgType'] }, 'plural'] }
        ]
      }
    ]
  }
},

{
  id: 'something',
  entries: [
    {
      id: 'email',
      entries: [
        { id: 'single', value: ['email'] },
        { id: 'plural', value: ['emails'] }
      ]
    },

    {
      id: 'notification',
      entries: [
        { id: 'single', value: ['notification'] },
        { id: 'plural', value: ['notifications'] }
      ]
    },

    {
      id: 'other',
      entries: [
        { id: 'single', value: ['item'] },
        { id: 'plural', value: ['items'] }
      ]
    }
  ]
}
```

zbraniecki: > I think maybe my confusion arises from the phrase "avoid blocking". I interpret that double-negative to mean that the standards body will avoid blocking, in the procedural sense of the word. However, it appears that this is the wrong sense of the word "block"; do you intend to mean it in the sense that we should not make design decisions that would prevent a potential future addition of the feature?

Ah, I can see that confusion now! Thank you for pointing it out.
Yes, my intended meaning of the word "blocking" was technical, not political.

> How about this: "The model will be designed to leave the door open for nested selectors being potentially reconsidered in the future."

That sounds great and is even stronger than my original one. I am happy with a stronger form because I really believe that we should keep the door open.

eemeli:
```suggestion
## 6: The model will be designed to leave the door open for nested selectors being potentially reconsidered in the future.
```

@stasm, would this form suggested by @sffc and seconded by @zbraniecki be acceptable to you?

stasm: Thanks for the additional discussion, everyone. I think it was the word _blocking_ which I wasn't sure about. The new wording is much clearer, thanks!

stasm: And yes, I also agree that the new wording is stronger, and I'm OK with it. My main issue with the previous wording was that I wasn't really sure what it entailed. The expectations of the new wording are easier for me to understand.

sffc: This isn't resolved yet.  It looks like dminor's wording got committed, instead of the new wording I proposed above.

sffc (changes requested):
(setting review bit)

docs/consensus_decisions.md
eemeli: Oh bother, I clicked the wrong button and messed up. Apologies. Fixing now, and then re-marking this thread as resolved.

docs/consensus_decisions.md
```diff
@@ -0,0 +1,69 @@
...
+All identified use cases of such constructions may be cleanly represented using a top-level selector that may use more than one input argument to select among a set of messages.
+Furthermore, we may enable complete reversibility of message transformations to and from languages such as MessageFormat 1 and Fluent by using message references.
+
+## 5: Top level selectors together with message references provides the same value as nested selectors at a lower cost.
+
+**Discussion:**
+Nested selectors provides capabilities that may be useful in avoiding variant permutation explosion in edge cases, but the use of them has not been evaluated in production localization systems to date.
+The group believes that the known value of this feature can be sufficiently covered by the combination of message references and top level selection features, which together provide a sufficient feature set at a lower cost to the ecosystem than nested selectors would do.
+
+## 6: The group does not intend to block the addition of nested selectors in the future revisions of the standard.
```
eemeli, line 64:
```suggestion
## 6: The model will be designed to leave the door open for nested selectors being potentially reconsidered in the future.
```

docs/consensus_decisions.md
```diff
@@ -0,0 +1,69 @@
...
+**Discussion:**
+Nested selectors provides capabilities that may be useful in avoiding variant permutation explosion in edge cases, but the use of them has not been evaluated in production localization systems to date.
+The group believes that the known value of this feature can be sufficiently covered by the combination of message references and top level selection features, which together provide a sufficient feature set at a lower cost to the ecosystem than nested selectors would do.
+
+## 6: The model will be designed to leave the door open for nested selectors being potentially reconsidered in the future.
+
+**Discussion:**
+The cost analysis of the nested selectors feature was performed in the absence of sufficient in-field experience of use in production systems.
+In result, the group's decision to not currently incorporate the feature is based on the lack of sufficient known value that would require them, which the group recognizes may change in the future.
+In result, it is the intent of the group to attempt to design MessageFormat 2 in such a way, that wouldn't block future revisions of the standard to be extended with nested selectors feature.
```
stasm, line 69:
```suggestion
In result, it is the intent of the group to design MessageFormat 2 in a way that wouldn't prevent future revisions of the standard to be extended with nested selectors feature.
```

eemeli: Yeah, this is a better fit for the updated language.

stasm (approved):

docs/consensus_decisions.md
```diff
@@ -66,4 +66,4 @@ The group believes that the known value of this feature can be sufficiently cove
 **Discussion:**
 The cost analysis of the nested selectors feature was performed in the absence of sufficient in-field experience of use in production systems.
 In result, the group's decision to not currently incorporate the feature is based on the lack of sufficient known value that would require them, which the group recognizes may change in the future.
```
aphillips, line 68: Both this paragraph and the following one say "in result". This is also oddly worded. I'd suggest:

> Because the inclusion of nested selectors is not a blocker for the initial release and because the need for nested selectors remains subject to debate, ultimately the group decided not to incorporate the feature into this version of the standard.

romulocintra (approved):

docs/consensus_decisions.md
eemeli: @aphillips Would you be open to having the current text be accepted provisionally in this PR, and then improving the language of Consensus 6 in a separate PR? I'm getting the sense that this one is a bit more divisive than the rest, and that it might be good to have a more focused discussion on its particulars.

aphillips: Go for it, although these suggestions are more about English prosody than content and I think we've about wrung any additional value out of this thread.

sffc (approved):

eemeli:
As promised, merging this before next Monday's meeting, now that it looks like we've reached consensus on this consensus. 🎉

PR #168, opened 2021-04-30
Clarify license for this repository

eemeli:
While working on #167, I noticed that this repository does not appear to provide any license for using any of its contents. This is required in order to be able to properly work with the test suite and any other data.

This was briefly discussed during the [2019 Dec 16 meeting](https://github.com/unicode-org/message-format-wg/blob/6b167e1/meetings/2019/notes-2019-12-16.md#process) by @srl295 and @sffc, from which I understand that the intent is to publish everything under the [Unicode License](http://www.unicode.org/faq/unicode_license.html).

This PR adds a `LICENSE` file to the repo root that clarifies this intent to already apply to its current contents. Note that its first paragraph is custom text, as our test suite and other parts have not yet been published in the places defined in the [Terms of Use](https://www.unicode.org/copyright.html).

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,49 @@
+UNICODE, INC. LICENSE AGREEMENT - DATA FILES AND SOFTWARE
+
+In addition to the definitions provided in the Terms of Use at
+https://www.unicode.org/copyright.html, Unicode Data Files ("Data Files")
+includes all data files and any associated documentation provided in this
+repository, and Unicode Software ("Software") includes all software provided
+in this repository.
+
+NOTICE TO USER: Carefully read the following legal agreement.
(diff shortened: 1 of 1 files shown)


romulocintra (approved):

romulocintra:
Not sure if we should also mention https://github.com/unicode-org/icu4x/blob/478c4a901faca10195edce477b583ed31f8fdd1f/LICENSE#L258.

srl295:
> Not sure if we should also mention https://github.com/unicode-org/icu4x/blob/478c4a901faca10195edce477b583ed31f8fdd1f/LICENSE#L258.

No. Just the Unicode license is fine.

srl295 (approved):
Good catch. FYI @alolita

sffc:
LGTM and thanks for catching this!

PR #170, opened 2021-05-02
Add case selection specification

### LINKED ISSUE #93: Concrete Message Format Use Case to work for next meeting
As we discussed in the last meeting, the group will propose a use case to be worked on for the next meeting.

**Goals**:
 - Allow to have the practical experience and generate tangible output to be discussed next session.
-  Apply the abstractions and comments we had on our MF long journey.
- Review all design principles and design process
- Failure and recovering :) 

**Process**:
- All proposals must be in comments the ones with more thumbs up will be considered in the next Chair Group Meeting to be a candidate for the work described above.

### LINKED ISSUE #99: Design Principle: Allow (or not) functions as data in data model?
In the data model, are we okay with treating functions as first-class data, such that we can include functions as a part of the data model for MF input?  

@zbraniecki 's presentation on the Fluent-based POC showed the use of [references to functions](https://github.com/zbraniecki/message-format-2.0-rs/pull/7/files#diff-6a87164025de4254d2818e3ca4bc813aR31) that can be used for custom placeholder formatting.  @mihnita 's previous presentation on the data model discussed the idea of [custom formatters for placeholders](https://docs.google.com/presentation/d/1KRY8XymAwOE0cUZYSoWmusStRiXHbe1j7PkRzJJDRUQ/edit?ts=5e4dbbf3#slide=id.g6e886633f4_0_0) by indicating where those formatting args could be stored, and this too might require function interfaces to be defined to enable user-defined custom formatters.

I'm in favor of saying that we can treat functions as data, and thus, include them as fields in our data model.  One counter-argument to the idea would be that not all programming languages allow first-class functions.

My perspective is that, for our purposes, declaring the type of a function and declaring a function interface are equivalent (arg types and return type).  Most languages these days have interfaces/traits or first-class fns, and even for the few that don't, they do not need those languages features to create an implementation compatible with the data model.

If we do define behavior as fn interfaces or first-class fns, then it might open up different ways of addressing problems or implementing details.  As a hypothetical example, if we want to generalize the notion of a default selection of a message variant in a multi-select message when placeholder values are absent, instead of relying on the structure of the message variants or adding fields to the data model, we could represent the selection process as a function.  There could be a spec for fn behavior and the rest could be left to implementation details.

### LINKED ISSUE #104: How do we handle and represent selection for a multi-select message?
For a multi-select message, there is the step of selecting a single pattern from the available patterns, followed by the formatting of that selected pattern.  This concerns the first step, which is how do we select, and how do we represent that in a data model?

The topic of how to perform selection came up specifically in https://github.com/zbraniecki/message-format-2.0-rs/issues/9, with the specific case of the uneven branches problem.  It offers alternatives for structuring the data model in ways that can represent a hierarchy/ordering of preference for selection among the patterns.

A different perspective to the problem is to represent selection as a function / interface / trait as suggested in https://github.com/unicode-org/message-format-wg/issues/99.  This requires us to consider defining functions/interfaces either alongside or within a data model, but it potentially decouples selection logic from affecting the shape of input data.

Details of how the algorithmic logic for selection might be defined is described in  https://github.com/unicode-org/message-format-wg/issues/101.  This covers cases where one selector arg is needed and many selector args are needed in the multi-select message.

### LINKED ISSUE #105: How do we support default case selection values in a multi-select message?
In a multi-select message, if the selection logic yields values that do not match up to any of the values/conditions corresponding to any of the patterns in the multi-select message, what do we do?  Do we require a default case to specified, or not?  If we do, how do we specify that default case?

One position in favor for required defaults advocates for a special selection value like like`OTHER` as used in ICU MessageFormat to represent default values, and that there should always be only one such default case across the multi-select message. https://github.com/zbraniecki/message-format-2.0-rs/issues/9#issuecomment-669537897

Another position favor for defaults is https://github.com/unicode-org/message-format-wg/issues/101.  This idea details the selection functions that are needed to be applied to selection args to produce the selection case values, but suggests that default cases can be interpreted when the selection functions return a special value (ex: the empty string `""`).  It says that the use of default values should be avoided / not over-used as a style guide, but we should allow default case checking to happen at runtime, not as a compile time error check.

Another position in favor for defaults is https://github.com/zbraniecki/message-format-2.0-rs/issues/9.  Here, the issue describes alternative, some of which encode defaults using an annotation / data struct field.  That annotation is separate from the case value(s) that the annotation is attached to.

### LINKED ISSUE #106: How do we support multiple selection args (selectors) in a multi-select message?
A multi-select message often has one variable used for selection.  But what if there are two or more?  How do we structure and represent the message accordingly?

This issue is somewhat related to https://github.com/unicode-org/message-format-wg/issues/103, and the example there is applicable here, too.

Several comments that are in favor of "flat" (non-nested) multi-select messages also represent the selection args (selectors) in a list, and likewise represent the selection function return values in a list to key each corresponding pattern in the multi-select message.  Ex: https://github.com/unicode-org/message-format-wg/issues/98#issuecomment-661463626 Ex: https://github.com/projectfluent/fluent/issues/4

One alternate formulation of a flat multi-select message uses a map instead of the prevalent 2 lists formulation (selector names and selector values).  A detailed discussion of the tradeoffs, which resulted in that formulation losing support from its proponent, can be [read here](https://docs.google.com/document/d/19x60Xjc5kA4Scbi8ijBsQpbMK9aubq3uBEvnbYXnbuU/edit).

If #103 results in allowing nested multi-select messages, then that might simplify how this question is understood.  What would that look like, and what are the tradeoffs?

eemeli:
Closes #93
Closes #99
Closes #104
Closes #105
Closes #106

I'm not sure what the plan is on us actually producing spec text, so I figured I'd propose a start to it by adding the `spec/` directory as a place to collect various parts of the specification as we agree on them. As with all our other work, it should be obvious that nothing is final until everything is final, and the contents of this directory may also change as our thinking changes and we take new aspects into consideration.

This PR adds a specification for the handling of case selection. This is an essential part of our [Deliverable 4](https://github.com/unicode-org/message-format-wg/blob/master/guidelines/goals.md#deliverables), "A specification for resolving translations at runtime, including interpolated data types and runtime errors."

A TypeScript implementation of the algorithm is [provided here](https://github.com/messageformat/messageformat/blob/2c97189bf2dad54d3cc0fe0d2c13a2f418b9aaa4/packages/messageformat/src/format-message.ts#L104). Its functioning has been verified against some pretty complex [MF1 & Fluent test cases](https://github.com/messageformat/messageformat/blob/2c97189bf2dad54d3cc0fe0d2c13a2f418b9aaa4/packages/messageformat/src/mf2-features.test.ts#L197).

The relatively high number of issues closed is due to our general silence on actually discussing this topic, at least so far. Previous to this, I'm only aware of #101 as a previously proposed algorithm (which this supersedes). While issue #99 is not directly about selector logic, the proposed algorithm relies on an answer "yes" to its question of allowing function references in the data model; this is supported by both proposed data models.

### Discussion

This selector structure provides a superset of the functionality provided by the `select`, `plural`, and `selectordinal` selectors of MessageFormat 1, the Fluent selector, as well as all other selector methods currently in use.

The algorithm is highly reliant on the case order being correct (see prereq. 7). This is intentional, as it allows for the runtime logic to be relatively simple, and for case selection precedence to be easily deduced.

For a selector function such as `plural`, the order of the keys that it returns is not significant.

It is intentional that a key list may be shorter than its selector list. Such a key list may be considered to have the default value set for each of the not-defined selectors.

The only requirement this algorithm places on the function registry is that it may provide something like a "selector function" matching the above-defined signature. How the registry might be organised or structured and whether this differs somehow from a "formatter function" is explicitly left undefined here.

To use a plural category selector on a numerical value with non-default options and arguments, an explicit function reference may be used. This allows e.g. for ordinal and range plurals to be supported.

This algorithm special-cases numerical values to use a pre-defined default function. It would be possible to extend this logic to also cover e.g. number ranges and gender selectors, but those could easily be implemented using explicit selector functions.

An implementation could define a warning to be emitted if the algorithm reaches step 3. That situation should not be considered as an error, but it could (and possibly should?) be noted at least during development and linting, as it may indicate a missing variant case.

### Open Questions

This algorithm does not define what a "number" is. It should be self-evident that a primitive `number` type would match this definition, but it is conceivable to also consider an object
```ts
{ type: "number", value: number, precision: number }
```
as a "number", and treat it accordingly (with an extension of the matching logic in steps 2.1-2.3).

The benefit of this would be an ability to express a value such as `1.0` as distinct from `1`, which would allow for its correct handling for plural category selection.

### DIFF
--- a/spec/case-selection.md
+++ b/spec/case-selection.md
@@ -1,0 +1,68 @@
+## Variant Case Selection
+
+> This is a fragment of the full MessageFormat 2 specification,
+> intended for later merge into a complete spec document.
+
+### Prerequisites
+
+The following assumptions are made about the **data model** and **function registry**,
+as prerequisites for the following algorithm:
+
+1. The data model supports a message type that allows for selecting
+   one variant **case** of a message based on the value of one or more **selectors**.
+
+2. The value of a selector may be determined by a literal value,
+   a function defined in the function registry,
+   a dynamic variable provided at runtime,
+   the formatted value of another message,
+   or some subset of these.
+   The process for determining the value of any of the above is defined elsewhere.
+
+3. For each selector, a default or **fallback** value is defined.
+   This may be a fixed value such as `"other"`,
+   or a string or integer value defined separately for each selector.
+
+4. In the data model, the selectors are defined by an ordered list.
+
+5. In the data model, the variant cases are defined by an ordered list of entries.
+   In addition to the case's actual message **value**,
+   each entry contains the **key** for the case as an ordered list of string and integer values.
+   These key values are aligned with the selectors.
+   The length of any key list is not greater than the length of the selector list.
+
+6. If the function registry contains a `plural` selector function,
+   that function may be called with one number argument,
+   and be expected to return a list of string and integer values.
+   The returned values correspond to the numerical value of the argument as well as
+   the identifier of the cardinal plural category of the argument.
+
+7. The cases are sorted, such that numerical keys are before string keys and
+   default or fallback key values are after all other key values.
+   If a fallback key has a number value, it is sorted after all non-fallback keys.
+
+### Algorithm
+
+1. The value `s` of each selector is determined:
+
+   1. The initial value `v` of the selector is resolved (see prereq. 2)
+   2. If `v` is a string value or a list of string and integer values, `s = v` is set.
+   3. Else if `v` is a number and a `plural` function is defined, `s = plural(v)` is set.
+   4. Else if `v` is a number, `s = v` is set.
+   5. Else, `v` is forcibly stringified: `s = String(v)`.
+
+   At this point, we have a list `S` of selector values,
+   each of which is a string, number, or a list of a string and integer values.
+
+2. For each case (in order), its key `K` is compared to the selector values.
+   Each string or integer `k` in `K` is compared to its corresponding selector value `s`
+   as well that selector's fallback value `d`:
+
+   1. If `s` is a string or number and `k == s`, the match succeeds.
+   2. Else if `s` if a list of string and integer values and `k` is in `s`, the match succeeds.
+   3. Else if `k == d`, the match succeds.
+   4. Else, the match fails.
+
+   If the match succeeds for all `k` in `K`,
+   the case is selected and its value is selected as the value of the whole message.
+
+3. If no case matches the selector value, an empty string is selected as the message value.


spec/case-selection.md
```diff
@@ -0,0 +1,68 @@
...
+   default or fallback key values are after all other key values.
+   If a fallback key has a number value, it is sorted after all non-fallback keys.
+
+### Algorithm
+
+1. The value `s` of each selector is determined:
+
+   1. The initial value `v` of the selector is resolved (see prereq. 2)
+   2. If `v` is a string value or a list of string and integer values, `s = v` is set.
+   3. Else if `v` is a number and a `plural` function is defined, `s = plural(v)` is set.
```
stasm, line 49: I'd prefer the data model to be more explicit, i.e. the selector should already encode that it will use the `plural` function for resolution.

eemeli: My intent with this is to handle cases where the data model representation does not make it clear whether the input value might or might not be a number. For instance, take the default message used by the [Fluent Playground](https://projectfluent.org/play/):

```
shared-photos =
    {$userName} {$photoCount ->
        [one] added a new photo
       *[other] added {$photoCount} new photos
    } to {$userGender ->
        [male] his stream
        [female] her stream
       *[other] their stream
    }.
```

Specifically, consider the `$photoCount` selector: at what point can it be known whether it represents a numeric value vs. e.g. the literal string `'one'`? The implicit `plural` allows us to ask that question only when we actually have the value; any earlier, we'd need additional input on its data type, or we'd need to add an explicit `plural` wrapper on every single selector.

As I mention in the discussion, the algorithm does of course allow for explicit `plural` wrappers, and that is in fact required when doing anything other than simple cardinal plural selection.

eemeli:
Converted back to draft, as this'll need to be rebased & refactored now that we have a `spec.md` in the repo.

eemeli:
Closing, as the design moved in a slightly different direction.

PR #189, opened 2021-08-18
(third) Implement formatToParts

stasm:
This is a prototype implementation of the `formatToParts` iterator discussed in #41. The approach taken is the one from https://github.com/unicode-org/message-format-wg/issues/41#issuecomment-900198029 by @eemeli with one modification:

- the iterator yields a flat sequence of _parts_,
- a _part_ can either be a  `FormattedPart {type: string, value: string}` or an `OpaquePart`,
- `FormattedPart` doesn't have the `source` field carrying the original raw value; it was easier to do it that way but also it's consistent with existing `formatToParts` specifications,
- patterns coming from message references are flattened into the final sequence.

Examples:

```
let payloadSize = 1.23;
"Transferred {NUMBER $payloadSize STYLE unit UNIT megabyte}."

{ type: 'literal', value: 'Transferred ' },
{ type: 'integer', value: '1' },
{ type: 'decimal', value: '.' },
{ type: 'fraction', value: '23' },
{ type: 'literal', value: ' ' },
{ type: 'unit', value: 'MB' },
{ type: 'literal', value: '.' }
```

```
let submitButton = new SomeUnstringifiableClass();
"Ready? Then {$submitButton}!"

{ type: 'literal', value: 'Ready? Then  ' },
{ type: 'opaque', value: SomeUnstringifiableClass {} },
{ type: 'literal', value: '!' }
```

### DIFF
--- a/experiments/stasm/third/README.md
+++ b/experiments/stasm/third/README.md
@@ -73,6 +73,7 @@ Run each example individually:
     npm run example phrases
     npm run example list
     npm run example number
+    npm run example opaque
 
 ## References
 
--- a/experiments/stasm/third/example/example_glossary.ts
+++ b/experiments/stasm/third/example/example_glossary.ts
@@ -1,6 +1,6 @@
 import {Argument, Message, Parameter} from "../impl/model.js";
 import {REGISTRY} from "../impl/registry.js";
-import {formatMessage, FormattingContext, StringValue} from "../impl/runtime.js";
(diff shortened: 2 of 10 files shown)


stasm:
(Sorry for the notification noise for everyone whom GitHub autoflagged for a review. I accidentally opened this PR against `master` which triggered the code owners reviews.)

PR #194, opened 2021-08-27
Add specification text outline

eemeli:
As a follow-on to discussion #191, this PR adds a bare-bones outline for us to start filling out with the MessageFormat 2.0 specification text.

The document includes an initial "Meta" section that specifies how additions and changes to the document may be proposed and accepted, along with a conflict resolution process. As it makes clear at its start:
> This document is the incomplete working draft of the MF2 specification. It _will_ experience breaking changes, and should not be depended on by anyone for anything. For now, there isn't even an expectation for various parts of the spec to be in agreement with each other.

The process by which the document may develop is intended to itself develop as we advance with this work. Its initial parameters are entirely adjustable or even replaceable if we so desire. The explicit intent of these rules is to enable a system with the following properties:
- The current state of the text is not required or expected to represent the consensus view of the full working group.
- Additions and changes may be made asynchronously of the plenary meetings.
- Individual changes are limited in scope.
- It's easier to have a more limited change land first, and expand it later.
- Adding content to new sections is relatively easy, and cannot be indefinitely blocked.
- Changing existing sections requires a bit more effort, but is still pretty painless.
- No-one is rushed to review changes, and may ask for additional time.
- Blocking progress requires explicit expression of what the concerns are.
- For contentious topics, proposing an alternative is the most effective way of raising the issue for discussion in the plenary meetings.

As an exercise in dogfooding, I'd like to recommend that we apply the proposed "change" rules to this PR, and require three approvals, all specific change requests addressed, and no other current spec framework proposals before landing this.

To submit your review, switch to the "Files changed" tab of this PR, where there's a "Review changes" button in the top right corner. Under that tab you may also comment on individual lines of the PR and even suggest specific changes to them (the ± button in the comment toolbar). Note that you may do so even if you're not a "member" of this repo, but do consider yourself a member of the working group.

As we're looking to eventually publish the spec as a UTS under the CLDR-TC, it makes sense to have it in this repo and format it using markdown, as that should allow us to eventually use the same tooling as is used for UTS#35.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -1,0 +1,227 @@
+# PROPOSED DRAFT: The MessageFormat 2.0 Specification
+
+## META
+
+This document is the incomplete working draft of the MF2 specification.
+It _will_ experience breaking changes,
+and should not be depended on by anyone for anything.
+For now, there isn't even an expectation for various parts of the spec
+to be in agreement with each other.
+
+To **add** content to or **change** a section of this document
+or to **reorganise** sections,
+please file a PR to this repo's `master` branch.
+Each PR should only either add or change one section
+(i.e. content directly under a `#` header or under a single `##` section title),
+or reorganise sections.
+Once a valid PR has been open for at least **two weeks**
+it may be merged if:
+
+- It is approved by:
+  - **one active member** of the WG,
+    if the PR adds a new section or reorganises sections
+  - **three active members** of the WG,
+    if the PR changes the contents of a section.
+- No active members of the WG have reviewed it with "Request changes",
+  or the specific concerns presented in their review have been addressed.
+  Such a block on progress should only apply to the explicit minimal scope of the PR,
+  so wanting it to be expanded should not prevent the smaller PR to be merged first.
+- If an active member of the WG has requested additional time to review the PR,
+  the PR has been open for **four weeks**,
+  or said member has concluded their review.
+- No other similarly approved PR exists with contents for the same section.
+  If the PRs cannot be combined or have all but one withdrawn,
+  the choice will be discussed at the next plenary meeting of the WG.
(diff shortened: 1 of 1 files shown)


macchiati:
We've been working under the assumption that it would be an additional Part
of UTS #35, which is in markdown (Note that we are fixing some of the
formatting for the markdown; right now it doesn't have the right header).

If y'all think it needs to be stand-alone, then we should discuss.

Mark

sffc (approved):
Great, then let's continue drafting in Markdown.  I think we decided to get the text drafted and discuss whether it goes into UTS 35 or into its own spec number once we see what we have.

macchiati:
Makes sense.

Mark

dminor (approved):
I'm happy to see this. I'm hoping it will give us a good way of making progress outside of plenary meetings. I think this format will make it easier for less active participants to keep track of progress and to contribute, which I think has been difficult with the emphasis on verbal, synchronous collaboration that we've had to date.

romulocintra (approved):
@eemeli  thanks for this excellent work !!! 

Should this spec.md be placed inside de docs/ or docs/spec folders?

srl295 (approved):
looks good as a start

eemeli:
@romulocintra I left the file in the repo root as it's sort of _the_ thing we're working on; this should give it the best visibility for anyone searching for it.

Eventually it probably does make sense to move the spec under e.g. `docs/`, but on the way there we probably ought to reorganise the repo otherwise as well; at least `exploration/` and `guidelines/` might make sense to have under `docs/`.

romulocintra:
Merging: I guess we already have enough consensus on this PR, it's a "live" document and will be changed several times. 

Thanks for this great job @eemeli

PR #195, opened 2021-09-11
Add Message Data Model

eemeli:
As a first step in actually filling out the spec, here's my proposal for the Message data model.

This _should_ be relatively uncontroversial, as we've previously agreed that messages may have top-level selectors. This data model is also as simple as possible, and does not e.g. allow for `number` values in selector keys. I'll later submit a PR for a selector algorithm that works with the proposed data model.

The definitions of PatternElement and Meta are left outside the scope of this PR.

As a stylistic choice, I've opted here to use normal text when referring to defined objects (which have a Capital first letter), but format their properties and methods as `code`. The types & interfaces use TypeScript for clarity.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -96,7 +96,60 @@ and eventually have it accepted as a Unicode Technical Standard (UTS).
 
 ## Data Model
 
-> _The data-only representation of a message._
+A Message provides the representation of a single message.
+It takes one of two forms,
+either as a PatternMessage or a SelectMessage.
+In a multi-locale context,
+a single message's choice of representation as one or the other
+should not affect its representation in other languages or locales.
+
+```ts
+type Message = PatternMessage | SelectMessage
+```
+
+A PatternMessage contains a list of PatternElement values,
+some of which are directly defined literal values,
+while others are placeholders with formatted values depending on additional data.
+
+```ts
+interface PatternMessage {
+  type: 'message'
+  value: PatternElement[]
+  meta?: Meta
+}
+```
+
+SelectMessage provides for the selection of one list of PatternElement values
+to use as the message's value when formatting,
+depending on the value of one or more Selector values.
+
+Each SelectCase is defined by a `key` of one or more string identifiers,
+and selection between them is made according to the corresponding Selector values.
+From this it follows that a valid SelectMessage must have at least as many `select` entries
+as its highest count of SelectCase `key` entries within its `cases`.
+
+As a SelectMessage is not a valid PatternElement,
+it is not possible for it to be placed within the `value` of another message.
+
+```ts
+interface SelectMessage {
+  type: 'select'
+  select: Selector[]
+  cases: SelectCase[]
+  meta?: Meta
+}
+
+interface Selector {
+  value: PatternElement
+  fallback?: string
+}
+
+interface SelectCase {
+  key: string[]
+  value: PatternElement[]
+  meta?: Meta
+}
+```
 
 ## Select Case Resolution
 


spec.md
```diff
@@ -96,7 +96,60 @@ and eventually have it accepted as a Unicode Technical Standard (UTS).
...
+  meta?: Meta
+}
+
+interface Selector {
+  value: PatternElement
+  fallback?: string
+}
+
+interface SelectCase {
+  key: string[]
```
zbraniecki, line 148: `suggestion (n-b)`: I believe we should consider select to be `number | string`. Comfortable to land this with a string only and file a spin-off to discuss extending with a number.

spec.md
```diff
@@ -96,7 +96,60 @@ and eventually have it accepted as a Unicode Technical Standard (UTS).
...
+type Message = PatternMessage | SelectMessage
+```
+
+A PatternMessage contains a list of PatternElement values,
+some of which are directly defined literal values,
+while others are placeholders with formatted values depending on additional data.
+
+```ts
+interface PatternMessage {
+  type: 'message'
```
zbraniecki, line 116: `question (n-b)`: why do you place `type` field on `PatternMessage` and `SelectMessage` but not on other nodes?

spec.md
```diff
@@ -96,7 +96,60 @@ and eventually have it accepted as a Unicode Technical Standard (UTS).
 
 ## Data Model
 
-> _The data-only representation of a message._
+A Message provides the representation of a single message.
+It takes one of two forms,
+either as a PatternMessage or a SelectMessage.
+In a multi-locale context,
+a single message's choice of representation as one or the other
+should not affect its representation in other languages or locales.
```
zbraniecki, line 104: `suggestion (n-b)`: I am not convinced that the last sentence is useful.

spec.md
```diff
@@ -96,7 +96,60 @@ and eventually have it accepted as a Unicode Technical Standard (UTS).
...
+to use as the message's value when formatting,
+depending on the value of one or more Selector values.
+
+Each SelectCase is defined by a `key` of one or more string identifiers,
+and selection between them is made according to the corresponding Selector values.
+From this it follows that a valid SelectMessage must have at least as many `select` entries
+as its highest count of SelectCase `key` entries within its `cases`.
+
+As a SelectMessage is not a valid PatternElement,
+it is not possible for it to be placed within the `value` of another message.
```
zbraniecki, line 132: `suggestion`: I don't think this sentence is useful.

spec.md
```diff
@@ -96,7 +96,60 @@ and eventually have it accepted as a Unicode Technical Standard (UTS).
...
+interface SelectMessage {
+  type: 'select'
+  select: Selector[]
+  cases: SelectCase[]
+  meta?: Meta
+}
+
+interface Selector {
+  value: PatternElement
+  fallback?: string
```
zbraniecki, line 144: `suggestion`: can you add a sentence binding `fallback` string to `key` on the appropriate level making valid `SelectMessage` required to have all `fallback` strings be present in the `key` field.

It also is going to be a bit tricky to enforce:

```js
{
  type: 'select',
  select: [
    { value: ..., fallback: "masculine"}
    { value: ..., fallback: "few"}
  ],
  cases: [
    { value: ..., key: ["masculine", "one"] },
    { value: ..., key: ["masculine", "few"] },
    { value: ..., key: ["feminine", "one"] },
    { value: ..., key: ["feminine", "many"] },
  ],
  meta: null
}
```

In this scenario if we resolved first `select` to `feminine` then the default for the second is useless.

I'm wondering if we can do better, for example:



```js
{
  type: 'select',
  select: [
    { value: ..., fallback: "masculine"}
    { value: ..., fallback: "few"}
  ],
  defaults: {
    "masculine": "few",
    "feminine": "many",
  },
  cases: [
    { value: ..., key: ["masculine", "one"] },
    { value: ..., key: ["masculine", "few"] },
    { value: ..., key: ["feminine", "one"] },
    { value: ..., key: ["feminine", "many"] },
  ],
  meta: null
}
```

and validate that?

Or give up on partial fallbacks:


```js
{
  type: 'select',
  select: [
    { value: ..., fallback: "masculine"}
    { value: ..., fallback: "few"}
  ],
  cases: [
    { value: ..., key: ["masculine", "one"] },
    { value: ..., key: ["masculine", "few"], default: true },
    { value: ..., key: ["feminine", "one"] },
    { value: ..., key: ["feminine", "many"] },
  ],
  meta: null
}
```

zbraniecki (changes requested):

spec.md
eemeli: This ends up (among other factors) connecting with the select case resolution algorithm, and how the matching is done. Happy to return to this at that point, and to amend as necessary.

spec.md
eemeli: Because in the resulting structure `PatternMessage` and `SelectMessage` need to be disambiguated from each other, while `Selector` and `SelectCase` are only ever in their assigned slots under `SelectMessage`. There is no situation in which there's uncertainty about the shape of the object found where either of the latter show up.

spec.md
eemeli: I don't it's ready to be submitted as a PR, but [here's my current draft](https://github.com/eemeli/message-format-wg/blob/select-messages/spec.md#select-case-resolution) of the select case resolution algorithm, which may provide relevant info on my views of the above. It probably makes more sense to first define at least the Formattables that it refers to.

In short, I think the `fallback` values _should_ rather than _must_ correspond to the case keys, though if they don't then it becomes more likely that the case selection fails, and we end up with an empty message. Which, to be clear, I think we should be able to handle, but strongly advise against.

And yes, that relationship ought to be made explicit in the spec.

The selection method I'd like to propose relies on case order providing an answer to the fallback vs. exact resolution. If the cases are ordered from more specific to less specific and keys are always checked from left to right, we can iterate through the cases and always select the first one that has keys which match either exactly or the fallback.

dminor: `suggestion (n-b)`: Please consider adding at least a placeholder definition of `fallback`.

spec.md
```diff
@@ -96,7 +96,60 @@ and eventually have it accepted as a Unicode Technical Standard (UTS).
...
+a single message's choice of representation as one or the other
+should not affect its representation in other languages or locales.
+
+```ts
+type Message = PatternMessage | SelectMessage
+```
+
+A PatternMessage contains a list of PatternElement values,
+some of which are directly defined literal values,
+while others are placeholders with formatted values depending on additional data.
```
dminor, line 112:
```suggestion
while others are placeholders with formatted values that depend on additional data.
```

dminor (approved):
Thank you, this seems like a good start to me.

spec.md
```diff
@@ -96,7 +96,60 @@ and eventually have it accepted as a Unicode Technical Standard (UTS).
...
+```
+
+SelectMessage provides for the selection of one list of PatternElement values
+to use as the message's value when formatting,
+depending on the value of one or more Selector values.
+
+Each SelectCase is defined by a `key` of one or more string identifiers,
+and selection between them is made according to the corresponding Selector values.
+From this it follows that a valid SelectMessage must have at least as many `select` entries
+as its highest count of SelectCase `key` entries within its `cases`.
```
eemeli, line 129:
```suggestion
From this it follows that a valid SelectMessage must have at least as many `select` entries
as its highest count of SelectCase `key` entries within its `cases`.
The `fallback` value of a Selector is used in addition to its `value`
when selecting one of the `cases` during formatting.
It should match exactly one of the corresponding SelectCase `key` values.
```

spec.md
eemeli: Added a definition/description of `fallback`.

spec.md
```diff
@@ -96,7 +96,63 @@ and eventually have it accepted as a Unicode Technical Standard (UTS).
...
+The `fallback` value of a Selector is used in addition to its `value`
+when selecting one of the `cases` during formatting.
+It should match exactly one of the corresponding SelectCase `key` values.
```
zbraniecki, line 132: `nit`: `should` is a loaded and ephemeral word. Are we ok if it doesn't? What then? Can it match two? What then?

eemeli: I would welcome a suggestion for clarifying this. The intent here is to indicate that with e.g. your earlier example:
```js
{
  type: 'select',
  select: [
    { value: ..., fallback: "masculine"},
    { value: ..., fallback: "few"}
  ],
  cases: [
    { value: ..., key: ["masculine", "one"] },
    { value: ..., key: ["masculine", "few"] },
    { value: ..., key: ["feminine", "one"] },
    { value: ..., key: ["feminine", "many"] }
  ]
}
```
The value of `select[0].fallback` should be either `"masculine"` or `"feminine"`, while `select[1].value` should to be one of `"one"`, `"few"`, `"many"`.

My preference here would be for the spec to use the term "should" rather than "must", as disobeying the latter would probably mean that we'd need to consider it a parsing error. I would prefer that be avoided, as it would mean that we'd need to build in relatively strong guarantees to rather remote parts of the toolchain to be able to also guarantee that result in its processing.

On the other hand, my preference here isn't absolute by any means, and I would welcome arguments to the contrary.

zbraniecki: I guess my question is - is it a parsing error or not?

And if not, what happens when the condition is not met?

eemeli: My position is that it should not be considered a parsing error, but instead result in all of the following:
- The default/recommended lint/validation rules should complain about it.
- During case selection, if no case matches the given selectors, the result is an empty list of pattern elements along with metadata indicating that case selection failed.
- During formatting, the representation of that implicit empty result may be determined by e.g. a `requireCaseMatch` option given to the formatter: if `false`, this is not an error, and the `formatToString()` result is en empty string. If `true`, this is an error, and some fallback representation of the message is used instead. It should be valid for a formatter to not provide this option, and always follow the `true` logic.

zbraniecki: I'm comfortable with moving forward with this approach for now and I'll likely suggest refinements later.

spec.md
eemeli: Sentence removed.

spec.md
eemeli: Sentence removed.

mihnita (changes requested):

spec.md
```diff
@@ -96,7 +96,63 @@ and eventually have it accepted as a Unicode Technical Standard (UTS).
...
+
+As a SelectMessage is not a valid PatternElement,
+it is not possible for it to be placed within the `value` of another message.
+
+```ts
+interface SelectMessage {
+  type: 'select'
+  select: Selector[]
+  cases: SelectCase[]
```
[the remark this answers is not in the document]
eemeli: Unfortunately that doesn't really work, because the `SelectCase` `key` is a `string[]`, and using an array type as a key would make the data model incompatible with JSON as well most other data representation formats. YAML might be the only one that would work.

spec.md
```diff
@@ -96,7 +96,63 @@ and eventually have it accepted as a Unicode Technical Standard (UTS).
...
+}
+
+interface Selector {
+  value: PatternElement
+  fallback?: string
+}
+
+interface SelectCase {
+  key: string[]
+  value: PatternElement[]
```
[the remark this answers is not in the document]
eemeli: That would be a possible alternative, but what would be the benefit?

zbraniecki (approved):
lgtm!

spec.md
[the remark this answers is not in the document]
mihnita: I understand.

But what about put in the spec what makes most sense from functionality perspective, and worry about syntax / implementation later?
What we need like an "abstract data type" feels like a map, no?
- the number of items in the 2 arrays are the same
- it's not ok to have doubles in the "select" array

It does not even supposed to compile
(although I have this a map in my ts implementation, and compiles, and works)

spec.md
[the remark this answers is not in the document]
mihnita: Semantics.

"This thing here is a simple message, same as the other"

The same reason people `#define` or `typedef` things in C/C++, or use `strong` instead of `bold`
"And array of X" is the same, but is it "the same thing" as a simple message?
Or just an implementation detail?

spec.md
mihnita: There was no response to the RTTI comment.

spec.md
mihnita: Not completely addressed.
My mistake to pile up on the existing one, maybe a bad habit from Google Docs.
I can copy in a different thread if you prefer.

spec.md
eemeli: Apologies, the preceding comment was added after this thread was previously resolved, and therefore hidden. Please unresolve threads if adding comments to them, so that others will note them as well.

I have been working with the assumption that the data model is composed of pure data, and therefore carries explicitly within that data sufficient information to identify each node, such as a `type` property.

If an implementation chooses to discard the `type` value during its internal processing, that is of course a valid option. However, that internal representation would not be universal.

In other words, I don't think the data model should make any assumptions about the capabilities of an implementation, and therefore something like `type` is needed here.

spec.md
eemeli: @mihnita It is by far easiest for all concerned if a single thread here considers a single topic. I shall presume that the relevant part for this thread is the assertion that a selector such as your example may not be represented using the proposed data model. If there's another topic that you would also like to cover, please add a standalone comment for that.

If I understood your pseudocode right, it's a selector on the grammatical case of a variable called `genitive`, with a default value of `'nominative'`, yes? And that there are three cases, of which the third is the fallback. This is how it could be represented with the data model that's proposed here:

```js
{
  type: 'select',
  select: [
    {
      value: {
        type: 'function',
        func: 'grammatical_case',
        args: [
          { type: 'variable', var_path: [{ type: 'literal', value: 'genitive' }] }
        ],
        options: { default: { type: 'literal', value: 'nominative' } }
      },
      fallback: '???'
    }
  ],
  cases: [
    {
      key: ['nominative'],
      value: [
        { type: 'literal', value: "Hello user who's name is " },
        { type: 'variable', var_path: [{ type: 'literal', value: 'userName' }] },
        { type: 'literal', value: '!' }
      ]
    },
    {
      key: ['genitive'],
      value: [
        { type: 'literal', value: "Hello user who's name is " },
        {
          type: 'function',
          func: 'grammatical_case',
          args: [{ type: 'literal', value: 'userName' }],
          options: { grammatical_case: { type: 'literal', value: 'genitive' } }
        },
        { type: 'literal', value: '!' }
      ]
    },
    {
      key: ['???'],
      value: [{ type: 'literal', value: 'Hello!' }]
    }
  ]
}
```

spec.md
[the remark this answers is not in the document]
eemeli: It's not _quite_ the same. Specifically, `SelectMessage` and `SelectCase` may both define a `meta` value that needs to be accounted for e.g. while formatting. This means that the computed `meta` value of a selected case is built from two different sources. The `meta` value of a `PatternMessage` is coming from a single source.

Therefore, keeping the two structures as separate interfaces makes sense, as they need to be handled differently.

spec.md
mihnita: In my understanding the data model is not data, that's why it's called "model"
Different level of abstraction.

A model can very well be UML, with boxes and arrows.

As such if there is a `Foo` type defined then there is already info about its type.
There is no need for a type field, otherwise it is redundant.

> If an implementation chooses to discard the type value during its internal processing

Or we can put it differently: the type should not be there. If an implementation chooses to have one ...


> I don't think the data model should make any assumptions about the capabilities of an implementation

Agree.
For me that also means that it should not add "noise" in the model just because of limitations in some implementation.

spec.md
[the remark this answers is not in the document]
mihnita: I think we are looking at things from a level that is not abstract enough, is too close to (some) implementations.

I think here we should talk at the level of "Abstract Data Types"
https://en.wikipedia.org/wiki/Abstract_data_type

At that level you say "Stack" or "Queue" and people understand what that means (FIFO vs LILO).
You say "Map" (or "Associative Array") that means certain behavior.

We lower the semantic level, and then we need to add it back as documented behavior
"the two arrays should be of equal size" and "no duplicate keys" and "how to fallback  / recover in case of duplicate keys"

It's like saying "a box with chairs and wheels to move people around" instead of Car.

The tendency is always increase the semantic content: we define types, sometimes just aliases to existing ones.
We defines css styles so that we don't deal with collections of attributes.

spec.md
mihnita: > single thread here considers a single topic

It was (is?) a single topic: "binding fallback string to key ..."
It is about how to represent that key: tag it at the top (in select), or on case (like Fluent does with `*`) or with a specially named key (like most programming languages)

----

> variable called genitive, with a default value of 'nominative', yes

No.

It is a fallback case that has nothing to do with any grammatical case.

It's saying _<< a correct Polish translation for "Hello {user}!" requires you to use vocative for the name. If you don't know how to form the vocative for that name at runtime, then fallback to "Hello!" >>_
So there is no user, no grammatical case, it is not a fallback to nominative.

spec.md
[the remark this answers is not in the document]
mihnita: Then I would define a type (MessageBody?) that is used in both places.
Higher level of abstraction.
MessageBody tells us more than an array of `PatternElement`
And it shares behavior, parsing, implementation.

We can say: these are correct ways to represent and handle the `MessageBody` vs describe how to correctly handle two arrays of PatternElement in two levels.

We name things and we rise the level of abstraction.

> may both define a meta value that needs to be accounted for e.g. while formatting. This means that the computed meta value of a selected case is built from two different sources. The meta value of a PatternMessage is coming from a single source

This is a good example of underspecified piece of data that "slips" in the data model without anyone knowing what's inside.

All of this: _"may define"_, the fact that it is possible to have _"computed meta"_ (what does it mean?), that _"coming from a single source"_, that _"meta value that needs to be accounted for e.g. while formatting"_ (so it influence rendering? Why is not in `options`?).

All of these are undocumented assumptions that only exist in one's mind.
Then we approve this, and two months down the line when we open the "Meta" can of worms and find problems the reaction is "sorry, but you approved it in September"

We should not approve undocumented black boxes.

In fact I would propose we remove the Meta for now.
I 100% agree that we need "something like it", but can't just put it the spec completely under the radar and without knowing what that is.

mihnita (changes requested):

spec.md
eemeli: > In my understanding the data model is not data, that's why it's called "model"
> Different level of abstraction.

Ok, we definitely need to talk about this at tomorrow's meeting. If "data model" is not a data-only representation of message data, then I don't know what it is.

> > I don't think the data model should make any assumptions about the capabilities of an implementation
> 
> Agree.
> For me that also means that it should not add "noise" in the model just because of limitations in some implementation.

... but that's literally making an assumption about the capabilities of an implementation, by assuming a specific capability?

spec.md
eemeli: Then I don't think I've understood what your example selector means:
```
grammatical_case(genitive,  default: nominative)
```
That looks like pseudocode for some sort of function. Is that accurate, or is it meant to be something different? What are its inputs, and what does it produce? Is it somehow implicitly getting `user` as an additional parameter?

spec.md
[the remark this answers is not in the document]
eemeli: > Then I would define a type (MessageBody?) that is used in both places.
> Higher level of abstraction.
> MessageBody tells us more than an array of `PatternElement`
> And it shares behavior, parsing, implementation.
> 
> We can say: these are correct ways to represent and handle the `MessageBody` vs describe how to correctly handle two arrays of PatternElement in two levels.
> 
> We name things and we rise the level of abstraction.

Is your request then to add a
```ts
type MessageBody = PatternElement[]
```
and to use that to type the `value` properties of `PatternMessage` and `SelectCase`?

> In fact I would propose we remove the Meta for now.
> I 100% agree that we need "something like it", but can't just put it the spec completely under the radar and without knowing what that is.

I would like to quote a part of the original description of this PR:

> The definitions of PatternElement and Meta are left outside the scope of this PR.

In other words, this PR makes absolutely no assertions about the definition or use of metadata values, other than noting (as you put it) that we need "something like it". Are there specific nodes that you think should not be able to contain metadata?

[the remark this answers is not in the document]
dminor: I also had concerns about leaving `Meta` undefined initially, but I think it is too much to require everything to be fully defined before including it here. We have a process that allows for `Meta` to be changed or removed after it is written, so if we're in agreement that we need something like it, I don't think there is any harm in including it as a placeholder.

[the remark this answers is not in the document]
mihnita: I think that if we agree we "need something like it" but we don't know what it looks like, then we should add an issue, so we don't forget, and put it in the spec later.

But to say "this part of the spec should be that way because of `Meta`, but I don't tell you what `Meta` is or does" feels wrong.
If it affects decisions on the current part of the spec then it kind of has to be there.
Otherwise it is _"just trust me"_

What worries me is that we push this through, and later we discover that the `Meta` as envisioned in one person's mind has limitations, does not do the right things, there are better ways, whatever.
And then the argument will be: we can't change this part of the spec, you agreed with it.

Just live it out.

This is the stated goal of this PR:
> As a first step in actually filling out the spec, here's my proposal for the Message data model.
> 
> This should be relatively uncontroversial, as we've previously agreed that messages may have top-level selectors. This data model is also as simple as possible, and does not e.g. allow for number values in selector keys. I'll later submit a PR for a selector algorithm that works with the proposed data model.

This looks controversial, and was not previously agreed.

So, leave it out?

spec.md
mihnita: In general I am worried about ivory tower specs that completely ignore what is possible / impossible to implement, and end up asking for stuff that is impossible to implement.

But this does not mean that everything in the spec should map 1:1 to existing concepts in all existing programming languages / frameworks.

This is why we should work in terms of interfaces.
We define the type, define what kind of operations are possible to do on it, and assess if various implementations would be able to implement that interface.
We don't say "we will not ask for tries, because C++ does not have tries". Well, C++ will need to implement a trie.

> assumption about the capabilities of an implementation, by assuming a specific capability?

No, I assess the if it is possible or not for an implementation to implement a certain capabilities.
I know (not assume) that there is a way to represent a map with non-string keys as json.

spec.md
mihnita: It is pseudocode, I wanted to represent what you did in that json example.
But I didn't want to use it "as is" because I am not 100% sure what that mean.

Yes, it is a function, and yes, would need arguments.
I tried to leave out things are not relevant to the discussion.

Why would I invent some locale convention, in one message ("???"), instead a "universal" convention?

I understand that this is what Fluent does.
But a generic, clear "ultimate default" is what virtually all programming languages do.
So, they are wrong, and Fluent is right?
And if we want something programmer friendly, as stated many times before, shouldn't we be biased toward well established conventions, unless proven harmful?

In my opinion it has problems, and I explained what those are.

The only pro for the proposed approach is that if the fallback is one of the existing cases there is no need to repeat that message twice (for instance once for nominative, and once for `default`

spec.md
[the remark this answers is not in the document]
mihnita: > Is your request then to add a
> `type MessageBody = PatternElement[]`
> and to use that to type the value properties of PatternMessage and SelectCase?

Yes.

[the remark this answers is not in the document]
mihnita: > this PR makes absolutely no assertions about the definition or use of metadata values

Not the PR directly, but your comment, when you say "can't do X because of `Meta`"
So it is an unspecified behavior of meta that affects decisions in this PR.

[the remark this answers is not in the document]
eemeli: `meta` values dropped, `MessageBody` added.

spec.md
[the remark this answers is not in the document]
eemeli: Transformed into a `Map<string[], PatternMessage>` in the abstract data type definition.

spec.md
eemeli: `type` values dropped from abstract data type definitions.

spec.md
eemeli: > It is pseudocode, I wanted to represent what you did in that json example.
> But I didn't want to use it "as is" because I am not 100% sure what that mean.

Are you perhaps referring to the data model examples that @zbraniecki posted earlier in this thread?

> Yes, it is a function, and yes, would need arguments.
> I tried to leave out things are not relevant to the discussion.

Perhaps these details are relevant, though? If we are to discuss whether a certain type of selection is possible or impossible, it would be good for us to establish exactly what it's doing. If you find it challenging to express in terms of function inputs and results, could you at least attempt to describe it in other words?

> Why would I invent some locale convention, in one message ("???"), instead a "universal" convention?

I wouldn't know, tbh. Clearly my initial read of your pseudocode was mistaken, that's why I'm asking for a clarficiation.

> I understand that this is what Fluent does.
> But a generic, clear "ultimate default" is what virtually all programming languages do.
> So, they are wrong, and Fluent is right?
> And if we want something programmer friendly, as stated many times before, shouldn't we be biased toward well established conventions, unless proven harmful?

Could you clarify how "Fluent is right" in this case? In its case selection, Fluent marks with a `*` one of the select cases to be the "ultimate default". This works well when selecting based on just one input, but it lacks precision when selecting on more than one input. Could it be that when talking of "Fluent" you mean the case selection method that's in the EZ proposal?

For MF2, we have previously reached a consensus decision to "Allow for selectors to select a case depending on the value of one or more input arguments." This is a different kind of selector than established conventions provide, and so we must by necessity find the best possible solution for our novel situation.

To see the need for assigning per-selector fallback values, consider the following scenario: A selector on two arguments, `$foo` and `$bar`. Then, let us presume that these cases are defined for them:
1. `$foo = fooX`, `$bar = barX`
2. `$foo = fooX`, `$bar = other`
3. `$foo = other`, `$bar = barX`
4. `$foo = other`, `$bar = other`

In other words, each of the cases matching `{fooX, other}` with `{barX, other}` are defined.

Now, what happens if we end up at runtime with (a) the values `$foo = fooX` and `$bar = ZZZ`? Or how about (b) the values `$foo = ZZZ` and `$bar = barX`?

If we define the fallback **per case** (as done by MF1, Fluent, and many others) we are effectively forced to select case 4 for both (a) and (b), because the arguments do not match any of the case keys.

If instead we define the fallback **per selector**, the case selection may select case 2 for (a) and case 3 for (b).

This is why the proposed data model includes the fallback per selector, because otherwise we cannot make the best choice in a situation like this.

eemeli:
Updated the proposal based on requests by @mihai, separating the abstract data type definitions of the interfaces from their canonical JSON equivalents as well as applying other requested changes, such as dropping references to metadata objects.

The proposed structure is now hopefully capable of reaching the appropriate level of abstraction, as we define nine types/interfaces instead of only five.

spec.md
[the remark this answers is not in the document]
romulocintra: @eemeli  question : we are duplicating interfaces to keep the JSON canonical representation?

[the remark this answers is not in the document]
eemeli: @romulocintra I would have preferred not needing to define interfaces separately as "abstract data" and their JSON compatible form, but as @mihnita is insisting on using a `Map` rather than an array + dropping the `type` values, this is necessary.

romulocintra (approved):
Great Work!! Unique was what I comment on before about duplicating interfaces to have canonical JSON  representation that I think we can think about in the future.

eemeli:
@mihnita Do you have any remaining concerns about this PR?

mihnita (changes requested):
> @mihnita Do you have any remaining concerns about this PR?

Sorry, yes.
I got hit (again) by the non-intuitive github UI.

I though I submitted the review, but I needed to look at the commit with Firefox and not logged-in to see that they are not actually pushed.
:-(

spec.md
```diff
@@ -96,7 +96,77 @@ and eventually have it accepted as a Unicode Technical Standard (UTS).
...
+while others are placeholders with formatted values that depend on additional data.
+
+```ts
+// Abstract data type
+interface PatternMessage {
+  value: MessageBody
+}
+
+// Canonical JSON representation
+interface PatternMessage {
```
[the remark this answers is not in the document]
eemeli: Yes, that's because it's the canonical JSON representation and not the abstract data type. You requested that they be defined separately.

spec.md
```diff
@@ -96,7 +96,77 @@ and eventually have it accepted as a Unicode Technical Standard (UTS).
...
+  cases: Map<string[], PatternMessage>
+}
+
+interface Selector {
+  value: PatternElement
+  fallback?: string
+}
+
+// Canonical JSON representation
+interface SelectMessage {
```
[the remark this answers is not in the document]
eemeli: Yes, that's because it's the canonical JSON representation and not the abstract data type. You requested that they be defined separately.

spec.md
```diff
@@ -96,7 +96,77 @@ and eventually have it accepted as a Unicode Technical Standard (UTS).
...
+  type: 'select'
```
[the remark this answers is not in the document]
eemeli: Yes, that's because it's the canonical JSON representation and not the abstract data type. You requested that they be defined separately.

spec.md
eemeli: @mihnita Do you understand the specific scenario that I'm presenting in my previous comment, and how that requires a per-selector falback approach?

spec.md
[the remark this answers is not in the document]
mihnita: > Yes, that's because it's the canonical JSON representation and not the abstract data type.

No, I requested to be removed because it is redundant, and we don't care about direct mapping to JSON.

Some programming languages already have info about type (RTTI).
And those that don't will need some kind of "type" field, but it is an implementation detail.
Serialization to/from JSON is possible, but not direct.

But that was and is not required.
This is one of the reasons we decided to go with a data model: to not be restricted by syntax.

Was a decision made 2 years ago:
https://github.com/unicode-org/message-format-wg/blob/main/guidelines/goals.md

And ignoring the goals, there isn't even one issue asking for a **direct** mapping to JSON.

spec.md
mihnita: I think I do.
And I don't think it requires per-selector default.

This can solve the same problem, with more flexibility:
```
switch (selector1(...), selector2(...), selector3(...)) {
    case [c1a, c2a, *] : ...
    case [c1b, *, c3a] : ...
    case [*, c2b, *] : ...
    case [*, *, *] : ...
}
```

Did you read the section about switch in the "whiteboard document"?
["6. Explicit default for selectors"](https://docs.google.com/document/d/1w6fxTfh4xeaqlKZ_QZK1sDpvTv1TqtBeh897aOT8qyk/edit?resourcekey=0-jRKq3bi_UzcnK-5gA8yjsw#heading=h.nkx529xpqnx)
Should I copy-paste it here?

eemeli: You proposal looks very interesting. Could you clarify exactly what the `*` signifies? Does it effectively correspond to the `other` fallback used by MF1, but on a per-selector basis?

> Did you read the section about switch in the "whiteboard document"? ["6. Explicit default for selectors"](https://docs.google.com/document/d/1w6fxTfh4xeaqlKZ_QZK1sDpvTv1TqtBeh897aOT8qyk/edit?resourcekey=0-jRKq3bi_UzcnK-5gA8yjsw#heading=h.nkx529xpqnx) Should I copy-paste it here?

Yes, I've read it, no need to copy-paste.

spec.md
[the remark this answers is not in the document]
eemeli: > > Yes, that's because it's the canonical JSON representation and not the abstract data type.
> 
> No, I requested to be removed because it is redundant, and we don't care about direct mapping to JSON.

To be clear, _you_ Mihai have made the assertion "I don't care about JSON." Others in the working group's discussions have repeatedly noted that having a JSON-compatible representation of the data model would either be beneficial, or even required. In the recording of our last session, such statements are made by @romulocintra and @gibson042 starting from the 1h 38min mark.

> Some programming languages already have info about type (RTTI). And those that don't will need some kind of "type" field, but it is an implementation detail. Serialization to/from JSON is possible, but not direct.
> 
> But that was and is not required. This is one of the reasons we decided to go with a data model: to not be restricted by syntax.

And this is why the data model's "abstract data type" is presented here separately from the canonical JSON representation, so that no such restrictions are imposed on data model operations.

If there is some way in which specifying the canonical JSON representation of the data model hinders work with it, that obviously needs to be taken into account. But as you yourself have agreed that "in implementation there is no difference" between these representations, no such hindrance on operations actually exists; the difference is only in the abstraction level.

So we're in a situation where:
1. You have requested that we separate the two different representations of the data model from each other, and this has now been done.
2. We share an understanding that defining a canonical JSON representation does not reduce our ability to perform operations on the abstract data type representation of the data model.
3. You have stated that you don't care about the JSON representation, while others have stated that they do care about the JSON representation.

Given the above, is the current state of affairs a sufficient compromise? Everyone gets what they want.

spec.md
mihnita: `*` represents a `default`
It is not the same as `other`, which was "hijacked" by plural to mean something.

See description in the thread above:
```
So in the 1:1 with Stas one of the ideas was to have a special "default" case that is not "other", because that was already "hijacked" by plurals.
And that would be "outside any enums", like it is in most programming languages.
```

spec.md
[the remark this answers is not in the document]
romulocintra: It's really hard to follow along this thread, Github UI is not helping and also the number of discussions overlaps and is not easy to understand sometimes 🤯🤯🤯🤯. 

## Spec Document

 I want to mention three things.

1. This is a Draft as a state on the first line of this document
2. It's a live document that will change and be adapted according to our progress.
3. Monday's brainstorm will help clarify and avoid some of this...


## JSON representation

About the JSON canonical representation of DM, I advocate that our data model should be compatible or easily converted to JSON without causing too much trouble or limiting the design of DM. We also agreed a long time ago that is not part of our Goals, but is a must if we can achieve it without "cost".  ✅  Looks like we got it so far.

### Should we care about JSON ?
We have in our deliverables list this point "A specification for a one-to-one mapping between the data model and XLIFF."

I think we should always care about the representations of DM unless they compromise the DM itself.

Now about JSON being present on the spec document, I will use the same analogy, should we have replicated interfaces for all other possible canonical representations? (Calling canonical but maybe not the right word)

TBH I asked about duplication of interfaces days ago because personally, I don't see this at the same level of the Data Model, but I also feel that if we do an effort of having different canonical representations or serialization of this model it must be connected somehow with the base representation as a whole.

We can have annexes, examples, Abstract Operations sections for JSON where those interfaces may appear(idea from ECMA specs) or any other thing that could explain this serialization or interface. 

We can try to research how those cases are solved in other bodies of standards

At moment the actual representation works for me, and while we progress we can come up with something else.

**Conclusion:**
If the unique friction point of this merge is having this duplication of interfaces to represent JSON, I think we should let it go or give a solution or alternative, that in my POV does not include the deletion of those interfaces.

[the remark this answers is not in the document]
romulocintra: Also would like to as @stasm  @DavidFatDavidF  @grhoten  to have a look at this PR you are already reviewers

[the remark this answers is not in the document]
mihnita: 1. I don't know what that means
I did not ask for 2 data models.
You mean that I asked to clarify what happens between parse, functions generating runtime values, and format?
And the fact that it results in 2 data models? (or one model with different states?)

2. Yes, a canonical JSON representation does not reduce out ability to do anything
But in a spec it add un-necessary noise.
It does not belong there, and it introduces "syntax artifacts" in the model.
That is the whole point to have a model: distance ourselves from a syntax, and only put there what is needed.
This is not needed, as I explained (the whole discution about programming languages with RTTI)

3. With 3 you are misrepresenting (again) my position.

I don't care about a DIRECT representation to JSON.
And nobody asked for that.

Nobody prevents a serialization to take this
```
interface PatternMessage {
  value: MessageBody
}
```
and generate a json with type:
```
{
   type: "PatternMessage "
   value: { ... }
}
```

_"while others have stated that they do care about the JSON representation."_
Can you provide a link to an issue that says "support direct serialization to JSON"?
I did give a link to "Goals and Non-Goals", which is one of the relatively few things approved by the group. It does not mention json.

I agree that JSON is nice to have.
Direct mapping to json, if it "distorts" the model, is not nice to have.

spec.md
eemeli: Have I got it right that the logic would work something like this?

In the example you gave, we have three selectors, each with independent values. Let's call these values `s1`, `s2`, and `s3`. They're coming from your `selector#(...)` calls. Let's consider then the selection process and testing the first `case [c1a, c2a, *] : ...`:

If `s1 = c1a, s2 = foo, s3 = bar`, this would not be selected because `foo != c2a`.

With `s1 = c1a, s2 = c2a, s3 = bar`, this would be a valid selection, essentially because `* == x` for any `x`, including `bar`.

Have I understood your proposal right? I.e. that the data model should allow for a SelectCase key list to include a value representing this `*` default/fallback value?

nciric:
This would be a great first PR where we can:
1. Land it soon
2. Define what a design doc is for this group
3. Use the design doc to hammer out the 3 options @mihnita proposed above
4. Amend the final spec with the final solution that came out from design doc

spec.md
eemeli: As requested by @mihnita on today's call, a note about the fallback values still being under discussion has been added to the PR.

spec.md
```diff
@@ -96,7 +96,51 @@ and eventually have it accepted as a Unicode Technical Standard (UTS).
...
+```
+
+SelectMessage provides for the selection of one list of PatternElement values
+to use as the message's value when formatting,
+depending on the value of one or more Selector values.
+
+Each of the SelectMessage `cases` is defined by a key of one or more string identifiers,
+and selection between them is made according to the corresponding Selector values.
+From this it follows that a valid SelectMessage must have at least as many `select` entries
+as its highest count of string entries within the keys of its `cases`.
```
stasm, line 126: _(Non-blocking)_ I'd suggest to drop the `From this it follows...` sentence.  We'll need to figure this out anyways when specifying the case selection logic, and I don't know if such a requirement will be needed.

PR #197, opened 2021-09-22
Add consensus decision on formatting function context

eemeli:
As discussed on Monday's call. For background, see [this presentation](https://docs.google.com/presentation/d/19Q90N6Z9HNF0t_bc2_jtUdYu2PKTPy1oFxuJsxrh2O8/edit?usp=sharing).

### DIFF
--- a/docs/consensus_decisions.md
+++ b/docs/consensus_decisions.md
@@ -20,6 +20,8 @@ For more details on the process that lead to these decisions, please refer to th
   The solution for [issue #127](https://github.com/unicode-org/message-format-wg/issues/127).
   Codified in [issue #137](https://github.com/unicode-org/message-format-wg/issues/137) during the [January 2021 meeting](https://github.com/unicode-org/message-format-wg/issues/146) of the working group.
   Discussed and accepted at the  [February 2021 meeting](https://github.com/unicode-org/message-format-wg/blob/HEAD/meetings/2021/notes-2021-02-15.md) of the working group.
+- **Consensus 7:**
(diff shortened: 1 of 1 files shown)


docs/consensus_decisions.md
```diff
@@ -67,3 +69,19 @@ The group believes that the known value of this feature can be sufficiently cove
...
+It is important to ensure that it is possible for an implementation to consider formatting functions as untrusted code,
+while not requiring that this is done.
+An implementation that does limit formatting function access is still expected to
+allow for e.g. message references and other parts of the specification to work.
+
+The "runtime context" here refers to information about the current message,
+other available messages,
+the available formatting functions,
+and any arguments or parameters that may have been made available when calling the message formatter.
+The formatting function would in all cases still have access to the explicit values and options with which it was called.
```
romulocintra, line 87: I miss some bullet points summarising the first paragraph and what we are agreeing with would be very handy and clarifying, wrote my understanding about the topic, hopefully it can help 

- This consensus doesn't discard that other that future proposed solutions can be used but 

 -  We agree that all security concerns that can arise must be addressed and avoided somehow, always when possible and translatable as design constraints or recommendations from our side.
    - security levels
    - implementation constrains
    - best practices
    - others

- We must ensure that non-malicious code can be used by using one of the proposed alternatives for formatting functions:
   - no access to anything
   - limit access via context or any other interface interface
   -  ~~full access~~

romulocintra (changes requested):
Could you try to give more detailed context in what we are agreeing on as i suggested in comment

PR #198, opened 2021-09-23
Add the Formattable section to the spec

eemeli:
As the proposal's text describes, a `Formattable` is a wrapper around values and their formatting options. It may be thought of as an abstraction that allows us to define the rest of the specification in a manner that's independent of the actual type of any runtime value.

The text briefly refers to MessageFormatParts. The section defining them has not yet been submitted for review, but some relevant discussion may be found in issue #41. Similarly, while PatternElements are also referred to in the preceding PR #195, they are not yet in the spec. Relevant discussion may be found in #190, which also includes some consideration of Formattables.

In the `third` proposal by @stasm, a corresponding abstraction uses the name `RuntimeValue`.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -222,19 +222,101 @@ interface Selector {
 
 ## Formattables
 
-> _Representations of runtime values with associated options, metadata and formatters._
+Formattable is a general interface for values which may be formatted
+to a string or to a sequence of MessageFormatParts,
+such as Messages, literals and runtime variables.
+Its purpose is to allow for each Message and PatternElement formatter to
+encapsulate any value while providing a fixed external interface.
+
+The value and any formatting options must be set during the construction of the Formattable.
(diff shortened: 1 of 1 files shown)


mihnita:
I think we are talking about this in https://github.com/unicode-org/message-format-wg/pull/190 ?
So maybe let's not "fork" the discussion and confuse things?

eemeli:
@mihnita While the discussion in #190 does also cover Formattables, that should not preclude a review of this particular PR.

mihnita (changes requested):
> @mihnita While the discussion in #190 does also cover Formattables, that should not preclude a review of this particular PR.

It does, because it is not clear at all that this concept is even needed.
And even less what it looks like.

eemeli:
Closing as out of date after #307.

PR #203, opened 2021-10-05
Create design-document-template.md

### LINKED ISSUE #202: Definition of Design Document
What are the minimum requirements to have a design document for MFWG?

Please provide comments with points you find important or examples that can be adopted in our WG as a template.

cc @eemeli  @mihnita  @stasm  @zbraniecki

romulocintra:
Proposal for a design doc template 

Create design document closes #202

### DIFF
--- a/guidelines/design-document-template.md
+++ b/guidelines/design-document-template.md
@@ -1,3 +1,10 @@
+<!-- 
+This document intends to provide a simple boilerplate to start a design document, fields aren't strictly required and other templates can be found here:
+- [Chromium Design Doc Template](https://docs.google.com/document/d/14YBYKgk-uSfjfwpKFlp_omgUq5hwMVazy_M965s_1KA/edit)
+- [ICU4X Desing docs](https://github.com/unicode-org/icu4x/tree/main/docs/design)
+- []
+-->
+ 
 # Design Document - My proposal 
 
 ## Overview 
--- a/guidelines/design-document-template.md
+++ b/guidelines/design-document-template.md
@@ -1,8 +1,5 @@
 # Design Document - My proposal 
 
-Author : **myemail@email.com**
-Date : Oct 2020
-
 ## Overview 
 
 ### Motivation
@@ -11,14 +8,9 @@ Motivation to have this design document
 ### Examples &  Prior Art 
 Examples or references in how other solved similar or same problems
 
-```
-Hi,{mf}!
-```
-
-
 ## Design 
 
-Discuss any background and motivation not already in the summary above. Give implementation details and all posible alternatives to solve the problem including affectations to other components.
+Discuss any background and motivation not already in the summary above. Give implementation details and possible alternatives to solve the problem including effects on other components.
 
 ### Code / Implementation / Experiments
 Code examples,  alternative solutions
--- a/guidelines/design-document-template.md
+++ b/guidelines/design-document-template.md
@@ -1,0 +1,35 @@
+# Design Document - My proposal 
+
+Author : **myemail@email.com**
+Date : Oct 2020
+
+## Overview 
+
+### Motivation
+Motivation to have this design document 
+
+### Examples &  Prior Art 
+Examples or references in how other solved similar or same problems
+
+```
+Hi,{mf}!
+```
+
+
+## Design 
+
+Discuss any background and motivation not already in the summary above. Give implementation details and all posible alternatives to solve the problem including affectations to other components.
+
+### Code / Implementation / Experiments
+Code examples,  alternative solutions
+
+### References/Links
+Link to mocks, specs, and related features.
+
+## Follow up
+
+### Consensus 
+List decisions or agreements
+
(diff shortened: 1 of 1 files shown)


mihnita:
I would not try to define here a design document template, unless we want to end with endless debates about that.
I would rather point to a few templates that are "good enough" and list what we really want to get from a design document.

---

Here is some info about the Design Documents at Google:
https://www.industrialempathy.com/posts/design-docs-at-google/

Not because it is perfect, but because it is good enough, and I am familiar with it (vs looking for design documents on the web, reading / understanding / judging what they mean, and if they are good enough.

If others can contribute something they know about, would be great.

---

And note that it is not a "universal recipe" for design documents.

That article say
> That is, given the context (facts), goals and non-goals (requirements), the design doc is the place to suggest solutions and show why a particular solution best satisfies those goals.
>
> The point of writing a document over a more formal medium is to provide the flexibility to express the problem set at hand in an appropriate manner. Because of this, there is no explicit guidance for how to actually describe the design.
>
> Having said that, a few best practices and repeating topics have emerged that make sense for a large percentage of design docs:

We can point to design documents or "deep dive" documents from ICU4X as examples.

---

What I think is critical to be in a design document that is not visible in code:
- alternatives considered & trade-offs
- the design doc is the place to suggest solutions and show **WHY** a particular solution best satisfies those goals

---

There is no need for code in a design document.
In fact I think it is counter-indicated, especially for something cross-platform, cross-programming-language like MF2.

The design document should rather contain pseudo-code to describe how things behave (not implementation),
and maybe public APIs / method signatures / interfaces.

Code can be used as a prof that something can be implemented, can do what it claims to be doing, and help the designed figure out the proper public APIs
But it is not part of the design doc, it only informs the design doc.

mihnita (changes requested):
I don't think we need to create a design doc template, if we don't want to spend weeks on it.

We should point to a few good existing templates, and some examples of design documents.

Because design documents
"They are more what you'd call guidelines than actual rules"
https://imageproxy.ifunny.co/crop:x-20,resize:640x,quality:90x75/images/ef717be29ae4c022fa0fb844dcdf70dfd9fe7ca3bf770262a60ba4c5471b24ef_1.jpg

romulocintra:
Did you check https://github.com/unicode-org/message-format-wg/blob/e335be9a06fe0aea54f9c93bdb5ed4ff262d6308/guidelines/design-document-template.md

mihnita (commented):
**Proposal**

We should:
* point to 2-3 good existing templates
* point to some examples of design documents (ICU4X?)
* list 2-3 things that we really-really want to be part of the design doc, even if everything else is discarded / optional, to not make things too heavyweight

guidelines/design-document-template.md
```diff
@@ -0,0 +1,35 @@
+# Design Document - My proposal 
+
+Author : **myemail@email.com**
+Date : Oct 2020
+
```
eemeli, line 5:
```suggestion
```

We don't need these duplicated in the doc, as they're rather directly available from the edit history.

guidelines/design-document-template.md
```diff
@@ -0,0 +1,35 @@
...
+Motivation to have this design document 
+
+### Examples &  Prior Art 
+Examples or references in how other solved similar or same problems
+
+```
+Hi,{mf}!
+```
+
+
```
eemeli, line 18:
```suggestion
```

This doesn't seem relevant.

guidelines/design-document-template.md
```diff
@@ -0,0 +1,35 @@
...
+## Design 
+
+Discuss any background and motivation not already in the summary above. Give implementation details and all posible alternatives to solve the problem including affectations to other components.
```
eemeli, line 21:
```suggestion
Discuss any background and motivation not already in the summary above. Give implementation details and possible alternatives to solve the problem including effects on other components.
```

Explicitly requiring consideration of _all_ alternatives is rather a lot to ask for.

guidelines/design-document-template.md
```diff
@@ -0,0 +1,35 @@
...
+
```
mihnita, line 22: What needs to be in a design document:

# Goals and non-goals

# Alternatives considered and trade-offs

Since we got to a design document it means there is no one solution that everybody agrees that is perfect.
So there are obviously trade-offs.

This would be the "meat" of the design document, as it explains how the trade-offs are balanced.


## APIs  (probably sub-section for each alternative in "Alternatives considered" 

The APIs section should include signatures.
Should also include explanations on what they do, if not obvious.
Behavior described as pseudo-code.
But behavior is not implementation.

It is also useful to show how the APIs are invoked by the developer, the proper sequence in normal use.

guidelines/design-document-template.md
mihnita: > implementation details

They don't belong in a design document.
Behavior does.

mihnita (changes requested):

PR #230, opened 2022-04-28
Add syntax proposal with EBNF

eemeli:
This PR adds [`spec/syntax.md`](https://github.com/unicode-org/message-format-wg/blob/syntax-proposal/spec/syntax.md) as a starting point for our MF2 grammar considerations. This document is based on the syntax specification prepared earlier by @stasm and represents an evolution of the "A" option presented in #229, taking into account comments and other responses to the presentation.

Given the time constraints, not all comments have been taken into account. We are not suggesting that this is a final syntax, but a sufficiently minimal compromise on which further changes may be applied.

Compared to the "A" option, the following changes have been made:
- All message-internal comments are removed.
- Unicode escapes are removed.
- "Plain" messages are added. These allow for the representation of messages that contain no selectors or placeholders with just the message's translatable contents.

The proposed grammar is LL(1), and a [parser](https://gist.github.com/eemeli/e1048dee25e0a17e6b59192b756059d1) for it is available. That was generated using the [REx Parser Generator](https://bottlecaps.de/rex/), with `-ll 1 -javascript -tree -main` command-line options.

We'd like the discussion about the syntax design to happen here on GitHub rather than in Google Docs comments, for clarity, discoverability and posterity. Our proposal for next steps and onwards progress is the following:

1. Agree that this is a sufficiently acceptable starting point for more specific syntax discussions, and merge this PR.
2. Add issues and/or PRs focusing on specific aspects of the syntax. We're not able to transfer all existing comments from the [MF2 Syntax presentation](https://docs.google.com/presentation/d/1kjHMz3pXEZmxC999_Fri3dj51uOCivlcy_fxwlm1cVk), so we kindly ask you to open new issues here on GitHub with your feedback.

To that end, we ask that you evaluate this proposal as a whole, and hopefully give your approval for merging it so that specific issues and changes may be suggested separately, rather than swamping this initial PR.

As agreed on Monday's call, Staś and I are happy to make ourselves available for in-person discussions before we present this again to the WG on our call on 2022-05-09.

Co-authored-by: Stanisław Małolepszy <sta@malolepszy.org>

### DIFF
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -154,7 +154,7 @@ which the runtime can use to construct a document tree structure for a UI framew
 A message with a single selector:
 
     {$count: number}
-        one [You have one notification.]
+        1 [You have one notification.]
         _ [You have {$count} notifications.]
 
 A message with a single selector which is an invocation of
@@ -173,9 +173,9 @@ which allows the message to query for presence of grammatical cases required for
 A message with 2 selectors:
 
     {$photoCount: number} {$userGender: equals}
-        one masculine [{$userName} added a new photo to his album.]
-        one feminine [{$userName} added a new photo to her album.]
-        one _ [{$userName} added a new photo to their album.]
+        1 masculine [{$userName} added a new photo to his album.]
+        1 feminine [{$userName} added a new photo to her album.]
+        1 _ [{$userName} added a new photo to their album.]
         _ masculine [{$userName} added {$photoCount} photos to his album.]
         _ feminine [{$userName} added {$photoCount} photos to her album.]
         _ _ [{$userName} added {$photoCount} photos to their album.]
@@ -325,7 +325,7 @@ Examples:
 
 ```
 $frac = {$count: number minFractionDigits=2}
-    one [One apple]
+    1 [One apple]
     _ [{$frac} apples]
 ```
 
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -526,7 +526,7 @@ whitespace is part of the translatable content and is recorded and stored verbat
 Whitespace is not significant outside translatable text.
 
 ```ebnf
-WhiteSpace ::= #x9 | #xD | #xA | #x20
+WhiteSpace ::= #x9 | #xD | #xA | #x20 /* ws: definition */
 ```
 
 ## Complete EBNF
@@ -558,9 +558,6 @@ Option ::= Name '=' (String | Nmtoken | Variable)
 /* Markup Tags */
 MarkupStart ::= Name Option*
 MarkupEnd ::= '/' Name
-
-/* Ignored tokens */
-Ignore ::= WhiteSpace /* ws: definition */
 
 <?TOKENS?>
 
@@ -597,5 +594,5 @@ TextEscape ::= Esc Esc | Esc '[' | Esc ']' | Esc '{' | Esc '}'
 StringEscape ::= Esc Esc | Esc '"'
 
 /* WhiteSpace */
-WhiteSpace ::= #x9 | #xD | #xA | #x20
+WhiteSpace ::= #x9 | #xD | #xA | #x20 /* ws: definition */
 ```
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -13,7 +13,6 @@
    1. [Selection](#selection)
    1. [Local Variables](#local-variables)
    1. [Complex Messages](#complex-messages)
-1. [Comparison with ICU MessageFormat 1.0](#comparison-with-icu-messageformat-10)
 1. [Productions](#productions)
    1. [Message](#message)
    1. [Plain](#plain)
@@ -44,10 +43,10 @@ the successor to ICU MessageFormat, henceforth called ICU MessageFormat 1.0.
 
 The design goals of the syntax specification are as follows:
 
-1. The syntax should be an incremental update over the ICU MessageFormat 1.0 syntax
-   in order to leverage the familiarity and the single-message model
-   that is ubiquitous in the localization tooling today,
-   and increase the chance of adoption.
+1. The syntax should leverage the familiarity with ICU MessageFormat 1.0
+   in order to lower the barrier to entry and increase the chance of adoption.
+   At the same time,
+   the syntax should fix the [pain points of ICU MessageFormat 1.0](../docs/why_mf_next.md).
 
    - _Non-Goal_: Be backwards-compatible with the ICU MessageFormat 1.0 syntax.
 
@@ -220,62 +219,6 @@ A complex message with 2 selectors and 3 local variable definitions:
         _ 1 [{$hostName} invites {$guestName} to their party.]
         _ 2 [{$hostName} invites {$guestName} and one other person to their party.]
         _ _ [{$hostName} invites {$guestName} and {$guestsOther} other people to their party.]
-
-## Comparison with ICU MessageFormat 1.0
-
-MessageFormat 2.0 improves upon the ICU MessageFormat 1.0 syntax through the following changes:
-
-1. In MessageFormat 2.0,
-   variants can only be defined at the top level of the message,
-   thus precluding any possible nestedness of expressions.
-
-   ICU MessageFormat 1.0:
-
-   ```
-   {foo, func,
-       foo1 {Value 1},
-       foo2 {
-           {bar, func,
-               bar1 {Value 2a}
-               bar2 {Value 2b}}}}
-   ```
-
-   MessageFormat 2.0:
-
-   ```
-   {$foo: func} {$bar: func}
-       foo1 [Value 1]
-       foo2 bar1 [Value 2a]
-       foo2 bar2 [Value 2b]
-   ```
-
-1. MessageFormat 2.0 differentiates between
-   the syntax used to introduce expressions (`{...}`) and
-   the syntax used to defined translatable content (`[...]`).
-
-1. MessageFormat 2.0 uses the dollar sign (`$`) as the sigil for variable references,
-   the colon (`:`) as the function call syntax,
-   and only allows named options to functions.
-   The purpose of this change is to help disambiguate between
-   the different parts of a placeholder (variable references, function names, literals etc.).
-
-   ICU MessageFormat 1.0:
-
-   ```
-   {when, date, short}
-   ```
-
-   MessageFormat 2.0:
-
-   ```
-   {$when: date style=short}
-   ```
-
-1. MessageFormat 2.0 doesn't provide the `#` shorthand inside variants.
-   Instead it allows local variables to be defined in the preamble,
-   which can then be referred to inside patterns similar to other variables.
-
-1. MessageFormat 2.0 doesn't require commas (`,`) inside placeholders.
 
 ## Productions
 
@@ -531,68 +474,6 @@ WhiteSpace ::= #x9 | #xD | #xA | #x20 /* ws: definition */
 
 ## Complete EBNF
 
-The following EBNF uses the [W3C flavor](https://www.w3.org/TR/xml/#sec-notation) of the BNF notation.
+The complete EBNF is available as [`message.ebnf`](./message.ebnf).
+It uses the [W3C flavor](https://www.w3.org/TR/xml/#sec-notation) of the BNF notation.
 The grammar is an LL(1) grammar without backtracking.
-
-```ebnf
-Message ::= Plain | Pattern | Preamble Variant+
-
-/* Preamble */
-Preamble ::= Selector+
-Selector ::= (Variable '=')? '{' Expression '}'
-
-/* Variants and Patterns */
-Variant ::= VariantKey* Pattern
-VariantKey ::= String | Nmtoken
-Pattern ::= '[' (Text | Placeable)* ']' /* ws: explicit */
-
-/* Placeables */
-Placeable ::= '{' (Expression | MarkupStart | MarkupEnd)? '}'
-
-/* Expressions */
-Expression ::= Operand Annotation? | Annotation
-Operand ::= String | Variable
-Annotation ::= ':' Name Option*
-Option ::= Name '=' (String | Nmtoken | Variable)
-
-/* Markup Tags */
-MarkupStart ::= Name Option*
-MarkupEnd ::= '/' Name
-
-<?TOKENS?>
-
-/* Plain */
-Plain ::= PlainStart (PlainChar* PlainEnd)?  /* ws: explicit */
-PlainChar ::= AnyChar - ('{' | '}')
-PlainStart ::= PlainChar - ('[' | '$' | WhiteSpace)
-PlainEnd ::= PlainChar - WhiteSpace
-
-/* Text */
-Text ::= (TextChar | TextEscape)+
-TextChar ::= AnyChar - ('[' | ']' | '{' | '}' | Esc)
-AnyChar ::= .
-
-/* Names */
-Variable ::= '$' Name /* ws: explicit */
-Name ::= NameStart NameChar* /* ws: explicit */
-Nmtoken ::= NameChar+ /* ws: explicit */
-NameStart ::= [a-zA-Z] | "_"
-            | [#xC0-#xD6] | [#xD8-#xF6] | [#xF8-#x2FF]
-            | [#x370-#x37D] | [#x37F-#x1FFF] | [#x200C-#x200D]
-            | [#x2070-#x218F] | [#x2C00-#x2FEF] | [#x3001-#xD7FF]
-            | [#xF900-#xFDCF] | [#xFDF0-#xFFFD] | [#x10000-#xEFFFF]
-NameChar ::= NameStart | [0-9] | "-" | "." | #xB7
-           | [#x0300-#x036F] | [#x203F-#x2040]
-
-/* Quoted strings */
-String ::= '"' (StringChar | StringEscape)* '"' /* ws: explicit */
-StringChar ::= AnyChar - ('"'| Esc)
-
-/* Escape sequences */
-Esc ::= '\'
-TextEscape ::= Esc Esc | Esc '[' | Esc ']' | Esc '{' | Esc '}'
-StringEscape ::= Esc Esc | Esc '"'
-
-/* WhiteSpace */
-WhiteSpace ::= #x9 | #xD | #xA | #x20 /* ws: definition */
-```
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -1,4 +1,4 @@
-# MessageFormat 2.0 Syntax
+# DRAFT MessageFormat 2.0 Syntax
 
 ## Table of Contents
 
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -1,0 +1,601 @@
+# MessageFormat 2.0 Syntax
+
+## Table of Contents
+
+1. [Introduction](#introduction)
+   1. [Design Goals](#design-goals)
+   1. [Design Restrictions](#design-restrictions)
+1. [Overview & Examples](#overview--examples)
+   1. [Simple Messages](#simple-messages)
+   1. [Simple Placeholders](#simple-placeholders)
+   1. [Formatting Functions](#formatting-functions)
+   1. [Markup Elements](#markup-elements)
+   1. [Selection](#selection)
+   1. [Local Variables](#local-variables)
+   1. [Complex Messages](#complex-messages)
+1. [Comparison with ICU MessageFormat 1.0](#comparison-with-icu-messageformat-10)
+1. [Productions](#productions)
+   1. [Message](#message)
+   1. [Plain](#plain)
+   1. [Preamble](#preamble)
+   1. [Variants](#variants)
+   1. [Patterns](#patterns)
+   1. [Placeables](#placeables)
+   1. [Expressions](#expressions)
+   1. [Markup Elements](#markup-elements)
+1. [Tokens](#tokens)
+   1. [Text](#text)
+   1. [Names](#names)
+   1. [Quoted Strings](#quoted-strings)
+   1. [Escape Sequences](#escape-sequences)
+   1. [Whitespace](#whitespace)
+1. [Complete EBNF](#complete-ebnf)
+
+## Introduction
+
+This document defines the formal grammar describing the syntax of a single message.
+A separate syntax shall be specified to describe collections of messages (_MessageResources_),
+including message identifiers, metadata, comments, groups, etc.
+
+The document is part of the MessageFormat 2.0 specification,
+the successor to ICU MessageFormat, henceforth called ICU MessageFormat 1.0.
+
+### Design Goals
+
+The design goals of the syntax specification are as follows:
+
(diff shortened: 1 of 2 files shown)


stasm:
I’ve compiled a list of topics mentioned in the comments on the slidedeck, and grouped them into themes. Most of these are still open questions, and agreeing to merge this PR doesn't mean they cease to be open questions. The intent of this PR is to have a relatively minimal base that we can then extend (or restrict even further) through a structured review process and discussion. Once this PR is merged, please feel free to open new issues to discuss specific feedback about the syntax

### Escaping 
* What are the general rules of escaping?
* Do we need Unicode escape sequences?
* What happens when an unknown sequences are encountered, e.g. `\a`?
* How do we make sure the need to escape is clear even when a message is stored in a general purpose container?
* Should the sets of escaped characters same inside the `Plain`, `Text`, and `String` productions?

### Markup
* How do ensure non HTML display elements are first class?
* What is the syntax for standalone elements?
* Can we use argument-less functions to represent standalone display elements?
* Do we need a different sigil for display element names?

### Functions
* Are we OK with a function call operator (`{$arg : func}`, whitespace optional), or do we want a sigil attached to function names (`{$arg :func}`)?
* How do we expect `$foo` to differ from `$foo: number` (if at all)?
* Is `{:func}` sufficient syntax for standalone function calls?

### Literals 
* Do we allow bare literals in placeholders, e.g. `{"foo"}`?
* Do literals in the argument position need to be delimited with quotes at all times? E.g. `{"42": number}`.

### Dynamic option values
* Do we want syntax like `{$arg func opt=$variable}`? What are the use-cases?
* How can we make it work such that func is cached/memoized?
* Can `$variable` above be supplied by the application code as an argument to MessageFormat?
* Can `$variable` refer to another placeholder found in the same message?
* Can `$variable` refer to a local variable defined by the message?

### Naming
* Variable / argument / input data.
* Local variable / alias / macro.

### Comments
* Do we want comments inside messages at all?
* Where do we put localization notes? How far away do we accept them to be from the things that they describe?

### Preamble
* If we start in the "code" mode, do we need to delimit the preamble?
* If so, what delimiters to use? `{ … }` or perhaps `{? … }`? Something else?
* Is it OK to make all local variable definitions selectors? I.e. the preamble is a single flat list of selectors, some of which are bound to local variables.

### Variants
* Should variant keys be delimited, too?
* Can there be fewer keys for a given pattern than there are selectors in the preamble?

### Delimiting patterns
* Why use `[ … ]` to delimit patterns rather than `{ … }`?

aphillips:
@stasm I have comments or opinions about nearly everything on your very good list above, but this isn't the place to discuss them. Are you planning to create an issue or issues around them? Or is the idea to just open discussion threads for whatever concerns us?

I'm going to give a positive review here so that we can merge it. I don't see any value in debating the wording before we can look at the doc as an artifact.

eemeli:
@aphillips It would probably be most helpful for any interested parties to open issues about topics that they'd like for the group to discuss. Staś and I may do so as well, but really this would work best as a collaborative effort.

One reason to prefer issues over discussions is that they should more naturally stay single-threaded on a single topic, and it's possible to explicitly close them either directly, or via a pull request. On GitHub, discussions don't have any clear conclusion.

echeran (changes requested):
From a process point of view, the intent of this PR is contrary to how we have done things all along, which is to seek consensus before merging things into the repo, especially if it's decision-making / high-level documents going into the `main` branch, even if it has taken extra time, ex: our [goals & non-goals doc](https://github.com/unicode-org/message-format-wg/blob/main/guidelines/goals.md) and our previous [mini-consensus decisions doc](https://github.com/unicode-org/message-format-wg/blob/main/docs/consensus_decisions.md).  

That alone is concerning, but if it also follows not reciprocating requests from one of the CLDR+ICU committee meeting proposal champions to be included in forming the original design in [the slide deck](https://docs.google.com/presentation/d/1kjHMz3pXEZmxC999_Fri3dj51uOCivlcy_fxwlm1cVk), and if converting that slide deck to this PR effectively drops the continuity of the discussion threads in the deck, then this PR feels premature at best.  We should make sure our processes remain inclusive / fair / consistent.

Steven's comment about marking the title as something non-official sounding is a good catch, and I would prefer the title additionally marked as "PROPOSED" until others have a chance to be included fully.  I think Addison's comment about a plan to address the discussion thread topics is also very important because there are clearly many and diverse questions and concerns raised in the slide deck that this PR doesn't really capture. (bonus points for continuity somehow so that previous commenters don't have to repeat themselves in any new venue(s)).

I don't think that PRs aren't great places for early-stage designs that have a multitude and variety of unresolved design elements.  I also have some questions on topics as well as how certain aspects relate to committee feedback, but I agree that this isn't the best place for that. We can & should address such topics, but I have concerns about merging this PR before that happens.

srl295:
> From a process point of view, the intent of this PR is contrary to how we have done things all along

I'm not really involved in the process, but If this PR doesn't reflect consensus for even an starting point, here's a friendly proposal: keep the `syntax-proposal` branch as just that, one particular syntax proposal branch. Entertain amendments to the proposal there. Maybe process wise it's not the right next step to merge into main.

stasm:
> From a process point of view, the intent of this PR is contrary to how we have done things all along, which is to seek consensus before merging things into the repo, especially if it's decision-making / high-level documents going into the `main` branch, even if it has taken extra time, ex: our [goals & non-goals doc](https://github.com/unicode-org/message-format-wg/blob/main/guidelines/goals.md) and our previous [mini-consensus decisions doc](https://github.com/unicode-org/message-format-wg/blob/main/docs/consensus_decisions.md).

Thanks, @echeran. I think you're right, I'm sorry about that. My intent was to make it easier for everyone to continue the discussion here on GitHub, but I understand your concerns. I opened #231 to discuss a change to the existing process which I think could work well going forward.

> That alone is concerning, but if it also follows not reciprocating requests from one of the CLDR+ICU committee meeting proposal champions to be included in forming the original design in [the slide deck](https://docs.google.com/presentation/d/1kjHMz3pXEZmxC999_Fri3dj51uOCivlcy_fxwlm1cVk), and if converting that slide deck to this PR effectively drops the continuity of the discussion threads in the deck, then this PR feels premature at best. We should make sure our processes remain inclusive / fair / consistent.

The intent was to help preserve continuity by creating something in the repo that others can comment on and open PRs against. My position is that a single PR wouldn't scale well, which is why I'd prefer landing something into the repo and iterating on it going forward.

(I'm OOO until May 4 and my ability to respond may be limited.)

stasm:
Re. the `syntax-proposal` branch proposed by @srl295 -- @eemeli and I considered something similar but we were concerned about the notification noise PRs and issues against such branch would create. That's why in #231 I'm suggesting we do something similar, but more general: a `head` branch for the active development. If the idea of the `head` branch doesn't get much support, I'm in favor of creating a special-purpose `syntax-proposal` branch that this PR could target.

eemeli:
During the meeting yesterday, we agreed to retarget this PR towards a new branch `develop` rather than `main`, as suggested by @srl295 and @stasm.

This will hopefully allow us to soon find consensus to land this, making it easier for further development to take place via issues and further PRs.

echeran (approved):
The biggest process concerns have been addressed, so I'll approve.

nit: I explain in #231 why I don't seen the benefit of creating yet another branch when we already have `experiments` serving the same functionality.  This feels messy <=> another potential instance of discontinuity.

> The intent was to help preserve continuity by creating something in the repo that others can comment on and open PRs against.

Okay. Given that there is a lot of valuable concrete technical discussion already in the comments of the slide deck, I imagine your plan to preserve the continuity of those discussions would be to create PRs for each of the topics listed and copy over those comments? That would SGTM.

markusicu:
Hi there, thanks for writing down a complete proposal. However, the more I look at it, the more I find aspects of the syntax confusing and unintuitive.

First, to a human, there is no clear mental model for how to start reading and writing a message. A message can just be plain text (yeah!), or text in square brackets, or "code". I see how you get that to work in the EBNF, but it's confusing.

I have always liked simple messages to be simple, with no surrounding syntax. However, that should include placeholders, requiring enclosing-delimiter syntax only when there are placeholders is weird, and allowing simple text both with and without delimiters is also weird.

Starting in "text" mode also has problems discussed here (uncertainty about trimming spaces), and we have problems with MF1 (ICU MessageFormat) where we don't want people to write literal text outside of a select-type placeholder.

So while it pains me to say this, bite the bullet, start in "code" mode, and always surround translatable text with delimiters. And *please* use curly braces for those delimiters; square brackets are common enough in real text. Together with using curly braces for placeholders, only `{}` are special and need escaping.

So like this:
```
{Hello world!}

{Hello {$name}!}
```

The next most confusing thing is figuring out which placeholder-looking item becomes an input to the selection, and how those match up to the variant value tuples. You really need to enclose the selectors so that it's clear which ones they are, and the number of selectors needs to be the same as the number of values in each value tuple. I suggest enclosing both the list of selectors and each value tuple in square brackets. Bracketing both selectors and tuples correlates them visually.

Also, it seems like a variable definition can have a side effect of its expression leaking into the selection syntax; don't overload unrelated functionalities.

With that, a message always starts in "code" mode with optional variable definitions; then it's either a single pattern, or a selector head followed by pairs of value tuples and patterns:
```
$f1={$something}  $f2={$else}
[{$count :plural offset=1} {$gender}]
[few female] {pattern {$1}}
[_ _] {the default pattern}
```

About the colon as a function prefix -- I am ok with using a colon, but it's a prefix, just like the dollar for variables, so it needs to attach to the function name, and there should be no space allow between the colon and the name. If you don't like the visuals, then pick a different symbol. (Earlier proposals used an `@` sign.)

Please don't allow whitespace everywhere. Traditional programming languages are super loose with spaces, but then people debate and enforce style guides for where to put them. Looking at my example above, I suggest not allowing spaces where I didn't put any. (For example, not around the `=`.)

The syntax so far is not clear about formatting functions and selection functions. It looks like "number" could be both, but they should be separate. "number" should just format, and "plural" should both format and select (remember that the two are intertwined).

Similarly, once a variable has been subject to select-and-format, *using* it should require different syntax from regular variables, because we need to be clear that the placeholder is replaced consistently in each variant. If you just write `{$count}` then a developer has to wonder whether their formatting options apply; and you cannot allow formatting options on a select-and-format variable in the variant, because then the output may well not fit the selected variant any more (e.g., changing the number of fraction digits on a plural). So pick a symbol other than the dollar, and don't allow functions and options there.

About the markup syntax: Are these free-form, implementation-dependent tokens? Do you expect them to be HTML or TTS hints or accessibility hints? Without any distinctions, this seems like [Unicode private use code points](https://www.unicode.org/faq/private_use.html) with all of their problems.

Or are the markup tokens registered functions? If so, then why not use the `{:function}` syntax, and maybe with a literal string before the function name (which could then be something like `:html`).

If we do need markup with special syntax, then it should fit better into the overall syntax (e.g., always an ASCII symbol after the opening `{`) and into the framework as a whole (e.g., registered entities).

Speaking of literals, don't enclose them in quotes. One of the stated goals here is to make messages usable as string literals in programming languages. Enclose literals in parentheses `{... {(5) :number} books}` or angle brackets `{... {<5> :number} books}` or pairs of pipe symbols `{... {|5| :number} books}` or similar.

Do we need variables inside placeholder options? We have seen this before, but I don't remember a good use case being presented. I see one example that suggests using it for grammatical agreement, but I strongly suspect that managing grammatical agreement will require a second pass over the output of the whole "formatting" pass for a message; a second pass so that it can look at everything in context -- rather than trying to make one isolated placeholder agree with the input to another. So unless we have a really good use case for variables in options, let's leave them out.

Phew, this turned out long, sorry :-)

Cc @macchiato

stasm:
Hi @markusicu, thank you for writing this down! I agree with a lot of your points, and with some others I agree but I picked different trade-offs. We should discuss all of them.

The idea behind this PR is to merge a starting point into the `develop` branch precisely so that we have a base against which we can file more issues. I see a material for a new issue in almost every paragraph of your comment:

- Start in "code" mode.
- Enclose selectors so that it's clear what they are.
- Don't overload selector syntax to define local variables.
- Use colon as a prefix, not an operator.
- Don't allow whitespace everywhere.
- Make it clear which functions are formatting functions and which are selection functions.
- Use a different namespace to reference local variables.
- Do markup elements map to registered functions?
- Pick a delimiter for literals other than the double quote.
- Do we need variables as option values?

Would you mind opening a new issues for each of these and tagging them with the `syntax` label? I can do this for you, too, just let me know.

romulocintra (approved):

romulocintra:
- [ ]  Waiting for @mihnita comments and approval to merge this PR

spec/syntax.md
```diff
@@ -0,0 +1,598 @@
...
+including message identifiers, metadata, comments, groups, etc.
+
+The document is part of the MessageFormat 2.0 specification,
+the successor to ICU MessageFormat, henceforth called ICU MessageFormat 1.0.
+
+### Design Goals
+
+The design goals of the syntax specification are as follows:
+
+1. The syntax should be an incremental update over the ICU MessageFormat 1.0 syntax
```
mihnita, line 47: The single message model in l10n has nothing to do with MF1, or with localization tooling.

This seems to conflate several independent issues:
- Translate 1 message at the time, independent of other messages. This is mostly dictated by localization frameworks. XLIFF and many localization tools have support for message grouping and references.
- Translate 1 source message to 1 target message (1:1 translation). This is a limitation in many localization tools, and currently gets in the way of proper support for plural (in MessageFormat, gettext, Android, others)
- The proposed syntax does not seem like an incremental update to MF1. In fact there are areas where it is clearly not functionally backward compatible (for examples select using "first match" instead of best match).

spec/syntax.md
```diff
@@ -0,0 +1,598 @@
...
+which the runtime can use to construct a document tree structure for a UI framework.
+
+    [{button}Submit{/button} or {link}cancel{/link}.]
+
+### Selection
+
+A message with a single selector:
+
+    {$count: number}
+        one [You have one notification.]
```
mihnita, line 157: All examples using `one` are incorrect, should use `1`

They would be OK if the message would contain the `{$count}` to be formatted.

Imagine using `21` instead of `1`.
Showing `"You have 21 notification"` is fine (in Russian :-).
Showing `"You have one notification"` when the count is 21 is incorrect.

spec/syntax.md
```diff
@@ -0,0 +1,598 @@
...
+  are translatable and which ones are part of the formatting logic definition.
+
+```ebnf
+Pattern ::= '[' (Text | Placeable)* ']' /* ws: explicit */
+```
+
+Examples:
+
+```
+[Hello, world!]
```
mihnita, line 371: The current syntax (intentionally) allows for simple messages without `[`, `]` (no selectors, no placeholders)
So this is not accurate.

In fact, the current syntax allows both `Hello world!` and `[Hello world!]`, which I don't think is a feature.

But for this PR I'm not arguing that, only that the examples here should reflect the current grammar.

spec/syntax.md
```diff
@@ -0,0 +1,598 @@
...
+        male 1 [{$hostName} invites {$guestName} to his party.]
+        male 2 [{$hostName} invites {$guestName} and one other person to his party.]
+        male _ [{$hostName} invites {$guestName} and {$guestsOther} other people to his party.]
+
+        _ 0 [{$hostName} does not give a party.]
+        _ 1 [{$hostName} invites {$guestName} to their party.]
+        _ 2 [{$hostName} invites {$guestName} and one other person to their party.]
+        _ _ [{$hostName} invites {$guestName} and {$guestsOther} other people to their party.]
+
+## Comparison with ICU MessageFormat 1.0
```
mihnita, line 224: I think we can remove this whole section.

The differences should be visible from the grammar and examples.

Or at least add this section later, once the syntax is a bit more "established".
Otherwise we would have to keep updating this section every time we change the syntax.

spec/syntax.md
```diff
@@ -0,0 +1,598 @@
...
+
+1. MessageFormat 2.0 doesn't provide the `#` shorthand inside variants.
+   Instead it allows local variables to be defined in the preamble,
+   which can then be referred to inside patterns similar to other variables.
+
+1. MessageFormat 2.0 doesn't require commas (`,`) inside placeholders.
+
+## Productions
+
+The specification defines the following grammar productions.
```
mihnita, line 282: Note / nitpick: there are currently quite a few things that are parsed but I would not consider valid.

Not actionable or blocker for this PR.
Unless you want to add a note to say "satisfying all rules of the final grammar" (adding "final")

For example `[Foo] one [Bar]` parses (as `[ TOKEN("["), Text("Foo"), TOKEN("]") ]`)
But that is not the only example.

mihnita (changes requested):

spec/syntax.md
eemeli:
```suggestion
        1 [You have one notification.]
```

spec/syntax.md
```diff
@@ -0,0 +1,598 @@
...
+        vocative [Hello, {$userName: person case=vocative}!]
+        accusative [Please welcome {$userName: person case=accusative}!]
+        _ [Hello!]
+
+A message with 2 selectors:
+
+    {$photoCount: number} {$userGender: equals}
+        one masculine [{$userName} added a new photo to his album.]
+        one feminine [{$userName} added a new photo to her album.]
+        one _ [{$userName} added a new photo to their album.]
```
eemeli, line 178:
```suggestion
        1 masculine [{$userName} added a new photo to his album.]
        1 feminine [{$userName} added a new photo to her album.]
        1 _ [{$userName} added a new photo to their album.]
```

spec/syntax.md
eemeli: That might be an artifact of the generated parser, as it does not consider trailing content to be an error.

spec/syntax.md
eemeli: This section is describing the `Pattern` construction, which does require `[` `]` as delimiters, and for which `[Hello, world!]` is a valid example. The un-delimited form is called `Plain`, and is described between the `Message` and `Preamble` sections above.

spec/syntax.md
eemeli: @mihnita Could you clarify whether you're asking for some specific change in this design goal?

spec/syntax.md
eemeli: I think keeping this would be good, as it makes reading the spec easier for someone who may be familiar with MF1. As is, the PR as a whole is optimised for reading rather than editing, and it includes other places as well where the same grammar is expressed in multiple ways. Yes, that does mean that changes will need to be reflected in more than one place, but I would posit that that's a bearable cost when considering the benefits.

spec/syntax.md
```diff
@@ -0,0 +1,598 @@
...
+```ebnf
+Preamble ::= Selector+
+Selector ::= (Variable '=')? '{' Expression '}'
+```
+
+Examples:
+
+```
+$frac = {$count: number minFractionDigits=2}
+    one [One apple]
```
eemeli, line 328:
```suggestion
    1 [One apple]
```

spec/syntax.md
mihnita: I have already marked this as is a nitpick, and suggested a change.
If you think it improves things, fine, implement it.
If not, fine too, close it.

spec/syntax.md
mihnita: OK, your call.

But I think that the differences are so much bigger in the general approach that comparing syntax details is irrelevant.

We will have custom formatter and custom selector functions, a standard function registry, better support for grammatical features, standard mapping to XLIFF, and this lists the `$`, `:`, and comma.

spec/syntax.md
mihnita: So both `[Hello, world!]` and `Hello, world!` are valid?
And worse, parsing them result in different representations in the data model?
But give the same result when rendered?

If that is intentional, let me know and I will open an issue, because that looks like bad design in my opinion.

spec/syntax.md
mihnita: I am asking for a full rewrite of bullet 1.
Since I don't even understand what it means, I can't suggest anything.
It connects completely unrelated issues, like _"we will paint this blue, for compatibility with Texas, and make it easier for people used to public transit"_
And insinuates (again) the baseless idea that we are sacrificing something for some fictitious MF 1 "incremental update"

spec/syntax.md
eemeli: Yes, it is intentional that both `[Hello, world!]` and `Hello, world!` are valid but different representations of a message that would format to a string as `Hello, world!`.

This is comparable to how some languages and formats allow for both `'Hello, world!'` and `"Hello, world!"` to be representations of the same string, while having different handling for their contents (example: shell scripts).

stasm: `Pain` and `Pattern` are different grammar productions, but they both result in the same data model.

spec/syntax.md
stasm: Mihai has a point here. We can write a short comparison document (comparing more than just the syntax) once the design is more fleshed out. As of today, I think it's a bit too early and it will add maintenance burden for us.

spec/syntax.md
mihnita: The " vs ' in some programming languages don't do the same thing:
```
# export name=Mihai
# echo 'Hello $name'
Hello $name
# echo "Hello $name"
Hello Mihai
```

Hawing several ways to do the same things is gratuitous.

TMTOWTDI is not currently seen as a good thing
https://en.wikipedia.org/wiki/There%27s_more_than_one_way_to_do_it

spec/syntax.md
stasm: Here's how I suggest to rephrase the first bullet:

> The syntax should leverage the familiarity with ICU MessageFormat 1.0 in order to lower the barrier to entry and increase the chance of adoption. At the same time, the syntax should fix the [pain points of ICU MessageFormat 1.0](https://github.com/unicode-org/message-format-wg/blob/main/docs/why_mf_next.md).

spec/syntax.md
zbraniecki: > TMTOWTDI is not currently seen as a good thing

I don't think this is a good analogy. The wikipedia article you provided ends with a quote from Python Zen, and Python is exactly allowing for multiple ways to format a string, with different tradeoffs: https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals

The current proposal uses the tradeoff between those two modes.

eemeli: Let's continue this discussion in a separate issue dedicated to the topic?

mihnita (approved):

macchiati:
I will not be able to attend tomorrow's meeting (Monday and Tuesday we're
moving into our new place in SF). I hope to be able to join subsequent
discussions.

To make sense of the differences, I tried putting together a comparison
at the following location, with my notes.

https://docs.google.com/document/d/1t9ZAgdVUb2ILUu-moEpS0g0T-3WZ07uCHwmw9WkMnt4/edit?usp=sharing

 I didn't have the time to be as thorough as I'd like to be, so my
apologies for that.

Mark

PR #273, opened 2022-05-18
Add self-closing MarkupEmpty element

### LINKED ISSUE #238: Markup: how do ensure non-HTML display elements are first class?

eemeli:
As discussed during this Monday's meeting, this PR adds the empty/standalone/self-closing `MarkupEmpty` element to the proposed spec, to represent elements that do not contain any body. This closes #238.

### DIFF
--- a/spec/message.ebnf
+++ b/spec/message.ebnf
@@ -10,7 +10,7 @@ VariantKey ::= String | Nmtoken
 Pattern ::= '[' (Text | Placeable)* ']' /* ws: explicit */
 
 /* Placeables */
-Placeable ::= '{' (Expression | MarkupStart | MarkupEnd)? '}'
+Placeable ::= '{' (Expression | MarkupStart | MarkupEnd | MarkupEmpty)? '}'
 
 /* Expressions */
 Expression ::= Operand Annotation? | Annotation
@@ -21,6 +21,7 @@ Option ::= Name '=' (String | Nmtoken | Variable)
 /* Markup Tags */
 MarkupStart ::= Name Option*
 MarkupEnd ::= '/' Name
+MarkupEmpty ::= Name Option* '/'
 
 <?TOKENS?>
 
--- a/spec/message.ebnf
+++ b/spec/message.ebnf
@@ -10,7 +10,7 @@ VariantKey ::= String | Nmtoken
 Pattern ::= '[' (Text | Placeable)* ']' /* ws: explicit */
 
 /* Placeables */
-Placeable ::= '{' (Expression | MarkupStart | MarkupEnd | MarkupEmpty)? '}'
+Placeable ::= '{' (Expression | Markup | MarkupEnd)? '}'
 
 /* Expressions */
 Expression ::= Operand Annotation? | Annotation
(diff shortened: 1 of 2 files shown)


spec/message.ebnf
```diff
@@ -21,6 +21,7 @@ Option ::= Name '=' (String | Nmtoken | Variable)
 /* Markup Tags */
 MarkupStart ::= Name Option*
 MarkupEnd ::= '/' Name
+MarkupEmpty ::= Name Option* '/'
```
stasm, line 24: This isn't LL(1) because it conflicts with the `MarkupStart`. One way to fix this formally is something like the following (`MarkupStart` may need to be renamed):

```ebnf
MarkupStart ::= Name Option* '/'?
MarkupEnd ::= '/' Name
```

This should be OK because the EBNF only describes the _grammar_, not the data model. Assuming we end up with `ElementStart`, `ElementEnd`, and `ElementEmpty` in the data model, the parser can yield `ElementStart` when `MarkupStart` is missing the slash, and `ElementEmpty` when the slash is present.

stasm (changes requested):
I agree with the change but I have comments about the formal implementation, and I'd like to see another iteration please.

spec/message.ebnf
eemeli: You're right. If keeping this syntax, it might make better to sense to have `Markup` and `MarkupEnd` in the syntax grammar, with `Markup` having a data model attribute like `isEmpty: boolean` determined by the presence (or not) of a trailing `/`.

The discussion in #269 is also relevant, as it may lead to different markup for Markup.

stasm: I realize that the lack of symmetry between `Markup` and `MarkupEnd` may be bothersome. How about a single `Markup` production?

```ebnf
Markup ::= Name Option* '/'? | '/' Name
```

eemeli: While that certainly technically works, it does rather look like two separate constructs crammed into one.

stasm: Right. EBNF requires a bit of formal accounting here and there to make it work, like the one I'm suggesting. Note that I only suggest to combine the _grammar productions_. The data model should have separate interfaces for all three element types.

eemeli: I updated the PR, merging MarkupStart and MarkupEmpty into one Markup element.

spec/syntax.md
```diff
@@ -370,13 +370,14 @@ $when: datetime month=2-digit
 ### Markup
 
 Markup elements provide a structured way to mark up parts of the content.
-There are two kinds of elements: start (opening) elements and end (closing) elements,
-each with its own syntax.
+There are two kinds of elements, each with its own syntax:
+Markup (opening) elements and MarkupEnd (closing) elements.
```
stasm, line 374:
```suggestion
Markup (opening or empty) elements and MarkupEnd (closing) elements.
```

spec/syntax.md
```diff
@@ -370,13 +370,14 @@ $when: datetime month=2-digit
...
+If a Markup element includes a trailing `/`,
+it is considered to be empty, i.e. self-closing.
```
stasm, line 376:
```suggestion
it is considered to be empty, i.e. self-closing.
Otherwise, it's an opening element.
```

stasm (approved):

eemeli:
Closing this, as the change proposed here has been obsoleted by subsequent conversation, esp. in #269.

PR #274, opened 2022-05-18
Use `*` instead of `_` as the catch-all variant key

### LINKED ISSUE #270: Variant Keys: Catchall should not be a valid Nmtoken
[`develop` syntax](https://github.com/unicode-org/message-format-wg/blob/972dc1f255f7a2f8a494c9abe8ae9f58827c5652/spec/syntax.md#variants) uses `_` as a "catch-all key", which can collide with possible values being selected against. It would be better to use a symbol that is not a valid Nmtoken, such as `*` (as is used in the analogous case of [HTTP `If-Match`](https://www.rfc-editor.org/rfc/rfc7232#section-3.1), where `*` matches any representation while `"*"` matches a representation with that specific entity-tag).

eemeli:
This closes #270 by doing as @gibson042 there proposes, and switching to using `*` as the catch-all key. This has the benefit over `_` in that the latter is a valid Nmtoken, and could cause a namespace conflict with user-defined keys.

### DIFF
--- a/spec/message.ebnf
+++ b/spec/message.ebnf
@@ -6,7 +6,7 @@ Selector ::= (Variable '=')? '{' Expression '}'
 
 /* Variants and Patterns */
 Variant ::= VariantKey* Pattern
-VariantKey ::= String | Nmtoken
+VariantKey ::= String | Nmtoken | '*'
 Pattern ::= '[' (Text | Placeable)* ']' /* ws: explicit */
 
 /* Placeables */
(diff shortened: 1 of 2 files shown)


eemeli:
Ping @gibson042, as I can't add you directly as a reviewer.

gibson042 (approved):
Thanks, LGTM!

stasm (approved):

mihnita (approved):

PR #276, opened 2022-05-20
Use (parentheses) rather than "quotes" around literal values

### LINKED ISSUE #263: Pick a delimiter for literals other than the double quote

> Speaking of literals, don't enclose them in quotes. One of the stated goals here is to make messages usable as string literals in programming languages. Enclose literals in parentheses `{... {(5) :number} books}` or angle brackets `{... {<5> :number} books}` or pairs of pipe symbols `{... {|5| :number} books}` or similar.

eemeli:
Closes #263 by replacing `"quotes"` with `(parentheses)` as delimiters for literal values. Correspondingly, the String token is renamed as Literal, though it's still expected to always be parsed into the data model as a string value.

Within a Literal, both `(` and `)` require escaping. Strictly this is only required for the closing parenthesis, but not also escaping the start looks odd, and should improve the experience for anyone looking at MF2 source in an editor that uses generic parenthesis-matching highlighting.

### DIFF
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -441,7 +441,7 @@ NameChar ::= NameStart | [0-9] | "-" | "." | #xB7
 
 ### Literal
 
-Any Unicode codepoint is allowed in literals,
+Any Unicode code point is allowed in literals,
 with the exception of its delimiters `(` and `)`,
 and `\` (which starts an escape sequence).
 
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -280,7 +280,7 @@ The key `*` is a "catch-all" key, matching all selector values.
 
 ```ebnf
 Variant ::= VariantKey* Pattern
-VariantKey ::= String | Nmtoken | '*'
+VariantKey ::= Literal | Nmtoken | '*'
 ```
 
 A well-formed message is considered valid if the following requirements are satisfied:
@@ -300,8 +300,8 @@ This serves 3 purposes:
   `hello = [Hello]` will unambiguously define the `Hello` message without the space in front of it.
 - The message should be conveniently embeddable in various programming languages
   without the need to escape characters commonly related to strings, e.g. `"` and `'`.
-  Such need may still occur when a singe or double quote is
-  used in the translatable content or to delimit a string literal.
+  Such need may still occur when a single or double quote is
+  used in the translatable content.
 - The syntax should make it as clear as possible which parts of the message body
   are translatable and which ones are part of the formatting logic definition.
 
@@ -327,7 +327,7 @@ Placeable ::= '{' (Expression | MarkupStart | MarkupEnd) '}'
 
 Expressions can either start with an operand, or be standalone function calls.
 
-The operand is a quoted string literal or a variable name.
+The operand is a literal or a variable name.
 The operand can be optionally followed by an _annotation_:
 a formatting function and its named options.
 Formatting functions do not accept any positional arguments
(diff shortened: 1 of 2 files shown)


stasm (approved):

spec/syntax.md
```diff
@@ -300,7 +300,7 @@ This serves 3 purposes:
 - The message should be conveniently embeddable in various programming languages
   without the need to escape characters commonly related to strings, e.g. `"` and `'`.
   Such need may still occur when a singe or double quote is
```
aphillips, line 302: typo `singe` => `single`

aphillips (approved):
I'm okay with this change, but remain uneasy about using such a common character pair. I buy @stasm's arguments about using separate open/close characters, otherwise I'd suggest using backtick (`U+0060` `). Or maybe use `<` and `>`? Those aren't as common in "normal" text (although they are not HTML friendly)

eemeli:
One challenge with backticks `` ` `` is that they're commonly used to indicate template or "code" strings. For instance in many markdown variants they require doubling the surrounding backticks to be escaped; I needed to use ``` `` ` `` ``` to get one to render right. They're also often visually difficult to distinguish from single quotes, so may make `` `mistakes'`` difficult to spot.

Angle brackets `<` `>` would need to be escaped in XML & HTML environments, which are rather common.

Hence parentheses `(` `)` as something like a least-worst option.

spec/syntax.md
eemeli:
```suggestion
  Such need may still occur when a single or double quote is
```

stasm:
FWIW, in #263 I was trying to say that for literals we might not need separate open/close delimiters, like I'm advocating for in case of patterns. I also have a concern about angle brackets which I shared in #263 (I'm trying to separate the discussion about the design from the discussion about the implementation).

mihnita:
I don't think that round brackets are not much better than `"`

It means that they become "something to escape", because they can be encountered in plain text.
(I don't know if more often than `"` or not)

TLDR: same as Addison, _"remain uneasy about using such a common character pair."_

---

I don't think that `< ... >` works either, as it conflicts with many existing "storage formats"

[Java can use .xml for resources](https://docs.oracle.com/javase/8/docs/api/java/util/Properties.html#loadFromXML-java.io.InputStream-), and some prefer them to `.properties` because they are Unicode.
.NET uses .xml files for localization ([resx files](https://docs.microsoft.com/en-us/dotnet/core/extensions/create-resource-files#ResxFiles))
The [Apple stringdict file format](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPInternational/StringsdictFileFormat/StringsdictFileFormat.html) used for localization is also xml.
[Android uses xml for translation](https://developer.android.com/guide/topics/resources/string-resource).

Even with HTML-like templating. I know of several localization systems for html that "embed" the localizable strings in html:
```<p i18n=true>Hello {$user}</p>```

**Comment by stasm** [comment]
> It means that they become "something to escape", because they can be encountered in plain text.

There are two "directions" of escaping that we should both be considering.

* Inside: what if I want to use the literal's delimiter as part of the literal?
* Outside: does the literal's delimiter collide with the container's delimiter for strings, or other special characters?

Choosing something else that the single quote, double quote, backtick or angle brackets optimizes for the "outside".

You're right that there may be cases when an "inside" conflict happens. The PR actually makes one of them clear in one of the examples:

```
{(Thu Jan 01 1970 14:37:00 GMT+0100 \(CET\)): datetime weekday=long}
```

Do you think a pipe would be a better choice here? Or any other character?

```
{|Thu Jan 01 1970 14:37:00 GMT+0100 (CET)|: datetime weekday=long}
```

**Comment by stasm** [comment]
FWIW, I think I'm OK with `{(Thu Jan 01 1970 14:37:00 GMT+0100 \(CET\)): datetime weekday=long}` as a trade-off.

**Comment by eemeli** [comment]
Rebased to account for changes in other merged PRs.

@mihnita Did you have concerns about preferring `"..."` over `(...)` for literals?

File: spec/syntax.md [hunk 2]
```diff
@@ -439,26 +439,26 @@ NameChar ::= NameStart | [0-9] | "-" | "." | #xB7
            | [#x0300-#x036F] | [#x203F-#x2040]
 ```
 
-### Quoted Strings
+### Literal
 
-Any Unicode codepoint is allowed in quoted string literals, with the exception of
-`"` (which ends the string literal),
+Any Unicode codepoint is allowed in literals,
```
markusicu, line 444:
```suggestion
Any Unicode code point is allowed in literals,
```

markusicu (approved):
lgtm

mihnita (approved):

romulocintra:
@eemeli needs rebase but ok to merge

PR #278, opened 2022-05-23
Add examples for multiline/complex resources for other languages

batjaa:
It's not clear how multi line/complex examples will look like in `.properties` & `.js` files.

### DIFF
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -156,6 +156,23 @@ A message with a single selector:
         1 [You have one notification.]
         * [You have {$count} notifications.]
 
+The same message defined in a `.properties` file:
+
+```properties
+app.greetings.notification = {$count: number}\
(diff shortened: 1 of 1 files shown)


aphillips:
Needs to be updated to match current syntax. Please revise to match current syntax by next plenary (23 January 2023)

aphillips:
@batjaa Thanks for this PR, which is now a bit stale. Is it your intention to update it? Note that the various ABNF changes and other work mean that the multiline presentation found in many of our documents is not actually part of the syntax (other than as permissible-but-not-meaningful whitespace).

We'll be considering this PR (among others) in the 2023-05-22 teleconference, if you'd care to turn up to discuss it, or let @zbraniecki know (since he often attends on behalf of Amazon). Thanks!

PR #282, opened 2022-06-09
Clarify the absolute nature of "any code point"

gibson042:
Adds explicit mention of cases that are often overlooked.

Ref #268

### DIFF
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -448,7 +448,7 @@ and `\` (which starts an escape sequence).
 
 This includes line-breaking characters (such as U+000A LINE FEED and U+000D CARRIAGE RETURN),
 other control characters (such as U+0000 NULL and U+0009 TAB),
-permanently reserved noncharacters (U+FDD0 through U+FDEF and U+<i>n</i>FFFE and U+<i>n</i>FFFF where n is 0 through 0x10),
+permanently reserved noncharacters (U+FDD0 through U+FDEF and U+<i>n</i>FFFE and U+<i>n</i>FFFF where <i>n</i> is 0x0 through 0x10),
 surrogate code points (U+D800 through U+DBFF),
 private-use code points (U+E000 through U+F8FF, U+F0000 through U+FFFFD, and U+100000 through U+10FFFD),
 and unassigned code points.
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -406,7 +406,7 @@ and `\` (which starts an escape sequence).
 ```ebnf
 Text ::= (TextChar | TextEscape)+ /* ws: explicit */
 TextChar ::= AnyChar - ('[' | ']' | '{' | '}' | Esc)
-AnyChar ::= .
+AnyChar ::= [#x0-#x10FFFF]
 ```
 
 ### Names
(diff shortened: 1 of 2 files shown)


mihnita (approved):

spec/syntax.md
```diff
@@ -406,7 +406,7 @@ and `\` (which starts an escape sequence).
 ```ebnf
 Text ::= (TextChar | TextEscape)+ /* ws: explicit */
 TextChar ::= AnyChar - ('[' | ']' | '{' | '}' | Esc)
-AnyChar ::= .
+AnyChar ::= [#x0-#x10FFFF]
```
eemeli, line 409: This change should also be reflected in the `message.ebnf` file.

spec/syntax.md
```diff
@@ -446,6 +446,15 @@ Any Unicode code point is allowed in literals,
 with the exception of its delimiters `(` and `)`,
 and `\` (which starts an escape sequence).
 
+This includes line-breaking characters (such as U+000A LINE FEED and U+000D CARRIAGE RETURN),
+other control characters (such as U+0000 NULL and U+0009 TAB),
+permanently reserved noncharacters (U+FDD0 through U+FDEF and U+<i>n</i>FFFE and U+<i>n</i>FFFF where <i>n</i> is 0x0 through 0x10),
+surrogate code points (U+D800 through U+DBFF),
+private-use code points (U+E000 through U+F8FF, U+F0000 through U+FFFFD, and U+100000 through U+10FFFD),
+and unassigned code points.
```
eemeli, line 454: Is there a benefit of including such an explicit and exhaustive list here? It feels like it doesn't bring any real benefit, while creating a potential maintenance burden if a theoretical future change were to alter any of these ranges.

eemeli (approved):
Looks good; some nitpicks below.

spec/syntax.md
aphillips: I'm not sure why we don't exclude the non-characters (`*FFFE` and `*FFFF` as well as the surrogates)? The other characters are just characters, but those shouldn't appear?

spec/syntax.md
gibson042: Done.

spec/syntax.md
aphillips: @gibson042 This could become complicated. It depends on whether we intend the syntax to be enforced by implementations, validated in tests, and the like. "Potentially troublesome" characters, such as RLI, are "garbage in/garbage out" (let the user beware). Unassigned characters can become assigned. Etc. But my comment was that non-character code points ought to be excluded (since we are doing text processing). 

This looks like a job for [`USVString`](https://webidl.spec.whatwg.org/#idl-USVString)?

gibson042: > my comment was that non-character code points ought to be excluded (since we are doing text processing)

For the record, I agree with this. A particular negative consequence of the current "anything goes" stance just occurred to me: any Message Format literal including a surrogate code point cannot be represented in UTF-8, which on its own seems to justify excluding _at least_ the same range as USVString.

mihnita: Hmmmm... Maybe we should have reached an agreement on this in an issue, before getting to a PR.

gibson042: This PR is purely editorial, and I think entirely appropriate for prompting such discussion even if it is ends up replaced before merging.

> I think we should think about this syntax as "in memory representation" I loaded the string from somewhere, and it is now in memory.

I strongly disagree. Syntax is something all parties must agree upon, and constrains both what can be communicated and how. It is not an in-memory representation (as evident by e.g. the `\\` and `\"` escape sequences), and even if it were, that would still not absolve the need to address representation in all the various possibilities for getting it _to_ memory.

> This would mean that we should not make this change.

For clarity, what change should we not make? The exclusion of surrogate code points (which I do support but have not included in this PR)?

> I have a slight preference for "anything goes, correct unicode or not". Coding units, not codepoints.

Now you have truly lost me, because code units only apply within an encoding (e.g., UTF-8 octets with special handling of leadings and continuations vs. UTF-16 hexadectets with special handling of surrogates).

romulocintra (approved):

PR #283, opened 2022-06-09
Add +start and -end sigils for markup elements

### LINKED ISSUE #238: Markup: how do ensure non-HTML display elements are first class?

eemeli:
As discussed in #269 and elsewhere, it would be beneficial for all placeholder content to start with some sigil or indicator. For markup, the current draft has used `{b}foo{/b}`; that's changes by this PR to `{+b}foo{-b}`.

It's also intentional that this more clearly uses a style that's different from HTML and XML, in order to help close #238.

### DIFF
--- a/spec/message.ebnf
+++ b/spec/message.ebnf
@@ -10,7 +10,7 @@ VariantKey ::= Literal | Nmtoken | '*'
 Pattern ::= '[' (Text | Placeholder)* ']' /* ws: explicit */
 
 /* Placeholders */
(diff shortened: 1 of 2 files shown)


mihnita:
I think we should not start creating PRs for issues that are not (yet) agreed on.
We are only splitting the debate in two places.

mihnita:
I will approve it to move things forward and not have these many open PRs.

It does not mean I agree with this syntax, so it should not be hold against me 1 month from now ("but you approved X")

There are still at least 2 open issues arguing about how to best handle "markup elements"

mihnita (approved):

PR #287, opened 2022-06-22
Use keywords in syntax

### LINKED ISSUE #252: Preamble: unify local variable definitions and selectors?

### LINKED ISSUE #253: Variants: should variant keys be delimited, too?

### LINKED ISSUE #256: Start in "code" mode exclusively

> First, to a human, there is no clear mental model for how to start reading and writing a message. A message can just be plain text (yeah!), or text in square brackets, or "code". I see how you get that to work in the EBNF, but it's confusing.

> I have always liked simple messages to be simple, with no surrounding syntax. However, that should include placeholders, requiring enclosing-delimiter syntax only when there are placeholders is weird, and allowing simple text both with and without delimiters is also weird.

> Starting in "text" mode also has problems discussed here (uncertainty about trimming spaces), and we have problems with MF1 (ICU MessageFormat) where we don't want people to write literal text outside of a select-type placeholder.

> So while it pains me to say this, bite the bullet, start in "code" mode, and always surround translatable text with delimiters.

### LINKED ISSUE #275: Delimited text looks like code
In general, existing templating and markup languages use delimiters such as `{...}`, `${...}` and `{{...}}` to indicate parts of the source as various forms of "code". In fact, I am not aware of any prior art other than MF1 select/plural variants that use curly braces or square brackets around content that is intended for consideration as "text", i.e. localisable content that is intended to become a top-level part of the formatted output. Fluent comes close, using `{...}` around an entire in-message select expression.

Given this state of affairs, I posit that humans, when scanning a source string, will initially pass over any delimited parts when looking for the content of a message. Furthermore, I posit that breaking from this assumption has made it more difficult for MF1 to reach a greater market share.

Given the above, I think MF2 should have a syntax which delimits variant keys, but not values. In past conversations on this, the strongest argument against such as approach has been that the lack of such delimiters means that the variant value would need to have its surrounding white space trimmed, making the representation of messages with leading or trailing white space more difficult.

I am aware of at least two possible solutions for trimming whitespace around variant values, either of which could provide a workable solution:
1. Trim all whitespace from the start and end of the variant value. This would require that in-content whitespace at the start and end would need to be explicitly escaped, e.g. as `{" "}`. As we are not including any in-message selectors in MF2, this would only be relevant when the whole message is expected to start of end with whitespace. I am not intimately familiar with how all translation tooling handles such cases, but my suspicion is that making such surrounding whitespace explicit would make it far less common for it to be accidentally removed, in particular when translating between languages that handle whitespace very differently.  

2. Define clear and minimal trimming rules for content. A likely pair of replacement rules might be to remove exactly one space from the start, and either a newline followed by spaces or a single space from the end. Expressed in code, that looks like this:
   ```ts
   function trimVariant(raw: string) {
     return raw.replace(/^ ?/, '').replace(/(\n *| )?$/, '');
   }
   ```
   With this approach, leading and trailing whitespace would not need to be escaped.

To illustrate the difference in these approaches, I've included below examples extraced from the syntax proposal and subsequent discussions. In each case, the original (with some variations in syntax) is first, followed by one or two examples of possible delimited-keys syntaxes.

```
1 [You have one notification.]
_ [You have {$count} notifications.]

———

[1] You have one notification.
[_] You have {$count} notifications.
```

```
vocative [Hello, {$userName: person case=vocative}!]
accusative [Please welcome {$userName: person case=accusative}!]
_ [Hello!]

———

[vocative] Hello, {$userName: person case=vocative}!
[accusative] Please welcome {$userName: person case=accusative}!
[_] Hello!
```

```
1 masculine [{$userName} added a new photo to his album.]
1 feminine [{$userName} added a new photo to her album.]
1 _ [{$userName} added a new photo to their album.]
_ masculine [{$userName} added {$photoCount} photos to his album.]
_ feminine [{$userName} added {$photoCount} photos to her album.]
_ _ [{$userName} added {$photoCount} photos to their album.]

———

[1 masculine] {$userName} added a new photo to his album.
[1 feminine] {$userName} added a new photo to her album.
[1 _] {$userName} added a new photo to their album.
[_ masculine] {$userName} added {$photoCount} photos to his album.
[_ feminine] {$userName} added {$photoCount} photos to her album.
[_ _] {$userName} added {$photoCount} photos to their album.

———

[1 masculine] {$userName} added a new photo to his album.
[1 feminine ] {$userName} added a new photo to her album.
[1 _        ] {$userName} added a new photo to their album.
[_ masculine] {$userName} added {$photoCount} photos to his album.
[_ feminine ] {$userName} added {$photoCount} photos to her album.
[_ _        ] {$userName} added {$photoCount} photos to their album.
```

```
one one [Some text...] one few [etc.]

———

[one one] Some text... [one few] etc.
```

As the original sources of these last two examples do not use `[...]` as delimiters, the corresponding key delimiters here are `{[...]}`. This issue isn't meant to be about the specific delimiters that are being used, but about which parts are being delimited in the syntax.
```
=1   fem  {the message 1 F}
=1    _   {the message 1 O}
one  masc {the message One M}
_    masc {the message O M}
_     _   {the message O O}

———

{[  1  fem  ]} the message 1 F
{[  1   _   ]} the message 1 O
{[ one masc ]} the message One M
{[  _  masc ]} the message O M
{[  _   _   ]} the message O O
```

```
[1 female] {{$name} added you to her circles.}
[1 male] {{$name} added you to his circles.}
[1 _] {{$name} added you to their circles.}
[_ _] {{$name} added you and {#count} others to their circles.}

———

{[1 female]} {$name} added you to her circles.
{[1 male]} {$name} added you to his circles.
{[1 _]} {$name} added you to their circles.
{[_ _]} {$name} added you and {#count} others to their circles.

———

{[1 female]} {$name} added you to her circles.
{[1 male  ]} {$name} added you to his circles.
{[1 _     ]} {$name} added you to their circles.
{[_ _     ]} {$name} added you and {#count} others to their circles.
```

Finally, to consider the case of strings with surrounding whitespace, the following examples show how they could look like first with delimited values, then delimited keys with escaped whitespace, and finally when working with the option-2 minimal trimming rules.

```
1 { and one more}
_ { and {$count} more}

———

{[ 1 ]} {" "}and one more
{[ _ ]} {" "}and {$count} more

———

{[ 1 ]}  and one more
{[ _ ]}  and {$count} more
```

```
1 {end of this paragraph
}
_ {end of these paragraphs
}


———

{[ 1 ]} end of this paragraph{"
"}
{[ _ ]} end of these paragraphs{"
"}

———

{[ 1 ]} end of this paragraph

{[ _ ]} end of these paragraphs

```

### LINKED ISSUE #286: Drop restriction on using keywords in the syntax
In recent conversations with @stasm and @mihnita, it's become obvious that especially with more complex messages involving local variables and multiple variants, it is not reasonable to expect a translator unfamiliar and untrained with MF2 syntax to work with the raw plaintext representation of a message and be able to immediately see which parts of it contain the actual message or messages to translate. Consider for instance this example from the current spec:

```
$countInt = {$count :number maximumFractionDigits=0}
$itemAcc = {$item :noun count=$count case=accusative}
one [You bought {$color :adjective article=indefinite accord=$itemAcc} {$itemAcc}.]
* [You bought {$countInt} {$color :adjective accord=$itemAcc} {$itemAcc}.]
```

There's so much going on here that it's effectively a necessity to use some form of tooling assistance to extract the two variants from this message and present them individually to a translator, along with the contextual information provided by the rest of the message.

As a consequence, it is not reasonable or beneficial for us to keep this restriction on our design:
> The syntax should not use nor reserve any keywords in any natural language, such as `if`, `match`, or `let`.

Following #252, we'll need to refactor the current preamble into two parts. When we do that, we should not be afraid of using keywords such as `let` and `match` in our syntax, esp. as it seems we have something like a consensus forming in #256 on starting in "code" mode, where such terms would be quite natural.

This change will help us reach a syntax that's easier to read and which has slightly fewer mysterious symbols and sigils that need to be deciphered.

eemeli:
- Closes #252
- Closes #253
- Closes #256
- Closes #275
- Closes #286

This proposal goes all-in on considering MF2 syntax as "code", and adds three keywords. As something of a side effect, this requires dropping the `Plain` construction.

#### `let`
Used to define the value of a local variable:
```
let $countInt = {$count :number maximumFractionDigits=0}
```

The syntax here is the same as previously in the preamble, with the addition of the `let`. The keyword is not strictly speaking necessary, but makes it easier to see that an assignment is taking place.

#### `match`
Defines the selectors for a multivariant message:
```
match {$countInt} {$user :gender}
```

Using `match` aligns with other pattern matching syntaxes, which more closely than `switch` maps to what we're doing in particular when selecting with more than one selector. Using `select` here wouldn't really work semantically.

#### `when`
Introduces a variant:
```
when 1 masc [...]
when many fem [...]
when * * [...]
```

Another possible alternative here would be `case`, but that somewhat overloads a term that we commonly use to refer to grammatical case.

Put together, these allow for a message that would with the current syntax be represented as
```
$countInt = {$count :number maximumFractionDigits=0}
{$user :gender}
1 masc [...]
many fem [...]
* * [...]
```

to instead read as:
```
let $countInt = {$count :number maximumFractionDigits=0}
match {$countInt} {$user :gender}
when 1 masc [...]
when many fem [...]
when * * [...]
```

While the latter clearly has more characters, it's much easier to figure out from it what's happening.

### Dropping `Plain`

Because this PR makes it possible for a message to start with `let` or `match`, it makes it practically speaking impossible to continue supporting the undelimited "plain" form of simple messages while maintaining the language as LL(1). Therefore, it's dropped.

Practically speaking, it should still be possible to write an efficient detector for plain vs. MF2 messages, should it be desirable in some spec or environment built on top of the MF2 spec (such as the message resource syntax):
1. If a message starts with `let`, `match` or `[`, it may be MF2. If it is, in the first two cases, it will always contain a `{` within the first few tokens.
2. Plain messages may not start with a `[` or contain a `{`.

As one benefit, dropping `Plain` means that MF2 will not provide two different ways to represent the same simple message (`[Hello]` vs. `Hello`).

### Never mind the delimiters

This PR uses the current `[...]` syntax in its examples as that's what's currently on the `develop` branch. It is not meant to offer any opinion on #255 and the option of using `{...}` instead. However, it is relevant to note that this PR does effectively make it necessary to have _some_ delimiters for message contents, concluding the discussion in #275.

### Still not Turing-complete

Given that this change makes MF2 look more like code, it's good to note that we're still at most defining a deterministic finite-state machine. Message processing does not have access to anything like a stack, and cannot modify its own state; formatting calls are independent of each other. This holds even with custom formatting functions, as long as they are require to be [pure functions](https://en.wikipedia.org/wiki/Pure_function).

### DIFF
--- a/spec/message.ebnf
+++ b/spec/message.ebnf
@@ -1,11 +1,11 @@
-Message ::= Plain | Pattern | Preamble Variant+
+Message ::= Declaration* ( Pattern | Selector Variant+ )
 
 /* Preamble */
-Preamble ::= Selector+
-Selector ::= (Variable '=')? '{' Expression '}'
+Declaration ::= 'let' WhiteSpace Variable '=' '{' Expression '}'
+Selector ::= 'match' ( '{' Expression '}' )+
 
 /* Variants and Patterns */
-Variant ::= VariantKey* Pattern
+Variant ::= 'when' ( WhiteSpace VariantKey )+ Pattern
 VariantKey ::= Literal | Nmtoken | '*'
 Pattern ::= '[' (Text | Placeholder)* ']' /* ws: explicit */
 
@@ -22,12 +22,6 @@ Option ::= Name '=' (Literal | Nmtoken | Variable)
 Markup ::= MarkupStart Option*
 
 <?TOKENS?>
-
-/* Plain */
-Plain ::= PlainStart (PlainChar* PlainEnd)?  /* ws: explicit */
-PlainChar ::= AnyChar - ('{' | '}')
-PlainStart ::= PlainChar - ('[' | '$' | WhiteSpace)
-PlainEnd ::= PlainChar - WhiteSpace
 
 /* Text */
 Text ::= (TextChar | TextEscape)+
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -15,8 +15,8 @@
    1. [Complex Messages](#complex-messages)
 1. [Productions](#productions)
    1. [Message](#message)
-   1. [Plain](#plain)
-   1. [Preamble](#preamble)
+   1. [Variable Declarations](#variable-declarations)
+   1. [Selectors](#selectors)
    1. [Variants](#variants)
    1. [Patterns](#patterns)
    1. [Placeholders](#placeholders)
@@ -66,9 +66,6 @@ The design goals of the syntax specification are as follows:
    `.properties`, YAML, XML, inlined as string literals in programming languages, etc.
    This includes a future _MessageResource_ specification.
 
-1. Simple messages that do not use any placeholders or selectors should (as far as possible)
-   be represented in the syntax with no additional characters than their actual contents.
-
 ### Design Restrictions
 
 The syntax specification takes into account the following design restrictions:
@@ -77,51 +74,35 @@ The syntax specification takes into account the following design restrictions:
    It should be possible to define a message entirely on a single line with no ambiguity,
    as well as to format it over multiple lines for clarity.
 
-1. The syntax should not use nor reserve any keywords in any natural language,
-   such as `if`, `match`, or `let`.
-
 1. The syntax should define as few special characters and sigils as possible.
 
 ## Overview & Examples
 
 ### Simple Messages
 
-A simple message without any variables does not need any syntax:
+All messages, including simple ones, need `[…]` delimiters:
 
-    Hello, world!
+    [Hello, world!]
 
 The same message defined in a `.properties` file:
 
 ```properties
-app.greetings.hello = Hello, world!
+app.greetings.hello = [Hello, world!]
 ```
 
 The same message defined inline in JavaScript:
 
 ```js
-let hello = new MessageFormat('Hello, world!')
+let hello = new MessageFormat('[Hello, world!]')
 hello.format()
 ```
 
 ### Simple Placeholders
 
-A message with an interpolated variable needs to be interpreted as a pattern,
-which uses `[…]` delimiters:
+Messages may contain placeholders within `{…}` delimiters,
+such as variables that are expected to be passed in as format paramters:
 
     [Hello, {$userName}!]
-
-The same message defined in a `.properties` file:
-
-```properties
-app.greetings.hello = [Hello, {$userName}!]
-```
-
-The same message defined inline in JavaScript:
-
-```js
-let hello = new MessageFormat('[Hello, {$userName}!]')
-hello.format({ userName: 'Anne' })
-```
 
 ### Formatting Functions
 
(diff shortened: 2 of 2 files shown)


spec/message.ebnf
```diff
@@ -1,11 +1,11 @@
-Message ::= Plain | Pattern | Preamble Variant+
+Message ::= Declaration* ( Pattern | Selector Variant+ )
 
 /* Preamble */
-Preamble ::= Selector+
-Selector ::= (Variable '=')? '{' Expression '}'
+Declaration ::= 'let' WhiteSpace Variable '=' '{' Expression '}'
```
stasm, line 4: I don’t think the EBNF requires that we be explicit about white space here.

```suggestion
Declaration ::= 'let' Variable '=' '{' Expression '}'
```

spec/message.ebnf
```diff
@@ -1,11 +1,11 @@
...
+Selector ::= 'match' ( '{' Expression '}' )+
 
 /* Variants and Patterns */
-Variant ::= VariantKey* Pattern
+Variant ::= 'when' ( WhiteSpace VariantKey )+ Pattern
```
stasm, line 8:
```suggestion
Variant ::= 'when' VariantKey+ Pattern
```

spec/syntax.md
```diff
@@ -229,63 +210,55 @@ if it meets additional semantic requirements about its structure, defined below.
...
+Selector ::= 'match' ( '{' Expression '}' )+
 ```
 
 Examples:
 
 ```
-$frac = {$count: number minFractionDigits=2}
-    1 [One apple]
-    * [{$frac} apples]
+let $frac = {$count: number minFractionDigits=2}
```
stasm, line 242: How about using a simpler example here, without a local variable?

```
match {$count :plural}
when 1 [One apple]
when * [{$count} apples]
```

stasm (approved):

spec/message.ebnf
eemeli: Without the explicit `WhiteSpace`, this is valid:
```
whenfoo [foo]
```

spec/message.ebnf
stasm: Actually, that's wrong. Sorry for the confusion! It's the other way around: we need `/* ws: explicit */` when we want to forbid _optional_ whitespace, not when we want to require at least some. I remembered wrong.

spec/syntax.md
eemeli: Let's instead add a second example, so that we continue to cover the local variable declarations here a bit as well?

````suggestion
match {$count :plural}
when 1 [One apple]
when * [{$count} apples]
```

```
let $frac = {$count: number minFractionDigits=2}
````

File: spec/syntax.md [hunk 4]
```diff
@@ -152,73 +133,73 @@ which the runtime can use to construct a document tree structure for a UI framew
 
 A message with a single selector:
 
-    {$count :number}
-        1 [You have one notification.]
-        * [You have {$count} notifications.]
+    match {$count :number}
+    when 1 [You have one notification.]
```
aphillips, line 137: I really don't care for the use of `when` proposed here. I don't see what value it adds. All of the selectors are clearly part of the `match` statement. The `when` is just extra noise which actually obscures the values.

Note that I still favor delimiting the selectors *and* the literal. The selector delimiter functions the same way that `when` does--identifying that there is an set of values in the `match`.

The single-line example (below, line 143) could then be like:

> match {:platform} [windows] {Settings} [*] {Preferences}

Or (without delims):

> match {:platform} windows {Settings} * {Preferences}

spec/syntax.md
```diff
@@ -77,52 +74,36 @@ The syntax specification takes into account the following design restrictions:
...
-   such as `if`, `match`, or `let`.
-
 1. The syntax should define as few special characters and sigils as possible.
 
 ## Overview & Examples
 
 ### Simple Messages
 
-A simple message without any variables does not need any syntax:
+All messages, including simple ones, need `[…]` delimiters:
```
aphillips, line 83: I still think `{}` brackets are better than `[]` ones because of the need to quote inside the literal.

spec/syntax.md
stasm: Sounds good, thanks.

spec/syntax.md
stasm: Ack. The wording here is just being consistent with the rest of the doc in its current state. I know that #255 is still very much open.

spec/syntax.md
stasm: I see two benefits to `when`:

1. It allows us to add more keywords in the future that come after the last variant, e.g. `meta` or `attribute`. Usually, programming languages solve this by putting the body of the statement inside a curly-brace-delimited block, but I think we said we didn't want to risk people forgetting the closing brace.

2. (This one is more subjective) It creates a visual correspondence between the `match` and each variant, in particular in case of multiline messages.

stasm: > Note that I still favor delimiting the selectors and the literal. The selector delimiter functions the same way that when does--identifying that there is an set of values in the match.

Ack here as well. I'm not opposed to delimiting variant keys, but it's true that I'm still hoping that we can use square brackets `[...]` for delimiting patterns (#255).

Technically, I think just relying on order here is good enough: `match <selector list> <variant list> <pattern value> <variant list> <pattern value>` is OK from the parser's point of view, unless we anticipate some other productions to follow all variants in the future (see point 1 in my comment just above).

spec/syntax.md
```diff
@@ -229,63 +210,61 @@ if it meets additional semantic requirements about its structure, defined below.
...
+A variable declaration is an expression binding a variable identifier
+within the scope of the message to the value of an expression.
+This local variable may then be used in other expressions within the same message.
 
 ```ebnf
-Plain ::= PlainStart (PlainChar* PlainEnd)?  /* ws: explicit */
-PlainChar ::= AnyChar - ('{' | '}')
-PlainStart ::= PlainChar - ('[' | '$' | WhiteSpace)
-PlainEnd ::= PlainChar - WhiteSpace
+Declaration ::= 'let' WhiteSpace Variable '=' '{' Expression '}'
```
stasm, line 227: `let` is clear enough, but I think there's an opportunity here to use a keyword that  directly relates to the name that we choose for these let bindings (#248). This would be self-explanatory and easier to learn and search for. For example:

    local $foo = {$count :number}   // ...and call them "locals"
    alias $foo = {$count :number}   // ...and call them "aliases"
    decl $foo = {$count :number}    // ...and call them "declarations"
    expr $foo = {$count :number}    // ...and call them "named expressions"

mihnita (approved):
Approved, with the understanding that this is not final, and to help with the tech preview.

I still have doubts about '[' vs '{'

And I'm intrigued by the idea to use keywords.
We should see what users say.

But I think it is an improvement because it makes everything start in code mode (no Plain)

romulocintra:
Consensus on MFWG Plenary waiting for @aphillips for the final approval then merge - this is independent of delimiters around variant keys and selectors - this PR it's only about adding keywords

aphillips (approved):
Based on discussion here and elsewhere I'm good to proceed with these changes.

spec/syntax.md
stasm: Since this PR is now closed, let's use #289 to continue this topic.

PR #290, opened 2022-07-18
Prohibit surrogate code units

gibson042:
Surrogate code units (U+D800 through U+DBFF) cannot be encoded into UTF-8.
Ref #268

This covers the most important part of #268, which is UTF-8 compatibility. I would also like to prohibit permanently reserved-for-Unicode-internal-use noncharacters (U+FDD0 through U+FDEF and U+<i>n</i>FFFE and U+<i>n</i>FFFF where <i>n</i> is 0x0 through 0x10, two of which are also invalid XML characters) and control characters (U+0000 through U+001F and U+007F through U+009F, the first 32 of which are not valid unescaped inside a JSON string or [with specific exceptions allowing tab, line feed, and carriage return] in XML), although those can be addressed in a followup (and any of them that should be allowed in string _contents_ will need a corresponding escape sequence, similar to how `\\` represents a single `\`).

### DIFF
--- a/spec/message.ebnf
+++ b/spec/message.ebnf
@@ -26,7 +26,7 @@ Markup ::= MarkupStart Option*
 /* Text */
 Text ::= (TextChar | TextEscape)+
 TextChar ::= AnyChar - ('{' | '}' | Esc)
-AnyChar ::= [#x0-#x10FFFF]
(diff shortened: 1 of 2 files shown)


eemeli (approved):

echeran (approved):

mihnita (commented):
Thank you!
M.

PR #294, opened 2022-08-19
Explain the liberal Text and Literal contents

### LINKED ISSUE #268: Strings: Are all code points preserved?
Per [`develop` syntax: Quoted Strings](https://github.com/unicode-org/message-format-wg/blob/972dc1f255f7a2f8a494c9abe8ae9f58827c5652/spec/syntax.md#quoted-strings), the only code points that may not appear between a string's opening and closing `"` unescaped are `"` itself and `\`, which means that multi-line strings in particular are valid. So I'm considering the behavior of e.g.
```
$text = {"foo
bar

baz"}

[{$text: makeLineBreaksVisible}]
```

There's also the even thornier problems of control characters such as U+0000 NULL and U+0009 TAB, permanently reserved noncharacters (U+FDD0 through U+FDEF and U+<i>n</i>FFFE and U+<i>n</i>FFFF where <i>n</i> is 0x0 through 0x10), and especially surrogate code points (U+D800 through U+DBFF), which are impossible to encode into UTF-8.

gibson042:
Closes #268

Text and Literal are very tolerant because a container format with its own
escape sequences is expected in common use.

This was prompted by post-meeting discussion with @echeran and @mihnita.

### DIFF
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -66,6 +66,13 @@ The design goals of the syntax specification are as follows:
    `.properties`, YAML, XML, inlined as string literals in programming languages, etc.
    This includes a future _MessageResource_ specification.
 
+   - _Non-Goal_: Support unnecessary escape sequences, which would theirselves require
+     additional escaping when embedded. Instead, we tolerate direct use of nearly all
+     characters (including line breaks, control characters, etc.) and rely upon escaping
+     in those outer formats to aid human comprehension (e.g., depending upon container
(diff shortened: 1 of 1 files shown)


echeran (approved):
Looks good, thanks!

stasm (approved):

eemeli (approved):

mihnita (approved):
Thank you!
M

PR #295, opened 2022-08-23
Create notes 2022-08-22.md

echeran:


meetings/2022/notes-2022-08-22.md
```diff
@@ -0,0 +1,214 @@
...
+ 
+STA: If we have a risky runtime, then the runtime is broken.  For the problem in placeholders, we expect that everything should be stringifiable.  For the first issue, I can see the problem there.  I thought we were working on the assumption that most tools don’t let translators touch messages.
+ 
+EAO: ECH, in Fluent, you can have different variants in a selection message. The syntax of Fluent doesn’t require an annotation for what type of value selection should be performed on. Only at runtime is that type is known.  At runtime, we wrap values as a NumberValue for plural selects.  The actual behavior ends up depending on the runtime value. In Fluent, it’s not possible to expect the input value be of a certain type, so we don’t know if the input is the number 1 or if it’s the string `”ONE”`.
```
eemeli, line 75:
```suggestion
EAO: ECH, in Fluent, you can have different variants in a selection message. The syntax of Fluent doesn’t require an annotation for what type of value selection should be performed on. Only at runtime is that type is known.  At runtime, we wrap values as a NumberValue for plural selects.  The actual behavior ends up depending on the runtime value. In Fluent, it’s not required to expect the input value be of a certain type, so we don’t know if the input is the number 1 or if it’s the string `”ONE”`.
```

We didn't cover this during the call, but Fluent does allow for using e.g. an explicit `NUMBER()` wrapper around a selector, which makes it always be handled as a number. So it's possible to have a Fluent message clearly communicate its expected input types, but it's not required.

meetings/2022/notes-2022-08-22.md
```diff
@@ -0,0 +1,214 @@
...
+ 
+STA: I acknowledge that we don’t want to put formatting functions in placeholders, that would be wasteful.. I think there is a point about consistency to be made. We all agree that we want to skip formatting functions in placeholders.  So if that is the case, then how is this case of selectors consistent with that of placeholders?
+ 
+EAO: I think there are 3 layers of type detection for linters, formatters, etc.  If we always have a function attached to a selector, then we don’t need to lookup what the type is. The second lookup is, say if we’re matching  `match {$foo} * {...}`, then we would want to specify in the syntax the types.  The third level is to look into the semantic comments in the message that declares that `$foo` is a number, which allows a linter / tooling to have access to that info.  My sense is that the place this becomes questionable, where the arguments are valid, are the questions about tooling and validation. We are likely to end up with more than one solution.  So maybe we end up with a selector without a function being a linting problem rather than a syntax problem.
```
eemeli, line 87:
```suggestion
EAO: I think there are 3 layers of type detection for linters, formatters, etc.  If we always have a function attached to a selector, then we don’t need to lookup what the type is. The second lookup is, say if we’re matching  `match {$foo} * {...}`, then we could have `$foo` set by a local variable with a well-defined type.  The third level is to look into the semantic comments in the message that declares that `$foo` is a number, which allows a linter / tooling to have access to that info.  My sense is that the place this becomes questionable, where the arguments are valid, are the questions about tooling and validation. We are likely to end up with more than one solution.  So maybe we end up with a selector without a function being a linting problem rather than a syntax problem.
```

eemeli (approved):

mihnita (approved):
Thank you!
M

PR #304, opened 2022-10-17
Add WIP draft document defining formatting behaviour

eemeli:
As we're discussing the MF2 formatting behaviour, it'd be good if we could start codifying it as well.

This PR adds an essentially empty document `spec/formatting.md` that's marked as a "WIP DRAFT" for this purpose.

Later PRs may then add sections to this that'll let us close issues on which we've reached agreement.

PR #305, opened 2022-10-17
Add variable resolution

### LINKED ISSUE #297: Hoisting local variables or not
Example:
```
let $foo = {$bar :number ...}
let $bar = {$count :number ...}
```

Is there a good use case for `$foo` to be able to refer to `$bar`?
I can't think of one.
Some programming languages do this, some don't, and I don't think those that don't have major problems.

Not allowing it would make things simpler:
* you don't have to wait until everything is defined to validate and report errors
* prevents circular references

Forked as a sub-issue from https://github.com/unicode-org/message-format-wg/issues/292

eemeli:
Builds on #304, which should be merged before this.

Closes #297

This PR does not attempt to provide exhaustive contents for variable resolution, but at least a start that should correspond to the consensus reached (so far) in issue #297. It also clarifies the precedence of local vs. external variables, on which we've had some discussion, but I'm not able to find a corresponding issue.

CC @macchiati, who I wasn't able to add directly as a reviewer.

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -7,3 +7,14 @@ when formatting a message for display in a user interface, or for some later pro
 
 The document is part of the MessageFormat 2.0 specification,
 the successor to ICU MessageFormat, henceforth called ICU MessageFormat 1.0.
+
+## Variable Resolution
+
+To resolve the value of a Variable,
+its Name is used to identify either a local variable,
+or a variable defined elsewhere.
+If a local variable and an externally defined one use the same name,
+the local variable takes precedence.
+
+It is an error for a local variable definition to
+refer to a local variable that's defined after it in the message.


spec/formatting.md
```diff
@@ -0,0 +1,20 @@
...
+
+## Variable Resolution
+
+To resolve the value of a Variable,
+its Name is used to identify either a local variable,
+or a variable defined elsewhere.
+If a local variable and an externally defined one use the same name,
+the local variable takes precedence.
+
+It is an error for a local variable definition to
```
macchiati, line 19: The language is a bit awkward; I suggest:

Any reference to a local variable before it is defined is an error.

macchiati (changes requested):

spec/formatting.md
```diff
@@ -0,0 +1,20 @@
...
+refer to a local variable that's defined after it in the message.
```
eemeli, line 20:
```suggestion
Any reference to a local variable before it is defined is an error.
```

spec/formatting.md
eemeli: @macchiati Good point; replaced with your suggestion.

stasm:
Perhaps we should discuss and try to resolve #248 at the plenary meeting today. It would make it easier to make progress on the formatting spec.

spec/formatting.md
```diff
@@ -7,3 +7,14 @@ when formatting a message for display in a user interface, or for some later pro
...
 The document is part of the MessageFormat 2.0 specification,
 the successor to ICU MessageFormat, henceforth called ICU MessageFormat 1.0.
+
+## Variable Resolution
+
+To resolve the value of a Variable,
+its Name is used to identify either a local variable,
+or a variable defined elsewhere.
+If a local variable and an externally defined one use the same name,
+the local variable takes precedence.
```
stasm, line 17: This implies that name shadowing is possible. My current thinking is that it should be OK, and I see use-cases for it. However, I don't recall an explicit discussion on the subject. It might also make sense to start with a stricter design and disallow shadowing at first.

stasm (approved):

spec/formatting.md
eemeli: I rather think that local variables must be allowed to shadow externally provided named arguments.

For instance, I could well imagine translation systems with the following attributes:
1. Translators are able to define additional local variables for a message.
2. A developer may pass the same basket of arguments to multiple messages, not all of which use each of them.

If shadowing were not allowed, then a combination of the above could result in a runtime error which would be difficult to discover ahead of time.

mihnita (approved):

spec/formatting.md
mihnita: I think that this part of the change (shadowing) is a bit of an overreach.
It is unrelated to the hoisting problem, and it was not commonly agreed.

In general I don't think it is a good idea to allow translators to override a developer decision.
So "Translators are able to define additional local variables for a message." is a cons in my book, not a pro.

And _"A developer may pass the same basket of arguments to multiple messages, not all of which use each of them."_ does not look like something related to shadowing.

There is a reason many linters warn about shadowing.

> "If shadowing were not allowed, then a combination of the above could result in a runtime error which would be difficult to discover ahead of time."

Not necessarily.
One can say that translators can't define variables (and that is a good thing).
It is an error, but the decision between runtime error and simply ignore the conflicting local variable is implementation dependent.
We should not use that as an argument for what the standard should do.

---

But TLDR: 

* I don't see valid use-cases for it. In fact the proposal we had used different sigils for local variables (macros at the time) and arguments.
* I don't recall an explicit discussion on the subject.
* It might also make sense to start with a stricter design and disallow shadowing at first.

---


Filed https://github.com/unicode-org/message-format-wg/issues/310

PR #308, opened 2022-10-19
Update README with examples and links to feedback collection

stasm:
I'd like to improve the README so that:

* It tells a better story of what MessageFormat is.
* It showcases a few examples of the draft syntax.
* It points to the existing tech preview implementations.
* It redirects users who want to share feedback to appropriate forums.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,8 +55,8 @@ See more examples and the formal definition of the grammar in [spec/syntax.md](.
 
 ### Implementations
 
-* Java: [`com.ibm.icu.message2`](https://unicode-org.github.io/icu-docs/apidoc/dev/icu4j/index.html?com/ibm/icu/message2/package-summary.html), part of ICU 72 released in October 2022, is a _tech preview_ implementation of the MessageFormat 2.0 syntax, together with parsing and formatting APIs.
-* JavcaScript: [`messageformat`](https://github.com/messageformat/messageformat/tree/master/packages/mf2-messageformat) 4.0 implements the MessageFormat 2.0 syntax, together with a polyfill of the runtime API proposed by ECMA-402.
+* Java: [`com.ibm.icu.message2`](https://unicode-org.github.io/icu-docs/apidoc/dev/icu4j/index.html?com/ibm/icu/message2/package-summary.html), part of ICU 72 released in October 2022, is a _tech preview_ implementation of the MessageFormat 2.0 syntax, together with a formatting API.
+* JavaScript: [`messageformat`](https://github.com/messageformat/messageformat/tree/master/packages/mf2-messageformat) 4.0 implements the MessageFormat 2.0 syntax, together with a polyfill of the runtime API proposed by ECMA-402.
 
 ## Sharing Feedback
 
--- a/README.md
+++ b/README.md
@@ -6,9 +6,9 @@ Welcome to the home page for the MessageFormat Working Group, a subgroup of the 
 
 The Message Format Working Group (MFWG) is tasked with developing an industry standard for the representation of localizable message strings to be a successor to [ICU MessageFormat](https://unicode-org.github.io/icu/userguide/format_parse/messages/). MFWG will recommend how to remove redundancies, make the syntax more usable, and support more complex features, such as gender, inflections, and speech. MFWG will also consider the integration of the new standard with programming environments, including, but not limited to, ICU, DOM, and ECMAScript, and with localization platform interchange. The output of MFWG will be a specification for the new syntax, which is expected to be on track to become a Unicode Technical Standard.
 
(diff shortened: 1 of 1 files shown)


romulocintra (approved):
Thank you @stasm  for this PR

echeran (changes requested):
Flipping the review bit so that we don't merge until Mark's comment is addressed, at least.  We shouldn't knowingly be precise on things that are in reality TBD / done differently.

README.md
```diff
@@ -1,20 +1,76 @@
...
-
-## Guidelines
+The Message Format Working Group (MFWG) is tasked with developing an industry standard for the representation of localizable message strings to be a successor to [ICU MessageFormat](https://unicode-org.github.io/icu/userguide/format_parse/messages/). MFWG will recommend how to remove redundancies, make the syntax more usable, and support more complex features, such as gender, inflections, and speech. MFWG will also consider the integration of the new standard with programming environments, including, but not limited to, ICU, DOM, and ECMAScript, and with localization platform interchange. The output of MFWG will be a specification for the new syntax.
```
eemeli, line 7: Not necessarily requiring action, but seeing what's different here would've been easier if this file used [semantic line breaks](https://sembr.org/).

eemeli (approved):

echeran (approved):

README.md
```diff
@@ -1,20 +1,76 @@
...
 
+- [Why ICU MessageFormat Needs a Successor](docs/why_mf_next.md)
 - [Goals and Non-Goals](guidelines/goals.md)
+- [Record of Consensus Decisions](docs/consensus_decisions.md)
+
+## MessageFormat 2.0 Draft Syntax
```
echeran, line 13:
```suggestion
## MessageFormat 2 Draft Syntax
```

README.md
```diff
@@ -1,20 +1,76 @@
...
+    when * 0 {{$hostName} does not give a party.}
+    when * 1 {{$hostName} invites {$guestName} to their party.}
+    when * 2 {{$hostName} invites {$guestName} and one other person to their party.}
+    when * * {{$hostName} invites {$guestName} and {$guestsOther} other people to their party.}
+
+See more examples and the formal definition of the grammar in [spec/syntax.md](./spec/syntax.md).
+
+### Implementations
+
+* Java: [`com.ibm.icu.message2`](https://unicode-org.github.io/icu-docs/apidoc/dev/icu4j/index.html?com/ibm/icu/message2/package-summary.html), part of ICU 72 released in October 2022, is a _tech preview_ implementation of the MessageFormat 2.0 syntax, together with a formatting API.
```
echeran, line 58:
```suggestion
* Java: [`com.ibm.icu.message2`](https://unicode-org.github.io/icu-docs/apidoc/dev/icu4j/index.html?com/ibm/icu/message2/package-summary.html), part of ICU 72 released in October 2022, is a _tech preview_ implementation of the MessageFormat 2 syntax, together with a formatting API.
```

README.md
```diff
@@ -1,20 +1,76 @@
...
+* JavaScript: [`messageformat`](https://github.com/messageformat/messageformat/tree/master/packages/mf2-messageformat) 4.0 implements the MessageFormat 2.0 syntax, together with a polyfill of the runtime API proposed for ECMA-402.
```
echeran, line 59:
```suggestion
* JavaScript: [`messageformat`](https://github.com/messageformat/messageformat/tree/master/packages/mf2-messageformat) 4.0 implements the MessageFormat 2 syntax, together with a polyfill of the runtime API proposed for ECMA-402.
```

echeran (commented):
Based on offline discussion with @stasm and @eemeli, I'm suggesting changing the "2.0" to "2" so that we don't give the impression that our current working specification has the formalism or completion that an unqualified semver like "2.0" implies.  Especially when measuring progress in relation to the CLDR-TC guidance on the scope of our goals for what 2.0 entails.

optional: the specification's docs on [syntax](https://github.com/unicode-org/message-format-wg/blob/main/spec/syntax.md) and [formatting](https://github.com/unicode-org/message-format-wg/blob/main/spec/formatting.md) both include "DRAFT" alongside the title, but if you agree with these changes, then it would be good to update those instances of "2.0" to "2", too.

stasm:
> optional: the specification's docs on [syntax](https://github.com/unicode-org/message-format-wg/blob/main/spec/syntax.md) and [formatting](https://github.com/unicode-org/message-format-wg/blob/main/spec/formatting.md) both include "DRAFT" alongside the title, but if you agree with these changes, then it would be good to update those instances of "2.0" to "2", too.

I agree with the direction of this change across the repo, but I think we should do it in a separate PR.

echeran:
> I agree with the direction of this change across the repo, but I think we should do it in a separate PR.

Sounds good.

PR #318, opened 2022-12-09
Specify 'format to parts' (issue #272)

mihnita:
Spec for "format to parts": starting point for discussion.
I don't have any illusions that it will go in "as is" :-)

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -18,3 +18,34 @@ the local variable takes precedence.
 
 It is an error for a local variable definition to
 refer to a local variable that's defined after it in the message.
+
+## Format to something other than text (“format to parts”)
+
+It was unanimously agreed that it would be very useful to be able to format a message to something other than plain text (see [issue #272](https://github.com/unicode-org/message-format-wg/issues/272)).
+
+Some use cases:
+
+* styling the output
+* replacing the output (for example replacing a date with a date picker)
+* attaching information to be used by a text-to-speech engine
+* attaching information to use for grammatical features (for example agreement between parts)
+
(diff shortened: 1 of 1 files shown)


eemeli:
This PR seems to be mixing in discussion with the actual proposed specification. Thus far, the content of `spec/formatting.md` has been text that is at least in its shape ready for inclusion directly in the final MF2 specification.

I would rather expect to see something much more terse like this for formatting targets:
> When formatting to a string, the output of a formatter implementation must by default match exactly the behavior specified here. Implementations may provide for configuration that changes this output, as well as providing other formatting targets, such as a sequence of formatted or partially resolved message parts.

mihnita:
> This PR seems to be mixing in discussion with the actual proposed specification.
...
> I agree that the language in the PR is not language that would go into a spec. 

I hear you.

One thing I tried to do was write it in a way that is readable by someone who wants to implement MF2 a number of years from now, without the whole context of the current discussions.

That is why I wanted to cover a bit what is "format to parts" (because it is not a well known term, it is a JavaScript Intl API).
Also a bit of context on the usefulness of such a functionality.

> When formatting to a string, the output of a formatter implementation must by default match exactly the behavior specified here. Implementations may provide for configuration that changes this output, as well as providing other formatting targets, such as a sequence of formatted or partially resolved message parts.

This PR has nothing to do with formatting to string.
And "sequence of formatted or partially resolved message parts" is too restrictive, as it seems to imply non-overlapping parts. Or a non-tree data structure.


Mihai

aphillips:
close in telecon 2023-05-22

PR #339, opened 2023-02-11
Actually omit surrogate code points

aphillips:
DBFF is only the high surrogates...

I probably would have done this rule differently, but I noticed the typo just now and can fast-fix it.


mihnita (approved):

eemeli:
@gibson042 Was there any reason why these were left out in #290?

echeran (approved):

aphillips:
I'm going to merge this. If there is some other reason to have only the one set of surrogates, we can fix later.

gibson042:
Nope, looks like it was just an oversight. :+1:

PR #344, opened 2023-02-13
Add explicit whitespace definitions

### LINKED ISSUE #340: whitespace in the EBNF
The EBNF is a bit handwavy about whitespace. As it is currently written, no whitespace is permitted in places where we often write spaces in our examples, e.g. around the `=` in statements like `let $foo = (bar)`

Go through the EBNF and ensure that we permit (or disallow!) whitespace appropriately. I intend to file a PR for this. 

Note that we should discuss whether we permit LWSP or WSP or just our `WhiteSpace` production. There are some okay-ish arguments for LWSP or just WSP.

**Describe why your solution should shape the standard**
Parsers will be written from the EBNF. It should be correct and complete.

**Additional context or examples**
See above.

stasm:
Fixes #340. This PR implements the following whitespace rules:

> 1. Is whitespace required around curly braces?
>    
>    * `let $foo={...}` or `let $foo = {...}`?
>    * `match{...}` or `match {...}`?

No, whitespace is _optional_ around braced expressions (as in `let $foo={...}` and `match{...}{...}`) nor around braced patterns (`when key{Hello}` is OK). Whitespace is _optional_ around the `=`.

> 2. Is space required after `let`?
>    
>    * `let$foo={...}` or `let $foo={...}`?

Yes, whitespace is required after `let`.

> 3. In a function call, is whitespace required after a literal option value?
>    
>    * `{:func opt=(literal)opt=literal}` or `{:func opt=(literal) opt=literal}`?

Yes, whitespace is required between all option pairs, regardless of whether the option value is a literal or not. Whitespace is _optional_ around the `=`.

> 4. Is whitespace required between consecutive literals?
>    
>    * `when (literal)(literal)` or `when (literal) (literal)`?

Yes, whitespace is required between all _when_ keys, regardless of the kind of the keys.

> 5. Is whitespace required around the default-key star?
>    
>    * `when*` or `when *`?
>    * `when **` or `when * *`?
>    * `when (literal)*` or `when (literal) *`?

Same as above, whitespace is required between all _when_ keys.

The grammar is now LL(1) with backtracking, however I noticed that with one functional change compared to the current `main`, we could make it LL(2) without backtracking. Namely, we can do it by forbidding empty placeholders.

---
There are a few follow-ups I'd like to open PRs for after this one:
* Convert to ABNF, as per #342.
* Rename some productions, e.g. `Function` → `FunctionName`, `Annotation` → `FunctionCall`. Please review this PR without paying too much attention to naming.
* Allow both lowercase and uppercase spelling of the `let`, `match`, and `when` keywords.


spec/message.ebnf
```diff
@@ -1,25 +1,20 @@
...
-VariantKey ::= Literal | Nmtoken | '*'
-Pattern ::= '{' (Text | Placeholder)* '}' /* ws: explicit */
+Pattern ::= '{' (Text | Placeholder)* '}'
+Selector ::= 'match' (s? '{' s? Expression s? '}')+
+Variant ::= 'when' (s Key)+ s? Pattern
+Key ::= Nmtoken | Literal | '*'
 
-/* Placeholders */
-Placeholder ::= '{' (Expression | Markup | MarkupEnd)? '}'
+Placeholder ::= '{' s? (Expression | Markup | MarkupEnd)? s? '}'
```
stasm, line 11: We can make the grammar be LL(2) without backtracking by forbidding empty placeholders:

```suggestion
Placeholder ::= '{' s? (Expression | Markup | MarkupEnd) s? '}'
```

I recall previous discussions with @eemeli about potentially allowing comment-only or metadata-only placeholders in the future. Any thoughts on this?

spec/message.ebnf
```diff
@@ -1,25 +1,20 @@
-Message ::= Declaration* ( Pattern | Selector Variant+ )
+Message ::= (s? Declaration)* s? Body s?
```
eemeli, line 1:
```suggestion
Message ::= s? (Declaration s)* Body s?
```
I think we should require a space after each declaration. Without that requirement, we would allow for e.g.:
```
let $foo={(foo)}{body}
```
or
```
let $foo={(foo)}let $bar={(bar)}{body}
```

spec/message.ebnf
```diff
@@ -1,25 +1,20 @@
...
 
-/* Preamble */
-Declaration ::= 'let' WhiteSpace Variable '=' '{' Expression '}'
-Selector ::= 'match' ( '{' Expression '}' )+
+Declaration ::= 'let' s Variable s? '=' s? '{' s? Expression s? '}'
+Body ::= Pattern | Selector (s? Variant)+
```
eemeli, line 4:
```suggestion
Body ::= Pattern | Selector (s Variant)+
```
Same as with `Message`, without a required space we would allow for
```
match {$foo}when 1{yes}when *{no}
```

spec/message.ebnf
```diff
@@ -1,25 +1,20 @@
...
 
-/* Variants and Patterns */
-Variant ::= 'when' ( WhiteSpace VariantKey )+ Pattern
-VariantKey ::= Literal | Nmtoken | '*'
-Pattern ::= '{' (Text | Placeholder)* '}' /* ws: explicit */
+Pattern ::= '{' (Text | Placeholder)* '}'
+Selector ::= 'match' (s? '{' s? Expression s? '}')+
```
eemeli, line 7:
```suggestion
Selector ::= 'match' (s '{' s? Expression s? '}')+
```
As discussed in #340.

spec/message.ebnf
```diff
@@ -1,25 +1,20 @@
...
+Variant ::= 'when' (s Key)+ s? Pattern
```
eemeli, line 8:
```suggestion
Variant ::= 'when' (s Key)+ s Pattern
```
As discussed in #340.

spec/message.ebnf
```diff
@@ -1,25 +1,20 @@
...
-/* Placeholders */
-Placeholder ::= '{' (Expression | Markup | MarkupEnd)? '}'
+Placeholder ::= '{' s? (Expression | Markup | MarkupEnd)? s? '}'
 
-/* Expressions */
-Expression ::= Operand Annotation? | Annotation
-Operand ::= Literal | Variable
-Annotation ::= Function Option*
-Option ::= Name '=' (Literal | Nmtoken | Variable)
+Expression ::= (Literal | Variable) (s Annotation)? | Annotation
```
eemeli, line 13: Is it intentional to drop out the "operand" term here?

eemeli (changes requested):
I think we should require more spaces; see line comments for specifics. My opinions here are not set in stone, but I would prefer a bit more weight to be placed on the reader experience. So while e.g. `match{$foo}{$bar}` is easily parsable by a machine, I would maintain that a human would find `match {$foo} {$bar}` faster to understand. As with all code, people will spend far more time reading rather than writing MF2.

spec/message.ebnf
eemeli: My current thinking is that it'd be fine to forbid empty placeholders in the syntax. In the data model I'd probably allow for them, but that's really an entirely separate question.

On a separate topic on this same line: Should we include something like this, to make the mapping to a data model more direct?

```suggestion
Placeholder ::= '{' s? PlaceholderBody s? '}'
PlaceholderBody ::= Expression | Markup | MarkupEnd
```

stasm:
> I think we should require more spaces; see line comments for specifics. My opinions here are not set in stone, but I would prefer a bit more weight to be placed on the reader experience. So while e.g. `match{$foo}{$bar}` is easily parsable by a machine, I would maintain that a human would find `match {$foo} {$bar}` faster to understand. As with all code, people will spend far more time reading rather than writing MF2.

I agree with the sentiment, but the grammar is not, in my opinion, the right vehicle to convey such guidelines. There's precedent and a strong convention in many programming languages that braced blocks can be positioned in a rather free-form manner relative to other symbols. I think we should respect it. Otherwise, it's going to turn out that our syntax is actually really quite whitespace-significant outside patterns, too.

spec/message.ebnf
stasm: Ah, an interesting point about being strict in the syntax and lenient in the data model. I like the approach.

stasm: Re. your other question: I don't think `PlaceholderBody` is needed. The grammar only describes a _parse tree_ (also known as a _concrete syntax tree_, CST). The CST will require further transformation before it becomes an AST. This is why I generally didn't attempt to map BNF productions to AST nodes.

spec/message.ebnf
```diff
@@ -29,12 +24,12 @@ TextChar ::= AnyChar - ('{' | '}' | Esc)
 AnyChar ::= [#x0-#x10FFFF] - [#xD800-#xDBFF]
```
aphillips, line 24:
```suggestion
AnyChar ::= [#x0-#x10FFFF] - [#xD800-#xDFFF]
```

Fix the bug we just removed :-).

stasm: Ah, good catch, thanks. Rebased on top of the current `main` and force-pushed.

spec/message.ebnf
```diff
@@ -44,13 +39,13 @@ NameChar ::= NameStart | [0-9] | "-" | "." | #xB7
...
 
 /* Escape sequences */
 Esc ::= '\'
 TextEscape ::= Esc Esc | Esc '{' | Esc '}'
 LiteralEscape ::= Esc Esc | Esc '(' | Esc ')'
 
-/* WhiteSpace */
-WhiteSpace ::= #x9 | #xD | #xA | #x20 /* ws: definition */
+/* White space */
+s ::= (#x9 | #xD | #xA | #x20)+
```
aphillips, line 51: Use `LWSP`, which is built-in, instead of defining our own?

spec/message.ebnf
```diff
@@ -1,25 +1,20 @@
...
-Declaration ::= 'let' WhiteSpace Variable '=' '{' Expression '}'
-Selector ::= 'match' ( '{' Expression '}' )+
+Declaration ::= 'let' s Variable s? '=' s? '{' s? Expression s? '}'
+Body ::= Pattern | Selector (s? Variant)+
 
-/* Variants and Patterns */
-Variant ::= 'when' ( WhiteSpace VariantKey )+ Pattern
-VariantKey ::= Literal | Nmtoken | '*'
-Pattern ::= '{' (Text | Placeholder)* '}' /* ws: explicit */
+Pattern ::= '{' (Text | Placeholder)* '}'
```
aphillips, line 6: `Text` is defined such that backslashes in the text must be escaped, e.g. `\u12AB` is required to be `\\u12AB` or `\n` is required to be `\\n`. I'm okay with requiring `{` and `}` literals to be escaped.

I am nervous that we're recreating the apostrophe disaster in MF1 by requiring every backslash to be escaped. Developers use e.g. `\n` or other escapes in strings all the time. Making double escapes out of them doesn't seem to be required except to allow for the literal `\{` or `\}` to appear in a string. 

I think we might be better off by making the placeholder markers self-escaping, e.g. if you want to have a literal `{` use `{{`. `{` and `}` are rare enough in real text that this might be much less of an impact than requiring every backslash to be doubled.

aphillips (commented):
Looking good as a start. Not a full review--just a couple of quick comments.

spec/message.ebnf
stasm: Looking at the definition in https://datatracker.ietf.org/doc/html/rfc5234#appendix-B, `LWSP` excludes standalone `LF`, which I don't think we want to do?

stasm: Also, the same appendix says that it _contains some basic rules that are in common use_ and that _[b]asic rules are in uppercase_. Does _common use_ mean that they're built-in, or perhaps that they're expected to be defined in a certain way?

The reason I'm asking is that I found an ABNF parser at https://author-tools.ietf.org/abnf and it complained about `DIGIT` and `ALPHA` being undefined. Bug or intended behavior?

spec/message.ebnf
stasm: I filed #345 to discuss empty placeholders. Perhaps we can do it even before this PR is ready to merge.

spec/message.ebnf
stasm: I filed #346 to discuss escaping independent of the whitespace.

spec/message.ebnf
eemeli: I would prefer keeping an explicit whitespace definition within the spec, rather than introducing an external dependency.

spec/message.ebnf
stasm: Yes, I didn't feel the need for the indirection it introduced. As I expressed elsewhere, I don't think the BNF has to directly map to the AST.

spec/message.ebnf
duerst: > Also, the same appendix says that it _contains some basic rules that are in common use_ and that _[b]asic rules are in uppercase_. Does _common use_ mean that they're built-in, or perhaps that they're expected to be defined in a certain way?

I think in an RFC, you would usually say that the following rules are imported from RFC 5234, and give a list of the rules you need.

> The reason I'm asking is that I found an ABNF parser at https://author-tools.ietf.org/abnf and it complained about `DIGIT` and `ALPHA` being undefined. Bug or intended behavior?

DIGIT and ALPHA are too frequent for such a bug not having been detected. I'd guess you have to add the rules you reference from RFC 5234 by hand.

aphillips: Bill's parser (which powers author-tools) has always complained about the built-ins. I think most I-D authors ignore the warnings (I know I learned to). For example, if you run BCP47's ABNF in the tool, it complains about all of the built-ins (and BCP47 is a published set of RFCs blessed by the editor):

```abnf
Language-Tag = langtag / privateuse / grandfathered
langtag = language [ "-" script ] [ "-" region ] *( "-" variant ) *( "-" extension ) [ "-" privateuse ]
; ALPHA UNDEFINED
language = ( 2*3ALPHA [ "-" extlang ] ) / 4ALPHA / 5*8ALPHA
extlang = 3ALPHA *2( "-" 3ALPHA )
script = 4ALPHA
; DIGIT UNDEFINED
region = 2ALPHA / 3DIGIT
variant = 5*8alphanum / ( DIGIT 3alphanum )
extension = singleton 1*( "-" ( 2*8alphanum ) )
singleton = DIGIT / %x41-57 / %x59-5A / %x61-77 / %x79-7A
privateuse = "x" 1*( "-" ( 1*8alphanum ) )
grandfathered = irregular / regular
irregular = "en-GB-oed" / "i-ami" / "i-bnn" / "i-default" / "i-enochian" / "i-hak" / "i-klingon" / "i-lux" / "i-mingo" / "i-navajo" / "i-pwn" / "i-tao" / "i-tay" / "i-tsu" / "sgn-BE-FR" / "sgn-BE-NL" / "sgn-CH-DE"
regular = "art-lojban" / "cel-gaulish" / "no-bok" / "no-nyn" / "zh-guoyu" / "zh-hakka" / "zh-min" / "zh-min-nan" / "zh-xiang"
alphanum = ( ALPHA / DIGIT )
obs-language-tag = primary-subtag *( "-" subtag )
primary-subtag = 1*8ALPHA
subtag = 1*8( ALPHA / DIGIT )
; CRLF UNDEFINED
registry = record *( "%%" CRLF record )
record = 1*field
field = ( field-name field-sep field-body CRLF )
field-name = ( ALPHA / DIGIT ) [ *( ALPHA / DIGIT / "-" ) ( ALPHA / DIGIT ) ]
; SP UNDEFINED
field-sep = *SP ":" *SP
```

stasm:
@eemeli All your unresolved comments are related to the whitespace requirement around `{...}`, both when used to wrap expressions and when used to wrap patterns. Are you willing to relax your opinion on this?

eemeli:
After some consideration, I think my only remaining concern is minification. I agree that there's some value in being lenient during parsing.

For existing web formats such as CSS, JSON & JS, when they're transmitted over the wire, it is common for any extra spaces, comments and such to be stripped out, which leads to a formulation that's sufficiently terse to need tooling before it's understandable by humans.

If we make spaces around `{...}` optional, we would end up with messages being transmitted in such form as well:

```
let $f={(foo)}match{$f}when 1{yes}when *{no}
```

On the other hand, if we require the spaces, the minimum size of this message would grow by six characters, but it would be much more human-readable:
```
let $f={(foo)}
match {$f}
when 1 {yes}
when * {no}
```

I would be interested to hear if anyone else has thoughts on this aspect, which I'm not aware of us explicitly discussing before. If you think that this is not of much concern, then I'd be fine relaxing my stance.

stasm:
> I would be interested to hear if anyone else has thoughts on this aspect, which I'm not aware of us explicitly discussing before.

It's fair to expect that people will want to attempt minification, although personally I wouldn't recommend it, mostly for practical reasons. Our syntax is already quite minimal and most of the contents of a message is the translation itself—which must not be minified. I can perhaps see some opportunity in removing leading whitespace from before `when` clauses, if the original formatting indents them. But otherwise, the benefits of minification are going to be minuscule.

alerque:
Compressing message content over the wire using GZIP or other standard mechanisms will be several orders of magnitude more effective than any possible minification, and also completely unaffected by any attempts to minify before compression.

eemeli:
> [...] I can perhaps see some opportunity in removing leading whitespace from before `when` clauses, if the original formatting indents them. But otherwise, the benefits of minification are going to be minuscule.

Actually, the benefits of minifying message resources can be significant, when considering a well-commented file as a starting point. All of the contextual information required for translation is unnecessary at runtime, and can easily make up a quarter or more of the total size. If/when used for web content, MF2 resources will be minified.

The application of an intelligent minifier is going to rely on a parser that understands the syntax it's working with, and it _will_ apply all possible savings, trimming whitespace wherever it can even if the difference between `match {...}` and `match{...}` is, as I would agree with you, minuscule.

To see what I mean, take a look at the static nominally human-readable text-based assets that are loaded alongside the HTML of this very page, and observe that they are all minified to the furtheset extent possible.

Given this, my point here then is that making the whitespaces around `{...}` that we're talking about optional will lead to minified MF2 content being just as unreadable as minified CSS, whereas making them required will make the minified output much more readable.

eemeli:
As a data point regarding minification, the total compressed size of all 280 FTL files currently in the Firefox repo is 274kB when raw, but 201kB minified; that's a saving of about 27%. I would expect minification of an MF2 resource format to provide similar file size savings for most users.

spec/message.ebnf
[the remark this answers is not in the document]
stasm: Thanks, @echeran. I tried to limit the changes in this PR to only those related to whitespace. That's why in the description I wrote: _ Please review this PR without paying too much attention to naming._ :)

The exact change you're suggesting is implemented in #347. Glad to see that we agree :)

spec/message.ebnf
stasm: > In addition to +1 on @eemeli's [comment](https://github.com/unicode-org/message-format-wg/pull/344/files#r1105089539) to allow for whitespace between `Declaration`s (the `let ...`) [...]

Did you mean to say "to allow" or "to require"? @eemeli's comment uses a stronger language.

> Same applies to the [comment](https://github.com/unicode-org/message-format-wg/pull/344/files#r1105091618) for whitespace between each variant in a selection message. At least, by "natural", I mean that I think this is how we have already written all of our example messages in our documentation so far.

I agree about the "natural". I think we should recommend readable formatting and lead by example. This PR is more about defining the bounds of what's considered valid, even if it's formatted horribly.

echeran:
> I would be interested to hear if anyone else has thoughts on this aspect, which I'm not aware of us explicitly discussing before. If you think that this is not of much concern, then I'd be fine relaxing my stance.

I also think it would be preferable to require whitespace around `{...}`, even if that is stricter. I think it increases the readability of the selector+variant portion because the whole thing becomes consistently space-separated.

We haven't talked about whitespace specifically, but here are my high level understandings of how we think about the syntax:

1. It should be able to represent the data model 1:1
2. It should be human-friendly

Explaining the above, starting with the 2nd point (human-friendly representation), I think the argument for eschewing delimiters (https://github.com/unicode-org/message-format-wg/issues/253, https://github.com/unicode-org/message-format-wg/issues/255) in favor of keywords (https://github.com/unicode-org/message-format-wg/pull/287) was that it would be more human-friendly. The space savings and regularity of delimiters were not the biggest criteria for those arguments in favor of keywords.

Regarding the first point, it should be obvious. Ever since we started talking about the syntax in earnest (https://github.com/unicode-org/message-format-wg/issues/178), we've wanted to have a syntax that represents anything that the data model can specify. And even though "human-friendly" can be a squishy thing to measure (ex: I think delimiters are more important for readability than keywords), I'm okay because I know there's always a 1:1 correspondence to any other representation/API I would find.

So we already have precedent for when readability in service of being human-friendly wins over saving a few characters. I think the minification arguments don't move me much because, at this point, the unresolved comments all relate to whitespace around `{...}`, which seems too relatively infrequent. I also trust / agree with @alerque 's sentiment that if we really care about space savings, then we would be better off starting with a compression like GZIP.

@stasm's point about parser leniency (should this be required vs. optional?) is expounded upon more over in the issue at https://github.com/unicode-org/message-format-wg/issues/340#issuecomment-1432852141, so I'll respond there.

spec/message.ebnf
echeran: @stasm Ah, yep, I meant "require".

alerque:
@eemeli How much of that difference is the removal of optional and inconsequential (but potentially hefty) blobs like comments an how much is reducing the syntax niceties like whitespace to a minimum?

eemeli:
> @stasm:
> I would guess most of the savings can be attributed to the removal of comments, the indentation of attributes and the indentation of nested selectors? None of which exist in MessageFormat 2.

Also in answer to @alerque: The vast majority (probably something like 95%) of the savings come from dropping comments.

> Taking a step back: are you advocating that minification is unavoidable and that we should attempt to preserve the readability of minified messages at the expense of the parser’s leniency?

I would like us to be explicit about the choice we're making. We're building a new format for encoding human messages that we've put some work into making human-readable. The ecosystem around us is one that now has years of experience applying AST transforms on source code for various purposes, so the complexity and cost of applying e.g. minification to MF2 is marginal. Therefore, we should presume that it will be used by at least some large fraction of MF2 emitters that presume that their output will primarily be consumed by machines rather than humans.

So while it is entirely valid for us to make the call that whitespace being optional results in a more lenient parser and that's desirable, I want our choice on this to be explicit rather than implicit.

stasm:
At the February 27 teleconference we agreed to move ahead with this PR. To summarize the current agreement:

There have been three major threads of discussion above:

1. whether to require whitespace around `{...}`, i.e. whether `match{...}` should be allowed,
2. whether to require whitespace between variant keys, i.e. whether `when***` and `when foo*` should be allowed,
3. whether to require whitespace between option pairs, i.e. whether `{:func opt=(x)opt=y}` should be allowed.

This PR implements the following answers:

1. no, whitespace around `{...}` is optional,
2. yes, whitespace between variant keys is required,
3. yes, whitespace between option pairs is required.

---

During the meeting we also discussed possibly further relaxing the whitespace requirements related to (2) and (3) above. I'll file a new PR with suggested changes to the BNF, and we will continue the discussion there.

eemeli (approved):
As discussed on the call today, I'm ok with this proceeding in its current shape.

aphillips (approved):
Please merge this. I believe this will almost immediately be overwritten by the ABNF version??

stasm:
> I believe this will almost immediately be overwritten by the ABNF version??

Yes. #347 will follow today or tomorrow.

PR #347, opened 2023-02-14
Define the grammar as an ABNF (RFC 5234)

### LINKED ISSUE #342: Choose BNF syntax for describing the grammar
Currently, `spec/message.ebnf` is written using the so-called W3C EBNF. It's one of a few BNF variations, commonly used in W3C and Unicode (I think?).

One of the nice-to-have reasons for picking it was that it's also supported by [REx](https://www.bottlecaps.de/rex/), an online parser generator, as well as the [Railroad Diagram Generator](https://www.bottlecaps.de/rr/ui), both by Gunther Rademacher. Having good tool support and being able to immediately test grammar ideas was beneficial to the rapid development-style of the syntax design process that we went through last year.

That said, I haven't been able to find other tools that supports this variant of BNF out of the box. This makes me a bit uncomfortable. Ideally, we would be able to define the grammar in a way that allows an arbitrary text snippet to be validated as MF2 syntax, parse it into a concrete syntax tree (CST), visualize both the rules and such tree, and even generate random strings that match the grammar, for the purpose of fuzz testing. I'm a bit disappointed by the state of the tooling in this regard.

* The original [BNF](https://en.wikipedia.org/wiki/Backus–Naur_form) which doesn't have some quality-of-life improvements of its successors, like syntax for alternatives or repetitions.
* [ABNF](https://en.wikipedia.org/wiki/Augmented_Backus–Naur_form) (_Augmented BNF_) which adds alternatives (`foo / bar`), optional symbols (`[foo]`), and repetitions (`n*m(foo)`).
  * ABNF has been standardized as [RFC 5234](https://datatracker.ietf.org/doc/html/rfc5234).
  * This variant is also defined at https://www.w3.org/Notation.html.
* Despite the above, some W3C standards (notably XML) use a different notation, which is commonly referred to as the [W3C EBNF](https://www.w3.org/TR/REC-xml/#sec-notation).
  * It uses `|` for alternatives, and Kleene operators for optionals and repetitions (`?`, `*`, `+`).
  * This variant is also used in some of [Unicode's documents](https://unicode.org/reports/tr18/#Notation).
* There's also another thing called _EBNF_ defined by [ISO/IEC 14997](https://www.iso.org/standard/26153.html), sometimes also called the [Wirth syntax notation](https://en.wikipedia.org/wiki/Wirth_syntax_notation). It uses `|` for alternatives, brackets for optional symbols (`[foo]`), and curly braces for repetitions (`{foo}`).
  * _Trivia_: This one is based on Niklaus Wirth's 1977 paper titled [_What can we do about the unnecessary diversity of notation for syntactic definitions?_](https://dl.acm.org/doi/10.1145/359863.359883) I sympathize with the question!
* There exist other variations, too, each with a different set of "minor modifications". E.g. a legacy search server in Windows used to have its [own EBNF](https://learn.microsoft.com/en-us/previous-versions/windows/desktop/indexsrv/description-of-the-ebnf-notation). Interestingly, it defined a special syntax for denoting a requiring space, which is something I wish was easily possible in other EBNFs (see #340).
* Lastly, there are BNF variants oriented towards parser generation, like the ones used by Yacc and Bison. They include snippets of C code ("actions") which are used to control the shape of the parse tree.

Outside the realm of context-free grammars (CFGs), it looks like [parser expression grammars](https://en.wikipedia.org/wiki/Parsing_expression_grammar) (PEGs) are also a popular choice for defining grammars of programming languages. E.g. Python switched to a PEG in [PEP 617](https://peps.python.org/pep-0617/).

There's also Tree-sitter developed by GitHub, which uses [parser combinators written in JavaScript](https://tree-sitter.github.io/tree-sitter/creating-parsers#writing-the-grammar) to define grammars, from which it can then generate parsers.

I'm opening this issue to discuss what requirements we have for the formal grammar of MessageFormat and to choose one of the available formats.

stasm:
_Based on #344. Fixes #342._

Convert our W3C EBNF grammar to ABNF, as defined by [RFC 5234](https://datatracker.ietf.org/doc/html/rfc5234).

I validated the grammar using https://author-tools.ietf.org/abnf.

I was also able to run some fuzz testing by:

* generating 1000 random files using [abnfgen](https://www.quut.com/abnfgen/),
* auto-converting the ABNF back to W3C EBNF using https://www.bottlecaps.de/convert/,
* generating an LL(2) parser from it using https://www.bottlecaps.de/rex/,
* and finally parsing the generated files with the parser.

Not all files parsed correctly. I'll investigate.

- [x] Remove `message.ebnf`.
- [x] Update BNF snipptes in `syntax.md`.
- [x] Update the _Complete EBNF_ section in `syntax.md`.


spec/message.abnf
```diff
@@ -0,0 +1,55 @@
...
+literal = "(" *(literal-char / literal-escape) ")"
+literal-char = any-char / "{" / "}"
+
+text-escape = "\" ("\" / "{" / "}")
+literal-escape = "\" ("\" / "(" / ")")
+
+s = 1*(%x09 / %x0D / %x0A / %x20)
+
+ALPHA = %x41-5A / %x61-7A
+DIGIT = %x30-39
```
alerque, line 55: These two rules are part of the core rule set and, as I read it, should not be redefined. The spec defines them as the same values you have here anyway.

```suggestion
```

stasm: Thanks. I wasn't sure if they were built-in, or just recommended to be defined by implementations.

spec/message.abnf
```diff
@@ -0,0 +1,52 @@
...
+key = nmtoken / literal / "*"
+
+placeholder = "{" [s] (expression / markup / markup-end) [s] "}"
+
+expression = ((literal / variable) [s annotation])
+           / annotation
+annotation = function *(s option)
+option = name [s] "=" [s] (literal / nmtoken / variable)
+
+markup = markup-start *(s option)
```
aphillips, line 20: This requires markup options to be k-v pairs, e.g. these are not okay:

> {+markup false}
> {+checkbox checked}

Do we want to be that strict?

aphillips (commented):
This is looking really good. Thanks Stas!

Most of my comments are corner cases and probably I should raise them as separate issues.

spec/message.abnf
```diff
@@ -0,0 +1,52 @@
+message = [s] *(declaration [s]) body [s]
+
+declaration = "let" s variable [s] "=" [s] "{" [s] expression [s] "}"
```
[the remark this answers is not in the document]
stasm: That's a good point. My plan was to file a follow-up for us to decide whether we want `let` *and* `LET`, or perhaps just the lowercase `let`. We can then use RFC 7405's case-sensitive literals, e.g. `%s"let"` and `%s"LET"` to encode the decision in the grammar.

spec/message.abnf
```diff
@@ -0,0 +1,52 @@
...
+
+text = 1*(text-char / text-escape)
+text-char = any-char / "(" / ")"
+
+variable = "$" name
+function = ":" name
+markup-start = "+" name
+markup-end = "-" name
+name = name-start *name-char
+nmtoken = 1*name-char
```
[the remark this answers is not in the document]
stasm: `name` and `nmtoken` are taken directly from https://www.w3.org/TR/xml/#NT-Name. And yes, the goal was to make it convenient to use (a) numerical option values (`fractionDigits=1`) and (b) option values such as `year=2-digit`.

More broadly, the use of `nmtoken` is intended to align MF2 with LDML. CLDR data is defined as attribute enums (e.g. in [ldmlSupplemental.dtd](https://github.com/unicode-org/cldr/blob/release-42/common/dtd/ldmlSupplemental.dtd)), which [implies that they are Nmtokens](https://www.w3.org/TR/xml/#sec-attribute-types).

spec/message.abnf
stasm: I'd prefer to start with the exact same option=value syntax as in function calls. Please file an issue if you'd like to suggest making markup different?

spec/message.abnf
```diff
@@ -0,0 +1,52 @@
...
+name-start = ALPHA / "_"
```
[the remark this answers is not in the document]
stasm: Please open a new issue :) This isn't changing in this PR.

spec/message.abnf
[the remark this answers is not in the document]
aphillips: `nmtoken` however is actually exposed to more than just LDML/CLDR constructs. Developers of selectors and formatters will use them to define key values and option values. That is, our user base is wider than just CLDR stuff or ICU developers. If we define values that are a strict superset of those restrictions then any XML, LDML, or CLDR values will work. LDML being a dialect of XML means that they are restricted by XML, but we don't ***have*** to be. Thoughts?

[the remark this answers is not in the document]
stasm: Wait, are you suggesting that keys and option values should be even less restricted than `nmtoken`?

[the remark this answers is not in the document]
stasm: My goal here (last year when I wrote this for my proposal) was to make sure that whatever appears in LDML attlists (now or in the future) can be a valid key or an option value. That's why I chose `Nmtoken` for these two productions.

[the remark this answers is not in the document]
aphillips: Yes. I'm suggesting that any characters we omit we know why we are omitting them. Some implementations will probably not check the values so long as they are able to parse the various tokens (which don't really depend on the characters inside the token); others will pedantically check the character ranges.

To be honest, most developers will write enumerated values in ASCII for keys and option values. But for cases when they don't the restrictions on users should make sense in our world (and not just because someone else did it)

spec/message.abnf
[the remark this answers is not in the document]
alerque: [RFC 7405](https://datatracker.ietf.org/doc/html/rfc7405) defines a case sensitive variant of ABNF. Instead of `"let"` as a string literal you would use `%s"let"` to define it as case sensitive or `%i"let"` for insensitive. Conversion to RFC 5234 readable syntax should be pretty easy in reverse if you were to write RFC 7405 in the first place.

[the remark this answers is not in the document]
alerque: Oops jinxed! My comment was started before the other reply, then I got sidetracked putting kids to bed.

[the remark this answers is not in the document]
stasm: I filed #349 for this.

spec/message.abnf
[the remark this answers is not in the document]
aphillips: Done as #350. You can close this comment.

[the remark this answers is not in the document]
stasm: Thank you!

spec/message.abnf
```diff
@@ -0,0 +1,52 @@
...
+option = name [s] "=" [s] (literal / nmtoken / variable)
+
+markup = markup-start *(s option)
+
+any-char = %x0-27 ; omit ( and )
+         / %x2A-5B ; omit \
+		 / %x5D-7A ; omit {
+		 / %x7C ; omit }
+		 / %x7E-D7FF
+		 / %xE000-10FFFF
```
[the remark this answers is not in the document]
stasm: FYI I refactored `text-char` and `literal-char` slightly in https://github.com/unicode-org/message-format-wg/pull/347/commits/6ee08f48b3abc5fe10fffaaba9874211e2cbda5e, without changing their meaning.

aphillips (approved):
Leaving aside my recently-filed issues that are not on topic, I think this is merge-worthy.

stasm:
Thanks, @aphillips. I researched using ABNF _after_ I started working on whitespace rules. This PR is unfortunately based on my work in #344. I'll keep it as a draft while #344 is being discussed.

spec/message.abnf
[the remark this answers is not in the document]
alerque: If one or two words turn out to me the only case sensitive bits, you might be able to skip the extensions with just an exact character code match for the lower case letters:

```abnf
let = %d154.145.164
```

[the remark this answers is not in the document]
stasm: Ah, good idea. Yes, there are only three keywords: `let`, `match`, and `when`.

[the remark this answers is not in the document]
stasm: > That's a good point. My plan was to file a follow-up for us to decide whether we want `let` _and_ `LET`, or perhaps just the lowercase `let`. We can then use RFC 7405's case-sensitive literals, e.g. `%s"let"` and `%s"LET"` to encode the decision in the grammar.

New plan: the W3C flavor of EBNF that this PR migrates _from_ is case-sensitive, which means that the currently agreed upon grammar only accepts lowercase `let`, `match`, and `when`. I filed #349 to discuss possibly relaxing this, but in the meantime I added one more commit to this PR (https://github.com/unicode-org/message-format-wg/pull/347/commits/84bbfa1dd751eb1915514cf7eb37e3834748bbf8) which explicitly encodes the lowercase keywords in the ABNF, to maintain the status quo.

spec/message.abnf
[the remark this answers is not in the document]
duerst: @aphillips 
> LDML being a dialect of XML

I hope you meant "LDML being an application of XML". XML doesn't have any dialects (except for XML 1.1, maybe).

spec/message.abnf
```diff
@@ -0,0 +1,58 @@
+message = [s] *(declaration [s]) body [s]
+
+declaration = let s variable [s] "=" [s] "{" [s] expression [s] "}"
```
aphillips, line 3: In writing a response to [this comment](https://github.com/unicode-org/message-format-wg/issues/350#issuecomment-1433903054), I noticed that we do not permit nested expressions.

My example wanted to look something like this:

```
let 1901 = :eq var1=|1901| var2={:locale_get key=variant}
```

spec/message.abnf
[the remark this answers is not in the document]
aphillips: @duerst exactly so

spec/message.abnf
stasm: I'm assuming you mean something like the following, using the current syntax?

    let $year = {:eq var1=1901 var2={:locale_get key=variant}}

Or perhaps this?

    let $year = {:eq var1=1901 var2={:locale_get} key=variant}

We considered this in early 2021, when we discussed the EM and EZ data models. The EZ model proposed something we dubbed the "treeness" of function calls. I did a presentation in August 2021 attempting to break the deadlock: [The Third Data Model](https://docs.google.com/presentation/d/1pex8lEIQ0dFs72ATxva0IprIP6xLe14Sop2Oy4xPXFo/edit?usp=sharing). The group agreed to move forward with the function call model it proposed:

* function calls cannot be nested,
* variables passed into functions would have their own data model nodes, i.e. `{kind: "Variabled", name: "foo"}` rather than be just strings resolved inside functions (like the EM proposal wanted).

eemeli: One significant part of that earlier conversation was to identify that local variables may be used to get the same effect, as in:
```
let $var2 = {:locale_get key=variant}
let $year = {:eq var1=1901 var2={$var2}}
```

aphillips: thanks @stasm and @eemeli. We should call out "no nesting" in the spec. In fact, I'll task myself with that in #353.

spec/message.abnf
```diff
@@ -0,0 +1,58 @@
...
+body = pattern
+     / (selectors 1*([s] variant))
+
+pattern = "{" *(text / placeholder) "}"
+selectors = match 1*([s] selector)
+selector = "{" [s] expression [s] "}"
+variant = when 1*(s key) [s] pattern
+key = nmtoken / literal / "*"
+
+placeholder = "{" [s] (expression / markup / markup-end) [s] "}"
```
aphillips, line 13: I think we can fix the markup by moving around some of the constructs. Perhaps:

```abnf
pattern = "{" *(text / placeholder / markup) "}"
markup = markup-start [s] *(text / placeholder) [s] markup-end
placeholder = "{" [s] expression [s] "}"
```

eemeli: This proposal changes three aspects of how markup currently works that would need to be discussed separately:
- It disallows markup within markup, so e.g. `{+b}{+i}...{-i}{-b}` becomes invalid.
- It requires markup start & end tags to be matched within a message.
- It swallows whitespace adjacent to a markup tag.

aphillips: @eemeli 

* It could be fixed to allow embedded markup.
* You're right that it would swallow whitespace and would need fixing that way
* It does require matched markup. We discussed this elsewhere and I filed #356 about that. Given that choice, this would not be appropriate.

So I'll resolve this comment. But `markup` feels unresolved to me. The failure states and fallback states of markup need to be carefully thought through.

alerque (approved):

spec/message.abnf
```diff
@@ -0,0 +1,57 @@
...
+
+expression = ((literal / variable) [s annotation])
+           / annotation
+annotation = function *(s option)
+option = name [s] "=" [s] (literal / nmtoken / variable)
+
+markup = markup-start *(s option)
```
eemeli, line 20: Nitpick: we could get away with not defining the `function` and `markup-start` productions, as they're only used here.
```suggestion
annotation = ":" name *(s option)
option = name [s] "=" [s] (literal / nmtoken / variable)

markup = "+" name *(s option)
```

spec/message.abnf
```diff
@@ -0,0 +1,57 @@
...
+name-start = ALPHA / "_"
+           / %xC0-D6 / %xD8-F6 / %xF8-2FF
+           / %x370-37D / %x37F-1FFF / %x200C-200D
+           / %x2070-218F / %x2C00-2FEF / %x3001-D7FF
+           / %xF900-FDCF / %xFDF0-FFFD / %x10000-EFFFF
+name-char = name-start / DIGIT / "-" / "." / %xB7
+          / %x0300-036F / %x203F-2040
+
+text-escape = "\" ("\" / "{" / "}")
+literal-escape = "\" ("\" / "(" / ")")
```
eemeli, line 55: Nitpick: these would be clearer without the grouping.
```suggestion
text-escape = "\\" / "\{" / "\}"
literal-escape = "\\" / "\(" / "\)"
```

eemeli (approved):
A few editorial suggestions inline, but I'm fine with merging even without them.

spec/message.abnf
stasm: Thanks, this looks better indeed. Can you make the same suggestion to `syntax.md`, please?

spec/message.abnf
eemeli: Sounds good!

spec/syntax.md
```diff
@@ -465,27 +476,33 @@ In particular, the grammatical feature data [specified in LDML](https://unicode.
...
 Escape sequences are introduced by the backslash character (`\`).
 They are allowed in translatable text as well as in literals.
 
-```ebnf
-Esc ::= '\'
-TextEscape ::= Esc Esc | Esc '{' | Esc '}'
-LiteralEscape ::= Esc Esc | Esc '(' | Esc ')'
+```abnf
+text-escape = "\" ("\" / "{" / "}")
+literal-escape = "\" ("\" / "(" / ")")
```
eemeli, line 505:
```suggestion
text-escape = "\\" / "\{" / "\}"
literal-escape = "\\" / "\(" / "\)"
```

spec/message.abnf
eemeli: Suggestion added.

aphillips (approved):

spec/message.abnf
```diff
@@ -0,0 +1,57 @@
...
+selector = "{" [s] expression [s] "}"
+variant = when 1*(s key) [s] pattern
+key = nmtoken / literal / "*"
+
+placeholder = "{" [s] (expression / markup / markup-end) [s] "}"
+
+expression = ((literal / variable) [s annotation])
+           / annotation
+annotation = function *(s option)
+option = name [s] "=" [s] (literal / nmtoken / variable)
```
gibson042, line 18: What are the expected semantics of an optional value that is a `nmtoken` but not a `name`, e.g. `foo=1`?

spec/message.abnf
```diff
@@ -0,0 +1,57 @@
...
+name-start = ALPHA / "_"
+           / %xC0-D6 / %xD8-F6 / %xF8-2FF
+           / %x370-37D / %x37F-1FFF / %x200C-200D
+           / %x2070-218F / %x2C00-2FEF / %x3001-D7FF
+           / %xF900-FDCF / %xFDF0-FFFD / %x10000-EFFFF
+name-char = name-start / DIGIT / "-" / "." / %xB7
+          / %x0300-036F / %x203F-2040
+
+text-escape = "\\" / "\{" / "\}"
+literal-escape = "\\" / "\(" / "\)"
```
gibson042, line 55: I recognize that this is in conflict with https://github.com/unicode-org/message-format-wg/pull/347#discussion_r1119976120 , but these strings are easy to misinterpret if the reader is not being careful. I personally like the approach used by [JSON](https://www.rfc-editor.org/rfc/rfc8259.html#section-7) of naming `\`:
```suggestion
text-escape    = backslash ( backslash / "{" / "}" )
literal-escape = backslash ( backslash / "(" / ")" )

backslash = %x5C ; U+005C REVERSE SOLIDUS "\"
```

spec/message.abnf
```diff
@@ -0,0 +1,57 @@
...
+             / %x5D-D7FF      ; omit surrogates
+             / %xE000-10FFFF
+
+variable = "$" name
+function = ":" name
+markup-start = "+" name
+markup-end = "-" name
+
+name = name-start *name-char
+nmtoken = 1*name-char
```
gibson042, line 45:
```suggestion
name    = name-start *name-char ; matches XML https://www.w3.org/TR/xml/#NT-Name
nmtoken = 1*name-char           ; matches XML https://www.w3.org/TR/xml/#NT-Nmtokens
```

spec/message.abnf
```diff
@@ -0,0 +1,57 @@
...
+
+declaration = let s variable [s] "=" [s] "{" [s] expression [s] "}"
+body = pattern
+     / (selectors 1*([s] variant))
+
+pattern = "{" *(text / placeholder) "}"
+selectors = match 1*([s] selector)
+selector = "{" [s] expression [s] "}"
+variant = when 1*(s key) [s] pattern
+key = nmtoken / literal / "*"
```
gibson042, line 11:
```suggestion
pattern   = "{" *(text / placeholder) "}"
selectors = match 1*([s] selector)
selector  = "{" [s] expression [s] "}"
variant   = when 1*(s key) [s] pattern
key       = nmtoken / literal / "*"
```

spec/message.abnf
```diff
@@ -0,0 +1,57 @@
...
+
+placeholder = "{" [s] (expression / markup / markup-end) [s] "}"
```
gibson042, line 13: I think denormalization would help with readability here.
```suggestion
placeholder = "{" [s] expression [s] "}"
            / "{" [s] markup-start *(s option) [s] "}"
            / "{" [s] markup-end [s] "}"
```

And given the reuse of `"{" [s] expression [s] "}"`, it might make sense for that to have a name of its own.

gibson042 (approved):
This is great! I support an initial mechanical conversion to ABNF, but have some suggestions for followup improvements.

spec/message.abnf
alerque: This is definitely easier to interpret. I did a couple head scratches when reading it in the diff here earlier, this would have been much less mental load to interpret.

spec/message.abnf
```diff
@@ -0,0 +1,57 @@
+message = [s] *(declaration [s]) body [s]
+
+declaration = let s variable [s] "=" [s] "{" [s] expression [s] "}"
+body = pattern
+     / (selectors 1*([s] variant))
```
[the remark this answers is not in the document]
stasm: Can we discuss using `RWS` and `OWS` in a separate PR? They touch every production in the ABNF.

I'm against column alignment for as long as we expect the grammar to change. They generate needless diffs. Let's do it once when the grammar stabilizes.

spec/message.abnf
```diff
@@ -0,0 +1,57 @@
...
+           / %x370-37D / %x37F-1FFF / %x200C-200D
+           / %x2070-218F / %x2C00-2FEF / %x3001-D7FF
+           / %xF900-FDCF / %xFDF0-FFFD / %x10000-EFFFF
+name-char = name-start / DIGIT / "-" / "." / %xB7
+          / %x0300-036F / %x203F-2040
+
+text-escape = "\\" / "\{" / "\}"
+literal-escape = "\\" / "\(" / "\)"
+
+s = 1*(%x09 / %x0D / %x0A / %x20)
```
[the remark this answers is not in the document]
stasm: I agree that using the builtin productions makes this more readable. Can you edit your suggestion to not change `s` to `RWS` (and we'll handle that in another PR)?

spec/message.abnf
```diff
@@ -0,0 +1,57 @@
...
+expression = ((literal / variable) [s annotation])
+           / annotation
+annotation = function *(s option)
+option = name [s] "=" [s] (literal / nmtoken / variable)
+
+markup = markup-start *(s option)
+
+let = %x6C.65.74
+match = %x6D.61.74.63.68
+when = %x77.68.65.6E
```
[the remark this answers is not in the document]
stasm: It felt like a good idea to not require RFC 7405 just for 3 keywords.

spec/message.abnf
stasm: It parses as a literal, `"1"`. The implementation of `:func` can interpret it as a number if it makes sense to do so for the `foo` option.

spec/message.abnf
stasm: I thought that @eemeli's suggestion already improved the readability of these rules, but I'll agree that introducing `backslash` makes them even easier to understand. Thanks!

spec/message.abnf
stasm: Thanks for suggesting this. As I said above, I'd prefer to not column-align the ABNF for now because I don't think the names of production are final and I expect the next few weeks to bring a few changes.

spec/message.abnf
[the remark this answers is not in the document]
macchiati: Can we avoid the hex, and just have:

let = "let"
match = "match"
when = "when"

Far more readable, and less error prone (how many people would see that let = %x6C.65.75 is wrong??)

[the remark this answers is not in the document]
stasm: My understanding is that `let = "let"` would make the `let` keyword case-insensitive, due to how literal text strings work in ABNF. That's why we'd need the case-sensitive extension: `let = %s"let"`. See also: https://github.com/unicode-org/message-format-wg/pull/347/files#diff-a33e1c859f2aaad86b37ea3b3b5e8d45331199d6e879c5aba38eca2f23f01865.

macchiati:
I understand why now (ugly syntax from ABNF)

spec/message.abnf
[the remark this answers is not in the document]
gibson042:
```suggestion
s = 1*( SP / HTAB / CR / LF )
```

spec/message.abnf
gibson042: Hrm, so the value of an option is either a variable or a literal, but the literal can be implicit rather than quoted? Are `{:func foo=|1|}` and `{:func foo=1}` therefore indistinguishable?

aphillips: There's a super-subtle difference between `nmtoken` and `literal` values.

An `nmtoken` value might be validated at parse time and the values that can be present in an `nmtoken` are restricted vs. the values permitted in a literal. The use of numbers is fairly common in existing formatters, Cf. `Intl.NumberFormat` options such as `maximumSignificantDigits`. But other values have limited (and enumerated) values which might be validated at parse time.

A `literal` value probably is a parsing error (invalid argument) when the function wants a number or enumerated value. MF's options are untyped, but the underlying implementation might not be.

There may be a "tripping hazard" here for users who can't see the difference between:

`{:func symbol=US$}` (invalid, as `$` is reserved) and `{:func symbol=|US$|}`

spec/message.abnf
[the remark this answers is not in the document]
aphillips: This can merit a comment in the ABNF to prevent folks wondering, e.g.

```abnf
let   = %x6C.65.74        ; "let"
match = %x6D.61.74.63.68  ; "match"
when  = %x77.68.65.6E     ; "when"
```

spec/message.abnf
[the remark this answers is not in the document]
stasm: Thank you!

spec/message.abnf
eemeli: > Hrm, so the value of an option is either a variable or a literal, but the literal can be implicit rather than quoted?

Yes.

> Are `{:func foo=|1|}` and `{:func foo=1}` therefore indistinguishable?

As far as I recall we have not had an explicit discussion on this, but my position would be that during formatting the `:func` handler should not be able to distinguish these two from each other.

spec/message.abnf
```diff
@@ -0,0 +1,61 @@
...
+            / "{" [s] markup-start *(s option) [s] "}"
+            / "{" [s] markup-end [s] "}"
+
+expression = ((literal / variable) [s annotation])
+           / annotation
+annotation = function *(s option)
+option = name [s] "=" [s] (literal / nmtoken / variable)
+
+markup = markup-start *(s option)
+
```
eemeli, line 23:
```suggestion
```

With the change to `placeholder`, this is now unused.

stasm: Ah, good catch. This will need more changes to `syntax.md`, too. I'll take care of them in a few.

stasm:
Let me list the suggestions to the PR that I'd like to discuss separately, after it lands, so that we can keep track of them all easily.

* [ ] Perhaps merge `name` and `nmtoken`. Discussed by @aphillips in https://github.com/unicode-org/message-format-wg/pull/347#discussion_r1106092803.
* [ ] Decide whether boolean options for markup should be allowed, e.g. `{+checkbox checked}`. Suggested by @aphillips in https://github.com/unicode-org/message-format-wg/pull/347#discussion_r1106122827.
* [ ] Perhaps drop the `function` and `markup-start` productions. Suggested by @eemeli in https://github.com/unicode-org/message-format-wg/pull/347#discussion_r1119966919.
* [ ] Use `RWS` and `OWS`. Suggested by @gibson042 in https://github.com/unicode-org/message-format-wg/pull/347#discussion_r1120504708 and https://github.com/unicode-org/message-format-wg/pull/347#discussion_r1120522098.
* [ ] Column-align productions in the entire file, or at least inside blocks of related productions. Suggested by @gibson042 in https://github.com/unicode-org/message-format-wg/pull/347#discussion_r1120595343
* [ ] Decide whether `foo=x` and `foo=|x|` should have different AST representation. See https://github.com/unicode-org/message-format-wg/pull/347#discussion_r1120556456.

spec/message.abnf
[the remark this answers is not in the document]
stasm: Add the comment suggested by @aphillips in 4205cd2.

stasm:
I'm going to merge this now. Please feel free to suggest further improvements by opening new PRs. Please remember to not only edit `message.abnf`, but to also make the corresponding changes in `syntax.md`.

PR #355, opened 2023-02-17
Update examples to indent 'when' lines by one space

stasm:
I'd like to suggest that, when formatting messages over multiple lines, the _recommended_ formatting for `when` clauses be (a) each one on its own line, and (b) indented by one space. This hopefully can help discern the "logic" parts of complex messages (`let`, `match`) from the "content" parts (`when`).

    let $count = {$count :number fractionDigits=0}
    match {$count :plural}
     when 1 {You have one notification.}
     when * {You have {$count} notifications.}

I'm opening a PR to demonstrate what it would look like. This would be a recommendation, not a mandate. No changes to the grammar are required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,8 +25,8 @@ Messages can interpolate arguments formatted using _formatting functions_:
 Messages can define variants which correspond to the grammatical (or other) requirements of the language:
 
     match {$count :number}
-    when 1 {You have one notification.}
-    when * {You have {$count} notifications.}
+     when 1 {You have one notification.}
+     when * {You have {$count} notifications.}
     
 The message syntax is also capable of expressing more complex translation, for example:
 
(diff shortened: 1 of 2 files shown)


alerque:
Sug min 2spc.

mihnita:
> Sug min 2spc

I agree. 2 and 4 are the most common options I've seen.
They also show it is OK to have more than one space.

I sometimes do stuff like this:
```
  when   1 {You have one notification.}
  when one {You have {$count} notification.}
  when   * {You have {$count} notifications.}
```

It aligns the keys (when more than one), the messages, and shows that the number of spaces is not relevant.

eemeli:
I have no strong opinion on this, but I kinda like having the let/match/when all starting from the same column. If indenting, then 2 or 4 spaces would certainly be preferable to just one.

stasm:
My *slight* preference for 1 space is due to how the first `match` expression aligns vertically with the first `when` key.

aphillips:
I'm not excited by this. 

The multiline presentation is a convention we've adopted for our convenience. The real syntax, when normalized/minified, probably eats the newlines and most of the spaces. I think we should avoid doing too much to suggest that the spacing or alignment has any meaning.

Like @mihnita I often align columns inside the `when` statement visually and I'm okay with doing that in our examples. I could be okay with some indentation if the group agrees with @stasm's idea 🤷  

If we adopt this we should:

* Call out the convention in our spec's document conventions section
* Call out in the same place (and elsewhere _only as needed_) that the whitespacing has no meaning

alerque:
When whitespace is used to align columns, rather than to suggest any
semantic nesting relationship I'd much rather see it placed *after* the
shorter fields than before. At least for me this avoids the mental
suggestion (even if I know the whitespace is not semantic in the language)
that it somehow changes the meaning.

```
let   $count = {$count :number fractionDigits=0}
match {$count :plural}
when  1 {You have one notification.}
when  * {You have {$count} notifications.}
```

At least for me this instantly visually clues me in that the spacing is just an alignment convenience, something that I did not understand from the version in the initial post above.

stasm:
> The real syntax, when normalized/minified, probably eats the newlines and most of the spaces.

I can imagine at least some developers formatting messages over multiple lines, and that would still be "real" syntax, as in, used in the real life. At least that's what I've been doing when writing tests for my experiments.

> I think we should avoid doing too much to suggest that the spacing or alignment has any meaning.

This (and the last comment from @alerque) is an interesting point which I didn't consider enough before. My attempt was aimed at improving legibility to help people learn the syntax, but I can see how this can actually go too far.

Also, seeing how much different opinions there are about how to format messages, I'm tempted to close this PR as a wontfix. I was expecting much less discussion about my proposal :) We don't have to recommend any particular formatting style if there isn't one obvious choice.

Is the following not controversial enough? No indents, no padding, no alignment.

```
let $count = {$count :number fractionDigits=0}
match {$count :plural}
when 1 {You have one notification.}
when * {You have {$count} notifications.}
```

aphillips:
@stasm Do you intend to abandon this PR?

stasm:
Yes, thanks for asking. I didn't check back until now. I can see three 👍 under my last comment. Thanks for the discussion!

PR #357, opened 2023-02-21
Define new error "Unknown Markup"

aphillips:
From a discussion about the ABNF with @eemeli I noticed that we don't have an error for when `markup` is not recognized. Proposing a fix.

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -130,7 +130,13 @@ These are divided into the following categories:
     
     ```
     {There are {+foo option=bar}no{-foo} wildebeest here}
+    ```
+
+    ```
     {There are {+foo option=bar}no wildebeest here}
+    ```
+
+    ```
     {There are no{-foo} wildebeest here}
     ```
 
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -121,6 +121,19 @@ These are divided into the following categories:
     when * {The value is not one.}
     ```
 
+  - **Unknown Markup errors** occur when an Expression includes
+    a reference to a markup handler which cannot be resolved.
+    
+    For example, attempting to format the following messages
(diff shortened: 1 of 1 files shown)


mihnita:
I think that markup is still not agreed on.
So adding some error for it only adds stuff to cleanup after.

mihnita:
Markup open issues:
* https://github.com/unicode-org/message-format-wg/issues/241
* https://github.com/unicode-org/message-format-wg/issues/262
* https://github.com/unicode-org/message-format-wg/issues/238

And to show that it was not agreed on, see PR https://github.com/unicode-org/message-format-wg/pull/283 in Jun 21, 2022:
> I will approve it to move things forward and not have these many open PRs.
> It does not mean I agree with this syntax, so it should not be hold against me 1 month from now ("but you approved X")
> There are still at least 2 open issues arguing about how to best handle "markup elements"

spec/formatting.md
```diff
@@ -121,6 +121,19 @@ These are divided into the following categories:
     when * {The value is not one.}
     ```
 
+  - **Unknown Markup errors** occur when an Expression includes
+    a reference to a markup handler which cannot be resolved.
+    
+    For example, attempting to format the following messages
+    MUST result in an Unknown Markup error if done within a context that
+    does not provide a handler for the markup `{+foo}` or `{-foo}`
```
eemeli, line 129: Technically, the braces aren't a part of the markup.
```suggestion
    does not provide a handler for the markup `+foo` or `-foo`
```

eemeli (approved):
This looks good, and well in line with the Unknown Function error.

Some nitpicks inline.

aphillips:
close in telecon 2023-05-22

PR #362, opened 2023-03-04
Provide selection matching options (best vs. first match)

aphillips:
Replacing the previous PR, which got tangled up with other changes on my fork. We need this doc for discussion.

### DIFF
--- a/exploration/selection-matching-options.md
+++ b/exploration/selection-matching-options.md
@@ -225,16 +225,34 @@ This exposes developers and translators to managing the complexity versus having
 
 #### Are there other complex matching cases? Or is `plural` everything?
 
-Currently there are no other complex rule-based selectors in ICU. However, there are a number of cases where complex matching might come into play. Some potential examples:
+Currently there are no other complex rule-based selectors in ICU. However, there are a number of cases where complex matching might come into play. The criteria for it being a complex match are:
+
+* The selector generates different output by locale (the way that plural generates `few` and `many` for e.g. Polish or fails to generate `one` in Japanese). If the states vary according to something other than locale all of the states have to be accounted for in the `root` locale resource and all translations (so it's not "complex" any more); or:
+* The selector can generate more than one match for the same value (compare with plural `=1` and `one`), with varying quality of match.
+* Any combination of the above.
+
+The key thing here is that the static text produced by the translator needs to reflect the grammatical needs of the language and depends on knowing something about the (invisible) value being inserted at runtime.
+
+Some potential examples (and this is "thinking out loud"):
 
 1. **Date/time based selection.** Date/time types, including the newer Temporal types, can present complex matching needs. While _incremental time_ values (such as `java.time.Instant`, `java.util.Date`, or JavaScript's `Date`) can resolve every field and be cast to any time zone, Other types, such as `java.time.ZonedDate`, are incomplete. There are different calendars that can affect presentation and selection as well. Some cases for complex time selection include:
    * **Relative time formats.** The values available (such as `yesterday`, `tomorrow`, `day after tomorrow`) vary by locale. 
-   * **Phrasing variations.** For the time `8:30` an American would say "_eight thirty_", while a Brit might say "_half nine_".
    * **Periodic time formats.** Recurring values might require message selection.
 
-1. **Measurement selection.** Selectors based on unit of measurement sometimes vary--even based on application-specific requirements. For example, turn-by-turn directions in mapping applications will choose between `km` and `m` (or `miles` and `feet`) based on some arbitrary distance, e.g. `0.5 miles` (2640 feet) vs. `1500 feet`
+1. **Gender or part-of-speech selection.** Grammatical gender is strongly linked to language and varies by language--very much like plurals. These types of selection might not have the multiple selection quirks of plurals, but will have varying shape by locale. 
 
-1. **Gender or part-of-speech selection.** Grammatical gender is strongly linked to language and varies by language--very much like plurals. These types of selection might not have the multiple selection quirks of plurals, but will have varying shape by locale.
+   For example, I built a "product name format" function into Amazon's devices. Each product knew (in each supported locale) its generic, short, medium, long, and full name, and each product's name could vary in gender/count/etc. per language. That is, the generic might be a "tablet" or "TV" (or whatever) and then the e.g. tablet might be called a "Fire", a "Fire 8 HDX", etc. 
+   
+   The software doesn't know which device it'll be built into (actually, it's built into all of them), so the formatter needs to select the correct pattern string according to device it is in at runtime. Rather than build separate strings for every device, we generated variations based on the (smaller) set of grammar variations per-locale. A simple message like `The {whatever} is ready` in French might look sort of like the following (in our syntax) in a French locale (and I'm omitting for clarity such things as enclitic handling, e.g. when it's `l'ordinateur` not `le ordinateur`):
+
+```
+match {:gender $product}
+when masculine {Le {:product format=generic} est prêt.}  ; le téléphone est prêt
+when feminine {La {:product format=generic} est prête.}  ; la télévision est prête
+when * {L'appareil est prêt.} 
+```
+
+   Notice that it isn't just the article that changes. And notice that the list of enumerated values changes by language (so German has three noun genders while English mostly has one)
 
 1. **Application specific selection.** Developers may need to write selectors with varying degrees of selection. For example, one might have a message that varies by category and then, for specific items, by sub-category:
 
--- a/exploration/selection-matching-options.md
+++ b/exploration/selection-matching-options.md
@@ -93,7 +93,7 @@ Best-Match selection evaluates the full list of _keys_ and selects the _variant_
 
 ### Sorted
 
-Sorted Matching evaluates the full list of _keys_ by sorting the matrix. Each _selector_ provides a "comparator" for values in its column (such as computing a weight for the value in its column). Rows that contain a non-matching value for any selector are eliminated as potential matches. The default value `*` always matches. The highest ranking _key_ is returned as the _pattern_. Ties are broken by column. If no matching row is found, returns an error.
+Sorted Matching evaluates the full list of _keys_ by sorting the matrix. Each _selector_ provides a "comparator" for values in its column (such as computing a weight for the value in its column). Rows that contain a non-matching value for any selector are eliminated as potential matches. The default value `*` always matches. Ordering is maintained for preceding columns, that is, _selector_ number 2 can only reorder items whose _selector_ number 1 key match. The highest ranking _key_ is returned as the _pattern_. Ties are broken by column. If no matching row is found, returns an error.
 
 **Pros**
 (All of the pros of best match plus:)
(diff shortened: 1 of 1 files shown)


macchiati (commented):
looks good

exploration/selection-matching-options.md
```diff
@@ -0,0 +1,512 @@
...
+We are discussing whether to change First-Match to another value. Currently voting looks like:
+
+| Person | Supports |
+|---|---|
+| APP | C-F, then B-M, no F-M |
+| STA | F-M, then B-M, the C-F with optional \* |
+| SCL | F-M, then C-F, then optional \* |
+| RGN | C-F optional \*, F-M, CF required \* |
+| MIH | B-M, C-F lexical sort, C-F required \* |
+| EAO | C-F, B-M (distant 2nd), C-F required \* behind that |
```
eemeli, line 14:
```suggestion
| EAO | C-F optional \*, F-M, C-F required \* |
```

eemeli (commented):
Added a bunch of syntax corrections, and a few other inline comments.

At least currently the "Sorted" and "Column-First" choices that the document presents are two implementations of the exact same algorithm. They should probably be merged, or is there some difference that I'm missing?

I don't think I understand the argument about best-match allowing for a smaller number of segments needing translation when new variants are added.

Let's say that we start with this message:
```
match {$count :plural}
when one {You have {$count} new message.}
when * {You have {$count} new messages.}
```

Now, let's say that a developer adds a `0` variant. With first-match, it needs to go first; with the others its position is not relevant:

```
match {$count :plural}
when 0 {You have no new messages.}
when one {You have {$count} new message.}
when * {You have {$count} new messages.}
```

Once this new version of the message has been formulated, how is the need to either translate just the "You have no new messages." segment or the whole message in any way dependent on how selection happens?

exploration/selection-matching-options.md
```diff
@@ -0,0 +1,512 @@
...
+
+First-Match selection evaluates the list of _keys_ row-by-row and selects the first _variant_ that matches the list of _selectors_ as the _pattern_.
+
+In the example message, the _variants_ are in a canonical order, so first-match produces the same order as best-match does.
+
+**Pros**
++ Allows developers to control the order of selection.
++ Allows translators to tailor the order of selection.
++ Can visually inspect match order.
++ May be more efficient when perfoming match (??)
```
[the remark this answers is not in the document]
aphillips: These would be separate arguments. Efficiency is about match speed/processing.

I can put "simplest" into this list separately, although, in my mind, it isn't a criterion on which *I* would base *my* choice.

exploration/selection-matching-options.md
```diff
@@ -0,0 +1,512 @@
...
+
+**Cons**
+- Requires developers to specify _variants_ in the correct order.
+- Requires translators to tailor the order of _variants_ if this is different from the source.
```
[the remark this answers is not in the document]
aphillips: No, it refers to the need for translators to tailor the order of the variants in their language/locale where required.

For example, I could have this for English:

```
when 0 { }
when one { }
when 2 { }
when * { }
```

In Russian the "exploded" list might be:

```
when 0 { }
when one { }
when 2 { }
when few { }
when many { }
when * { }
```

... but Russian requires that the `2` key appear before the `few`/`many` **and** `one` keys (so that it is evaluated before one of those other keys "capture" the number `2`. (In `ru` the `one` plural rule fires for numbers like 21, 31, 41, etc.)

exploration/selection-matching-options.md
```diff
@@ -0,0 +1,512 @@
...
+- Requires all translation tooling and runtime processing to preserve the order of the _variants_
+- Translation tools that "explode" the selection matrix to support languages with additional needs (such as the way that the `pl` locale requires keywords `few` and `many` for plurals whereas the `en` locale does not) to order generated _variants_ correctly and to interpolate the developer's intent.
```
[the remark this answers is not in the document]
aphillips: You left out the interpolation of the developer's intent, which I think is important. When the list is ordered manually, it is not always clear when the developer is being "clever" somehow.

exploration/selection-matching-options.md
```diff
@@ -0,0 +1,512 @@
...
+
+
+As an aside, how does the above express the `when` clause for the value `2.00`? It can't use a literal (what if the decimal separator were `,`!!) and the _nmtoken_ `2.00` could be complicated to handle? Also, does `2` match?
+
+#### How does this compare to programming language constructs (such as switch)?
+
+It's difficult to say if the MF2 `match` statement should work like familiar selection methods in programming languages. Internationalization APIs, such as resource managers, MF1, and date/number skeletons have tended towards "do what I want", hiding the need for both developers and translators to know about cultural and lingusitic variation and account for it in code. Modern I18N APIs hide most of this complexity. Some of the analogous cases in I18N APIs are:
+* resource fallback with sparse population
+* skeletons for dates, such as (`yyyyMMddjm`) do not require translators to touch "picture strings" (such as `MM/dd/yyyy HH:mm a`) to handle the time or date separators `/` and `:`, the use of 24-hour time vs. 12-hour time, the order of the fields
```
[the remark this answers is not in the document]
aphillips: I'm not following your comment here? Are you suggesting adding a bullet item for locale negotiation or are you commenting on skeletons?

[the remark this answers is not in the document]
eemeli: I meant that adding a bullet on locale negotation or [language matching](https://www.unicode.org/reports/tr35/tr35.html#LanguageMatching) might be appropriate, as it's also dealing with selecting the best choice from a list, given a input list of preferred locales.

exploration/selection-matching-options.md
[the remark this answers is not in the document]
eemeli: Ah, ok. I have been presuming that this "explosion" is mostly handled by tooling/automation, which would know how to sort the variants.

exploration/selection-matching-options.md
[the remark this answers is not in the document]
aphillips: I'll add that.

exploration/selection-matching-options.md
```diff
@@ -0,0 +1,547 @@
...
+- Requires all translation tooling and runtime processing to preserve the order of the _variants_
+- Translation tools that "explode" the selection matrix to support languages with additional needs (such as the way that the `pl` locale requires keywords `few` and `many` for plurals whereas the `en` locale does not) must order the generated _variants_ correctly.
+
+## Best-Match
+
+Best-Match selection evaluates the full list of _keys_ and selects the _variant_ that, according to the matching strategy used, "best matches" the list of _selectors_ as the _pattern_. There are three best-match selection strategies in this document:
+
+1. Sorted Matching
+1. Column-First with required default `*`
+1. Column-First without required default `*`
```
stasm, line 83:
```suggestion
1. Column-First with required default `*`
1. Column-First without required default `*`
1. Scored Best Matching
```

exploration/selection-matching-options.md
```diff
@@ -0,0 +1,547 @@
...
++ Variants can be written in any order and produce a consistent result.
++ Selector developers can write complex matches that produce different quality matches for the same value. For example, `{|1| :plural}` matches both the variant `1` and the variant `one`, but prefers the value `1`. The plural _selector_ does not need to communicate with the other _selectors_ in order to arrive at the best matching pattern.
++ Translators do not need to worry about the order of variants or need to reorder variants (which can be difficult to do when only the translation segment for the pattern is shown or when only a changed or generated _variant_ is exposed to translation.
++ Translation tools do not have to preserve the order of _variants_ and are free to send only the translatable segment (the pattern) for translation.
+
+**Cons**
+- Developers cannot override the order that the _selector_ provides unless this is exposed as a feature of the given _selector_.
+- More complex matching implementation; may be slower?
+
+### Sorted
```
stasm, line 95: Is this section meant as the base on which the following ones build? Or just some of the following? In any case, I was a bit confused by the fact that it's presented as a fourth alternative while the list above mentions only three.

exploration/selection-matching-options.md
```diff
@@ -0,0 +1,547 @@
...
++ Translators do not need to worry about the order of variants or need to reorder variants (which can be difficult to do when only the translation segment for the pattern is shown or when only a changed or generated _variant_ is exposed to translation.
++ Translation tools do not have to preserve the order of _variants_ and are free to send only the translatable segment (the pattern) for translation.
++ Easier to evaluate visually than sorting strategies.
+
+**Cons**
+- Developers cannot override the order that the _selector_ provides unless this is exposed as a feature of the given _selector_.
+- Can require more processing than First-Match
+
+
+### Column-First with Optional `*`
```
stasm, line 197: The question of whether the `*` is required or optional seems related, but orthogonal. I think it's a valid question for all strategies presented. Perhaps move it to its own section? (In this PR or later.)

exploration/selection-matching-options.md
```diff
@@ -0,0 +1,547 @@
...
+```
+
+This exposes developers and translators to managing the complexity versus having the API take care of it.
+
+
+#### Are there other complex matching cases? Or is `plural` everything?
+
+Currently there are no other complex rule-based selectors in ICU. However, there are a number of cases where complex matching might come into play. The criteria for it being a complex match are:
+
+1. The selector generates different output by locale (the way that plural generates `few` and `many` for e.g. Polish or fails to generate `one` in Japanese). If the states vary according to something other than locale all of the states have to be accounted for in the `root` locale resource and all translations (so it's not "complex" any more); or:
```
stasm, line 402: I don't really understand this use-case. How does this result in complex matching?

exploration/selection-matching-options.md
```diff
@@ -0,0 +1,547 @@
...
+2. The selector can generate more than one match for the same value (compare with plural `1` and `one`), with varying quality of match.
+3. Any combination of the above.
+
+The key thing here is that the static text produced by the translator needs to reflect the grammatical needs of the language and depends on knowing something about the (invisible) value being inserted at runtime.
+
+Some potential examples of (1) (and this is "thinking out loud"):
+
+1. **Date/time based selection.** Date/time types, including the newer Temporal types, can present complex matching needs. While _incremental time_ values (such as `java.time.Instant`, `java.util.Date`, or JavaScript's `Date`) can resolve every field and be cast to any time zone, Other types, such as `java.time.ZonedDate`, are incomplete. There are different calendars that can affect presentation and selection as well. Some cases for complex time selection include:
+   * **Relative time formats.** The values available (such as `yesterday`, `tomorrow`, `day after tomorrow`) vary by locale. Here's one example in the [German CLDR charts](https://unicode-org.github.io/cldr-staging/charts/latest/summary/de.html#1d45310cbcf1b2e5)
```
stasm, line 411: Why is it important that the values very by locale? I assume the issue is that the value of `1` matches both `tomorrow` and `in 1 day`, but that seems to be the manifestation of the second criteria above? ("The selector can generate more than one match for the same value.")

stasm (commented):
Thank you for writing this, it helps a lot. Since it's in `exploration`, I'm assuming we can merge it before we reach consensus on the question itself?

exploration/selection-matching-options.md
aphillips: This is *also* a manifestation of (1). English has words for `tomorrow` and `yesterday`, but used counted phrases (`in x days`, `x days ago`) for other values. German (and some other languages) have _words_ for two days ago/from now, e.g. `vorgestern`. So the relative time format used varies by locale.

eemeli: No need to look beyond the worlds of English even to find "ereyesterday" and "overmorrow". :)

aphillips:
From the 2023-03-31 teleconference, merging this as reference material for the future. We can make additional edits to this once we land the consensus changes to the spec.

PR #364, opened 2023-03-08
Replace `nmtoken` with `unquoted`

eemeli:
Currently, a valid `nmtoken` like `42` or `foo` can be used directly as an option value `key=42 bar=foo`, but needs to be quoted when used as an argument: `{|42| :number|`, `{|foo|}`. This is a bit of a wtf, and there should be no need for this difference.

This is a relic from when the syntax considered a placeholder with just an `nmtoken` to be a markup-start, but this was changed in #283. This PR allows for all `nmtoken` excpet for ones starting with a `-` to be used as an expression argument.

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -363,7 +363,7 @@ These are divided into the following categories:
     uses a `:plural` selector function which requires its input to be numeric:
 
     ```
-    match {horse :plural}
+    match {|horse| :plural}
     when 1 {The value is one.}
     when * {The value is not one.}
     ```
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -445,7 +445,7 @@ and ends with U+007D RIGHT CURLY BRACKET `}`.
 Between the brackets, the following contents are used:
 
 - Expression with Literal Operand: U+007C VERTICAL LINE `|`
-  followed by the Quoted or Unquoted value,
+  followed by the value of the Literal,
   and then by U+007C VERTICAL LINE `|`
 
   Examples: `{|your horse|}`, `{|42|}`
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -345,7 +345,7 @@ These are divided into the following categories:
     does not provide for the function `:func` to be successfully resolved:
 
     ```
-    {The value is {|horse| :func}.}
+    {The value is {horse :func}.}
     ```
 
     ```
@@ -363,7 +363,7 @@ These are divided into the following categories:
     uses a `:plural` selector function which requires its input to be numeric:
 
     ```
-    match {|horse| :plural}
+    match {horse :plural}
     when 1 {The value is one.}
     when * {The value is not one.}
     ```
@@ -390,7 +390,7 @@ These are divided into the following categories:
      an option `field` to be provided with a string value,
 
   ```
-  {Hello, {|horse| :get field=name}!}
+  {Hello, {horse :get field=name}!}
   ```
 
   ```
@@ -445,10 +445,10 @@ and ends with U+007D RIGHT CURLY BRACKET `}`.
 Between the brackets, the following contents are used:
 
 - Expression with Literal Operand: U+007C VERTICAL LINE `|`
-  followed by the value of the Literal,
+  followed by the Quoted or Unquoted value,
   and then by U+007C VERTICAL LINE `|`
 
-  Examples: `{|horse|}`, `{|42|}`
+  Examples: `{|your horse|}`, `{|42|}`
 
 - Expression with Variable Operand: U+0024 DOLLAR SIGN `$`
   followed by the Variable Name of the Operand
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -35,8 +35,9 @@ quoted-char = %x0-5B         ; omit \
             / %x7D-D7FF      ; omit surrogates
             / %xE000-10FFFF
 
-unquoted = unquoted-start *name-char ; based on https://www.w3.org/TR/xml/#NT-Nmtoken,
-                                     ; but cannot start with U+002D HYPHEN-MINUS or U+003A COLON ":"
+; based on https://www.w3.org/TR/xml/#NT-Nmtoken,
+; but cannot start with U+002D HYPHEN-MINUS or U+003A COLON ":"
+unquoted = unquoted-start *name-char
 unquoted-start = name-start / DIGIT / "."
                / %xB7 / %x300-36F / %x203F-2040
 
@@ -52,8 +53,9 @@ reserved-char  = %x00-08        ; omit HTAB and LF
                / %x7E-D7FF      ; omit surrogates
                / %xE000-10FFFF
 
-name = name-start *name-char ; based on https://www.w3.org/TR/xml/#NT-Name,
-                             ; but cannot start with U+003A COLON ":"
+; based on https://www.w3.org/TR/xml/#NT-Name,
+; but cannot start with U+003A COLON ":"
+name = name-start *name-char
 name-start = ALPHA / "_"
            / %xC0-D6 / %xD8-F6 / %xF8-2FF
            / %x370-37D / %x37F-1FFF / %x200C-200D
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -7,12 +7,15 @@ body = pattern
 pattern = "{" *(text / expression) "}"
 selectors = match 1*([s] expression)
 variant = when 1*(s key) [s] pattern
-key = nmtoken / literal / "*"
+key = literal / "*"
 
 expression = "{" [s] (((literal / variable) [s annotation]) / annotation) [s] "}"
 annotation = (function *(s option)) / reserved
 
-option = name [s] "=" [s] (literal / nmtoken / variable)
+literal = quoted / unquoted
+variable = "$" name
+function = (":" / "+" / "-") name
+option = name [s] "=" [s] (literal / variable)
 
 ; reserved keywords are always lowercase
 let   = %x6C.65.74        ; "let"
@@ -26,14 +29,16 @@ text-char = %x0-5B         ; omit \
           / %x7E-D7FF      ; omit surrogates
           / %xE000-10FFFF
 
-literal      = "|" *(literal-char / literal-escape) "|"
-literal-char = %x0-5B         ; omit \
-             / %x5D-7B        ; omit |
-             / %x7D-D7FF      ; omit surrogates
-             / %xE000-10FFFF
+quoted      = "|" *(quoted-char / quoted-escape) "|"
+quoted-char = %x0-5B         ; omit \
+            / %x5D-7B        ; omit |
+            / %x7D-D7FF      ; omit surrogates
+            / %xE000-10FFFF
 
-variable = "$" name
-function = (":" / "+" / "-") name
+unquoted = unquoted-start *name-char ; based on https://www.w3.org/TR/xml/#NT-Nmtoken,
+                                     ; but cannot start with U+002D HYPHEN-MINUS or U+003A COLON ":"
+unquoted-start = name-start / DIGIT / "."
+               / %xB7 / %x300-36F / %x203F-2040
 
 ; reserve additional sigils for future use
 reserved       = reserved-start reserved-body
@@ -47,18 +52,18 @@ reserved-char  = %x00-08        ; omit HTAB and LF
                / %x7E-D7FF      ; omit surrogates
                / %xE000-10FFFF
 
-name    = name-start *name-char ; based on https://www.w3.org/TR/xml/#NT-Name, but cannot start with U+003A COLON ":"
-nmtoken = 1*name-char           ; equal to https://www.w3.org/TR/xml/#NT-Nmtoken
+name = name-start *name-char ; based on https://www.w3.org/TR/xml/#NT-Name,
+                             ; but cannot start with U+003A COLON ":"
 name-start = ALPHA / "_"
            / %xC0-D6 / %xD8-F6 / %xF8-2FF
            / %x370-37D / %x37F-1FFF / %x200C-200D
            / %x2070-218F / %x2C00-2FEF / %x3001-D7FF
            / %xF900-FDCF / %xFDF0-FFFD / %x10000-EFFFF
-name-char = name-start / DIGIT / "-" / "." / ":"
-          / %xB7 / %x0300-036F / %x203F-2040
+name-char  = name-start / DIGIT / "-" / "." / ":"
+           / %xB7 / %x300-36F / %x203F-2040
 
 text-escape     = backslash ( backslash / "{" / "}" )
-literal-escape  = backslash ( backslash / "|" )
+quoted-escape   = backslash ( backslash / "|" )
 reserved-escape = backslash ( backslash / "{" / "|" / "}" )
 backslash       = %x5C ; U+005C REVERSE SOLIDUS "\"
 
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -478,7 +478,8 @@ The _name_ token is used for variable names (prefixed with `$`),
 function names (prefixed with `:`, `+` or `-`),
 as well as option names.
 It is based on XML's [Name](https://www.w3.org/TR/xml/#NT-Name),
-with the restriction that it MUST NOT start with an ASCII digit and certain basic combining characters.
+with the restriction that it MUST NOT start with `:`,
+as that would conflict with _function_ start characters.
 Otherwise, the set of characters allowed in names is large.
 
 ```abnf
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -22,7 +22,8 @@
        1. [Reserved Sequences](#reserved)
 1. [Tokens](#tokens)
    1. [Keywords](#keywords)
-   1. [Text and Literals](#text-and-literals)
+   1. [Text](#text)
+   1. [Literals](#literals)
    1. [Names](#names)
    1. [Escape Sequences](#escape-sequences)
    1. [Whitespace](#whitespace)
@@ -307,7 +308,7 @@ The key `*` is a "catch-all" key, matching all selector values.
 
 ```abnf
 variant = when 1*(s key) [s] pattern
-key = nmtoken / literal / "*"
+key     = literal / "*"
 ```
 
 A _well-formed_ message is considered _valid_ if the following requirements are satisfied:
@@ -357,21 +358,29 @@ _Functions_ do not accept any positional arguments other than the _literal_ or _
 ```abnf
 expression = "{" [s] (((literal / variable) [s annotation]) / annotation) [s] "}"
 annotation = (function *(s option)) / reserved
-option = name [s] "=" [s] (literal / nmtoken / variable)
+option = name [s] "=" [s] (literal / variable)
 ```
 
 Expression examples:
 
 ```
-{|1.23|}
+{1.23}
 ```
 
 ```
-{|1.23| :number maxFractionDigits=1}
+{|-1.23|}
+```
+
+```
+{1.23 :number maxFractionDigits=1}
 ```
 
 ```
 {|Thu Jan 01 1970 14:37:00 GMT+0100 (CET)| :datetime weekday=long}
+```
+
+```
+{|My Brand Name| :linkify href=|foobar.com|}
 ```
 
 ```
@@ -410,22 +419,20 @@ The grammar defines the following tokens for the purpose of the lexical analysis
 ### Keywords
 
 The following three keywords are reserved: `let`, `match`, and `when`.
+Reserved keywords are always lowercase.
 
 ```abnf
-; reserved keywords are always lowercase
 let   = %x6C.65.74        ; "let"
 match = %x6D.61.74.63.68  ; "match"
 when  = %x77.68.65.6E     ; "when"
 ```
 
-### Text and Literals
+### Text
 
-_Text_ is the translatable content of a _pattern_, and _Literal_ is used for matching
-variants and providing input to expressions.
-Any Unicode code point is allowed in either, with the exception of
-the relevant delimiters (`{` and `}` for Text, `|` for Literal),
-`\` (which starts an escape sequence), and
-surrogate code points U+D800 through U+DFFF (which cannot be encoded into UTF-8).
+_Text_ is the translatable content of a _pattern_.
+Any Unicode code point is allowed,
+except for surrogate code points U+D800 through U+DFFF.
+The characters `\`, `{`, and `}` MUST be escaped as `\\`, `\{`, and `\}`.
 
 All code points are preserved.
 
@@ -438,12 +445,31 @@ text-char = %x0-5B         ; omit \
           / %xE000-10FFFF
 ```
 
+### Literals
+
+_Literal_ is used for matching variants and providing input to _expressions_.
+_Quoted_ literals may include content with any Unicode code point,
+except for surrogate code points U+D800 through U+DFFF.
+The characters `\` and `|` MUST be escaped as `\\` and `\|`.
+_Unquoted_ literals have a much more restricted range that
+is intentionally close to the XML's [Nmtoken](https://www.w3.org/TR/xml/#NT-Nmtoken),
+with the restriction that it MUST NOT start with `-` or `:`,
+as those would conflict with _function_ start characters.
+
+All code points are preserved.
+
 ```abnf
-literal = "|" *(literal-char / literal-escape) "|"
-literal-char = %x0-5B         ; omit \
-             / %x5D-7B        ; omit |
-             / %x7D-D7FF      ; omit surrogates
-             / %xE000-10FFFF
+literal = quoted / unquoted
+
+quoted         = "|" *(quoted-char / quoted-escape) "|"
+quoted-char    = %x0-5B         ; omit \
+               / %x5D-7B        ; omit |
+               / %x7D-D7FF      ; omit surrogates
+               / %xE000-10FFFF
+
+unquoted       = unquoted-start *name-char
+unquoted-start = name-start / DIGIT / "."
+               / %xB7 / %x300-36F / %x203F-2040
 ```
 
 ### Names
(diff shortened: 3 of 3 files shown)


aphillips:
You're overlooking my concurrent PR about reserving the other ASCII symbols as potential sigals. This would have to be changed to account for that.

You're also overlooking that `nmtoken` is not a synonym for `literal`. It has a different and more restrictive namespace which applies to function authors. It's meant to allow enumerated values (keywords) to be used in passing into functions. The `nmtoken` cannot contain, for example, spaces or non-namelike characters. The fact that there is overlap with literal is not the same thing as saying that the `nmtoken` ***is*** a literal.

As a reader, I'm also not sure what an "argtoken" represents in the syntax? I like that the current syntax has production names that correspond to functionality (it's easier to understand as an MF2 author what to do). I would suggest that you rename `argtoken` more like what it's function is in the syntax. Perhaps `unquoted-literal`? With that thought, I would then probably take it the next step and redefine `literal` in a way that makes it either an "unquoted" literal or a quoted one. E.g:

```abnf
literal = `|` ( literal-char / literal-escape)* `|`
        / unquoted-start unquoted-char*     ; to be unquoted a literal cannot start/end/contain some characters
```

stasm:
A few thoughts:

* This should be carefully considered together with #360, as @aphillips suggests. The more characters we reserve in #360, the more exceptions we will require when using literals in argument positions.
* I recall an argument about designing MF2 syntax to avoid the unquoted literal (`{foo}`) because it can be confused with the placeholder name in MF1. Is this still something we're concerned about?
* This is also something we can easily do later.
* What are the primary use-cases for literals in argument positions?
  * Number literals? Do we really expect things like `{5 :number}`, or are they rather academic?
  * Markup element names, if no special markup syntax is available? `{em :html.open}`
  * Message identifiers for message references provided via custom functions? `{app.menu.edit.select-all :msgref}`
* If the primary use-case is _names_, perhaps we should consider `(literal / name / variable)` as the operand?

aphillips:
@stasm mentioned:

> I recall an argument about designing MF2 syntax to avoid the unquoted literal ({foo}) because it can be confused with the placeholder name in MF1. Is this still something we're concerned about?

I don't think it can be a consideration. An MF1 pattern has to somehow acquire `{`/`}` around it to become an MF2 pattern. Presumably such a translation would also replace `{foo}` with `{$foo}`. If you forgot to do that, the result would not be syntactically wrong but not very useful (hopefully you would notice in testing the pattern...?)

> What are the primary use-cases for literals in argument positions?
>
>*    Number literals? Do we really expect things like {5 :number}, or are they rather academic?

Number literals would be useful for getting localized number formatting and `formatToParts` functionality on what are otherwise hardcoded numeric values.

>*    Markup element names, if no special markup syntax is available? {em :html.open}

We really really need to discuss the approach to markup.

>*    Message identifiers for message references provided via custom functions? {app.menu.edit.select-all :msgref}

That's a good case. The identifier could be quoted as a literal if the ID did not match our syntactical restrictions. This would make the production:

```abnf
expression = (literal / variable / name) [s annotation]) / annotation
```

The reason to have a `name` there is the same as having a number or date or other value there: the value is hardcoded in a locale-neutral manner and will be formatted at runtime, but you want the translator to know what the value is for the purposes of translation.

Perhaps examples would look like:

```
You have {5 :number} days to return an item before your account will be charged.
Your {tv :productName format=short} is not connected to an antenna.
We will be closed on {2023-08-01 :date} to observe International Mahjong Day.
```

eemeli:
> @aphillips:
> You're overlooking my concurrent PR about reserving the other ASCII symbols as potential sigals. This would have to be changed to account for that.

I don't think so? As far as I can tell, none of the characters that have been considered for reservation in #360 are valid in `nmtoken` or `argtoken`.

> You're also overlooking that `nmtoken` is not a synonym for `literal`. It has a different and more restrictive namespace which applies to function authors. It's meant to allow enumerated values (keywords) to be used in passing into functions. The `nmtoken` cannot contain, for example, spaces or non-namelike characters. The fact that there is overlap with literal is not the same thing as saying that the `nmtoken` _**is**_ a literal.

Hmm. Perhaps my wording was a bit awkward somehow? I agree with you about `nmtoken` having a clearly more restricted namespace, and I'm not trying to say that it's a `literal`.

> As a reader, I'm also not sure what an "argtoken" represents in the syntax? I like that the current syntax has production names that correspond to functionality (it's easier to understand as an MF2 author what to do). I would suggest that you rename `argtoken` more like what it's function is in the syntax. [...]

The intent was for `argtoken` to be understood as a token that's valid as an argument, much like `nmtoken` is a token that's valid as an option value. We could even merge these into one, by forbidding an `nmtoken` from starting with a `-` character.

I don't really have any strong opinion about these names, and would be happy to update them as necessary.

> @stasm:
> What are the primary use-cases for literals in argument positions?
> * Number literals? Do we really expect things like `{5 :number}`, or are they rather academic?
> * Markup element names, if no special markup syntax is available? `{em :html.open}`
> * Message identifiers for message references provided via custom functions? `{menu.edit.select-all :msgref}`
>
> If the primary use-case is _names_, perhaps we should consider `(literal / name / variable)` as the operand?

> @aphillips:
> The reason to have a `name` there is the same as having a number or date or other value there: the value is hardcoded in a locale-neutral manner and will be formatted at runtime, but you want the translator to know what the value is for the purposes of translation.

I think all of the use cases in the preceding two messages are in general valid and valuable. `name` would be sufficient for all but numbers and dates, but it'd be great not needing quotes for those either.

stasm:
> We could even merge these into one, by forbidding an `nmtoken` from starting with a `-` character.

Aligning our `nmtoken` to be at least as "wide" as the XML's `Nmtoken` gives us compatibility with CLDR data defined in LDML. We can discuss potentially _relaxing_ `nmtoken`, but I'd like to avoid narrowing it.

aphillips:
> @eemeli:
> Hmm. Perhaps my wording was a bit awkward somehow? I agree with you about nmtoken having a clearly more restricted namespace, and I'm not trying to say that it's a literal.

Well... we're kind of saying it's a `literal` here, no? It's a kind of restricted literal _syntactically_, because it is unquoted. But it's not a "different thing" functionally (for which implementations have to apply different processing logic). 

That's why I suggested pushing the change down into the `literal` definition in the ANBF. If we choose to use `nmtoken` for unquoted literal, that would change my suggestion to:

```abnf
literal = '|' (literal-char / literal-escape)* '|'
        / nmtoken ; unquoted literals
```

>@eemeli:
> name would be sufficient for all but numbers and dates, but it'd be great not needing quotes for those either.

`nmtoken` covers these, since it can start with a digit 😄 

If we want numbers, there is the problem of negative numbers. `{-42 :number}` parses currently as markup or possibly as "reserved sigal" later.

mihnita:
> Number literals? Do we really expect things like {5 :number}, or are they rather academic?

It is a convenience, if you want.
This is used to handle hardcoded values, known when you write the code, but that are still locale dependent.
(for example Arabic uses "native digits" in some countries, and ASCII in others).
This is also something that can be configured, at least on Windows and Android, so it is a user preferences.
So it can't be just stored in the string and expect the translator to "translate it with the proper digits"

This also applies to decimal/thousand separators in bigger numbers, with dates / times, etc.

With MF1 (and most other system) the solution is to make that fixed value a parameter.
I've seen quite a few bugs, and had to explain it several times.
So it is not academic, it is a real thing.

We can argue if it is useful enough to complicate the syntax just for that.

On that my vote is yes.

I think it is also useful for markup, for example `...{img :html src=foo.jpg}...`

---

+1 to change this to be consistent with the values in options.
Meaning `literal / nmtoken / variable` (current ebnf: `option = name [s] "=" [s] (literal / nmtoken / variable)`

I do find the minus a bit troublesome though. Because I can do `{...key=-2}` but can't do `{-2 :number}`

mihnita:
> be at least as "wide" as the XML's Nmtoken gives

Although it sounds good at the first look (I even voted your comment with a thumb up :-),
that's a problem because the XML definition (https://www.w3.org/TR/xml/#d0e804) allows for `':'` as a starting character, but not a minus.

So `{:foo}` is a nmtoken, but overlaps with a function. This is also valid `{foo:baz:bar :fun}`.  And `{...key=-42}` is invalid (not a nmtoken, starts with minus).

stasm:
Great catch, @mihnita. Looks like we'll need to sort it out in the ABNF:

* by adding `:` only to `name-char` (and thus, to `nmtoken`), and accepting that `name` is not the same as XML's `Name` but at least `nmtoken` is aligned, or
* by adding `:` to `name-start`, aligning both `name` and `nmtoken` with their counterparts in XML. This however could result in weirdness around variable (`$:foo`) and function (`::foo`) names...

spec/message.abnf
```diff
@@ -34,15 +34,17 @@ literal-char = %x0-5B         ; omit \
 variable = "$" name
 function = (":" | "+" | "-") name
```
aphillips, line 35: `|` should be `/` no?

spec/syntax.md
```diff
@@ -357,19 +357,19 @@ Function calls do not require an operand as an argument,
...
+expression = "{" [s] (((literal / argtoken / variable) [s annotation]) / annotation) [s] "}"
 annotation = function *(s option)
 option = name [s] "=" [s] (literal / nmtoken / variable)
 ```
 
 Expression examples:
 
 ```
-{|1.23|}
+{1.23}
```
aphillips, line 368: we shouldn't remove the quoted examples, since they are still valid. We can augment the examples with unquoted ones. It would be useful to add or change the quoted ones to be something where the quotes add value, e.g. perhaps more like:

```
{1.23}
{|1.23|}
{1.23 :number maxFractionDigits=1}
{|-1.23| :number maxFractionDigits=2}
{|My Brand Name| :linkify href=|foobar.com|}
```

spec/syntax.md
```diff
@@ -452,6 +452,9 @@ Otherwise, the set of characters allowed in names is large.
 The _nmtoken_ token doesn't have _name_'s restriction on the first character
 and is used as variant keys and option values.
 
+The _argtoken_ is used for expression arguments,
+and matches _nmtoken_ except for not allowing `-` as a first character.
```
aphillips, line 456: As above, perhaps add text about why `argtoken` exists, e.g.:

> `argtoken` exists as a convenience, since it allows some literal values to be unquoted.

spec/syntax.md
```diff
@@ -452,6 +452,9 @@ Otherwise, the set of characters allowed in names is large.
...
+
 _Note:_ The Name and Nmtoken symbols are intentionally defined to be
 the same as XML's [Name](https://www.w3.org/TR/xml/#NT-Name) and [Nmtoken](https://www.w3.org/TR/xml/#NT-Nmtokens)
 in order to increase the interoperability with data defined in XML.
```
aphillips, line 460: code quote `Name` and `Nmtoken` and match the casing in the ABNF. I wouldn't call them "symbols". I would probably call them productions (since that's the ABNF terminology) or just "nothing", i.e. "_Note:_ `name` and `nmtoken` are intentionally defined..."

I would also be more direct here, changing "... in order to increase the interoperability with data defined in XML" to something more like:

> ... in order to ensure current and future compatibility with LDML (UAX35) and it's various names and values, since these are defined using XML.

aphillips (commented):
Editorial comments.

spec/message.abnf
eemeli: Oops, indeed. Filed #380 for that.

spec/syntax.md
eemeli: I added a couple such examples.

spec/syntax.md
eemeli: Done.

spec/syntax.md
eemeli: Should these changes be done in a separate non-normative editorial PR?

spec/message.abnf
```diff
@@ -49,13 +49,16 @@ reserved-char  = %x00-08        ; omit HTAB and LF
...
+argtoken = arg-start *name-char
 name-start = ALPHA / "_"
            / %xC0-D6 / %xD8-F6 / %xF8-2FF
            / %x370-37D / %x37F-1FFF / %x200C-200D
            / %x2070-218F / %x2C00-2FEF / %x3001-D7FF
            / %xF900-FDCF / %xFDF0-FFFD / %x10000-EFFFF
 name-char = name-start / DIGIT / "-" / "." / ":"
           / %xB7 / %x0300-036F / %x203F-2040
+arg-start = name-start / DIGIT / "."
+          / %xB7 / %x0300-036F / %x203F-2040
```
aphillips, line 61: technically `%x0300-036F` in both of these productions should be `%x300-36F`. The Unicadet in me prefers four digits (except when five or six are needed), but we should be consistent in our ABNF and with common ABNF practices.

spec/syntax.md
```diff
@@ -457,6 +465,10 @@ Otherwise, the set of characters allowed in names is large.
 The _nmtoken_ token doesn't have _name_'s restriction on the first character
 and is used as variant keys and option values.
 
+The _argtoken_ is used for expression arguments,
+and matches _nmtoken_ except for not allowing `-` or `:` as a first character.
+It allows some literal argument values to be unquoted.
```
aphillips, line 470: This is the part of this PR that keeps bringing me back for another look. I don't know what an "expression argument" is, especially given that other tokens found inside an expression can be considered arguments (notably, `options` can be considered arguments to a `function` or a `reserved`).

Since the purpose of `argtoken` is to allow unquoted literals, I think I might prefer if we dealt with them by recasting `literal` and naming `argtoken` as `unquoted`, e.g.:

```abnf
; literal data values; some values, depending on content, can appear
; without quoting | delimiters
literal  = unquoted / ("|" *(literal-char / literal-escape) "|")
unquoted = arg-start *name-char
```

Otherwise I'd want to rewrite aspects of this text to clarify "expression argument"

aphillips (changes requested):

spec/syntax.md
eemeli: I've now renamed the added tokens as `unquoted` and `unquoted-start`.

spec/message.abnf
eemeli: Done.

spec/message.abnf
```diff
@@ -55,7 +55,10 @@ name-start = ALPHA / "_"
            / %x2070-218F / %x2C00-2FEF / %x3001-D7FF
            / %xF900-FDCF / %xFDF0-FFFD / %x10000-EFFFF
 name-char = name-start / DIGIT / "-" / "." / ":"
-          / %xB7 / %x0300-036F / %x203F-2040
+          / %xB7 / %x300-36F / %x203F-2040
+unquoted = unquoted-start *name-char
```
aphillips, line 59: nit: insert a blank link before `unquoted` to separate the new production from the ones about `name` and `nmtoken` above (or, potentially, move the `unquoted` production with `name` and `nmtoken`???)

eemeli:
```suggestion

unquoted = unquoted-start *name-char
```

spec/syntax.md
```diff
@@ -457,6 +465,10 @@ Otherwise, the set of characters allowed in names is large.
 The _nmtoken_ token doesn't have _name_'s restriction on the first character
 and is used as variant keys and option values.
 
+The _unquoted_ token is used for expression arguments,
+and matches _nmtoken_ except for not allowing `-` or `:` as a first character.
+It allows some literal argument values to be unquoted.
```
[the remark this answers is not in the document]
eemeli:
```suggestion
The _unquoted_ token allows some literal values to appear unquoted in an _expression_.
It matches _nmtoken_ except for not allowing `-` or `:` as a first character.
```

spec/message.abnf
```diff
@@ -9,7 +9,7 @@ selectors = match 1*([s] expression)
 variant = when 1*(s key) [s] pattern
```
aphillips, line 9: This is one of two places (with the `option` definition on line 15) where a literal can appear but which cannot be unquoted. I'm wondering if this can introduce confusion for authors? Is this what we want? Here's a couple of examples just to brain-tease. I recognize that `unquoted` is easily confused with `nmtoken` when parsing and that these are probably processed differently, that is, `100` is not equal to `|100|` (the literal is a string, while the nmtoken might be a number). Weakly typed languages might not care, but strongly typed ones might be unhappy.

Is there a way to change our syntax so that _both_ quoted and unquoted literals are always distinct from other types of tokens (such that if I don't have to quote a literal in the `match` or `expression` start I don't have to quote it anywhere)?

```
match horse
when |horse| {requires a quote}
when * {{horse :something option=|horse|} option requires a quote}

match 100
when 100 {100 is nmtoken}
when |100| {100 is literal}
when * {{100 :something option=100} option is nmtoken}
when foo {{100 :something option=|100|} option is literal}
```

eemeli: Just to clarify, an `nmtoken` is valid as a `key`, and `unquoted` is a strict subset of `nmtoken`. Therefore, an `unquoted` token may be used as a `key`.

In my mental model, the MF2 spec level understanding of the type of the value of a `literal`, `nmtoken` or `unquoted` should always be a string. Considering the examples above, a formatting function implementing `:something` should only be passed bags of options like `{ option: 'horse' }` or `{ option: '100' }` irrespective of their presentation in the message syntax as `horse`, `|horse|`, `100`, or `|100|`. Any further parsing of a value as e.g. a number would be entirely the responsibility of the formatting function.

aphillips: The thing that makes me wary is that a parser can "capture" the same value in two different ways--as an `nmtoken` or as an (unquoted) `literal`--suggesting that implementations could produce differing outcomes for the same input (which isn't what we're saying). We should take extra care to describe the relationship between literals and tokens. We don't *say* that `unquoted` is a subset of `nmtoken`--there is neither an ABNF nor specification relationship. 

One challenge is that some functions (such as `plural`) expect _specific_ `nmtoken` values as keys and then separately support a (huge) set of literals, e.g.:

```
match :plural($foo)
when few {an nmtoken recognized by plural}
when 100 {is it an nmtoken or a literal?}
when |-3| {definitely a literal}
when * {etc.}
```

This might be okay, since MF2 is untyped and leaves the parsing up to the functions. But it makes me think we might go back and take up my previous suggestion of merging `unquoted` into the `literal` production. It would be clearer (to me) if (for example) the production for `key` (`nmtoken / literal / "*"`) incorporated `unquoted` via `literal`

eemeli: With the current structure of the ABNF, `unquoted` cannot be a subset of `literal` because that would make the parsing of `key` and `option` ambiguous. For example, with `option=100` we would not be able to say whether the `100` is an `unquoted` via `literal`, or if it's an `nmtoken`.

If we were to make such a change, we could need to limit `nmtoken` so that it _must_ start with either `-` or `:`. This doesn't sound like a good idea to me.

stasm: > The thing that makes me wary is that a parser can "capture" the same value in two different ways--as an `nmtoken` or as an (unquoted) `literal`--suggesting that implementations could produce differing outcomes for the same input (which isn't what we're saying). We should take extra care to describe the relationship between literals and tokens. We don't _say_ that `unquoted` is a subset of `nmtoken`--there is neither an ABNF nor specification relationship.

We shouldn't think of the ABNF as the definition of the *data model*. Instead, it describes a set of validation rules which can be used to filter and accept the input. As @eemeli says, both `literal` and `unquoted` can end up being encoded in the data model as the same thing. For example, in the implementation that I experimented with, all three: variant keys, expression operands (arguments), and option values are encoded as instances of the [`Literal`](https://github.com/stasm/message2/blob/4abf43f2023b6e20d8ee1d462684d0741ece791b/syntax/ast.ts#L116) class in the data model.

stasm: > With the current structure of the ABNF, `unquoted` cannot be a subset of `literal` because that would make the parsing of `key` and `option` ambiguous. For example, with `option=100` a parser would not be able to say whether the `100` is an `unquoted` via `literal`, or if it's an `nmtoken`.

Do we need to be able to tell them apart outside of the context in which they appear? We can probably know which is which by inferring it from their position relative to other tokens in the input.

spec/message.abnf
```diff
@@ -55,7 +55,11 @@ name-start = ALPHA / "_"
            / %x2070-218F / %x2C00-2FEF / %x3001-D7FF
            / %xF900-FDCF / %xFDF0-FFFD / %x10000-EFFFF
 name-char = name-start / DIGIT / "-" / "." / ":"
-          / %xB7 / %x0300-036F / %x203F-2040
+          / %xB7 / %x300-36F / %x203F-2040
+
+unquoted = unquoted-start *name-char
```
stasm, line 60: I'd like to bikeshed the name a bit. `unquoted` suggested that it's an *unquoted literal*, but in fact, it's much more limited than that. It sits somewhere between `name` and `nmtoken`.

spec/message.abnf
eemeli: I don't think so? That's effectively reflected in #382 as well. But the ABNF can't be ambiguous.

spec/message.abnf
eemeli: Suggestions are welcome. I started with `argtoken`, but renamed to `unquoted` on @aphillips's request.

spec/message.abnf
aphillips: @stasm The ABNF is not the definition of the _data model_, but it should, to the extent it is reasonable to do so, be _consistent with_ the data model. I think the thing that is perilous is to have overlap between `literal`, `unquoted`, and `nmtoken` and then describe these separately--with separate requirements--in the data model or elsewhere in the specification. How do I know if `option=100` should apply `unquoted` or `nmtoken` rules to `100`? Is `option=|100|` a different value? Of course, we can avoid that by bringing items together using a production (some productions in the ABNF exist for the convenience of the specification).

Perhaps change this:

```abnf
option = name [s] "=" [s] (literal / nmtoken / variable / unquoted)
```

to something like this:

```abnf
option = name [s] "=" [s] value
value = literal / nmtoken / variable / unquoted
```

(I suggest this for `option` and not for `key` because `key` already does this)

spec/message.abnf
```diff
@@ -9,7 +9,7 @@ selectors = match 1*([s] expression)
 variant = when 1*(s key) [s] pattern
 key = nmtoken / literal / "*"
 
-expression = "{" [s] (((literal / variable) [s annotation]) / annotation) [s] "}"
+expression = "{" [s] (((literal / unquoted / variable) [s annotation]) / annotation) [s] "}"
 annotation = (function *(s option)) / reserved
 
 option = name [s] "=" [s] (literal / nmtoken / variable)
```
eemeli, line 15: @aphillips The ABNF change you suggest in https://github.com/unicode-org/message-format-wg/pull/364#discussion_r1199057894 would make it ambiguous whether the `value` of `foo=42` would get parsed as an `nmtoken` or an `unquoted`, so I presume that this is what you meant:

```suggestion
option = name [s] "=" [s] value
value = literal / nmtoken / variable
```


Would applying this change address your remaining concerns about this PR?

aphillips: @eemeli Not... exactly? My request to add `unquoted` to `option` was to make clear to readers that the value side of the option can contain an unquoted literal (which is what `unquoted` is). I agree that `nmtoken` is a pure superset of `unquoted`. Maybe the way out of this would be to add a comment to the ABNF (and specification) about the relationship between `unquoted` and `nmtoken`?

macchiati:
> it's actually a non-goal to try to align the ABNF with the data model

I think that's a mistake. It very valuable to use consistent terminology in
the spec and in the BNF. That way when people are reading the spec and see
a term X, they can find what X means syntactically in the BNF. It might
mean that the BNF is not minimal, but *that* is a non-goal. For example,
we've found it quite useful in the CLDR spec for terms like simple_unit,
single_unit, mixed_unit, etc.

eemeli:
Updated as discussed on today's call & rebased on latest `main`. As proposed here, `nmtoken` is dropped and `literal` can now be either `quoted` or `unquoted`.

This has the effect of making `-` and `:` not allowed as first characters of variant keys or named options, as they were previously. My understanding of today's discussion was that a subsequent change might re-allow for those in a separate PR.

I've closed most of the line discussions above as they were either resolved or outdated by this change.

spec/formatting.md
```diff
@@ -363,7 +363,7 @@ These are divided into the following categories:
     uses a `:plural` selector function which requires its input to be numeric:
 
     ```
-    match {|horse| :plural}
+    match {horse :plural}
     when 1 {The value is one.}
```
stasm, line 367: This is the part that makes me a bit skeptical. I see `{horse` in one line and `{The` in the next; they look the same but are completely different syntax. Prefixes like `$`, `:`, and `|` make it easier for me to visually recognize placeholders.

Do you think we could keep the quoted literals in the examples in this file, as a stylistic choice?

spec/syntax.md
```diff
@@ -438,55 +445,65 @@ text-char = %x0-5B         ; omit \
...
+### Literals
+
+_Literal_ is used for matching variants and providing input to _expressions_.
+_Quoted_ literals may include content with any Unicode code point,
+except for surrogate code points U+D800 through U+DFFF.
+The characters `\` and `|` MUST be escaped as `\\` and `\|`.
+_Unquoted_ literals have a much more restricted range that
+is intentionally close to the XML's [Nmtoken](https://www.w3.org/TR/xml/#NT-Nmtoken),
+with the restriction that it MUST NOT start with `-` or `:`,
+as those would conflict with _function_ start characters.
```
stasm, line 457: These two lines don't feel _normative_ the way the rest of this passage does. Perhaps remove them?

spec/syntax.md
```diff
@@ -438,55 +445,65 @@ text-char = %x0-5B         ; omit \
...
 ```
 
 ### Names
 
 The _name_ token is used for variable names (prefixed with `$`),
 function names (prefixed with `:`, `+` or `-`),
 as well as option names.
-A name MUST NOT start with an ASCII digit and certain basic combining characters.
+It is based on XML's [Name](https://www.w3.org/TR/xml/#NT-Name),
+with the restriction that it MUST NOT start with an ASCII digit and certain basic combining characters.
```
stasm, line 481: The phrasing here suggested that XML's `Name` allows digits as the first character.

stasm (approved):

spec/syntax.md
eemeli: Good point. The only actual divergence is the colon.

```suggestion
with the restriction that it MUST NOT start with `:`,
as that would conflict with _function_ start characters.
```

spec/formatting.md
```diff
@@ -445,10 +445,10 @@ and ends with U+007D RIGHT CURLY BRACKET `}`.
 Between the brackets, the following contents are used:
 
 - Expression with Literal Operand: U+007C VERTICAL LINE `|`
-  followed by the value of the Literal,
+  followed by the Quoted or Unquoted value,
```
aphillips, line 448: This change is wrong, since you're defining a quoted literal here. I would suggest de-naming Literal here:

```suggestion
  followed by the value of the literal,
```

spec/message.abnf
```diff
@@ -26,14 +29,16 @@ text-char = %x0-5B         ; omit \
...
-             / %xE000-10FFFF
+quoted      = "|" *(quoted-char / quoted-escape) "|"
+quoted-char = %x0-5B         ; omit \
+            / %x5D-7B        ; omit |
+            / %x7D-D7FF      ; omit surrogates
+            / %xE000-10FFFF
 
-variable = "$" name
-function = (":" / "+" / "-") name
+unquoted = unquoted-start *name-char ; based on https://www.w3.org/TR/xml/#NT-Nmtoken,
```
aphillips, line 38: move the comment above the rule for legibility

spec/message.abnf
```diff
@@ -47,18 +52,18 @@ reserved-char  = %x00-08        ; omit HTAB and LF
                / %x7E-D7FF      ; omit surrogates
                / %xE000-10FFFF
 
-name    = name-start *name-char ; based on https://www.w3.org/TR/xml/#NT-Name, but cannot start with U+003A COLON ":"
-nmtoken = 1*name-char           ; equal to https://www.w3.org/TR/xml/#NT-Nmtoken
+name = name-start *name-char ; based on https://www.w3.org/TR/xml/#NT-Name,
```
aphillips, line 55: same comment: move comment above line

aphillips (changes requested):
Partial review... catching up

spec/syntax.md
aphillips: This isn't quite how I would approach this. I think it needs more explanation about literals in general and then about the unquoted ones. The relationship to Nmtoken (which we broke on purpose) isn't that relevant any more. Perhaps:

```suggestion
_Literal_ values are used to pass data to various parts of a `message`:
* As the value of a `key` in a `when` statement
* As the `argument` in an `expression`
* As the `value` in an `option`

A `Literal` is a sequence of _Unicode code points_ and can include any Unicode character. Surrogate code points are not allowed. 

The characters `\\` U+005C REVERSE SOLIDUS and `|` U+007C VERTICAL BAR **_must_** be escaped (as `\\` and `\|` respectively) when they appear in the value of a `Literal`.

Spaces are significant in a `Literal`.

A `Quoted` literal is surrounded by `|` characters. 

A `Literal` can be `Unquoted` when its content matches that production. The content restrictions for `Unquoted` follow best practices for the use of Unicode in formal grammars and are intentionally similar to, for example, XML's [Nmtoken](https://www.w3.org/TR/xml/#NT-Nmtoken).
```

macchiati: > +A `Literal` is a sequence of _Unicode code points_ and can include any Unicode character. Surrogate code points are not allowed. 

Should be:

A `Literal` is a sequence of _Unicode code points_, and can contain any Unicode code points except for surrogate code points and non-character code points.

Reason: "Unicode character" would mean "assigned Unicode character", which is unnecessarily fragile across versions.

aphillips: Actually, non-characters (U+FFFF for example) are not excluded. Only surrogate code points are. This is consistent with e.g. DOMString and USVString.

macchiati: That's fine

PR #365, opened 2023-03-12
Allow colons in nmtokens

stasm:
The current definition of the `nmtoken` production doesn't match the XML's one. The same is true of `name`, but names are less of an issue. `nmtokens` are frequently used in LDML to define attribute values which are likely to be used a variant keys and option values.

This was originally spotted by @mihnita in https://github.com/unicode-org/message-format-wg/pull/364#issuecomment-1465029138. I [suggested](https://github.com/unicode-org/message-format-wg/pull/364#issuecomment-1465033819) fixing it in one of the two following manners:

> * by adding `:` only to `name-char` (and thus, to `nmtoken`), and accepting that `name` is not the same as XML's `Name` but at least `nmtoken` is aligned, or
> * by adding `:` to `name-start`, aligning both `name` and `nmtoken` with their counterparts in XML. This however could result in weirdness around variable (`$:foo`) and function (`::foo`) names...

This PR implements the first solution.

### DIFF
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -42,15 +42,15 @@ function = ":" name
 markup-start = "+" name
 markup-end = "-" name
 
-name    = name-start *name-char ; matches XML https://www.w3.org/TR/xml/#NT-Name
-nmtoken = 1*name-char           ; matches XML https://www.w3.org/TR/xml/#NT-Nmtokens
+name    = name-start *name-char ; based on https://www.w3.org/TR/xml/#NT-Name, but cannot start with U+003A COLON ":"
+nmtoken = 1*name-char           ; equal to https://www.w3.org/TR/xml/#NT-Nmtoken
 name-start = ALPHA / "_"
            / %xC0-D6 / %xD8-F6 / %xF8-2FF
            / %x370-37D / %x37F-1FFF / %x200C-200D
            / %x2070-218F / %x2C00-2FEF / %x3001-D7FF
            / %xF900-FDCF / %xFDF0-FFFD / %x10000-EFFFF
-name-char = name-start / DIGIT / "-" / "." / %xB7
-          / %x0300-036F / %x203F-2040
+name-char = name-start / DIGIT / "-" / "." / ":"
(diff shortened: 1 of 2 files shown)


stasm:
It's worth noting that even without this change, it's still possible to use literals for values which don't match our `nmtoken` production. Also, AFAICT, LDML doesn't actually define any exotic attribute values right now. However, aligning with XML exactly means that any valid tokenized XML attribute value can be represented without resorting to the literal syntax.

eemeli (approved):

aphillips (approved):
This looks good. We can revisit later if needed.

spec/message.abnf
```diff
@@ -42,15 +42,15 @@ function = ":" name
 markup-start = "+" name
 markup-end = "-" name
 
-name    = name-start *name-char ; matches XML https://www.w3.org/TR/xml/#NT-Name
-nmtoken = 1*name-char           ; matches XML https://www.w3.org/TR/xml/#NT-Nmtokens
+name    = name-start *name-char ; based on https://www.w3.org/TR/xml/#NT-Name, but cannot start with U+003A COLON ":"
+nmtoken = 1*name-char           ; equal to https://www.w3.org/TR/xml/#NT-Nmtoken
```
gibson042, line 46: Is the colon in `{:functionName}` syntax sufficiently valuable to justify deviation from **exact** match of the XML productions? If so, I question whether alignment with XML is actually well-motivated, and if not then perhaps it would make more sense to use characters excluded from Nmtoken—``` !"#$%&'()*+,/;<=>?@[\]^`{|}~ ```.

Note also that if leading `:` is prohibited because of [`function`](https://github.com/unicode-org/message-format-wg/blob/main/spec/syntax.md#names), then leading `-` should be prohibited for the same reason.
```suggestion
name    = name-start *name-char ; based on https://www.w3.org/TR/xml/#NT-Name, but cannot
                                ; start with U+003A COLON ":" or U+002D HYPHEN-MINUS "-"
nmtoken = 1*name-char           ; equal to https://www.w3.org/TR/xml/#NT-Nmtoken
```

PR #368, opened 2023-03-13
Draft of the registry specification

stasm:
This PR includes 2 files:

* `registry.dtd` is the schema for defining regestries in XML; it is normative.
* `registry.md` is the non-normative documentation explaining the motivation and the schema. It also includes examples.

This PR is based on my [old spec proposal from January 2022](https://github.com/unicode-org/message-format-wg/discussions/218), and the more recent [presentation](https://github.com/unicode-org/message-format-wg/blob/main/meetings/2023/notes-2023-02-06.md) that I did to resume the work on the design of the registry.

For now, I've focused on describing custom functions by defining their signatures. A single signature corresponds to one set of: the current locale, the argument, and the options bag.

I didn't address all feedback from our February 6 meeting in this PR. Looking at my notes, here are the topics for future discussions:

* [ ] Not all options should be locale-specific.
* [ ] Some options should be common to all signatures of a given function.
* Support other data types besides functions:
  * [ ] markup,
  * [ ] metadata (comments, max length, screenshot URL, etc.),
  * [ ] global variables.
* [ ] Describe the interface of runtime arguments and local variables (i.e. the return types of formatting functions). Right now the validation of arguments and option values only applies to literal values.

### DIFF
--- a/spec/registry.dtd
+++ b/spec/registry.dtd
@@ -1,0 +1,33 @@
+<!ELEMENT registry (function*|pattern*)>
+
+<!ELEMENT function (description|signature+)>
+<!ATTLIST function name NMTOKEN #REQUIRED>
+
+<!ELEMENT description (#PCDATA)>
+
+<!ELEMENT pattern EMPTY>
+<!ATTLIST pattern id ID #REQUIRED>
+<!ATTLIST pattern regex CDATA #REQUIRED>
+
+<!ELEMENT signature (input*|param*|match*)>
+<!ATTLIST signature type (match|format) #REQUIRED>
+<!ATTLIST signature position (open|close|standalone) "standalone">
+<!ATTLIST signature locales NMTOKENS #IMPLIED>
+
+<!ELEMENT input EMPTY>
+<!ATTLIST input title CDATA #IMPLIED>
+<!ATTLIST input values NMTOKENS #IMPLIED>
+<!ATTLIST input pattern NMTOKEN #IMPLIED>
+
+<!ELEMENT param EMPTY>
+<!ATTLIST param title CDATA #IMPLIED>
+<!ATTLIST param name NMTOKEN #REQUIRED>
+<!ATTLIST param values NMTOKENS #IMPLIED>
+<!ATTLIST param default NMTOKEN #IMPLIED>
+<!ATTLIST param pattern IDREF #IMPLIED>
+<!ATTLIST param required (true|false) "false">
+<!ATTLIST param readonly (true|false) "false">
+
+<!ELEMENT match EMPTY>
+<!ATTLIST match values NMTOKENS #IMPLIED>
+<!ATTLIST match pattern NMTOKEN #IMPLIED>
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -1,0 +1,162 @@
+# WIP DRAFT MessageFormat 2.0 Registry
+
+_This document is non-normative._
+
+The implementations and tooling can greatly benefit from a structured definition of formatting and matching functions available to messages at runtime.
+The _registry_ is a mechanism for storing such declarations in a portable manner.
+
+## Goals
+
+The goal of the registry is to provide a machine-readable description of MessageFormat extensions (custom functions) in order to:
+
+* support the localization roundtrip,
+* validate message correctness, and
+* improve the authoring experience.
+
+## Use Cases
+
+The registry enables the following usage scenarios:
+
+* Generate variant keys for a given locale during XLIFF extraction.
+* Verify the exhaustiveness of variant keys given a selector.
+* Type-check values passed into functions.
+* Validate that matching functions are only called in selectors.
+* Validate that formatting functions are only called in placeholders.
+* Forbid edits to certain function options (e.g. currency options).
+* Autocomplete function and option names.
+* Display on-hover tooltips for function signatures with documentation.
+* Display/edit known message metadata.
+* Restrict input in GUI by providing a dropdown with all viable option values.
+
+## Data Model
+
+The registry contains descriptions of function signatures.
+[`registry.dtd`](./registry.dtd) describes its data model.
+
+The main building block of the registry is the `<function>` element.
+It represents an implementation of a custom function available to translation at runtime.
+A function defines a human-readable _description_ of its behavior
+and one or more machine-readable _signatures_ of how to call it.
+Named `<pattern>` elements can optionally define regex validation rules for input, optional values, and variant keys.
+
+The `<signature>` element defines one particular signature of the custom function,
+i.e. the set of arguments and named options that can be used together in a single call to the function.
+The `type` attribute specifies the calling context of the signature.
+A `type="format"` function can only be called inside a placeholder inside translatable text.
+A `type="match"` function can only be called inside a selector.
+Signatures with a non-empty `locales` attribute are locale-specific and only available in translations in the given languages.
+
+A signature may define the type of input it accepts with zero or more `<input>` elments.
+In MessageFormat 2.0 functions can only ever accept at most one positional argument.
+Multiple `<input>` elements can be used to define different validation rules for this single argument input,
+together with appropriate human-readable descriptions.
+
+A signature may also define one or more `<param>` elements representing _named options_ to the function.
+Parameters are optional by default,
+unless the `required` attribute is present.
+They accept either a finite enumeration of values (the `values` attribute)
+or validate they input with a regular expression (the `pattern` attribute).
+Read-only parameters (the `readonly` attribute) can be displayed to translators in CAT tools, but may not be edited.
+
+Matching-function signatures additionally include one or more `<match>` elements
+to define the keys against which they can match when used as selectors.
+
+## Example
+
+The following `registry.xml` is an example of a registry file
+which may be provided by an implementation to describe its built-in functions.
+For the sake of brevity, only `locales="en"` is considered.
+
+```xml
+<?xml version="1.0" encoding="UTF-8" ?>
+<!DOCTYPE registry SYSTEM "./registry.dtd">
+
+<registry>
+    <function name="platform">
+        <description>Match the current OS.</description>
+        <signature type="match">
+            <match values="windows linux macos android ios"/>
+        </signature>
+    </function>
+
+    <pattern id="anyNumber" regex="-?[0-9]+(\.[0-9]+)"/>
+    <pattern id="positiveInteger" regex="[0-9]+"/>
+    <pattern id="currencyCode" regex="[A-Z]{3}"/>
+
+    <function name="number">
+        <description>
+            Format a number.
+            Match a numerical value against CLDR plural categories or against a number literal.
+        </description>
+
+        <signature type="match" locales="en">
+            <input pattern="anyNumber"/>
+            <param name="type" values="cardinal ordinal"/>
+            <param name="minimumIntegerDigits" pattern="positiveInteger"/>
+            <param name="minimumFractionDigits" pattern="positiveInteger"/>
+            <param name="maximumFractionDigits" pattern="positiveInteger"/>
+            <param name="minimumSignificantDigits" pattern="positiveInteger"/>
+            <param name="maximumSignificantDigits" pattern="positiveInteger"/>
+            <match values="one other"/>
+            <match pattern="anyNumber"/>
+        </signature>
+
+        <signature type="format" locales="en">
+            <input pattern="anyNumber"/>
+            <param name="minimumIntegerDigits" pattern="positiveInteger"/>
+            <param name="minimumFractionDigits" pattern="positiveInteger"/>
+            <param name="maximumFractionDigits" pattern="positiveInteger"/>
+            <param name="minimumSignificantDigits" pattern="positiveInteger"/>
+            <param name="maximumSignificantDigits" pattern="positiveInteger"/>
+            <param name="style" readonly="true" values="decimal currency percent unit" default="decimal"/>
+            <param name="currency" readonly="true" pattern="currencyCode"/>
+        </signature>
+    </function>
+</registry>
+```
+
+A localization engineer can then extend the registry by defining the following `customRegistry.xml` file.
+
+```xml
+<?xml version="1.0" encoding="UTF-8" ?>
+<!DOCTYPE registry SYSTEM "./registry.dtd">
+
+<registry>
+    <function name="noun">
+        <description>Handle the grammar of a noun.</description>
+        <signature type="format" locales="en">
+            <input title="Noun id"/>
+            <param name="article" values="definite indefinite"/>
+            <param name="plural" values="one other"/>
+            <param name="case" values="nominative genitive" default="nominative"/>
+        </signature>
+    </function>
+
+    <function name="adjective">
+        <description>Handle the grammar of an adjective.</description>
+        <signature type="format" locales="en">
+            <input title="Adjective id"/>
+            <param name="article" values="definite indefinite"/>
+            <param name="plural" values="one other"/>
+            <param name="case" values="nominative genitive" default="nominative"/>
+        </signature>
+        <signature type="format" locales="en">
+            <input title="Adjective id"/>
+            <param name="article" values="definite indefinite"/>
+            <param name="accord"/>
+        </signature>
+    </function>
+</registry>
+```
+
+Messages can now use the `:noun` and the `:adjective` functions.
+The following message references the first signature of `:adjective`,
+which expects the `plural` and `case` options:
+
+    {You see {$color adjective article=indefinite plural=one case=nominative} {$object :noun case=nominative}!}
+
+The following message references the second signature of `:adjective`,
+which only expects the `accord` option:
+
+    let $obj = {$object :noun case=nominative}
+    {You see {$color adjective article=indefinite accord=$obj} {$obj}!}


spec/registry.dtd
```diff
@@ -0,0 +1,33 @@
...
+<!ATTLIST pattern regex CDATA #REQUIRED>
+
+<!ELEMENT signature (input*|param*|match*)>
+<!ATTLIST signature type (match|format) #REQUIRED>
+<!ATTLIST signature position (open|close|standalone) "standalone">
+<!ATTLIST signature locales NMTOKENS #IMPLIED>
+
+<!ELEMENT input EMPTY>
+<!ATTLIST input title CDATA #IMPLIED>
+<!ATTLIST input values NMTOKENS #IMPLIED>
```
eemeli, line 19: If we end up allowing for some `nmtoken`-ish values to be used unquoted as arguments (#364), it would be good to match that rule here.

spec/registry.dtd
```diff
@@ -0,0 +1,33 @@
...
+<!ATTLIST input pattern NMTOKEN #IMPLIED>
+
+<!ELEMENT param EMPTY>
+<!ATTLIST param title CDATA #IMPLIED>
+<!ATTLIST param name NMTOKEN #REQUIRED>
+<!ATTLIST param values NMTOKENS #IMPLIED>
+<!ATTLIST param default NMTOKEN #IMPLIED>
+<!ATTLIST param pattern IDREF #IMPLIED>
+<!ATTLIST param required (true|false) "false">
+<!ATTLIST param readonly (true|false) "false">
```
eemeli, line 29:
```suggestion
<!ATTLIST param translate (true|false) "true">
```

Given that developers and translators will both be working with MF2, it might be clearer to use `translate` rather than `readonly` to communicate the intent of this attribute.

As a related question, shouldn't this also be available for `<input>`?

spec/registry.dtd
```diff
@@ -0,0 +1,33 @@
...
+<!ATTLIST function name NMTOKEN #REQUIRED>
+
+<!ELEMENT description (#PCDATA)>
+
+<!ELEMENT pattern EMPTY>
+<!ATTLIST pattern id ID #REQUIRED>
+<!ATTLIST pattern regex CDATA #REQUIRED>
+
+<!ELEMENT signature (input*|param*|match*)>
+<!ATTLIST signature type (match|format) #REQUIRED>
```
eemeli, line 13: Is there a particular reason why it'd be better to have `<signature type="match">` and `<signature type="format">` rather than e.g. `<match-signature>` and `<format-signature>`?

Separating the two would allow for attributes like `position` and child elements like `<match>` to only be valid for one element, and not the other.

spec/registry.dtd
```diff
@@ -0,0 +1,33 @@
...
+<!ATTLIST signature position (open|close|standalone) "standalone">
```
eemeli, line 14: Is this meant to be a building block for markup support? This attribute does not appear to be documented or explained.

spec/registry.md
```diff
@@ -0,0 +1,162 @@
...
+
+The `<signature>` element defines one particular signature of the custom function,
+i.e. the set of arguments and named options that can be used together in a single call to the function.
+The `type` attribute specifies the calling context of the signature.
+A `type="format"` function can only be called inside a placeholder inside translatable text.
+A `type="match"` function can only be called inside a selector.
+Signatures with a non-empty `locales` attribute are locale-specific and only available in translations in the given languages.
+
+A signature may define the type of input it accepts with zero or more `<input>` elments.
+In MessageFormat 2.0 functions can only ever accept at most one positional argument.
```
eemeli, line 50:
```suggestion
Functions can only ever accept at most one positional argument.
```
The context here is rather clear, and including a major.minor version number sets up an unnecessary value that may need to be updated later.

spec/registry.md
```diff
@@ -0,0 +1,162 @@
...
+Multiple `<input>` elements can be used to define different validation rules for this single argument input,
+together with appropriate human-readable descriptions.
+
+A signature may also define one or more `<param>` elements representing _named options_ to the function.
```
eemeli, line 54: As these are _named options_, why is the element called `<param>` and not e.g. `<option>`?

spec/registry.md
```diff
@@ -0,0 +1,162 @@
...
+        </signature>
+    </function>
+</registry>
+```
+
+Messages can now use the `:noun` and the `:adjective` functions.
+The following message references the first signature of `:adjective`,
+which expects the `plural` and `case` options:
+
+    {You see {$color adjective article=indefinite plural=one case=nominative} {$object :noun case=nominative}!}
```
eemeli, line 156:
```suggestion
    {You see {$color :adjective article=indefinite plural=one case=nominative} {$object :noun case=nominative}!}
```

spec/registry.md
```diff
@@ -0,0 +1,162 @@
...
+
+The following message references the second signature of `:adjective`,
+which only expects the `accord` option:
+
+    let $obj = {$object :noun case=nominative}
+    {You see {$color adjective article=indefinite accord=$obj} {$obj}!}
```
eemeli, line 162:
```suggestion
    {You see {$color :adjective article=indefinite accord=$obj} {$obj}!}
```

eemeli (commented):
A first pass with some inline comments & questions.

One key factor that seems unaddressed is the representation of variable inputs. In other words, if my message includes something like
```
{$foo :datetime dateStyle=long}
```
How can I represent an expectation that the input `$foo` ought to be an actual date object of whatever description, rather than e.g. a string representation of a date?

Edit: D'oh, that's your last checkbox above.

spec/registry.dtd
```diff
@@ -0,0 +1,33 @@
+<!ELEMENT registry (function*|pattern*)>
+
+<!ELEMENT function (description|signature+)>
+<!ATTLIST function name NMTOKEN #REQUIRED>
```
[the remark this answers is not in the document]
stasm: The `ID` type is also subject to a validity constraint about there only being one element of the given name in the XML document: https://www.w3.org/TR/xml/#id.

I wasn't sure if we wanted to be this strict. Perhaps it's OK to have functions named the same as regex patterns? Or to have two function definitions with the same name? That's why I went for `name` and `NMTOKEN`.

spec/registry.dtd
stasm: In general, I'm not fond of `true` being the default value. I think I'd prefer to name these attributes such that their lack is equal to `false`.

stasm: Re. `readonly`/`translate` for `<input>` -- good point. I think I'll drop the ability to define more than one `<input>` element per signature, then.

spec/registry.md
eemeli: Do we actually refer to the "argument" as such anywhere in the current spec? I've no strong opinions on that, though `option` is clearly preferable over `param`.

spec/registry.dtd
stasm: I don't think #364 is related, actually. My intent here was to allow specifying either an enumeration of nmtokens or a regex to validate arguments that are MF2 literals. This seems orthogonal to whether these literals are quoted or not?

spec/registry.dtd
stasm: > Re. `readonly`/`translate` for `<input>` -- good point. I think I'll drop the ability to define more than one `<input>` element per signature, then.

Done in 831a9cdaca37aa26d6299ae1680d3ea134f33e99.

spec/registry.dtd
eemeli: I meant that if #364 lands, then _nearly_ all `NMTOKEN` values may be used without quoting, except for the ones that start with `:` and `-`. Given that, it might be appropriate to leave those out of the values that are definable by `values` rather than `pattern`.

stasm: I don't think we can enforce such restriction in the DTD alone. LDML does it by extending DTD with _annotations_: https://unicode.org/reports/tr35/#57-dtd-annotations.

spec/registry.dtd
stasm: That's a good point. Done in 39f13a99db4f0391ebabd8534dbf54e190c6c827, thanks!

spec/registry.md
```diff
@@ -0,0 +1,156 @@
...
+        </formatSignature>
+    </function>
+
+    <function name="adjective">
+        <description>Handle the grammar of an adjective.</description>
+        <formatSignature locales="en">
+            <input title="Adjective id"/>
+            <option name="article" values="definite indefinite"/>
+            <option name="plural" values="one other"/>
+            <option name="case" values="nominative genitive" default="nominative"/>
```
catamorphism, line 135: I'm a little confused about why there's a `case` option for an adjective if this signature is only defined in locale `"en"`.

spec/registry.md
```diff
@@ -0,0 +1,156 @@
...
+        </formatSignature>
+        <formatSignature locales="en">
+            <input title="Adjective id"/>
+            <option name="article" values="definite indefinite"/>
+            <option name="accord"/>
```
catamorphism, line 140: The `accord` option doesn't have a pattern or an enumeration, which surprises me given lines 51-52 (which suggest to me that every option has one or the other).

spec/registry.md
```diff
@@ -0,0 +1,156 @@
...
+
+A _signature_ defines one particular set of arguments and named options that can be used together in a single call to the function.
+`<formatSignature>` corresponds to a function call inside a placeholder inside translatable text.
+`<matchSignature>` coresponds to a function call inside a selector.
+Signatures with a non-empty `locales` attribute are locale-specific and only available in translations in the given languages.
+
+A signature may define the positional argument of the function with the `<input>` element.
+A signature may also define one or more `<option>` elements representing _named options_ to the function.
+Options are optional by default,
+unless the `required` attribute is present.
```
catamorphism, line 50: Would it be an error for an option to have both a `required` and a `default` attribute?

spec/registry.md
```diff
@@ -0,0 +1,156 @@
...
+        </formatSignature>
+    </function>
+</registry>
+```
+
+Messages can now use the `:noun` and the `:adjective` functions.
+The following message references the first signature of `:adjective`,
+which expects the `plural` and `case` options:
+
+    {You see {$color :adjective article=indefinite plural=one case=nominative} {$object :noun case=nominative}!}
```
catamorphism, line 150: What happens if you write `{You see {$color :adjective article=indefinite}}`? Since none of the options in the example are `required`, it's unclear which of the two signatures for `:adjective` should be used. Or does it not matter because the idea is that `:adjective` has a single implementation for all signatures?

spec/registry.md
```diff
@@ -0,0 +1,156 @@
...
+
+A signature may define the positional argument of the function with the `<input>` element.
+A signature may also define one or more `<option>` elements representing _named options_ to the function.
+Options are optional by default,
+unless the `required` attribute is present.
+They accept either a finite enumeration of values (the `values` attribute)
+or validate they input with a regular expression (the `pattern` attribute).
+Read-only options (the `readonly` attribute) can be displayed to translators in CAT tools, but may not be edited.
+
+Matching-function signatures additionally include one or more `<match>` elements
```
catamorphism, line 55: Is it "one or more" or "zero or more"? The DTD implies that there can be zero `<match>` elements, as does the second signature for `number` in the example below.

catamorphism (commented):
I left some inline comments. Overall, the major thing that's unclear to me is the split between formatting and matching functions. More examples (both of valid and invalid messages) might help. I'm also not sure what a matching function *returns*, in terms of the code that implements the interface.

spec/registry.md
```diff
@@ -0,0 +1,156 @@
...
+## Use Cases
+
+The registry enables the following usage scenarios:
+
+* Generate variant keys for a given locale during XLIFF extraction.
+* Verify the exhaustiveness of variant keys given a selector.
+* Type-check values passed into functions.
+* Validate that matching functions are only called in selectors.
+* Validate that formatting functions are only called in placeholders.
+* Forbid edits to certain function options (e.g. currency options).
```
aphillips, line 25: Not sure that "currency" is a good example here

spec/registry.md
```diff
@@ -0,0 +1,156 @@
...
+    <pattern id="positiveInteger" regex="[0-9]+"/>
+    <pattern id="currencyCode" regex="[A-Z]{3}"/>
+
+    <function name="number">
+        <description>
+            Format a number.
+            Match a numerical value against CLDR plural categories or against a number literal.
+        </description>
+
+        <matchSignature locales="en">
```
aphillips, line 86: I'm very curious about the `locales` attribute--curious in this case is a euphemism for "nervous about". I would prefer if functions were mostly locale neutral (I can declare, for example, a number format in any message in any locale).

spec/registry.dtd
```diff
@@ -0,0 +1,36 @@
+<!ELEMENT registry (function*|pattern*)>
+
+<!ELEMENT function (description|signature+)>
```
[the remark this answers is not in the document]
stasm: Good catch, thanks.

spec/registry.md
```diff
@@ -0,0 +1,156 @@
+# WIP DRAFT MessageFormat 2.0 Registry
+
+_This document is non-normative._
+
+The implementations and tooling can greatly benefit from a structured definition of formatting and matching functions available to messages at runtime.
```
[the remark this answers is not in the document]
stasm: I've added a paragraph explaining formatting vs. matching functions to this PR. We should discuss the naming further. For example, I'm not sure if it's appropriate to call `{a +html}` a *formatting* function. (In the current state, the PR doesn't reflect the recent change to function open/close prefixes.)

spec/registry.md
```diff
@@ -0,0 +1,156 @@
...
+* improve the authoring experience.
+
+## Use Cases
+
+The registry enables the following usage scenarios:
+
+* Generate variant keys for a given locale during XLIFF extraction.
+* Verify the exhaustiveness of variant keys given a selector.
+* Type-check values passed into functions.
+* Validate that matching functions are only called in selectors.
```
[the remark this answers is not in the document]
stasm: You're right, let's be explicit about this. I'd prefer to leave this out of this PR, however, because I think that we will first need to decide whether it's allowed to match on a function-less expression (i.e. `match {$foo}` instead of `match {$foo :func}`).

spec/registry.md
```diff
@@ -0,0 +1,156 @@
...
+## Data Model
+
+The registry contains descriptions of function signatures.
+[`registry.dtd`](./registry.dtd) describes its data model.
+
+The main building block of the registry is the `<function>` element.
+It represents an implementation of a custom function available to translation at runtime.
+A function defines a human-readable _description_ of its behavior
+and one or more machine-readable _signatures_ of how to call it.
+Named `<pattern>` elements can optionally define regex validation rules for input, option values, and variant keys.
```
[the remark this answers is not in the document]
stasm: I don't know yet how to validate runtime types. I'm suggesting to use regexes to validate literal values. Let's discuss what sort of type validation is even possible for  some implementations.

spec/registry.md
```diff
@@ -0,0 +1,156 @@
...
+
+A _signature_ defines one particular set of arguments and named options that can be used together in a single call to the function.
+`<formatSignature>` corresponds to a function call inside a placeholder inside translatable text.
+`<matchSignature>` coresponds to a function call inside a selector.
+Signatures with a non-empty `locales` attribute are locale-specific and only available in translations in the given languages.
+
+A signature may define the positional argument of the function with the `<input>` element.
+A signature may also define one or more `<option>` elements representing _named options_ to the function.
+Options are optional by default,
```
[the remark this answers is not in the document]
stasm: Yes, thank you! I struggled with expressing this well.

spec/registry.md
```diff
@@ -0,0 +1,156 @@
...
+unless the `required` attribute is present.
+They accept either a finite enumeration of values (the `values` attribute)
```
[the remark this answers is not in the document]
stasm: Hmm, interesting. I'm not opposed to removing `values` and keeping `pattern` as the only option. The reasons why I opted for both were:
* `values` allows the values to be specified inline (`<match values="one other"/>`) whereas `pattern` refers to a `<pattern>` element by id. I felt it was more convenient in particular for plural categories, cases and genders to be able to defined them inline rather than define a regex for each possible combination.
* Some supplemental data in CLDR uses `values="<enumeration>"`, too, e.g. https://github.com/unicode-org/cldr/blob/release-43/common/supplemental/grammaticalFeatures.xml, so it's easy to reuse this data as-is.

Do you feel strongly about this?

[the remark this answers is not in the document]
eemeli: Generating autocompletion values for an editor is easy from an explicit list in `values`, but _really hard_ from a regexp `pattern`. I'd much rather keep `values` even if only for that.

spec/registry.md
stasm: English doesn't make it easy to build meaningful examples of grammatical features :) I'll try to come up with something better.

spec/registry.md
stasm: Right, this is still WIP and underspec'ed. I'd like to continue the discussion about validating runtime values outside of this PR.

stasm:
Thank you, @eemeli, @catamorphism, and @aphillips for your reviews. I was able to address some feedback, but not all of it yet. I'll continue over the weekend. I'm already seeing a few topics that I'd like to leave out of this PR and discuss separately:

* How to represent human-readable meta-data, such as the now-removed `title` attribute.
* Add open/close concepts to function signatures. Are they still _formatting_ functions, or do we need a new category for them?
* What should happen when a formatting function is used for matching, and vice versa? Should implementations support it, or raise errors?
  * Possibly related: should `match {$foo}` be allowed, or do we want to always require a function name in selectors?
* How can we validate runtime types? E.g. `{$count :number}` -- what can we realistically do to know anything about `$count`?
  * Related: are regexes good enough for validating literals?

I'd like to suggest to try to merge this PR without answering the above questions, and continue iterating on it later.

spec/registry.md
stasm: Oh, I think I like it!

aphillips:
@stasm 

Thanks for your (continuing) work on this. This will, of course, be a topic for Monday's call. I will also have recommendations on the other PRs by then.

Do you have opinions about the questions you raise? Would it be productive to try to bring in solutions? Or do you want to raise separate issues to discuss these? I'd like to iterate quickly on these where possible........

spec/registry.md
```diff
@@ -0,0 +1,156 @@
...
+
+The implementations and tooling can greatly benefit from a structured definition of formatting and matching functions available to messages at runtime.
+The _registry_ is a mechanism for storing such declarations in a portable manner.
+
+## Goals
+
+The goal of the registry is to provide a machine-readable description of MessageFormat extensions (custom functions) in order to:
+
+* support the localization roundtrip,
+* validate message correctness, and
```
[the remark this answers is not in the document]
stasm: I merged "Use Cases" into "Goals", which should hopefully make this clearer.

spec/registry.md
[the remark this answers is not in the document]
catamorphism: @stasm No, I don't feel strongly, and @eemeli has a good point, so I withdraw the question :)

spec/registry.md
catamorphism: Yeah, I know :) I was just thinking the example could be from a different language, maybe.

eemeli (approved):
We still need to iterate on the registry, but I think we should merge this so that we can do so.

spec/registry.dtd
```diff
@@ -0,0 +1,34 @@
...
+<!ATTLIST pattern id ID #REQUIRED>
+<!ATTLIST pattern regex CDATA #REQUIRED>
+
+<!ELEMENT signature (input?|param*|match*)>
+<!ATTLIST signature type (match|format) #REQUIRED>
+<!ATTLIST signature position (open|close|standalone) "standalone">
+<!ATTLIST signature locales NMTOKENS #IMPLIED>
+
+<!ELEMENT input EMPTY>
+<!ATTLIST input title CDATA #IMPLIED>
```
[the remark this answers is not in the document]
stasm: Filed #405 to discuss this further.

spec/registry.md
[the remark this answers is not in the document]
stasm: Filed #409 to discuss this further.

spec/registry.md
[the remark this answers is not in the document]
stasm: Filed #407 about regexes and #408 about runtime types.

spec/registry.md
stasm: Filed #410 to discuss this further.

PR #371, opened 2023-03-29
Drop separate syntax constructs for markup

eemeli:
This is a minimally opinionated extension of our `{$foo :bar opt=42}` placeholder syntax to support markup elements.

The idea here is to drop "markup" as an explicit concept in the syntax and to allow for three different function prefix sigils:
- `:` for standalone content
- `+` for starting or opening elements
- `-` for ending or closing elements

The exact meaning of these standalone/open/close concepts is left undefined, allowing for each of the following styles to be equivalently expressible:

```
{Click {|a| +html}here{|a| -html} to continue}
```

```
{Click {+html.a}here{-html.a} to continue}
```

```
{Click {+ph}<a>{-ph}here{+ph}</a>{-ph} to continue}
```

### DIFF
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -1,6 +1,6 @@
 message = [s] *(declaration [s]) body [s]
 
-declaration = let s variable [s] "=" [s] "{" [s] expression [s] "}"
+declaration = let s variable [s] "=" [s] placeholder
 body = pattern
      / (selectors 1*([s] variant))
 
@@ -11,9 +11,6 @@ variant = when 1*(s key) [s] pattern
 key = nmtoken / literal / "*"
 
 placeholder = "{" [s] expression [s] "}"
-            / "{" [s] markup-start *(s option) [s] "}"
-            / "{" [s] markup-end [s] "}"
-
 expression = ((literal / variable) [s annotation])
            / annotation
 annotation = function *(s option)
@@ -38,9 +35,7 @@ literal-char = %x0-5B         ; omit \
              / %xE000-10FFFF
 
 variable = "$" name
-function = ":" name
-markup-start = "+" name
-markup-end = "-" name
+function = (":" | "+" | "-") name
 
 name    = name-start *name-char ; matches XML https://www.w3.org/TR/xml/#NT-Name
 nmtoken = 1*name-char           ; matches XML https://www.w3.org/TR/xml/#NT-Nmtokens
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -1,17 +1,15 @@
 message = [s] *(declaration [s]) body [s]
 
-declaration = let s variable [s] "=" [s] placeholder
+declaration = let s variable [s] "=" [s] expression
 body = pattern
      / (selectors 1*([s] variant))
 
-pattern = "{" *(text / placeholder) "}"
-selectors = match 1*([s] placeholder)
+pattern = "{" *(text / expression) "}"
+selectors = match 1*([s] expression)
 variant = when 1*(s key) [s] pattern
 key = nmtoken / literal / "*"
 
-placeholder = "{" [s] expression [s] "}"
-expression = ((literal / variable) [s annotation])
-           / annotation
+expression = "{" [s] (((literal / variable) [s annotation]) / annotation) [s] "}"
 annotation = function *(s option)
 option = name [s] "=" [s] (literal / nmtoken / variable)
 
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -5,8 +5,7 @@ body = pattern
      / (selectors 1*([s] variant))
 
 pattern = "{" *(text / placeholder) "}"
-selectors = match 1*([s] selector)
-selector = "{" [s] expression [s] "}"
+selectors = match 1*([s] placeholder)
 variant = when 1*(s key) [s] pattern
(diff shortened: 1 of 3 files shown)


eemeli:
This change simplifies parsing a bit, as `declaration`, `pattern` and `selectors` may now all refer to a single `placeholder`, rather than each needing to build their own wrappers around `expression`.

For clarity, I'm also suggesting here to change our language to say that a `match` statement may contain multiple `selector`s, rather than being itself a "selector" with multiple "expressions", and that the variants that follow are not a part of the `match` statement.

spec/message.abnf
```diff
@@ -1,19 +1,15 @@
...
      / (selectors 1*([s] variant))
 
 pattern = "{" *(text / placeholder) "}"
-selectors = match 1*([s] selector)
-selector = "{" [s] expression [s] "}"
+selectors = match 1*([s] placeholder)
 variant = when 1*(s key) [s] pattern
 key = nmtoken / literal / "*"
 
 placeholder = "{" [s] expression [s] "}"
```
stasm, line 12: Do we need `placeholder` at all now?

stasm (commented):
I’m looking forward to the discussion on Friday about whether open/close are needed in the syntax. Assuming they are, I think this approach is better than our current syntax. That said, I’m still not fond of plus and minus as prefixes. Also, perhaps it would be better to keep the starting `:` in all cases? As in, `:+strong` for the opening tag.

spec/message.abnf
eemeli: If we didn't have `placeholder`, the `"{" [s] ... [s] "}"` would need to be repeated in three different places. On the other hand, `expression` could be dropped by inlining its contents into `placeholder`, which is the only place where it's now used.

eemeli:
> [...] Also, perhaps it would be better to keep the starting `:` in all cases? As in, `:+strong` for the opening tag.

But why? At least to me `:+strong` and `:-strong` look much uglier than `+strong` and `-strong`, for no real benefit.

spec/message.abnf
stasm: I meant dropping `placeholder` and moving the braces to `expression`. I prefer “expression” because I wouldn’t call the RHS in a `let` declaration a “placeholder”.

eemeli: Would you prefer including that change in this PR, or as a separate follow-up?

eemeli: Done.

spec/message.abnf
```diff
@@ -1,21 +1,15 @@
...
 variant = when 1*(s key) [s] pattern
 key = nmtoken / literal / "*"
 
-placeholder = "{" [s] expression [s] "}"
-            / "{" [s] markup-start *(s option) [s] "}"
-            / "{" [s] markup-end [s] "}"
-
-expression = ((literal / variable) [s annotation])
-           / annotation
+expression = "{" [s] (((literal / variable) [s annotation]) / annotation) [s] "}"
```
aphillips, line 12: For my action item I am working on ABNF, but we might as well talk about it here. I was going to suggest:

```abnf
expression = "{" [s] (((literal / variable) [s annotation])
             / annotation) [s] "}"
annotation = (function / reserved) *(s option)
function = ":" name
reserved = (%21-%26 / %2a / %2b / %2d / %2f
           / %3b-%40 / %5e / %7e) name ; reserved sigil characters, specifically these: !"#$%&*+-/^~
```

I'm reluctant to embrace `+` and `-` being "function introducers". I would rather just have a single starter for the "thing we've been calling a function" and having another name for other types of function. If these are different constructs, they should parse to another production name. That can be accommodated by pushing it down a level:

```abnf
annotation = named *(s option)
named = function / reserved
function = ":" name
reserved = [!@#$%^&*+-<>?] name
```
 This would allow something to become unreserved by adding a named production:

```abnf
annotation = named *(s option)
named = function / reserved
function = ":" name
fragment = "#" name
reserved = [!@$%^&*+-<>?] name ; deleted # from here
```

eemeli: I suspect that ultimately anything that may have an argument and/or a bag of options could be represented by some sort of "function". The proposed shape of `reserved` encodes this as the only possible direction for the extension of the spec, so ultimately anything that would be later un-reserved could effectively be described as a "function introducer".

If we don't want to limit ourselves this way, then really the `reserved` construct ought to be defined on a higher level, allowing for more freedom (using this PR's syntax as a baseline):
```abnf
expression = "{" [s] (((literal / variable) [s annotation]) / annotation / reserved) [s] "}"
reserved = reserved-start reserved-body
reserved-start = "!" / "@" / "#" / "$" / "%" / "^" / "&" / "*" / "<" / ">" / "?"
reserved-body = *(reserved-char / reserved-escape / ("{" reserved-body "}") / literal)
reserved-char = %x0-5B         ; omit \
              / %x5D-7A        ; omit { | }
              / %xE000-10FFFF
reserved-escape = backslash ( backslash / "|" / "{" / "}" )
```

Now, independently of how we might proceed with `reserved`, I do think that "open" and "close" placeholders are sufficiently common that they definitely should have their own top-level introducers `+` and `-`. This leaves us with at least 11 more to use for other concepts, should that prove desirable later on.

mihnita (approved):
Thank you!

eemeli:
Merging, as agreed during today's call.

PR #372, opened 2023-03-29
Add column-first pattern selection

### LINKED ISSUE #93: Concrete Message Format Use Case to work for next meeting

### LINKED ISSUE #104: How do we handle and represent selection for a multi-select message?

### LINKED ISSUE #105: How do we support default case selection values in a multi-select message?

### LINKED ISSUE #106: How do we support multiple selection args (selectors) in a multi-select message?

### LINKED ISSUE #208: Case Selection - Design Document
This issue encapsulates the work done for Case Selection and may include more documentation.

Design Document : https://docs.google.com/document/d/1nT-_-ZMknbtKvUc9lR9v9esOGiXK-OOPwJv7Ccl-2EI/edit#heading=h.6x3fflpb0yjs

cc @mihnita  please expand description or any other information needed

Related issues:
Closes #93
Closes #99
Closes #104
Closes #105
Closes #106

Open PR's: 
#170

eemeli:
This is an attempt to explicitly document how case/variant selection happens with messages that have a `when` selector statement. The method presented here implements the "column-first" selection method that was agreed upon at the last meeting.

A JS implementation of the algorithm is available; see #369 for more information on that.

The overall intent is to minimally but sufficiently define selection, such that two implementations that use similar custom selector functions will make the same selection, when given the same input message and formatting context. In a number of places details are left for each implementation to fill in for themselves, as each may have a different internal representations of resolved and unresolved values and may perform value matching in different ways.

By necessity, the method definition needs to use more formal language than what we have so far in the spec. For that, I've borrowed some of the conventions of the TC39 spec and hope that it's sufficiently readable as is, without us needing separate definitions of e.g. what a "list" is.

- Closes #93
- Closes #104
- Closes #105
- Closes #106
- Closes #208
- See also #101
- See also #170
- See also #333
- See also #369

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -43,7 +43,7 @@ Then, the remaining _variants_ are sorted lexicographically by key preference,
 with earlier _selectors_ having higher priority than later ones.
 Finally, the highest-sorted _variant_ is selected.
 
-This selecion method is defined in more detail below.
+This selection method is defined in more detail below.
 An implementation MAY use any pattern selection method,
 as long as its observable behaviour matches the results of the method defined here.
 
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -19,6 +19,199 @@ the local variable takes precedence.
 It is an error for a local variable definition to
 refer to a local variable that's defined after it in the message.
 
+## Pattern Selection
+
+When formatting a message with one or more _selectors_,
+the _pattern_ of one of the _variants_ must be selected for formatting.
+
+When a message has a single _selector_,
+an implementation-defined method compares each key to the _selector_
+and determines which of the keys match, and in what order of preference.
+A catch-all key will always match, but is always the least preferred choice.
+During selection, the _variant_ with the best-matching key is selected.
+
+In a message with more than one _selector_,
+each _variant_ also has a corresponding number of keys.
+These correspond to _selectors_ by position.
+The same implementation-defined method as above is used to compare
+the corresponding key of each _variant_ to its _selector_,
+to determine which of the keys match, and in what order of preference.
+In order to select a single _variant_,
+the full list of _variants_ will need to be filtered and sorted.
+First, each _variant_ with a key that does not match its _selector_ is left out.
+Then, the remaining _variants_ are sorted lexicographically by key preference,
+with earlier _selectors_ having higher priority than later ones.
+Finally, the highest-sorted _variant_ is selected.
+
+This selecion method is defined in more detail below.
+An implementation MAY use any pattern selection method,
+as long as its observable behaviour matches the results of the method defined here.
+
+### Resolve Selectors
+
+First, resolve the values of each _selector_:
+
+1. Let `res` be a new empty list of resolved values that support selection.
+1. For each _expression_ `exp` of the message's _selectors_,
+   1. Let `rv` be the resolved value of `exp`.
+   1. If selection is supported for `rv`:
+      1. Append `rv` as the last element of the list `res`.
+   1. Else:
+      1. Let `nomatch` be a resolved value for which selection always fails.
+      1. Append `nomatch` as the last element of the list `res`.
+      1. Emit a Selection Error.
+
+The shape of the resolved values is determined by each implementation,
+along with the manner of determining their support for selection.
+
+### Resolve Preferences
+
+Next, using `res`, resolve the preferential order for all message keys:
+
+1. Let `pref` be a new empty list of lists of strings.
+1. For each index `i` in `res`:
+   1. Let `keys` be a new empty list of strings.
+   1. For each _variant_ `var` of the message:
+      1. Let `key` be the `var` key at position `i`.
+      1. If `key` is not the catch-all key `'*'`:
+         1. Let `ks` be the decoded value of `key`.
+         1. Append `ks` as the last element of the list `keys`.
+   1. Let `rv` be the resolved value at index `i` of `res`.
+   1. Let `matches` be the result of calling the method MatchSelectorKeys(`rv`, `keys`)
+   1. Append `matches` as the last element of the list `pref`.
+
+The method MatchSelectorKeys is determined by the implementation.
+It takes as arguments a resolved _selector_ value `rv` and a list of string keys `keys`,
+and returns a list of string keys in preferential order.
+The returned list must only contain unique elements of the input list `keys`,
+and it may be empty.
+The most preferred key is first, with the rest sorted by decreasing preference.
+
+### Filter Variants
+
+Then, using the preferential key orders `pref`,
+filter the list of _variants_ to the ones that match with some preference:
+
+1. Let `vars` be a new empty list of _variants_.
+1. For each _variant_ `var` of the message:
+   1. For each index `i` in `pref`:
+      1. Let `key` be the `var` key at position `i`.
+      1. If `key` is the catch-all key `'*'`:
+         1. Continue the inner loop on `pref`.
+      1. Let `ks` be the decoded value of `key`.
+      1. Let `matches` be the list of strings at index `i` of `pref`.
+      1. If `matches` includes `ks`:
+         1. Continue the inner loop on `pref`.
+      1. Else:
+         1. Continue the outer loop on message _variants_.
+   1. Append `var` as the last element of the list `vars`.
+
+### Sort Variants
+
+Finally, sort the list of variants `vars` and select the _pattern_:
+
+1. Let `sortable` be a new empty list of (integer, _variant_) tuples.
+1. For each _variant_ `var` of `vars`:
+   1. Let `tuple` be a new tuple (-1, `var`).
+   1. Append `tuple` as the last element of the list `sortable`.
+1. Let `len` be the integer count of items in `pref`.
+1. Let `i` be `len` - 1.
+1. While `i` >= 0:
+   1. Let `matches` be the list of strings at index `i` of `pref`.
+   1. Let `minpref` be the integer count of items in `matches`.
+   1. For each tuple `tuple` of `sortable`:
+      1. Let `matchpref` be an integer with the value `minpref`.
+      1. Let `key` be the `tuple` _variant_ key at position `i`.
+      1. If `key` is not the catch-all key `'*'`:
+         1. Let `ks` be the decoded value of `key`.
+         1. Let `matchpref` be the integer position of `ks` in `matches`.
+      1. Set the `tuple` integer value as `matchpref`.
+   1. Call the method SortVariants(`sortable`).
+   1. Set `i` to be `i` - 1.
+1. Let `var` be the _variant_ element of the first element of `sortable`.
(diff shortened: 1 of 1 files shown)


spec/formatting.md
```diff
@@ -19,6 +19,199 @@ the local variable takes precedence.
...
+with earlier _selectors_ having higher priority than later ones.
+Finally, the highest-sorted _variant_ is selected.
+
+This selecion method is defined in more detail below.
+An implementation MAY use any pattern selection method,
+as long as its observable behaviour matches the results of the method defined here.
+
+### Resolve Selectors
+
+First, resolve the values of each _selector_:
```
aphillips, line 52: I don't understand this? What are the "values" of a _selector_?

I think a _selector_ has two inputs: an _expression_ (either a variable or a literal) and a list of keys. Resolving the "value" of the _expression_ seems to be early-binding in a way that isn't required to implement the spec. Or is "value" something else?

spec/formatting.md
```diff
@@ -19,6 +19,199 @@ the local variable takes precedence.
...
+      1. Let `nomatch` be a resolved value for which selection always fails.
+      1. Append `nomatch` as the last element of the list `res`.
+      1. Emit a Selection Error.
+
+The shape of the resolved values is determined by each implementation,
+along with the manner of determining their support for selection.
+
+### Resolve Preferences
+
+Next, using `res`, resolve the preferential order for all message keys:
```
aphillips, line 69: what does "preferential order" mean here? Can you clarify what you mean?

I think you're asking the selector to resolve all possible values in their preferred order. This is fine for a selector with an enumerated set of values (like plural rules), but limiting for implementations that might be somewhat dynamic (like plural explicit values, where the infinite set of numbers might be valid key values)

spec/formatting.md
```diff
@@ -19,6 +19,199 @@ the local variable takes precedence.
...
+   1. Let `rv` be the resolved value at index `i` of `res`.
+   1. Let `matches` be the result of calling the method MatchSelectorKeys(`rv`, `keys`)
+   1. Append `matches` as the last element of the list `pref`.
+
+The method MatchSelectorKeys is determined by the implementation.
+It takes as arguments a resolved _selector_ value `rv` and a list of string keys `keys`,
+and returns a list of string keys in preferential order.
+The returned list must only contain unique elements of the input list `keys`,
+and it may be empty.
+The most preferred key is first, with the rest sorted by decreasing preference.
```
aphillips, line 88: I think this might be phrased oddly?

As I noted above, I'd rather think of the _selector_ as having the burden of matching, not "the implementation"

spec/formatting.md
```diff
@@ -19,6 +19,199 @@ the local variable takes precedence.
...
+
+The method SortVariants is determined by the implementation.
+It takes as an argument a `sortable` list of (integer, _variant_) tuples,
+which it modifies in place using some stable sorting algorithm.
+The method does not return anything.
+The list is sorted according to the tuple's first integer element,
+such that a lower number is sorted before a higher one,
+and entries that have the same number retain their order.
+
+### Examples
```
aphillips, line 142: Can you write an example with plural? This example is nice for its simplicity, but the 👿 is in the details (as they say). Showing `when 1` and `when one` and `when *` together would make this clearer to me. Perhaps that can be an *additional* example?

aphillips (changes requested):
I'm struggling to understand this algorithm, partially for wording reasons and partially because I think that you and I think about matching differently. This seems to do what I think it should do, but may (as noted in my comments) restrict how selectors can be implemented or might have some quirks with unbounded value spaces. Have a look and see what you think.

spec/formatting.md
```diff
@@ -19,6 +19,199 @@ the local variable takes precedence.
...
+When formatting a message with one or more _selectors_,
+the _pattern_ of one of the _variants_ must be selected for formatting.
+
+When a message has a single _selector_,
+an implementation-defined method compares each key to the _selector_
+and determines which of the keys match, and in what order of preference.
+A catch-all key will always match, but is always the least preferred choice.
+During selection, the _variant_ with the best-matching key is selected.
+
+In a message with more than one _selector_,
```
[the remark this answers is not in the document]
eemeli: The intent with these paragraphs is to provide an easier-to-understand but accurate representation of the selection method than what's detailed in the subsequent sections. The split into two cases (N=1, then N>1) is intended to first explain the more common and more straightforward situation, before extending to the more complex.

> I think it is clearer to think of a _selector_ as a function. Then one can describe single-select and the later mulitple-select more generically.

To me, this doesn't quite match with our current [definition](https://github.com/unicode-org/message-format-wg/blob/main/spec/syntax.md#selectors) of the word:
> A **_selector_** is a statement containing one or more expressions which will be used to choose one of the _variants_ during formatting.

If accepted, this is modified by PR #371 to read:
> A `match` statement contains one or more **_selectors_** which will be used to choose one of the _variants_ during formatting.

I read this as describing a _selector_ primarily as a data structure rather than a function, which means that until we actually define such a function or other construct, we can't use it for further explanations.

> I think the single-vs-multiple is a red herring as well: they should work identically.

They do work identically; it's just that with only one we don't need to combine multiple preference orders, which is the main source of our complexity here.

spec/formatting.md
eemeli: Happy to play with the semantics here, but what I intend the "resolved value" here to mean is probably pretty close to your understanding of a _selector_ as a function. For example, let's say that we start from the following syntax representation of a selector:
```
{$count :plural minimumFractionDigits=2}
```
This may be parsed e.g. into the following data model representation, which we can still claim to represent the _selector_:
```js
{
  type: 'expression',
  name: 'plural',
  operand: { type: 'variable', name: 'count' },
  options: [
    { name: 'minimumFractionDigits', value: { type: 'nmtoken', value: '2' } }
  ]
}
```
For the sake of simplicity, let's say that `$count` may resolve to an externally provided integer value `42`. With that, the resolution could in practice be done by calling the handler for `:plural` that's defined in our registry:
```js
let rv = pluralHandler(42, { minimumFractionDigits: '2'})
```
The important bit here is that `rv` doesn't need to be a concrete formatted value like `'42.00'`, but it could be something like:
```ts
function matchPlurals(
  pr: Intl.PluralRules,
  value: number,
  keys: Set<string>
): string[] {
  const matches: string[] = [];
  const str = String(value);
  if (keys.has(str)) matches.push(str);
  const cat = pr.select(value);
  if (keys.has(cat)) matches.push(cat);

  return matches;
}

// The `bind()` returns a curried function with a single argument `keys`
let rv = matchPlurals.bind(null,
  new Intl.PluralRules('en', { minimumFractionDigits: 2 }),
  42
);
```

So while this does do a little bit of early binding for resolving `$count` and `:plural`, as well as constructing a PluralRules instance, it still leaves the actual selection to be performed later. This is ultimately necessary to ensure that we emit the right errors at the right time. 

For example, if the value given as the `minimumFractionDigits` option were instead `-1`, the `new Intl.PluralRules()` call would have failed with a range error. Our selection method should ensure that an error for this is emitted exactly once, and at an appropriate time compared to errors in any other selector.

It's important to note that this doesn't lead to the requirement that an implementation must be as eager in its resolution, as long as its observable behaviour (including error emission) matches the method presented here.

spec/formatting.md
eemeli:
```suggestion
Next, using `res`, resolve the preferential order for all variant keys:
```

The "all" here is meant to refer to all the keys in the current message, rather than all possible keys. Would "all variant keys" communicate that better? If so, please feel free to commit this change.

spec/formatting.md
eemeli: Added a third example using plurals.

PR #374, opened 2023-04-10
Propose reserving additional sigils for future use

aphillips:
This proposal proposes to reserve sigils for future use with reserved sigils allowed in the same place that function names are currently, e.g. standalone (`{%foo}`), with options (`{%foo key=val}`), or subsidiary to variable/literals (`{|quote me| %foo}` or `{$var %foo}`).

A case could be made for reserving a comment syntax that allows a placeholder with prose in it (and I would suggest using the `%` for it). @eemeli's proposal is effectively this. I didn't propose this because of the impact on parsers in the future. Let's discuss.

### DIFF
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -11,12 +11,10 @@ variant = when 1*(s key) [s] pattern
 key = nmtoken / literal / "*"
 
 placeholder = "{" [s] expression [s] "}"
-            / "{" [s] markup-start *(s option) [s] "}"
-            / "{" [s] markup-end [s] "}"
 
 expression = ((literal / variable) [s annotation])
            / annotation
-annotation = function *(s option)
+annotation = (function / markup / reserved) *(s option)
 option = name [s] "=" [s] (literal / nmtoken / variable)
 
 ; reserved keywords are always lowercase
@@ -39,8 +37,8 @@ literal-char = %x0-5B         ; omit \
 
 variable = "$" name
 function = ":" name
-markup-start = "+" name
-markup-end = "-" name
+markup = ("+" / "-") name
+reserved = ("!" / "@" / "#" / "$" / "%" / "^" / "&" / "*" / "<" / ">" / "?") name
 
 name    = name-start *name-char ; matches XML https://www.w3.org/TR/xml/#NT-Name
 nmtoken = 1*name-char           ; matches XML https://www.w3.org/TR/xml/#NT-Nmtokens


eemeli (commented):
@aphillips Did I understand right from the call that you'd be updating this PR?

aphillips:
@eemeli That's correct.

aphillips:
Note: this PR is still missing the work on the spec prose.

spec/message.abnf
```diff
@@ -32,6 +33,11 @@ literal-char = %x0-5B         ; omit \
              / %xE000-10FFFF
 
 variable = "$" name
+
+; reserve additional sigils for future use
+reserved = reserved-start text
+reserved-start = ("!" / "@" / "#" / "%" / "^" / "&" / "*" / "<" / ">" / "?" / "~")
```
eemeli, line 39: This is really cute, but it would disallow literal values that include `\{}`.

Currently this is a valid expression:
```
{:foo key=|{bar}|}
```
But if we were to unreserve `@` and try to give it the same semantics as with `:`, this would be a parse error as the expression would end before the literal does:
```
{@foo key=|{bar}|}
```

Leaving out the nesting from my proposal in https://github.com/unicode-org/message-format-wg/pull/371#discussion_r1155874966, I think we still need new `reserved-char` and `reserved-escape` definitions:

```suggestion
reserved = reserved-start reserved-body
reserved-start = "!" / "@" / "#" / "%" / "^" / "&" / "*" / "<" / ">" / "?" / "~"
reserved-body = *(reserved-char / reserved-escape / literal)
reserved-char = %x0-5B         ; omit \
              / %x5D-7A        ; omit { | }
              / %x7E-D7FF      ; omit surrogates
              / %xE000-10FFFF
reserved-escape = backslash ( backslash / "{" / "|" / "}" )
```

eemeli (changes requested):

spec/syntax.md
```diff
@@ -345,19 +345,18 @@ Whitespace within a _pattern_ is meaningful and MUST be preserved.
...
-other than the operand in front of them.
+A _literal_ or _variable_ ***may*** be optionally followed by an _annotation_. 
 
-Function calls do not require an operand as an argument,
-but an _expression_ must not be completely empty.
+An _annotation_ consists of a _function_ and its named _options_.
+
+_Functions_ do not accept any positional arguments other than the _literal_ or _variable_ in front of them.
+
+_Reserved_ sequences start with a reserved character and are intended for future standardization.
```
eemeli, line 356: Should we include some explicit language about whether implementations are allowed to define their own meanings for parts of the reserved space?

eemeli (approved):
Looks good! I think we'll want to add a section on `reserved` to the syntax text, but I like the general shape of this.

spec/syntax.md
aphillips: We should not allow that unless we specifically reserve one or more sigils for private use. It was on my mind writing this to propose private use, but I deferred those thoughts in favor of getting this in.

What we don't want is for vendor foo to use (let's say) `~` for private use and then MF2.1 to adopt `~` for something else. Private agreement is another thing and leaving room for that is probably a Good Idea.

spec/message.abnf
```diff
@@ -9,8 +9,9 @@ selectors = match 1*([s] expression)
 variant = when 1*(s key) [s] pattern
 key = nmtoken / literal / "*"
 
-expression = "{" [s] (((literal / variable) [s annotation]) / annotation) [s] "}"
+expression = "{" [s] (((literal / variable) [s annotation]) / annotation / reserved) [s] "}"
 annotation = function *(s option)
```
eemeli, line 13: Just had a thought: Should we allow for reserved operators to accept a positional argument? That could be achieved with this change:

```suggestion
expression = "{" [s] (((literal / variable) [s annotation]) / annotation) [s] "}"
annotation = (function *(s option)) / reserved
```

aphillips: We probably should do this. One callout: `reserved` is _space significant_ and has no ending marker (other than the closing `}` of the `expression` itself). An implication of this for parsers is that one can't identify a placeholder substring and `trim()` the contents before parsing. This needs to be called out in the spec.

spec/syntax.md
```diff
@@ -108,7 +109,7 @@ let hello = new MessageFormat('{Hello, world!}')
 hello.format()
 ```
 
-### Expressions
+### Placeholder Expressions
```
eemeli, line 112: As discussed at our last meeting, an expression is also used for variable definitions and for selectors; not just placeholders. I think changing that terminology should be beyond the scope of this PR.

eemeli (changes requested):
A few requested changes documented inline. Furthermore, the error handling & fallback string representations in `formatting.md` should be updated as a part of this change.

spec/syntax.md
eemeli: I've filed #378 to discuss this separately.

spec/syntax.md
```diff
@@ -495,6 +497,8 @@ Inside _patterns_,
 whitespace is part of the translatable content and is recorded and stored verbatim.
 Whitespace is not significant outside translatable text, except where required by the syntax.
 
+***NOTE:*** Whitespace **_is_** significant in the `reserved` production and implementations need to be careful not to trim trailing whitespace from reserved sequences.
```
eemeli, line 500: As an alternative solution, we could also tie our future hands a little, and explicitly make the trailing whitespace of `reserved` not significant. Realistically, I don't think that would cause any problems, especially as we're already considering trailing whitespace non-significant in other expressions.

macchiati:
BTW, big fan of [Semantic Line Breaks](https://sembr.org/)

spec/syntax.md
aphillips: I made the links work in the spec doc by expanding the term here. Note that the term "placeholder" doesn't appear in the ABNF and is not currently defined anywhere (else). The syntax of an expression includes the `{`/`}` markers and so I concluded that it was reasonable to say that all expressions are placeholders. The text recognizes that placeholder expressions can appear in variable definitions and selectors.

The easier way to fix this would be to use singular instead of plural and I'll do that for now.

spec/syntax.md
```diff
@@ -345,20 +346,17 @@ Whitespace within a _pattern_ is meaningful and MUST be preserved.
 
 ### Expressions
 
-**_Expressions_** can either start with an operand or a function call.
+_Expressions_ ***must*** start with a _literal_, a _variable_, or an _annotation_, or consist of a _reserved_ string. An _expression_ ***must not*** be empty.
```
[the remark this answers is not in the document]
aphillips: I will touch up the formatting.

I missed correcting this when making the change. I mean, _technically_, it's correct, but the reserved sequence is just part of an annotation :-).

spec/syntax.md
```diff
@@ -398,6 +396,10 @@ Message examples:
 {{+h1 name=above-and-beyond}Above And Beyond{-h1}}
 ```
 
+#### Reserved
+
+_Reserved_ sequences start with a reserved character and are intended for future standardization. A reserved sequence can contain arbitrary text in which whitespace _is_ significant. The reserved sequence terminates with the end of the expression in which it appears. While a reserved sequence is technically "well-formed", unrecognized reserved sequences have no meaning and implementations **_may_** emit an error when encountered during formatting.
```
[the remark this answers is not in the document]
aphillips: Part of this paragraph should be replicated into formatting.md, but I wanted to have a warning-flavored bit of text here.

spec/syntax.md
aphillips: We can do that, but we would need to change the ABNF. The problem is that all of our other productions are bounded in some way and the exterior whitespace is applied by the optional production `[s]`. Because reserved sequences are not delimited, there is no way to know where the whitespace belongs to the reserved or to the `[s]`. Future standardization would most likely use productions like `name` or `nmtoken` and the resulting trailing whitespace would "emerge" from the murky fog of `reserved` to be captured by `[s]`. Note that this is one of the reasons I originally proposed reserving the sigils to the function-like syntax.

I tend to agree with wanting to make the exterior whitespace not significant, in part because it interferes with user expectations. These two sequences are logically identical:

```
{      $foo :bar    }
{$foo :bar}
```

But these two are not:

```
{    $foo ~bar    }
{$foo ~bar}
```

I think we should fix this. Probably:

```abnf
reserved = reserved-start reserved-body reserved-end
reserved-end = ; reserved-body less SP / HTAB/ CR / LF
```

eemeli: Why do we need to distinguish the trailing spaces of `reserved` in the ABNF? Would it not be enough to include some statement like this in the text on reserved about future use?
> Future uses of _reserved_ expressions MUST NOT assign any significance to trailing whitespace.

spec/syntax.md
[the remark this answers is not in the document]
aphillips: removing the space-at-the-end makes this thread irrelevant

spec/syntax.md
aphillips: We should make the parser return the right tokens. See what you think of the revised abnf.

PR #381, opened 2023-05-17
Clarify that variables declarations may override previous ones

eemeli:
Naming things is hard, so we should make it as easy as possible. Specifically, I'm thinking here about the named parameters/arguments that a formatting call makes available within a message via variable references.

Let's say we have a numerical argument like `count` that we'd like to format using some set of options. That raw value will be initially represented as `$count` within the message, but what about the value-with-formatter-and-options that we'd like to derive from it? Could that not also be `$count`?

To use a minimal example, I would like for something like this to be valid and have the obvious meaning:
```
let $count = {$count :number minimumFractionDigits=2}
```

This PR adds spec language and examples clarifying that this is valid.

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -14,9 +14,11 @@ To resolve the value of a Variable,
 its Name is used to identify either a local variable,
 or a variable defined elsewhere.
 If a local variable and an externally defined one use the same name,
-the local variable takes precedence.
+the local variable takes precedence after its variable declaration.
+A variable declaration may refer in its expression
+to a previous or external variable with the same name as the declaration's target.
 
-It is an error for a local variable definition to
+It is an error for a variable declaration to
 refer to a local variable that's defined after it in the message.
 
 ## Error Handling
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -14,7 +14,7 @@ To resolve the value of a _variable_,
 its _name_ is used to identify either a local variable,
 or a variable defined elsewhere.
 
-If more than one _declaration_ binds a value to the same _name_,
+If the left-hand side _variable_ of more than one _declaration_ uses the same _name_,
 or if an externally defined variable and a _declaration_ use the same _name_,
 the resolved value of the most recent _declaration_ preceding the _variable_ is used.
 
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -18,7 +18,7 @@ If more than one _declaration_ binds a value to the same _name_,
 or if an externally defined variable and a _declaration_ use the same _name_,
 the resolved value of the most recent _declaration_ preceding the _variable_ is used.
 
-A _declaration_ MAY overwrite the value of a _variable_ for later parts of the _message_.
+A _declaration_ MAY override the value of a _variable_ for later parts of the _message_.
 
 Attempting to resolve a _variable_ with no preceding _declaration_ or external definition
 binding a value to its _name_ results in an Unresolved Variable error.
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -10,16 +10,18 @@ the successor to ICU MessageFormat, henceforth called ICU MessageFormat 1.0.
 
 ## Variable Resolution
 
-To resolve the value of a Variable,
-its Name is used to identify either a local variable,
+To resolve the value of a _variable_,
+its _name_ is used to identify either a local variable,
 or a variable defined elsewhere.
-If a local variable and an externally defined one use the same name,
-the local variable takes precedence after its variable declaration.
-A variable declaration may refer in its expression
-to a previous or external variable with the same name as the declaration's target.
 
-It is an error for a variable declaration to
-refer to a local variable that's defined after it in the message.
+If more than one _declaration_ binds a value to the same _name_,
+or if an externally defined variable and a _declaration_ use the same _name_,
+the resolved value of the most recent _declaration_ preceding the _variable_ is used.
+
+A _declaration_ MAY overwrite the value of a _variable_ for later parts of the _message_.
+
+Attempting to resolve a _variable_ with no preceding _declaration_ or external definition
+binding a value to its _name_ results in an Unresolved Variable error.
 
 ## Error Handling
 
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -275,6 +275,20 @@ This local variable can then be used in other expressions within the same messag
 declaration = let s variable [s] "=" [s] expression
 ```
 
+Declaration examples:
+
+```
+let $foo = {|A literal value|}
+```
+
+```
+let $foo = {:func opt=$bar}
+```
+
+```
+let $foo = {$foo :number minimumFractionDigits=2}
+```
+
 ### Selectors
 
 A `match` statement contains one or more **_selectors_**


spec/formatting.md
```diff
@@ -14,9 +14,11 @@ To resolve the value of a Variable,
 its Name is used to identify either a local variable,
 or a variable defined elsewhere.
 If a local variable and an externally defined one use the same name,
-the local variable takes precedence.
+the local variable takes precedence after its variable declaration.
+A variable declaration may refer in its expression
+to a previous or external variable with the same name as the declaration's target.
```
aphillips, line 19: I think this is good, but I think we could be more explicit. Perhaps instead of this paragraph we should say:

> Each `declaration` is processed sequentially. 
> A `declaration` **_may_** refer to previously declared `variables` (including itself) or to external variables in its `expression`. 
> A `declaration` that shares a name with a previous `declaration` or to an external variable overwrites the preceding or external variable.

---

Note the use of 2119 keywords and keyword quoting, which is our standard.

spec/formatting.md
```diff
@@ -14,9 +14,11 @@ To resolve the value of a Variable,
...
 
-It is an error for a local variable definition to
+It is an error for a variable declaration to
 refer to a local variable that's defined after it in the message.
```
aphillips, line 22:
```suggestion
It is an error for a `declaration` to refer to an undefined variable.
```

This statement is trickier than it looks. A `declaration` can refer to an external variable and there is no list of these. We should probably say which type of error it is?

aphillips (commented):
Good generally, but a couple of comments (mainly editorial)

spec/formatting.md
stasm: I don't feel strongly about the wording, although I like that @aphillips's suggestions is slightly more explicit. That said:

> Each declaration is processed sequentially.

I agree that processing declarations should generally work sequentially. At the same time, do we need to be prescriptive about it in terms of the implementation?

> A declaration may refer to previously declared variables (including itself) or to external variables in its expression.

Here, OTOH, I think we should rather say `(excluding itself)`. Otherwise we're getting into a teritorry of circular references.

eemeli: Apologies for nitpicking, but I think it would be helpful for us to use valid syntax when arguing about spec details. To wit:

```
let $foo = {|some value|}
let $foo = {$foo :functionA}
let $foo = {$foo :functionB}
let $foo = {||}
```

stasm: So, it's not really referring to *itself*, is it? It's referring to a different value, by identity, but with the same name. (I guess it really depends on whether we think of `let $foo` as a new variable, or a new name binding...)

catamorphism:
Also see https://github.com/unicode-org/message-format-wg/issues/310

spec/formatting.md
eemeli: This is covered in the Error Handling section, which specifies that this would be an Unresolved Variable error.

spec/formatting.md
catamorphism, line 19: I'm not sure what "target" means here, and it isn't defined; can this be rephrased as something like:

"In a variable declaration, `let v = e`, `e` may refer to `v`. While resolving `e`, the meaning of `v` is the meaning of the most recently defined local or external variable named `v`."

It's hard to be more precise than that without talking about lexical scope and environments. In fact, for that reason, my personal preference is to disallow name shadowing. If name shadowing is allowed, and the spec stays at this level of formality, it will be harder to validate whether implementations comply with the spec.

spec/formatting.md
catamorphism, line 22: Should this say something like: "It is an error for the right-hand side of a variable declaration to refer to a local variable that's defined after it in the message"? Otherwise, a reader who doesn't have the use-mention distinction totally clear in their mind would think that you're not allowed to write:

```
let $foo = 5
let $foo = 6
```

spec/syntax.md
```diff
@@ -275,6 +275,20 @@ This local variable can then be used in other expressions within the same messag
...
+let $foo = {|A literal value|}
+```
+
+```
+let $foo = {:func opt=$bar}
+```
+
+```
+let $foo = {$foo :number minimumFractionDigits=2}
+```
```
catamorphism, line 290: Even though it's explained in the formatting doc, is it worth explaining here as well that in the third example, either there must be a previous in-scope definition of `$foo`, or this declaration is syntactically correct but semantically invalid?

spec/formatting.md
aphillips: Yes, but this requires someone to go read a separate location in the spec to find out what type of error it is. We don't need to define error handling here, but it's friendlier to readers to say:

> A `declaration` that refers to an undefined variable results in an `Unresolved Variable error`. 
> Local variables **_must_** be defined before any `declaration` or `expression` that refers to them.

eemeli:
I've rewritten the text, mostly based on language proposed by @aphillips and @catamorphism. Please re-review?

aphillips (approved):
This could be committed. I made one suggestion on wording and one suggestion for an addition (of what is, I believe, an unstated assumption)

spec/formatting.md
```diff
@@ -10,14 +10,18 @@ the successor to ICU MessageFormat, henceforth called ICU MessageFormat 1.0.
...
+its _name_ is used to identify either a local variable,
 or a variable defined elsewhere.
-If a local variable and an externally defined one use the same name,
-the local variable takes precedence.
 
-It is an error for a local variable definition to
-refer to a local variable that's defined after it in the message.
+If more than one _declaration_ binds a value to the same _name_,
+or if an externally defined variable and a _declaration_ use the same _name_,
+the resolved value of the most recent _declaration_ preceding the _variable_ is used.
```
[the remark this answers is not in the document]
eemeli: I think the function purity / lack of side effects ought to be addressed with respect to function resolution, rather than here.

---

The proposed "is processed sequentially" statement is a bit problematic, as it may be interpreted to require and enforce an eager processing model. Consider this message:

```
let $foo = {:really-heavy-operation}
match {$bar}
when 0 {Nothing here}
when * {Have a {$foo}}
```

In this case, we should make sure that in the `bar == 0` case an implementation is not required to determine and then throw away the value of `$foo`.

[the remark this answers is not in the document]
catamorphism: Note that there is already an open issue about whether to require eager or lazy evaluation of local variables:

https://github.com/unicode-org/message-format-wg/issues/299

Thta said, I don't think either version of the wording commits to either eager or lazy evaluation. It depends on the interpretation of the word "processed". "Processing" could mean evaluating (resolving?) the right-hand side of the declaration and binding the left-hand side to the resulting value, or it could mean binding the left-hand side to a thunk representing the right-hand side.

catamorphism (changes requested):
I wrote some comments, but I think those can mostly be addressed in future PRs. The only change I'd like to see before merging is the "overwrite"/"override" one.

spec/formatting.md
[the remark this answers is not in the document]
aphillips: @eemeli I don't think we have to say anything about early or late binding of variables: we should not require early binding but we should not require late binding either. It's an implementation detail. What is important is that declarations behave as if they were evaluated in order.

I think we can let the text go in as-is and revisit.

spec/formatting.md
```diff
@@ -10,14 +10,18 @@ the successor to ICU MessageFormat, henceforth called ICU MessageFormat 1.0.
...
+
+A _declaration_ MAY override the value of a _variable_ for later parts of the _message_.
```
mihnita, line 21: I think that last time we proposed to not allow overriding.
Compared to programming, these messages are a relatively short "code block"
There should be no problem to just use a different name.
It just introduces complexity without any benefit that I can imagine.

spec/formatting.md
```diff
@@ -10,14 +10,18 @@ the successor to ICU MessageFormat, henceforth called ICU MessageFormat 1.0.
...
-its Name is used to identify either a local variable,
+To resolve the value of a _variable_,
+its _name_ is used to identify either a local variable,
 or a variable defined elsewhere.
-If a local variable and an externally defined one use the same name,
-the local variable takes precedence.
 
-It is an error for a local variable definition to
-refer to a local variable that's defined after it in the message.
+If more than one _declaration_ binds a value to the same _name_,
```
eemeli, line 17: Let's avoid the word "binds", as requested by @mihnita in https://github.com/unicode-org/message-format-wg/issues/310#issuecomment-1596280816.

```suggestion
If the left-hand side _variable_ of more than one _declaration_ uses the same _name_,
```

eemeli:
Closing, as we did not reach consensus on this in the WG. The action is now on @mihnita and @catamorphism to champion an alternative PR.

PR #385, opened 2023-05-22
Clarifications to pattern selection

catamorphism:
* State up front that the catch-all key `*` is never passed to the implementation-defined key-comparison method
  * Clarify the relationship between selector length and key length
  * Clarify the type signature for `SortVariants`
  * Fix typo

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -131,18 +131,22 @@ Finally, sort the list of variants `vars` and select the _pattern_:
          1. Let `ks` be the resolved value of `key`.
          1. Let `matchpref` be the integer position of `ks` in `matches`.
       1. Set the `tuple` integer value as `matchpref`.
-   1. Call the method SortVariants(`sortable`).
+   1. Set `sortable` to be `SortVariants(sortable)`.
    1. Set `i` to be `i` - 1.
 1. Let `var` be the _variant_ element of the first element of `sortable`.
 1. Select the _pattern_ of `var`.
 
-The method SortVariants is determined by the implementation.
-It takes as an argument a `sortable` list of (integer, _variant_) tuples,
-which it modifies in place using some stable sorting algorithm.
-The method does not return a value: `sortable` is its out-parameter.
-The list is sorted according to the tuple's first integer element,
-such that a lower number is sorted before a higher one,
-and entries that have the same number retain their order.
+`SortVariants` is a method whose single argument is
+a list of (integer, _variant_) tuples.
+It returns a list of (integer, _variant_) tuples.
+Any implementation of `SortVariants` is acceptable
+as long as it satisfies the following requirements:
+1. Let `sortable` be an arbitrary list of (integer, _variant_) tuples.
+1. Let `sorted` be `SortVariants(sortable)`.
+1. `sorted` is the result of sorting `sortable` using the following comparator:
+    1. `(i1, v1)` <= `(i2, v2)` if and only if `i1 <= i2`.
+1. The sort is stable (pairs of tuples from `sortable` that are equal
+   in their first element have the same relative order in `sorted`).
 
 ### Examples
 
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -27,11 +27,14 @@ the _pattern_ of one of the _variants_ must be selected for formatting.
 When a message has a single _selector_,
 an implementation-defined method compares each key to the _selector_
 and determines which of the keys match, and in what order of preference.
-A catch-all key will always match, but is always the least preferred choice.
-During selection, the _variant_ with the best-matching key is selected.
+The list of keys passed to this implementation-defined method
+does not include the catch-all key `*`.
+The formatter treats the catch-all key `*`
+as a match for any _selector_,
+but with the lowest possible preference.
 
 In a message with more than one _selector_,
-each _variant_ also has a corresponding number of keys.
+the number of keys in each _variant_ equals the number of _selectors_.
 These correspond to _selectors_ by position.
 The same implementation-defined method as above is used to compare
 the corresponding key of each _variant_ to its _selector_,
@@ -134,7 +137,7 @@ Finally, sort the list of variants `vars` and select the _pattern_:
 The method SortVariants is determined by the implementation.
 It takes as an argument a `sortable` list of (integer, _variant_) tuples,
 which it modifies in place using some stable sorting algorithm.
-The method does not return anything.
+The method does not return a value: `sortable` is its out-parameter.
 The list is sorted according to the tuple's first integer element,
 such that a lower number is sorted before a higher one,
 and entries that have the same number retain their order.
@@ -221,7 +224,7 @@ when * * {Otherwise}
 Presuming a more powerful implementation which supports selection on numerical values,
 and provides a `:plural` function that matches keys by their exact value
 as well as their plural category (preferring the former, if possible),
-and an Enligh-language formatting context in which
+and an English-language formatting context in which
 the variable reference `$count` resolves to the number `1`,
 pattern selection proceeds as follows for this message:
 
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -82,6 +82,7 @@ Next, using `res`, resolve the preferential order for all message keys:
    1. For each _variant_ `var` of the message:
       1. Let `key` be the `var` key at position `i`.
       1. If `key` is not the catch-all key `'*'`:
+         1. Assert that `key` is a _literal_.
          1. Let `ks` be the resolved value of `key`.
          1. Append `ks` as the last element of the list `keys`.
    1. Let `rv` be the resolved value at index `i` of `res`.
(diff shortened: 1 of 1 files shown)


eemeli (approved):

catamorphism:
As per #384 I added another commit that changes references to "decoding" keys to "resolving" them. I'm not sure if the text I added is clear enough, but am open to suggestions. (Also, maybe this should land after #382 does, since it now refers to the "Literal Resolution" section that isn't there yet.)

spec/formatting.md
```diff
@@ -77,7 +82,7 @@ Next, using `res`, resolve the preferential order for all message keys:
    1. For each _variant_ `var` of the message:
       1. Let `key` be the `var` key at position `i`.
       1. If `key` is not the catch-all key `'*'`:
-         1. Let `ks` be the decoded value of `key`.
+         1. Let `ks` be the resolved value of `key`.
```
eemeli, line 85:
```suggestion
         1. Assert that `key` is a _literal_.
         1. Let `ks` be the resolved value of `key`.
```

eemeli (commented):
The current proposal seems a bit cumbersome in how it explicitly links with references the method introduction, its body, and another named section of the spec. It might be easier to drop it and include assertions about `key` values in the method. The suggested change below could be prepended to all three instances, and the added lines 49-54 dropped.

eemeli (approved):

spec/formatting.md
```diff
@@ -27,11 +27,13 @@ the _pattern_ of one of the _variants_ must be selected for formatting.
 When a message has a single _selector_,
 an implementation-defined method compares each key to the _selector_
 and determines which of the keys match, and in what order of preference.
-A catch-all key will always match, but is always the least preferred choice.
-During selection, the _variant_ with the best-matching key is selected.
+The list of keys passed to this implementation-defined method
+does not include the catch-all key `*`.
+The catch-all key `*` is treated as a match for any _selector_,
+but with the lowest possible preference.
```
aphillips, line 33: Wait... why are we doing this? What is wrong with letting the "implementation-defined method" decide that the `*` key is the best match?

What you're requiring here is that the selector method return "no match" in order to reach `*` and that the higher-level caller of the "implementation-defined method" then choose the fallback vs. allowing the selector method to make that determination. We haven't discussed this previously, so I don't understand why we are changing this text.

spec/formatting.md
```diff
@@ -27,11 +27,13 @@ the _pattern_ of one of the _variants_ must be selected for formatting.
...
 
 In a message with more than one _selector_,
-each _variant_ also has a corresponding number of keys.
+the number of keys in each _variant_ equals the number of _selectors_.
```
aphillips, line 36: use normative language


```suggestion
the number of keys in each _variant_ **_must_** equal the number of _selectors_.
```

spec/formatting.md
```diff
@@ -134,7 +139,7 @@ Finally, sort the list of variants `vars` and select the _pattern_:
 The method SortVariants is determined by the implementation.
 It takes as an argument a `sortable` list of (integer, _variant_) tuples,
 which it modifies in place using some stable sorting algorithm.
-The method does not return anything.
+The method does not return a value: `sortable` is its out-parameter.
```
aphillips, line 142: "out-parameter"?

I actually find this a bit odd or maybe constraining. This defines a method whose output is a side-effect on the input value, which is unusual and it forbids writing a method with a signature like:

```java
public MessageKeyList sortVariants(final MessageKeyList input)
```

Why is it necessary for us to force this?

spec/formatting.md
```diff
@@ -221,7 +226,7 @@ when * * {Otherwise}
 Presuming a more powerful implementation which supports selection on numerical values,
 and provides a `:plural` function that matches keys by their exact value
 as well as their plural category (preferring the former, if possible),
-and an Enligh-language formatting context in which
+and an English-language formatting context in which
 the variable reference `$count` resolves to the number `1`,
 pattern selection proceeds as follows for this message:
```
aphillips, line 231: Good typo fix.

I find this example nearly impossible to read?

I don't agree with the idea of an "English-language formatting context". It should be in terms of a locale.

I'll come back an suggest an edit when I have a few more minutes.

aphillips (changes requested):

spec/formatting.md
eemeli: I don't think this is intended as a change to the selection method, but a clarification of what's already in the Resolve Preferences section below.

If you think that should be changed, that might be best discussed in a separate issue or PR.

spec/formatting.md
eemeli: This is intended as a convenience for keeping the method simple. Recall that the section includes this near the top:
> An implementation MAY use any pattern selection method, as long as its observable behaviour matches the results of the method defined here.

On the other hand, if the sense is that it would be easier to read if the implementation-defined method returned some value that was then re-assigned to `sortable`, that's certainly a possible change.

spec/formatting.md
catamorphism: @aphillips Yes, as Eemeli said, it's meant to be a clarification of what's already there. The existing definitions of "Resolve Preferences", "Filter Variants", and "Sort Variants" only pass literals to "the implementation-defined method". So the method can do whatever it wants if it's passed anything that's not a literal, such as `*`. Saying "The list of keys passed to this implementation-defined method does not include the catch-all key `*`" is another way of saying that these definitions don't rely on undefined behavior.

spec/formatting.md
catamorphism: @aphillips The way the algorithm is currently written, `sortable` has to change on each iteration of the `while` loop. However, I think I've found a way to leave it open whether `SortVariants` modifies the list in-place and returns the same list as its result, or returns a new list. Let me know what you think.

spec/formatting.md
catamorphism: The edit could maybe be a separate PR?

catamorphism:
I re-requested review from @aphillips to confirm that it's OK for me to resolve the 3 open conversations, and re-requested from @eemeli for the sorting change that I made in response to the suggestion from @aphillips ( https://github.com/unicode-org/message-format-wg/pull/385/commits/f08b9b60a94321d17fb82b8e16d7ae3b63ae770a ), which seems significant enough to merit another look.

spec/formatting.md
```diff
@@ -123,21 +127,26 @@ Finally, sort the list of variants `vars` and select the _pattern_:
       1. Let `matchpref` be an integer with the value `minpref`.
       1. Let `key` be the `tuple` _variant_ key at position `i`.
       1. If `key` is not the catch-all key `'*'`:
-         1. Let `ks` be the decoded value of `key`.
+         1. Assert that `key` is a _literal_.
+         1. Let `ks` be the resolved value of `key`.
          1. Let `matchpref` be the integer position of `ks` in `matches`.
       1. Set the `tuple` integer value as `matchpref`.
-   1. Call the method SortVariants(`sortable`).
+   1. Set `sortable` to be `SortVariants(sortable)`.
```
eemeli, line 134: Nitpicky, but let's be a bit more explicit here, and follow the same style of expression as on line 83.

```suggestion
   1. Set `sortable` to be the result of calling the method `SortVariants(sortable)`.
```

spec/formatting.md
```diff
@@ -123,21 +127,26 @@ Finally, sort the list of variants `vars` and select the _pattern_:
...
-The method does not return anything.
-The list is sorted according to the tuple's first integer element,
-such that a lower number is sorted before a higher one,
-and entries that have the same number retain their order.
+`SortVariants` is a method whose single argument is
+a list of (integer, _variant_) tuples.
+It returns a list of (integer, _variant_) tuples.
+Any implementation of `SortVariants` is acceptable
+as long as it satisfies the following requirements:
+1. Let `sortable` be an arbitrary list of (integer, _variant_) tuples.
```
eemeli, line 144:
```suggestion
as long as it satisfies the following requirements:

1. Let `sortable` be an arbitrary list of (integer, _variant_) tuples.
```

eemeli (approved):
It's still good. Two stylistic nitpicks inline.

spec/formatting.md
eemeli: @aphillips Would you prefer the current proposed change to be included or dropped from this PR? Getting this merged would make it easier for you to file a subsequent PR with your proposed rewrite.

spec/formatting.md
aphillips: Actually, on  second thought... it's a typo correct. Fix the typo. We'll address the text separately.

spec/formatting.md
aphillips: On the one hand, I would prefer not to merge non-consensus text wherever possible. But I agree that the lack of other reviewers leaves little evidence about whether this should or should not go.

Let's merge this. I will file my comments as a separate issue and create a PR with them once this is merged. Sound like a plan?

aphillips (approved):
As noted, I'm not keen on the current structure of the text, but the changes here make the *existing* text consistent with the changes to the ABNF. Let's put this in and sort out my concerns separately.

PR #388, opened 2023-06-05
Change the "pattern selection" text

aphillips:
In PR #385 from @catamorphism  I made the [comment](https://github.com/unicode-org/message-format-wg/pull/385#discussion_r1211860808) which suggested a different approach to pattern selection. This PR addresses that comment.

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -31,31 +31,32 @@ refer to a local variable that's defined after it in the message.
 
 ## Pattern Selection
 
-When a _message_ contains a _match_ statement with one or more _selectors_, it needs to determine
-which _variant_ will be used to provide the _pattern_ for the formatting operation. This is done by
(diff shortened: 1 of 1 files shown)


spec/formatting.md
eemeli, line 1: Could you reformat the changes to this file to use [semantic line breaks](https://sembr.org/)? It would make the diff much easier to read.

eemeli (approved):
Mostly would like yet moar newlines. The line 35 change is the only one adding a few words.

This is a bit less precise than the previous text, but I'm fine with that.

aphillips:
@eemeli Thanks for the review. I have committed your changes to the line-breaks. The wording change was also committed (since it was an improvement/oversight on my part).

mihnita (approved):

PR #389, opened 2023-06-06
Make pattern selection example 3 clearer

aphillips:
Editorial changes to make this example easier to understand.

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -238,18 +238,20 @@ when * * {Otherwise}
 
 #### Example 3
 
-Presuming a more powerful implementation which supports selection on numerical values,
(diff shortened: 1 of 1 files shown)


eemeli (approved):
This text is indeed less clunky. Needs newlines, though.

I liked the `>` styling you used for the example in #388. Should we apply that to all examples, and mention somewhere that such text is non-normative? Then we wouldn't need to specify that at every instance.

eemeli:
@aphillips I added a merge commit from `main` to this branch in order to simplify the diff presented by GitHub. Squashing base branch changes onto a PR branch tends to mess it up pretty badly.

stasm (approved):

PR #392, opened 2023-06-17
Harmonise spec style

eemeli:
This fixes the following editorial issues in the spec:
- Always style normative key words using all caps: MUST, MAY, etc.
- Always use **_definition_** and _reference_ styles
- Always use `>` block quotes for examples
- Always use semantic line breaks
- Apply Prettier markdown styling
- Update ToC in readme


eemeli:
Requesting fast-tracking for these editorial fixes.

aphillips (approved):

aphillips:
These are strictly editorial changes to make the formatting consistent. I see no need to hold these back.

PR #395, opened 2023-06-17
Explicitly define `operand` in syntax

eemeli:
While working on the data model and formatting, I found it a bit challenging to not have the _operand_ of an expression defined in the syntax, even though it needs to be referred to all over the place.

So let's add that, and move the _function_ sigil description to the right section of the spec.

### DIFF
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -125,29 +125,23 @@ A simple _expression_ is a bare variable name:
 ### Formatting Functions
 
 A _function_ is named functionality, possibly with _options_, that format,
-process, or operate on a _variable_.
+process, or operate on an _operand_ which may be either a _literal_ or a _variable_.
 
-For example, a _message_ with an interpolated `$date` _variable_ formatted with the `:datetime` _function_:
+For example, a _message_ with a `$date` _variable_ formatted with the `:datetime` _function_:
 
     {Today is {$date :datetime weekday=long}.}
 
-A _message_ with an interpolated `$userName` _variable_ formatted with
+A _message_ with a `$userName` _variable_ formatted with
 the custom `:person` _function_ capable of
 declension (using either a fixed dictionary, algorithmic declension, ML, etc.):
 
     {Hello, {$userName :person case=vocative}!}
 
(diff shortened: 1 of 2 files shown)


eemeli:
To bikeshed on the name a bit, one possible alternative to `operand` could be `value`; that would be a bit less arcane, and would make it easier to use it also in `option`:

```abnf
expression = "{" [s] ((value [s annotation]) / annotation) [s] "}"
value = literal / variable
option = name [s] "=" [s] value
```

macchiati:
Not a blocker, but the value of 'operand' is that it reduces confusion with 'value' used as a normal English word.

macchiati:
In your very next PR, you have in the text "...determines the value of a part of the message..."

stasm (approved):
You can also change line 19, which currently reads:

spec/syntax.md
```diff
@@ -347,19 +341,27 @@ Whitespace within a _pattern_ is meaningful and MUST be preserved.
...
 
 _Functions_ do not accept any positional arguments
-other than the _literal_ or _variable_ in front of them.
+other than the _operand_ in front of them.
+
+_Functions_ use one of the following prefix sigils:
+
+- `:` for standalone content
+- `+` for starting or opening elements
+- `-` for ending or closing elements
```
aphillips, line 360: Instead of "elements", say "_expressions_", since that is our name for a placeable element.

aphillips (changes requested):
Looks good, but one comment.

spec/syntax.md
eemeli, line 360:
```suggestion
- `+` for starting or opening _expressions_
- `-` for ending or closing _expressions_
```

eemeli: Done.

eemeli:
> You can also change line 19, which currently reads:
> 
> ```
> option = name [s] "=" [s] (literal / variable)
> ```

Noticing that is what prompted me to consider _value_, actually, as "operand" doesn't really work conceptually as the value of an option. :/

PR #396, opened 2023-06-17
Add missing formatting sections

eemeli:
This is still incomplete and will need plenty of additional work, but I thought I'd share at least the current shape of my thoughts.

@aphillips You mentioned being able to potentially help out a bit with this? It's missing at least some description of what actually goes on in formatting a resolved pattern, but what else do we need to include? The bidi stuff is explicitly left out here, as that's progressing in its own PR.

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -12,8 +12,8 @@ their handling during formatting is specified here as well.
 
 Formatting of a _message_ is defined by the following operations:
 
-- **_Resolution_** determines the value of a part of the message,
-  such as an _expression_, with reference to the current _formatting context_.
+- **_Expression Resolution_** determines the value of an _expression_,
+  with reference to the current _formatting context_.
   This can include multiple steps,
   such as looking up the value of a variable and calling formatting functions.
   The resolved value is not necessarily in the shape it will finally take,
@@ -30,9 +30,12 @@ Formatting of a _message_ is defined by the following operations:
   > The shape of this value is implementation-dependent,
   > and different implementations may choose to perform different levels of resolution.
 
+  The resolution of _text_ is rather straighforward,
+  and is detailed under _literal resolution_.
+
 - **_Pattern Selection_** determines which of a message's _patterns_ is formatted.
   For a message with no _selectors_, this is simple as there is only one _pattern_.
-  With _selectors_, this will depend on their _resolution_.
+  With _selectors_, this will depend on their resolution.
 
 - **_Formatting_** takes the resolved values of the selected _pattern_,
   and formats them in the desired shape.
@@ -44,7 +47,7 @@ Formatting of a _message_ is defined by the following operations:
   > rather than as a representation of its HTML source.
 
 Formatter implementations are not required to expose
-the _resolution_ and _pattern selection_ operations to their users,
+the _expression resolution_ and _pattern selection_ operations to their users,
 or even use them in their internal processing,
 as long as the final _formatting_ result is made available to users
 and the observable behavior of the formatter matches that described here.
@@ -52,7 +55,7 @@ and the observable behavior of the formatter matches that described here.
 ## Formatting Context
 
 A message's **_formatting context_** represents the data and procedures that are required
-for the message's _resolution_, _pattern selection_ and _formatting_.
+for the message's _expression resolution_, _pattern selection_ and _formatting_.
 
 At a minimum, it includes:
 
@@ -75,9 +78,7 @@ At a minimum, it includes:
 
 Implementations MAY include additional fields in their _formatting context_.
 
-## Resolution
-
-### Expression Resolution
+## Expression Resolution
 
 _Expressions_ are used in _declarations_, _selectors_, and _patterns_.
 
@@ -154,9 +155,8 @@ the following steps are taken:
 4. Call the function implementation with the following arguments:
 
    - The current _locale_.
-   - The resolved value of the _operand_, or if there is no _operand_,
-     an empty value.
    - The resolved mapping of _options_.
+   - If the _expression_ includes an _operand_, its resolved value.
 
    The shapes of the resolved _operand_ and _option_ values are implementation-defined.
 
@@ -175,7 +175,7 @@ the following steps are taken:
 
 ### Fallback Resolution
 
-The _resolution_ of an _expression_ may fail in the following cases:
+The resolution of an _expression_ may fail in the following cases:
 
 - A _variable_ _operand_ fails to resolve.
 - A _function_ _annotation_ fails to resolve.
@@ -274,7 +274,7 @@ or if this is not available or empty, the U+FFFD REPLACEMENT CHARACTER `�`.
 
 ### Resolve Selectors
 
-First, resolve the values of each _selector_:
+First, resolve the values of each _selector_ _expression_:
 
 1. Let `res` be a new empty list of resolved values that support selection.
 1. For each _expression_ `exp` of the message's _selectors_,
@@ -777,7 +777,7 @@ or contains some error which leads to further errors,
 an implementation which does not emit all of the errors
 SHOULD prioritise Syntax and Data Model errors over others.
 
-When an error occurs in the _resolution_ of an _option_,
+When an error occurs in the resolution of an _option_,
 the surrounding _expression_ MUST be processed as if the _option_ were not present.
 This MAY allow the _expression_ to resolve to a non-fallback _value_,
 though an error MUST still be emitted.
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -19,17 +19,6 @@ Formatting of a _message_ is defined by the following operations:
   The resolved value is not necessarily in the shape it will finally take,
   but is "formattable", i.e. it contains everything required by the eventual formatting.
 
-  > For example, the resolved value of the _expression_ `{|0.40| :number style=percent}`
-  > could be an object such as
-  >
-  > ```
-  > { value: Number('0.40'),
-  >   formatter: NumberFormat(locale, { style: 'percent' }) }
-  > ```
-  >
-  > The shape of this value is implementation-dependent,
-  > and different implementations may choose to perform different levels of resolution.
-
   The resolution of _text_ is rather straighforward,
   and is detailed under _literal resolution_.
 
@@ -38,13 +27,9 @@ Formatting of a _message_ is defined by the following operations:
   With _selectors_, this will depend on their resolution.
 
 - **_Formatting_** takes the resolved values of the selected _pattern_,
-  and formats them in the desired shape.
-  This specification only defines formatting messages as a single concatenated string,
-  but implementations SHOULD provide formatters for additional shapes
-  as appropriate for their setting.
-
-  > For example, a formatter in a web browser could format a message as a DOM fragment
-  > rather than as a representation of its HTML source.
+  and produces the formatted result for the _message_.
+  Depending on the implementation, this result could be a single concatenated string,
+  an array of objects, an attributed string, or some other locally appropriate data type.
 
 Formatter implementations are not required to expose
 the _expression resolution_ and _pattern selection_ operations to their users,
@@ -90,6 +75,20 @@ implementations SHOULD NOT immediately fully format the value for output.
 In _selectors_, the resolved value of an _expression_ is used for _pattern selection_.
 
 In a _pattern_, the resolved value of an _expression_ is used in its _formatting_.
+
+The shapes of resolved values are implementation-dependent,
+and different implementations MAY choose to perform different levels of resolution.
+
+> For example, the resolved value of the _expression_ `{|0.40| :number style=percent}`
+> could be an object such as
+>
+> ```
+> { value: Number('0.40'),
+>   formatter: NumberFormat(locale, { style: 'percent' }) }
+> ```
+>
+> Alternatively, it could be an instance of an ICU4J `FormattedNumber`,
+> or some other locally appropriate value.
 
 Depending on the presence or absence of an _operand_
 and a _function_ or _reserved_ _annotation_,
@@ -498,29 +497,44 @@ each _text_ and _expression_ part of the selected _pattern_ must be resolved and
 
 _Formatting_ is a mostly implementation-defined process,
 as it depends on the implementation's shape for resolved values
-and the target shape of the formatting.
+and the result type of the formatting.
 
 Formatting errors MAY be emitted during _formatting_,
-as formatting is not defined on every resolved value.
+as formatting is not necessarily defined on every resolved value.
 A formatter MAY provide a value to use in such a case instead of a _fallback value_.
 
-_Formatting_ MAY produce formatted messages in the following shapes,
+_Formatting_ MAY produce formatted messages with the following data types,
 as well as any others:
 
 - A single concatenated string.
+- A string with associated attributes for portions of its text.
 - A flat sequence of objects corresponding to each resolved value.
 - A hierarchical structure of objects that group spans of resolved values,
   such as sequences delimited by "open" and "close" _function_ _annotations_.
 
-Implementations SHOULD provide _formatting_ output shapes that match user needs,
-such that e.g. re-parsing a message formatted as a string is not required.
+Implementations SHOULD provide _formatting_ result types that match user needs,
+including situations that require further processing of formatted messages.
 Implementations SHOULD encourage users to consider a formatted localised string
 as an opaque data structure, suitable only for presentation.
+
+### Examples
+
+_This section is non-normative._
+
+1. An implementation might choose to return an interstitial object
+   so that the caller can "decorate" portions of the formatted value.
+   In ICU4J, the `NumberFormatter` class returns a `FormattedNumber` object,
+   so a _pattern_ such as `{This is my number {42 :number}}` might return
+   the character sequence `This is my number `
+   followed by a `FormattedNumber` object representing the value `42` in the current locale.
+
+2. A formatter in a web browser could format a message as a DOM fragment
+   rather than as a representation of its HTML source.
 
 ### Formatting Fallback Values
 
 If the resolved _pattern_ includes any _fallback values_
-and the formatting target is a concatenated string or a sequence of strings,
+and the formatting result is a concatenated string or a sequence of strings,
 the string representation of each _fallback value_ MUST be the concatenation of
 a U+007B LEFT CURLY BRACKET `{`,
 the _fallback value_ as a string,
@@ -586,7 +600,7 @@ which MAY also introspect the _pattern_'s _text_ values
 and identify situations where isolate characters are not needed
 or where additional or different isolation would produce better results.
 
-If an implementation provides formatting to non-string targets,
+If an implementation provides formatting to non-string result types,
 it SHOULD provide similar strategies for enabling bidirectional isolation,
 where appropriate.
 
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -111,7 +111,7 @@ after any character escape has been converted to the escaped character.
 When a _literal_ is used as an _operand_
 or on the right-hand side of an _option_,
 the formatting function MUST treat its resolved value the same
-independently of whether the value was originally _quoted_ or _unquoted_.
+whether its value was originally _quoted_ or _unquoted_.
 
 > For example,
 > the _option_ `foo=42` and the _option_ `foo=|42|` are treated as identical.
@@ -136,19 +136,21 @@ this MUST also be considered a failure.
 
 ### Function Resolution
 
-To resolve the value of an _expression_ with a _function_ _annotation_,
+To resolve an _expression_ with a _function_ _annotation_,
 the following steps are taken:
 
 1. If the _expression_ includes an _operand_, resolve its value.
    If this fails, use a _fallback value_ for the _expression_.
 2. Based on the _function_ starting sigil and _name_,
    find the appropriate function implementation from the _function registry_.
-   If this fails, emit an Unknown Function error
+   If the registry does not define an implementation for this _name_,
+   emit an Unknown Function error
    and use a _fallback value_ for the _expression_.
-3. Resolve the _option_ values as a mapping of string identifiers to values.
-   For each _option_, resolve its right-hand side value.
-   If this succeeds, assign it to the mapping with the _name_ of the _option_.
-   If this fails, do not set any value in the resolved mapping for this _option_.
+3. Resolve the _option_ values to a mapping of string identifiers to values.
+   For each _option_:
+     * If its right-hand side successfully resolves to a value,
+       bind the _name_ of the _option_ to the resolved value in the mapping.
+     * Otherwise, do not bind the _name_ of the _option_ to any value in the mapping.
 4. Call the function implementation with the following arguments:
 
    - The current _locale_.
@@ -156,9 +158,9 @@ the following steps are taken:
      an empty value.
    - The resolved mapping of _options_.
 
-   The shapes of the resolved _operand_ and _option_ values is implementation-defined.
+   The shapes of the resolved _operand_ and _option_ values are implementation-defined.
 
-   An implementation MAY include additional arguments for the function call,
+   An implementation MAY pass additional arguments to the function,
    as long as reasonable precautions are taken to keep the function interface
    simple and minimal, and avoid introducing potential security vulnerabilities.
 
@@ -175,11 +177,11 @@ the following steps are taken:
 
 The _resolution_ of an _expression_ may fail in the following cases:
 
-- A failure in the resolution of a _variable_ _operand_.
-- A failure in the resolution of a _function_ _annotation_.
-- The presence of a _reserved_ _annotation_.
+- A _variable_ _operand_ fails to resolve.
+- A _function_ _annotation_ fails to resolve.
+- The _expression_ has a _reserved_ _annotation_.
 
-In each such case, and error MUST be emitted
+In each such case, an error MUST be emitted
 and a **_fallback value_** used for the _expression_.
 This value depends on the shape of the _expression_:
 
@@ -206,12 +208,13 @@ _Option_ names and values are not included in the _fallback value_.
 
 When an error occurs in an _expression_ with a _variable_ _operand_
 and the _variable_ refers to a local _declaration_,
-the _fallback value_ is formatted based on the _expression_ of the _declaration_,
+the _fallback value_ is formatted based on the _expression_
+on the right-hand side of the _declaration_,
 rather than the _expression_ in the _selector_ or _pattern_.
 
 > For example,
-> attempting to format either of the following messages within a context that
-> does not provide for the function `:func` to be successfully resolved:
+> in a context in which the function `:func` fails to resolve,
+> attempting to format either of the following messages:
 >
 > ```
 > let $var = {|horse| :func}
@@ -230,7 +233,7 @@ _Pattern selection_ is not supported for _fallback values_.
 
 ## Pattern Selection
 
-When a _message_ contains a _match_ statement with one or more _expressions_,
+When a _message_ contains a _match_ construct with one or more _expressions_,
 the implementation needs to determine which _variant_ will be used
 to provide the _pattern_ for the formatting operation. 
 This is done by ordering and filtering the available _variant_ statements
@@ -491,14 +494,14 @@ when *   {Other match}
 ## Formatting
 
 After _pattern selection_,
-each _text_ and _expression_ of the selected _pattern_ must be resolved and formatted.
+each _text_ and _expression_ part of the selected _pattern_ must be resolved and formatted.
 
 _Formatting_ is a mostly implementation-defined process,
 as it depends on the implementation's shape for resolved values
 and the target shape of the formatting.
 
 Formatting errors MAY be emitted during _formatting_,
-as the successful _resolution_ of an _expression_ does not guarantee success in its _formatting_.
+as formatting is not defined on every resolved value.
 A formatter MAY provide a value to use in such a case instead of a _fallback value_.
 
 _Formatting_ MAY produce formatted messages in the following shapes,
@@ -653,8 +656,9 @@ These are divided into the following categories:
     > when * 1 {Second is one}
     > ```
 
-  - **Duplicate Option Name errors** occur when the same _name_ is used
-    for more than one _option_ in the same _expression_.
+  - **Duplicate Option Name errors** occur when the same _name_ 
+    appears on the left-hand side
+    of more than one _option_ in the same _expression_.
 
     > Example invalid messages resulting in a Duplicate Option Name error:
     >
@@ -761,8 +765,8 @@ During selection, an _expression_ handler MUST only emit Resolution and Selectio
 During formatting, an _expression_ handler MUST only emit Resolution and Formatting errors.
 
 Resolution and Formatting errors in _expressions_ that are not used
-in _pattern selection_ or _formatting_ MAY be ignored
-as such do not impact the current message's formatting.
+in _pattern selection_ or _formatting_ MAY be ignored,
+as they do not affect the output of the formatter.
 
 In all cases, when encountering an error,
 a message formatter MUST provide some representation of the message.
@@ -775,9 +779,9 @@ SHOULD prioritise Syntax and Data Model errors over others.
 
 When an error occurs in the _resolution_ of an _option_,
 the surrounding _expression_ MUST be processed as if the _option_ were not present.
-This MAY allow for the _expression_ _resolution_ not resulting in a _fallback value_,
+This MAY allow the _expression_ to resolve to a non-fallback _value_,
 though an error MUST still be emitted.
 
 When an error occurs within a _selector_,
 the _selector_ MUST NOT match any _variant_ _key_ other than the catch-all `*`
-and a Resolution or Selector error is emitted.
+and a Resolution or Selector error MUST be emitted.
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -1,24 +1,124 @@
-# WIP DRAFT MessageFormat 2.0 Formatting Behaviour
+# DRAFT MessageFormat 2.0 Formatting
 
 ## Introduction
 
 This document defines the behaviour of a MessageFormat 2.0 implementation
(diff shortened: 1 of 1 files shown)


eemeli:
This might be ready now? I'll go over it again before the call tomorrow, as there's undoubtedly stuff still missing.

spec/formatting.md
```diff
@@ -1,24 +1,113 @@
...
 
 This document defines the behaviour of a MessageFormat 2.0 implementation
 when formatting a message for display in a user interface, or for some later processing.
 
-The document is part of the MessageFormat 2.0 specification,
-the successor to ICU MessageFormat, henceforth called ICU MessageFormat 1.0.
+Message formatting is defined in terms of the following stages:
 
-## Literal Resolution
+- **_Resolution_** determines the value of a part of the message,
```
aphillips, line 10: You might be missing an "assignment" stage (processing of `let` statements), which occurs before selection (and might contain some "resolutions")?

spec/formatting.md
```diff
@@ -1,24 +1,113 @@
...
+# DRAFT MessageFormat 2.0 Formatting Behaviour
 
 ## Introduction
 
 This document defines the behaviour of a MessageFormat 2.0 implementation
 when formatting a message for display in a user interface, or for some later processing.
 
-The document is part of the MessageFormat 2.0 specification,
-the successor to ICU MessageFormat, henceforth called ICU MessageFormat 1.0.
+Message formatting is defined in terms of the following stages:
```
aphillips, line 8: I'm not sure if "stages" is the right way to think of this. In particular, some "stages" contain (or can be thought of as-if they contain) another stage. Perhaps "operations" would be more useful here?

spec/formatting.md
```diff
@@ -1,24 +1,113 @@
...
 
-## Literal Resolution
+- **_Resolution_** determines the value of a part of the message,
+  such as an _expression_, with reference to the current _formatting context_.
+  This may include multiple steps,
```
aphillips, line 12: In general, avoid normative keywords unless used normatively

```suggestion
This can include multiple steps, 
```

spec/formatting.md
```diff
@@ -1,24 +1,113 @@
...
+  > { value: Number('0.40'),
+  >   formatter: NumberFormat(locale, { style: 'percent' }) }
+  > ```
+  >
+  > The shape of this value is completely implementation-dependent,
+  > and different implementations may choose to perform different levels of resolution.
+
+- **_Pattern Selection_** determines which of a message's _patterns_ is formatted.
+  For a message with no _selectors_, this is simple as there is only one _pattern_.
+  With _selectors_, this will depend on their _resolution_.
```
aphillips, line 30: instead of selectors doing resolution, why not point to the section about selection?

spec/formatting.md
```diff
@@ -1,24 +1,113 @@
...
+  > For example, the resolved value of the _expression_ `{|0.40| :number style=percent}`
+  > could be an object such as
+  >
+  > ```
+  > { value: Number('0.40'),
+  >   formatter: NumberFormat(locale, { style: 'percent' }) }
+  > ```
+  >
+  > The shape of this value is completely implementation-dependent,
+  > and different implementations may choose to perform different levels of resolution.
```
aphillips, line 26: here `MAY` is normative and should be formatted normatively

spec/formatting.md
```diff
@@ -1,24 +1,113 @@
...
-The resolved value of _text_, _literal_ and _nmtoken_ tokens
+  > For example, the resolved value of the _expression_ `{|0.40| :number style=percent}`
+  > could be an object such as
+  >
+  > ```
+  > { value: Number('0.40'),
+  >   formatter: NumberFormat(locale, { style: 'percent' }) }
+  > ```
+  >
+  > The shape of this value is completely implementation-dependent,
```
aphillips, line 25: "completely" is unnecessary. Remove.

spec/formatting.md
```diff
@@ -1,24 +1,113 @@
...
+
+- **_Pattern Selection_** determines which of a message's _patterns_ is formatted.
+  For a message with no _selectors_, this is simple as there is only one _pattern_.
+  With _selectors_, this will depend on their _resolution_.
+
+- **_Formatting_** takes the resolved values of the selected _pattern_,
+  and formats them in the desired shape.
+  This specification only defines formatting messages as a single concatenated string,
+  but implementations SHOULD provide formatters for additional shapes
+  as appropriate for their setting.
```
aphillips, line 36: This is what I was mentioning in the call. We should start to think in terms of `formatToParts`. We can't say how the implementation works or what the name of objects/classes/data structures are. But we can describe formatting as resolving the _message_ into a logical sequence of (um, erm) values. 

---

A formatted message as a whole has some properties associated with it: locale, base direction, and a sequence of "parts"

Each "part" also has a set of properties. The parts are in a logically ordered sequence or array. Each part has a locale and a base direction property. Additional properties MAY be defined by the implementation.

There are two kinds of "part": a "literal part" and an "expression" part.

Each "literal part" consists of a string.

An "expression part" can be resolved to a sequence of zero or more "literal parts".

The string output of a message is the concatenated sequence of resolved literal parts. 

> Here is a simple terrible example:
> Inputs:
> | What | Value | Description |
> |---|---|---|
> | Locale | ar-AE | Locale to use for formatting |
> | `date` | `2023-06-19` | Data value passed to formatter |
>
> Message:
> 
> `{The example date is {$date :datetime skeleton=yMMMd}}`
>
> Output:
> ```
> {
>    "locale": "ar-AE",
>    "direction": "ltr",
>    "parts": [
>        {
>            "type": "literal",
>            "locale": "ar-AE",
>            "direction": "ltr",
>            "value": "The example date is "
>        },
>        {
>            "type": "expression",
>            "locale": "ar-AE",
>            "direction": "rtl",
>            "value": [
>                   { "type": "literal", "locale":"ar-AE", "dir": "rtl", "name": "day", "19" },
>                   { "type": "literal", "locale":"ar-AE", "dir": "rtl", "name": "separator", " " },
>                   { "type": "literal", "locale":"ar-AE", "dir": "rtl", "name": "month", "يونيو" },
>                   { "type": "literal", "locale":"ar-AE", "dir": "rtl", "name": "separator", " " },
>                   { "type": "literal", "locale":"ar-AE", "dir": "rtl", "name": "year", "2023" },
>            ]
>        },
>    ]
> }
> ```

Callers can consume the sequence of parts in order to perform additional processing, such as markup. The above example might be formatted into an HTML context thusly:

> ```html
> <p lang="ar-AE" dir="ltr">The example date is 
>     <span dir="rtl" id="date">19 <em>يونيو</em> 2023</span></p>
> ```

aphillips (changes requested):
A couple of short edits and one really long comment with an alternative approach.

spec/formatting.md
```diff
@@ -1,24 +1,113 @@
-# WIP DRAFT MessageFormat 2.0 Formatting Behaviour
+# DRAFT MessageFormat 2.0 Formatting Behaviour
```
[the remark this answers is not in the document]
eemeli: Ach, that's hard as the speller in question is between my ears. But dropped "behaviour" nevertheless.

spec/formatting.md
eemeli: Switched to "operations" for now. Not sure it's really right, but it's not horrible either.

spec/formatting.md
eemeli: That's done implicitly through variable resolution. There's also this on line 736:

> Resolution and Formatting errors in _expressions_ that are not used
> in _pattern selection_ or _formatting_ MAY be ignored
> as such do not impact the current message's formatting.

In other words, an "assignment" stage is explicitly left out to allow for implementations that either:
1. Eagerly resolve all declarations immediately, or
2. Lazily only resolve declarations that are required by the message.

If we did include an assignment stage, we would effectively mandate the former even when an implementation could determine that it never needed the value of a declaration.

spec/formatting.md
eemeli: Hmmm. This should probably be reformulated a bit differently, as it's currently in an example's `>` block which ought to mean it's non-normative.

spec/formatting.md
eemeli: The intent here is to establish a dependency of pattern selection on resolution.

spec/formatting.md
eemeli: > Each "literal part" consists of a string.
> 
> An "expression part" can be resolved to a sequence of zero or more "literal parts".

This is not always true. For one example, consider this message:
```
{This is my image: {$img}. Isn't it pretty?}
```

When formatting this, the value of `$img` could be a representation of the image itself, rather than any sequence of literal parts.

Next, consider this message:
```
{This is my image: {flower.png :image}. Isn't it pretty?}
```

Here, the message doesn't include any variables, but it does make use of a custom `:image` function, which could format as a representation of the image that is similarly non-stringifiable.

In other words, we cannot make any assumptions about the shape of external variables or the return values of custom functions.

eemeli:
@cdaringe @aphillips Apologies, I ended up needing to rebase rather than merge to account for today's spec changes; hence the force-push. Will try to avoid those going forward.

macchiati:
BTW, the term 'shape' as used here is not standard English, and probably confusing. I think 'structure' (or something similar) would be much more understandable.

spec/formatting.md
```diff
@@ -1,24 +1,124 @@
...
+  > ```
+  >
+  > The shape of this value is implementation-dependent,
+  > and different implementations may choose to perform different levels of resolution.
+
+- **_Pattern Selection_** determines which of a message's _patterns_ is formatted.
+  For a message with no _selectors_, this is simple as there is only one _pattern_.
+  With _selectors_, this will depend on their _resolution_.
+
+- **_Formatting_** takes the resolved values of the selected _pattern_,
```
catamorphism, line 37: The earlier definition of "resolution" said "**_Resolution_** determines the value of a part of the message", but this definition of "formatting" implies that the output of resolution is one or more values. So for consistency, either the earlier definition should say "values" instead of "value", or this definition should rely on a shared definition of "value" that includes multiple parts.

spec/formatting.md
```diff
@@ -1,24 +1,124 @@
...
+
+- Information on the current **_locale_**,
+  potentially including a fallback chain of locales.
+  This will be passed on to formatting functions.
+
+- Information on the base directionality of the message and its _text_ tokens.
+  This will be used by strategies for bidirectional isolation.
+
+- A mapping of string identifiers to values,
+  defining variable values that may be used during _variable resolution_.
```
catamorphism, line 67: Is it worth saying here that this mapping is for "external variables", as distinct from the variables defined with `let`-declarations?

spec/formatting.md
```diff
@@ -1,24 +1,124 @@
...
+### Literal Resolution
+
+The resolved value of a _text_ or a _literal_ is
+the character sequence of the _text_ or _literal_
+after any character escape has been converted to the escaped character.
+
+When a _literal_ is used as an _operand_
+or on the right-hand side of an _option_,
+the formatting function MUST treat its resolved value the same
+independently of whether the value was originally _quoted_ or _unquoted_.
```
catamorphism, line 114:
```suggestion
whether its value was originally _quoted_ or _unquoted_.
```

spec/formatting.md
```diff
@@ -1,24 +1,124 @@
...
+
+> For example,
+> the _option_ `foo=42` and the _option_ `foo=|42|` are treated as identical.
+
+The resolution of a _text_ or _literal_ token MUST always succeed.
```
catamorphism, line 119: I don't think the term "token" has been defined yet.

spec/formatting.md
```diff
@@ -29,15 +129,114 @@ the local variable takes precedence.
...
 refer to a local variable that's defined after it in the message.
 
+The resolution of a _variable_ MAY fail if no value is identified for its _name_.
+If this happens, an Unresolved Variable error MUST be emitted.
+If a _variable_ would resolve to a _fallback value_,
+this MUST also be considered a failure.
+
+### Function Resolution
+
+To resolve the value of an _expression_ with a _function_ _annotation_,
```
catamorphism, line 139:
```suggestion
To resolve an _expression_ with a _function_ _annotation_,
```
As I understand the earlier definition, values aren't resolved; expressions are resolved to values.

spec/formatting.md
```diff
@@ -29,15 +129,114 @@ the local variable takes precedence.
...
+the following steps are taken:
+
+1. If the _expression_ includes an _operand_, resolve its value.
+   If this fails, use a _fallback value_ for the _expression_.
+2. Based on the _function_ starting sigil and _name_,
+   find the appropriate function implementation from the _function registry_.
+   If this fails, emit an Unknown Function error
```
catamorphism, line 146:
```suggestion
   If the registry does not define an implementation for this name, emit an Unknown Function error
```

spec/formatting.md
```diff
@@ -29,15 +129,114 @@ the local variable takes precedence.
...
+   and use a _fallback value_ for the _expression_.
+3. Resolve the _option_ values as a mapping of string identifiers to values.
```
catamorphism, line 148:
```suggestion
3. Resolve the _option_ values to a mapping of string identifiers to values.
```

spec/formatting.md
```diff
@@ -29,15 +129,114 @@ the local variable takes precedence.
...
+   For each _option_, resolve its right-hand side value.
+   If this succeeds, assign it to the mapping with the _name_ of the _option_.
+   If this fails, do not set any value in the resolved mapping for this _option_.
```
catamorphism, line 151:
```suggestion
   For each _option_:
     * If its right-hand side successfully resolves to a value, bind the _name_ of the _option_ to the resolved value in the mapping.
     * Otherwise, do not bind the _name_ of the _option_ to any value in the mapping.
```

spec/formatting.md
```diff
@@ -29,15 +129,114 @@ the local variable takes precedence.
...
+4. Call the function implementation with the following arguments:
+
+   - The current _locale_.
+   - The resolved value of the _operand_, or if there is no _operand_,
+     an empty value.
```
catamorphism, line 156: Not sure what "an empty value" means concretely.

spec/formatting.md
```diff
@@ -29,15 +129,114 @@ the local variable takes precedence.
...
+   - The resolved mapping of _options_.
+
+   The shapes of the resolved _operand_ and _option_ values is implementation-defined.
```
catamorphism, line 159:
```suggestion
   The shapes of the resolved _operand_ and _option_ values are implementation-defined.
```

spec/formatting.md
```diff
@@ -29,15 +129,114 @@ the local variable takes precedence.
...
+
+   An implementation MAY include additional arguments for the function call,
```
catamorphism, line 161:
```suggestion
   An implementation MAY pass additional arguments to the function,
```

spec/formatting.md
```diff
@@ -29,15 +129,114 @@ the local variable takes precedence.
...
+   If the call fails or does not return a valid value,
+   emit a Resolution error and use a _fallback value_ for the _expression_.
+
+### Fallback Resolution
+
+The _resolution_ of an _expression_ may fail in the following cases:
+
+- A failure in the resolution of a _variable_ _operand_.
+- A failure in the resolution of a _function_ _annotation_.
+- The presence of a _reserved_ _annotation_.
```
catamorphism, line 180:
```suggestion
- A _variable_ _operand_ fails to resolve.
- A _function_ _annotation_ fails to resolve.
- The _expression_ has a _reserved_ _annotation_.
```

spec/formatting.md
```diff
@@ -29,15 +129,114 @@ the local variable takes precedence.
...
+
+In each such case, and error MUST be emitted
```
catamorphism, line 182:
```suggestion
In each such case, an error MUST be emitted
```

spec/formatting.md
```diff
@@ -29,15 +129,114 @@ the local variable takes precedence.
...
+
+  > Examples: `:platform`, `+tag`, `-tag`
+
+- Otherwise: The U+FFFD REPLACEMENT CHARACTER `�` character
+
+_Option_ names and values are not included in the _fallback value_.
+
+When an error occurs in an _expression_ with a _variable_ _operand_
+and the _variable_ refers to a local _declaration_,
+the _fallback value_ is formatted based on the _expression_ of the _declaration_,
```
catamorphism, line 209:
```suggestion
the _fallback value_ is formatted based on the _expression_ on the right-hand side of the _declaration_,
```

spec/formatting.md
```diff
@@ -29,15 +129,114 @@ the local variable takes precedence.
...
+rather than the _expression_ in the _selector_ or _pattern_.
+
+> For example,
+> attempting to format either of the following messages within a context that
+> does not provide for the function `:func` to be successfully resolved:
```
catamorphism, line 214:
```suggestion
> in a context in which the function `:func` fails to resolve,
> attempting to format either of the following messages:
```

spec/formatting.md
```diff
@@ -29,15 +129,114 @@ the local variable takes precedence.
...
+> {The value is {$var}.}
+> ```
+>
+> ```
+> let $var = {|horse|}
+> {The value is {$var :func}.}
+> ```
+>
+> would in both cases result in the _pattern_ _expression_
+> resolving to a _fallback value_ of `|horse|`.
```
catamorphism, line 227:
```suggestion
> yields a _fallback value_ of `|horse|`.
```

spec/formatting.md
```diff
@@ -29,15 +129,114 @@ the local variable takes precedence.
...
+
+_Pattern selection_ is not supported for _fallback values_.
```
catamorphism, line 229:
```suggestion
_Pattern selection_ is not defined on _fallback values_.
```

spec/formatting.md
```diff
@@ -29,15 +129,114 @@ the local variable takes precedence.
...
+
 ## Pattern Selection
 
 When a _message_ contains a _match_ statement with one or more _expressions_,
```
catamorphism, line 233:
```suggestion
When a _message_ contains a _match_ construct with one or more _expressions_,
```

spec/formatting.md
```diff
@@ -284,9 +488,48 @@ when *   {Other match}
 
 4. The pattern `{Exact match}` of the most preferred `1` variant is selected.
 
-## Handling Bidirectional Text
+## Formatting
+
+After _pattern selection_,
+each _text_ and _expression_ of the selected _pattern_ must be resolved and formatted.
```
catamorphism, line 494:
```suggestion
each _text_ and _expression_ part of the selected _pattern_ must be resolved and formatted.
```

(Not sure if we have a standard name for the parts/elements/components of a pattern yet.)

spec/formatting.md
```diff
@@ -284,9 +488,48 @@ when *   {Other match}
...
+
+_Formatting_ is a mostly implementation-defined process,
+as it depends on the implementation's shape for resolved values
+and the target shape of the formatting.
+
+Formatting errors MAY be emitted during _formatting_,
+as the successful _resolution_ of an _expression_ does not guarantee success in its _formatting_.
```
catamorphism, line 501:
```suggestion
as formatting is not defined on every resolved value.
```

spec/formatting.md
```diff
@@ -284,9 +488,48 @@ when *   {Other match}
...
+_Formatting_ MAY produce formatted messages in the following shapes,
+as well as any others:
+
+- A single concatenated string.
+- A flat sequence of objects corresponding to each resolved value.
+- A hierarchical structure of objects that group spans of resolved values,
+  such as sequences delimited by "open" and "close" _function_ _annotations_.
+
+Implementations SHOULD provide _formatting_ output shapes that match user needs,
+such that e.g. re-parsing a message formatted as a string is not required.
```
catamorphism, line 513: Not sure I understand this -- should it be "re-parsing an output shape that is a single formatted string"?

spec/formatting.md
```diff
@@ -409,6 +653,20 @@ These are divided into the following categories:
     > when * 1 {Second is one}
     > ```
 
+  - **Duplicate Option Name errors** occur when the same _name_ is used
+    for more than one _option_ in the same _expression_.
```
catamorphism, line 657:
```suggestion
  - **Duplicate Option Name errors** occur when the same _name_ 
    appears on the left-hand side
    of more than one _option_ in the same _expression_.
```

spec/formatting.md
```diff
@@ -497,96 +755,29 @@ These are divided into the following categories:
...
+Syntax and Data Model errors MUST be emitted as soon as possible.
+
+During selection, an _expression_ handler MUST only emit Resolution and Selection errors.
+During formatting, an _expression_ handler MUST only emit Resolution and Formatting errors.
 
-During selection, an _expression_ handler must only emit Resolution and Selection errors.
-During formatting, an _expression_ handler must only emit Resolution and Formatting errors.
+Resolution and Formatting errors in _expressions_ that are not used
+in _pattern selection_ or _formatting_ MAY be ignored
+as such do not impact the current message's formatting.
```
catamorphism, line 765:
```suggestion
in _pattern selection_ or _formatting_ MAY be ignored,
as they do not affect the output of the formatter.
```

spec/formatting.md
```diff
@@ -497,96 +755,29 @@ These are divided into the following categories:
...
-should prioritise Syntax and Data Model errors over others.
+SHOULD prioritise Syntax and Data Model errors over others.
 
-When an error occurs in the resolution of an _expression_,
-the _expression_ in question is processed as if the option were not defined.
-This may allow for the fallback handling described below to be avoided,
-though an error must still be emitted.
+When an error occurs in the _resolution_ of an _option_,
+the surrounding _expression_ MUST be processed as if the _option_ were not present.
+This MAY allow for the _expression_ _resolution_ not resulting in a _fallback value_,
```
catamorphism, line 778:
```suggestion
This MAY allow the _expression_ to resolve to a non-fallback _value_,
```

spec/formatting.md
```diff
@@ -497,96 +755,29 @@ These are divided into the following categories:
...
+though an error MUST still be emitted.
 
 When an error occurs within a _selector_,
-the _selector_ must not match any _variant_ _key_ other than the catch-all `*`
+the _selector_ MUST NOT match any _variant_ _key_ other than the catch-all `*`
 and a Resolution or Selector error is emitted.
```
catamorphism, line 783: Should this be "MUST BE emitted"?

spec/formatting.md
eemeli: The plural here is intended to refer to the distinct _text_ and _expression_ parts of a _pattern_. Each such part of _pattern_ would still resolve to one value each.

spec/formatting.md
eemeli: I'd prefer not adding to that here atm, as variable resolution is still being iterated upon.

spec/formatting.md
eemeli: It's from here: https://github.com/unicode-org/message-format-wg/blob/main/spec/syntax.md#tokens

spec/formatting.md
eemeli: It would be something like `undefined`, `null`, or `void`, depending on the programming language.

spec/formatting.md
eemeli: The current phrasing is intended to refer only to the result of the resolution, rather than commenting on the result of the formatting. With "yields", this distinction is lost.

spec/formatting.md
eemeli: The current wording matched that in "Resolve Selectors", where step 2.ii. asks "If selection is supported for `rv`". If you think this should be changed, then that method line should also be changed correspondingly.

spec/formatting.md
eemeli: We don't; that's another possible addition to the syntax like `operand` that we might want to make for the sake of clarity.

spec/formatting.md
eemeli: I think so? As in, there should be no need for someone to format MF2 to a string, and then right away try to parse that string.

spec/formatting.md
eemeli: Yes.
```suggestion
and a Resolution or Selector error MUST be emitted.
```

eemeli:
Thank you @cdaringe for a thorough review! I'm marking this as "Ready for review", as it, well, clearly is.

spec/formatting.md
catamorphism: In that case, maybe it's worth adding something after line 15 saying that the result of resolving a pattern is a list of values that results from independently resolving each of its parts? If the definition of formatting a pattern requires a list as input, then maybe it's worth saying that the output of resolving a pattern is a list, even if the shape of a "value" is being left abstract.

spec/formatting.md
catamorphism: Can that be cited in this document?

spec/formatting.md
catamorphism: In that case, maybe it's worth adding something after line 15 saying that the result of resolving a pattern is a list of values that results from independently resolving each of its parts? If the definition of formatting a pattern requires a list as input, then maybe it's worth saying that the output of resolving a pattern is a list, even if the shape of a "value" is being left abstract.

spec/formatting.md
catamorphism: There are programming languages that don't have any of those, so perhaps change to something like "a distinct argument that denotes the absence of a value, which depends on the underlying implementation and programming language"?

spec/formatting.md
catamorphism: I'll withdraw this comment. As long as it's consistent, "supported for" is fine.

spec/formatting.md
catamorphism: I understand now, but I still find the text as it is a bit hard to follow. Maybe it would be easier to read split into two sentences, like:

```
Implementations SHOULD provide _formatting_ output shapes that match user needs.
For example, a user should never need to immediately re-parse the string output by the formatter.
Instead, the formatter itself should directly produce a more structured representation.
```

That's still a bit verbose, but the point to get across is that `string` shouldn't serve as an intermediate data structure, I think.

catamorphism (commented):
Looks good. The additional comments I made are just suggesting a few minor wording changes for clarity. The only one that needs to be added, in my opinion, is citing the definition of "token". Everything else is up to your discretion.

spec/formatting.md
eemeli: Your comments lead me to realise that really practically all of the "resolution" section is about "expression resolution". So I'm retitling accordingly, hopefully adding some clarity here as well.

spec/formatting.md
eemeli: I believe it effectively already is, as we've established in `spec/README.md` that e.g. the italic "_text_" is a reference to the bold-italic "**_text_**" definition that's a subsection of the "Tokens" section of `syntax.md`.

I don't know how exactly we'll do the final rendering of this, but I would presume that we would at that stage linkify these terms accordingly.

spec/formatting.md
eemeli: Fair point. Rephrasing to avoid the term and make the operand an optional argument.

spec/formatting.md
eemeli: Would this be clearer?
```suggestion
including situations that require further processing of formatted messages.
```

eemeli:
> BTW, the term 'shape' as used here is not standard English, and probably confusing. I think 'structure' (or something similar) would be much more understandable.

@macchiati I'm happy to iterate on the language. As it's used here, "shape" tries to be sufficiently generic to allow for implementations to use values that are e.g. just strings, objects with fields, or instances of classes with methods. At least my sense of "structure" strongly implies something in the middle of that spectrum, while potentially leaving out entirely non-object values.

Is there a different word that we could use with this sort of meaning, or is my understanding of "structure" somehow skewed?

spec/formatting.md
aphillips: Message format can have parts that are objects.

Literal parts of a message are *always* strings by definition.

Expressions can, as you note, represent objects and these might not be immediately stringable. But the existence of string resolution (note your text does this!!) means that all expressions can ultimately be represented as a sequence of literals. It is tempting to want to make it a list of literals *or* expressions. But ultimately what my text says is that you **can** resolve the deepest nesting of an expression to a string.

The string might be `` or `img/png:base64gooHere` or something. And users do not have to force the object to become a string (they can peek at the expression "type" and get the image, for example).

I'm open to a lot of change here: my proposal above is basically the back of my cocktail napkin while thinking about message resolution. Allowing "shapes" other than string is fine, but having our standard require that one be able to produce a character sequence means that everything can ultimately call `toString`.

To @macchiati's point, "shape" is kind of a vague word. Perhaps:

```suggestion
- **_Formatting_** takes the resolved values of the selected _pattern_,
  and returns the formatted result for the _message_.
  This specification defines formatting of each _message_ as a _string_.
 Implementations MAY return a _message_ using a different, locally appropriate, 
 data type (such as an attributed string) or as a logical sequence of 
 values as appropriate for that implementation.
 
 > For example, an implementation might choose to return an interstitial
 > object so that the caller can "decorate" portions of the formatted value.
 > The `NumberFormatter` class in ICU4J, for example, returns a `FormattedNumber`
 > object, so a _pattern_ such as `{This is my number {42 :number}}` might return
 > the character sequence `This is my number ` followed by a `FormattedNumber`
 > object representing the value `42` in the current locale.
```

macchiati:
The term _shape of a value_ is completely opaque IMO; the only thing it suggests to me is a visual shape, like the visual outline of that value as rendered.

I would have no idea at all that you meant something like the underlying structure. Addison's suggestion of the term 'result' is better in that it doesn't suggest a completely wrong interpretation, and is broad enough to encompass a wide variety of possibilities.

spec/formatting.md
eemeli: I've taken some of the suggested language and incorporated it into the Formatting section, rather than this Introduction. Also moved out & refactored the resolution example slightly.

spec/formatting.md
eemeli: Applied the above change.

eemeli:
@macchiati Point taken. I've dropped the term "shape" (and also "target" while at it) and now refer to the "result" or "result type" when referring to what's produced by formatting.

---

@aphillips Responding to your comments on [this line](https://github.com/unicode-org/message-format-wg/pull/396#discussion_r1234466707) here to preserve them when that thread gets resolved:
> Expressions can, as you note, represent objects and these might not be immediately stringable. But the existence of string resolution (note your text does this!!) means that all expressions can ultimately be represented as a sequence of literals. It is tempting to want to make it a list of literals _or_ expressions. But ultimately what my text says is that you **can** resolve the deepest nesting of an expression to a string.
> 
> The string might be `` or `img/png:base64gooHere` or something. And users do not have to force the object to become a string (they can peek at the expression "type" and get the image, for example).

This only _mostly_ works. It fails e.g. when the identity of an object (such as an image) matters, or when the string representation of a function used to define a value refers to variables that were in scope during its definition, but are not available afterwards.

Or are we perhaps talking about different things? To me, the key thing here is that it's not possible to construct all possible non-string formatting results if an intermediate result forces all of the values to be strings.

> I'm open to a lot of change here: my proposal above is basically the back of my cocktail napkin while thinking about message resolution. Allowing "shapes" other than string is fine, but having our standard require that one be able to produce a character sequence means that everything can ultimately call `toString`.

That's possible, yes, but not always useful. The string could very well end up something like `'[object Object]'` which has almost no utility.

aphillips:
@eemeli noted:

> Or are we perhaps talking about different things? To me, the key thing here is that it's not possible to construct all possible non-string formatting results if an intermediate result forces all of the values to be strings.

I think this is the disconnect: we agree that stringification is a terminal result. My suggestions carefully do not require the interstitial results to be a string. Since our specification defines a terminal string form, we need to specify what an implementation does in these cases (which shouldn't be **_too_** specific and probably should be very permissive, i.e. it's whatever the function or expression wants it to be)

Ultimately, though, my meta-point is: we should not defer "formatToParts" down the road much further. We should deal with it here to ensure that implementations can expose non-string resolution of parts, including *nested* sequences. Your original reaction was to my saying:

> An "expression part" can be resolved to a sequence of zero or more "literal parts".

Notice that this allows the string resolution for an expression to be empty. And it requires that an "expression part" be ultimately resolvable to a literal. What it doesn't say (it probably should) is that an "expression part" doesn't have to directly resolve to a literal. 

I think your reaction is that you read this text to mean that the literal parts are *always* resolved to a literal:

> The string output of a message is the concatenated sequence of resolved literal parts.

We can and should add the necessary support for non-string "expression parts". But your proposed text and the back of my napkin are both dealing with the string resolution bit. Would it help if the above said:

> The string output of a message is the concatenated sequence of all parts once they have been resolved to a literal. 
> _Expression parts_ SHOULD NOT be resolved to a literal until required to do so by the caller (e.g. in a `toString` function or method) or because that is the preferred output by the expression's implementer (as in the `datetime` example in this section)

Would this representation in my fake JSON make sense:

```json
{
   "locale": "ar-AE",
   "direction": "ltr",
   "parts": [
       {
           "type": "literal",
           "locale": "ar-AE",
           "direction": "ltr",
           "value": "Your image is "
       },
       {
           "type": "expression",
           "locale": "ar-AE",
           "direction": "rtl",
           "value": [
                  { "type": "image", "locale":"ar-AE", "dir": "rtl", "name": "image", "src": "image.jpg" }
           ]
       },
      {
         "type": "literal",
         "locale": "ar-AE",
         "direction": "ltr",
         "value": " Isn't it pretty?"
      }
   ]
}
```

spec/formatting.md
```diff
@@ -1,24 +1,124 @@
...
+for the message's _expression resolution_, _pattern selection_ and _formatting_.
+
+At a minimum, it includes:
+
+- Information on the current **_locale_**,
+  potentially including a fallback chain of locales.
+  This will be passed on to formatting functions.
+
+- Information on the base directionality of the message and its _text_ tokens.
+  This will be used by strategies for bidirectional isolation.
```
aphillips, line 52: Not just! It should also be exposed as a property of the message so that the caller can properly set the base paragraph direction. Or is this what you mean by "strategies for bidirectional isolation"? If the latter, I'd suggest:

```suggestion
This can be used by callers to set the base direction of the _message_ upon display.
```

eemeli: Included in #411.

PR #397, opened 2023-06-19
Change the syntax of the ::open and :/close function calls

stasm:
This is a follow-up to #364, which made it possible to use unquoted literals in the argument position in placeholders. However, due to the current syntax of +open and -close function calls, arguments that are number literals must still be quoted, e.g. `{|-1| :number}`.

This PR proposes to change the syntax of markup-like function calls:

    BEFORE: {+button title=|Click me!|}Submit{-button}
    AFTER:  {::button title=|Click me!|}Submit{:/button}

The benefit of using a two-char-long prefix is that we effectively establish the colon `:` as the general-purpose function introducer.

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -483,14 +483,14 @@ Between the brackets, the following contents are used:
   > Example: `{:platform}`
 
 - Opening _expression_ with no operand:
-  U+002B PLUS SIGN `+` followed by the _function_ _name_
+  U+003A COLON `:` followed by U+003A COLON `:` and then by the _function_ _name_
 
-  > Example: `{+tag}`
+  > Example: `{::tag}`
 
 - Closing _expression_ with no operand:
-  U+002D HYPHEN-MINUS `-` followed by the _function_ _name_
+  U+003A COLON `:` followed by U+002F SOLIDUS `/` and then by the _function_ _name_
 
-  > Example: `{-tag}`
+  > Example: `{:/tag}`
 
 - Otherwise: The U+FFFD REPLACEMENT CHARACTER `�` character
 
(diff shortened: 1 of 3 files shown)


eemeli:
> due to the current syntax of +open and -close function calls, arguments that are number literals must still be quoted, e.g. `{|-1| :number}`.

To qualify this a bit, only literals that represent negative numbers must be quoted with our current syntax; for example
```
{1 :number}
```
is completely fine.

I also continue to be interested in having anyone provide examples of existing real-world messages in any format that include negative number literals in any form. I contend that this is an exceedingly rare usage pattern that should not drive our syntax deliberations.

For one real-world data point, much of the Firefox code base is formatted with Fluent, which also supports markup as well as numerical literals as formatting function arguments and option values. It currently uses markup in a little under 5% of all messages, and does not include a single negative number literal.

aphillips:
@eemeli I don't think that percentages of messages is that useful a measure. The majority of strings have no replacements. The majority of strings with placeables have only one. Etc. What does concern me in this case is the "accessibility" or "usability" of the features we build: how surprising or unsuprising are the features? How consistent are they?

I think for negative numbers, there are domains where they are more prevalent and domains where they are less so. I am unsurprized that FF's code base has few of them. Other domains may find they are more prevalent and might include:

* finance (gain/loss, performance)
* weather (low temperatures, changes)
* sports (over/under)

When a user needs to write a message with a negative literal value, they shouldn't be surprised to find that it is a syntax error.

```
match {$temperature :lt}
when -20 {Is it cold?}
when -10 {Significant risk of chapped lips}
when 0 {Freezing}
when 10 {Bring a coat}
when 20 {Not bad}
when * {Is it hot in here?}
```

stasm:
I'm going to close this for now, while I prepare a document about a more holistic sigil design.

PR #399, opened 2023-06-19
Add negative-start rule

eemeli:


spec/message.abnf
```diff
@@ -36,11 +36,11 @@ quoted-char = %x0-5B         ; omit \
             / %x7D-D7FF      ; omit surrogates
             / %xE000-10FFFF
 
-; based on https://www.w3.org/TR/xml/#NT-Nmtoken,
-; but cannot start with U+002D HYPHEN-MINUS or U+003A COLON ":"
-unquoted = unquoted-start *name-char
+; based on https://www.w3.org/TR/xml/#NT-Nmtoken
```
aphillips, line 39: this is increasingly less true. Perhaps "similar to https://...."?

spec/syntax.md
```diff
@@ -463,8 +463,9 @@ The characters `\` and `|` MUST be escaped as `\\` and `\|`.
 
 **_Unquoted_** literals have a much more restricted range that
 is intentionally close to the XML's [Nmtoken](https://www.w3.org/TR/xml/#NT-Nmtoken),
-with the restriction that it MUST NOT start with `-` or `:`,
-as those would conflict with _function_ start characters.
+with the restrictions that it MUST NOT start with `:`,
+and when starting with `-`, must be followed by a `.` or a digit.
+These restrictions are required not to conflict with _function_ start characters.
```
aphillips, line 468: This is now a run-on sentence. Perhaps:

```suggestion
is intentionally close to XML's [Nmtoken](https://www.w3.org/TR/xml/#NT-Nmtoken).
To make _unquoted_ literals distinct from _function_ names, a literal 
MUST be quoted if it begins with a `:` or if it begins with a `-` that is 
**not** followed by a `.` or a digit. Literals beginning with a `-` followed
by a `.` or a digit might be used to pass a negative number value to
a _function_ or as a _key_ for a selector.
```

@eemeli An example I didn't think of calling out previously is that unquoted literals are also used on the right hand side of an `option`. I think there are **_lots_** of places where a negative number would be passed there--that would be the most common use:

```
 {$foo :function option=-42}
```

spec/syntax.md
```diff
@@ -477,9 +478,10 @@ quoted-char    = %x0-5B         ; omit \
                / %x7D-D7FF      ; omit surrogates
                / %xE000-10FFFF
 
-unquoted       = unquoted-start *name-char
+unquoted       = (unquoted-start / negative-start) *name-char
 unquoted-start = name-start / DIGIT / "."
                / %xB7 / %x300-36F / %x203F-2040
+negative-start = "-" ( DIGIT / "." )
```
aphillips, line 484: Could this be combined?

```abnf
unquoted-start = name-start
               / DIGIT
               / "."
               / %B7 
               / %300-36F 
               / %203F-2040
               / ("-" ( DIGIT / "." ))
```

Also, remind me why we added the combining mark range of U+0300-036F to the quoted literal start? Wouldn't we be better off (given that unquoted literals are a **convenience**) in not adding them? Or U+203F and U+2040?

aphillips (changes requested):

spec/message.abnf
eemeli:
```suggestion
; similar to https://www.w3.org/TR/xml/#NT-Nmtoken
```

spec/syntax.md
eemeli: > Could this be combined?

Yes, they could be combined. I separated them because you mention on Slack earlier that:
> This change has the unhappy aspect of making `unquoted-start` sometimes consume multiple characters. 

> Also, remind me why we added the combining mark range of U+0300-036F to the quoted literal start? Wouldn't we be better off (given that unquoted literals are a **convenience**) in not adding them? Or U+203F and U+2040?

It's because we've been trying to stay as close as possible to the [XML Nmtoken](https://www.w3.org/TR/xml/#NT-Nmtoken), which includes these characters. In other words, we currently have:

```abnf
```

and at least technically we could "simplify" these so that we had

```abnf
name-char  = unquoted-start / "-" / ":"
```

But that seems like it would ultimately raise more questions.

spec/syntax.md
eemeli: I edited your comment to get the line breaks right, but then still needed to apply the commit locally to get it to stick. I left out the last sentence, as it didn't seem necessary and was a bit confusing in listing only some of the places where a negative literal could hereby be used.

spec/syntax.md
aphillips: I see... it requires looking back at `Nmtoken` and its definition in XML. That makes it clear.

I did have a comment on Slack about not having the weird two-character side effect, but then today we discussed in teleconference that we were "okay" as a group with that. With that in mind, I thought maybe the single production would be clearer.

eemeli: I'm fine either way; the end result is the same.

gibson042: > It's because we've been trying to stay as close as possible to the [XML Nmtoken](https://www.w3.org/TR/xml/#NT-Nmtoken), which includes these characters.

I've asked before, and will repeat the question: what is the value of staying _close_ to XML Nmtoken but not matching it exactly? If there's going to be deviation (and unexplained deviation at that), what makes that a better foundation than e.g. [UAX 31 Identifiers](https://www.unicode.org/reports/tr31/#Default_Identifier_Syntax), or for that matter of dropping the pretense that this syntax has any relevant relationship to another one?

aphillips: @gibson042 The main reason we were sticking close to Nmtoken is for compatibility with names found in CLDR, although I tend to agree with you that "sorta consistent with" is not a really useful feature (and is probably a tripping hazard): we want to ensure that `name` values found in CLDR are valid without modification. The `unquoted` production is recycling the production (but I think not for any technical reason--see below)

@eemeli My "simplification" was meant to make top-level productions (what people are trying to understand) as logical as possible. I think it is cleaner to define all of the name-like constructions like:

```abnf
foo       = foo-start *foo-char
foo-start = ...
foo-char  = ...
```

Or, if recycling `name-char`:

```abnf
foo       = foo-start *name-char
foo-start = ...
bar       = bar-start *name-char
bar-start = ...
...
name-char = ...
```

I am going against my slack comment in suggesting we allow consuming two characters because it seems a simpler compromise (that we can call out with a comment) than having alternate `-start` productions.

To @gibson042's point, there is no real reason why we need to use `name-char` in `unquoted` except that we already know it doesn't have any whitespace or other troublesome characters in it and implementers probably already have some code to validate that set. Is that enough of a reason?

peter-b: We really do not need a N+1th incompatible specification for identifiers. It would be greatly preferable to either:

- use an identifier grammar production that exactly matches some existing standard
- use UAX 31 

Any other choice requires an extraordinarily strong justification, which has not been provided here.

eemeli: I've combined the rules as suggested in the first comment here, so closing this thread. Will reply separately about the identifier basis.

eemeli:
From @gibson042:
> I've asked before, and will repeat the question: what is the value of staying _close_ to XML Nmtoken but not matching it exactly? If there's going to be deviation (and unexplained deviation at that), what makes that a better foundation than e.g. [UAX 31 Identifiers](https://www.unicode.org/reports/tr31/#Default_Identifier_Syntax), or for that matter of dropping the pretense that this syntax has any relevant relationship to another one?

UAX 31 identifiers support a much more limited set of strings than we want to support as unquoted values; for example values such as `42`, `13.0` and `html:b` are currently valid unquoted literals, but are not valid UAX 31 identifiers. XML Nmtoken is _really_ close to being exactly what we want, with two specific starting-character exclusions we've needed to carve out.

One key differentiator here is that for `unquoted` we're not trying to match an "identifier" grammar, but a "token" grammar. As far as I'm aware, no real regularity exists on this, and the XML Nmtoken seems like a pretty good starting point.

---

From @peter-b:
> We really do not need a N+1th incompatible specification for identifiers. It would be greatly preferable to either:
> * use an identifier grammar production that exactly matches some existing standard
> * use UAX 31
> 
> Any other choice requires an extraordinarily strong justification, which has not been provided here.

Please correct me if I'm mistaken, but this seems like a point that's more about `name` rather than `unquoted`, yes? If this is the case, replies to the following should be spun off into their own issue rather than being tracked under this PR:

For `name`, using UAX 31 identifiers would be conceivable as a starting point, provided that the grammar also allowed for a path-splitting character such as `.`, `-`, or `:` to support constructs like `html:b` or `foo.bar`. This is not an uncontroversial topic, and when settling on a syntax the WG opted to allow each implementation to figure out for itself e.g. whether the `.` in `foo.bar` has any special meaning, rather than making the call at the syntax level. Basing `name` on XML Name supports this, while other choices would require this discussion to be reopened.

gibson042:
> XML Nmtoken is _really_ close to being exactly what we want, with two specific starting-character exclusions we've needed to carve out.

"_Needed_" seems like an overly strong characterization. What problems would ensue from updating the syntax to match XML Nmtoken exactly (keeping in mind that it is possible for source to be syntactically valid while still being rejected for higher-level semantic reasons)? And if the deviation is **actually** necessary, shouldn't the reason be documented in a grammar comment, either directly or with a referencing link?

eemeli:
> What problems would ensue from updating the syntax to match XML Nmtoken exactly (keeping in mind that it is possible for source to be syntactically valid while still being rejected for higher-level semantic reasons)?

XML Nmtoken may start with either the `:` or `-` characters, which are function start indicators in MF2. This causes an overlap at the beginning of an expression, which _may_ have an operand. Effectively, matching the rule exactly for `unquoted` would require us to change the function start characters, or disallow `unquoted` as an operand; only allowing it as an option value.

> And if the deviation is **actually** necessary, shouldn't the reason be documented in a grammar comment, either directly or with a referencing link?

This clarification is provided in the `syntax.md`, to which this PR adds this text:

```markdown
To make _unquoted_ literals distinct from _function_ names,
a literal MUST be quoted if it begins with a `:`
or if it begins with a `-` that is **not** followed by a `.` or a digit.
```

At least my understanding is that our syntax is defined by both the `message.abnf` as well as the `syntax.md`, and that duplication of the latter into the former would not necessarily be beneficial -- if you're looking for explanations, then `syntax.md` already provides that, along with the ABNF rules.

spec/syntax.md
```diff
@@ -462,9 +462,10 @@ except for surrogate code points U+D800 through U+DFFF.
 The characters `\` and `|` MUST be escaped as `\\` and `\|`.
 
 **_Unquoted_** literals have a much more restricted range that
```
aphillips, line 464: I think this is missing some information that could help avoid repeats of the conversation in this PR.

```suggestion
A _literal_ MAY appear **_unquoted_** when its content matches
the `unquoted` production. 

Similar to other productions in the message syntax,
the content of an _unquoted_ literal is identical to XML's [Nmtoken]
(https://www.w3.org/TR/xml/#NT-Nmtoken) except that specific characters
meaningful to other portions of the syntax are not allowed at the start. Specifically:

* A _literal_ MUST be quoted if it begins with a `:` (U+003A COLON).
* An _unquoted_ literal MUST NOT begin with a `-`  (U+002D HYPHEN MINUS)
unless that character is followed by a `.` (U+002E FULL STOP) or an ASCII digit.
This allows literals which might be used to represent negative numbers, such 
as `-42` or `-.345`, to appear unquoted.
```

spec/message.abnf
```diff
@@ -36,11 +36,11 @@ quoted-char = %x0-5B         ; omit \
             / %x7D-D7FF      ; omit surrogates
             / %xE000-10FFFF
 
-; based on https://www.w3.org/TR/xml/#NT-Nmtoken,
-; but cannot start with U+002D HYPHEN-MINUS or U+003A COLON ":"
+; similar to https://www.w3.org/TR/xml/#NT-Nmtoken
```
aphillips, line 39: I think this is distracting. The similarity to (and differences from) Nmtoken is noted in the spec. It doesn't really add anything to the understanding of the ABNF. I think I'd rather comment on what `unquoted` is:

```suggestion
; literals that do not require quoting: 
; containing no whitespace and with a restricted range of characters
```

aphillips (approved):
This addresses my previous comments, but this PR's discussion thread makes me think we should make some changes to the text to clarify what we're doing.

spec/syntax.md
eemeli: I'm fine with being more verbose about the description here, but the above suggestion is a bit problematic:
- The first paragraph is somewhat tautological, given that this is how practically all of the syntax is defined.
- I don't think we have any other productions that refer to or approximate XML Nmtoken. Our _name_ is rather close to XML Name, but its first-character rules are quire different.

gibson042:
> XML Nmtoken may start with either the `:` or `-` characters, which are function start indicators in MF2. This causes an overlap at the beginning of an expression, which _may_ have an operand. Effectively, matching the rule exactly for `unquoted` would require us to change the function start characters, or disallow `unquoted` as an operand; only allowing it as an option value.

If XML Nmtoken is sufficiently important to use as a foundation for unquoted literals, then having overlap between it and function prefix sigils seems like an unnecessary misstep. Without dragging this too far into the weeds, I will note that most ASCII punctuation characters are excluded from NameChar and therefore available for that use in MF2, particularly all of those that are currently included in ``` reserved-start = "!" / "@" / "#" / "%" / "^" / "&" / "*" / "<" / ">" / "?" / "~" ```. Accordingly, of the three PRs relating to unquoted negative number operands, I think `{#open}...{/close}` of #398 heads in the best direction (although it would still need to replace `{:functionName}` with something like `{!functionName}` so that valid Nmtoken instances like ":functionName" can be used as unquoted operands.

aphillips:
@gibson042 That's a good point. If we want to claim that our naming is rooted in XML, why not just go all the way and _use_ `Name` and `Nmtoken` and choose sigils (there are plenty available) that don't conflict with these? The main thing is that we'd need to change our (arbitrary) choice of `:` too. It would greatly simplify the logic _behind_ our naming and perhaps our syntax.

macchiati:
The important issue is that our syntax be even as natural and non-error prone as possible. I don't see that there is much if any value to adhering strictly to Nmtoken if that adherence compromises the syntax.

gibson042:
Sure, but are you claiming that adopting Nmtoken _would_ compromise the syntax? Is `{:foo}` really more natural than `{!foo}`/`{@foo}`/`{#foo}`/etc.? Is it really less error prone to accept `{-1 :bar}` while rejecting `{+1 :bar}`?

aphillips:
@macchiati I agree. However, if we can achieve full compatibility with our chosen namespace scheme by choosing a different (somewhat randomly chosen) sigil, doesn't that (slightly) reduce potential errors? The syntax is only required to be internally consistent, but its better if it is also widely compatible with the contexts in which it will appear and if exterior software requires only minimal modification to interact with our constructs.

Of course, it isn't as if the world is constructed around `Nmtoken`. It's nice that our naming matches _something_, but that something is not exactly like most programming language namespaces and it _will_ conflict with naming conventions established in some other resource, templating, scripting, or programming contexts somewhere. As a reminder, the reason we chose `Nmtoken` (and `Name`) was to maximize compatibility with (potential) LDML constructs, since CLDR uses XML. The "carve out" for various sigils doesn't conflict with naming in LDML currently and future conflicts are under our control. If we don't care about other XML dialects, non-conformance with `Nmtoken` (et al) may ultimately not matter. Most users will name their variables and functions something rational (probably an ASCII alphanumeric token).

stasm:
I continue to believe that this PR treats the symptom rather than the root cause. I think there are two blockers before we can consider this PR for merging:

* Finish the design of the open/close spannables in #470, including their syntax.
* Document the use-cases for unquoted literals in #478.

eemeli:
Just to note, on the call of 4 September we agreed to merge this on or immediately after the end of the Seville colloquium on 13 September, unless we explicitly decide against doing so. No such decision has been made.

I agree that this is treating the symptom rather than the root cause, but that something very much like this is required for negative number support with our current syntax. We should patch this now, while acknowledging that it may be wholly superseded by later changes.

stasm:
That agreement was based on the assumption that we would have completed the design of open/close in Seville. That didn't happen and the design is still ongoing in #470.

I want to this PR to stay open because I want this to be a friction point, so that we don't get complacent with the current proposed design on open/close spannables.

An alternative which I'd be happy to go with instead is to remove the current open/close features from the spec on the `main` branch, since it's pending a refactor in #470 anyways. We can then fix the negative literals the proper way, and *then* discuss if we want to compromise in order to enable the `+` and `-` syntax for spannables.

aphillips:
Held for discussion in 2023-11-13 call

aphillips:
I believe this is now obsolete and should be closed in favor of #553

PR #400, opened 2023-06-19
docs: add roundtrip to glossary

cdaringe:
# Problem

["roundtrip" is used liberally](https://github.com/search?q=repo%3Aunicode-org%2Fmessage-format-wg+roundtrip&type=code) in captured documentation, but has no captured definition in the glossary

# Solution

Define the term

### DIFF
--- a/docs/glossary-and-resources.md
+++ b/docs/glossary-and-resources.md
@@ -35,6 +35,7 @@
 * **locale fallback** - offering a reasonable substitute locale when the requested locale's resources are not available. Results may vary depending on context (ex: audio vs. text vs. video).
 * **locale matching** - computing the locale fallback.
 * **resource** - files bundled with an application that are loaded in by the executable code.  UI strings, etc. and their locale-specific translations are typically stored as resources.
+* **roundtrip (RTT)** - the process of translating text into another language, then translating the result back into the original language.
 * **translation merging** - in l10n TMSes, the document-level interpolation of translated content. In other words, the replacing of translatable units in the source document with their equivalent translated units.  See *interpolate*.
 * **selector** - see *placeholder type*.
 * **specification** - the rules we decide that describe what is passed to the API for message formatting (structure of data, syntax, etc.).
--- a/docs/glossary-and-resources.md
+++ b/docs/glossary-and-resources.md
@@ -35,7 +35,7 @@
 * **locale fallback** - offering a reasonable substitute locale when the requested locale's resources are not available. Results may vary depending on context (ex: audio vs. text vs. video).
 * **locale matching** - computing the locale fallback.
 * **resource** - files bundled with an application that are loaded in by the executable code.  UI strings, etc. and their locale-specific translations are typically stored as resources.
-* **roundtrip (RTT)** - the process of translating text into another language, then translating the result back into the original language.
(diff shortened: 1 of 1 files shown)


docs/glossary-and-resources.md
```diff
@@ -35,6 +35,7 @@
 * **locale fallback** - offering a reasonable substitute locale when the requested locale's resources are not available. Results may vary depending on context (ex: audio vs. text vs. video).
 * **locale matching** - computing the locale fallback.
 * **resource** - files bundled with an application that are loaded in by the executable code.  UI strings, etc. and their locale-specific translations are typically stored as resources.
+* **roundtrip (RTT)** - the process of translating text into another language, then translating the result back into the original language.
```
eemeli, line 38: This seems like an excellent way to grow our glossary, esp. as it seems that the meaning of this term is not immediately obvious. Here's my understanding:

```suggestion
* **roundtrip (RTT)** - the process of transforming a message into another format or represenation, then transforming it back into the original format.
```

eemeli (changes requested):

docs/glossary-and-resources.md
cdaringe: i more or less ripped off https://en.wikipedia.org/wiki/Round-trip_translation

shucks, is their definition too narrow? or is ours too wide? i sincerely do not know 😄

eemeli: I think it's more that we're working in a rather specialized context. The original and most common use of this term in our context is with respect to messages in ICU MessageFormat (aka MF1) syntax, and their representation in an MF2 data model.

Because MF1 can include inline selectors and MF2 can't, an MF1-to-MF2 transformer needs to duplicate any leading and trailing content around an MF1 selector to each of its variants. If that message is then transformed back into MF1, information is lost about its exact original syntax, but the resulting MF1 message will still format exactly the same as before.

Early on, an [explicit choice](https://github.com/unicode-org/message-format-wg/blob/main/docs/consensus_decisions.md#4-only-allow-for-selectors-at-the-top-level-of-a-message) was made to be ok with this.

cdaringe: if you think others would be fine _overloading_ the definition just for this project, fine by me :) i generally would try to avoid overloading the term if its meaning is well known to the target audience... but perhaps it's close enough to not make a fuss!

eemeli (approved):
> if you think others would be fine _overloading_ the definition just for this project, fine by me :) i generally would try to avoid overloading the term if its meaning is well known to the target audience... but perhaps it's close enough to not make a fuss!

Alas, in this case I think it's a matter of recognising that in our parlance we've _already_ overloaded this term.

@aphillips, let's fast-track this one once @cdaringe's CLA is sorted out.

aphillips:
@cdaringe Please agree to the Contributor License Agreement (see link above).

cdaringe:
Thx @aphillips. My representation is on behalf of my employer, so I'm working with them to sign the agreement s.t. I can sign on behalf of them. Is there any special place for my employer to sign?

macchiati:
Your employer will need to sign a corporate CLA, see https://www.unicode.org/policies/licensing_policy.html#CLA

cdaringe:
Signing on behalf of self, working in the corporate sig for long term

PR #402, opened 2023-06-24
Require variable names to be globally unique

catamorphism:
See https://github.com/unicode-org/message-format-wg/issues/310

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -124,8 +124,22 @@ To resolve the value of a _variable_,
 its _name_ is used to identify either a local variable,
 or a variable defined elsewhere.
 
-It is an error for the right-hand side of a local variable declaration to
-refer to a local variable that's declared after it in the message.
+When a _variable_is used as an _operand_,
+the _name_ of this variable MUST either be:
+1. The _name_ of a _variable_ of a _declaration_ that occurs before the
+_operand_, or:
+1. An externally defined variable.
+
+In case 1, the `entire` _declaration_ MUST appear before the _operand_.
+For example, resolution of the following declaration should fail:
+
+```
+let $var = {$var}
+```
+
+because although a _variable_ with _name_ `var` occurs before
+the expression `{$var}`, the entire declaration `let $var = {$var}`
+does not occur before the expression `{$var}`.
 
 Variable names are required to be globally unique. That is,
 for any local variable declaration of the form `let $v = e`:
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -123,11 +123,15 @@ The resolution of a _text_ or _literal_ token MUST always succeed.
 To resolve the value of a _variable_,
 its _name_ is used to identify either a local variable,
 or a variable defined elsewhere.
-If a local variable and an externally defined one use the same name,
-the local variable takes precedence.
 
-It is an error for a local variable definition to
-refer to a local variable that's defined after it in the message.
+It is an error for the right-hand side of a local variable declaration to
+refer to a local variable that's declared after it in the message.
+
+Variable names are required to be globally unique. That is,
+for any local variable declaration of the form `let $v = e`:
+* It is an error if `v` is the left-hand side of a local variable declaration
+that appears before `let $v = e` in the message.
+* It is also an error if `v` is an externally defined variable.
 
 The resolution of a _variable_ MAY fail if no value is identified for its _name_.
 If this happens, an Unresolved Variable error MUST be emitted.
@@ -229,6 +233,24 @@ rather than the _expression_ in the _selector_ or _pattern_.
 > resolving to a _fallback value_ of `|horse|`.
 
 _Pattern selection_ is not supported for _fallback values_.
+
+When there are multiple _declarations_ for the same _variable_,
+the fallback string is formatted based on the _expression_ of
+the first declaration of that _variable_.
+
+> For example, attempting to format the following message:
+>
+> ```
+> let $var = {|cart|}
+> let $var = {|horse|}
+> {The value is {$var}.}
+> ```
+>
+> would result in this formatted string representation:
+>
+> ```
+> The value is {|cart|}.
+> ```
 
 ## Pattern Selection
 
@@ -702,6 +724,32 @@ These are divided into the following categories:
     > match {$var}
     > when 1 {The value is one.}
     > when * {The value is not one.}
+    > ```
+
+  - **Variable Redefinition errors** occur when
+    the left-hand side of a variable declaration is already defined,
+    either by a previous local variable declaration,
+    or by an external definition.
+
+    > For example, attempting to format the following message
+    > must result in a Variable Redefinition error,
+    > because there is another declaration for `var`
+    > that textually precedes
+    > the declaration of `var` to be `{|horse|}`.
+    >
+    > ```
+    > let $var = {|cart|}
+    > let $var = {|horse|}
+    > {The value is {$var}.}
+    > ```
+    >
+    > Attempting to format the following message in a context
+    > where `var` is externally defined
+    > should also result in a Variable Redefinition error.
+    >
+    > ```
+    > let $var = {|horse|}
+    > {The value is {$var}.}
     > ```
 
   - **Unknown Function errors** occur when an _expression_ includes
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -744,7 +744,7 @@ These are divided into the following categories:
     > ```
 
   - **Variable Redefinition errors** occur when
-    the left-hand side of a variable declaration is already defined,
+a `declaration` contains a `variable` that is already defined
     either by a previous local variable declaration,
     or by an external definition.
 
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -749,7 +749,7 @@ These are divided into the following categories:
     or by an external definition.
 
     > For example, attempting to format the following message
-    > must result in a Variable Redefinition error,
+    > results in a Variable Redefinition error,
     > because there is another declaration for `var`
     > that textually precedes
     > the declaration of `var` to be `{|horse|}`.
@@ -762,7 +762,7 @@ These are divided into the following categories:
     >
     > Attempting to format the following message in a context
     > where `var` is externally defined
-    > should also result in a Variable Redefinition error.
+    > also results in a Variable Redefinition error.
     >
     > ```
     > let $var = {|horse|}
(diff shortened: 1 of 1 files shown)


spec/formatting.md
```diff
@@ -23,11 +23,15 @@ are treated as identical.
 To resolve the value of a _variable_,
 its _name_ is used to identify either a local variable,
 or a variable defined elsewhere.
-If a local variable and an externally defined one use the same name,
-the local variable takes precedence.
 
-It is an error for a local variable definition to
-refer to a local variable that's defined after it in the message.
+It is an error for the right-hand side of a local variable declaration to
```
aphillips, line 27: the right and left hand sides have names in the ABNF. You should use those. In this case it should be something like:

```suggestion
The `expression` part of a `declaration` MUST NOT refer to a `variable` that is declared after it in the `message`.
```

(Technically we could just say "an `expression` MUST NOT refer to a `variable` that has not been assigned in a previous `declaration` or passed in...")

spec/formatting.md
```diff
@@ -23,11 +23,15 @@ are treated as identical.
...
+refer to a local variable that's declared after it in the message.
+
+Variable names are required to be globally unique. That is,
```
aphillips, line 30: Use normative language.

```suggestion
Each `variable` MUST be unique within a message. It is a (what type?) error if a `declaration` uses a name assigned by a previous `declaration`.
```

spec/formatting.md
```diff
@@ -23,11 +23,15 @@ are treated as identical.
...
+for any local variable declaration of the form `let $v = e`:
+* It is an error if `v` is the left-hand side of a local variable declaration
+that appears before `let $v = e` in the message.
+* It is also an error if `v` is an externally defined variable.
```
aphillips, line 34: We discussed this briefly in the recent call (talking about changing the sigil for external values). I think this is bad because it is easy for code to suddenly produce an error when the input data (which is dynamic) collides with a declaration.

spec/formatting.md
```diff
@@ -428,6 +432,32 @@ These are divided into the following categories:
     > when * {The value is not one.}
     > ```
 
+  - **Variable Redefinition errors** occur when
+    the left-hand side of a variable declaration is already defined,
```
aphillips, line 436: Get rid of "left-hand side". The formal name `variable` and the term `definition` work fine.

```suggestion
a `declaration` contains a `variable` that is already defined
```

spec/formatting.md
```diff
@@ -428,6 +432,32 @@ These are divided into the following categories:
...
+    either by a previous local variable declaration,
+    or by an external definition.
+
+    > For example, attempting to format the following message
+    > must result in a Variable Redefinition error,
```
aphillips, line 441: must => MUST

... although examples are by definition non-normative, which means you should avoid normative language.... so...

must result => results

spec/formatting.md
```diff
@@ -428,6 +432,32 @@ These are divided into the following categories:
...
+    >
+    > ```
+    > let $var = {|cart|}
+    > let $var = {|horse|}
+    > {The value is {$var}.}
+    > ```
+    >
+    > Attempting to format the following message in a context
+    > where `var` is externally defined
+    > should also result in a Variable Redefinition error.
```
aphillips, line 454: see above. And "should" is super wrong 😉 

Just say "results" or "also results"

spec/formatting.md
```diff
@@ -590,3 +620,21 @@ rather than the _expression_ in the _selector_ or _pattern_.
 > ```
 > The value is {|horse|}.
 > ```
+
+When there are multiple _declarations_ for the same _variable_,
+the fallback string is formatted based on the _expression_ of
+the first declaration of that _variable_.
```
aphillips, line 626: what about if `var` were externally defined? And why is this a good fallback string?

aphillips (changes requested):
I oppose banning redefinitions so don't want to merge this at all. I have a couple of technical comments about that.

Most comments below, however, are with my chair hat on and pertain to formatting and normative language.

mihnita:
I think it would be good to use `const` instead of `let` or `var`, as that would make it a lot more obvious that they should not be overridden, without reading the spec.
I was proposed by Addison last Monday, and I think it is a good idea.

---

I also had a meeting with Stas last week, and he would really like to see something that mitigates the shadowing between these "local variables" (consts :-) and the "externally defined variables" (edv :-) or "arguments"

I proposed another sigil, but he would rather avoid that, and proposed either `$$`, or `_` in the name.
So local consts would be
```
const $_foo = {$foo :number ...}
const $foo = {...} // ERROR, local variable name does not start with _
const $_foo = {...} // ERROR, overrides a "externally defined variables" / argument
```

One of the added benefits (in addition to shadowing) is that we can do better static checking.
If the message uses any `_` named placeholder, is must also have a `const` in the same message.
That is better than using the same convention for locals / externally defined.

I am fine with `_`. I can even live with `$$`, although not a fan :-)

spec/formatting.md
mihnita: +1 to Addison's phrasing.

spec/formatting.md
mihnita: in the spirit of the generic comment I left on this PR, I suggest we use "constants" everywhere in this PR (and `const` keyword). (of course not for the passed-in ones)

spec/formatting.md
mihnita: Nitpick: `let $v = e` => `let $v = {$e}` (everywhere)

spec/formatting.md
mihnita: > When there are multiple _declarations_ for the same _variable_,

That would be an error.

spec/formatting.md
```diff
@@ -428,6 +432,32 @@ These are divided into the following categories:
...
+    > Attempting to format the following message in a context
+    > where `var` is externally defined
+    > should also result in a Variable Redefinition error.
+    >
+    > ```
+    > let $var = {|horse|}
+    > {The value is {$var}.}
+    > ```
+
   - **Unknown Function errors** occur when an _expression_ includes
```
mihnita, line 461: Nitpick: would be good to have a code example with a function here

spec/formatting.md
```diff
@@ -590,3 +620,21 @@ rather than the _expression_ in the _selector_ or _pattern_.
...
+
+When there are multiple _declarations_ for the same _variable_,
+the fallback string is formatted based on the _expression_ of
+the first declaration of that _variable_.
+
+> For example, attempting to format the following message:
+>
+> ```
+> let $var = {|cart|}
+> let $var = {|horse|}
```
mihnita, line 632: The second declaration (`let $var = {|horse|}`) would be an error.

mihnita:
> I suggested the HTML approach of requiring a dash - inside the name, e.g. $my-foo. I considered $$ as a prefix, but don't have a strong opinion about it.

Thank you for the correction.

(Personally I am not a fan of `-` in variable names.
But the misrepresentation of what you said was not intentional.
Sorry)

spec/formatting.md
aphillips: @mihnita Note that I'm using production names from the ABNF. If we decided to make the `variables` constant then we should change the name of the production at the same time.

eemeli (changes requested):
I don't think disallowing value definitions is a good idea.

For an alternative approach, could we consider changing the `let` keyword to be `set`, and for references to _declarations_ to be changed to _definitions_ or _assignments_? This might make it easier to consider them as assignments rather than initializations.

I do not think using `const` would be a good idea, esp. as many such "constants" would be variable depending on the formatting call arguments.

spec/formatting.md
catamorphism: I made an attempt to rephrase this. 

This would be much easier to specify if, in the grammar, `declaration` and `operand` were changed:

```
declaration = let s variable-definition [s] "=" [s] expression
operand = literal / variable-reference
option = name [s] "=" [s] operand
variable-definition = variable
variable-reference = variable
```

If this change were made, then I'd be able to write something like:

"The _variable_ of any `variable-reference` must be preceded by a full _declaration_ whose `variable-definition` has the same _variable_."

and could avoid referring to the enclosing context of "the operand".

spec/formatting.md
catamorphism: Rewrote this; for now I'll defer the `const` vs. `let` question and try to focus on making the language precise.

spec/formatting.md
catamorphism: @aphillips I agree that it would be preferable to change the sigil for externally-defined variables. If that change lands before this PR does (i.e. if #403 is resolved), then I'll update this PR to reflect that.

@mihnita I think my rewrites that I just pushed eliminate metavariables anyway, but if I end up re-adding any I'll keep that in mind.

spec/formatting.md
catamorphism: Do you mean a function in the declaration of `$var`, or a function in the expression that uses `$var`, or both?

spec/formatting.md
catamorphism: > what about if `var` were externally defined?

Rewrote to include that case as well.

> And why is this a good fallback string?

See https://github.com/unicode-org/message-format-wg/blob/main/spec/formatting.md?plain=1#L208

I wanted to be as consistent as possible with the existing spec, which says that fallback values should be formatted based on the RHS of a declaration. I thought the most consistent thing if there are multiple declarations is to pick one of them. Do you have another suggestion?

catamorphism: > > When there are multiple _declarations_ for the same _variable_,
> 
> That would be an error.
> 

@mihnita I'm not sure what you mean -- doesn't there have to be a fallback string to replace any expression that resolves to an error?

catamorphism:
> I think it would be good to use `const` instead of `let` or `var`, as that would make it a lot more obvious that they should not be overridden, without reading the spec. I was proposed by Addison last Monday, and I think it is a good idea.
> 

I would be fine with that change, but perhaps if we get to the point where everyone is OK with this PR as-is, then renaming `let` to `const` could be a separate PR afterward?

spec/formatting.md
catamorphism: @mihnita I can't find the context for this comment since I've since pushed more commits -- I'm "resolving", but if the issue is still there, please comment again.

catamorphism:
> I don't think disallowing value re-definitions is a good idea.
> 
> For an alternative approach, could we consider changing the `let` keyword to be `set`, and for references to _declarations_ to be changed to _definitions_ or _assignments_? This might make it easier to consider them as assignments rather than initializations.

I filed a new issue for this: #413

cdaringe (approved):

eemeli:
As discussed during the last MFWG call, changes to how we handle variable names probably needs a design document.

catamorphism:
I think this pull request is made obsolete by https://github.com/unicode-org/message-format-wg/commit/5065a51fe47906f4052d8c21dad54eba5d6dc58e .

PR #411, opened 2023-07-03
Fix normative vs. non-normative keywords in formatting.md

eemeli:
This cleans up a few things that were not included in PR #396. Namely:
- A formatting context suggestion from @aphillips.
- Avoid using non-normative "may".
- Add resolution rules for private reserved annotations.

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -311,9 +311,10 @@ Next, using `res`, resolve the preferential order for all message keys:
 The method MatchSelectorKeys is determined by the implementation.
 It takes as arguments a resolved _selector_ value `rv` and a list of string keys `keys`,
 and returns a list of string keys in preferential order.
-The returned list must only contain unique elements of the input list `keys`,
-and it MAY be empty.
-The most preferred key is first, with the rest sorted by decreasing preference.
+The returned list MUST contain only unique elements of the input list `keys`.
+The returned list MAY be empty.
+The most-preferred key is first,
+with each successive key appearing in order by decreasing preference.
 
 ### Filter Variants
 
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -103,7 +103,7 @@ one of the following is used to resolve the value of the _expression_:
 - Else, the _expression_ has a _reserved_ _annotation_.
   If the _annotation_ uses a `private-start` character that the implementation supports,
   its value is resolved according to the implementation's specification.
-  Else, a Reserved Syntax error is emitted and a fallback value is used as its value.
+  Else, an Unsupported Expression error is emitted and a fallback value is used as its value.
 
 ### Literal Resolution
 
@@ -725,18 +725,18 @@ These are divided into the following categories:
     > when * {The value is not one.}
     > ```
 
-  - **Reserved Syntax errors** occur when an expression uses syntax reserved for
-    future standardization,
+  - **Unsupported Expression errors** occur when an expression uses
+    syntax reserved for future standardization,
     or for private implementation use that is not supported by the current implementation.
 
     > For example, attempting to format this message
-    > would always result in an Reserved Syntax error:
+    > would always result in an Unsupported Expression error:
     >
     > ```
     > {The value is {@horse}.}
     > ```
     >
-    > Attempting to format this message would result in an Reserved Syntax error
+    > Attempting to format this message would result in an Unsupported Expression error
     > if done within a context that does not support the `^` private use sigil:
     >
     > ```
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -103,7 +103,7 @@ one of the following is used to resolve the value of the _expression_:
 - Else, the _expression_ has a _reserved_ _annotation_.
   If the _annotation_ uses a `private-start` character that the implementation supports,
   its value is resolved according to the implementation's specification.
-  Otherwise, a fallback value is used as its value.
+  Else, a Reserved Syntax error is emitted and a fallback value is used as its value.
 
 ### Literal Resolution
 
@@ -177,15 +177,15 @@ the following steps are taken:
 
 ### Fallback Resolution
 
-The resolution of an _expression_ MAY fail in the following cases:
+A **_fallback value_** is the resolved value emitted when an _expression_ cannot be resolved.
+
+An _expression_ fails to resolve when:
 
 - A _variable_ _operand_ fails to resolve.
 - A _function_ _annotation_ fails to resolve.
 - The _expression_ has a _reserved_ _annotation_.
 
-In each such case, an error MUST be emitted
-and a **_fallback value_** used for the _expression_.
-This value depends on the shape of the _expression_:
+The _fallback value_ depends on the contents of the _expression_:
 
 - _expression_ with _literal_ _operand_: U+007C VERTICAL LINE `|`
   followed by the value of the Literal,
@@ -497,7 +497,7 @@ when *   {Other match}
 ## Formatting
 
 After _pattern selection_,
-each _text_ and _expression_ part of the selected _pattern_ must be resolved and formatted.
+each _text_ and _expression_ part of the selected _pattern_ is resolved and formatted.
 
 _Formatting_ is a mostly implementation-defined process,
 as it depends on the implementation's shape for resolved values
@@ -695,7 +695,7 @@ These are divided into the following categories:
   - **Unresolved Variable errors** occur when a variable reference cannot be resolved.
 
     > For example, attempting to format either of the following messages
-    > must result in an Unresolved Variable error if done within a context that
+    > would result in an Unresolved Variable error if done within a context that
     > does not provide for the variable reference `$var` to be successfully resolved:
     >
     > ```
@@ -712,7 +712,7 @@ These are divided into the following categories:
     a reference to a function which cannot be resolved.
 
     > For example, attempting to format either of the following messages
-    > must result in an Unknown Function error if done within a context that
+    > would result in an Unknown Function error if done within a context that
     > does not provide for the function `:func` to be successfully resolved:
     >
     > ```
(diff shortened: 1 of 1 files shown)


spec/formatting.md
```diff
@@ -174,7 +177,7 @@ the following steps are taken:
 
 ### Fallback Resolution
 
-The resolution of an _expression_ may fail in the following cases:
+The resolution of an _expression_ MAY fail in the following cases:
```
aphillips, line 180: Do we really mean _MAY_ here? This isn't normative permission, really. In this case, if one of the following things happens, the expression almost certainly fails to resolve. (To be fair, it might not fail, if the failure were something that was recoverable or irrelevant, e.g. `{$foo :bar option=$unknown}` where `option` is an option that `:bar` doesn't care about).

I think this section is slightly upside down. I'd probably change it by bringing line 186 forward and modifying it to actually define the term in question, kind of like so:

> A **_fallback value_** is the value emitted when an _expression_ cannot be resolved.
>
> Some of the reasons that an _expression_ might fail to be resolved include:
> * A _variable_ _operand_ fails to resolve.
> * A _function_'s _annotation_ fails to resolve.
> * The _expression_ contains or consists of a _reserved_ _annotation_.
>
> The _fallback value_ depends on the content of the _expression_:

(Note that the last line of my suggestion replaces an occurrence of the word "shape" :wink in the current text)

spec/formatting.md
```diff
@@ -309,7 +312,7 @@ The method MatchSelectorKeys is determined by the implementation.
 It takes as arguments a resolved _selector_ value `rv` and a list of string keys `keys`,
 and returns a list of string keys in preferential order.
 The returned list must only contain unique elements of the input list `keys`,
-and it may be empty.
+and it MAY be empty.
 The most preferred key is first, with the rest sorted by decreasing preference.
```
aphillips, line 316: The term `MUST` is also normative and it is intended to be normative, so:

```suggestion
The returned list MUST contain only unique elements of the input list `keys`.
The returned list MAY be empty.
The most-preferred key is first, with each successive key appearing in order by
decreasing preference.
```

aphillips (changes requested):
Good start. Some comments about normative language.

spec/formatting.md
eemeli: The list here is intended to be exhaustive, so I would say that MAY is appropriate. As in, we do not allow for failure in literal resolution, and failure to resolve a variable option value does not directly cause an annotation failure. The MAY is allowing for expression resolution failure in the listed cases, but not otherwise.

macchiati: The word "may" followed by a list of cases will not be understood by native speakers as being exhaustive (aka complete). Moreover, MAY is not the right term to use in such a case. 

"MAY   This word, or the adjective "OPTIONAL", mean that an item is
   truly optional.  One vendor may choose to include the item because a
   particular marketplace requires it or because the vendor feels that
   it enhances the product while another vendor may omit the same item." 

What you are conveying is

* when there is a _variable_ _operand_ that fails to resolve, the implementation might or might not resolve an expression
* when there is a _function_ _annotation_ that fails to resolve, the implementation might or might not resolve an expression
* in other circumstances as well, the implementation might or might not resolve an expression.

It depends on what you are trying to say. If you want to say that the resolution **only** fails in these two situations, and it **does** always fail in either of these two situations, you'd write:

The resolution of an _expression _ MUST_ fail in all and only the following cases: 

But as Addison says, it is probably better to just reorganize the text.

...

aphillips: @eemeli 

> The list here is intended to be exhaustive, so I would say that MAY is appropriate. As in, we do not allow for failure in literal resolution, and failure to resolve a variable option value does not directly cause an annotation failure. The MAY is allowing for expression resolution failure in the listed cases, but not otherwise.

If the list is exhaustive, then the normative term to use is `MUST`, as in:

> The resolution of an _expression_ MUST fail if any of the following conditions are met:

But I would still recommend my reformulation. To make the list exhaustive, reword my suggestion:

```
Some of the reasons that an _expression_ might fail to be resolved include:
```

to say:

```
An `expression` fails to resolve when:
```

eemeli: Point taken; I was wrong. Applied changes following @aphillips suggestions.

eemeli:
Noticed and fixed a few more uses of non-normative "must" and added a Reserved Syntax error. Its emission criteria is defined under Expression Resolution.

spec/formatting.md
```diff
@@ -721,6 +725,26 @@ These are divided into the following categories:
...
+
+    > For example, attempting to format this message
+    > would always result in an Reserved Syntax error:
+    >
+    > ```
+    > {The value is {@horse}.}
+    > ```
+    >
+    > Attempting to format this message would result in an Reserved Syntax error
+    > if done within a context that does not support the `^` private use sigil:
```
aphillips, line 740: I think this might be better as:

```suggestion
> Attemping to format the following message would result in a _Reserved Syntax error_
> if done in a context that does not support the private use sigils or where no private
> use function `^private` is available:
```

eemeli: I don't think we want to refer to "private use functions" as a concept. We've been pretty good so far to not assign any meaning to reserved annotations, as they really could mean anything.

Your `_underlines_` do remind me that we ought to apply a pass of `**_definition_**` and `_use_` styling to errors.

aphillips: @eemeli ... and yeah, we need to do a doc sweep for formatting and MUSTard consistency.

eemeli: The language that I'm proposing is attempting to define this as a well-defined error in an implementation that does not attach any meaning to `^`, and it's explicitly not defining what might happen in an implementation that does attach some significance to the sigil. It could, for instance, be a custom comment format.

> I think maybe another important consideration is that we don't want interstitial processes to yak on messages that the ultimate consumer might be able to process (or might rely on), e.g. translation tools should not reject `^private`. Only the act of formatting should produce an error for reserved or private sequences (the "I don't know what this is" error)

Agreed. That's why this is classified as a Resolution error and not a Syntax error, despite the name.

aphillips: > The language that I'm proposing is attempting to define this as a well-defined error in an implementation that does not attach any meaning to ^, and it's explicitly not defining what might happen in an implementation that does attach some significance to the sigil. It could, for instance, be a custom comment format.

Understood. I think you can't say "doesn't attach _any_ meaning to `^`" because that sigil always means "private use gorp follows". "There is a private use expression here, but I don't know what it means" or "there is a private use expression here, but I don't support formatting these".

I think what I was describing (the implementation supports private use and the sigil in question, but something is wrong with the private use expression) should just be a garden variety _Resolution error_? If we weren't trying to portmanteau the reserved and private-use sigils in a single section, you'd have some repetitive text, but we can elucidate the private use requirements separately at our leisure. 

To your point about the name, I would change it from "Reserved Syntax error" to avoid the misleading term "syntax". Perhaps "reserved expression error" or (more-correct in the ABNF) "reserved annotation error"? Or, given that we want the same errors for reserved and private-use, perhaps "Unsupported expression error"?

eemeli: Renamed the error as "Unsupported Expression", as that seems suitably inclusive of both future-spec-reserved and private-reserved annotations.

spec/formatting.md
```diff
@@ -721,6 +725,26 @@ These are divided into the following categories:
     > when * {The value is not one.}
     > ```
 
+  - **Unsupported Expression errors** occur when an expression uses
+    syntax reserved for future standardization,
+    or for private implementation use that is not supported by the current implementation.
```
aphillips, line 730: I think we should try to be consistent when talking about private use. Also, remember your formatting. I'd suggest:

```suggestion
or for _annotations_ that use a _private-use_ sequence unsupported by the current implementation.
```

aphillips (changes requested):
All comments are editorial, otherwise I would approve.

spec/formatting.md
```diff
@@ -721,6 +725,26 @@ These are divided into the following categories:
     > when * {The value is not one.}
     > ```
 
+  - **Unsupported Expression errors** occur when an expression uses
```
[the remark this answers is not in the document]
eemeli: The current plural language matches that used by all the other errors. Could fixes to the style used for the error descriptions be handled as a separate PR, rather than further increasing the scope of this PR?

spec/formatting.md
eemeli: Your proposed change does not match our current description of _reserved_, which does not so differentiate annotations reserved for future spec versions compared to those reserved for private implementation use. Furthermore, we do not have a definition of _private-use_ to refer to.

It's clear that the language around _reserved_ could be updated, but that really ought to happen holistically rather than as a part of this PR, which so far is only touching the formatting parts of the spec.

spec/formatting.md
```diff
@@ -721,6 +725,26 @@ These are divided into the following categories:
...
+    syntax reserved for future standardization,
+    or for private implementation use that is not supported by the current implementation.
+
+    > For example, attempting to format this message
+    > would always result in an Unsupported Expression error:
```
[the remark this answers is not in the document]
eemeli: Could this fix also be included in a separate PR of all our error definitions and mentions? The current proposal matches the style used by all our current error references.

spec/formatting.md
```diff
@@ -721,6 +725,26 @@ These are divided into the following categories:
...
+    >
+    > ```
+    > {The value is {@horse}.}
+    > ```
+    >
+    > Attempting to format this message would result in an Unsupported Expression error
+    > if done within a context that does not support the `^` private use sigil:
```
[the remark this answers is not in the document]
eemeli: This too seems like it should be a part of a separate subsequent PR updating our error definitions, as the current language matches that used by all our other error examples.

spec/formatting.md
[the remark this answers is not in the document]
aphillips: I tend to favor "not digging the hole deeper" when committing text, but since we'll probably merge this in an hour, I can come back and make a more holistic set of edits later.

With regard to this comment (and noting that it rhymes with other comments in other PRs), the style I prefer is something like:

> **Term** is (definition here).
>
> (discussion of _term_'s attributes, usage, triggers, etc., with _term_ referencing the definition)

I tend to prefer singular, because automatic linking of references to the definition is easier when the term is originally singular. Unlike W3C, we don't have markup to use for alternate forms of the term and will have to write the link generation eventually as a script (if we decide we want linking). 🤷

spec/formatting.md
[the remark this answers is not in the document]
aphillips: See ⬆️ 

The current error references should be edited to italicize terms like our spec's front matter says we do, so this change really should be committed and other, inconsistent, text fixed. I will file an issue to do formatting and take that as an action after I click "Comment" here 😉

PR #359, opened 2023-02-28
Change the literal delimiter to the vertical pipe character.

stasm:
Closes #263.

PR #414, opened 2023-07-07
Use `"` or `'` instead of `|` as quote delimiter

eemeli:
Please see the following discussions for context on how we ended up with `|quotes|`:
- #263
- #276
- #359

Now that we've made it possible to use unquoted values as operands, I realise that my opinion on this topic has changed, and so I'm proposing that we allow both `'quoted'` and `"quoted"` literals. I have three reasons for this:
1. With the ability to write an expression like `{42 :number}`, I'm much less bothered by using string-y quote characters for delimiting literal values. In particular, if/once we also allow for negative numbers to not be quoted.
2. Fundamentally, I believe that this would make our syntax less _weird_. We've spent quite a bit of effort on ensuring that MF2 syntax is relatively readable, so that people don't feel like they need to read the spec to understand what a message means, or how to write a message. Using `|` as literal delimiters (or `()` before the current choice) is really rather novel, and that's not really a good thing.
3. Allowing either `'` or `"` should reduce the need to escape these delimiters either due to characters in the literal value, or due to the delimiters of a surrounding resource format.

As one negative consequence of this proposed change, this might mean a little more care is necessary when moving the source of a message from one encapsulation to another one. I think this is well justified given the benefits.

### DIFF
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -30,10 +30,12 @@ text-char = %x0-5B         ; omit \
           / %x7E-D7FF      ; omit surrogates
           / %xE000-10FFFF
 
-quoted      = "|" *(quoted-char / quoted-escape) "|"
-quoted-char = %x0-5B         ; omit \
-            / %x5D-7B        ; omit |
-            / %x7D-D7FF      ; omit surrogates
+quoted      = "'" *(quoted-char / DQUOTE / quoted-escape) "'"
+            / DQUOTE *(quoted-char / "'" / quoted-escape) DQUOTE
+quoted-char = %x0-21         ; omit "
+            / %x23-26        ; omit '
+            / %x28-5B        ; omit \
+            / %x5D-D7FF      ; omit surrogates
             / %xE000-10FFFF
 
 ; based on https://www.w3.org/TR/xml/#NT-Nmtoken,
@@ -47,12 +49,15 @@ unquoted-start = name-start / DIGIT / "."
 reserved       = ( reserved-start / private-start ) reserved-body
 reserved-start = "!" / "@" / "#" / "%" / "*" / "<" / ">" / "/" / "?" / "~"
 private-start  = "^" / "&"
-reserved-body  = *( [s] 1*(reserved-char / reserved-escape / literal))
+reserved-body  = *( [s] 1*(reserved-char / reserved-escape / quoted))
 reserved-char  = %x00-08        ; omit HTAB and LF
                / %x0B-0C        ; omit CR
                / %x0E-19        ; omit SP
-               / %x21-5B        ; omit \
-               / %x5D-7A        ; omit { | }
+               / %x21           ; omit "
+               / %x23-26        ; omit '
+               / %x28-5B        ; omit \
+               / %x5D-7A        ; omit {
+               / %x7C           ; omit }
                / %x7E-D7FF      ; omit surrogates
                / %xE000-10FFFF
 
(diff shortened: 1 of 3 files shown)


spec/message.abnf
```diff
@@ -47,12 +49,15 @@ unquoted-start = name-start / DIGIT / "."
 reserved       = ( reserved-start / private-start ) reserved-body
 reserved-start = "!" / "@" / "#" / "%" / "*" / "<" / ">" / "/" / "?" / "~"
 private-start  = "^" / "&"
-reserved-body  = *( [s] 1*(reserved-char / reserved-escape / literal))
+reserved-body  = *( [s] 1*(reserved-char / reserved-escape / quoted))
```
eemeli, line 52: This is an unrelated fix that I noticed while working on this PR.

aphillips: **Please make a separate PR for this.**

The intention was to allow `reserved` to define whatever structure it wants. For example, allowing "comments" might look something like:

```
{% This is a "comment". Do not translate it}
{$foo % This is a "comment" about $foo: do not translate it}
```

The contents of  `reserved` must appear within a placeholder, so the characters `{` and `}` must be escaped--but everything else (including quotes, whichever ones we use) remain undefined.

Hence `reserved-char` should match `text-char`. The production `reserved-char` currently omits whitespace, but the optional whitespace part of `reserved-body` just adds the spaces back.

aphillips:
(as contributor):
I heartily agree with your reasoning about using normal quotation marks. After all this time, `|horse|` just looks weird to me.

However, I note that our syntax will be embedded into file formats (I cited JSON in the #263 thread) or into code. These syntaxes already use `'` or `"` for quotes, so ours will have to be escaped in those contexts. Indeed, I believe the reason you want to add both quotes is because JS allows such quote alternation--but other syntaxes are more rigorous about one or the other and do not. Messages being ported between runtime environments with different quoting rules will be prone to failures due to over- and underzealous escaping.

(as chair):
This issue was previously resolved in February after long debate. This cannot be merged until we discuss it as a group. You need to write up the pros/cons for consideration.

stasm:
I think I understand @eemeli’s position, but my current view is to oppose this change.

The premise of the presented argument seems to be that the current delimiter syntax is unusual. That’s true, but more crucially, I’d say our _unquoted_ literal syntax is even more unusual. So we’re already in the realm of “weirdness”, and extending it a bit by choosing `|` as the delimiter seems reasonable to me given the “why”.

We expect quotes literals to be rare. We tend to give two corollaries from the rareness argument, which contradict each other:

1. Because it’s rare (we’ve taken care to optimize for the common case), it’s OK if the syntax for the rare case is a bit exotic. 

2. Because it’s rare, people will use it less often, and won’t always remember the correct syntax if the syntax is too exotic.

I think that just looking at these two points of view won’t lead us to a consensus. That’s where trade-offs and use-cases come into play. 

Using regular quotes as delimiters comes with a cost: embedding into most other programming languages and container formats will require caution. Using exotic delimiters prevents this one class of issues.

Exotic syntax is OK if it’s well justified. I think users will understand and learn to use it.

spec/message.abnf
eemeli: Done: #415.

mihnita:
I 100% agree with the "weirdness" part.

But I think the original reasons for `|` was that:
- we need some kind of quoting, because of possible spaces in the values quoted
- we didn't like " or ' because the conflict with a lot of existing file formats

I since none of that changed, so I would not change this.

TLDR: same concerns as Addison & Stas

cdaringe:
+0. Voicing pure neutrality on this one. This is 🎨 . The OCaml'er and pgpsql'er in me is fine with the pipes. The more common lang developer in me (TS, rust) is fine with the quotes. We're in a DSL. I like to think that consumers of DSLshave a natural intuition of  "DSL rules apply here, not my own presumptions", ...but given the correlation/pairing that MF2 has with programming concepts, it's a nice byproduct to have literal grouping aligned between paradigms. `"` is the unambiguous, ubiquitous char for bounding text content in written language, so I'd lean that way, but not powerfully enough to voice a strong stake.

eemeli:
Converting this to draft until #477 is eventually accepted, at which point this can be rebased or abandoned.

PR #415, opened 2023-07-10
Fix `reserved-body` to use `quoted` rather than `literal`

eemeli:
This was missed in #364, which was introduced before but merged after #374. Previously included in #414.

`reserved-body` needs to allow for `quoted` values (which were named `literal` at the time of its writing) to support expressions like
```
{^foo |some {text} here|}
```
as the `{}` within the `|` quotes does not need to be escaped.

As is, the ABNF is a bit broken for `reserved-body` as there are characters that are valid as either `reserved-char` or `literal`.

### DIFF
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -47,7 +47,7 @@ unquoted-start = name-start / DIGIT / "."
 reserved       = ( reserved-start / private-start ) reserved-body
 reserved-start = "!" / "@" / "#" / "%" / "*" / "<" / ">" / "/" / "?" / "~"
 private-start  = "^" / "&"
-reserved-body  = *( [s] 1*(reserved-char / reserved-escape / literal))
+reserved-body  = *( [s] 1*(reserved-char / reserved-escape / quoted))
 reserved-char  = %x00-08        ; omit HTAB and LF
                / %x0B-0C        ; omit CR
                / %x0E-19        ; omit SP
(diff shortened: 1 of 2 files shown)


aphillips:
Actually, I think this might be wrong. The idea was that the contents of `reserved-body` would be completely opaque and have no predefined structure. Thus I think it might be:

```
reserved = (reserved-start / private-start) *text
```

Note that `text` requires that `{` and `}` be escaped, which is necessary because those characters are part of the enclosing syntax.

eemeli:
> Actually, I think this might be wrong. The idea was that the contents of `reserved-body` would be completely opaque and have no predefined structure. Thus I think it might be:
> 
> ```
> reserved = (reserved-start / private-start) *text
> ```
> 
> Note that `text` requires that `{` and `}` be escaped, which is necessary because those characters are part of the enclosing syntax.

My comment on this in the original PR still stands: https://github.com/unicode-org/message-format-wg/pull/374#discussion_r1167447318

aphillips:
If one unreserved a sigil, e.g. `@`, the specification would include a production for the unreserved value. Updated processors would parse inside the resulting placeholder. Non-updated processors would ignore the entire placeholder. There is no need to reserve the literal quote sigil in a private use or reserved block because it doesn't mean anything there--and it is important that it not mean anything there.

The `reserved-body` production does not allow `{@foo bar={|literal|}}` (which is your example) because the `{` and `}` inside the body have to be escaped--but that escaping would prevent `bar`'s assignment when `@` becomes unreserved. We currently don't allow any nested placeholders, so escaping `{` and `}` works. But reserving the quote character means that the reserved body has to be changed.

aphillips:
@eemeli Fair enough. I created #417 as a placeholder for fixing the reserved production.

aphillips (approved):

PR #419, opened 2023-07-12
Styling and structure changes to syntax.md

aphillips:
This includes fixing the style of **_definitions_** and their _references_.

In the course of doing this change, I made some non-editorial changes:

* Introduce terms and definitions to clarify _selectors_, _match statement_, and _selector_ and concepts. Our text was difficult to read because there was no way to refer to the `match` clearly (because the production is called `selectors`) or to refer to a specific _selector_ (or just the selector expressions minus the keyword `match`). Hopefully this is a step in the right direction. This transformation is incomplete, since there are several places that refer to `match` rather than _match statement_, etc.

* I split `private-use` from `reserved`, although that really should be a separate PR.

* I modified the structure and description of the _quoted_ and _unquoted_ literals to make their relationship clearer.

* I reorganized and made normative text in several places. I believe that the normative changes are not a material change: they are just applying the SHOUTY KEYWORDS appropriately or making explicit what is only implied by the ABNF.

* I added `quoted` to the list of places where whitespace is meaningful. I think this was previously an oversight.

I expect tons of comments/hate mail ;-)

### DIFF
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -471,8 +471,8 @@ option = name [s] "=" [s] (literal / variable)
 A **_reserved_** _annotation_ is an _annotation_ whose syntax is reserved
 for future standardization.
 
-A _reserved_ annotation starts with a reserved character.
-A _reserved_ _annotation_ MAY be empty or contain arbitrary text.
+A _reserved_ _annotation_ starts with a reserved character.
+A _reserved_ _annotation_ MAY be empty or contain arbitrary text after its first character.
 This allows maximum flexibility in future standardization,
 as future definitions are expected to define additional semantics and constraints
 on the contents of these _annotations_.
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -281,8 +281,9 @@ of the selection process.
 
 A **_match statement_** signals that the _message_ contains more than one
 _variant_ that can potentially be used to format as output.
-A _match_ statement MUST begin with the keyword `match`.
-A _match_ statement MUST be followed by one or more _selector expressions_.
+A _match statement_ MUST begin with the keyword `match`.
+A _match statement_ MUST contain one or more _selector expressions_.
+A _match statement_ MUST be followed by at least one _variant_.
 
 ```abnf
 selectors = match 1*([s] expression)
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -520,8 +520,8 @@ text-char = %x0-5B         ; omit \
 A **_literal_** is a character sequence that appears outside
 of _text_ in various parts of a _message_.
 A _literal_ can appear in a _declaration_, as a _key_ value,
-as an _operand_, or in the value of an _option_. A _literal_ MAY
-include any Unicode code point
+as an _operand_, or in the value of an _option_.
+A _literal_ MAY include any Unicode code point
 except for surrogate code points U+D800 through U+DFFF.
 
 All code points are preserved.
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -424,7 +424,7 @@ pattern = "{" *(text / expression) "}"
 ### Expressions
 
 An _expression_ can appear as a local variable value, as a _selector_, and within a _pattern_.
-Each _expression_ MUST start with an _operand_ or an _annotation_.
+The contents of each _expression_ MUST start with an _operand_ or an _annotation_.
 An _expression_ MUST NOT be empty.
 
 An **_operand_** is a _literal_ or a _variable_ to be evaluated in an _expression_.
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -329,7 +329,7 @@ that varies in content or form depending on values determined at runtime.
 A **_selector expression_** is an _expression_ that will be used as part 
 of the selection process.
 
-A **_match statement_** indicates that the _message_ contains more than one
+A **_match statement_** indicates that the _message_ contains at least one
 _variant_ that can potentially be used to format as output.
 A _match statement_ MUST begin with the keyword `match`.
 A _match statement_ MUST contain one or more _selector expressions_.
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -104,15 +104,18 @@ to be formatted as a unit.
 All _patterns_, including simple ones, begin with U+007B LEFT CURLY BRACKET `{` 
 and end with U+007D RIGHT CURLY BRACKET `}`.
 
->    {Hello, world!}
+> A _message_ consisting of a simple _pattern_:
+>>```
+>>{Hello, world!}
+>>```
 
->The same message defined in a `.properties` file:
+>The same _message_ defined in a `.properties` file:
 >
 >>```properties
 >>app.greetings.hello = {Hello, world!}
 >>```
 
->The same message defined inline in JavaScript:
+>The same _message_ defined inline in JavaScript:
 >
 >>```js
 >>let hello = new MessageFormat('{Hello, world!}')
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -6,7 +6,7 @@
    1. [Design Goals](#design-goals)
    1. [Design Restrictions](#design-restrictions)
 1. [Overview & Examples](#overview--examples)
-   1. [Messages](#message)
+   1. [Messages and Patterns](#messages-and-patterns)
    1. [Expressions](#expression)
    1. [Formatting Functions](#function)
    1. [Selection](#selection)
@@ -91,11 +91,17 @@ The syntax specification takes into account the following design restrictions:
 
 _This section is non-normative._
 
-### Message
+### Messages and Patterns
 
 A **_message_** is the complete template for a specific message formatting request.
 
-All _messages_, including simple ones, begin with U+007B LEFT CURLY BRACKET `{` 
+All _messages_ MUST contain either a _pattern_ or _selectors_. 
+A _message_ MAY contain zero or more _declarations_.
+(Implication: the empty string is not a valid message)
+
+A **_pattern_** is a combination of _text_ and _placeholders_ 
+to be formatted as a unit.
+All _patterns_, including simple ones, begin with U+007B LEFT CURLY BRACKET `{` 
 and end with U+007D RIGHT CURLY BRACKET `}`.
 
 >    {Hello, world!}
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -576,7 +576,7 @@ The grammar defines the following tokens for the purpose of the lexical analysis
 
 ### Keywords
 
-A **_keyword_** is a reserved token that is part of the _message_ syntax.
+A **_keyword_** is a reserved token that has a unique meaning in the _message_ syntax.
 
 The following three keywords are reserved: `let`, `match`, and `when`.
 Reserved keywords are always lowercase.
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -89,15 +89,15 @@ The syntax specification takes into account the following design restrictions:
 
 ## Overview & Examples
 
-_This section is non-normative._
-
 ### Messages and Patterns
 
 A **_message_** is the complete template for a specific message formatting request.
 
-All _messages_ MUST contain either a _pattern_ or _selectors_. 
-A _message_ MAY contain zero or more _declarations_.
+All _messages_ MUST contain a _body_.
+The _body_ of a _message_ consists of either a _pattern_ or of _selectors_. 
 An empty string is not a valid _message_.
+
+A _message_ MAY also contain zero or more _declarations_ before the _body_.
 
 A **_pattern_** is a combination of _text_ and _placeholders_ 
 to be formatted as a unit.
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -374,6 +374,7 @@ A _well-formed_ message is considered _valid_ if the following requirements are 
 ### Patterns
 
 A **_pattern_** is a sequence of translatable elements.
+A _pattern_ MAY be empty.
 Patterns MUST be delimited with `{` at the start, and `}` at the end.
 This serves 3 purposes:
 
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -240,10 +240,10 @@ A complex message with 2 _selectors_ and 3 local variable _declarations_:
 
 The specification defines the following grammar productions.
 
-A message satisfying all rules of the grammar is considered _well-formed_.
+A **_well-formed_** message satisifies all of the rules of the grammar.
 
-Furthermore, a well-formed message is considered _valid_
-if it meets additional semantic requirements about its structure, defined below.
+A **_valid_** message meets the additional semantic requirements about 
+the structure and functionality defined below.
 
 ### Message
 
@@ -272,8 +272,17 @@ declaration = let s variable [s] "=" [s] expression
 
 ### Selectors
 
-A `match` statement contains one or more **_selectors_**
-which will be used to choose one of the _variants_ during formatting.
+**_Selectors_** are a _match statement_ followed by one or more _variants_.
+_Selectors_ provide the ability for a _message_ to use a _pattern_
+that varies in content or form depending on values determined at runtime.
+
+A **_selector expression_** is an _expression_ that will be used as part 
+of the selection process.
+
+A **_match statement_** signals that the _message_ contains more than one
+_variant_ that can potentially be used to format as output.
+A _match_ statement MUST begin with the keyword `match`.
+A _match_ statement MUST be followed by one or more _selector expressions_.
 
 ```abnf
 selectors = match 1*([s] expression)
@@ -296,8 +305,9 @@ selectors = match 1*([s] expression)
 
 ### Variants
 
-A **_variant_** is a keyed _pattern_.
-The keys are used to match against the _selectors_ defined in the `match` statement.
+A **_variant_** is a _pattern_ associated with a set of _keys_. Each _variant_ 
+MUST begin with the _keyword_ `when`,be followed by a sequence of _keys_, and
+terminate with a valid _pattern_.
 The key `*` is a "catch-all" key, matching all selector values.
 
 ```abnf
@@ -341,14 +351,21 @@ Whitespace within a _pattern_ is meaningful and MUST be preserved.
 
 ### Expressions
 
-**_Expressions_** MUST start with an _operand_ or an _annotation_.
+An **_expression_** is a _selector_ or a placeholder in a _pattern_
+that is evaluated during the formatting process. Each _expression_ MUST 
+start with an _operand_ or an _annotation_.
 An _expression_ MUST NOT be empty.
 
-An **_operand_** is either a _literal_ or a _variable_.
+An **_operand_** is either a _literal_ or a _variable_ that is the "subject"
+of an _expression_, that is, it is the value upon which the _expression_ will
+be evaluated.
 An _operand_ MAY be optionally followed by an _annotation_.
 
 An **_annotation_** consists of a _function_ and its named _options_,
 or consists of a _reserved_ sequence.
+
+A **_function_** is functionality used to evaluate, format, select, or otherwise 
+process an _operand_, or, if lacking an _operand_, its _annotation_. 
 
 _Functions_ do not accept any positional arguments
 other than the _operand_ in front of them.
@@ -412,9 +429,10 @@ option = name [s] "=" [s] (literal / variable)
 
 #### Reserved
 
-**_Reserved_** annotations start with a reserved character
-and are intended for future standardization
-as well as private implementation use.
+A **_reserved_** _annotation_ is an _annotation_ whose syntax is reserved
+for future standardization.
+
+A _reserved_ annotation starts with a reserved character.
 A _reserved_ _annotation_ MAY be empty or contain arbitrary text.
 This allows maximum flexibility in future standardization,
 as future definitions are expected to define additional semantics and constraints
@@ -446,11 +464,27 @@ reserved-char  = %x00-08        ; omit HTAB and LF
                / %xE000-10FFFF
 ```
 
+#### Private-Use
+
+A **_private-use_** _annotation_ is an _annotation_ whose syntax is reserved
+for use by a specific implementation or by private agreement between multiple
+implementations.
+
+A _private-use_ _annotation_ starts with a `private-start` character. A _private-use_
+_annotation_ MAY be empty. The meaning and semantics of a _private-use_ _annotation_ 
+depend on the implementation. 
+
+**NOTE:** Users are cautioned that _private-use_ sequences cannot be reliably exchanged
+and can result in errors during formatting. It is generally a better idea to use
+the function registry to define additional formatting or annotation options.
+
 ## Tokens
 
 The grammar defines the following tokens for the purpose of the lexical analysis.
 
 ### Keywords
+
+A **_keyword_** is a reserved token that is part of the _message_ syntax.
 
 The following three keywords are reserved: `let`, `match`, and `when`.
 Reserved keywords are always lowercase.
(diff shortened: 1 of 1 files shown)


eemeli (changes requested):
Please leave out the _private-use_ changes from this PR.

spec/syntax.md
```diff
@@ -341,15 +352,22 @@ Whitespace within a _pattern_ is meaningful and MUST be preserved.
...
-**_Expressions_** MUST start with an _operand_ or an _annotation_.
+An **_expression_** is a _selector_ or a placeholder in a _pattern_
+that is evaluated during the formatting process. Each _expression_ MUST 
+start with an _operand_ or an _annotation_.
 An _expression_ MUST NOT be empty.
 
-An **_operand_** is either a _literal_ or a _variable_.
+An **_operand_** is either a _literal_ or a _variable_ that is the "subject"
+of an _expression_, that is, it is the value upon which the _expression_ will
+be evaluated.
```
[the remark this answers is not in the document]
aphillips: I tend to agree, but was needing to write a definition from whole cloth.

Watch for a new revision.

spec/syntax.md
```diff
@@ -341,15 +352,22 @@ Whitespace within a _pattern_ is meaningful and MUST be preserved.
 
 ### Expressions
 
-**_Expressions_** MUST start with an _operand_ or an _annotation_.
+An **_expression_** is a _selector_ or a placeholder in a _pattern_
```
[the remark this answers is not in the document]
aphillips: Yes.

spec/syntax.md
```diff
@@ -412,9 +430,10 @@ option = name [s] "=" [s] (literal / variable)
 
 #### Reserved
 
-**_Reserved_** annotations start with a reserved character
-and are intended for future standardization
-as well as private implementation use.
+A **_reserved_** _annotation_ is an _annotation_ whose syntax is reserved
+for future standardization.
```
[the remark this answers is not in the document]
aphillips: I will remove the private use section and create a separate PR (including an ABNF change) to host it.

spec/syntax.md
```diff
@@ -534,8 +582,10 @@ name-char  = name-start / DIGIT / "-" / "." / ":"
 
 ### Escape Sequences
 
-Escape sequences are introduced by U+005C REVERSE SOLIDUS `\`
-and allow the appearance of lexically meaningful characters
+An **_escape sequence_** is a two-character sequence starting with
+`\` U+005C REVERSE SOLIDUS.
```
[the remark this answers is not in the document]
aphillips: This is the W3C I18N chair in me struggling to get out... (we're working on [this masterpiece of formatting joy](https://w3c.github.io/bp-i18n-specdev/#char_ref_template))

spec/syntax.md
```diff
@@ -91,159 +91,198 @@ The syntax specification takes into account the following design restrictions:
...
-### Messages
-
-All messages, including simple ones, are enclosed in `{…}` delimiters:
+### Message
 
-    {Hello, world!}
+A **_message_** is the complete template for a specific message formatting request.
 
-The same message defined in a `.properties` file:
+All _messages_, including simple ones, begin with U+007B LEFT CURLY BRACKET `{`
```
mihnita, line 98: Is this accurate though?

A message can contain declarations, that start with `let $foo = {...}`,
or can contain a with a selector, starting with `match`.
In these cases the message does not start with `{`

The text below in the document seems to also consider these elements (declarations & selector) as part of the message:

> A _message_ containing a _declaration_ defining a local variable `$whom` which is then used twice inside the pattern:
>
> `let $whom = {$monster :noun case=accusative}`
> ...

and

> A message can have multiple selectors.
> A message with a single _selector_:
> 
> `match {$count :number}`

aphillips: You are correct: we start in code mode. Note that this is not text that I originated in this PR, so we've had a spec bug for a while.

aphillips: See if 7bd78fa fixes it.

aphillips:
I observe that there is a **lot** of repetition between the sections in syntax.md, which leads to forward definitions and the temptation to do stuff in two places.

spec/syntax.md
```diff
@@ -91,159 +91,198 @@ The syntax specification takes into account the following design restrictions:
...
+>>    {Hello, {$userObj :person firstName=long}!}
+>>```
 
-A _message_ with a `$userObj` _variable_ formatted with
-the custom `:person` _function_ capable of
-plucking the first name from the object representing a person:
+_Functions_ can be _standalone_, or can be an _opening element_ or _closing element_.
 
-    {Hello, {$userObj :person firstName=long}!}
+A **_standalone_** _function_ is not expected to be paired with another _function_.
```
mihnita, line 170: The `standalone`, `open`, `close` are properties of the expression, not the function.

aphillips: Is it _expression_ or _placeholder_? That is, does it have meaning to say that this is an "open expression":

```
let $foo = {$bar +open}
```

aphillips: ... also we should say that `standalone` is the default for expressions such as `{$foo}` that have only an operand.

spec/syntax.md
```diff
@@ -91,150 +91,198 @@ The syntax specification takes into account the following design restrictions:
...
 _This section is non-normative._
 
-### Messages
+### Messages and Patterns
 
-All messages, including simple ones, are enclosed in `{…}` delimiters:
+A **_message_** is the complete template for a specific message formatting request.
 
-    {Hello, world!}
+All _messages_ MUST contain either a _pattern_ or _selectors_.
```
eemeli, line 98: This statement is a bit odd, as _selectors_ includes at least one _pattern_. How about this instead?

```suggestion
All _messages_ MUST contain one or more _patterns_.
```

spec/syntax.md
```diff
@@ -91,150 +91,198 @@ The syntax specification takes into account the following design restrictions:
...
-### Messages
+### Messages and Patterns
 
-All messages, including simple ones, are enclosed in `{…}` delimiters:
+A **_message_** is the complete template for a specific message formatting request.
 
-    {Hello, world!}
+All _messages_ MUST contain either a _pattern_ or _selectors_. 
+A _message_ MAY contain zero or more _declarations_.
+(Implication: the empty string is not a valid message)
```
eemeli, line 100: This doesn't need to be parenthetical, even though it is re-stating an implication.

```suggestion
As an implication, an empty string is not a valid message.
```

spec/syntax.md
```diff
@@ -91,150 +91,198 @@ The syntax specification takes into account the following design restrictions:
...
+All _patterns_, including simple ones, begin with U+007B LEFT CURLY BRACKET `{` 
+and end with U+007D RIGHT CURLY BRACKET `}`.
 
-```properties
-app.greetings.hello = {Hello, world!}
-```
+> A _message_ consisting of a simple _pattern_:
+>>```
+>>{Hello, world!}
+>>```
```
eemeli, line 110: There should be no need to double-indent conde samples. This applies to all other instances as well:

```suggestion
> A _message_ consisting of a simple _pattern_:
> ```
> {Hello, world!}
> ```
```

spec/syntax.md
```diff
@@ -91,150 +91,198 @@ The syntax specification takes into account the following design restrictions:
...
 
-A _message_ with a `$userObj` _variable_ formatted with
-the custom `:person` _function_ capable of
-plucking the first name from the object representing a person:
+_Functions_ can be _standalone_, or can be an _opening element_ or _closing element_.
 
-    {Hello, {$userObj :person firstName=long}!}
+A **_standalone_** _function_ is not expected to be paired with another _function_.
+An **_opening element_** is a _function_ that SHOULD be paired with a _closing function_.
+A **_closing element_** is a _function_ that SHOULD be paired with an _opening function_.
```
eemeli, line 175: If we're including this in the spec, should these SHOULD statements also mention the expected order of the expressions, i.e. opening before closing?

spec/syntax.md
```diff
@@ -91,150 +91,198 @@ The syntax specification takes into account the following design restrictions:
 
 _This section is non-normative._
```
eemeli, line 92: If this section is now introducing normative text, this statement should be removed.

spec/syntax.md
```diff
@@ -91,150 +91,198 @@ The syntax specification takes into account the following design restrictions:
...
+An _opening element_ MAY be present in a message without a corresponding _closing element_,
+and vice versa.
 
-    {{+button}Submit{-button} or {+link}cancel{-link}.}
+>A message with two markup-like _functions_, `button` and `link`,
+>which the runtime can use to construct a document tree structure for a UI framework:
+>
+>>```
+>>    {{+button}Submit{-button} or {+link}cancel{-link}.}
+>>```
```
eemeli, line 185: With the triple backticking, the four-space indentations should also be removed. This also applies to other places where this change is applied.

spec/syntax.md
```diff
@@ -91,150 +91,198 @@ The syntax specification takes into account the following design restrictions:
...
 which defines the value of a named local variable.
 
-A _message_ containing a _declaration_ defining a local variable `$whom` which is then used twice inside the pattern:
-
-    let $whom = {$monster :noun case=accusative}
-    {You see {$quality :adjective article=indefinite accord=$whom} {$whom}!}
-
-A message defining two local variables:
-`$itemAcc` and `$countInt`, and using `$countInt` as a selector:
+>A _message_ containing a _declaration_ defining a local variable `$whom` which is then used twice inside the pattern:
```
eemeli, line 239:
```suggestion
>A _message_ containing a _declaration_ defining a local variable `$whom`
>which is then used twice inside the pattern:
```

spec/syntax.md
```diff
@@ -352,14 +401,12 @@ Whitespace within a _pattern_ is meaningful and MUST be preserved.
 
 ### Expressions
 
-An **_expression_** is a _selector_ or a placeholder in a _pattern_
-that is evaluated during the formatting process. Each _expression_ MUST 
-start with an _operand_ or an _annotation_.
+An **_expression_** is a _selector_ or a _placeholder_ in a _pattern_
+that is evaluated during the formatting process.
```
eemeli, line 405: Sorry, should've noticed this earlier: This leaves out declarations as a possible place for expressions.

eemeli (changes requested):

spec/syntax.md
eemeli: > Is it _expression_ or _placeholder_? That is, does it have meaning to say that this is an "open expression":
> 
> ```
> let $foo = {$bar +open}
> ```

We have not explicitly discussed this (yet!), but yes, that's what we have right now.

> ... also we should say that `standalone` is the default for expressions such as `{$foo}` that have only an operand.

I'm pretty sure that such a default is not currently defined, and that this is probably best considered within the context of discussing our function registry and our implementations' behaviour assigning function handlers to expressions that do not include an annotattion.

This PR should not make any such statements.

spec/syntax.md
```diff
@@ -341,15 +401,20 @@ Whitespace within a _pattern_ is meaningful and MUST be preserved.
...
 ### Expressions
 
-**_Expressions_** MUST start with an _operand_ or an _annotation_.
+An **_expression_** is a _selector_ or a _placeholder_ in a _pattern_
+that is evaluated during the formatting process.
+Each _expression_ MUST start with an _operand_ or an _annotation_.
 An _expression_ MUST NOT be empty.
 
-An **_operand_** is either a _literal_ or a _variable_.
+An **_operand_** is a _literal_ or a _variable_ to be evaluated in a _placeholder_.
```
eemeli, line 409: Declarations and selectors also have operands.

```suggestion
An **_operand_** is a _literal_ or a _variable_ to be evaluated in an _expression_.
```

spec/syntax.md
aphillips: Will do.

aphillips: Done in 5aef879

eemeli (commented):
Only a few remaining line comments left from me; see also the previous unresolved ones.

spec/syntax.md
aphillips: Probably, but we also allow unpaired appearance of these, so `{{-foo}something{+foo}}` might just be a fragment where the missing `{+foo}` and `{-foo}` are external to the message (externality is the argument for allowing unpaired open/close functions). 

There is nothing in the syntax that actually causes pairing to happen--and currently there is nothing in the registry that would lead to pairing either. Increasingly I think that the open/close syntax is either (a) merely a naming convention that carries no meaning or (b) woefully underspecified.

eemeli: I'm fine to consider changes to this separately from this PR.

spec/syntax.md
```diff
@@ -272,8 +321,18 @@ declaration = let s variable [s] "=" [s] expression
 
 ### Selectors
 
-A `match` statement contains one or more **_selectors_**
-which will be used to choose one of the _variants_ during formatting.
+**_Selectors_** are a _match statement_ followed by one or more _variants_.
+_Selectors_ provide the ability for a _message_ to use a _pattern_
+that varies in content or form depending on values determined at runtime.
```
catamorphism, line 326:
```suggestion
A _message_ can vary its content or form
depending on values determined at runtime.
_Selectors_ make this variation possible.
```

It's not really the _pattern_ that varies (the static, syntactic construct), but the formatted output (the runtime result). Each syntactic pattern is fixed; the choice of pattern varies.

spec/syntax.md
```diff
@@ -272,8 +321,18 @@ declaration = let s variable [s] "=" [s] expression
...
+
+A **_selector expression_** is an _expression_ that will be used as part 
+of the selection process.
+
+A **_match statement_** indicates that the _message_ contains more than one
```
catamorphism, line 331:
```suggestion
A **_match statement_** indicates that the _message_ contains at least one
```

**error?**

A `match` with exactly one variant is well-formed.

spec/syntax.md
```diff
@@ -296,8 +355,10 @@ selectors = match 1*([s] expression)
 
 ### Variants
 
-A **_variant_** is a keyed _pattern_.
-The keys are used to match against the _selectors_ defined in the `match` statement.
+A **_variant_** is a _pattern_ associated with a set of _keys_.
+Each _variant_ MUST begin with the _keyword_ `when`,
+be followed by a sequence of _keys_,
```
catamorphism, line 360:
```suggestion
be followed by a non-empty sequence of _keys_,
```

spec/syntax.md
```diff
@@ -331,25 +394,45 @@ This serves 3 purposes:
...
 > ```
-
-Whitespace within a _pattern_ is meaningful and MUST be preserved.
+>
+> An empty _pattern_:
+> ```
+> {}
+> ```
+>
+> Some _patterns_ with _expressions_:
```
catamorphism, line 409:
```suggestion
> Some _patterns_ that contain _text_ and/or _expressions_:
```

spec/syntax.md
```diff
@@ -331,25 +394,45 @@ This serves 3 purposes:
...
+>
+> A _pattern_ containing three spaces:
+> ```
+> {   }
+> ```
 
 ### Expressions
 
-**_Expressions_** MUST start with an _operand_ or an _annotation_.
+An _expression_ can appear as a local variable value, as a _selector_, and within a _pattern_.
```
catamorphism, line 423: **error?**

I think this is repeated from line 136.

spec/syntax.md
```diff
@@ -331,25 +394,45 @@ This serves 3 purposes:
...
+Each _expression_ MUST start with an _operand_ or an _annotation_.
```
catamorphism, line 424:
```suggestion
The contents of each _expression_ MUST start with an _operand_ or an _annotation_.
```

(I think there's something missing about curly brackets having to enclose an expression?)

spec/syntax.md
```diff
@@ -331,25 +394,45 @@ This serves 3 purposes:
...
 An _expression_ MUST NOT be empty.
 
-An **_operand_** is either a _literal_ or a _variable_.
+An **_operand_** is a _literal_ or a _variable_ to be evaluated in an _expression_.
```
catamorphism, line 427:
```suggestion
An **_operand_** MUST appear in an _expression_. An _operand_ is either a _literal_ or a _variable_.
```

spec/syntax.md
```diff
@@ -331,25 +394,45 @@ This serves 3 purposes:
...
 An _operand_ MAY be optionally followed by an _annotation_.
 
 An **_annotation_** consists of a _function_ and its named _options_,
 or consists of a _reserved_ sequence.
 
+A **_function_** is functionality used to evaluate, format, select, or otherwise 
+process an _operand_, or, if lacking an _operand_, its _annotation_. 
+
 _Functions_ do not accept any positional arguments
 other than the _operand_ in front of them.
```
catamorphism, line 437:
```suggestion
A _function_ has at most one positional argument
(its _operand_, if present).
```

spec/syntax.md
```diff
@@ -452,6 +536,8 @@ The grammar defines the following tokens for the purpose of the lexical analysis
 
 ### Keywords
 
+A **_keyword_** is a reserved token that is part of the _message_ syntax.
```
catamorphism, line 539:
```suggestion
A **_keyword_** is a reserved token that has a unique meaning in the _message_ syntax.
```

(Technically many possible tokens are part of the _message_ syntax, not just keywords)

spec/syntax.md
```diff
@@ -481,19 +567,32 @@ text-char = %x0-5B         ; omit \
 
 ### Literals
 
-**_Literal_** is used for matching variants and providing input to _expressions_.
-
-**_Quoted_** literals may include content with any Unicode code point,
+A **_literal_** is a character sequence that appears outside
+of _text_ in various parts of a _message_.
```
catamorphism, line 571: This phrasing seems backwards, since _text_ can only appear at the top level of a pattern and _literal_s can appear in many places, but I'm not sure how to rephrase it.

spec/syntax.md
```diff
@@ -481,19 +567,32 @@ text-char = %x0-5B         ; omit \
...
 
-**_Unquoted_** literals have a much more restricted range that
+All code points are preserved.
+
+A **_quoted_** literal begins and ends with U+005E VERTICAL BAR `|`.
+The characters `\` and `|` within a _quoted_ literal MUST be 
+escaped as `\\` and `\|`.
+
+An **_unquoted_** literal is a _literal_ that does not require the `|`
+quotes around it to be distinct from the rest of the _message_ syntax.
```
catamorphism, line 584: I think this language makes sense to someone writing a parser, but not to a user necessarily. Perhaps something like "a _literal_ that does not include any of the following characters..." and then refer to where the list of forbidden characters is defined?

catamorphism (commented):
Most of my comments are stylistic and it's up to your judgment whether to incorporate them. I explicitly marked the comments where I think there might be an error in the text with "**error**?"

spec/syntax.md
```diff
@@ -89,161 +89,210 @@ The syntax specification takes into account the following design restrictions:
...
 
-_This section is non-normative._
+### Messages and Patterns
 
-### Messages
+A **_message_** is the complete template for a specific message formatting request.
 
-All messages, including simple ones, are enclosed in `{…}` delimiters:
+All _messages_ MUST contain a _body_.
+The _body_ of a _message_ consists of either a _pattern_ or of _selectors_.
```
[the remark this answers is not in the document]
aphillips: Your comment is exactly my proposal to change the ABNF (see the discussion of the refactoring of this document in today's call)! The current syntax doesn't support talking about a selector cleanly.

spec/syntax.md
```diff
@@ -89,161 +89,210 @@ The syntax specification takes into account the following design restrictions:
...
-_This section is non-normative._
+### Messages and Patterns
 
-### Messages
+A **_message_** is the complete template for a specific message formatting request.
 
-All messages, including simple ones, are enclosed in `{…}` delimiters:
+All _messages_ MUST contain a _body_.
+The _body_ of a _message_ consists of either a _pattern_ or of _selectors_. 
+An empty string is not a valid _message_.
```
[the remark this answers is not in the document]
aphillips: That's correct.

spec/syntax.md
```diff
@@ -89,161 +89,210 @@ The syntax specification takes into account the following design restrictions:
...
 
-    {Hello, world!}
+A _message_ MAY also contain one or more _declarations_ before the _body_.
 
-The same message defined in a `.properties` file:
+A **_pattern_** is a combination of _text_ and _placeholders_
```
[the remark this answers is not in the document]
aphillips: In other specification regimes (think: W3C) we do auto linking of pluralized terms. While I can technically agree with excluding the "s", I think it's simpler not to.

"sequence" is a good change.

spec/syntax.md
```diff
@@ -89,161 +89,210 @@ The syntax specification takes into account the following design restrictions:
...
+The _body_ of a _message_ consists of either a _pattern_ or of _selectors_. 
+An empty string is not a valid _message_.
 
-    {Hello, world!}
+A _message_ MAY also contain one or more _declarations_ before the _body_.
 
-The same message defined in a `.properties` file:
+A **_pattern_** is a combination of _text_ and _placeholders_ 
+to be formatted as a unit.
+All _patterns_, including simple ones, begin with U+007B LEFT CURLY BRACKET `{`
```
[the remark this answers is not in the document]
aphillips: This is the original text, modified slightly. Basically a "simple" pattern has no placeholders:

```
{Hello, world}
```

spec/syntax.md
```diff
@@ -89,161 +89,210 @@ The syntax specification takes into account the following design restrictions:
...
-An _expression_ can appear as a local variable value, as a _selector_, and within a _pattern_.
+A **_placeholder_** is an _expression_ that appears inside of a _pattern_
+and which will be replaced during the formatting of the _message_.
 
-A simple _expression_ is a bare variable name:
+An _expression_ begins with U+007B LEFT CURLY BRACKET `{` 
+and ends with U+007D RIGHT CURLY BRACKET `}`.
 
-    {Hello, {$userName}!}
+An _expression_ can appear as a local variable value, as a _selector_, and within a _pattern_.
```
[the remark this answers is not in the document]
aphillips: Location is important in describing the syntax, so I think the original text is suitable for that purpose. I agree that we *also* need to describe *why* one might use an expression (not just "physical" or syntactic description, but _functional_ description too). I think those can be kept separate.

Note: a lot of your comments on this PR are about *changing* the normative aspects of the text, while this PR is fundamentally about formatting. The main semi-normative set of text changes are that I put all definitions into the form:

> A **_thing_** is {definition here}.

(with a clear definition that is *not* about the structure/serialization)

spec/syntax.md
```diff
@@ -89,161 +89,210 @@ The syntax specification takes into account the following design restrictions:
...
 
-### Formatting Functions
+> A simple _expression_ containing a variable:
+>
+>```
+>{Hello, {$userName}!}
+>```
```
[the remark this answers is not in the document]
aphillips: Yes, it should say:

```suggestion
A _message_ containing a simple _expression_ consisting of a variable name:
```

spec/syntax.md
```diff
@@ -89,161 +89,210 @@ The syntax specification takes into account the following design restrictions:
...
+>{{+button}Submit{-button} or {+link}cancel{-link}.}
+>```
 
-An opening element MAY be present in a message without a corresponding closing element,
-and vice versa.
 
 ### Selection
 
-A _selector_ selects a specific _pattern_ from a list of available _patterns_
+A **_selector_** selects a specific _pattern_ from a list of available _patterns_
```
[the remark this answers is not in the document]
aphillips: See my discussion/proposal elsewhere. This is out of scope to this PR

spec/syntax.md
aphillips: See the forthcoming refactor :-(... this is one reason that exists.

spec/syntax.md
```diff
@@ -89,161 +89,210 @@ The syntax specification takes into account the following design restrictions:
...
 
-A _message_ containing a _declaration_ defining a local variable `$whom` which is then used twice inside the pattern:
-
-    let $whom = {$monster :noun case=accusative}
-    {You see {$quality :adjective article=indefinite accord=$whom} {$whom}!}
-
-A message defining two local variables:
-`$itemAcc` and `$countInt`, and using `$countInt` as a selector:
+>A _message_ containing a _declaration_ defining a local variable `$whom`
+>which is then used twice inside the pattern:
```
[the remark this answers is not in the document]
aphillips: Actually, either are correct. The body here consists of a pattern.

spec/syntax.md
aphillips: Your proposed change loses the "definition form" that this PR is adding, that is, that a definition say *what* something is before imposing anything on it.

> An **_operand_** is _whatever it is_.

It can then go on to say that it MUST (only) appear in an expression (note that your phrase is misleading, since an expression  is not required to have an operand)

spec/syntax.md
aphillips: I tend to agree, but... an unquoted literal is not just a literal that doesn't include specific characters. What marks it as unquoted is that it is missing its quotes! That is, `|42|` is a _quoted_ literal that doesn't need to be quoted.

spec/syntax.md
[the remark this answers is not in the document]
aphillips: DONE

spec/syntax.md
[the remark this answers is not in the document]
aphillips: Done 'sequence'

PR #420, opened 2023-07-12
First draft of some registry functions

mihnita:


mihnita:
The `.md` needs updating, but I though I should do that only after we finish polishing the `.xml`.

macchiati:
Some comments from a very quick scan:

>       <option name="minimumFractionDigits" values="0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20"/>

The number could be a BigDecimal, and you don't want to list all of those possible values! Maybe just \d+?

>             Match a numerical value against CLDR plural categories or against a number literal.
should be something like:

Match a **formatted** numerical value against CLDR plural categories or against a number literal.

That is, I'm looking for some way we can make it harder (or at least alert people to )make the following mistake:

match {$count}
when one {You have a book}
when * {You have {$count :number maximumFractionDigits=0} books}

That is, when $count = 1.1, this fails, with *You have 1 books

mihnita:
Thank you!

> The number could be a BigDecimal, and you don't want to list all of those possible values! Maybe just \d+?

The info comes from ECMAScript, which limits the possible values:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#minimumfractiondigits

The intention was to start with something that is common between ICU and JS.

spec/registry.xml
```diff
@@ -0,0 +1,250 @@
...
+    One can also imagine phone numbers, emails, url, etc.
+
+    Even the regex for anyNumber (from the .md) is not OK:
+      .3  : Invalid (no digit before dot)
+      3   : Invalid (no dot + one digit after)
+            Brackets in regexp means grouping, not optional.
+      002 : Valid (OK to start with zero)
+      000 : Valid (OK to start with zero)
+  -->
+  <pattern id="iso8601" regex="\d{4}-\d{2}-\d{2}T\d{2}:\d{2}\d{2}"/>
```
aphillips, line 25: ISO8601 and RFC3339 are in the process of being extended and that work is important for support of Temporals. So we probably want to revisit here.

spec/registry.xml
```diff
@@ -0,0 +1,250 @@
...
+
+  <function name="datetime">
+    <!-- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat -->
+    <description>Language-sensitive date and time formatting</description>
```
aphillips, line 29: Language => Locale?

spec/registry.xml
```diff
@@ -0,0 +1,250 @@
...
+
+    <formatSignature>
+      <input pattern="iso8601"/>
+      <!-- The predefined date formatting style to use. -->
+      <option name="dateStyle" values="full long medium short"/>
```
aphillips, line 34: should the list be in the smallest-to-largest order?

spec/registry.xml
```diff
@@ -0,0 +1,250 @@
...
+      <option name="calendar" values="buddhist chinese coptic dangi ethioaa ethiopic gregory hebrew indian islamic islamic-umalqura islamic-tbla islamic-civil islamic-rgsa iso8601 japanese persian roc"/>
+      <!-- Numbering system to use. -->
+      <option name="numberingSystem" values="arab arabext bali beng deva fullwide gujr guru hanidec khmr knda laoo latn limb mlym mong mymr orya tamldec telu thai tibt"/>
+      <!-- The time zone to use. The only value implementations must recognize
+         is "UTC"; the default is the runtime's default time zone.
+         Implementations may also recognize the time zone names of the IANA
+         time zone database, such as "Asia/Shanghai", "Asia/Kolkata",
+         "America/New_York".
+      -->
+      <option name="timeZone" pattern="timeZoneId"/>
```
aphillips, line 47: Don't forget offset time zones.

spec/registry.xml
```diff
@@ -0,0 +1,250 @@
...
+    <!-- Whether to use 12-hour time (as opposed to 24-hour time). -->
+    <!-- I think it overlaps with "hourCycle"?
+    <option name="hour12" values="true false"/>
+    -->
+
+  </function>
+
+  <function name="number">
+    <!-- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat -->
+    <description>Language-sensitive number formatting</description>
```
aphillips, line 111: s/Language-sensitive/Locale-sensitive/g (won't repeat this later)

spec/registry.xml
```diff
@@ -0,0 +1,250 @@
...
+      <option name="offset" values="positiveInteger"/>
+      <!-- The minimum number of integer digits to use.
+        A value with a smaller number of integer digits than this number will be
+        left-padded with zeros (to the specified length) when formatted.
+      -->
+      <option name="minimumIntegerDigits" values="1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20 21" default="1"/>
+      <!-- The minimum number of fraction digits to use.
+        The default for plain number and percent formatting is 0;
+        the default for currency formatting is the number of minor unit digits provided by
+        the ISO 4217 currency code list (2 if the list doesn't provide that information).
```
aphillips, line 204: ... unless there is an override that says otherwise, e.g. some curriences with the cash format (e.g. technically `COP` has two digits but practically it has none)

spec/registry.xml
```diff
@@ -0,0 +1,250 @@
...
+        Similar to exact values in number formatting, which does not come from NumberFormat(ter).
+        It is supported by both ICU and Fluent, and it is used for constructs like this:
+        "You invited A, B, and 2 more guests."
+      -->
+      <option name="offset" values="positiveInteger"/>
+      <!-- The minimum number of integer digits to use.
+        A value with a smaller number of integer digits than this number will be
+        left-padded with zeros (to the specified length) when formatted.
+      -->
+      <option name="minimumIntegerDigits" values="1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20 21" default="1"/>
```
aphillips, line 200: do we really need a list?

spec/registry.xml
```diff
@@ -0,0 +1,250 @@
...
+        the default for percent formatting is the larger of minimumFractionDigits and 0.
+      -->
+      <option name="maximumFractionDigits" values="0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20"/>
+      <!-- The minimum number of significant digits to use. -->
+      <option name="minimumSignificantDigits" values="1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20 21" default="1"/>
+      <!-- The maximum number of significant digits to use. -->
+      <option name="maximumSignificantDigits" values="1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20 21" default="21"/>
+
+      <match pattern="anyNumber"/>
+      <match values="zero one two few many"/>
```
aphillips, line 220: does * need to appear here? Some locales produce `other` as output from plural, which is effectively `*` in our syntax. That is, do we want to be able to say "this selector can request the default message"?

spec/registry.xml
```diff
@@ -0,0 +1,250 @@
...
+    -->
+
+  </function>
+
+  <function name="plural">
+    <!-- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/PluralRules/PluralRules
+      https://unicode-org.github.io/icu-docs/apidoc/dev/icu4j/com/ibm/icu/text/MessageFormat.html
+    -->
+    <description>Language-sensitive plural selection</description>
+    <matchSignature>
```
eemeli, line 186: My preference would be to include this `<matchSignature>` under `<function name="plural">`. Given that this is likely to be a lively topic of discussion, it might be easier to separate it out to a separate PR.

spec/registry.xml
```diff
@@ -0,0 +1,250 @@
...
+  <!-- ICU does not have this, and one might argue it is just syntactic sugar on top of "select".
+    ICU rejected this a few times.
+  -->
+  <function name="gender">
+    <description>Gender selection</description>
+    <matchSignature>
+      <match values="feminine masculine"/>
+    </matchSignature>
+  </function>
+
```
eemeli, line 241: I don't think these should be included here. Happy to discuss them separately.

```suggestion
```

spec/registry.xml
```diff
@@ -0,0 +1,250 @@
...
+  <!-- TODO: IGNORED for now are the functions that are ICU only or ECMAScript only. To discuss?
+    list : ECMAScripy only
+      https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat
+    relativeTime : ECMAScripy only
+      https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat
+    spellout : ICU only
+    duration : ICU only
+  -->
```
eemeli, line 249: Please raise these as separate issue(s) or PR(s) instead.

```suggestion
```

eemeli (changes requested):
This looks like a much bigger subset of the JS Intl.NumberFormat & Intl.DateTimeFormat formatters' options than I expected, which is great! Besides the line comments, I would find it easier to consider exactly which parts are being left out: Could a list of such options with their reasonings be provided?

spec/registry.xml
mihnita: Done

spec/registry.xml
mihnita: Agree. I that what we have now (regex + list of values) is not powerful enough.
Created https://github.com/unicode-org/message-format-wg/issues/422 for this (and more)

spec/registry.xml
mihnita: Created https://github.com/unicode-org/message-format-wg/issues/422 for this (and more)

mihnita:
> There is, however, no need to have that limitation here, and artificially limit more powerful implementations.
> 
> Instead, what I think we should do is document that if the value exceeds what the implementation is capable of, the maximum supported by the implementation is used.

Some of the group members strongly opposed including anything that is not supported by ECMAScript.
So I tried to start as non-controversial as possible.

I anticipate (and hope) that we will discuss PR in the next meeting.

spec/registry.dtd
```diff
@@ -9,25 +9,25 @@
 <!ATTLIST pattern id ID #REQUIRED>
 <!ATTLIST pattern regex CDATA #REQUIRED>
 
-<!ELEMENT formatSignature (input?|option*)>
+<!ELEMENT formatSignature (input*,option*)>
```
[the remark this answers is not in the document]
mihnita: Undo.

spec/registry.xml
```diff
@@ -0,0 +1,250 @@
...
+
+    One can also imagine phone numbers, emails, url, etc.
+
+    Even the regex for anyNumber (from the .md) is not OK:
+      .3  : Invalid (no digit before dot)
+      3   : Invalid (no dot + one digit after)
+            Brackets in regexp means grouping, not optional.
+      002 : Valid (OK to start with zero)
+      000 : Valid (OK to start with zero)
+  -->
```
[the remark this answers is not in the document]
mihnita: Removed, and created issue https://github.com/unicode-org/message-format-wg/issues/422 for this (and more)

spec/registry.xml
```diff
@@ -0,0 +1,250 @@
...
+    -->
+    <!-- The format matching algorithm to use. -->
+    <!--
+    <option name="formatMatcher" values="basic best-fit" default="best-fit"/>
+    -->
+    <!-- Whether to use 12-hour time (as opposed to 24-hour time). -->
+    <!-- I think it overlaps with "hourCycle"?
+    <option name="hour12" values="true false"/>
+    -->
+
```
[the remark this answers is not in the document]
mihnita: Removed, created issue https://github.com/unicode-org/message-format-wg/issues/422 for this (and more)

spec/registry.xml
```diff
@@ -0,0 +1,250 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?xml-model href="registry.dtd" type="application/xml-dtd"?>
+<registry>
+  <pattern id="anyNumber" regex="-?[0-9]+(\.[0-9]+)"/>
```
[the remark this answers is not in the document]
mihnita: Hard / impossible to represent properly with just regexp.
Created issue https://github.com/unicode-org/message-format-wg/issues/422 for this (and more)

spec/registry.xml
mihnita: ?

This `<matchSignature>` (line 186) is under `<function name="plural">` (line 181).
We just have a a comment and `<description>` in between.
And `description` is mandatory to be present, and mandatory to be before `matchSignature` (per DTD). and

spec/registry.xml
mihnita: Literal selection is really needed. Exists in ICU, exists in Fluent.
And I've seen it used a lot, very handy to have.

Removed gender, added it to issue https://github.com/unicode-org/message-format-wg/issues/422

mihnita:
Removed, created issue https://github.com/unicode-org/message-format-wg/issues/422 for this (and more)

spec/registry.xml
mihnita: Removed, created issue https://github.com/unicode-org/message-format-wg/issues/422 for this (and more)

spec/registry.xml
```diff
@@ -0,0 +1,250 @@
...
+      000 : Valid (OK to start with zero)
+  -->
+  <pattern id="iso8601" regex="\d{4}-\d{2}-\d{2}T\d{2}:\d{2}\d{2}"/>
+
+  <function name="datetime">
+    <!-- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat -->
+    <description>Language-sensitive date and time formatting</description>
+
+    <formatSignature>
+      <input pattern="iso8601"/>
```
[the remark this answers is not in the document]
mihnita: Removed, created issue https://github.com/unicode-org/message-format-wg/issues/422 for this (and more)

Should see the `iso8601` pattern as a temporary placeholder.

spec/registry.dtd
```diff
@@ -9,25 +9,25 @@
...
 
-<!ELEMENT matchSignature (input?|option*|match*)>
+<!ELEMENT matchSignature (input*,option*,match*)>
 <!ATTLIST matchSignature locales NMTOKENS #IMPLIED>
 
 <!ELEMENT input EMPTY>
 <!ATTLIST input values NMTOKENS #IMPLIED>
 <!ATTLIST input pattern NMTOKEN #IMPLIED>
-<!ATTLIST input readonly (true|false) "false">
+<!ATTLIST input l10nAccess (hidden|readonly|values|pattern|freeform) "readonly">
```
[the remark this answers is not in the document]
mihnita: Removed, created issue https://github.com/unicode-org/message-format-wg/issues/423

spec/registry.xml
mihnita: Following the ECMA order:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat#datestyle

spec/registry.xml
mihnita: It is 100% what ECMA says:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat#timezone

I tried to not include anything that is not currently supported by ECMAScript, as we know there is strong opposition to that.

I'm for it, happy to discuss, but I didn't change this xml.

Added a topic in https://github.com/unicode-org/message-format-wg/issues/422

spec/registry.xml
mihnita: Agree. But this is 100% ECMAScript quote.

mihnita:
Reverted most changes to the .dtd, except for the ones that were clearly unintentional, and prevented a valid .xml.

For example:
```
<!ELEMENT registry (function*|pattern*)>
```

This means that the registry can contain (0 or more `function`) OR (0 or more `pattern`)
But `function` AND `pattern` is illegal, it is either / or.

This makes the examples in the .md invalid, and the registry can't cover the use cases we need.

spec/registry.xml
```diff
@@ -172,17 +143,16 @@
...
 
   <function name="plural">
-    <!-- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/PluralRules/PluralRules
+    <!--
+      https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/PluralRules/PluralRules
       https://unicode-org.github.io/icu-docs/apidoc/dev/icu4j/com/ibm/icu/text/MessageFormat.html
     -->
-    <description>Language-sensitive plural selection</description>
+    <description>Locale-sensitive plural selection.
+      It matches a **formatted** numerical value against CLDR plural categories or against a number literal.
```
aphillips, line 154: I probably would phrase this slightly differently:

```suggestion
Evaluates the operand as a numerical value, including, where appropriate, any associated formatting options, to determine the most appropriate matching CLDR plural category or numeric literal value.
```

>***Note***
> My comment is also intended to maybe model what other descriptions might want to do.

spec/registry.xml
aphillips: I think it is an interesting minor issue for the registry spec.

spec/registry.xml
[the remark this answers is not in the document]
eemeli: Here's a regexp matching the JSON number spec:
```suggestion
  <pattern id="anyNumber" regex="-?(0|([1-9]\d*))(\.\d*)([eE][-+]?\d+)"/>
```

spec/registry.xml
eemeli: Ach, sorry, mistyped earlier. To clarify:

My preference would be to include this `<matchSignature>` under `<function name="number">`. Given that this is likely to be a lively topic of discussion, it might be easier to separate it out to a separate PR.

spec/registry.xml
eemeli: I would like for us to have a separate discussion about `<function name="select">`. I think we agree on the ultimate behaviour we want to enable, but disagree about some of the details in how to achieve it.

spec/registry.xml
```diff
@@ -0,0 +1,202 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?xml-model href="registry.dtd" type="application/xml-dtd"?>
+<registry>
+  <!-- All regex here are to be seen as provisory. See issue #422. -->
+  <pattern id="anyNumber" regex="-?[0-9]+(\.[0-9]+)"/>
+  <pattern id="positiveInteger" regex="[0-9]+"/>
```
eemeli, line 6:
```suggestion
  <pattern id="positiveInteger" regex="0|([1-9]\d*)"/>
```

spec/registry.xml
aphillips: @eemeli Create an issue and Agenda+ it.

Note that `SelectFormat` exists in ICU4J's message format and so would be part of compatibility for ICU's MFv1 implementation. Matching _application-defined_ enumerated data values is where it is most useful today and it is a powerful (and often necessary) tool for developers to provide message selection in place of concatenation.

I think it is super unfortunate that gender selection is often the example given for a select format. Gender and other grammatical variation stuff really needs to depend on something standardized and better defined, much like the way that `plural` was created.

eemeli: Done: #425

macchiati:
So this is ECMAScriptMessageFormatting only?

More seriously, if the restriction is removed, it simply allows clients to
pass in more digits of accuracy; doesn't mean that the implementations have
to support that in formatting.

spec/registry.xml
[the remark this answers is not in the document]
mihnita: "0", // fails because there is no dot
"123", // fails because there is no dot
"+2", // fails because plus is not allowed
"3.14" // fails because it's missing the `eE...`, which is not optional

The regexp misses some `*` here and there.

---

But that's not the point.
The point is that in general regex are not strong enough to represent everything we might need.

spec/registry.xml
mihnita: > My preference would be to include this <matchSignature> under <function name="number">. 

I think this is covered by the issue you opened (https://github.com/unicode-org/message-format-wg/issues/425). Thanks.

spec/registry.xml
mihnita: My "rules of thumb" to include something or not were:
1. is already supported by Fluent & MessageFormat
2. the "building blocks" supported by ECMAScript & ICU (think formatters, PluralRules)
3. known to be useful (might be a duplicate of 1. If both Fluent & MFv1 have it, it's probably useful)
4. relatively easy to implement

That's what I had in mind when I wrote this.

---

> I think we agree on the ultimate behaviour we want to enable, but disagree about some of the details in how to achieve it.

I am not aware of such a disagreement on plain text select, which is basic string compare.

spec/registry.xml
```diff
@@ -0,0 +1,202 @@
...
+        the ISO 4217 currency code list (2 if the list doesn't provide that information).
+      -->
+      <option name="minimumFractionDigits" values="0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20"/>
+      <!-- The maximum number of fraction digits to use.
+        The default for plain number formatting is the larger of minimumFractionDigits and 3;
+        the default for currency formatting is the larger of minimumFractionDigits and the number of minor
+        unit digits provided by the ISO 4217 currency code list (2 if the list doesn't provide that information);
+        the default for percent formatting is the larger of minimumFractionDigits and 0.
+      -->
+      <option name="maximumFractionDigits" values="0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20"/>
```
eemeli, line 138: Of some relevance to us, TG2 is rather likely to very shortly update this and `minimumFractionDigits` to support values of up to 100 rather than 20: tc39/ecma402#786

spec/registry.md
```diff
@@ -90,7 +90,7 @@ For the sake of brevity, only `locales="en"` is considered.
     <function name="number">
         <description>
             Format a number.
-            Match a numerical value against CLDR plural categories or against a number literal.
+            Match a **formatted** numerical value against CLDR plural categories or against a number literal.
```
ryzokuken, line 93: Does the imply that the input is a localized number that'd be parsed and then matched? Also since this function includes both selection and formatting, should this also say something about the formatting part?

spec/registry.xml
```diff
@@ -0,0 +1,202 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?xml-model href="registry.dtd" type="application/xml-dtd"?>
+<registry>
+  <!-- All regex here are to be seen as provisory. See issue #422. -->
+  <pattern id="anyNumber" regex="-?[0-9]+(\.[0-9]+)"/>
+  <pattern id="positiveInteger" regex="[0-9]+"/>
+  <pattern id="currencyCode" regex="[A-Z]{3}"/>
+  <pattern id="timeZoneId" regex="[a-zA-Z/]+"/>
```
ryzokuken, line 8: We could use a more restricted grammar that only allows valid IANA timezone IDs, but I don't feel very strongly about it. Exhibit A: https://tc39.es/proposal-temporal/#prod-TimeZoneIANAName

spec/registry.xml
```diff
@@ -0,0 +1,202 @@
...
+  <pattern id="anythingNotEmpty" regex=".+"/>
+  <pattern id="iso8601" regex="\d{4}-\d{2}-\d{2}T\d{2}:\d{2}\d{2}"/>
```
ryzokuken, line 10:
```suggestion
  <pattern id="iso8601" regex="\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}"/>
```

spec/registry.xml
```diff
@@ -0,0 +1,202 @@
...
+        The default for plain number formatting is the larger of minimumFractionDigits and 3;
+        the default for currency formatting is the larger of minimumFractionDigits and the number of minor
+        unit digits provided by the ISO 4217 currency code list (2 if the list doesn't provide that information);
+        the default for percent formatting is the larger of minimumFractionDigits and 0.
+      -->
+      <option name="maximumFractionDigits" values="0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20"/>
+      <!-- The minimum number of significant digits to use. -->
+      <option name="minimumSignificantDigits" values="1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20 21" default="1"/>
+      <!-- The maximum number of significant digits to use. -->
+      <option name="maximumSignificantDigits" values="1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20 21" default="21"/>
```
ryzokuken, line 142: For all of these `*Digits` `<option>`s, I wonder if we could add a convenient shorthand in the DTD... Perhaps a simple `<range start="x" end="y"/>` definition could work?

ryzokuken (approved):
This is great! Thanks for kicking this off.

spec/registry.xml
mihnita: The current registry spec does not allow us to do this, only regex and list of values.

I suggested changing the spec to allow for URLs to external specs.
Added it as a discussion point in https://github.com/unicode-org/message-format-wg/issues/422

spec/registry.xml
mihnita: I've updated all of these instances to `positiveInteger`
We might even define regex for positive integers with / without zero, if we want to be more precise.

spec/registry.md
mihnita: No. I've merged a suggestion from Addison which does not mention formatting.

There is an ongoing discussion in issue https://github.com/unicode-org/message-format-wg/issues/425
Especially this comment https://github.com/unicode-org/message-format-wg/issues/425#issuecomment-1640648720

eemeli:
> Between several people commenting on this, and the news that ECMAScript considers removing the restriction, I changed this to positiveInteger. I hope it will not be controversial :-)

The JS limit isn't being removed; it's being increased from 20 to 100. Attempting to format a number with a `minimumFractionDigits` value greater than that will still throw a RangeError. As long as we consider that to be acceptable behaviour for a comforming implementation, I'm fine with this option accepting any positive integer value.

spec/registry.dtd
```diff
@@ -1,6 +1,6 @@
-<!ELEMENT registry (function*|pattern*)>
+<!ELEMENT registry (function,pattern)*>
```
stasm, line 1: Thanks for noticing this. I filed #434 to discuss and fix this independently of this PR, so that you can focus it on `registry.xml`.

spec/registry.xml
stasm: > I think our spec makes the presence of `*` mandatory. So I didn't include `other` for plural and the other selectors.

Why omit `other` if the spec only requires `*`?

spec/registry.xml
ryzokuken: The Temporal polyfill includes a Regexp that works for this purpose, but unfortunately it isn't a literal but I imagine that if we agreed on having a restrictive regex here that only allows valid IANA tzids, then we could come up with a simplified regex that works in this case.

spec/registry.xml
```diff
@@ -0,0 +1,147 @@
...
+      <option name="currencySign" values="accounting standard" default="standard"/>
+      <!-- The formatting that should be displayed for the number. -->
+      <option name="notation" values="standard scientific engineering compact" default="standard"/>
+      <!-- Numbering system to use. -->
+      <option name="numberingSystem" values="arab arabext bali beng deva fullwide gujr guru hanidec khmr knda laoo latn limb mlym mong mymr orya tamldec telu thai tibt"/>
+      <!-- When to display the sign for the number. -->
+      <!-- "negative" value is Experimental. -->
+      <option name="signDisplay" values="auto always exceptZero never" default="auto"/>
+      <!-- The formatting style to use. -->
+      <option name="style" values="decimal currency percent unit" default="decimal"/>
```
eemeli, line 111: To consider: If we were to leave out `currency` and `unit` formatting from the default, then we wouldn't need to say how compound units need to work.

spec/registry.xml
mihnita: I think that in the longer run (not first commit) we should also add support for [Unicode Time Zone Identifiers](https://unicode.org/reports/tr35/#Time_Zone_Identifiers)

The main reason would be stability, which the Olson database IDs does not offer (see linked)

spec/registry.xml
mihnita: My thinking (why I excluded it): because they mean the same thing?
If I see `other` in the registry, I expect to also see `other` handled in the `when`

stasm: This assumption is unfortunately false, likely on accident in CLDR, although I miss the context. For Polish, `other` is specifically used for fractions, while `many` is what you'd expect to be a good `other` in most other languages.

PR #421, opened 2023-07-12
Implement `private-use` separately from `reserved`

aphillips:
- creates a separate section in the syntax spec
- creates a separate rule in abnf
- adjusts the text in `reserved` to suit

Note that I did not include changing `reserved` to use `*(text)` in this PR, but did change `private-use` to do so.

### DIFF
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -42,11 +42,15 @@ unquoted = unquoted-start *name-char
 unquoted-start = name-start / DIGIT / "."
                / %xB7 / %x300-36F / %x203F-2040
 
-; reserve additional sigils for use by future versions of this
-; specification or for private use by implementations
-reserved       = ( reserved-start / private-start ) reserved-body
-reserved-start = "!" / "@" / "#" / "%" / "*" / "<" / ">" / "/" / "?" / "~"
+
+; reserve sigils for private-use by implementations
+private-use    = private-start 1*(text)
 private-start  = "^" / "&"
+
+; reserve additional sigils for use by 
+; future versions of this specification
+reserved       = reserved-start reserved-body
+reserved-start = "!" / "@" / "#" / "%" / "*" / "<" / ">" / "/" / "?" / "~"
 reserved-body  = *( [s] 1*(reserved-char / reserved-escape / literal))
 reserved-char  = %x00-08        ; omit HTAB and LF
                / %x0B-0C        ; omit CR
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -10,8 +10,8 @@ variant = when 1*(s key) [s] pattern
 key = literal / "*"
 
 expression = "{" [s] ((operand [s annotation]) / annotation) [s] "}"
-operand = literal / variable
-annotation = (function *(s option)) / reserved
+operand    = literal / variable
+annotation = (function *(s option)) / reserved / private-use
 
 literal = quoted / unquoted
 variable = "$" name
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -20,6 +20,7 @@
    1. [Patterns](#patterns)
    1. [Expressions](#expressions)
       1. [Reserved Sequences](#reserved)
+      2. [Private-Use Sequences](#private-use)
 1. [Tokens](#tokens)
    1. [Keywords](#keywords)
    1. [Text](#text)
@@ -413,17 +414,13 @@ option = name [s] "=" [s] (literal / variable)
 #### Reserved
 
 **_Reserved_** annotations start with a reserved character
-and are intended for future standardization
-as well as private implementation use.
+and are intended for future standardization.
 A _reserved_ _annotation_ MAY be empty or contain arbitrary text.
 This allows maximum flexibility in future standardization,
-as future definitions are expected to define additional semantics and constraints
+as future definitions MAY define additional semantics and constraints
 on the contents of these _annotations_.
 A _reserved_ _annotation_ does not include trailing whitespace.
 
-Implementations MAY define their own meaning and semantics for
-_reserved_ annotations that start with
-the U+0026 AMPERSAND `&` or U+005E CIRCUMFLEX ACCENT `^` characters.
 Implementations MUST NOT assign meaning or semantics to
 an _annotation_ starting with `reserved-start`:
 these are reserved for future standardization.
@@ -433,9 +430,8 @@ While a reserved sequence is technically "well-formed",
 unrecognized reserved sequences have no meaning and MAY result in errors during formatting.
 
 ```abnf
-reserved       = ( reserved-start / private-start ) reserved-body
+reserved       = reserved-start reserved-body
 reserved-start = "!" / "@" / "#" / "%" / "*" / "<" / ">" / "/" / "?" / "~"
-private-start  = "^" / "&"
 reserved-body  = *( [s] 1*(reserved-char / reserved-escape / literal))
 reserved-char  = %x00-08        ; omit HTAB and LF
                / %x0B-0C        ; omit CR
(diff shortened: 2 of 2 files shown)


spec/syntax.md
```diff
@@ -446,6 +442,37 @@ reserved-char  = %x00-08        ; omit HTAB and LF
                / %xE000-10FFFF
 ```
 
+#### Private-Use
+
+A **_private-use_** _annotation_ is an _annotation_ whose syntax is reserved
+for use by a specific implementation or by private agreement between multiple
+implementations.
```
eemeli, line 449:
```suggestion
A **_private-use_** _annotation_ is an _annotation_ whose syntax is reserved
for use by a specific implementation or by private agreement between multiple implementations. 
```

eemeli (changes requested):

spec/message.abnf
```diff
@@ -42,11 +42,15 @@ unquoted = unquoted-start *name-char
 unquoted-start = name-start / DIGIT / "."
                / %xB7 / %x300-36F / %x203F-2040
 
-; reserve additional sigils for use by future versions of this
-; specification or for private use by implementations
-reserved       = ( reserved-start / private-start ) reserved-body
-reserved-start = "!" / "@" / "#" / "%" / "*" / "<" / ">" / "/" / "?" / "~"
+
+; reserve sigils for private-use by implementations
+private-use    = private-start 1*(text)
```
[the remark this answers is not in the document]
aphillips: It should be `*(text)` as you note.

I think the bug is in `reserved` 😄 

From your comment, I suspect you and I are thinking differently about the reserved/private constructs. Let me see if I can clarify.

The way I wrote `private-use` above is that parsers do not parse anything within a private use placeholder (except for the optional operand). It is up to the private use implementation to define what, if any, structure appears in the private use "blob". A smart implementer would probably define their private use syntax to be consistent with the rest of MF... but, hey, it's their funeral.

I think it makes sense to define `reserved` in the same way and we discussed making it utterly opaque to allow for innovations in syntax in the future. I want to send a note to our future selves that, if they unreserve a sigil, it SHOULD (2119 keyword intentional) probably work just like the rest of the syntax in terms of parsing. But those folks might ignore us.

I understood your point in an earlier comment to suggest that reserved sigils continue to parse word tokens, which is why you want to include `literal` in the production. That way if a sigil is unreserved "it just works". But I would tend to say that an newly-unreserved sigil would "just work" because it would then be part of `function` or some other new ABNF.

Functionally, using `*text` would set up our syntax so that once an implementation sees a `reserved` or (unknown) `private-use` sigil, they can consume the entire placeholder up to the (not escaped) `}` (and potentially emit the error we defined in formatting.md).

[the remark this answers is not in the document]
eemeli: > I understood your point in an earlier comment to suggest that reserved sigils continue to parse word tokens, which is why you want to include `literal` in the production. That way if a sigil is unreserved "it just works". But I would tend to say that an newly-unreserved sigil would "just work" because it would then be part of `function` or some other new ABNF.

My point is that if we use `*text`, we will not be able to do that later without breaking changes.

A key reason we're introducing `reserved` is to ensure that something like MF 2.1 will be able to take e.g. `!` and allow for its use as a `function` introducer. If we do that, we allow for a valid MF 2.1 expression like
```
{!new key=|no } escape|}
```
where the `|no } escape|` is a valid `quoted` value.

If we now define `reserved` as e.g.
```abnf
reserved = reserved-start *text
```
then it would mean that parsing an MF 2.1 message with an MF 2.0 processor could produce a syntax error.

Therefore the need to explicitly allow for `quoted` in `reserved-body`: If it's not there to start with, it can never be added later to any construct that we un-reserve.

For an alternative way to avoid this issue, we could require `{}` to always be escaped as `\{` and `\}` in `quoted`.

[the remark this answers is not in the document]
aphillips: I see it now.

It would be odd to require quoting `{`/`}` in a literal (since that would be different than quoting regimes in any programming language I can think of). Since `reserved` isn't allowed and `private-use` is discouraged (but not deprecated), most users will encounter this as a quirk when writing a literal.

Alright, you've convinced me.

eemeli (commented):
The current shape of the changes to the syntax looks fine to me. There are some changes to `formattting.md` that should be applied at the same time, as its references to _reserved_ need to be updated to also deal with _private-use_.

aphillips:
@eemeli noted:

>  There are some changes to formattting.md that should be applied at the same time, as its references to reserved need to be updated to also deal with private-use.

Done in 77ef04e.

spec/formatting.md
```diff
@@ -92,18 +92,20 @@ and different implementations MAY choose to perform different levels of resoluti
...
-- Else, the _expression_ has a _reserved_ _annotation_.
-  If the _annotation_ uses a `private-start` character that the implementation supports,
-  its value is resolved according to the implementation's specification.
-  Else, an Unsupported Expression error is emitted and a fallback value is used as its value.
+- Else, if the _expression_ has a _private-use_ _annotation_,
+  its resolved value is defined according to the implementations's specification.
+- Else, if the _expression_ has a _reserved_ _annotation_,
+  an Unsupported Expression error is emitted and a fallback value is used as its value.
+  Else the contents are syntactically invalid and
+  an Unsupported Expression error is emitted and a fallback value is used as its value.
```
eemeli, line 108: I don't think "syntactically invalid" can apply here. If we'd encountered a syntax error, then we wouldn't have an _expression_ to resolve. We could in theory still get here if the message had been passed in as a (broken) data model and the implementation didn't sufficiently validate such, but then we can't presume that it originally came from any syntax and should treat it as a data model error.

In either case, we can trust that pattern selection has already resolved such a broken message to a fallback value.

```suggestion
- Else, the _expression_ has a _reserved_ _annotation_.
  An Unsupported Expression error is emitted and a fallback value is used as its value.
```

spec/syntax.md
```diff
@@ -410,20 +411,56 @@ option = name [s] "=" [s] (literal / variable)
...
+
+> Here are some examples of what _private-use_ sequences might look like:
+>> ```
+>> {Here's private use with an operand: {$foo &bar}}
+>> {Here's a placeholder that is entirely private-use: {&anything here}}
+>> {Here's a private-use function that uses normal function syntax: {$operand ^foo option=|literal|}}
+>> {The character \| has to be paired or escaped: {&private || |something between| or isolated: \| }}
+>> {Stop {& "translate 'stop' as a verb" might be a translator instruction or comment }}
+>> {Protect stuff in {^ph}<a>{^/ph}private use{^ph}</a>{^/ph}}
+>>```
```
eemeli, line 452: No need for double indent.

aphillips: Done.

spec/formatting.md
aphillips: Done.

spec/formatting.md
```diff
@@ -92,18 +92,18 @@ and different implementations MAY choose to perform different levels of resoluti
...
 - Else, if the _expression_ has a _function_ _annotation_,
   its resolved value is defined by _function resolution_.
-- Else, the _expression_ has a _reserved_ _annotation_.
-  If the _annotation_ uses a `private-start` character that the implementation supports,
-  its value is resolved according to the implementation's specification.
-  Else, an Unsupported Expression error is emitted and a fallback value is used as its value.
+- Else, if the _expression_ has a _private-use_ _annotation_,
+  its resolved value is defined according to the implementations's specification.
+- Else, the _expression_ has a _reserved_ _annotation_,
+  an Unsupported Expression error is emitted and a fallback value is used as its value.
```
eemeli, line 106: This seems like a run-on sentence, and the "it" could be specified a bit more explicitly. Yes, I realise that I'm in part correcting my own suggestion here.

```suggestion
- Else, the _expression_ has a _reserved_ _annotation_.
  An Unsupported Expression error is emitted and
  the expression resolves to a fallback value.
```

spec/syntax.md
```diff
@@ -19,7 +19,8 @@
    1. [Variants](#variants)
    1. [Patterns](#patterns)
    1. [Expressions](#expressions)
-      1. [Reserved Sequences](#reserved)
+      1. [Private-Use Sequences](#private-use)
+      2. [Reserved Sequences](#reserved)
```
eemeli, line 23: We might want to start using explicit `<a name="foo"></a>` anchors at some point, rather than having two different names for spec sections.

eemeli (approved):

PR #418, opened 2023-07-11
Fix styling to be consistent

aphillips:
Addresses #416

This commit is the first of several to address styling and formatting consistency issues in the specification.

* The 2119 keywords paragraph now quotes RFC 8174 verbatim.

PR #429, opened 2023-07-17
(EARLY REVIEW/SPECULATIVE) Refactoring syntax spec

aphillips:
In my effort to make the formatting correct, I found that the spec was repetitious and somewhat difficult to read. Here is a prototype refactor for early review.

This makes some normative changes, particularly it splits up `selectors` into named parts. 

Some sections could stand to be sub-sections. Have a look at it (preview it in the branch `aphillips-syntax-refactor` if that helps) and see if this is worth bringing to the larger group.

Note: the TOC has not been updated, so ignore it.

### DIFF
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -166,7 +166,7 @@ declaration = let s variable [s] "=" [s] expression
 
 ### Body
 
-The **_body_** of a message consists of either a _pattern_ or a _selector_.
+The **_body_** of a message consists of either a _pattern_ or a _match statement_.
 
 ### Pattern
 
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -192,7 +192,7 @@ various formats regardless of the container's whitespace trimming rules.
 > Example. In a Java `.properties` file, the message `hello` has exactly three spaces before and after
 > the word "Hello":
 > ```properties
-> hello = {  Hello  }
+> hello = {   Hello   }
 > ```
 
 ### Text
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -89,6 +89,8 @@ The syntax specification takes into account the following design restrictions:
 
 ## Messages and Syntax
 
+
+
 ### Messages
 
 A **_message_** is the complete template for a specific message formatting request.
@@ -126,6 +128,11 @@ An empty string is not a valid _message_.
 >let hello = new MessageFormat('{Hello, world!}')
 >hello.format()
 >```
+
+A _message_ satisfying all rules of the grammar is considered _well-formed_.
+
+Furthermore, a _well-formed_ _message_ is considered _valid_
+if it meets additional semantic requirements about its structure, defined below.
 
 ### Variable Declarations
 
@@ -177,6 +184,21 @@ A _message_ can only be considered _well-formed_ if the following requirements a
 selector = match_statment 1*(variant)
 ```
 
+>A message with a _selector_:
+>
+>```
+>match {$count :number}
+>when 1 {You have one notification.}
+>when * {You have {$count} notifications.}
+>```
+
+>A message with a _selector_ formatted on a single line:
+>
+>```
+>match {:platform} when windows {Settings} when * {Preferences}
+>```
+
+
 ### Match
 
 A **_match statement_** or just `match` is the portion of a _selector_ that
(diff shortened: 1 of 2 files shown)


eemeli:
Before starting to review this, I would like us to establish clearly whether we are continuing to apply [semantic line breaks](https://sembr.org/) to our markdown content, or follow some other line wrapping practice.

My preference would be to continue with semantic line breaks, as it seems that we persist in revisiting the text, but I'd really rather not spend more time adjusting line breaks in PRs, so would be fine with an alternative if it reaches consensus among others.

aphillips:
We're doing semantic line breaks. 

What I'm looking for here with the early review is a sense of whether this change is something to chase after at all. If this is worth pursuing, I'll go back and button up the text before you (and the rest of the WG) review it seriously at the detail level. Most of the text here was stolen from the pre-formatting-fix text, although some I stole from the on-going PR.

Is this better or should I focus on the other PR?

eemeli:
I would very much prefer to first get #418 and #419 merged before considering further large editorial changes to the spec. I would also find it easier to consider editorial and normative changes separately, if that's possible.

aphillips:
We'll get those merged first, since we've already expended effort to make the text and syntax there correct.

However... I'm looking for a reaction to a proposal: do you find this structure better than the current one? If not or if it's not significantly better, I won't bother with it.

stasm:
I actually like this a lot, thank you @aphillips!

Back when I originally wrote my spec proposal (#218) I struggled with the same repetition that you noticed. I wanted the _Overview & Examples_ section to come first, to showcase the proposal before specifying it, but it was difficult to explain the examples without defining the productions and the concepts. So I ended up with repeating some parts which didn't feel very "spec-like" to me, and which carried over to the spec on `main`.

To address @eemeli's concern, I still think it would be good to have some kind of _Overview_ before the _Messages and Syntax_ section. It doesn't have to be as extensive as it is today; in fact, it shouldn't try to showcase every possible feature of the syntax — I think we should expect other documents to serve that purpose (e.g. tutorials, getting-started articles, etc.). But I'd like to have a few examples up front to help visualize what MesageFormat is and what problem space it exists in.

Perhaps something similar to the examples in our [current README](https://github.com/unicode-org/message-format-wg/blob/0a8281ccfcedaa93ba534e79675db22ae8aea400/README.md)? I can try to come up with something over the weekend.

spec/syntax.md
```diff
@@ -87,285 +87,253 @@ The syntax specification takes into account the following design restrictions:
...
 
-A message satisfying all rules of the grammar is considered _well-formed_.
+A **_selector_** selects a specific _pattern_ from a list of available
+_variants_ in a _message_.
+_Selectors_ provide the ability for a _message_ to use a _pattern_ that 
+varies in content or form depending on values determined at runtime.
 
-Furthermore, a well-formed message is considered _valid_
-if it meets additional semantic requirements about its structure, defined below.
+A _selector_ consists of a _match_ statement followed by at least one _variant_.
```
stasm, line 228: I've always thought about this the other way around: expressions can either be placeholders or selectors, and the `match` statement is where you put multiple selectors into.

Also, I'd consider variants to be part of the `match` statement, rather than follow after it.

So in my mind it's (pseudo-ABNF):

```abnf
match-stmt = 'match' 1*selector 1*variant
```

rather than:

```abnf
selector = 'match' 1*expression 1*variant
```

spec/syntax.md
```diff
@@ -87,285 +87,253 @@ The syntax specification takes into account the following design restrictions:
...
-    when 1 masculine {{$userName} added a new photo to his album.}
-    when 1 feminine {{$userName} added a new photo to her album.}
-    when 1 * {{$userName} added a new photo to their album.}
-    when * masculine {{$userName} added {$photoCount} photos to his album.}
-    when * feminine {{$userName} added {$photoCount} photos to her album.}
-    when * * {{$userName} added {$photoCount} photos to their album.}
+### Body
 
-### Local Variables
+The **_body_** of a message consists of either a _pattern_ or a _selector_.
```
stasm, line 169: In line with my other comment about the `match` statement containing selectors and variant, I'd phrase this as:

```suggestion
The **_body_** of a message consists of either a _pattern_ or a _match statement_.
```

aphillips:
@stasm Thanks!

Note that README.md is already a part of the spec and I think demonstrates some of the cases. We could put more there.

spec/syntax.md
aphillips: I like your reformulation very much. It solves a lot of circular jargon problems surrounding what the `expression`s in `selectors` are.

eemeli (commented):
Commenting mostly to clear this from my review queue.

As discussed on this week's call, I think a prerequisite for the changes proposed here is moving the intro-to-MF2 parts to the README or to some other overview doc. I would also find it easier if the structural changes to the prose and the ABNF changes were not mixed into one set of changes, but I understand that separating them might not be possible.

aphillips:
I am closing this PR in favor of the new one.

@eemeli I did **not** include the ABNF changes in that PR. I will submit those in an independent PR. You can see my intentions in the prose.

I also did **not** include changes outside syntax.md for now.

PR #431, opened 2023-07-19
Add Missing Selector Annotation error

eemeli:
See #425 for context.

Adds this requirement to the spec, and a corresponding data model error:

```markdown
Each _selector_ MUST have an _annotation_,
```

Edit: Added "directly or indirectly" to the original proposal, to account for messages like:
```
let $one = {|The one| :func}
let $two = {$one}
match {$two}
when 1 {Value is one}
when * {Value is not one}
```

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -674,7 +674,7 @@ These are divided into the following categories:
     > when * 1 {Second is one}
     > ```
 
-  - **Missing Selector Annotation errors** occur when the message
+A **_Missing Selection Annotation error_** is an error that occurs when the _message_
     contains a _selector_ that does not have an _annotation_,
     or contains a _variable_ that does not directly or indirectly reference a _declaration_ with an _annotation_.
 
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -381,15 +381,15 @@ _This section is non-normative._
 
 #### Example 1
 
-Presuming a minimal implementation which only supports string values
-and matches keys by using string comparison,
+Presuming a minimal implementation which only supports `:string` annotation
+which matches keys by using string comparison,
 and a formatting context in which
 the variable reference `$foo` resolves to the string `'foo'` and
 the variable reference `$bar` resolves to the string `'bar'`,
 pattern selection proceeds as follows for this message:
 
 ```
-match {$foo} {$bar}
+match {$foo :string} {$bar :string}
 when bar bar {All bar}
 when foo foo {All foo}
 when * * {Otherwise}
@@ -420,7 +420,7 @@ Alternatively, with the same implementation and formatting context as in Example
 pattern selection would proceed as follows for this message:
 
 ```
-match {$foo} {$bar}
+match {$foo :string} {$bar :string}
 when * bar {Any and bar}
 when foo * {Foo and any}
 when foo bar {Foo and bar}
@@ -645,13 +645,13 @@ These are divided into the following categories:
     > Example invalid messages resulting in a Variant Key Mismatch error:
     >
     > ```
-    > match {$one}
+    > match {$one :func}
     > when 1 2 {Too many}
     > when * {Otherwise}
     > ```
     >
     > ```
-    > match {$one} {$two}
+    > match {$one :func} {$two :func}
     > when 1 2 {Two keys}
     > when * {Missing a key}
     > when * * {Otherwise}
@@ -663,15 +663,42 @@ These are divided into the following categories:
     > Example invalid messages resulting in a Missing Fallback Variant error:
     >
     > ```
-    > match {$one}
+    > match {$one :func}
     > when 1 {Value is one}
     > when 2 {Value is two}
     > ```
     >
     > ```
-    > match {$one} {$two}
+    > match {$one :func} {$two :func}
     > when 1 * {First is one}
     > when * 1 {Second is one}
+    > ```
+
+  - **Missing Selector Annotation errors** occur when the message
+    contains a _selector_ that does not have an _annotation_,
+    or contains a _variable_ that does not reference a _declaration_ with an _annotation_.
+
+    > Example invalid messages resulting in a Missing Selector Annotation error:
+    >
+    > ```
+    > match {$one}
+    > when 1 {Value is one}
+    > when * {Value is not one}
+    > ```
+    >
+    > ```
+    > let $one = {|The one|}
+    > match {$one}
+    > when 1 {Value is one}
+    > when * {Value is not one}
+    > ```
+    >
+    > ```
+    > let $one = {|The one| :func}
+    > let $two = {$one}
+    > match {$two}
+    > when 1 {Value is one}
+    > when * {Value is not one}
     > ```
 
   - **Duplicate Option Name errors** occur when the same _name_ 
@@ -703,7 +730,7 @@ These are divided into the following categories:
     > ```
     >
     > ```
-    > match {$var}
+    > match {$var :func}
     > when 1 {The value is one.}
     > when * {The value is not one.}
     > ```
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -678,7 +678,7 @@ These are divided into the following categories:
     contains a _selector_ that does not have an _annotation_,
     or contains a _variable_ that does not directly or indirectly reference a _declaration_ with an _annotation_.
 
-    > Example invalid messages resulting in a Missing Selector Annotation error:
+    > Example invalid messages resulting in a _Missing Selector Annotation error_:
     >
     > ```
     > match {$one}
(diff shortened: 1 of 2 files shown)


aphillips:
Meh. I don't think we should do this. I would rather do one of the following:

(a) Make `:select` (or `:equals` or whatever we call it) the default selector and normatively say that no other selectors can be inferred (other selectors must be explicitly declared). This ensures that processing is not programming language or runtime dependent, which helps translation tools do consistent stuff:

```
match {$two}
when 1 {two is a string with only U+0031 in it}
when * {two is some other string}
```

(b) *Syntactically* require the annotation, e.g.:

```abnf
switch = matcher *variant
matcher = match 1*selector
; annotation is not optional unlike in expression
selector = "{" [s] (operand s annotation)/(annotation) [s] "}"
```

This makes the selector function explicit and makes it a parse error not to say which one to use. It also means that you can't use a declaration for the selector (only as the value of a selector), which I agree is kind of odd.

---

I'll also call out: in our discussion of mutable vs. immutable variables, @mihnita's "13/42" example comes into play here. If we say that this evaluates as "13 and 42":

```
let $a = {13}
let $b = {$a}
let $a = 42
{{$a} and {$b}}
```

... but then say that a selector is late binding:

```
let $a = {$count :plural}
match {$a}
when 1 {exactly 1}
when one {one plural rule}
when * {other plural rule}
```

... that seems like an inconsistency in favor of "42 42"

eemeli:
> Meh. I don't think we should do this. I would rather do one of the following:
> 
> (a) Make `:select` (or `:equals` or whatever we call it) the default selector and normatively say that no other selectors can be inferred (other selectors must be explicitly declared). This ensures that processing is not programming language or runtime dependent, which helps translation tools do consistent stuff:
> 
> ```
> match {$two}
> when 1 {two is a string with only U+0031 in it}
> when * {two is anything else}
> ```

I don't think this kind of message is at all common, so I would find it odd to have it be the default behaviour. What is the benefit we get from not requiring a selector here?

> (b) _Syntactically_ require the annotation, e.g.:
> 
> ```abnf
> switch = matcher *variant
> matcher = match 1*selector
> ; annotation is not optional unlike in expression
> selector = "{" [s] (operand s annotation)/(annotation) [s] "}"
> ```
> 
> This makes the selector function explicit and makes it a parse error not to say which one to use. It also means that you can't use a declaration for the selector (only as the value of a selector), which I agree is kind of odd.

I don't like this. This would mean we'd need two different expression parsers, which seems really cumbersome. Requiring an annotation would also guide developers towards using different options bags for the selection vs. formatting.

> I'll also call out: in our discussion of mutable vs. immutable variables, @mihnita's "13/42" example comes into play here. If we say that this evaluates as "13 and 42":
> 
> ```
> let $a = {13}
> let $b = {$a}
> let $a = 42
> {{$a} and {$b}}
> ```
> 
> ... but then say that a selector is late binding:
> 
> ```
> let $a = {$count :plural}
> match {$a}
> when 1 {exactly 1}
> when one {one plural rule}
> when * {other plural rule}
> ```
> 
> ... that seems like an inconsistency in favor of "42 42"

The way that I would read this is that we first create a "plural" instance and assign it to `$a`. Then we use that as a selector. I don't really see how that qualifies as late binding?

spec/syntax.md
```diff
@@ -309,6 +309,8 @@ A _well-formed_ message is considered _valid_ if the following requirements are
 
 - The number of keys on each _variant_ MUST be equal to the number of _selectors_.
 - At least one _variant's_ keys MUST all be equal to the catch-all key (`*`).
+- Each _selector_ MUST have an _annotation_,
+  or contain a _variable_ that references a _declaration_ with an _annotation_.
```
stasm, line 313: Would this make the following message invalid?

```
let $foo = {$count :plural}
let $bar = {$foo}
let $baz = {$bar}
match {$baz} ...
```

eemeli: Yes.

stasm: Ah, I see your "and that's just silly" comment in the PR's description. While I agree that it's not something I'd expect to see in a real-life message, I think it's an important litmus test for how easy it is to build a mental model of how a message is evaluated.

Adding arbitrary rules to this effect ("references can have as many jumps as needed, unless used in a selector, when it can only be one jump to an expression with an annotation") is going to make MF2 harder to learn.

Furthermore, such rules don't really help, because as a translator, I can still do to bypass such checks:

```
let $foo = {$count :noop}
match {$foo} ...
```

Nor do they prevent us from errors when non-matching functions are used:

```
let $foo = {$count :datetime}
match {$foo} ...
```

eemeli:
> #425 has gotten so huge that it's difficult to see what exactly this PR is trying to address. Can you summarize what problem do you think we'd solve by adding this restriction?

That's fair. 😅

The main beneficiary of requiring annotations on selectors is translation tooling, which can then better handle locale-sensitive selectors such as plurals.

A secondary benefit is the avoidance of apparently-but-not-really valid messages, such as the one presented by @macchiati in https://github.com/unicode-org/message-format-wg/issues/425#issuecomment-1642959919:
```
match {$count}
when one {...}
when * {...}
```

spec/syntax.md
eemeli:
```suggestion
  or contain a _variable_ that directly or indirectly references a _declaration_ with an _annotation_.
```

spec/formatting.md
```diff
@@ -663,17 +663,44 @@ These are divided into the following categories:
...
     > ```
-    > match {$one} {$two}
+    > match {$one :func} {$two :func}
     > when 1 * {First is one}
     > when * 1 {Second is one}
     > ```
 
+  - **Missing Selector Annotation errors** occur when the message
+    contains a _selector_ that does not have an _annotation_,
+    or contains a _variable_ that does not reference a _declaration_ with an _annotation_.
```
eemeli, line 679:
```suggestion
    or contains a _variable_ that does not directly or indirectly reference a _declaration_ with an _annotation_.
```

eemeli:
@aphillips @macchiati @mihnita @stasm I've now updated this proposal following our discussion, to also account for transitive references to annotated expressions.

spec/formatting.md
```diff
@@ -663,17 +663,44 @@ These are divided into the following categories:
...
     > ```
     >
     > ```
-    > match {$one} {$two}
+    > match {$one :func} {$two :func}
     > when 1 * {First is one}
     > when * 1 {Second is one}
     > ```
 
+  - **Missing Selector Annotation errors** occur when the message
```
aphillips, line 677: Don't dig the formatting hole deeper. When you introduce the term, do so in the form:

```suggestion
A **_Missing Selection Annotation error_** is an error that occurs when the _message_
```

That is, the form is:

> A **_thing_** is {definition} with _terms_ italicized.

aphillips (changes requested):
Minor change requested.

spec/formatting.md
```diff
@@ -663,17 +663,44 @@ These are divided into the following categories:
...
+    > match {$one :func} {$two :func}
     > when 1 * {First is one}
     > when * 1 {Second is one}
     > ```
 
+A **_Missing Selection Annotation error_** is an error that occurs when the _message_
+    contains a _selector_ that does not have an _annotation_,
+    or contains a _variable_ that does not directly or indirectly reference a _declaration_ with an _annotation_.
+
+    > Example invalid messages resulting in a Missing Selector Annotation error:
```
aphillips, line 681: This says "Missing Selector Annotation error" but the definition says "Missing Selection Annotation error". Choose one.

Also, the name should be italicized here.

aphillips (changes requested):

aphillips:
Note: I filed #433 to track choosing a name for `:string`. I think we should merge this as reflecting consensus (assuming it does in tomorrow's call) and follow it up with a name change (if we decide to change the name) separately.

spec/formatting.md
```diff
@@ -663,17 +663,44 @@ These are divided into the following categories:
...
     > ```
     >
     > ```
-    > match {$one} {$two}
+    > match {$one :func} {$two :func}
     > when 1 * {First is one}
     > when * 1 {Second is one}
     > ```
 
+A **_Missing Selection Annotation error_** is an error that occurs when the _message_
```
eemeli, line 677:
```suggestion
  - A **_Missing Selector Annotation error_** is an error that occurs when the _message_
```

spec/formatting.md
eemeli:
```suggestion
    > Example invalid messages resulting in a _Missing Selector Annotation error_:
```

aphillips (approved):

stasm (approved):
LGTM after dropping the "at most one jump" requirement on annotations in local variables.

mihnita (approved):

PR #434, opened 2023-07-23
Fix registry.dtd to not use alternatives in function, formatSignature, matchSignature

stasm:
A spin-off from #420, in which @mihnita noticed that `registry.dtd` uses alternatives in a wrong way. For example, `<!ELEMENT formatSignature (input?|option*)>` means that `formatSignature` is allowed to have as children either: at most one `input` OR any number of `options`, but not both.

Instead, @mihnita suggested using sequences: `<!ELEMENT formatSignature (input?,option*)>`, which this PR does. Note that sequences require the children to appear in a specific order, which isn't something that's useful to us. However, I'm not aware of any way of lifting this requirement that also allows to enforce at most on `input` or exactly one `description`.


spec/registry.dtd
```diff
@@ -1,6 +1,6 @@
-<!ELEMENT registry (function*|pattern*)>
+<!ELEMENT registry (function|pattern)*>
```
stasm, line 1: In #420 @mihnita suggested changing this to:

```dtd
<!ELEMENT registry (function,pattern)*>
```

...but IIUC, this requires `function` and `pattern` to appear one after the other, in this specific order, i.e. interleaved. Instead, in this PR I'm changing the original `(function*|pattern*)` — which wrongly meant _any number of `function` OR any number of `pattern`, but never both_ — to `(function|pattern)*`.

stasm:
#420 was merged today, but before I rebase this PR on top of it, I'd like to make sure that we're OK with using DTD sequences:

```dtd
<!ELEMENT function (description,(formatSignature|matchSignature)+)>
<!ELEMENT formatSignature (input?,option*)>
<!ELEMENT matchSignature (input?,option*,match*)>
```

Using sequences here means that, for instance, the following `formatSignature` would not be valid, because `option` comes before `input`:

```xml
<formatSignature>
    <option name="dateStyle" values="full long medium short"/>
    <input pattern="iso8601"/>
</formatSignature>
```

aphillips (approved):

PR #438, opened 2023-07-24
 Add details to "Missing Selector Annotation error" section

catamorphism:
I found this text to be confusing: "...or contains a _variable_ that does not directly or indirectly reference a _declaration_ with an _annotation_". Read literally, this implies _all_ variables must be annotated or indirectly reference expressions that have annotations. I tried to clarify that this only refers to variables used in a selector context.

Also, one of the "invalid" examples was actually valid, according to the "directly or indirectly" rule, because `$two` indirectly refers to `{|The one| : func}`, which has an annotation. I added a "valid examples" section since I wasn't sure how else to explain the rule about indirect references.

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -677,8 +677,14 @@ These are divided into the following categories:
     > ```
 
   - A **_Missing Selector Annotation error_** is an error that occurs when the _message_
-    contains a _selector_ that does not directly or indirectly reference
-    an _expression_ with an _annotation_.
+    contains a _selector_ without an _annotation_.
+
+    An implementation MUST NOT emit this error for a _selector_ _expression_
+    that would have an _annotation_
+    if each local _variable_ in it were replaced
+    with the _expression_ of its binding _declaration_,
+    repeating this process until the _expression_
+    contains no local _variable_s.
 
     > Example invalid messages resulting in a _Missing Selector Annotation error_:
     >
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -695,24 +695,8 @@ These are divided into the following categories:
     > when * {Value is not one}
     > ```
     >
-    > Example valid messages that do not result in a _Missing Selector Annotation error_:
-    >
-    >
-    > ```
-    > match {$two :func}
-    > when 1 {Value is one}
-    > when * {Value is not one}
-    > ```
-    >
-    > ```
-    > let $one = {|The one| :func}
-    > let $two = {$one}
-    > match {$two}
-    > when 1 {Value is one}
-    > when * {Value is not one}
-    > ```
-    
-  - **Duplicate Option Name errors** occur when the same _name_ 
+
+- **Duplicate Option Name errors** occur when the same _name_ 
     appears on the left-hand side
     of more than one _option_ in the same _expression_.
 
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -677,8 +677,8 @@ These are divided into the following categories:
     > ```
 
   - A **_Missing Selector Annotation error_** is an error that occurs when the _message_
-    contains a _selector_ that does not have an _annotation_,
-    or contains a _variable_ that does not directly or indirectly reference a _declaration_ with an _annotation_.
+    contains a _selector_ that does not directly or indirectly reference
+    an _expression_ with an _annotation_.
 
     > Example invalid messages resulting in a _Missing Selector Annotation error_:
     >
@@ -695,6 +695,15 @@ These are divided into the following categories:
     > when * {Value is not one}
     > ```
     >
+    > Example **valid** messages that do _not_ result in a _Missing Selector Annotation error_:
+    >
+    >
+    > ```
+    > match {$two :func}
+    > when 1 {Value is one}
+    > when * {Value is not one}
+    > ```
+    >
     > ```
     > let $one = {|The one| :func}
     > let $two = {$one}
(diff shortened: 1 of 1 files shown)


spec/formatting.md
```diff
@@ -695,14 +695,26 @@ These are divided into the following categories:
...
     > let $one = {|The one| :func}
     > let $two = {$one}
     > match {$two}
     > when 1 {Value is one}
     > when * {Value is not one}
     > ```
-
+    >
+    > The second message is valid because `$two` refers to the expression `$one`,
+    > and `$one` refers to the expression ` {|The one| :func}`, which has an annotation.
```
eemeli, line 716: None of the other examples are explained, so this shouldn't be either.

```suggestion
```

spec/formatting.md
```diff
@@ -695,14 +695,26 @@ These are divided into the following categories:
     > when * {Value is not one}
     > ```
     >
+    > Example **valid** messages that do _not_ result in a _Missing Selector Annotation error_:
```
eemeli, line 698: We've assigned special meaning to bold/italic text styling, so we shouldn't use them for _emphasis_.

```suggestion
    > Example valid messages that do not result in a _Missing Selector Annotation error_:
```

eemeli (approved):

spec/formatting.md
```diff
@@ -677,8 +677,8 @@ These are divided into the following categories:
     > ```
 
   - A **_Missing Selector Annotation error_** is an error that occurs when the _message_
-    contains a _selector_ that does not have an _annotation_,
-    or contains a _variable_ that does not directly or indirectly reference a _declaration_ with an _annotation_.
+    contains a _selector_ that does not directly or indirectly reference
+    an _expression_ with an _annotation_.
 
     > Example invalid messages resulting in a _Missing Selector Annotation error_:
```
aphillips, line 683: Making the wording flow a bit better

```suggestion
> **Example**.
> Here are some invalid _messages_ that would result in a _Missing Selector Annotation error_:

File: spec/formatting.md [hunk 4]
```diff
@@ -695,14 +695,23 @@ These are divided into the following categories:
     > when * {Value is not one}
     > ```
     >
+    > Example valid messages that do not result in a _Missing Selector Annotation error_:
```
aphillips, line 698: As above...

```suggestion
> By contrast, here are some valid messages that do not result in a _Missing Selector Annotation error_:
```

Is it that helpful to show valid messages? Shouldn't we focus on triggering behavior?

aphillips (changes requested):

spec/formatting.md
catamorphism: This text is modeled on the existing text for the other kinds of data model errors. Perhaps a separate PR could add this change in a way that's consistent for all the error descriptions?

spec/formatting.md
catamorphism: I'm not sure how to clarify what "does _not_ directly or indirectly reference an expression with an annotation" (emphasis added) means without contrasting it with a message that _does_ indirectly reference an expression with an annotation. Do you have any suggestions for how to make that language clearer without using a valid example?

catamorphism: I thought of another way to phrase the condition for validity that didn't require a valid example to explain -- let me know what you think.

spec/formatting.md
```diff
@@ -677,8 +677,14 @@ These are divided into the following categories:
...
-    contains a _selector_ that does not have an _annotation_,
-    or contains a _variable_ that does not directly or indirectly reference a _declaration_ with an _annotation_.
+    contains a _selector_ without an _annotation_.
+
+    An implementation MUST NOT emit this error for a _selector_ _expression_
+    that would have an _annotation_
+    if each local _variable_ in it were replaced
+    with the _expression_ of its binding _declaration_,
+    repeating this process until the _expression_
+    contains no local _variable_s.
```
eemeli, line 687: This approach is a bit problematic because option values cannot be expressions, and this is clearly including their replacement as well.

@stasm This is an example of a situation where aligning our data model closely with our syntax is making it more difficult to speak about it. If we went the other way than proposed in #436 and e.g. defined a _variable expression_ in the data model, we could use that term here.

spec/formatting.md
aphillips: I haven't looked at the proposed text yet, but I will note: examples are good and even if we think of a better way of explaining something, we should be looking for opportunities to incorporate examples. My above comment might go towards some style guide activity, and, in a more general way, start to build some infrastructure to support conversion of our spec into a proper HTML document (I'll create a task for myself to capture).

spec/formatting.md
catamorphism: @eemeli I see your point.

The validity property is easy to express in terms of types:

```
The type of a local _variable_ is the type of the _expression_ in its _declaration_.
The type of an external variable is `Unannotated`.

The type of an _expression_ is:
* `Annotated` if the _expression_ has an _annotation_.
* `Annotated` if the _operand_ of the expression is a _variable_ whose type is `Annotated`.
* `Unannotated` otherwise.

A `selector` must have type `Annotated`.
```

I'm not sure how this language would be received, but we are talking about a static dataflow property, whether or not we call it that :)

I also agree that it would be easier to write an unambiguous specification if the object language was the language described by the abstract syntax, rather than the concrete syntax.

aphillips:
I've added this to the 2023-12-11 agenda. I think we either (a) need to reject this change and let implementers do what we mean or (b) clarify how to say what we mean.

aphillips:
Will reject this if no supporting comments by call time tomorrow.

aphillips:
See above comment about resolution

PR #439, opened 2023-07-24
Add JSON Schema & XML DTD definitions of message data model

eemeli:
PR #393 left out the JSON Schema definition of the data model, which it originally referred to as its canonical description. This PR re-presents that, but now together with a synonymous `message.dtd` document. Unlike before, these definitions are here referenced not as canonical, but as equivalent to the TypeScript definitions.

The data model documents are also moved to `spec/data-model/`, to increase their distance slightly from the rest of the spec.

The intent with including these definitions is to increase the utility of the data model for applications that do have a need for representing messages as structured data rather than syntax. By providing ready definitions, we make it easier for such applications to naturally align for compatibility, without imposing any requirements that all implementations must comply with.


aphillips (approved):

ryzokuken (approved):

eemeli:
@aphillips I resolved the conflicts on this, and this should now be good to merge. I'll update #444 after that's done, so that it too can be merged.

PR #441, opened 2023-07-27
Specification refactor "v2"

aphillips:
This is the second attempt at the specification refactoring. I started with all of our PRs merged Monday.

This is a much more significant reformulation. It is intended to be read as prose as well as specifying individual productions. 

I organized the sections "top down", with some sub-sections containing others, e.g.:

> # Message
> ## Declarations
> ## Body
> ### Pattern
> ### Matcher
> #### Selector
> #### Variant
> ##### Key

There is then a set of productions about Expressions and then finally the remaining terms, such as keywords, whitespace, and such. You might wish to view this rendered first rather than diving into the diff specifically.

---

I have followed commentary about separating "abstract" vs. "concrete" (e.g. in #440). This proposal embraces the use of ABNF constructs whose purpose is "abstract". We don't have to do this. We can define terms in the spec that do not appear in the ABNF. E.g.:

> A <dfn>**_placeholder_**</dfn> refers to any _expression_ that appears inside of a _pattern_.

(and then use `expression` in the ABNF exclusively), a la:

```abnf
pattern = "{" 1* (text / expression) "}"
```

---

This PR does not include a repaired TOC and I have not yet removed the requirements to elsewhere. I moved some examples to the end, with an intention of moving them out altogether. 

Have a read. 📚

### DIFF
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -151,7 +151,7 @@ A **_declaration_** binds a _variable_ identifier to the value of an _expression
 This local variable can then be used in other _expressions_ within the same _message_.
 _Declarations_ are optional: many messages will not contain any _declarations_.
 
-```
+```abnf
 declaration = let s variable [s] "=" [s] expression
 ```
 
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -264,7 +264,7 @@ matcher = match 1*(selector) 1*(variant)
 ### Selector
 
 A **_selector_** is an _expression_ that ranks or excludes the
- _keys_ from the available _variants_ in a _message_.
+_variants_ based on the value of its corresponding _key_ in each _variant_.
 The combination of _selectors_ in a _matcher_ thus determines 
 which _pattern_ will be used during formatting.
 
@@ -305,8 +305,8 @@ be followed by a sequence of _keys_,
 and terminate with a valid _pattern_.
 The number of _keys_ in each _variant_ MUST match the number of _selectors_ in the _matcher_.
 
-_Keys_ are separated from the keyword `when` and from each other by whitespace.
-Whitespace is not required between the last _key_ and the _pattern_.
+Each _key_ is separated from the keyword `when` and from each other by whitespace.
+Whitespace is permitted but not required between the last _key_ and the _pattern_.
 
 ```abnf
 variant = when 1*(s key) [s] pattern
@@ -322,7 +322,7 @@ A _key_ can be either a _literal_ value or the "catch-all" key `*`.
 The **_catch-all key_** is a special key, represented by `*`,
 that matches all values for a given _selector_.
 
-### Expression
+## Expressions
 
 An **_expression_** is a part of a _message_ that will be determined
 during the _message_'s formatting.
@@ -335,14 +335,35 @@ an _annotation_,
 or an _operand_ followed by an _annotation_;
 or it can consist of a _private-use_ or _reserved_ sequence.
 
-A _selector_ or a _placeholder_ are both _expressions_.
-The value portion of a _declaration_ is also an expression.
-
 ```abnf
 expression = "{" [s] ((operand [s annotation]) / annotation) [s] "}"
 operand = literal / variable
 annotation = (function *(s option)) / private-use / reserved
 ```
+
+There are several types of _expression_ that can appear in a _message_.
+All _expressions_ share a common syntax. The types of _expression_ are:
+1. The value of a _declaration_
+2. A _selector_
+3. A _placeholder_ in a _pattern_
+
+> Examples of different types of _expression_
+>
+> Declarations:
+> ```
+> let $x = {|This is an expression|}
+> let $y = {$operand :function option=operand}
+> ```
+> Selectors:
+> ```
+> match {$selector :functionRequired}
+> ```
+> Placeholders:
+> ```
+> {This contains an {|expression|}}
+> {This references an {$operand}}
+> {This expression calls a {$operand :function with=options}}
+> ```
 
 ### Operand
 
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -423,8 +423,8 @@ No other positional arguments are possible.
 _Functions_ can be _standalone_, or can be an _opening element_ or _closing element_.
 
 A **_<dfn>standalone</dfn>_** _function_ is not expected to be paired with another _function_.
-An **_<dfn>opening element</dfn>_** is a _function_ that SHOULD be paired with a _closing function_.
-A **_<dfn>closing element</dfn>_** is a _function_ that SHOULD be paired with an _opening function_.
+An **_<dfn>opening element</dfn>_** is a _function_ that SHOULD be paired with a _closing element_.
+A **_<dfn>closing element</dfn>_** is a _function_ that SHOULD be paired with an _opening element_.
 
 An _opening element_ MAY be present in a message without a corresponding _closing element_,
 and vice versa.
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -451,7 +451,8 @@ _Options_ are not required.
 
 ##### Options
 
-An **_<dfn>option</dfn>_** is a key-value pair containing arguments passed to a _function_.
+An **_<dfn>option</dfn>_** is a key-value pair
+containing a named argument that is passed to a _function_.
 
 An _option_ has a _name_ and a _value_.
 The _name_ is separated from the _value_ by an U+003D EQUALS SIGN `=` along with
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -653,8 +653,8 @@ name-char  = name-start / DIGIT / "-" / "." / ":"
 
 > **Note**
 > _External variables_ can be passed in that are not valid _names_.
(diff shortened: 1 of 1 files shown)


spec/syntax.md
```diff
@@ -88,127 +88,212 @@ The syntax specification takes into account the following design restrictions:
...
->
->```
->match {$count :number}
->when 1 {You have one notification.}
->when * {You have {$count} notifications.}
->```
+A _message_ can only be considered _valid_ if the following requirements are
+satisfied:
+* The number of _keys_ on each _variant_ MUST be equal to the number of _selectors_.
+* At least one _variant_ MUST exist whose _keys_ are all equal to the "catch-all" key `*`.
```
eemeli, line 263: Running Prettier on this will replace the `*` with `-` and add a space before the list. Might be easier to just use its styling for markdown?

spec/syntax.md
```diff
@@ -229,272 +314,145 @@ A message can have multiple selectors.
...
+An _expression_ can contain an _operand_, 
+an _annotation_, 
+or an _operand_ followed by an _annotation_;
+or it can consist of a _private-use_ or _reserved_ sequence.
 
 ```abnf
-selectors = match 1*([s] expression)
+expression = "{" [s] ((operand [s annotation]) / annotation) [s] "}"
+operand = literal / variable
+annotation = (function *(s option)) / private-use / reserved
```
eemeli, line 358: This is defined in its own section.

```suggestion
```

spec/syntax.md
```diff
@@ -229,272 +314,145 @@ A message can have multiple selectors.
...
->
-> A _pattern_ containing three spaces:
-> ```
-> {   }
-> ```
+### Function
 
-### Expressions
+A **_function_** is a used to evaluate, format, select, or otherwise process an _operand_,
+or, if lacking an _operand_, its _options_.
```
eemeli, line 407: This implies that a function with an operand does not care about its options, and that a function without an operand requires options. The previous reference to processing its annotation was really clumsy, but less incorrect.

This sentence should probably be completely refactored.

spec/syntax.md
```diff
@@ -229,272 +314,145 @@ A message can have multiple selectors.
...
+A _function_ accepts only an _operand_ as a positional argument.
 
-An _expression_ can appear as a local variable value, as a _selector_, and within a _pattern_.
-The contents of each _expression_ MUST start with an _operand_ or an _annotation_.
-An _expression_ MUST NOT be empty.
+A _function_ consists of a prefix sigil followed by a _name_.
+The following sigils are used for _functions_:
+* `:` for standalone content
+* `+` for starting or opening _expressions_
+* `-` for ending or closing _expressions_
```
eemeli, line 414: Further down you define _opening element_ and _closing element_. Should those terms be used here as well?

spec/syntax.md
```diff
@@ -672,13 +611,17 @@ name-char  = name-start / DIGIT / "-" / "." / ":"
            / %xB7 / %x300-36F / %x203F-2040
 ```
 
+> **Note**
+> _External variables_ names are constrained by this namespace.
```
eemeli, line 615: Strictly speaking not true. Anything not fitting _name_ just won't be matched by any variable in the message.

spec/syntax.md
```diff
@@ -704,3 +647,80 @@ s = 1*( SP / HTAB / CR / LF )
 The grammar is formally defined in [`message.abnf`](./message.abnf)
 using the ABNF notation,
 as specified by [RFC 5234](https://datatracker.ietf.org/doc/html/rfc5234).
+
+
+
+### Complex Messages
```
eemeli, line 653: Just delete this section: It's not meant to be here, and it's in the git history.

spec/syntax.md
```diff
@@ -704,3 +647,80 @@ s = 1*( SP / HTAB / CR / LF )
...
+>when male * {{$hostName} invites {$guestName} and {$guestsOther} other people to his party.}
+>
+>when * 0 {{$hostName} does not give a party.}
+>when * 1 {{$hostName} invites {$guestName} to their party.}
+>when * 2 {{$hostName} invites {$guestName} and one other person to their party.}
+>when * * {{$hostName} invites {$guestName} and {$guestsOther} other people to their party.}
+>```
+
+
+> Expression examples:
```
eemeli, line 684: Did you have a place in mind for these expression examples, and/or a reason why not to include them in this doc?

eemeli (commented):
This reads much better than the previous update. I found it easier to evaluate as a whole, here: https://github.com/unicode-org/message-format-wg/blob/aphillips-specification-refactor/spec/syntax.md

Comments below from a first pass; I may have more thoughts on this later. Not sure if I noticed all the small changes that are applied here.

spec/syntax.md
```diff
@@ -88,127 +88,212 @@ The syntax specification takes into account the following design restrictions:
...
+### Well-formed vs. Valid Messages
+
+A _message_ is **_well-formed_** if it satisfies all the rules of the grammar.
+
+A _message_ is **_valid_** if it is _well-formed_ and **also** meets the additional content restrictions
+and semantic requirements about its structure defined below.
+
+## The Message
+
+A <dfn>**_message_**</dfn> is the complete template for a specific message formatting request.
```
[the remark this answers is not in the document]
aphillips: Yes. I'm starting to think about how we script the markdown into a formal specification. Placing `dfn` tags around definitions (and not _just_ relying on styling) will make that easier. (We can rely on styling for references to term definitions, of course)

My intention is to go back and add `dfn` tags around the rest of the definitions.

spec/syntax.md
aphillips: Good callout and one for the style guide.

spec/syntax.md
aphillips: That's correct. I ran into this issue when drafting this section and it occurred to me that we needed to deal with the fact that `name` is constrained but callers will not be constrained. I'll refactor this sentence to be correct, but it is the formatting spec that needs instructions on handling of external variables, which I recorded in #442.

spec/syntax.md
aphillips: Along with "complex messages", I just pushed them out of the way for the nonce. I'm reluctant to delete them without looking more closely at the `expression` section. I'll refactor these into the text.

spec/syntax.md
aphillips: Note that this sentence doesn't make sense as written. Refactored this section.

eemeli (commented):
1. We should drop the ToC. It requires unnecessary manual maintenance, and it's replicating built-in functionality that GitHub provides for rendered markdown via the outline button at the far right of the Preview/Code/Blame toolbar.

2. I think adding the `<dfn>` tags is jumping the gun. We do ultimately need to transform our spec format to whatever is required for the UTS/UTR format that we'll be publishing it in, but it'll be just as easy for automation to then find the definitions as `<b><i>...</i></b>` compared to `<b><i><dfn>...</dfn></i></b>`. But if adding it, see line comments fixing a few mistakes.

spec/syntax.md
```diff
@@ -393,30 +394,80 @@ operand    = literal / variable
...
+A _function_'s entry in the _function registry_ will define 
+whether the _function_ is a _selector_ or formatter (or both),
+whether an _operand_ is required, 
+what form the values of an _operand_ can take,
+what _options_ and _option_ values are valid, 
+and what outputs might result. 
+See [function registry](./) for more information.
+
+A _function_ MAY accept an _operand_.
+No other positional arguments are possible.
```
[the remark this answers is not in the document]
aphillips: > If we say "accept", doesn't that introduce a potential dependency between the function syntax definition and the function registry?

Not really? It introduces a dependency between a given function's implementation and the appearance of an operand.

This text (the original and either your or my formulation) speaks to what a function can do (it is not required to accept an operand, even if one is provided). The second sentence is redundant, but was preserved from the existing document.

I think either version is subtly misleading, since the actual state of affairs is more like: a function can appear in an annotation, which may or may not follow an operand in an expression--and the operand serves as input to the function. In the syntax, `function` is not actually the functionality: the `function` is just the **name** of the functionality.

So maybe we should deal with this in `annotation`, for example:

> An _annotation_ can appear in an _expression_ by itself or following a single _operand_.
> When following an _operand_, the _operand_ serves as input to the _annotation_.

Then we should eliminate discussion here of operands.

spec/syntax.md
```diff
@@ -393,30 +394,80 @@ operand    = literal / variable
...
+
+_Functions_ can be _standalone_, or can be an _opening element_ or _closing element_.
+
+A **_<dfn>standalone</dfn>_** _function_ is not expected to be paired with another _function_.
+An **_<dfn>opening element</dfn>_** is a _function_ that SHOULD be paired with a _closing function_.
+A **_<dfn>closing element</dfn>_** is a _function_ that SHOULD be paired with an _opening function_.
```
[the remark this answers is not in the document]
aphillips: Missed this one. Thanks.

spec/syntax.md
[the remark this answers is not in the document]
eemeli: The annotation text sounds like a good approach; I'd much rather do that than argue about semantics.

PR #443, opened 2023-07-29
Drop the Literal quoted flag from the data model

eemeli:
At the moment, the data model includes a boolean `quoted` property on the `Literal` construct. This should be dropped, as it's not meant to effect anything during the runtime.

As the data model is extensible by implementations, this does allow for an implementation to add the field back in as a private extension, should it have a need to track this information.

### DIFF
--- a/spec/data-model.md
+++ b/spec/data-model.md
@@ -101,8 +101,9 @@ The `Literal` and `VariableRef` correspond to the the _literal_ and _variable_ s
 When they are used as the `body` of an `Expression`,
 they represent _expression_ values with no _annotation_.
 
-An _unquoted_ value is represented by a `Literal` with `quoted: false`,
-while a _quoted_ value would have `quoted: true`.
+Both _quoted_ and _unquoted_ values are represented by `Literal`,
+as the use or lack of quotation is a presentation detail
+which has no effect on the meaning of the _literal_.
 The `value` of `Literal` is the "cooked" value (i.e. escape sequences are processed).
 
 In a `VariableRef`, the `name` does not include the initial `$` of the _variable_.
@@ -110,7 +111,6 @@ In a `VariableRef`, the `name` does not include the initial `$` of the _variable
 ```ts
 interface Literal {
   type: 'literal'
-  quoted: boolean
   value: string
 }
 
--- a/spec/data-model.md
+++ b/spec/data-model.md
@@ -101,9 +101,8 @@ The `Literal` and `VariableRef` correspond to the the _literal_ and _variable_ s
 When they are used as the `body` of an `Expression`,
 they represent _expression_ values with no _annotation_.
 
-Both _quoted_ and _unquoted_ values are represented by `Literal`,
-as the use or lack of quotation is a presentation detail
-which has no effect on the meaning of the _literal_.
+`Literal` represents all literal values, both _quoted_ and _unquoted_.
+The presence or absence of quotes is not preserved by the data model.
 The `value` of `Literal` is the "cooked" value (i.e. escape sequences are processed).
 
 In a `VariableRef`, the `name` does not include the initial `$` of the _variable_.


spec/data-model.md
```diff
@@ -101,16 +101,16 @@ The `Literal` and `VariableRef` correspond to the the _literal_ and _variable_ s
 When they are used as the `body` of an `Expression`,
 they represent _expression_ values with no _annotation_.
 
-An _unquoted_ value is represented by a `Literal` with `quoted: false`,
-while a _quoted_ value would have `quoted: true`.
+Both _quoted_ and _unquoted_ values are represented by `Literal`,
+as the use or lack of quotation is a presentation detail
+which has no effect on the meaning of the _literal_.
```
aphillips, line 106: Perhaps:

```suggestion
`Literal` represents all literal values, both _quoted_ and _unquoted_. 
The presence or absence of quotes is not preserved by the data model
because the quote marks have no meaning.
```

spec/data-model.md
```diff
@@ -101,16 +101,16 @@ The `Literal` and `VariableRef` correspond to the the _literal_ and _variable_ s
...
 The `value` of `Literal` is the "cooked" value (i.e. escape sequences are processed).
```
aphillips, line 107: I would avoid the jargon-ish "cooked" here and I would note the non-inclusion of the quotes (where they exist and as you did for reserved's sigils elsewhere)

```suggestion
The `value` of `Literal` does not include surrounding quotes (where present)
and replaces `quoted-escape` sequences with the unescaped character.
```

aphillips (changes requested):
Good start. Some wording tweaks suggested.

spec/data-model.md
eemeli: With your suggested change, it's unclear to what extent the "have no meaning" phrase applies, which creates some confusion as the marks are a significant presentation detail in the syntax. I do like "presence or absence of quotes" though; that's more appropriately passive than what I had before.

spec/data-model.md
eemeli: I'd like for this to be considered in a separate issue or PR, for two reasons:
1. The data model uses the "raw" and "cooked" terms also with respect to `Text` and `Reserved`, which ought to be updated simultaneously. I would rather keep that outside the scope of this rather focused PR.
2. At the moment, the data model is described and explained via equivalences with the MF2 syntax. If there is a desire to describe it as an explicit result of parsing the syntax, as suggested by the term "replaces" here, that's a much bigger change that ought to be accompanied by some additional documentation.

spec/data-model.md
aphillips: Just full stop after data model then?

```suggestion
`Literal` represents all literal values, both _quoted_ and _unquoted_.
The presence or absence of quotes is not preserved by the data model.
```

spec/data-model.md
aphillips: If "raw" and "cooked" are terms, define them has terms (in a terminology section that needs to be added) and link them.

A different way of saying this would be to go more Unicode-like:

```suggestion
The `value` of `Literal` is the code point sequence contained by the _literal_,
with external syntax (such as quotes) removed
and escape sequences resolved to the characters that they represent.
```

This would apply to any representation, not just MF2. For example, a JS string would replace `\u20ac` notation with `€` in the `Literal`.

> If there is a desire to describe it as an explicit result of parsing the syntax, as suggested by the term "replaces" here, that's a much bigger change that ought to be accompanied by some additional documentation.

I think we should stipulate that the data model representation can round-trip any MF2 string without loss of information,  although doing so would canonicalize the representation, such as syntax (non-literal) whitespace and the presence or absence of quotes around literals such that the resulting round-trip string might not be a character-by-character match to the original.

ryzokuken (approved):
In the absence of this flag in the data model, do we still need to make this explicit distinction between "quoted" and "unquoted" values ? 

> `Literal` represents all literal values, both _quoted_ and _unquoted_.
> The presence or absence of quotes is not preserved by the data model.

If this is implied by the absence of the flag, then perhaps it's more confusing to leave it in as opposed to just dropping these lines.

eemeli:
> In the absence of this flag in the data model, do we still need to make this explicit distinction between "quoted" and "unquoted" values ?

I think it's good to include, to clarify that the mapping of these potentially separately representable syntax rules into a single data model interface is wholly intentional.

PR #444, opened 2023-07-29
Clarify that Reserved may also represent private-use in data model

eemeli:
The data model `Reserved` interface may also be used for private-use annotations that are not supported by the implementation. This PR adds text clarifying that to be the case, and explicitly mentions that private-use syntax that is supported by the implementation may use a different interface.

The `source` definition is also corrected to _not_ include the starting sigil. This was the only point in the data model where a part of the syntax was showing up twice, without a really good reason why.

### DIFF
--- a/spec/data-model.md
+++ b/spec/data-model.md
@@ -145,16 +145,21 @@ interface Option {
 }
 ```
 
-A `Reserved` represents an _expression_ with a _reserved_ _annotation_.
-The `sigil` corresponds to the starting sigil of the _reserved_.
+A `Reserved` represents an _expression_ with a _reserved_ or _private-use_ _annotation_.
+The `sigil` corresponds to the starting sigil of the _annotation_.
 The `source` is the "raw" value (i.e. escape sequences are not processed)
-and includes the starting `sigil`.
+and does not include the starting `sigil`.
 
 Implementations MUST NOT rely on the set of `sigil` values remaining constant,
 as future versions of this specification MAY assign other meanings to such sigils.
 
 If the _expression_ includes a _literal_ or _variable_ before the _annotation_,
 it is included as the `operand`.
+
+When parsing the syntax of a _message_ that includes a _private-use_ _annotation_
+supported by the implementation,
+the implemenation MAY represent it in the data model using a different interface
+as appropriate for the semantics and meaning that it attaches to that _annotation_.
 
 ```ts
 interface Reserved {
--- a/spec/data-model.md
+++ b/spec/data-model.md
@@ -150,8 +150,10 @@ The `sigil` corresponds to the starting sigil of the _annotation_.
 The `source` is the "raw" value (i.e. escape sequences are not processed)
 and does not include the starting `sigil`.
 
-Implementations MUST NOT rely on the set of `sigil` values remaining constant,
-as future versions of this specification MAY assign other meanings to such sigils.
+> **Note**
+> Implementers should be aware that future versions of this specification
+> might assign meaning to _reserved_ `sigil` values
+> and add new interfaces to this data model.
 
 If the _expression_ includes a _literal_ or _variable_ before the _annotation_,
 it is included as the `operand`.
--- a/spec/data-model/README.md
+++ b/spec/data-model/README.md
@@ -80,8 +80,7 @@ The `value` of `Text` is the "cooked" value (i.e. escape sequences are processed
 
 Implementations MUST NOT rely on the set of `Expression` `body` values being exhaustive,
 as future versions of this specification MAY define additional expressions.
-If encountering a `body` with an unrecognised value,
-an implementation SHOULD treat it as it would an `Unsupported` value.
+A `body` with an unrecognized value SHOULD be treated as an `Unsupported` value.
 
 ```ts
 interface Pattern {
--- a/spec/data-model/README.md
+++ b/spec/data-model/README.md
@@ -81,7 +81,7 @@ The `value` of `Text` is the "cooked" value (i.e. escape sequences are processed
 Implementations MUST NOT rely on the set of `Expression` `body` values being exhaustive,
 as future versions of this specification MAY define additional expressions.
 If encountering a `body` with an unrecognised value,
-an implementation SHOULD treat it as it would a `Reserved` value.
+an implementation SHOULD treat it as it would an `Unsupported` value.
 
 ```ts
 interface Pattern {
@@ -95,7 +95,7 @@ interface Text {
 
 interface Expression {
   type: 'expression'
-  body: Literal | VariableRef | FunctionRef | Reserved
+  body: Literal | VariableRef | FunctionRef | Unsupported
 }
 ```
 
@@ -148,7 +148,7 @@ interface Option {
 }
 ```
 
-A `Reserved` represents an _expression_ with a _reserved_ or _private-use_ _annotation_.
+An `Unsupported` represents an _expression_ with a _reserved_ or _private-use_ _annotation_.
 The `sigil` corresponds to the starting sigil of the _annotation_.
 The `source` is the "raw" value (i.e. escape sequences are not processed)
 and does not include the starting `sigil`.
@@ -167,8 +167,8 @@ the implemenation MAY represent it in the data model using a different interface
 as appropriate for the semantics and meaning that it attaches to that _annotation_.
 
 ```ts
-interface Reserved {
-  type: 'reserved'
+interface Unsupported {
+  type: 'unsupported'
   sigil: '!' | '@' | '#' | '%' | '^' | '&' | '*' | '<' | '>' | '/' | '?' | '~'
   source: string
   operand?: Literal | VariableRef
--- a/spec/data-model/README.md
+++ b/spec/data-model/README.md
@@ -148,23 +148,27 @@ interface Option {
 }
 ```
 
-An `Unsupported` represents an _expression_ with a _reserved_ or _private-use_ _annotation_.
+An `Unsupported` represents an _expression_ with a
+_reserved_ _annotation_ or a _private-use_ _annotation_ not supported
+by the implementation.
 The `sigil` corresponds to the starting sigil of the _annotation_.
 The `source` is the "raw" value (i.e. escape sequences are not processed)
 and does not include the starting `sigil`.
 
 > **Note**
-> Implementers should be aware that future versions of this specification
-> might assign meaning to _reserved_ `sigil` values
-> and add new interfaces to this data model.
+> Be aware that future versions of this specification
+> might assign meaning to _reserved_ `sigil` values.
+> This would result in new interfaces being added to
+> this data model.
 
 If the _expression_ includes a _literal_ or _variable_ before the _annotation_,
 it is included as the `operand`.
 
 When parsing the syntax of a _message_ that includes a _private-use_ _annotation_
 supported by the implementation,
-the implemenation MAY represent it in the data model using a different interface
-as appropriate for the semantics and meaning that it attaches to that _annotation_.
+the implementation SHOULD represent it in the data model
+using an interface appropriate for the semantics and meaning
+that the implementation attaches to that _annotation_.
 
 ```ts
 interface Unsupported {
--- a/spec/data-model/message.dtd
+++ b/spec/data-model/message.dtd
@@ -9,7 +9,7 @@
 <!ATTLIST key default (true | false) "false">
 
 <!ELEMENT pattern (#PCDATA | expression)*>
-<!ELEMENT expression (literal | variable | function | reserved)>
+<!ELEMENT expression (literal | variable | function | unsupported)>
 
 <!ELEMENT literal (#PCDATA)>
 <!ATTLIST literal quoted (true | false) #REQUIRED>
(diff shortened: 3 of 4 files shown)


spec/data-model.md
```diff
@@ -145,17 +145,22 @@ interface Option {
 }
 ```
 
-A `Reserved` represents an _expression_ with a _reserved_ _annotation_.
-The `sigil` corresponds to the starting sigil of the _reserved_.
+A `Reserved` represents an _expression_ with a _reserved_ or _private-use_ _annotation_.
+The `sigil` corresponds to the starting sigil of the _annotation_.
 The `source` is the "raw" value (i.e. escape sequences are not processed)
-and includes the starting `sigil`.
+and does not include the starting `sigil`.
```
aphillips, line 151: Is this true? The escape sequences prevent falsely finding an ending marker. The unescaped `source` is the "raw" value. Reserializing the `source` would reinsert the escapes (this is how one avoids double-escaping traps)

spec/data-model.md
```diff
@@ -145,17 +145,22 @@ interface Option {
 }
 ```
 
-A `Reserved` represents an _expression_ with a _reserved_ _annotation_.
-The `sigil` corresponds to the starting sigil of the _reserved_.
+A `Reserved` represents an _expression_ with a _reserved_ or _private-use_ _annotation_.
+The `sigil` corresponds to the starting sigil of the _annotation_.
```
aphillips, line 149: I think this is incorrect.

Reserved represents a portion of the syntax that is wholly invalid--not permitted to be used--but which might become part of a future incarnation. If a part of reserved were used in a future incarnation, existing (pre the unreserving) implementations would continue with whatever behavior reserved provides. Probably this means, as you have it here, passing reserved gunk through the data model without processing.

Private use is different. Private use is a portion of the syntax that _is valid_, but which may not be functional in a given implementation. 

What I would do is define private use and reserved separately and use the same language where appropriate.

Also: note that private-use is a feature of our specification, which is why I put it before reserved in the text.

spec/data-model.md
```diff
@@ -145,17 +145,22 @@ interface Option {
...
 
 Implementations MUST NOT rely on the set of `sigil` values remaining constant,
 as future versions of this specification MAY assign other meanings to such sigils.
 
 If the _expression_ includes a _literal_ or _variable_ before the _annotation_,
 it is included as the `operand`.
 
+When parsing the syntax of a _message_ that includes a _private-use_ _annotation_
+supported by the implementation,
+the implemenation MAY represent it in the data model using a different interface
```
aphillips, line 161: I think we might need some stronger guidance here. I agree with the "MAY" you have, but would suggest something along the lines of:

> _Private-use_ _annotations_ are specific to a particular "private agreement"
> between the various implementations that support a given form of _private-use_.
>
> When parsing the syntax of a _message_ that includes a _private-use_ _annotation_
> unrecognized by the implementation, the _annotation_ MUST be processed
> identically to a _reserved_ _annotation_.
>
> An implementation that supports a given _private-use_ _annotation_ MUST
> define the specific interface to support the semantics, structure, and meaning
> that it provides. Use of existing data model interfaces is RECOMMENDED, although
> an implementation MAY use any interface appropriate for its needs.

spec/data-model.md
aphillips, line 151: Also, by removing the starting sigil, you remove the ability for an implementation to tell how reserved (or private) sequence was introduced. If the data model were serialized and sent to an implementation that supported one or another sigil, the receiver would have no way of knowing what the sigil was. Is there a reason you removed the sigil?

spec/data-model.md
```diff
@@ -145,17 +145,22 @@ interface Option {
...
-A `Reserved` represents an _expression_ with a _reserved_ _annotation_.
-The `sigil` corresponds to the starting sigil of the _reserved_.
+A `Reserved` represents an _expression_ with a _reserved_ or _private-use_ _annotation_.
+The `sigil` corresponds to the starting sigil of the _annotation_.
 The `source` is the "raw" value (i.e. escape sequences are not processed)
-and includes the starting `sigil`.
+and does not include the starting `sigil`.
 
 Implementations MUST NOT rely on the set of `sigil` values remaining constant,
 as future versions of this specification MAY assign other meanings to such sigils.
```
aphillips, line 154: I think this normative text is unenforceable. If I hardcoded the existing sigils for functions, pua, and reserved into my code, that's my business. I think any test we make of an MFv2 "1.0" implementation would absolutely depend on implementations treating the set of sigil values as constant (for that version).

I think what you mean is:

```suggestion
> **Note**
> Implementers should be aware that future versions of this specification
> might assign meaning to _reserved_ `sigil` values.
```

PS> I use a non-normative "should" because Notes are non-normative.

aphillips (changes requested):

spec/data-model.md
eemeli: The `sigil` in this case is the `^`, `*` or other identified sigil from which we can tell that this annotation is private-use or reserved. Including it or not in `source` is a matter of choice; I'm here proposing that it should not be, as that's how all the other sigils are handled in the spec.

spec/data-model.md
eemeli: Could you give an example of a data model use case where it would be useful for the _private-use_ and _reserved_ syntax rules to be represented by different data structures, rather than this single `Reserved`?

Given that supported private-use syntax would end up using a different interface altogether, I'm failing to come up with a realistic scenario where I'd like unsupported private-use and reserved to be handled differently. I suppose a linter error for a message including one or the other could use a slightly different text, but that can be detected from the `sigil`.

spec/data-model.md
eemeli: It's because it's already in the separate `sigil` field.

spec/data-model.md
eemeli: I don't think the "parsing the syntax" sentence belongs in the data model spec, but in the syntax. It's defining a MUST about what the syntax means, and if included here that would get watered down by the qualifier at the top of this doc:
> Implementations are not required to use this data model for their internal representation of messages.

On review, I now also note that the Extensions section will need to be updated to account for private-use. Something like your third paragraph is probably needed there.

spec/data-model.md
eemeli: Sure; the MUST NOT is indeed probably pretty impossible to break. I think the note might read better with this expansion?

```suggestion
> **Note**
> Implementers should be aware that future versions of this specification
> might assign meaning to _reserved_ `sigil` values
> and add new interfaces to this data model.
```

aphillips: Yes, that.

spec/data-model.md
aphillips: My comment here isn't about sigils. It is that preserving the "raw" value without unescaping produces a risk of double-escaping (the receiver cannot tell if the escape sequence is intentional or not). 

What is the value of preserving the "raw" (escaped) representation of the annotation in the data model, given that the purpose of the escapes is either ambiguity in parsing MF2 or in the runtime environment?

macchiati:
>  Implementations MUST NOT rely on the set of `sigil` values remaining
constant,
> as future versions of this specification MAY assign other meanings to
such sigils.

Not a desirable situation, since conformant implementations become
non-conformant. What other standards  (incl Unicode) have done to avoid
this problem is to put symbols/ids into two buckets:
1. private-use, and
2. reserved future versions of the spec.

Clients that use PU then don't have to worry about being non-conformant to
a future version of the specification.

spec/data-model.md
eemeli: The problem is that if we process the string, then we need to process it into a sequence of reserved-char sequences and quoted literals, because their escapes are different.

Say we didn't do that, and just applied the escapes. Then we could end up with a cooked source value like
```
one|two|three|four
```

With that, we can't know which of the following would be its original source form:
1. `one|two|three\|four`
2. `one|two\|three|four`
3. `one\|two|three|four`
4. `one\|two\|three\|four`

As the implementation or future spec version undoubtedly cares very much whether a `|` is escaped or not, we would not be able to reconstruct the annotation source.

My presumption is that almost always when encountering unsupported syntax we either don't care about its syntax source, or we just want to pass it through untouched; hence the current approach. If you think this behaviour should be reconsidered, could you raise a separate issue or submit a PR rather than increasing the scope of this one?

aphillips: I think this describes a bug in the syntax, where we were "clever" to smuggle in the quoted production and the whitespace production. I think the argument was something like: "if we unreserved a sigil, then it would just parse correctly". But I think we're better off if reserved and private are opaque to the ABNF. An unreserved sigil is subject to whatever ABNF is applied to it (either the existing annotation syntax or new syntax). Pre-unreserved implementations still won't parse into the sigil's space, so won't be affected.

In other words, I think we should modify the ABNF thusly:

```abnf
private-use    = private-start text
private-start  = "^" / "&"

; reserve additional sigils for use by 
; future versions of this specification
reserved       = reserved-start text
```

The only escapes in text are around `{` and `}` (and `\` in case one needs `\}` as a character sequence). Brackets retain syntactic meaning everywhere. The current `reserved-body` approach means that implementations parse `reserved` and `private-use` into word tokens and literals, even though the implementation is not allowed to interpret them. I think true opacity is the right approach. (In that case (1), (2), (3), and (4) are different character sequences).

aphillips: (and, yeah, we've had this discussion before [in #374](https://github.com/unicode-org/message-format-wg/pull/374#discussion_r1167447318))

Your argument there was:

> Currently this is a valid expression:
>```
>{:foo key=|{bar}|}
>```
>If we were to unreserve @ and try to give it the same semantics as with :, this would be a parse error due to the unescaped >inner {:
>```
>{@foo key=|{bar}|}
>```

The problem then is: `quoted` allows `{` and `}` unescaped (because they are "quoted"). Adding these characters to the escaped list for `quoted` simplifies the ABNF a tiny amount (quoted-escape and reserve-escape are the same and we lose some productions) and lets implementations parse out expressions by matching (unescaped) `{`/`}` pairs. The cost is that `{`/`}` must be escaped in a literal. 

While fewer escapes is better than more escapes in quoted literals, I don't think that is very onerous compared to having `private-use` (`reserved` scares me less, as we might never use it) be "semi-parsed". I was convinced before that the ABNF would be okay using what we have now because the character sequences could be squeezed in (that I wouldn't actually have to parse the contents). But the data model shows why I was shy in the first place: `private-use` and `reserved` turn out to have parsed structure where we want opacity.

aphillips: Opened #446 to discuss.

spec/data-model/README.md
```diff
@@ -81,7 +81,7 @@ The `value` of `Text` is the "cooked" value (i.e. escape sequences are processed
 Implementations MUST NOT rely on the set of `Expression` `body` values being exhaustive,
 as future versions of this specification MAY define additional expressions.
 If encountering a `body` with an unrecognised value,
-an implementation SHOULD treat it as it would a `Reserved` value.
+an implementation SHOULD treat it as it would an `Unsupported` value.
```
aphillips, line 84: Consider using a more positive statement:

```suggestion
A `body` with an unrecognized value SHOULD be treated as
an `Unsupported` value.
```

Will `body` always have a sigil, that is, will it fit the description below in which unsupported is always reserved or private use? Or do we need to add text below that says (effectively) "reserved, unsupported private-use, or any unrecognized sequence"

spec/data-model/README.md
```diff
@@ -148,20 +148,27 @@ interface Option {
...
+> might assign meaning to _reserved_ `sigil` values
+> and add new interfaces to this data model.
 
 If the _expression_ includes a _literal_ or _variable_ before the _annotation_,
 it is included as the `operand`.
 
+When parsing the syntax of a _message_ that includes a _private-use_ _annotation_
+supported by the implementation,
+the implemenation MAY represent it in the data model using a different interface
+as appropriate for the semantics and meaning that it attaches to that _annotation_.
```
aphillips, line 167: Perhaps a bit stronger:

```suggestion
the implementation SHOULD represent it in the data model
using an interface appropriate for the semantics and meaning
that the implementation attaches to that _annotation_.
```

spec/data-model/README.md
```diff
@@ -148,20 +148,27 @@ interface Option {
 }
 ```
 
-A `Reserved` represents an _expression_ with a _reserved_ _annotation_.
-The `sigil` corresponds to the starting sigil of the _reserved_.
+An `Unsupported` represents an _expression_ with a _reserved_ or _private-use_ _annotation_.
```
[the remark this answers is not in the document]
eemeli: Intentional, because we're at least so far not referring to data model constructs from elsewhere in the spec, and their names would very closely conflict with syntax rules. The data model is currently entirely built on top of the rest, and we should be careful about establishing a dependency on it from elsewhere.

spec/data-model/README.md
```diff
@@ -148,20 +148,27 @@ interface Option {
...
 The `source` is the "raw" value (i.e. escape sequences are not processed)
-and includes the starting `sigil`.
+and does not include the starting `sigil`.
 
-Implementations MUST NOT rely on the set of `sigil` values remaining constant,
-as future versions of this specification MAY assign other meanings to such sigils.
+> **Note**
+> Implementers should be aware that future versions of this specification
+> might assign meaning to _reserved_ `sigil` values
+> and add new interfaces to this data model.
```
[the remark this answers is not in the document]
eemeli: The fix looks good; it improves on your previous suggestion in https://github.com/unicode-org/message-format-wg/pull/444#discussion_r1278315019.

aphillips (approved):

spec/data-model/README.md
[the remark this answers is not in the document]
aphillips: Wonderful what a cup of coffee ☕ and fresh eyes 👀 will do sometimes...

PR #447, opened 2023-07-31
Bringing the spec description in sync with the abnf

mihnita:
The `.abnf` is now this:
```
variable = "$" name
```

In `{b :html}` the `:html` is the function, not `b` (similar to `{43 :number}` and `{|20231217T1930| :datetime}`)

### DIFF
--- a/exploration/0000-design-proposal-template.md
+++ b/exploration/0000-design-proposal-template.md
@@ -1,5 +1,13 @@
 # Design Proposal Template
 
+Status: **Accepted**
+
+_Use the status **Proposed** when creating a design proposal._
+_The status will be changed to **Accepted** when the design proposal is accepted as working group consensus._
+_The status will be changed to **Rejected** if the design proposal ultimately is not adopted by the working group._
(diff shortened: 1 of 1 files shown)


aphillips:
You're not wrong. Note the changes in #441 and the move towards defining standalone/opening/closing "elements", which should properly be termed _expression_.

mihnita:
> Note the changes in #441

If it helps with the refactoring you are trying to do I can either delay this until after you land #441, or you can make the change directly in #441.

eemeli:
I think this is blocked by #441, and I'd really rather not introduce further increases to its scope from its current rather massive extent, but discuss changes like this after that PR has been merged.

aphillips:
Per teleconference, please refector to fit in the syntax.md file and then ping me to merge.

eemeli (changes requested):
As discussed on prior meetings, this PR should be rebased on the latest `main` for further consideration.

mihnita:
Replaced by #466
This one was too messy to merge. And was done in my main, making it worse.

PR #448, opened 2023-08-02
Exploration: sigils inside expressions

stasm:
This is work in progress, in a very rough shape right now. It’s a draft of a draft. I probably spent too long thinking about this in isolation and should have reached out earlier to be able to iterate on it faster. Instead, I got a bit stuck.

So I’m sharing this now, incomplete as it is. I’d like to make an introduction to this topic at the next week’s meeting.

Rendered: [exploration/sigils.md](https://github.com/stasm/message-format-wg/blob/sigils/exploration/sigils.md)

----

tl;dr:

I suggest that we take another stab at discussing markup, focusing on intended use-cases and requirements, and then design the syntax for it. In particular, while the current syntax strongly suggests that open, close, and standalone are properties of functions, there are good reasons to consider them as properties of expressions or even placeholders.

### DIFF
--- a/exploration/sigils.md
+++ b/exploration/sigils.md
@@ -265,11 +265,11 @@ We can require a data model error be emitted when this happens.
 
 ### Q04: Should formatting and selectors annotations have different syntax?
 
-	See [#260](https://github.com/unicode-org/message-format-wg/issues/260).
+See [#260](https://github.com/unicode-org/message-format-wg/issues/260).
 
 ### Q05: Do we need namespaced local variables?
 
-	See [#403](https://github.com/unicode-org/message-format-wg/issues/403).
+See [#403](https://github.com/unicode-org/message-format-wg/issues/403).
 
 ## Alternatives for the syntax
 
@@ -358,7 +358,19 @@ Using paired square brackets:
 * new special characters: `[` and `]`
 * can use `{[foo]}` for standalone placeholders
 
-### Separate expression syntax
+### Extra placeholder-only syntax
+
+This is the same as above, but applied only to placeholders.
+Possible with the following change to our ABNF:
+
+```diff
+-expression = "{" [s] ((operand [s annotation]) / annotation) [s] "}"
++expression = (operand [s annotation]) / annotation
++selector = "{" [s] expression [s] "}"
++placeholder = "{" [s] expression [s] "}"
+```
+
+### Separate expression and placehodler syntax
 
 Delimit expressions with one pair of curly braces and then embed them into placeholders which use another pair of curly braces as delimiters.
 
@@ -373,6 +385,7 @@ Delimit expressions with one pair of curly braces and then embed them into place
 		{% {$x}} ... {/ {$x}}
 
 * can forbid open/close in local declarations and selectors on the syntax level
+* a large depature from the current syntax
 * visually noisy, especially near the edges of a pattern: `{{{$username}}}` may be considered a bit excessive
 * another meaning for `{` and `}`
 
@@ -404,6 +417,7 @@ Delimit expressions with square brackets, placeholders with curly braces.
 
 * can forbid open/close in local declarations and selectors on the syntax level
 * may be combined with the brackets-as-the-function-call approach detailed above: `let $foo = [$count :number]` but `let $foo = $count`.
+* a very large depature from the current syntax
 * visually noisy
 * new special chars: `[` and `]`
 
--- a/exploration/sigils.md
+++ b/exploration/sigils.md
@@ -17,258 +17,259 @@ I've attempted to present some possible approaches in the final part of the docu
 
 Already established decisions which impose limitations on the design, sometimes in a good way.
 
-* **C01** `|` is the delimiter for quoted literals.
+### C01: `|` is the delimiter for quoted literals.
 
-* **C02** Unquoted literals are `nmtokens`, but they cannot start with `:` and `-`.
+### C02: Unquoted literals are `nmtokens`, but they cannot start with `:` and `-`.
 
-* **C03** `*` is the catch-all variant key.
+### C03: `*` is the catch-all variant key.
 
-* **C04** `$` is the sigil for variables.
+### C04: `$` is the sigil for variables.
 
-	In contrast to ICU MessageFormat (1.0), we want ot decorate variables with a sigil to make them stand out visually, and to make it clearer that their names are "code" and should not be translated.
+In contrast to ICU MessageFormat (1.0), we want ot decorate variables with a sigil to make them stand out visually, and to make it clearer that their names are "code" and should not be translated.
 
-* **C05** The operand in annotated expressions comes before the annotation.
+### C05: The operand in annotated expressions comes before the annotation.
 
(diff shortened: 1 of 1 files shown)


eemeli:
This is a fascinating exploration, and I find myself wanting to respond to it at multiple different levels. However, before doing so here (and I would like to ask others to also desist), I have the following meta-level concerns that I would like us to address:
1. We've been working on this spec for a while, and in its current shape it at least _looks_ like it's nearly ready to publish. We had previously aimed to get this out together with ICU 74 this fall, but we've already missed that target. The next one is ICU 75, the release candidate of which is expected in March 2024. Working backwards from there we should establish at the very least "soft freeze" dates for the syntax and other spec parts to allow time for implementation work.

   I would estimate that we have at most 10 regular meetings before such a soft freeze date, less if we want a realistic buffer for ourselves. We should evaluate how much of that time we are willing and interested in spending on potentially wide-ranging explorations of ground on which we have previously established consensus, and/or find ways to make more time available outside those meetings. In doing so, we ought to be realistic about how much actual time each of us has available for these efforts.

   We've been at this for nearly four years now. I for one would very much like to finally get the 2.0 spec out next spring.

2. Commenting on this issue will not be the right way to discuss the actual contents of this exploration. Honestly, this sounds like it might need its own GitHub Discussions category to present and discuss its constraints, requirements, questions, and proposed alternatives. But I think I'll leave that up to @stasm.

stasm:
Thanks, @eemeli, for starting with the meta-discussion. I don't mean to disrupt our current progress, and I agree that it's good to first decide how much we want to invest in this.

A large part of the exploration are the *agreed* constraints and requirements. Hopefully documenting them is worthwhile regardless of what comes out of this particular discussion. I found myself wishing that such a list existed already when I was thinking about the sigil design.

I've also tried to avoid personal opinions in the doc. The *proposed* constraints and requirements, as well as the open questions, are invitation for discussion that I think we should have and document the outcome of, even if we decide to not change anything. I hope it wouldn't be a wasted effort to at least see how far we are from consensus about them.

The section about syntax alternatives is an exploration of options available to us wrt. to designing syntax that meets the requirements. Please read it as "here's what we *could* do" rather than "here's what we *should* do". The shift from "could" to "should" needs to be informed by use-cases and requirements.

Lastly, at least for some outcomes of the discussion, I think the resulting changes to the syntax would be rather surgical. For instance, replacing `:` as the function introducer with something else is in fact a minimal change in terms of lines of spec changed. We can also consider *removing* some features from the curent spec if we feel that they are not justified enough.

I'm going to refrain from discussing the specifics for now so that I can take some rest over the weekend :)

My ask for right now is:

* Please skim through the doc before the meeting next Monday.
* Think about how you feel about the *proposed* contraints and requirements. Anything you'd change or add?
* Do you feel that it would be helpful to discuss and document the answers to the *open* questions listed in the doc?
* Let's discuss next steps on Monday.

macchiati:
We software engineers should always be wary of the occupational hazard:
Il meglio è l'inimico del bene.

exploration/sigils.md
```diff
@@ -0,0 +1,417 @@
...
+	// Or perhaps repeat the annotation?
+	{{$x +html}Hello{$x -html}}
+
+> [!IMPORTANT]
+> If so:
+> Open/close is not a property of the annotation.
+> It is at most a property of the expression, which can comprise just the operand.
+> Or perhaps even a property of the placeholder (see below).
+
+### Q03: Should it be possible to change the open/close role once assigned?
```
mihnita, line 258: NO!
:-D

exploration/sigils.md
```diff
@@ -0,0 +1,417 @@
...
+
+	{img :html}
+	{img +html}
+	{img -html}
+
+* at odds with P01; is the function called `html` or `:html`?
+
+Double char sigils to group similar sigils under a common introducer.
+
+	{img :html}
```
mihnita, line 291: I think that technically these should be
:html
::html
/:html
+:html
and so on.

`:html` means function.
And if open/close/etc are not function attributes, they they belong "outside" the `:html`

Not saying it is agreement yet that these are function attributes.
But if we agree or not, it affects the order of the sigils / sub-sigils.

---

In general I think that if open / close are function attributes they can be represented in the options bag: `{a :html type:closed}`, there is no need to single them out, except for convenience of typing.

But if the same flag shows again and again across many functions, and can be used (by humans and linters) without any understanding or access to the registry, then I argue it is not a function attribute.
I can tell that `"...{foo -bar}...{foo +bar}..."` is wrong even if I have no clue what `bar` is / does.

mihnita (commented):
General comment: if we look again at the whole syntax and shake this boat, then I suggest to consider the need for other placeholder attributes / flags.
Things that are not function specific, but have "global", well, understood meanigns.

Some examples:
* can clone
* can delete
* don't reorder
* not rendered

---

**`can clone` => this is OK:**
source: foo {ph} bar
target: foox {ph} barx {ph}

**`can delete` => this is OK:**
source: foo {ph} bar
target: foox barx

If we think html, some tags can be doubled or removed, some don't.
For example the presence of an `id` means there cloning / removing are forbidden, because they break functionality.

Similar XLIFF concepts:
http://docs.oasis-open.org/xliff/xliff-core/v2.1/os/xliff-core-v2.1-os.html#cancopy
https://docs.oasis-open.org/xliff/xliff-core/v2.1/os/xliff-core-v2.1-os.html#candelete

---

**`don't reorder` => this is NOT OK:**
source: foo {ph1} bar {ph2} 
target: foox {ph2} barx {ph1} 

---

**`not rendered`** means that the placeholder does not result in visible readable content. Goes beyond text, to (for example) images.
Allows for linguistic post-processing to ignore non-render-able attributes:

`"La  sauvage"` (the wild bee) renders as "La 🐝 sauvage", and "bee" is "abeille" in French, so a corrected string would be "L'🐝 sauvage"

So  an `img` html tag is "renderable.
A bold / italic / a is not. BUT! is can become renderable if the css has a `:pre`, for example.
So just by looking at `"foo <span class='zyz'>...` I can't tell if it is renderable or not.
An explicit flag added by the dev might help.

XLIFF: http://docs.oasis-open.org/xliff/xliff-core/v2.1/os/xliff-core-v2.1-os.html#canreorder

eemeli:
As the exploration touches quite heavily on open/close expressions, I figured it might be useful to look at some frequency data on them. At Mozilla, we localize our products using Pontoon, which currently holds about 160k localisable messages. Of these, about 11k (7%) include something like open/close elements.

Digging down, of those 11k messages with open/close elements, the numbers of messages with multiple options is rather low:
| key=value pairs | # of messages |
| -- | -- |
| 1 | 6414 |
| 2 | 268 |
| 3 | 33 |
| 4 | 9 |
| 5 | 3 |
| 6+ | 0 |

This dataset includes messages in a variety of formats; a vast majority of the higher-count pairs are pure-HTML tags. Of the messages using Fluent, no message includes an element with more than two key=value options.

To me, this frequency distribution would suggest that the user need to e.g. deduplicate open/close elements' options bags via local variables is likely to be rather marginal. If anyone else has any similar statistics from their databases that could be shared, that would be great.

PR #450, opened 2023-08-14
Define @attributes on expressions

### LINKED ISSUE #426: How to express expression locale?
Through our past discussions on bidirectional isolation (#28, #315) and formatting to non-string targets (#41, #272), I believe that we've established a consensus on wanting to support output that may mingle content from more than one locale and script. Going further, I think we've established that the resolved locale of an expression may be considered as a separate field of data from its formatted value.

We have not yet, however, established how the locale of an expression should be defined on the input side of things. Is it an option that some functions might have, or should we treat it as something a bit special like the existing ICU and JS Intl formatters do?

If we expect `locale` to be one option among many, then I think we should clearly establish what its name and value should be. Is it `lc`, `locale`, `locales` or something else? Can its value be something other than a BCP-47 tag? If providing fallback tags, how are they defined and delimited?

If we want to treat locale as somehow special, then we need to use some special syntax for it. Something like this?
```
{In French, "{|bonjour| @fr}" is a greeting}
```

If we only establish a common practice for locale as an option, then I think one of our base formatters should be one that allows for an expression such as the above:
```
{In French, "{|bonjour| :string locale=fr}" is a greeting}
```

eemeli:
Fixes #426

As discussed during today's call, we have use cases for attributes that pertain to an expression, rather than acting as function options. The examples presented in #426 refer just to the locale of an expression, but other valuable attributes include the translatability of an expression, and attributes such as `canCopy`, `canDelete`, `canOverlap`, and others defined in the [XLIFF 2.1 spec](http://docs.oasis-open.org/xliff/xliff-core/v2.1/os/xliff-core-v2.1-os.html#d0e4957). Many of these do not have an impact on message formatting, but can be very important for translation work.

This PR proposes the following syntax for this, using the examples included in the PR:

> A _message_ in English including a French _literal_ that should not be translated:
>
> ```
> {In French, "{|bonjour| @locale=fr @translate=no}" is a greeting}
> ```
>
> A _message_ with a `$date` _variable_ formatted with the `:datetime` _function_ using the root "und" locale:
>
> ```
> {Today is {$date :datetime weekday=long @locale=und}.}
> ```
>
> A _message_ with a `$start` _variable_ that should in translations be kept as the first element of the message.
>
> ```
> {{$start @can-reorder=no} is the beginning.}
> ```

Effectively, after the _function_ _options_, _expression attributes_ may be defined with the same _option_ syntax but with `@` as a prefix for each name. The following attributes are explicitly defined:
- `@locale`: a sequence of IETF BCP 47 language tags
- `@translate`: a literal value of either `yes` or `no`.

All other custom attributes must include a hyphen `-` in their name (such as `@can-reorder` above); everything else is ignored. Mismatching names can't be treated as a syntax error, because that would limit our later ability in a 2.1 spec version to define more attributes.

This PR also unreserves the `@` character and updates the data model descriptions as necessary.

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -105,6 +105,9 @@ one of the following is used to resolve the value of the _expression_:
 - Else, the _expression_ has a _reserved_ _annotation_,
   an Unsupported Expression error is emitted and a fallback value is used as its value.
 
+_Expression attribute resolution_ defines the handling of _expression attributes_
+that have an impact on _message_ formatting.
+
 ### Literal Resolution
 
 The resolved value of a _text_ or a _literal_ is
@@ -151,12 +154,16 @@ the following steps are taken:
    and use a _fallback value_ for the _expression_.
 3. Resolve the _option_ values to a mapping of string identifiers to values.
    For each _option_:
-     * If its right-hand side successfully resolves to a value,
-       bind the _name_ of the _option_ to the resolved value in the mapping.
-     * Otherwise, do not bind the _name_ of the _option_ to any value in the mapping.
-4. Call the function implementation with the following arguments:
+   - If its right-hand side successfully resolves to a value,
+     bind the _name_ of the _option_ to the resolved value in the mapping.
+   - Otherwise, do not bind the _name_ of the _option_ to any value in the mapping.
+4. Determine the _expression_ _locale_.
+   If the _expression_ has a valid `@locale` _expression attribute_,
+   use its resolved value as the _expression_ _locale_.
+   Otherwise, look up the _message_ _locale_ from the _formatting context_.
+5. Call the function implementation with the following arguments:
 
-   - The current _locale_.
+   - The _expression_ _locale_.
    - The resolved mapping of _options_.
    - If the _expression_ includes an _operand_, its resolved value.
 
@@ -170,7 +177,7 @@ the following steps are taken:
    their access to the _formatting context_ SHOULD be minimal and read-only,
    and their execution time SHOULD be limited.
 
-5. If the call succeeds,
+6. If the call succeeds,
    resolve the value of the _expression_ as the result of that function call.
    If the call fails or does not return a valid value,
    emit a Resolution error and use a _fallback value_ for the _expression_.
@@ -234,6 +241,24 @@ rather than the _expression_ in the _selector_ or _pattern_.
 > resolving to a _fallback value_ of `|horse|`.
 
 _Pattern selection_ is not supported for _fallback values_.
+
+### Expression Attribute Resolution
+
+The only _expression attribute_ that MUST be supported is `@locale`.
+If its value is set by a _literal_,
+it MUST be a comma-delimited sequence of IETF BCP 47 language tags.
+If its value is set by a _variable_,
+it MUST be either a single IETF BCP 47 language tag,
+or a sequence of IETF BCP 47 language tags.
+
+The resolved value of a valid `@locale` _expression attribute_
+is a sequence of IETF BCP 47 language tags.
+If a `@locale` has an invalid value,
+emit a Resolution error and ignore the _expression attribute_ in further processing.
+
+Implementations MAY ignore all other _expression attributes_ during formatting.
+Implementations MAY assign meaning during formatting to other _expression attributes_
+which use a _name_ that contains at least one U+002D HYPHEN-MINUS `-` character.
 
 ## Pattern Selection
 
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -244,7 +244,7 @@ _Pattern selection_ is not supported for _fallback values_.
 
 ### Expression Attribute Resolution
 
-The only _expression attribute_ that MUST be supported is `@locale`.
+Implementations MUST support the _expression attribute_ `@locale`.
 If its value is set by a _literal_,
 it MUST be a comma-delimited sequence of IETF BCP 47 language tags.
 If its value is set by a _variable_,
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -245,11 +245,11 @@ _Pattern selection_ is not supported for _fallback values_.
 ### Expression Attribute Resolution
 
 Implementations MUST support the _expression attribute_ `@locale`.
-If its value is set by a _literal_,
-it MUST be a comma-delimited sequence of IETF BCP 47 language tags.
-If its value is set by a _variable_,
-it MUST be either a single IETF BCP 47 language tag,
-or a sequence of IETF BCP 47 language tags.
+The value of `@locale` MUST be a sequence of
+one or more well-formed IETF BCP47 language tags.
+If the value is set by a _literal_,
+the _literal_ MUST consist of a comma-delimited sequence of
+well-formed IETF BCP47 language tags.
 
 The resolved value of a valid `@locale` _expression attribute_
 is a sequence of IETF BCP 47 language tags.
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -545,8 +545,11 @@ reserved-char  = %x00-08        ; omit HTAB and LF
 
 An **_<dfn>expression attribute</dfn>_** is a key-value pair
 containing a named argument that is attached to the _expression_ as a whole.
-They are intended to convey information about the _expression_
-that is not necessarily applicable as a _function_ _option_.
+These might be used to convey information about the _expression_
+that are not necessarily applicable as _function_ _options_,
+such as attributes related to handling during the translation process.
+They might also be used to override contextual or implementation-specific values
+used by a _function_ during formatting, such as the locale.
 For example, the `@locale` attribute overrides the locale of the _message_ for its _expression_.
 Attributes like `@can-delete` and `@translate` might not have any impact on formatting,
 acting instead as instructions to translators.
@@ -557,17 +560,16 @@ but otherwise use the same syntax as _options_.
 The _name_ of an _expression attribute_ MUST either be one of the following,
 or it MUST contain at least one U+002D HYPHEN-MINUS `-` character:
 
-- `language`: SHOULD resolve to a sequence of IETF BCP 47 language tags;
+- `locale`: a sequence of well-formed IETF BCP47 language tags
   see _expression attribute resolution_.
-- `translate`: SHOULD have a literal value of either `yes` or `no`.
+- `translate`: a literal value of either `yes` or `no`
 
 To ensure future extensibility of this specification,
 all other _expression attributes_ with a _name_ not containing `-` MUST be ignored.
 
 Multiple _expression attributes_ are permitted in an _expression_.
 Each _expression attribute_ is separated by whitespace.
-If present, _expression attributes_ MUST be the last non-whitespace content in an _expression_,
-and MUST NOT be followed by any _function_ _options_.
+If present, _expression attributes_ MUST be the last non-whitespace content in an _expression_.
 
 ```abnf
 expression-attribute = "@" option
@@ -581,15 +583,14 @@ expression-attribute = "@" option
 > {In French, "{|bonjour| @locale=fr @translate=no}" is a greeting}
 > ```
 >
-> A _message_ with a `$date` _variable_ formatted with the `:datetime` _function_
-> using the root "und" locale:
->
+> A _message_ with a `$date` _variable_ formatted using the `:datetime` _function_
+> and overriding the locale to use French formatting:
+>```
+>{The date would be displayed in French as {$date :datetime @locale=fr}.}
 > ```
-> {Today is {$date :datetime weekday=long @locale=und}.}
-> ```
 >
-> A _message_ with a `$start` _variable_ that should in translations
-> be kept as the first element of the message.
+> A _message_ with a `$start` _variable_ that should be kept in translations
+> as the first element of the message.
 >
 > ```
 > {{$start @can-reorder=no} is the beginning.}
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -321,7 +321,7 @@ during the _message_'s formatting.
 An _expression_ MUST begin with U+007B LEFT CURLY BRACKET `{`
 and end with U+007D RIGHT CURLY BRACKET `}`.
 An _expression_ MUST NOT be empty.
-An _expression_ can contain an _operand_,
+An _expression_ MUST contain an _operand_,
 an _annotation_,
 or an _operand_ followed by an _annotation_.
 This MAY be followed by one or more _expression attributes_.
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -318,15 +318,17 @@ that matches all values for a given _selector_.
 An **_<dfn>expression</dfn>_** is a part of a _message_ that will be determined
 during the _message_'s formatting.
 
-An _expression_ MUST begin with U+007B LEFT CURLY BRACKET `{` 
+An _expression_ MUST begin with U+007B LEFT CURLY BRACKET `{`
 and end with U+007D RIGHT CURLY BRACKET `}`.
 An _expression_ MUST NOT be empty.
-An _expression_ can contain an _operand_, 
-an _annotation_, 
+An _expression_ can contain an _operand_,
+an _annotation_,
 or an _operand_ followed by an _annotation_.
+This MAY be followed by one or more _expression attributes_.
 
 ```abnf
-expression = "{" [s] ((operand [s annotation]) / annotation) [s] "}"
+expression = "{" [s] expression-body *(s expression-attribute) [s] "}"
+expression-body = (operand [s annotation]) / annotation
 operand = literal / variable
 annotation = (function *(s option)) / private-use / reserved
 ```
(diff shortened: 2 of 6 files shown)


macchiati:
The @can-reorder=no doesn't look very useful. Not a blocker for this PR, but it should have a better example in the future.

It isn't clear what it would mean, but I'm guessing: 

in the translation, 
- all placeholders that appear before this placeholder appear before this placeholder in the original.
- all placeholders that appear after this placeholder appear after this placeholder in the original.

So if I had some paired placeholders like the following (pseudocode, cr = can-reorder), the translator couldn't reorder the text as needed for their language

```
.x. {bold-start @cr=no} .y. {bold-end @cr=no} .z.  {italic-start @cr=no} .w. {italic-end @cr=no} .u.
```
The translator couldn't reorder the parts, eg
```
.X. {italic-start @cr=no} .W. {italic-end @cr=no} .U. {bold-start @cr=no} .Y. {bold-end @cr=no} .Z.  
```

spec/formatting.md
```diff
@@ -151,12 +154,16 @@ the following steps are taken:
...
+   - If its right-hand side successfully resolves to a value,
+     bind the _name_ of the _option_ to the resolved value in the mapping.
+   - Otherwise, do not bind the _name_ of the _option_ to any value in the mapping.
+4. Determine the _expression_ _locale_.
+   If the _expression_ has a valid `@locale` _expression attribute_,
+   use its resolved value as the _expression_ _locale_.
+   Otherwise, look up the _message_ _locale_ from the _formatting context_.
+5. Call the function implementation with the following arguments:
+
+   - The _expression_ _locale_.
```
aphillips, line 166: I would expose the expression attributes here also. Locale may not be the only one--I certainly have timezone in mind and other attributes may be provided by the implementation or our specification that affect the formatter.

spec/formatting.md
```diff
@@ -235,6 +242,24 @@ rather than the _expression_ in the _selector_ or _pattern_.
 
 _Pattern selection_ is not supported for _fallback values_.
 
+### Expression Attribute Resolution
+
+The only _expression attribute_ that MUST be supported is `@locale`.
+If its value is set by a _literal_,
+it MUST be a comma-delimited sequence of IETF BCP 47 language tags.
```
aphillips, line 249: Do we wish to further constrain the language tags to be Unicode Locale Identifiers (ULIs)?

This text is non-specific about the conformance level of the language tags. Is well-formed enough? Valid? Canonicalized?

spec/formatting.md
```diff
@@ -235,6 +242,24 @@ rather than the _expression_ in the _selector_ or _pattern_.
 
 _Pattern selection_ is not supported for _fallback values_.
 
+### Expression Attribute Resolution
+
+The only _expression attribute_ that MUST be supported is `@locale`.
```
aphillips, line 247: I think this needs to start with the general case before jumping into the built-in attribute(s).

I would also not write this quite this way, in case we add more later. The same effect can be had with:

```suggestion
The _expression attribute_ `@locale` MUST be supported.
```

Or perhaps:

```suggestion
Implementations MUST support the _expression attribute_ `@locale`.
```

spec/formatting.md
```diff
@@ -235,6 +242,24 @@ rather than the _expression_ in the _selector_ or _pattern_.
...
+If its value is set by a _literal_,
+it MUST be a comma-delimited sequence of IETF BCP 47 language tags.
+If its value is set by a _variable_,
+it MUST be either a single IETF BCP 47 language tag,
+or a sequence of IETF BCP 47 language tags.
```
aphillips, line 252: I wouldn't differentiate literals and variables, since the content ends up the same:

```suggestion
The value of `@locale` MUST be a comma-delimited sequence of well-formed IETF BCP47 language tags.
```

(we can provide instructions on processing the language priority list, including whitespace handling and such if needed)

spec/formatting.md
```diff
@@ -235,6 +242,24 @@ rather than the _expression_ in the _selector_ or _pattern_.
...
+
+The resolved value of a valid `@locale` _expression attribute_
+is a sequence of IETF BCP 47 language tags.
+If a `@locale` has an invalid value,
+emit a Resolution error and ignore the _expression attribute_ in further processing.
```
aphillips, line 257: Should we say that the value reverts to that of the _message_?

spec/formatting.md
```diff
@@ -235,6 +242,24 @@ rather than the _expression_ in the _selector_ or _pattern_.
...
+
+Implementations MAY ignore all other _expression attributes_ during formatting.
+Implementations MAY assign meaning during formatting to other _expression attributes_
+which use a _name_ that contains at least one U+002D HYPHEN-MINUS `-` character.
```
aphillips, line 261: Where did this come from? Why is the hyphen important? Why can't implementations "assign meaning" to unhyphenated values?

spec/formatting.md
```diff
@@ -235,6 +242,24 @@ rather than the _expression_ in the _selector_ or _pattern_.
...
+If its value is set by a _variable_,
+it MUST be either a single IETF BCP 47 language tag,
+or a sequence of IETF BCP 47 language tags.
+
+The resolved value of a valid `@locale` _expression attribute_
+is a sequence of IETF BCP 47 language tags.
+If a `@locale` has an invalid value,
+emit a Resolution error and ignore the _expression attribute_ in further processing.
+
+Implementations MAY ignore all other _expression attributes_ during formatting.
```
aphillips, line 259: What does "ignore" mean? Not process them?

spec/syntax.md
```diff
@@ -318,15 +318,17 @@ that matches all values for a given _selector_.
 An **_<dfn>expression</dfn>_** is a part of a _message_ that will be determined
 during the _message_'s formatting.
 
-An _expression_ MUST begin with U+007B LEFT CURLY BRACKET `{` 
+An _expression_ MUST begin with U+007B LEFT CURLY BRACKET `{`
 and end with U+007D RIGHT CURLY BRACKET `}`.
 An _expression_ MUST NOT be empty.
-An _expression_ can contain an _operand_, 
-an _annotation_, 
+An _expression_ can contain an _operand_,
```
aphillips, line 324: I think we should promote "can" to normative "MUST" here. It was previously fine as "can" because there was nothing else permitted besides these three options.

```suggestion
An _expression_ MUST contain an _operand_,
```

spec/syntax.md
```diff
@@ -526,18 +528,73 @@ unrecognized reserved sequences have no meaning and MAY result in errors during
...
 
+### Expression Attributes
+
+An **_<dfn>expression attribute</dfn>_** is a key-value pair
+containing a named argument that is attached to the _expression_ as a whole.
+They are intended to convey information about the _expression_
+that is not necessarily applicable as a _function_ _option_.
+For example, the `@locale` attribute overrides the locale of the _message_ for its _expression_.
+Attributes like `@can-delete` and `@translate` might not have any impact on formatting,
+acting instead as instructions to translators.
```
aphillips, line 552: The syntax specification should only include attributes that we formally define all the way through. If these are examples, they should be treated as such, e.g. "If an implementation supported attributes such as..."

spec/syntax.md
```diff
@@ -526,18 +528,73 @@ unrecognized reserved sequences have no meaning and MAY result in errors during
...
                / %x7E-D7FF      ; omit surrogates
                / %xE000-10FFFF
 ```
 
+### Expression Attributes
+
+An **_<dfn>expression attribute</dfn>_** is a key-value pair
+containing a named argument that is attached to the _expression_ as a whole.
+They are intended to convey information about the _expression_
+that is not necessarily applicable as a _function_ _option_.
```
aphillips, line 549: I think this is too opinionated for us to include at this juncture. I would tone it down:

```suggestion
These might be used to convey information about the _expression_ that are not necessarily applicable as _options_ to a _function_, such as attributes related handling during the translation process.
Or they might be used to override contextual or implementation-specific values used by a _function_ during formatting, such as the locale, time zone, or other settings.
```

spec/syntax.md
```diff
@@ -526,18 +528,73 @@ unrecognized reserved sequences have no meaning and MAY result in errors during
...
+For example, the `@locale` attribute overrides the locale of the _message_ for its _expression_.
+Attributes like `@can-delete` and `@translate` might not have any impact on formatting,
+acting instead as instructions to translators.
+
+_Expression attributes_ are prefixed by a U+0040 COMMERCIAL AT `@` sign,
+but otherwise use the same syntax as _options_.
+
+The _name_ of an _expression attribute_ MUST either be one of the following,
+or it MUST contain at least one U+002D HYPHEN-MINUS `-` character:
```
aphillips, line 558: I think you're trying to create a reserved namespace of "no hyphens" for use by the specification and then allow implementations to define their own expression attributes by forcing them to include hyphens. If that's the case, _we should say that_. Personally, I would have gone for some other mechanism, such as reserving `_` (which is a valid `name` starter) for implementation defined items, e.g.:

> {$foo :function option=value @locale=fr-FR @canReorder=no @_custom=val}

spec/syntax.md
```diff
@@ -526,18 +528,73 @@ unrecognized reserved sequences have no meaning and MAY result in errors during
...
+
+- `language`: SHOULD resolve to a sequence of IETF BCP 47 language tags;
```
aphillips, line 560: `language` or `locale`??

Also, in formatting.md you said it MUST resolve to a language tag. It should say MUST here too. The syntax won't be responsible for checking it (in terms of ABNF) but let's not confuse people by mixing normative language.

Or avoid the issue by saying:

```suggestion
- `locale`: a sequence of well-formed IETF BCP47 language tags
```

spec/syntax.md
```diff
@@ -526,18 +528,73 @@ unrecognized reserved sequences have no meaning and MAY result in errors during
...
+  see _expression attribute resolution_.
+- `translate`: SHOULD have a literal value of either `yes` or `no`.
```
aphillips, line 562: Same comment here.

spec/syntax.md
```diff
@@ -526,18 +528,73 @@ unrecognized reserved sequences have no meaning and MAY result in errors during
...
+
+To ensure future extensibility of this specification,
+all other _expression attributes_ with a _name_ not containing `-` MUST be ignored.
```
aphillips, line 565: How about just allowing unrecognized attributes to be ignored?

spec/syntax.md
```diff
@@ -526,18 +528,73 @@ unrecognized reserved sequences have no meaning and MAY result in errors during
...
+
+Multiple _expression attributes_ are permitted in an _expression_.
+Each _expression attribute_ is separated by whitespace.
+If present, _expression attributes_ MUST be the last non-whitespace content in an _expression_,
+and MUST NOT be followed by any _function_ _options_.
```
aphillips, line 570: Well... or anything else, like an annotation with no options or an operand. Just stop after "content in an _expression_"?

spec/syntax.md
```diff
@@ -526,18 +528,73 @@ unrecognized reserved sequences have no meaning and MAY result in errors during
...
+>
+> ```
+> {In French, "{|bonjour| @locale=fr @translate=no}" is a greeting}
+> ```
+>
+> A _message_ with a `$date` _variable_ formatted with the `:datetime` _function_
+> using the root "und" locale:
+>
+> ```
+> {Today is {$date :datetime weekday=long @locale=und}.}
```
aphillips, line 588: The root locale might not be the most evocative? Perhaps mirror the above example:

```suggestion
> A _message_ with a `$date` _variable_ formatted using the `:datetime` _function_
> and overriding the locale to use French formatting:
>```
>{The date would be displayed in French as {$date :datetime @locale=fr}
```

(Note that this doesn't require "translate=no" because the datetime function is doing the work.
```

File: spec/syntax.md [hunk 17]
```diff
@@ -526,18 +528,73 @@ unrecognized reserved sequences have no meaning and MAY result in errors during
...
+> ```
+>
+> A _message_ with a `$start` _variable_ that should in translations
+> be kept as the first element of the message.
```
aphillips, line 592: should in translations be kept => should be kept in translations

aphillips (changes requested):
Good start.

Be careful of whether it is @language or @locale.

I think this is too opinionated about usage and needs to be slightly genericized.

See comments about namespacing. I think checking the interior of the name is too much work? I also think that we should explicitly define what implementations are allowed to define for themselves.

One normative thing is that unrecognized items are ignored, but does that mean that they are not passed downstream? Are they deleted from the data model, for example? We know how to process them, just not what they mean...

spec/formatting.md
```diff
@@ -151,12 +154,16 @@ the following steps are taken:
    and use a _fallback value_ for the _expression_.
 3. Resolve the _option_ values to a mapping of string identifiers to values.
    For each _option_:
-     * If its right-hand side successfully resolves to a value,
-       bind the _name_ of the _option_ to the resolved value in the mapping.
-     * Otherwise, do not bind the _name_ of the _option_ to any value in the mapping.
-4. Call the function implementation with the following arguments:
-
-   - The current _locale_.
+   - If its right-hand side successfully resolves to a value,
```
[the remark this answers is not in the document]
eemeli: This text isn't changed from the previous; just the bullet point style is fixed.

spec/formatting.md
```diff
@@ -151,12 +154,16 @@ the following steps are taken:
...
+     bind the _name_ of the _option_ to the resolved value in the mapping.
+   - Otherwise, do not bind the _name_ of the _option_ to any value in the mapping.
```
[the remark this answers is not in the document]
eemeli: This text isn't changed from the previous; just the bullet point style is fixed.

spec/formatting.md
```diff
@@ -151,12 +154,16 @@ the following steps are taken:
...
+4. Determine the _expression_ _locale_.
+   If the _expression_ has a valid `@locale` _expression attribute_,
```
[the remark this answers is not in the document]
eemeli: That's defined in the _expression attribute resolution_ section.

spec/formatting.md
eemeli: At least for the JS `Intl.DateTimeFormat`, `timeZone` is a regular option. Should we not handle it as such?

For other attributes, there is this below:
> Implementations MAY assign meaning during formatting to other _expression attributes_

Should the text here make a reference to that as well, or is that implicit permission sufficiently clear?

spec/formatting.md
eemeli: I'm very happy to take input on this; I'm honestly not sure what the right specification here would be.

spec/formatting.md
eemeli: No need, because that's implicit in what happens if an expression has no `@locale` attribute.

spec/formatting.md
eemeli: Yes.

spec/formatting.md
eemeli: If we allow implementations to use any attribute names, then we won't be able to introduce any additional attributes later without a danger of breaking userspace.

With the `-` requirement, we allow for custom attributes like `@can-delete`, but leave space for later MF 2.x spec versions to define other attributes.

spec/syntax.md
eemeli: This is beyond the scope of this PR, which for this line is only removing the trailing space.

spec/syntax.md
eemeli:
```suggestion
- `translate`: a literal value of either `yes` or `no`.
```

spec/syntax.md
eemeli: The reason for creating the namespace is explicitly said a few lines further down:
> To ensure future extensibility of this specification,
all other _expression attributes_ with a _name_ not containing `-` MUST be ignored.

An underscore `_` is a possible alternative as well. I went with `-` because that's the [requirement](https://html.spec.whatwg.org/multipage/custom-elements.html#valid-custom-element-name) used by custom HTML elements, and that seemed like pretty decent prior art to follow.

spec/syntax.md
eemeli: Because then we could not later assign a meaning to them without potentially breaking user code.

spec/syntax.md
eemeli, line 570:
```suggestion
If present, _expression attributes_ MUST be the last non-whitespace content in an _expression_.
```

spec/syntax.md
eemeli, line 592:
```suggestion
> A _message_ with a `$start` _variable_ that should be kept in translations
> as the first element of the message.
```

spec/formatting.md
```diff
@@ -105,6 +105,9 @@ one of the following is used to resolve the value of the _expression_:
 - Else, the _expression_ has a _reserved_ _annotation_,
   an Unsupported Expression error is emitted and a fallback value is used as its value.
 
+_Expression attribute resolution_ defines the handling of _expression attributes_
+that have an impact on _message_ formatting.
```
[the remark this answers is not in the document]
eemeli: I'm not so sure about attributes getting passed as such to formatting or selection functions. If that makes sense, then why not pass them as function options?

Locale is special because
1. It's already handled as special by existing formatters, and
2. it may get a separate field in the formatted-parts output.

I can see how e.g. timezone could be handled similarly specially by setting up some aspect of the environment before a formatter call rather than considering it as a function option, but for that I'd expect it to be a thing that the implementation does, and which affects the formatter only indirectly.

> The attributes need to be resolved if they are present.

That's not really in line with what we're saying about other ignored code. For example, we allow an implementation to ignore errors in expressions that aren't needed for the current formatting. Similarly, we should ignore errors in attributes that don't impact the formatting.

spec/syntax.md
aphillips: Reserving a namespace is a separate thing from ignoring unrecognized values. These can work in tandem. An MF2.0 implementation would ignore the unrecognized value `@foo` that MF2.1 unreserves, as well as `@_bar` that some other proprietary implementation is using.

Alternatively (and as I suggested elsewhere), as long as the values are syntactically correct, the implementation can just parse the values without imputing meaning to them and pass them. It only has to apply validation rules to attributes it "knows" (and do we define an error for that???)

spec/formatting.md
aphillips: > For other attributes, there is this below:
>> Implementations MAY assign meaning during formatting to other expression attributes
> Should the text here make a reference to that as well, or is that implicit permission sufficiently clear?

You're permitting _implementations_ to assign meaning, but here I'm suggesting that _functions_ can assign meaning to attribute values, including those unknown to the implementation. They can only do this if they have access to the values, of course.

spec/syntax.md
aphillips: The starts-with-underscore I find unattractive, but it fits with `name` without any changes and only requires a one-character lookahead to figure it out. I like `-` okay, I guess, but we really should resolve our naming issues so that we don't have a motley collection of naming schemes.

I filed #453 to track.

aphillips (changes requested):

spec/formatting.md
eemeli: That works. Your proposed text already clarifies an empty value to be an error, as the sequence need to consist of "one or more" tags.

spec/syntax.md
eemeli: Sure, that makes sense. I had not earlier caught the concern about an expression consisting of only attributes being non-empty.

spec/syntax.md
eemeli: Consider the following hypothetical, in a world where we allow implementations to assign their own meaning to any attributes not explicitly defined in the spec like `@locale` and `@translate`:

Let's say that an implementation defines `@fallback` to provide a value that should be used for a placeholder, should its formatting fail for whatever reason. This doesn't really make sense as a function option, because it's covering the behaviour of what happens if the function throws an error. Other implementations could just ignore this attribute, and use the spec-defined fallback representation.

Then let's say that in a 2.1 version of the MF spec we'd like to define `@fallback` to mean a fallback locale to use, if the function doesn't support the message locale, or the one set by `@locale`. This hypothetical attribute would have the same semantics as `@locale`, but it would append values to the language priority list rather than replacing it.

But in this world we could not do that without potentially breaking userspace, because an implementation as defined above could exist, along with a pre-existing body of messages using `@fallback` with a completely different meaning. In this case, the meanings of "fallback" are quite different, but even should they align, it's very likely that the detailed implementation and requirements could differ for the attributes. In other words, we would not be able to define _any_ future attributes in the spec without a danger of breaking userspace.

spec/formatting.md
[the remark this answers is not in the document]
aphillips: > >     The attributes need to be resolved if they are present.
> That's not really in line with what we're saying about other ignored code. For example, we allow an implementation to ignore errors in expressions that aren't needed for the current formatting. Similarly, we should ignore errors in attributes that don't impact the formatting.

The attributes need to be parsed, even if they are not recognized. The parser does not necessarily know what is supported downstream or in an implementation.

> Similarly, we should ignore errors in attributes that don't impact the formatting.

Depends on what "error" means. If there is a syntax error, we can't ignore that, no? (some of this is about what "we" means here:  It is up to the function to decide if an attribute is "bad" or just to ignore it. "Our" job is to get the options and variables to the function.)

mihnita:
> {{$start @can-reorder=no} is the beginning.}

This is not the intended XLIFF usage.
The use case for reorder is not allowing this kind of translation:
```
src: ...{ph1} .... {ph2}....
trg: ...{ph2} .... {ph1}....
```
The reason this makes sense in XLIFF is that it is an interchange format.
It is used to extract strings from other formats and represent them in a unified manner.
So "no reorder" would be used if the string in the original format was something like ".... %d ... %s ...".
(yes, bad i18n, but xliff should be able to represent that)

But I don't think it has a use case for MF2.

eemeli:
Based on discussion during today's call, I'll be submitting a [design document](https://github.com/unicode-org/message-format-wg/blob/main/exploration/0000-design-proposal-template.md) that this PR may be solving.

eemeli:
Closing in favour of #592.

PR #451, opened 2023-08-15
Add GitHub push action enforcing Prettier style

eemeli:
The styles we use to write markdown vary, and the result is occasionally a bit messy.

Prettier defines some pretty decent defaults, and using them allows us to care less about the style we use. This PR adds a GitHub Action that runs Prettier after every commit that touches a `.md` file and pushes any changes to the current branch with a `style: Apply Prettier` commit message.

The first commit here fixes the one place where the styling appeared to cause an issue in `spec/syntax.md`; the second commit has been automatically generated by the action.

### DIFF
--- a/.github/workflows/prettier.yaml
+++ b/.github/workflows/prettier.yaml
@@ -1,0 +1,30 @@
+name: Apply Prettier style
+
+on:
+  push:
+    paths:
+      - "**.md"
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  prettier:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: "20.x"
+      - run: npm install --no-save prettier@3
+      - run: npx prettier --write .
+      - name: git config
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+      - run: git add .
+      - name: git commit & push any changes
+        run: |
+          git diff-index --quiet HEAD || (git commit -m "style: Apply Prettier" && git push)
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -1,10 +1,9 @@
 ---
 name: Feature request
 about: Suggest an idea or feature for Message Format
-title: ''
-labels: ''
-assignees: ''
-
+title: ""
+labels: ""
+assignees: ""
 ---
 
 **Is your feature request related to a problem? Please describe.**
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 .DS_Store
+node_modules/
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,6 +10,7 @@ the information on the Contributor License Agreement below. In addition, you sho
 2. Watch this repository (use the "Watch" button in the upper right corner)
 
 <!-- boilerplate follows - do not edit -->
+
 ## Contributor License Agreement
 
 In order to contribute to this project, the Unicode Consortium must have on file a Contributor License Agreement (CLA) covering your contributions, either an individual or a corporate CLA. Pull Requests will not be merged until the correct CLA is signed. Which version needs to be signed depends on who owns the contribution being made: you as the individual making the contribution or your employer. _It is your responsibility to determine whether your contribution is owned by your employer._ Please review [The Unicode Consortium Intellectual Property, Licensing, and Technical Contribution Policies][policies] for further guidance on which CLA to sign, as well as other information and guidelines regarding the Consortium’s licensing and technical contribution policies and procedures.
@@ -19,7 +20,6 @@ In order to contribute to this project, the Unicode Consortium must have on file
 - **Corporate CLA**: If you have determined that a Corporate CLA is appropriate, please check the [public list of Corporate CLAs][unicode-corporate-clas] that the Consortium has on file. If your employer has already signed a CLA, then when you open your first Pull Request, an automated comment will appear that contains a link you can follow to declare your affiliation with this employer. If your employer has not already signed a CLA, you will need to arrange for your employer to sign the Corporate CLA, as described in [How to Sign a Unicode CLA][signing].
 
 Unless otherwise noted in the LICENSE file, this project is released under the free and open-source [Unicode License][unicode-license], also known as Unicode, Inc. License Agreement - Data Files and Software.
-
 
 [policies]: https://www.unicode.org/policies/licensing_policy.html
 [unicode-corporate-clas]: https://www.unicode.org/policies/corporate-cla-list/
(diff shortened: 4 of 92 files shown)


eemeli:
It looks like the action's email address will need to match one that's acceptable to the CLA bot.

@aphillips @macchiati Is there any such bot email address yet that could be used?

srl295:
I'll add it as an exception. 
Same as dependabot. 

Ok to fix tomorrow?

eemeli:
Tomorrow's completely fine; this is fixing a long-running annoyance.

Right now the email address is empty in the generated commits, so please tell me which address to use once you've got it set up? For example, `github-actions@github.com` could work.

.github/workflows/prettier.yaml
```diff
@@ -0,0 +1,26 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: "20.x"
+      - run: npm install --no-save prettier@3
+      - run: npx prettier --write .
+      - name: git config
+        run: |
+          git config user.name 'github-action'
```
srl295, line 21: maybe it should be `actions` because https://github.com/actions

srl295:
I don't know what a good email would be. But i wonder if `actions` would be a better username since that's already reserved. Someone could sign up and spoof this one!

Anyway i added it to the allow list so you should be able to pass this

srl295 (approved):
LGTM can update the id later

eemeli:
I set the action to use `github-actions <github-actions@github.com>`. Not sure if the CLA checked has re-run itself though?

srl295:
> github-actions seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).

srl295:
> I set the action to use `github-actions <github-actions@github.com>`. Not sure if the CLA checked has re-run itself though?

seemed to make it not work… i just added `github-actions` as the exclusion (by the github id not the email)

.github/workflows/prettier.yaml
eemeli: Ah, wait: Did i misunderstand, and you meant that the name should be `actions` rather than `github-actions`?

.github/workflows/prettier.yaml
```diff
@@ -0,0 +1,26 @@
...
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: "20.x"
+      - run: npm install --no-save prettier@3
+      - run: npx prettier --write .
+      - name: git config
+        run: |
+          git config user.name 'github-actions'
+          git config user.email 'github-actions@github.com'
```
srl295, line 22: https://github.com/orgs/community/discussions/26560#discussioncomment-3531273

```suggestion
          git config user.name "github-actions[bot]"
          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
```

srl295 (approved):

srl295:
🎉

eemeli:
Phew, I think it's now sorted? Thank you @srl295 for your help here!

srl295:
image
seems ok

srl295 (approved):
💯 for persistence

stasm (changes requested):
> This PR adds a GitHub Action that runs Prettier after every commit that touches a `.md` file and pushes any changes to the current branch with a `style: Apply Prettier` commit message.

Please don't auto-commit to `main` like this. It will generate needless diffs, making browsing the history and blaming files more difficult. Instead, let's set up a CI check on all PRs and require it to pass before merging. All changes to the spec should go through PRs anyways.

eemeli:
This would have the same effect on PRs as a CI check, because it's run on all pushes on all forks. The benefit of doing it this way is that the style corrections are automatic rather than blocking.

For example, if a PR review suggestion commit introduces some trailing whitespace to a spec PR, this would immediately fix that, rather than blocking the PR with a failing check which would need manual action.

The only way this would add a `spec/` changing commit directly on `main` is if someone else first edits it directly, without going through the PR process.

srl295:
> > This PR adds a GitHub Action that runs Prettier after every commit that touches a `.md` file and pushes any changes to the current branch with a `style: Apply Prettier` commit message.
> 
> Please don't auto-commit to `main` like this. It will generate needless diffs, making browsing the history and blaming files more difficult. Instead, let's set up a CI check on all PRs and require it to pass before merging. All changes to the spec should go through PRs anyways.

that's what CLDR and Unicodetools do.  if you forget, your PR just doesn't pass until it's fixed.

Auto committing to a PR seems reasonable.

.github/workflows/prettier.yaml
```diff
@@ -0,0 +1,26 @@
+name: Apply Prettier style
+
+on:
+  push:
+    paths:
+      - "**.md"
```
srl295, line 6: - note the job should be marked as needing write permissions https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs we've dealt with that before. 
- could be only on PRs not on main

stasm:
> This would have the same effect on PRs as a CI check, because it's run on all pushes on all forks. The benefit of doing it this way is that the style corrections are automatic rather than blocking.

Oh, do you mean that there *will* be automatic commits in the PR's branch, but since we merge by squashing, they will never make it to `main`?

srl295:
@eemeli i would recommend having the action only apply to PRs and not to main.

macchiati:
We chose in CLDR and ICU to have the equivalent to Prittier be run as a
test, and fail if the format would change. That way what the author has in
the PR is exactly what is merged in. It does place the onus on the author
to run the prettier before committing to a PR, but has a cleaner record in
github.

.github/workflows/prettier.yaml
eemeli: Added the `contents` and `pull-requests` write permissions, and rebased to force another action run to verify that those are sufficient for the action.

eemeli:
> @stasm:
> Oh, do you mean that there _will_ be automatic commits in the PR's branch, but since we merge by squashing, they will never make it to `main`?

Yes. Going forward, to submit a PR you'll need a branch here or a fork of this repo that'll include the `.github/workflows/prettier.yaml` that's added by this PR, and it'll trigger the action which will (if necessary) add a commit to the PR branch with the style fixes.

> @srl295:
> i would recommend having the action only apply to PRs and not to main.

The problem is that if/as we do occasionally commit directly to `main` e.g. most recently with 828d4cb adding meeting notes, if those commits do introduce something that doesn't match the Prettier styles, they would get picked up by all subsequent PR branches until one of them is merged -- it's much trickier to configure Prettier to only apply itself to the files actually changed on a branch, rather than all files.

Allowing commits also on `main` only really affects meeting notes and similar administrativia, not the spec. If we do want to prevent the action from running on `main`, then really we should protect the branch and require all its changes to land via PRs. This would probably inconvenience @aphillips more than anyone else; do you have an opinion on such a change?

stasm (approved):
As agreed in today's meeting, let's try this as-is and see how many auto-commits actually make it to `main`.

PR #456, opened 2023-08-19
Allow expressions without annotations to dynamically choose a function

### LINKED ISSUE #42: Automatic selection of formatter by argument type
Fluent implicitly formats arguments in three situations:

* Numbers in placeables are formatted using `NumberFormatter`
* Dates in placeables are formatted using `DateFormatter`
* Numbers in selector are formatted using `PluralRules`

The result is that the following are roughly equivalent:

```
key = { PLURAL($emailCount) ->
    [one] You have one email
  *[other] You have { NUMBER($emailCount) } emails
key2 = Today is { DATETIME($date) }
```

and

```
key = { $emailCount ->
    [one] You have one email
  *[other] You have { $emailCount } emails
key2 = Today is { $date }
```

The localizer can still decide to use the explicit formatter if they want to pass parameters:

```
key2 = Today is { DATETIME($date, month: "long") }
```

but in most cases, the implicit formatting does the job.

aphillips:
Fixes #42 

Also addresses some problems with normative terms in the same text.

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -96,7 +96,7 @@ and a _function_, _private-use_, or _reserved_ _annotation_,
 the resolved value of the _expression_ is determined as follows:
 
 If the _expression_ contains a _reserved_ _annotation_,
-an `Unsupported Expression` error is emitted and a fallback value is used as its value;
+an `Unsupported Expression` error is emitted and a fallback value is used as its value.
 
 Else, if the _expression_ contains a _private-use_ _annotation_,
 its resolved value is defined according to the implementation's specification;
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -129,8 +129,8 @@ Else, if the _expression_ consists of a _literal_,
 its resolved value is defined by _literal resolution_.
 
 > **Note**
-> This means that a _literal_ value with no _annotation_ is always treated
-> as a string.
+> This means that a _literal_ value with no _annotation_
+> is always treated as a string.
 > To represent values that are not strings as a _literal_,
 > an _annotation_ needs to be provided:
 >
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -102,7 +102,7 @@ Else, if the _expression_ contains a _private-use_ _annotation_,
 its resolved value is defined according to the implementation's specification.
 
 Else, if the _expression_ contains an _annotation_,
-its resolved value is defined by _function resolution_;
+its resolved value is defined by _function resolution_.
 
 Else, the _expression_ will contain only an _operand_,
 which consists of either a _literal_ or a _variable_.
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -111,10 +111,10 @@ If the _expression_ consists of a _variable_,
 its resolved value is defined by _variable resolution_.
 An implementation MAY perform additional processing
 when resolving the value of the _expression_.
-For example, it could apply _function resolution_ using a _function_
-and a set of _options_ chosen based on the value or type of the _variable_.
 
-> For example, given a _message_ like this:
+> For example, it could apply _function resolution_ using a _function_
+> and a set of _options_ chosen based on the value or type of the _variable_.
+> So, given a _message_ like this:
 >```
 >{Today is {$date}}
 >```
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -120,11 +120,10 @@ when resolving the value of the _expression_.
 > {Today is {$date}}
 > ```
 >
-> If the value passed in the _variable_ were a date object, such as a
-> JavaScript `Date` or a Java `java.util.Date` or `java.time.Temporal`,
+> If the value passed in the _variable_ were a date object,
+> such as a JavaScript `Date` or a Java `java.util.Date` or `java.time.Temporal`,
 > the implementation could interpret the _placeholder_ `{$date}` as if
-> the pattern included the function `:datetime` with some set of default
-> options.
+> the pattern included the function `:datetime` with some set of default options.
 
 Else, if the _expression_ consists of a _literal_,
 its resolved value is defined by _literal resolution_.
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -115,10 +115,18 @@ to further perform _function resolution_.
 
 Else, if the _expression_ consists of a _literal_,
 its resolved value is defined by _literal resolution_.
-If the _literal_ appears in a _declaration_, no further resolution is permitted.
-Otherwise, an implementation MAY use the contents of the _literal_
-to choose a _function_ and _options_ to attempt to perform _function resolution_.
-If such a _function resolution_ fails, the _literal_ value MUST be the resolved value.
+
+> **Note**
+> This means that a _literal_ value with no _annotation_ is always treated
+> as a string.
+> To represent values that are not strings as a _literal_,
+> an _annotation_ needs to be provided:
+>```
+>let $aNumber = {1234 :number}
+>let $aDate = {|2023-08-30| :datetime}
+>let $aFoo = {|some foo| :foo}
+>{You have {42 :number}}
+>```
 
 ### Literal Resolution
 
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -113,8 +113,8 @@ An implementation MAY perform additional processing
 when resolving the value of the _expression_.
 For example, it could apply _function resolution_ using a _function_
 and a set of _options_ chosen based on the value or type of the _variable_.
-> **Example**
-> Given a _message_ like this:
+
+> For example, given a _message_ like this:
 >```
 >{Today is {$date}}
 >```
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -505,14 +505,17 @@ _Formatting_ is a mostly implementation-defined process,
 as it depends on the implementation's shape for resolved values
 and the result type of the formatting.
 
-Formatting errors MAY be emitted during _formatting_,
-as formatting is not necessarily defined on every resolved value.
-A formatter MAY provide a value to use in such a case instead of a _fallback value_.
+Formatting is not necessarily defined on every resolved value.
+An implementation SHOULD emit formatting errors during _formatting_,
+but MAY provide a value to use in such a case instead of a _falback value_.
 
-_Formatting_ MAY produce formatted messages with the following data types,
-as well as any others:
+An implementation MAY use the value of an _expression_'s _operand_
+to choose a _formatting function_ and default _options_ if no
+_annotation_ is supplied by the _expression_.
 
-- A single concatenated string.
+Implementations MAY represent the result of _formatting_ using the most
+appropriate data type or structure. Some examples of these include:
+- A single string concatenated from the resolved _pattern_ and its parts.
 - A string with associated attributes for portions of its text.
 - A flat sequence of objects corresponding to each resolved value.
 - A hierarchical structure of objects that group spans of resolved values,
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -99,7 +99,7 @@ If the _expression_ contains a _reserved_ _annotation_,
 an `Unsupported Expression` error is emitted and a fallback value is used as its value.
 
 Else, if the _expression_ contains a _private-use_ _annotation_,
-its resolved value is defined according to the implementation's specification;
+its resolved value is defined according to the implementation's specification.
 
 Else, if the _expression_ contains an _annotation_,
 its resolved value is defined by _function resolution_;
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -508,8 +508,8 @@ and the result type of the formatting.
 Resolved values cannot always be formatted by a given implementation.
 When such an error occurs during _formatting_, 
 an implementation SHOULD emit a _formatting error_ and produce a 
-_formatting fallback value_ for the _placeholder_ that produced the error.
-An implementation MAY substitute a value to use instead of a _falback value_.
+_fallback value_ for the _placeholder_ that produced the error.
+A formatting function MAY substitute a value to use instead of a _fallback value_.
 
 An implementation MAY use the value of an _expression_'s _operand_
 to choose a _formatting function_ and default _options_ if no
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -115,10 +115,12 @@ when resolving the value of the _expression_.
 > For example, it could apply _function resolution_ using a _function_
 > and a set of _options_ chosen based on the value or type of the _variable_.
 > So, given a _message_ like this:
->```
->{Today is {$date}}
->```
->If the value passed in the _variable_ were a date object, such as a
+>
+> ```
+> {Today is {$date}}
+> ```
+>
+> If the value passed in the _variable_ were a date object, such as a
 > JavaScript `Date` or a Java `java.util.Date` or `java.time.Temporal`,
 > the implementation could interpret the _placeholder_ `{$date}` as if
 > the pattern included the function `:datetime` with some set of default
@@ -132,12 +134,13 @@ its resolved value is defined by _literal resolution_.
 > as a string.
 > To represent values that are not strings as a _literal_,
 > an _annotation_ needs to be provided:
->```
->let $aNumber = {1234 :number}
->let $aDate = {|2023-08-30| :datetime}
->let $aFoo = {|some foo| :foo}
->{You have {42 :number}}
->```
+>
+> ```
+> let $aNumber = {1234 :number}
+> let $aDate = {|2023-08-30| :datetime}
+> let $aFoo = {|some foo| :foo}
+> {You have {42 :number}}
+> ```
 
 ### Literal Resolution
 
@@ -542,13 +545,14 @@ as it depends on the implementation's shape for resolved values
 and the result type of the formatting.
 
 Resolved values cannot always be formatted by a given implementation.
-When such an error occurs during _formatting_, 
-an implementation SHOULD emit a _formatting error_ and produce a 
+When such an error occurs during _formatting_,
+an implementation SHOULD emit a _formatting error_ and produce a
 _fallback value_ for the _placeholder_ that produced the error.
 A formatting function MAY substitute a value to use instead of a _fallback value_.
 
 Implementations MAY represent the result of _formatting_ using the most
 appropriate data type or structure. Some examples of these include:
+
 - A single string concatenated from the parts of the resolved _pattern_.
 - A string with associated attributes for portions of its text.
 - A flat sequence of objects corresponding to each resolved value.
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -109,9 +109,20 @@ which consists of either a _literal_ or a _variable_.
 
 If the _expression_ consists of a _variable_,
 its resolved value is defined by _variable resolution_.
-An implementation MAY use the value of the _variable_ or
-its resolved value to choose a _function_ and _options_
-to further perform _function resolution_.
+An implementation MAY perform additional processing
+when resolving the value of the _expression_.
+For example, it could apply _function resolution_ using a _function_
+and a set of _options_ chosen based on the value or type of the _variable_.
+> **Example**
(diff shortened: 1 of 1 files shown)


spec/formatting.md
```diff
@@ -505,14 +505,17 @@ _Formatting_ is a mostly implementation-defined process,
...
-A formatter MAY provide a value to use in such a case instead of a _fallback value_.
+Formatting is not necessarily defined on every resolved value.
+An implementation SHOULD emit formatting errors during _formatting_,
+but MAY provide a value to use in such a case instead of a _falback value_.
 
-_Formatting_ MAY produce formatted messages with the following data types,
-as well as any others:
+An implementation MAY use the value of an _expression_'s _operand_
+to choose a _formatting function_ and default _options_ if no
+_annotation_ is supplied by the _expression_.
```
eemeli, line 514: I think this belongs in the Variable Resolution section, not here. As I understand it, we want to make it possible for something like `{$num}` and `{$num :number}` to format the same, if `$num` is a number.

To achieve that, it's much easier to say that an implementation may apply a default annotation to a variable reference that refers to an external variable, depending on its type and value.

If we add this here, then a conforming implementation would need to include in its resolved-value shape information about each value's original expression shape, which is otherwise unnecessary. That's an unnecessary restriction on implementations.

eemeli (changes requested):

spec/formatting.md
```diff
@@ -505,14 +505,17 @@ _Formatting_ is a mostly implementation-defined process,
 as it depends on the implementation's shape for resolved values
 and the result type of the formatting.
 
-Formatting errors MAY be emitted during _formatting_,
-as formatting is not necessarily defined on every resolved value.
-A formatter MAY provide a value to use in such a case instead of a _fallback value_.
+Formatting is not necessarily defined on every resolved value.
+An implementation SHOULD emit formatting errors during _formatting_,
+but MAY provide a value to use in such a case instead of a _falback value_.
```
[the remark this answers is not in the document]
aphillips: You're right that the proposed normative sentence doesn't make sense.

The problem with the original sentence is that it is not usefully normative: "MAY" is trying to permit formatting errors to be emitted during formatting, but when else would formatting errors occur if not during formatting? As a direction to implementers it's not useful and it's not testable.

My proposal, of course, is worse, because it seems to imply that every formatting operation ought to produce errors 😭.

spec/formatting.md
aphillips: I thought about that initially. One challenge is the need to do a bunch of changes if we do that. Currently the parent "expression resolution" section says:

> Depending on the presence or absence of an operand and a function, private-use, or reserved annotation, one of the following is used to resolve the value of the expression:
> *  If the expression contains no annotation, its resolved value is determined by literal resolution or variable resolution, depending on the shape of the operand.
> *  Else, if the expression has a function annotation, its resolved value is defined by function resolution.
> * _... more else clauses dealing with private/reserved..._

I think that we can't ignore literals. It's one thing if a literal is just a blob of text, but what about:

```
let $when = {|2023-07-15T00:00:00Z|}
let $howMany = {4200}
{You bought {$howMany} on {$when}}
```

You might say that this is variable resolution, but then what about:

```
{You bought {4200} on {2023-07-16}}
```

If we want to allow (but not force) implementations to infer a type and thus a formatter, we would need to recast literal and variable resolution to pass through function resolution. I suspect this might be the right thing to do anyway (the ultimate default formatter is the one that places stringified literals or stringified variables) rather than having them be separate paths as they are now.

spec/formatting.md
```diff
@@ -505,14 +505,19 @@ _Formatting_ is a mostly implementation-defined process,
...
-as well as any others:
-
-- A single concatenated string.
+Resolved values cannot always be formatted by a given implementation.
+When such an error occurs during _formatting_, 
+an implementation SHOULD emit a _formatting error_ and produce a 
+_formatting fallback value_ for the _placeholder_ that produced the error.
+An implementation MAY substitute a value to use instead of a _falback value_.
+
+An implementation MAY use the value of an _expression_'s _operand_
```
mihnita, line 514: > implementation MAY use the value of an
implementation MAY use the type of an
?

spec/formatting.md
mihnita: I think that a literal without a function is just the literal, there is no good way to infer a function for it.
At least not "in the engine" (in MF2 code itself), while still remaining function agnostic.

spec/formatting.md
aphillips: We don't have types. Implementations might have types. But that type would naturally be bound up in the operand's value (it would be the value's _type_ 😸)

spec/formatting.md
aphillips: @mihnita Fair enough. But remember that the keys of a selector are literals:

```
when 0 {This is the =0 message}
```

If a literal is always and *only* a character sequence, how can I assign a number to a variable? A date?

Of course there are sticky issues. What does this print?

```
{That is {NaN :number}}  <- CLDR has data for formatting this
```

eemeli: I think we should require an explicit function for formatting literals; by default they should always end up as strings.

> If a literal is always and _only_ a character sequence, how can I assign a number to a variable? A date?

By having a `:number` or `:datetime` function that can accept string input. For example `{42 :number}`.

> Of course there are sticky issues. What does this print?
> 
> ```
> {That is {NaN :number}}  <- CLDR has data for formatting this
> ```

This is implementation-dependent behaviour, unless we define the supported input values of `:number` rather explicitly. With the JS proposal in its current shape, you'd end up with `That is {|NaN|}` because it only accepts JSON numerical values for string input.

spec/formatting.md
```diff
@@ -505,14 +505,19 @@ _Formatting_ is a mostly implementation-defined process,
...
-
-_Formatting_ MAY produce formatted messages with the following data types,
-as well as any others:
-
-- A single concatenated string.
+Resolved values cannot always be formatted by a given implementation.
+When such an error occurs during _formatting_, 
+an implementation SHOULD emit a _formatting error_ and produce a 
+_formatting fallback value_ for the _placeholder_ that produced the error.
+An implementation MAY substitute a value to use instead of a _falback value_.
```
eemeli, line 512:
```suggestion
_fallback value_ for the _placeholder_ that produced the error.
A formatting function MAY substitute a value to use instead of a _fallback value_.
```

I'm pretty sure that we don't need to specify _formatting fallback value_, but just _fallback value_.

Also, the intent of the original text was to allow for a formatting function to supply a substitution. By default we avoid revealing any runtime values as doing so could have security concerns, but a formatter that's confident about its input could opt to use a different fallback.

For instance, a dynamic message reference formatter could use the resolved message identifier as its fallback value rather than the name of the variable. This would allow a missing message to format as something like `{action.cancel}` rather than `{$msgId}`.

eemeli (changes requested):

spec/formatting.md
aphillips: The difference here is something else, though. The original text said "an implementation", which means the message formatting engine. My text says "a formatting function", which (a) allows the function to provide an alternate fallback value (just as your comment spells out) and (b) means that implementations have to provide the ability for a formatting function to supply the fallback value. 

In fact, that seems like a potential requirement or a potential option, so I'll clarify the text some more.......

eemeli: The original text actually said "formatter"; your PR currently says "implemantation"; it's only my suggestion above that uses "formatting function".

spec/formatting.md
aphillips: We can _allow_ implementations to "cast" literals that they recognize. For many implementations and users this will seem natural and many implementations will probably do mainly the right thing here. The alternative, as you point out @eemeli, would be to require non-literals to use an annotation.

This is, to some degree, a loss of functionality for strongly typed languages vs. MF1, where a pattern like `You have {0}` will format a `Number` or a `Date` (or its `Temporal` or `Calendar` friends) as those types when they are passed and not just call `toString()`.

The prose I'm proposing would allow (MAY) this but not require it. Do you object to that?

spec/formatting.md
aphillips: My bad. I'm reading this comment wrong.

spec/formatting.md
eemeli: At the moment, we say this [in the spec](https://github.com/unicode-org/message-format-wg/blob/main/spec/formatting.md#literal-resolution):
> The resolution of a _text_ or _literal_ token MUST always succeed.

If we allow for casting literals to non-string types, we'll need to remove that assertion, and we will no longer be able to say with certainty e.g. that this message will always format to the same string in all implementations:

```
{Enter code {12345} to continue}
```

That could come out as e.g. any of the following, even in just "English":
- Enter code 12 345 to continue
- Enter code 12,345 to continue
- Enter code 12.345 to continue

> This is, to some degree, a loss of functionality for strongly typed languages vs. MF1, where a pattern like `You have {0}` will format a `Number` or a `Date` (or its `Temporal` or `Calendar` friends) as those types when they are passed and not just call `toString()`.

This is not the case. What we do with unannotated literals does not affect what we do with unannotated variable references. If a variable resolves to a number, we can use number formatting for it. Same if it's a Date, or any other value type that the implementation has defined handling for.

> The prose I'm proposing would allow (MAY) this but not require it. Do you object to that?

Yes, because I think that allowing such value casts adds unnecessary uncertainty without a clear and significant benefit.

mihnita: > @mihnita Fair enough. But remember that the keys of a selector are literals:

But they have a function that looks at them and interprets them.
That is the matched function.

In the `when 0 {This is the =0 message}` it is the `plural` function.
As the matcher functions take the key(s), the input, and tell us if they match or not.
So it is still a function doing the interpretation and the casting.
And if one implements a `choice` matcher, it would be that function looking at the key in `when |<10| {...}`

mihnita: > let $x = 12345

`$x` contains a literal at this point.

> let $x = $countFromArgument

`$x` might contain a number here. If the programming languages have some kind of type system.

Note that even MF1 for C has type information here, as the real values to be formatted are wrapped in a 
[`Formattable`](https://unicode-org.github.io/icu-docs/apidoc/released/icu4c/classicu_1_1Formattable.html)

eemeli: > > If we allow for casting literals to non-string types, we'll need to remove that assertion
> 
> I disagree, I think. The _resolution_ of a text or literal would always succeed, but the formatting might be different. These are separate operations.

We're getting into semantics, but you are indeed right if we consider the string-without-pattern-matching-applied as "resolved". In other words, we could leave the assertion in, but we would need to add in a much more restrictive definition of "resolved value" than we currently have, because that could not include the resolution of a string `'12345'` as a numerical value.

> > we will no longer be able to say with certainty e.g. that this message will always format to the same string in all implementations
> 
> It is already the case that the same message might not format to the same string in all implementations--because, for example, they are running different CLDR versions or the locale resolves slightly differently on a given platform.

In general yes, but my point was about the specific string including an explicit literal. It does currently format to the same string in all possible implementations.

> The example you provided with `{12345}` is a good one, since it highlights the problem of untyped or weakly typed values, which this proposal brings to the fore.

In my understanding, the value represented in `{12345}` is always a string, and only ever a string. Its value could be parsed as a number, but it's still a string.

> > > This is, to some degree, a loss of functionality for strongly typed languages vs. MF1
> >
> > This is not the case
>
> No, I'm pretty sure it is the case. MF1 lets you write simple patterns that don't declare a formatter and uses reflection to assign the formatter at runtime, e.g. `{0}` formats a `Number`, `Date`, or `String` differently and using the appropriate formatter. Whether that capability is important is what we're discussing 😉.

Just as a sanity check, I'm presuming that the `{0}` you're using is an MF1 reference to an external variable, yes? If so, that's completely separate from this discussion on literal values that are wholly and completely defined in the message syntax.

> Wouldn't you feel a bit of surprise if you didn't get the same behavior when you did a local assignment?

Not if we make it clear that the value of an unannotated literal is always a string.

> On the one hand, I do think that requiring the annotation is super helpful (particularly to translators, who cannot see what type is being passed), but that suggests that we require an annotation all the time.

The we should also define a `:string` function.

> The question of clear and significant benefit is whether developers can get "do what I mean" behavior. We should seek group consensus here.

Agreed.

macchiati:
One other consideration: we should discuss whether enabling the conversion
of MF1 messages to MF2 messages is a desired feature. I suspect the
biggest barrier to that would be if a conformant implementation *cannot*
perform automatic choice of functions, with strongly typed languages.

aphillips:
@macchiati We discussed this in teleconference today. Implementations will be allowed to choose how to process values based on type without having a specific annotation, so that level of compatibility will be present.

aphillips:
Here is the first cut at adopting the consensus from the 2023-08-28 call. I suspect that examples are needed. In particular, while the normative text allows implementations to make the kind of inference we're talking about here, it might not be clear to readers. 

Let's make the normative gunk first, although I welcome suggestions for how to make this clear. I'm going to toy with some examples.

spec/formatting.md
```diff
@@ -93,17 +93,32 @@ and different implementations MAY choose to perform different levels of resoluti
...
+An implementation MAY use the value of the _variable_ or
+its resolved value to choose a _function_ and _options_
+to further perform _function resolution_.
+
+Else, if the _expression_ consists of a _literal_,
+its resolved value is defined by _literal resolution_.
+If the _literal_ appears in a _declaration_, no further resolution is permitted.
+Otherwise, an implementation MAY use the contents of the _literal_
+to choose a _function_ and _options_ to attempt to perform _function resolution_.
+If such a _function resolution_ fails, the _literal_ value MUST be the resolved value.
```
eemeli, line 121:
```suggestion
```

I don't think this reflects the consensus achieved during the call; rather the opposite.
https://github.com/unicode-org/message-format-wg/blob/a53a0619a7dfcb0ce5cec5b1eb99e323d7dbbc85/meetings/2023/notes-2023-08-28.md?plain=1#L259-L263

spec/formatting.md
```diff
@@ -93,17 +93,32 @@ and different implementations MAY choose to perform different levels of resoluti
...
+its resolved value is defined by _function resolution_;
+
+Else, the _expression_ will contain only an _operand_,
+which consists of either a _literal_ or a _variable_.
+
+If the _expression_ consists of a _variable_,
+its resolved value is defined by _variable resolution_.
+An implementation MAY use the value of the _variable_ or
+its resolved value to choose a _function_ and _options_
+to further perform _function resolution_.
```
eemeli, line 114: What does "the value of the _variable_ or its resolved value" mean?

Furthermore, this language implicitly restricts an implementation from providing custom behaviour for values e.g. based on their type that is not explicitly made available via _functions_ and _options_. As I argued on the call:
https://github.com/unicode-org/message-format-wg/blob/a53a0619a7dfcb0ce5cec5b1eb99e323d7dbbc85/meetings/2023/notes-2023-08-28.md?plain=1#L345-L349

spec/formatting.md
```diff
@@ -505,14 +520,15 @@ _Formatting_ is a mostly implementation-defined process,
...
+Resolved values cannot always be formatted by a given implementation.
+When such an error occurs during _formatting_, 
+an implementation SHOULD emit a _formatting error_ and produce a 
+_fallback value_ for the _placeholder_ that produced the error.
+A formatting function MAY substitute a value to use instead of a _fallback value_.
 
-- A single concatenated string.
+Implementations MAY represent the result of _formatting_ using the most
+appropriate data type or structure. Some examples of these include:
+- A single string concatenated from the resolved _pattern_ and its parts.
```
eemeli, line 531:
```suggestion
- A single string concatenated from the parts of the resolved _pattern_.
```

I find it odd to refer separately to the _pattern_ and the _parts_. That implies that the _pattern_ includes something that's not a _part_ but does affect the result.

eemeli (changes requested):

spec/formatting.md
aphillips: Done, but with a note added.

spec/formatting.md
aphillips: "value or resolved value" is to allow for declaration chaining

```
let $foo = {$someExternalVariableThatIsOfTypeNumber}
let $bar = {$foo}
let $baz = {$bar}
{{$baz} can be a number}
```

How about:

```suggestion
An implementation MAY use the resolved value of a _variable_
to further perform _function resolution_.
```

Note that I used my original permissive language about choosing a function and options because it doesn't make much sense to "apply function resolution" if there is no function! This doesn't forbid an implementation from applying unspecified resolution, such as your person object example. Although, honestly, if you have a `com.foo.person` object and have a "PersonFormatter" function, isn't that "registering the person formatter function"? My text does **not** say "... choose an _annotation_" or "...apply an annotation". It goes directly to the function and its options. I think that leap removes the need for a `:person` keyword. Do we need to say that explicitly?

eemeli: The suggestion is better, but still problematic. The explicit reference to _function resolution_ mean that this step of that process needs to be considered: https://github.com/unicode-org/message-format-wg/blob/92595639fe284c55608b2272b72eeccd8d33a793/spec/formatting.md?plain=1#L147-L151

In other words, this language still requires for something like `:person` to be addressable for any special handling to be assignable to a `com.foo.person` object.

I would prefer something more like this, without an explicit _function resolution_ reference:

```suggestion
An implementation MAY select and apply additional processing
based on the resolved value of a _variable_,
and use the result of such processing as the resolved value of the _expression_.
```

aphillips: What you're suggesting is that if I have a pattern like this:

```
{{$person}}
```

... that an implementation can decide, based on the value of the variable `$person` to supply a formatter function (along with whatever options, i.e. arguments that such a function takes), even though it does not appear in the implementation's "registry" of functions. Note that the registry might not exist as a physical file for a given implementation. 

That seems like a weird feature to have, since it can only be accessed by passing objects that the implementation recognizes and can't be controlled (the pattern can't pass any options). I suppose the formatter could be externally configured, sort of like how ICU4J let's you set a specific formatter in MF1 (generally icky).

In any case, I'm not against allowing some interpolation, but I think gently nudging folks to functions and options in their registry makes sense--if only as a way of making what we're saying understandable. 

How about:

```suggestion
An implementation MAY perform additional processing
when resolving the value of a _variable_ based on the value of the _variable_.
For example, it could apply _function resolution_ using a _function_
and a set of _options_ chosen based on the value or type of the _variable_.
```

spec/formatting.md
```diff
@@ -93,17 +93,54 @@ and different implementations MAY choose to perform different levels of resoluti
...
+> the implementation could interpret the _placeholder_ `{$date}` as if
+> the pattern included the function `:datetime` with some set of default
+> options.
+
+Else, if the _expression_ consists of a _literal_,
+its resolved value is defined by _literal resolution_.
+
+> **Note**
+> This means that a _literal_ value with no _annotation_ is always treated
+> as a string.
```
eemeli, line 134:
```suggestion
> This means that a _literal_ value with no _annotation_
> is always treated as a string.
```

eemeli (approved):
I added a merge from `main` to fix the Prettier action. A small bunch of stylistic fixes inline; otherwise this starts to look like an improvement on what we've currently got.

spec/formatting.md
```diff
@@ -93,17 +93,54 @@ and different implementations MAY choose to perform different levels of resoluti
...
-- Else, if the _expression_ has a _function_ _annotation_,
-  its resolved value is defined by _function resolution_.
-- Else, if the _expression_ has a _private-use_ _annotation_,
-  its resolved value is defined according to the implementations's specification.
-- Else, the _expression_ has a _reserved_ _annotation_,
-  an Unsupported Expression error is emitted and a fallback value is used as its value.
+the resolved value of the _expression_ is determined as follows:
+
+If the _expression_ contains a _reserved_ _annotation_,
+an `Unsupported Expression` error is emitted and a fallback value is used as its value;
```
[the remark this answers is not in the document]
aphillips: I used the semi-colons on purpose, as a style that is similar to many if-else "algorithms" in other specs, but periods work just as well in our current formatting.

stasm (commented):
Catching up with the developments of the past two weeks. LGTM :)

PR #457, opened 2023-08-22
Add some capabilities for the builtin functions

ryzokuken:
As a precursor to finalizing and standardizing the set of builtin functions, we should document the "goals" of the builtin API.

This PR aims to kick off that process by adding a simple list of capabilities that we aim to expose to users. This list is close to the set of functions exposed by the ECMA-402 API and a superset of the API documented by #420. Some notable exceptions include:

* The list of formatters excludes relative time formatting and durations. It is my understanding that `RelativeTimeFormat` was never the most elegant part of the Intl API and its future use would be discouraged. Durations could be added as a follow-on, but I wasn't sure about it.

* The list of selectors include a "gender" selector. This functionality doesn't exist in 402 and I didn't yet investigate whether it's exposed by ICU. That said, it was a recurring theme in a number of documents I read (in this repository, I think?) and I felt it might be uniquely useful within the context of MessageFormat so I decided to include it here but don't feel very strongly either way. 

This is super WIP so feel free to tear it apart but I hope it's a decent launch-off point for our work.


ryzokuken:
Will figure out what's the best way to deal with the CLA.

spec/registry.md
```diff
@@ -186,3 +186,17 @@ which only expects the `accord` option:
 
     let $obj = {$object :noun case=nominative}
     {You see {$color :adjective article=indefinite accord=$obj} {$obj}!}
+
+## Built-in Registry
+
+As hinted in the example above, implementations may <!-- FIXME: may vs should --> provide built-in functions to their users in order to perform common operations in a standardized manner instead of having to rely on custom functions for everything. The working group agrees that these builtins provided by implementations should expose the following capabilities to their users:
```
aphillips, line 192: The default registry will contain function signatures that every implementation is **_required_** to provide. So this is neither MAY nor SHOULD: it is MUST.

Note that RFC2119 keywords MUST be capitalized.

You should think of this document as being part of the specification. This section shouldn't refer to "the working group" or what our thoughts are. Adopt the point of view of an implementer ("what do I have to do here?") or a user ("how do I use this?")

```suggestion
The **_<dfn>built-in registry</dfn>_** is the set of _functions_, together with
their options, that are present in every implementation.
Implementations MUST provide each of the built-in functions defined in the 
_built-in registry_, including all of the options.
They MAY include additional options, although implementers are cautioned to
avoid options which might conflict with future standardization.

These include the following capabilities:
```

aphillips: Please provide line breaks.

spec/registry.md
```diff
@@ -186,3 +186,17 @@ which only expects the `accord` option:
...
+
+* Ability to format the following types of data in a locale-sensitive manner:
+    * Dates and times
+    * Numbers
+    * Lists
+* Ability to select for the following types of information in a locale-sensitive manner:
+    * Plural rules
+    * Gender
```
aphillips, line 200: Probably not gender in this version. This is not yet standardized in CLDR and there is a question about whether to also handle grammatical case.

spec/registry.md
```diff
@@ -186,3 +186,17 @@ which only expects the `accord` option:
...
+
+## Built-in Registry
+
+As hinted in the example above, implementations may <!-- FIXME: may vs should --> provide built-in functions to their users in order to perform common operations in a standardized manner instead of having to rely on custom functions for everything. The working group agrees that these builtins provided by implementations should expose the following capabilities to their users:
+
+* Ability to format the following types of data in a locale-sensitive manner:
+    * Dates and times
+    * Numbers
+    * Lists
+* Ability to select for the following types of information in a locale-sensitive manner:
```
aphillips, line 198: A key thing about selectors: they select between different pattern strings. Perhaps:

```suggestion
* Ability to select between pattern strings in a locale-sensitive way based on:
* ```

File: spec/registry.md [hunk 4]
```diff
@@ -186,3 +186,17 @@ which only expects the `accord` option:
...
+    * Plural rules
```
aphillips, line 199: The "basic" or default selector (whose name we are discussing in #433) is what ICU4J calls `SelectFormat`. That should be listed here.

Try to make a list of items that could be selectors. Maybe start by making a _long_ list of everything and we can try to triage the list by crossing things out (rather than trying to remember if we decided not to include it... 😉). My back of envelope for selectors would maybe look like:

* select/equals (string/value equality; can be used to build things like a poor person's gender selector)
* plural
* ordinal

Potentially below the line:
* grammatical gender
* case
* date/time query (before/after/has-field-value)
* choice (see Java's `ChoiceFormat`)

spec/registry.md
```diff
@@ -186,3 +186,17 @@ which only expects the `accord` option:
...
+    * Gender
+
+The co-located `registry.xml` file aims to define such a set of built-in functions, inspired by the design of the ICU4X and JavaScript's ECMA-402 APIs.
```
aphillips, line 202: This belongs in the design doc. Note that compatibility with MF1 in ICU4J/Java is kind of important here as well.

aphillips (changes requested):
Thank you for starting on this!!

In the call this week we established a new bit of process. I think this PR would be better off as a design document. See [here](https://github.com/unicode-org/message-format-wg/blob/main/exploration/0000-design-proposal-template.md) for the template. Create a new proposal using the template under /main/exploration and that'll be easier to discuss. As you can see from my comments below, trying to be "specification like" and working on what to put in the built-in registry doubles the effort.

ryzokuken:
Okay moved the text to the exploration dir, made some structural changes and tried to incorporate your comments.

@aphillips how does it look?

exploration/0001-builtin-registry-capabilities.md
```diff
@@ -0,0 +1,52 @@
...
+
+<details>
+	<summary>Metadata</summary>
+	<dl>
+		<dt>Contributors</dt>
+		<dd>@ryzokuken</dd>
+		<dt>First proposed</dt>
+		<dd>2023-08-22</dd>
+		<dt>Pull Request</dt>
+		<dd>#457</dd>
```
eemeli, line 11:
```suggestion
		<dd><a href="https://github.com/unicode-org/message-format-wg/pull/457">#457</a></dd>
```

exploration/0001-builtin-registry-capabilities.md
```diff
@@ -0,0 +1,52 @@
...
+	</dl>
+</details>
+
+## Objective
+
+Document the capabilities that the set of built-in functions would expose to the user in order to ensure it can be designed in the most minimalistic way that would still cater to the most pressing requirements, partly inspired by ICU4X anbd JavaScript's Intl API and backwards-compatible with MF1 in ICU4J.
```
eemeli, line 17:
```suggestion
Document the capabilities that the set of built-in functions would expose to the user in order to ensure it can be designed in the most minimalistic way that would still cater to the most pressing requirements, partly inspired by ICU4X and JavaScript's Intl API and backwards-compatible with MF1 in ICU4J.
```

exploration/0001-builtin-registry-capabilities.md
```diff
@@ -0,0 +1,52 @@
...
+All conforming implementations of this specification MUST provide a set of <dfn>built-in functions</dfn>, which allows the developers to perform the following operations:
+
+* Formatting the following types of data in a locale-sensitive manner:
+    * Dates and times
+    * Numbers
+    * Lists
+* Selecting between pattern strings in a locale-sensitive manner based on:
+	* String equality
+    * Plural rules
+    * Ordinality
```
eemeli, line 48: Isn't ordinality a subset of plural rules? I know that MF1 has separately `plural` and `selectordinal`, but wouldn't it make sense for our plural selector to include something like `type=ordinal` as an option instaed?

eemeli (changes requested):
For some reason the prettier formatter hasn't run on this branch, and that'll need to be fixed. Otherwise, some minor issues inline.

exploration/0001-builtin-registry-capabilities.md
ryzokuken: In my basic understanding of linguistics, there's not a whole lot of overlap between the two concepts. We could obviously decide to design the API either way, I have no strong opinions but we should err on the side of not confusing the user between two otherwise distinct linguistic concepts.

aphillips: @eemeli Yes, you're correct and we might cover it using an option. But it is good to ensure we remember it on the checklist of things we need to build in.

ryzokuken:
The prettier action seems to be broken, here and elsewhere (#456). Should I find an issue about it?

eemeli:
> The prettier action seems to be broken, here and elsewhere (#456). Should I file an issue about it?

See #464.

ryzokuken:
I don't have the permissions to rerun this job but now it should have access to my branch :crossed_fingers:

ryzokuken:
Weird, looks like prettier behaves differently but either way, this should be good now. :D

exploration/0001-builtin-registry-capabilities.md
```diff
@@ -0,0 +1,54 @@
...
+All conforming implementations of this specification MUST provide a set of <dfn>built-in functions</dfn>, which allows the developers to perform the following operations:
+
+- Formatting the following types of data in a locale-sensitive manner:
+  - Dates and times
+  - Numbers
+  - Lists
+- Selecting between pattern strings in a locale-sensitive manner based on:
+  - String equality
+    - Plural rules
+    - Ordinality
```
eemeli, line 50:
```suggestion
  - String equality
  - Plural rules
  - Ordinality
```

eemeli (approved):
Minor indent fix inline, but otherwise looks good to land as "proposed". So this is conditional approval presuming that #461 is accepted first; otherwise all the TBD needs to be D.

ryzokuken:
@eemeli dunno why prettier did that but thanks for the fix! I'm assuming that we're happy with #461 and that we can flesh out this document further after merging but in the meantime I'll start working on the remaining sections

ryzokuken:
Added some pretty tentative text for the other sections so that atleast all sections exist.

ryzokuken:
@eemeli @aphillips sorry for dropping the ball on this somewhat. I've now put in some improvements that I felt brought the doc closer to what you expected and applied some of Addison's comments. Please feel free to let me know if there's anything else here that needs rewording or polishing.

exploration/0001-builtin-registry-capabilities.md
```diff
@@ -16,31 +16,35 @@ Status: **Proposed**
...
+Like many standardized APIs out there, the biggest constraint on the resulting set of _built-in functions_ would be that until we decide to make another major update to MessageFormat, we'd not be able to make breaking changes to the function signatures since that would invalidate the promise of a consistent API across implementations and versions.
 
 Another constraint is that since MessageFormat would be used alongside other internationalization features in many programming environments, the final interface should be somewhat consistent to reduce developer confusion.
 
 ## Proposed Design
 
-All conforming implementations of this specification MUST provide a set of <dfn>built-in functions</dfn>, which allows the developers to perform the following operations:
+The **_<dfn>built-in registry</dfn>_** is the set of _functions_, together with their options, that are present in every implementation.
+Implementations MUST provide each of the built-in functions defined in the _built-in registry_, including all of the options.
+They MAY include additional options, although implementers are cautioned to avoid options which might conflict with future standardization.
```
eemeli, line 46: This is fine, but could include a clarification (potentially as a later change) that the exact output produced for any of these functions or options is left up to each implementation.

So for example a conformant implementation could "support" the `notation` option for `:number` by not throwing an error if it's used, but producing the same output no matter what the value the option has.

exploration/0001-builtin-registry-capabilities.md
```diff
@@ -53,4 +57,4 @@ All conforming implementations of this specification MUST provide a set of <dfn>
 
 ## Alternatives Considered
 
-The alternative would be to not include any built-in functions, essentially outsourcing the task of curating and maintaining commonly used functions to the ecosystem. The details of how this could work out in practice are unclear.
+The alternative would be to not include any _built-in functions_, essentially outsourcing the task of curating and maintaining commonly used functions to the ecosystem.
```
eemeli: Could also list as an alternative the preceding form of this PR, which defines but does not require the interfaces.

eemeli (approved):

exploration/0001-builtin-registry-capabilities.md
```diff
@@ -58,3 +59,4 @@ These include the following capabilities:
 ## Alternatives Considered
 
 The alternative would be to not include any _built-in functions_, essentially outsourcing the task of curating and maintaining commonly used functions to the ecosystem.
+Another alternative would be to specify the signatures of the functions in a default registry for interoperability between implementations but not require implementations to implement all of them.
```
aphillips, line 62: This could use more work.

```suggestion
The working group has considered a number of ways to address core functionality. These include:

**No default registry** Each implementation would be free to define
its own set of functions and options for each.

- (+) Each implementation would be able to define functions
  and options according to existing I18N APIs
  which would be familiar to users in that language or runtime
- (-) Messages would not be portable between implementations
- (-) Users would have to learn the variations between implementations
- (-) Tooling would have to be adjusted for each variation

**Informative registry** The default registry could be made informative
with implementations allowed to pick-and-choose (or ignore) entries.
This would provide better interoperability for implementations 
claiming to implement the default registry, while allowing platform-specific
variation.

**Non-extensible registry entries.** Like this design, but not allowing
implementation specific options. 
 
```

aphillips (changes requested):

ryzokuken:
@aphillips thanks for all the suggestions and apologies for the delay. Some of your suggestions I applied as-is, but mostly I tweaked the draft to make it closer to what I think you expected. Let me know what you think

exploration/0001-builtin-registry-capabilities.md
```diff
@@ -0,0 +1,88 @@
...
+
+From formatting date and time values to performing selection based on plural rules, the use cases for a minimal set of
+common tools standardized across implementations are virtually everywhere.
+While a number of applications might require more complex functions with custom logic, a well-provisioned default
+registry could significantly lower the barrier of entry as well as the need to write and maintain many custom functions.
+
+## Requirements
+
+Implementations are required to implement and expose the entire _built-in registry_ that would provide the capabilities
+in this document.
```
aphillips, line 41: The requirements section should be requirements on this design, rather than what this design requires. A comprehensive set of requirements could be rather extensive, but it's still a good idea to spend the effort. One way to do that is to populate this section as you write the design.

exploration/0001-builtin-registry-capabilities.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Constraints
+
+Like many standardized APIs out there, the biggest constraint on the resulting _built-in registry_ would be that until
+we decide to make another major update to MessageFormat, we'd not be able to make breaking changes to the function
+signatures since that would invalidate the promise of a consistent API across implementations and versions.
```
aphillips, line 47: In some ways, this is too weak. It also belongs in requirements.

```suggestion
The names of functions and associated options defined by the registry
will be embedded into messages across the MF2 ecosystem.
These function signatures **_MUST_** be stable and cannot be changed
later.
```

exploration/0001-builtin-registry-capabilities.md
```diff
@@ -0,0 +1,88 @@
...
+
+Another constraint is that since MessageFormat would be used alongside other internationalization features in many
+programming environments, the final interface should be somewhat consistent to reduce developer confusion.
```
aphillips, line 50: Not sure what the constraint is exactly?

exploration/0001-builtin-registry-capabilities.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Proposed Design
+
+The **_<dfn>built-in registry</dfn>_** is the set of functions, together with their options, that are present in every
+implementation.
+Implementations MUST provide each of the functions defined in the _built-in registry_, including all of the options.
+The ouptut of each function is implementation-defined.
```
aphillips, line 57:
```suggestion
The output of each function is implementation-defined.
```

(typo)

exploration/0001-builtin-registry-capabilities.md
```diff
@@ -0,0 +1,88 @@
...
+They MAY include additional options, although implementers are cautioned to avoid options which might conflict with
+future standardization.
```
aphillips, line 59: tweak

```suggestion
Implementations MAY include additional options for built-in functions,
although implementers are cautioned to avoid adding options
that might conflict with future standardization.
```

aphillips (changes requested):
Looking good. Some minor tweaks.

The use cases and requirements need to be fleshed out as we go.

@eemeli has started work on the number selection design in another PR, so maybe don't focus on that.

I think one really interesting area we need to work out are the options for numbers and for date/time values. ICU has developed "skeletons" for both of these, but JS generally has not adopted skeletons in favor of (somewhat verbosely named) bags of options. We have some consensus to use the option bags, but we should spell out the options for each and then check that it matches what both JS and ICU provide.

I think it would be interesting to consider providing an **_optional but defined_** set of options for skeletons. That is, not require an implementation to support them, but require implementations that do support them to define them using our definition.

Anyway, fix the typos and then (if the group concurs) we can merge for further work.

ryzokuken:
@aphillips applied your straightforward suggestions and rephrased the remaining item in "constraints". Some of the sections are far from ideal indeed but I do feel that fleshing things out as we go would be the mode I'd opt for nonetheless since this would just be a "proposed" draft at the moment.

aphillips (approved):

aphillips:
@ryzokuken thanks!

Let's merge this and iterate on it.

PR #459, opened 2023-08-27
Limit Prettier action to selected directories & events

eemeli:
It turns out that our `main` branch is in fact protected, and that the action can't write to it.

So rather than trying to fix the styles everywhere, this limits the action to just `docs/`, `exploration/` and `spec/`.

It should also now run for `pull_request` events, in case a PR is made from a fork that isn't allowing actions.


aphillips (approved):
Thanks for debugging this.

PR #461, opened 2023-08-28
Add proposed/accepted status to design docs

eemeli:
We should track and signal the status of a design document, so that we can iterate on them across multiple PRs. To keep it simple, we probably don't need more than two statuses to start with, "proposed" and "accepted".

CC @ryzokuken, who also has a design doc PR in flight.

### DIFF
--- a/exploration/0000-design-proposal-template.md
+++ b/exploration/0000-design-proposal-template.md
@@ -2,7 +2,9 @@
 
 Status: **Accepted**
 
-_Use either **Proposed** or **Accepted** as the status._
+_Use the status **Proposed** when creating a design proposal._
+_The status will be changed to **Accepted** when the design proposal is accepted as working group consensus._
+_The status will be changed to **Rejected** if the design proposal ultimately is not adopted by the working group._
 _This is intended to allow new design documents to be first proposed and then iterated over time,_
 _rather than needing to have all their details agreed in a single PR._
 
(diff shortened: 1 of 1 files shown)


exploration/0000-design-proposal-template.md
```diff
@@ -1,5 +1,11 @@
 # Design Proposal Template
 
+Status: **Accepted**
+
+_Use either **Proposed** or **Accepted** as the status._
```
aphillips, line 5: I would say:

```suggestion
_Use the status **Proposed** when creating a design proposal._
_The status will be changed to **Accepted** when the design proposal is accepted as working group consensus._
_The status will be changed to **Rejected** if the design proposal ultimately is not adopted by the working group._

aphillips (changes requested):
But otherwise ready...

eemeli:
Requesting fast-tracking.

stasm:
In my mind, a design doc would only be merged into `main` when it's accepted. Furthermore, I think that splitting a design doc into multiple PRs may again make it more difficult to track the discussion around it. Instead, I think we should encourage long-lived PRs and multiple rounds of reviews.

ryzokuken (approved):

eemeli:
@stasm Could you review #458, as that's effectively the prototypical reason for this change?

It's currently at 35 comments in 13 threads of review comments, most of which are hidden as resolved. If we don't make this change, we won't be able to merge the PR until all expression attribute design questions are answered, including e.g. whether it's the proper vessel for communicating "open" vs. "closed" concepts which have not yet been mentioned at all.

aphillips (approved):

PR #462, opened 2023-08-29
Specify pattern resolution behavior for patterns with syntax errors

catamorphism:
While testing my ICU4C implementation on the [JSON tests](https://github.com/messageformat/messageformat/blob/e0087bff312d759b67a9129eac135d318a1f0ce7/packages/mf2-messageformat/src/__fixtures/test-messages.json), I noticed a test case that reflects unspecified behavior:

```
    "src": "no braces",
    "exp": "{no braces}",
```

(See [line 235 of test-messages.json](https://github.com/messageformat/messageformat/blob/e0087bff312d759b67a9129eac135d318a1f0ce7/packages/mf2-messageformat/src/__fixtures/test-messages.json#L235)).

This pull request adds spec text stating that patterns that do not begin with a '{' format to their contents enclosed in '{'/'}'.

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -237,6 +237,21 @@ _Pattern selection_ is not supported for _fallback values_.
 
 ## Pattern Selection
 
+### Messages without Selectors
+
+When a _message_ does not contain a _match_ construct,
+and does not begin with a U+007B LEFT CURLY BRACKET `{`,
+the result of pattern selection MUST be the concatenation of
+a U+007B LEFT CURLY BRACKET `{`,
+the contents of the message as a string,
+and a U+007D RIGHT CURLY BRACKET `}`.
+
+> For example, the following message:
+> `no braces`
+> would format to a string as `{no braces}`.
+
+### Messages with Selectors
+
 When a _message_ contains a _match_ construct with one or more _expressions_,
 the implementation needs to determine which _variant_ will be used
 to provide the _pattern_ for the formatting operation.
@@ -278,7 +293,7 @@ the result of pattern selection MUST be a pattern resolving to a single _fallbac
 using the message's fallback string defined in the _formatting context_
 or if this is not available or empty, the U+FFFD REPLACEMENT CHARACTER `�`.
 
-### Resolve Selectors
+#### Resolve Selectors
 
 First, resolve the values of each _selector_ _expression_:
 
@@ -295,7 +310,7 @@ First, resolve the values of each _selector_ _expression_:
 The shape of the resolved values is determined by each implementation,
 along with the manner of determining their support for selection.
 
-### Resolve Preferences
+#### Resolve Preferences
 
 Next, using `res`, resolve the preferential order for all message keys:
 
@@ -320,7 +335,7 @@ The returned list MAY be empty.
 The most-preferred key is first,
 with each successive key appearing in order by decreasing preference.
 
-### Filter Variants
+#### Filter Variants
 
 Then, using the preferential key orders `pref`,
 filter the list of _variants_ to the ones that match with some preference:
@@ -340,7 +355,7 @@ filter the list of _variants_ to the ones that match with some preference:
          1. Continue the outer loop on message _variants_.
    1. Append `var` as the last element of the list `vars`.
 
-### Sort Variants
+#### Sort Variants
 
 Finally, sort the list of variants `vars` and select the _pattern_:
 


eemeli (changes requested):
I think that's a bug in the test, and the spec should not need to be changed to accommodate it. This is a syntax error, and its expected behaviour is already covered:

> If the message being formatted has any Syntax or Data Model errors, the result of pattern selection MUST be a pattern resolving to a single fallback value using the message's fallback string defined in the formatting context or if this is not available or empty, the U+FFFD REPLACEMENT CHARACTER `�`.

spec/formatting.md
```diff
@@ -237,6 +237,21 @@ _Pattern selection_ is not supported for _fallback values_.
...
+When a _message_ does not contain a _match_ construct,
+and does not begin with a U+007B LEFT CURLY BRACKET `{`,
+the result of pattern selection MUST be the concatenation of
+a U+007B LEFT CURLY BRACKET `{`,
+the contents of the message as a string,
+and a U+007D RIGHT CURLY BRACKET `}`.
+
+> For example, the following message:
+> `no braces`
+> would format to a string as `{no braces}`.
```
aphillips, line 251: I'm not sure this is "pattern selection". This is a parse error, since the message is not, well, a message. What's more, the message "no braces {some braces}" is not a message.

I think the thing that is missing in `formatting.md` is the first step, which is: parsing. This should result in an syntax error. We could then go on to define what happens to the string that wants to be a message. 

To be honest, I would not add the curly brackets. If you give me a string and it produces a syntax error, I should probably give you back the string as you gave it to me. If you thought it should be a valid message, you can debug it. If you are a runtime process, you can just display it (as we didn't damage it).

---

Note well: I argued long and hard that a plain old string with no placeholders should be a valid message. This is the old argument code-mode vs. string-mode. I am not reopening this, but we're going to tangle with the fact that you have to grope strings before feeding them to MF2 as a result.

aphillips (changes requested):
I think this is the right idea, but in the wrong place. See my longer comment below.

aphillips:
@eemeli noted: 
> I think that's a bug in the test, and the spec should not need to be changed to accommodate it. This is a syntax error, and its expected behaviour is already covered:
> 
> > If the message being formatted has any Syntax or Data Model errors, the result of pattern selection MUST be a pattern resolving to a single fallback value using the message's fallback string defined in the formatting context or if this is not available or empty, the U+FFFD REPLACEMENT CHARACTER `�`.

I agree with your comment, but the structure of `formatting.md` is:

- resolve expressions
- select pattern
- format results

So the quoted text is buried inside "select pattern", whereas, as I noted, probably we want to parse the message before resolving declarations or trying to select the pattern. 

Does that make sense?

eemeli:
> I agree with your comment, but the structure of `formatting.md` is:
> 
> * resolve expressions
> * select pattern
> * format results
> 
> So the quoted text is buried inside "select pattern", whereas, as I noted, probably we want to parse the message before resolving declarations or trying to select the pattern.

The text I quoted is from the fourth top-level part of `formatting.md`, "Error Handling".

aphillips:
@eemeli 
> The text I quoted is from the fourth top-level part of formatting.md, "Error Handling".

Okay. But do you agree that we're missing the parsing?

eemeli:
@catamorphism Are you still looking to advance this PR, or should it be considered abandoned and closed?

catamorphism:
> I think that's a bug in the test, and the spec should not need to be changed to accommodate it. This is a syntax error, and its expected behaviour is already covered:
> 
> > If the message being formatted has any Syntax or Data Model errors, the result of pattern selection MUST be a pattern resolving to a single fallback value using the message's fallback string defined in the formatting context or if this is not available or empty, the U+FFFD REPLACEMENT CHARACTER `�`.

What should the expected result be in the test? Currently the expected result is [`{no braces}`]](https://github.com/messageformat/messageformat/blob/e0087bff312d759b67a9129eac135d318a1f0ce7/packages/mf2-messageformat/src/__fixtures/test-messages.json#L235).

catamorphism:
> @catamorphism Are you still looking to advance this PR, or should it be considered abandoned and closed?

I just returned from leave today. Yes, I think this question still requires clarification, if only by changing the test (see previous comments).

eemeli:
> What should the expected result be in the test? Currently the expected result is [`{no braces}`](https://github.com/messageformat/messageformat/blob/e0087bff312d759b67a9129eac135d318a1f0ce7/packages/mf2-messageformat/src/__fixtures/test-messages.json#L235).

While you were out, we kinda went and started to do a number on the syntax, so for this test the expected result is now just `no braces`, as the braces are no longer required at least for simple messages like this.

In general, for syntax and data model errors the expected result is `{�}` -- which is also the format's logo now.

catamorphism:
I think this PR is obsolete now due to syntax changes. If I notice related issues I'll start a new PR.

PR #463, opened 2023-08-29
(Design) Formatted Parts

eemeli:
This is pre-empting #458 and #461, but I figure they'll likely land before this.

The proposed design is close to that used by the JS polyfill, but there are some differences.


exploration/0003-formatted-parts.md
```diff
@@ -0,0 +1,187 @@
...
+interface MessageTextPart {
+  type: "text";
+  value: string;
+}
+```
+
+For MessageExpressionPart, the `source` corresponds to the expression's fallback value.
+The parts' `dir` and `locale` may be defined if overridden by an expression attribute or set otherwise.
+Each part should have at most one of `value` or `parts` defined;
+some may have none.
```
aphillips, line 76: Does this make sense? Or would it be better to say:

```suggestion
Each part MUST have either a `value` or `parts` defined. 
A part MAY have a `value` that is the empty string.
A part MAY have a `parts` that is an empty list.
```

exploration/0003-formatted-parts.md
```diff
@@ -0,0 +1,187 @@
...
+
+The bidi isolation strategies included in the spec may require
+the insertion of MessageBiDiIsolationParts in the formatted-parts output.
+
+```ts
+interface MessageBiDiIsolationPart {
+  type: "bidiIsolation";
+  value: "\u2066" | "\u2067" | "\u2068" | "\u2069"; // LRI | RLI | FSI | PDI
+}
+```
```
aphillips, line 97: Maybe? But since the parts have a direction, the bidi isolation parts are not necessarily required. Let's consider pattern:

```
{Today is {$date :datetime}}
```

In locale `ar-AE`, the output of the `:datetime` function might be "30‏/08‏/2023" with a `dir` of `rtl`. Note that the string looks bad here because it is an LTR context and not bidi isolated. There are U+200F RLMs after `08` and `30` to assist, but the string needs to be wrapped.

Ignoring that the date has interior parts, getting a MessageExpressionPart:

```
value: "30‏/08‏/2023"
dir: "rtl"
```

... could result in a formatted string (in an HTML context) like:

```
Today is <span dir=rtl>30‏/08‏/2023</span>.
```

This draws correctly without any further intervention.

If this became a list of parts it still works:

```
parts: [
   { name: "day", value: "30\u200f" },
   { name: "sep", value: "/"},
   { name: "month", value: "08\u200f"},
   { name: "sep", value: "/"},
   { name: "year", value: "2023"}
]
dir: "rtl"
```

The caller could use this to produce:

```html
Today is <span dir=rtl>
<span class="day">30&#x200f;</span>/08&#x200f;2023</span>.
```

[Test](https://w3c.github.io/i18n-discuss/explainers/bidi-html-demo.html?text=Today%20is%20%3Cspan%20class=magenta%20dir=rtl%3E30\u200f/08\u200f/2023%3C/span%3E.)

You wouldn't want Unicode isolates in this case.

exploration/0003-formatted-parts.md
```diff
@@ -0,0 +1,187 @@
...
+
+Some of the MessageExpressionPart instances may be further defined
+without reference to the function registry.
+
+Unannotated expressions with a _literal_ operand
+are represented by MessageStringPart.
+As with MessageTextPart,
+the `value` of MessageStringPart is always a string.
```
aphillips, line 105: Is there a need for the distinction? It is simply the source?

exploration/0003-formatted-parts.md
```diff
@@ -0,0 +1,187 @@
...
+  type: "string";
+  source: string;
+  value: string;
+  dir?: "ltr" | "rtl" | "auto";
+  locale?: string;
+}
+```
+
+Unannotated expressions with a _variable_ operand
+that is not formatted according to its type
```
aphillips, line 118: Do you mean:

```suggestion
whose type is not recognized by the implementation
or for which no default formatter is available
```

exploration/0003-formatted-parts.md
```diff
@@ -0,0 +1,187 @@
...
+interface MessageUnknownPart {
+  type: "unknown";
+  source: string;
+  value: unknown;
+}
+```
+
+When the resolution or formatting of a placeholder fails,
+it is represented in the output by MessageFallbackPart.
+No `value` is provided; when formatting to a string,
```
aphillips, line 131: Should we provide the fallback value? I think we have some text in the spec that allows implementations or functions to supply their own fallback.

---

Question: should a goal be that the string output of a message be equivalent to concatenating the string representation of its parts? Or at least that a test be that one can assemble the string output from the parts?

exploration/0003-formatted-parts.md
```diff
@@ -0,0 +1,187 @@
...
+
+```ts
+interface MessageFallbackPart {
+  type: "fallback";
+  source: string;
+}
+```
+
+Formatting functions defined in the registry
+will need accompanying formatted-parts representations.
```
aphillips, line 142: This sentence confused me, since, when I read it, I thought it was requiring a "parts" interface or that you were using too strong a word with "will". I think what you really mean is:

```suggestion
Each function defined in the registry MUST define its "formatted-parts" representation.
A function can define either a unitary string `value` or a `parts` representation.
Where possible, a function SHOULD provide a `parts` representation if its output 
might reasonably consist of multiple fields.
```

(I added the sentences after the first as a further suggestion)

exploration/0003-formatted-parts.md
```diff
@@ -0,0 +1,187 @@
...
+Where available, such a formatted value should itself be represented by `parts`
+rather than a unitary string `value`.
+These sub-parts should not need fields beyond their `type` and `value`,
+and in most cases it's presumed that the sub-part `value` would be a string.
+
+```ts
+interface MessageDateTimePart {
```
aphillips, line 149: Can we define the "parts" so that they are generic rather than each type having its own special part type?

exploration/0003-formatted-parts.md
```diff
@@ -0,0 +1,187 @@
...
+  type: "datetime";
+  source: string;
+  parts: Iterable<{ type: string; value: unknown }>;
```
aphillips, line 152: Notice in my example earlier in this review that I exposed the field name as a field in the parts. This becomes important when trying to decorate an Iterable whose contents shift around due to the locale/localized formatting. Dates have this feature (YMD, DMY, MDY). So do currency values (which may or may not have a decimal part, may have the symbol first or last, and may or may not have a space around the symbol). That's how the screen shots of currency values (from amazon.com and amazon.fr) get decorated.

aphillips (changes requested):
Good to see this. Many initial comments...

exploration/0003-formatted-parts.md
eemeli: The suggestion would make the proposed MessageFallbackPart invalid, as it does not include either `value` or `parts`. It's conceivable for other parts to also exist which do not include either, such as open/close expressions without an annotation.

exploration/0003-formatted-parts.md
eemeli: I agree that in some cases bidi isolation of formatted-parts output may be achieved e.g. via `<span dir=...>` tags, but I don't think that's universally true for every use case for formatted parts.

This is one reason why e.g. in the JS Intl.MF proposal my proposal for bidi handling includes the `"none"` strategy as an alternative to the default (and required) `"compatibility"` strategy.

Effectively, I think we should include the MessageBiDiIsolationPart by default, but allow for an implementation to provide a way to get rid of them, or just have a consumer of the output ignore all the `type: "bidiIsolation"` parts.

exploration/0003-formatted-parts.md
eemeli: In addition to `source` the `dir` and `locale` are not present on MessageTextPart.

exploration/0003-formatted-parts.md
eemeli: > Should we provide the fallback value? I think we have some text in the spec that allows implementations or functions to supply their own fallback.

Fallback customization is only available for syntax and data model errors, which default to `�`. That would be used as the `source` value here.

> Question: should a goal be that the string output of a message be equivalent to concatenating the string representation of its parts? Or at least that a test be that one can assemble the string output from the parts?

I think the latter. With the current proposal, it's doable like this:

```js
function stringifyParts(parts) {
  let res = ''
  for (let part of parts) {
    if (part.type === 'fallback') res += `{${part.source}}`
    else if ('value' in part) res += String(part.value)
    else if ('parts' in part) {
      for (let sub of part.parts) res += String(sub.value)
    }
  }
  return res
}
```

exploration/0003-formatted-parts.md
eemeli: Yes, that's the MessageExpressionPart definition above, which these interfaces also match. The definitions here are giving more specificity about what e.g. `:datetime` and `:number` end up producing, i.e. that they have explicit `type` identifiers and define `parts` rather than `value`.

exploration/0003-formatted-parts.md
eemeli: Yes, that's the `"type"` field here. I picked that rather than `"name"` because it's used by the JS [Intl formatters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/formatToParts#description).

aphillips: That was super non-obvious to me (hence this conversation!), particularly since the `type` fields in the examples seemed to be focused on the "type" of formatter (datetime, number) rather than on the parts field within them. Admittedly, the MF2-level parts will be at the placeholder level. Interior parts are the problem of the formatter. But this was not at all clear and probably could use an example.

exploration/0003-formatted-parts.md
aphillips: Would an empty `value` or empty `parts` satisfy that? Or a fallback could have a string expression? Empty strings don't result in the erroneous emission of the string `null` 😉 

I understand that it would "break" the current definition: we should decide what the shapes should be and make consistent.

exploration/0003-formatted-parts.md
aphillips: Why don't text parts have a direction and language? Are they inheriting from the message?

Aside: we don't necessarily want to span every section of text, that is, this is better:

```html
<span lang=foo dir=ltr>blah blah blah 
   <span lang=bar dir=auto>placeholder here</span> blah blah</span>
```

than this:

```html
<span lang=foo dir=ltr>blah blah blah </span>
   <span lang=bar dir=auto>placeholder here</span><span lang=foo dir=ltr> blah blah</span>
```

... which is why text parts would inherit language and base paragraph direction from the message.

eemeli: Text parts do inherit the language and direction from the message, so they'll be whatever the formatter was called with. But they're not necessary to be repeated on every part, because the caller already knows what they asked for, and they may well be formatting the message in a context with e.g. `<body lang=foo dir=ltr>` setting them at a much higher level.

exploration/0003-formatted-parts.md
eemeli: Let's make sure that it's explained in more detail in the spec PR, should this design doc be accepted.

exploration/0003-formatted-parts.md
eemeli: For fallback when formatting to a string the `{...}` make sense as a visual indicator, but for a formatted-parts consumer some different representation could be better. So including an explicit `value` or `parts` would be misleading.

For open/close, it doesn't make sense to define their explicit parts shapes in this spec, but for JS I have them as:

```ts
interface MessageMarkupPart {
  type: 'open' | 'close';
  source: string;
  name: string;
  value?: unknown;
  options: { [key: string]: unknown };
}
```

There, the `value` would be `'b'` for `{b +html}`, but it would not be set for `{+html.b}`. Setting it to an empty string would be misleading, as `{+foo}` and `{|| +foo}` could easily mean different things.

exploration/0003-formatted-parts.md
aphillips: I agree that span is not the only way that bidi will be implemented. `none` is easy: one can just ignore the bidi/direction. Obviously sometimes folks will want bidi controls. Or they might want to make an attributed string, I suppose, in some UI frameworks. 

I'm reluctant to create a "bidi isolation part" because it separates the direction information from the value it is associated with. It's harder if one is doing decoration, such as building some control using various tags, not all of which are span, some of which have classes or other attributes. The only place where the isolation part is easier is stringifying the message (since one just consumes it. 

```javascript
function stringifyParts(parts) {
  let res = ''
  for (let part of parts) {
    res += bidiStrategyImpl.start(part.dir);  // none does nothing, compatibility does a control, etc.
    if (part.type === 'fallback') res += `{${part.source}}`
    else if ('value' in part) res += String(part.value)
    else if ('parts' in part) {
      for (let sub of part.parts) res += String(sub.value)
    }
    res += bidiStrategyImpl.end(part.dir);
  }
  return res
}
```

eemeli: Doing as you propose would require implementations to provide explicit `bidiStrategyImpl.start()` and `bidiStrategyImpl.end()` interfaces, or the behaviour of format-to-parts-to-string could not be guaranteed to match the behaviour of format-to-string.

That seems like a much bigger ask than defining what the bidi isolation parts would look like if they are to be included in the parts output by the requested bidi isolation strategy.

aphillips: Format-to-string's output would depend on what the bidi isolation strategy was, no? 

The start/end thing was just an example. Note that with isolation parts you have to "read ahead" to find out what is being isolated if what one is doing with the parts is decorating a control or generating HTML.

eemeli: > Format-to-string's output would depend on what the bidi isolation strategy was, no?

Yes.

> The start/end thing was just an example. Note that with isolation parts you have to "read ahead" to find out what is being isolated if what one is doing with the parts is decorating a control or generating HTML.

Sure, but either the bidi isolation parts need to be included, or some way of obtaining their equivalent is required.

eemeli:
Just to note, progress on this design doc appears to be blocked by lack of review for the past two months.

exploration/0003-formatted-parts.md
```diff
@@ -0,0 +1,205 @@
...
+		<dd>#000</dd>
+	</dl>
+</details>
+
+## Objective
+
+Messages often include placeholders that,
+when formatted, contain internal structure ("parts")
+that the caller might want access to
+for the purposes of styling, presentation, or manipulation.
```
catamorphism, line 22:
```suggestion
when formatted, contain internal structure ("parts").
Preserving this structure in a formatted message
may be helpful to the caller, who can then
manipulate the parts.
For example, a caller may want to style or present
messages with the same content differently
if those messages have different internal structure.
```

exploration/0003-formatted-parts.md
```diff
@@ -0,0 +1,205 @@
...
+
+This proposal defines a formatted-parts target for MessageFormat 2.
+
+## Background
+
+Past examples have shown us that if we don't provide a formatter to parts,
+the string output will be re-parsed and re-processed by users.
```
catamorphism, line 29: It might be useful to explain why it's undesirable for users to re-parse a string (or cite some other document that explains it).

exploration/0003-formatted-parts.md
```diff
@@ -0,0 +1,205 @@
...
+or otherwise set by the implementation.
+Each part should have at most one of `value` or `parts` defined;
+some may have none.
+
+```ts
+interface MessageExpressionPart {
+  type: string;
+  source: string;
+  parts?: Iterable<{ type: string; value: unknown }>;
+  value?: unknown;
```
catamorphism, line 97: Can this be something like:

```
contents?: <Iterable<{ type: string; value: unknown }> | unknown>;
```
to capture that `parts` and `value` can't both be present?

exploration/0003-formatted-parts.md
eemeli:
```suggestion
the string output will be re-parsed and re-processed by users.
Recent examples of web browsers needing to account for such user behaviour are available from
[June 2022](https://github.com/WebKit/WebKit/commit/1dc01f753d89a85ee19df8e8bd75f4aece80c594) and
[November 2022](https://bugs.chromium.org/p/v8/issues/detail?id=13494).
```

exploration/0003-formatted-parts.md
eemeli: I don't think so. Using a single field for both `parts` and `value` would mean that it would be possible for a unitary value to masquerade as a sequence of parts, and be treated as such.

aphillips:
Pinging @mihnita: You have an action to review/comment on this doc.

aphillips (approved):

mihnita:
My main concern (also expressed in Seville) is that the proposed model is conceptually a tree
(2 levels deep, but maybe more, with the deeper levels unspecified).

> top level, the formatted-parts result is an iterable sequence of parts
One of them is `MessageExpressionPart`, which contains `parts?: Iterable<{ type: string; value: unknown }>`

So from what I understand the `MessageExpressionPart` contains what a formatter function returns.
But the proposal stops there, and does not go to explain what the parts returned by a formatter would be.

Let's take an example: `I'll be out of office {$vacation, :dateRange}`
The result might be something like `"I'll be out of office November 27-December 7, 2023"`

If I read this proposal correctly, the "November 27-December 7, 2023" would be described by an `MessageExpressionPart`, but there is no explanation on what that looks like.

So the most useful part for people's use cases ("how can format the months differently") is left out.

---

I looked at existing implementations of similar concepts in iOS / macOS, Android, ICU4C / ICU4J, Eclipse SWT.

And all of them use "flat" structures: a text, with information about ranges.

---

It is not a bad design in itself, standalone.

But my main concerns are:
 - it is stops the description to where it starts being interesting
 - the "impedance" between this proposal and existing implementations (nesting vs flat text with attributes)

Do we want to go in that direction?

I will document the existing implementations in a separate comment, for readability.

mihnita:
## macOS & iOS

[`AttributedString`](https://developer.apple.com/documentation/foundation/attributedstring)

```
var attributedString = AttributedString("This is a string with empty attributes.")
var container = AttributeContainer()
container[AttributeScopes.AppKitAttributes.ForegroundColorAttribute.self] = .red
attributedString.mergeAttributes(container, mergePolicy: .keepNew)
```
The attributes are in an [`AttributeContainer`](https://developer.apple.com/documentation/foundation/attributecontainer) (key-value)

**APIs to access:**
* Access a range: https://developer.apple.com/documentation/foundation/attributedstring/3765864-subscript
* Access indices: https://developer.apple.com/documentation/foundation/attributedstring/accessing_indicies_within_an_attributed_string
* Iterate: `AttributedString.Runs` : An iterable view into segments of the attributed string, each of which indicates where a run of identical attributes begins or ends. https://developer.apple.com/documentation/foundation/attributedstring/runs

mihnita:
## Android

[`android.text.Spanned`](https://developer.android.com/reference/android/text/Spanned)

API: to access the span info: [`T[] getSpans(int start, int end, Class<T> type)`](https://developer.android.com/reference/android/text/Spanned#getSpans(int,%20int,%20java.lang.Class%3CT%3E))

Classes:
```
Spanned(i)
	Spannable (i)
		SpannableStringBuilder (c) => mutable content & markup
		SpannableString (c) => immutable content, muttable markup
```

Example:
```
Spannable spannable = new SpannableString("0123456789_ABCDEFGHIJKLMNOPQRSTUVWXYZ_abcdefghijklmnopqrstuvwxyz");
spannable .setSpan(new ForegroundColorSpan(Color.BLUE), 10, 20, Spannable.SPAN_EXCLUSIVE_EXCLUSIVE);
```

Spans supported: https://developer.android.com/reference/android/text/style/package-summary
(and can be extended with custom ones)

mihnita:
## ICU4J

[`FormattedValue`](https://unicode-org.github.io/icu-docs/apidoc/dev/icu4j/com/ibm/icu/text/FormattedValue.html)

Classes implementing it:
`DateIntervalFormat.FormattedDateInterval`, `FormattedMessage`, `FormattedNumber`, `FormattedNumberRange`, `ListFormatter.FormattedList`, `PlainStringFormattedValue`, `RelativeDateTimeFormatter.FormattedRelativeDateTime`

Two ways to iterate:

* `nextPosition(ConstrainedFieldPosition cfpos)` // `cfpos` contains the info.
* `AttributedCharacterIterator toCharacterIterator()` => inherited from JDK (`java.text.AttributedCharacterIterator`)

## ICU4C

[`icu::FormattedValue`](https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/classicu_1_1FormattedValue.html)

Implemented in `icu::FormattedDateInterval`, `icu::number::FormattedNumber`, `icu::FormattedList`, `icu::number::FormattedNumberRange`, and `icu::FormattedRelativeDateTime`.

Iteration: `nextPosition (ConstrainedFieldPosition &cfpos, UErrorCode &status) const =0`

There is no equivalent of `toCharacterIterator()`

mihnita:
## SWT
[`StyledText`](https://archive.eclipse.org/eclipse/downloads/documentation/2.0/html/plugins/org.eclipse.platform.doc.isv/reference/api/org/eclipse/swt/custom/StyledText.html)

APIs: `setStyleRange(StyleRange range)`, `StyleRange[] getStyleRanges()`

aphillips:
@mihnita 
> If I read this proposal correctly, the "November 27-December 7, 2023" would be described by an MessageExpressionPart, but there is no explanation on what that looks like.

It contains an `Iterable` of `parts`. The idea is that the formatting function (`DateTimeRangeFormatter` in this case) knows what parts there are and supplies that information. This way you can decorate the MessageExpressionPart _as a thing_ and still decorate subparts (such as the month names).

This could be used to create an attributed string just as much as it could be used to create a sequence of DOM nodes. What is important is that the various boundaries are accessible to the caller and in the correct order for the localized message. It's important to know where the placeables were, but then know _inside_ the placeable where sub-field boundaries are.

Conversely, an implementation of MF2 could return an attributed string describing the parts defined here (using attributes instead of objects)?

exploration/0003-formatted-parts.md
mihnita: So the parts are technically not formatted yet?
I can't get from the proposal what a `sub.value` is.
Can it be a date?

exploration/0003-formatted-parts.md
```diff
@@ -0,0 +1,212 @@
...
+  dir?: "ltr" | "rtl" | "auto";
```
mihnita, line 105: It is unclear to me why have `dir` and `locale` at this level.

We don't need a locale to format anything, because the parts should be already formatted.
The whole proposal is called "Formatted Parts"

The locale might be needed to render things.

Or to process the formatting result (fix grammatical agreements as a post-step, fix "a apple" to "an apple" (en) or "La abeille" => "L'abbeile" (fr), or to sentence case the result of "{item} is foo"

But then that is something that is needed for the whole collection of parts, not on `MessageExpressionPart` only.

eemeli: These are needed to allow embedding content in a message that uses a different script or locale than the surrounding message.

mihnita: Kind of clear what they can be used for.

But this info is on the `MessageExpressionPart`, which comes from an expression.
And an expression can't create this info out of nothing, it is probably something we passed as a parameter.
So if I already know the info (because I passed it to the expression), having it on the `MessageExpressionPart` is useless duplication.

mihnita:
> It contains an `Iterable` of `parts`. The idea is that the formatting function (`DateTimeRangeFormatter` in this case) knows what parts there are and supplies that information.

Yes, but it looks like we are mixing the inputs with the outputs.
The input would be the pattern, with parameters, and pass that to MF2, which formats to parts.
MF2 already invoked `DateTimeRangeFormatter`, which already returned some kind of "sub-parts".

Now I need to be able to look at the output and access these parts and sub-parts without `DateTimeRangeFormatter`, which was input.

> Conversely, an implementation of MF2 could return an attributed string describing the parts defined here (using attributes instead of objects)?

Yes, but that implementation would not be standard.
Because there is no equivalent of `MessageExpressionPart` with a list of items in attributed strings.
All attributes are at the same level, there is no part with sub-parts.

> MF2 doesn't know or need to know about what kinds of parts (if any) a subordinate formatter has or returns.

Does not need to know, but if the formatters return parts that are "compatible" with what MF2 returns that is good.
First, because there is consistency. I don't iterate the MF2 parts in the result one way, and then the parts in a range format in another way.

If they "part id" is a string then I can do something css selection.

Example:
```
msg = "I'll be out of office {$vacation, :dateRange}!"

parts = "I'll be out of office November 27-December 7, 2023!"
attributedString "vacation" = "November 27-December 7, 2023"
attributedString "start" = range of "November 27"
attributedString "end" = range of "December 7"
attributedString "common" = range of  ", 2023"
attributedString "month" = range of  "November"
attributedString "month" = range of  "December"
attributedString "day" = range of  "27"
attributedString "day" = range of  "7"
```
The "attributes" overlap, same as bold / italic / other attributes overlap.
range of actually means the start offset + end offset.
But `range of  "November"` is more readable than `[22, 30]`
All the APIs described above use that style.

But then I can iterate and I get something like this:
```
"I'll be out of office" => []
"November" => ["vacation", "start", "month"]
"27" => ["vacation", "start", "day"]
"December" => ["vacation", "end", "month"]
"-" => ["vacation", "separator"]
"7" => ["vacation", "end", "day"]
", " => ["vacation", "common"]
"2023" => ["vacation", "common", "year"]
"!" => []
```

And I can do the "css selection"-like operations:
- `select( parts, "vacation;month")` => range of "November"
- `select( parts, "month")` => collection of ranges (covering "November" and "December")

---

If we leave the sub-parts undefined it is more inconsistent in how I access the parts / subparts.
And the whole thing is a lot less useful.

mihnita:
I don't see a way to represent the result of "markdown" in the parts.
Which would map naturally to a range of text with attribute(s).

mihnita:
I don't think that `MessageBiDiIsolationPart` is needed, or that it actually works.
And it is unclear who would be responsible to create it.

Since it is a part like all others, I can in theory have 
`[ MessageTextPart MessageBiDiIsolationPart MessageTextPart MessageBiDiIsolationPart MessageTextPart]`

Since `MessageBiDiIsolationPart` is not created by an expression, it means it was in the pattern (translated or not). Which means it was in plain text somehow.
So it is something that was actually plain text.
And if it was plain text, and it is one of `LRI | RLI | FSI | PDI`, then why do we need a special part for it?
Leave them as plain text, and I can remove them as I want, or convert them to `<span dir=...>` or to `<bdi>`, or whatever.

eemeli:
The primary reason why the type of the formatted parts of an expression is left as `unknown` rather than e.g. `string` is to support formatting to non-string targets. Consider for example message references, which we've left out of the spec with the understanding and expectation that they may be implemented on top of it.

If we agree that it's reasonable to have a non-string formatted value for an expression (for example, an inline image or date selector) and we don't want to reopen the discussion on message references, then we need the spec to support having a message pattern that includes an expression that acts as a message reference which resolves to a formatted message with a non-string part in it.

In a message resource, that would look something like this:

```
foo = I have an image {$img}, isn't it pretty?

bar = They said, "{foo :msgref img=$img}"
```

With the proposed approach, the formatted parts representation of `bar` could look something like:

```js
[
  { type: 'text', value: 'They said, "' },
  { type: 'message', source: '|foo|', parts: [
    { type: 'text', value: 'I have an image ' },
    { type: 'unknown', source: '$img', value: <img/> }, // value defined by an input variable
    { type: 'text', value: ", isn't it pretty?" }
  ]},
  { type: 'text', value: '"' }
]
```

Effectively, if we support any non-string value in the formatted result, we need to either do something like what's proposed here, or we need implement at least message references directly in MF2 and determine some custom solution for them.

> It also looks to me like the result can't be used as is, or serialized, as it is not yet "formatted" The `parts?: Iterable<{ type: string; value: unknown }>` still contain unformatted(?) values (`value: unknown`).

The intent is that these values are indeed formatted, but that the spec does not require the formatting to be to a `string`. A library of formatters could of course provide a stricter guarantee about its output types.

> I looked at existing implementations of similar concepts in iOS / macOS, Android, ICU4C / ICU4J, Eclipse SWT.
> 
> And all of them use "flat" structures: a text, with information about ranges.

Thank you for these links. Reviewing these, it looks like they could be divided in two categories:
1. Very restrictive (ICU4C, ICU4J, SWT). These only allow very specific information to be attached to strings, and would not allow e.g. associating a URL with a range.
2. Very liberal (iOS/MacOS, Android). These allow any values to be associated with text ranges (`AttributeContainer` values, the `Object what` of `Spannable` spans).

I'm not sure that the former is suitable for our purposes, but let us consider the latter. With a range-based approach, the formatted result of my example message from above could be:

```js
{
  text: 'They said, "I have an image [object HTMLImageElement], isn\'t it pretty?"',
  ranges: [
    { start: 12, end: 71, type: 'message', source: '|foo|' },
    { start: 28, end: 53, type: 'unknown', source: '$img', value: <img/> }
  ]
}
```

This encodes the same information, and may be directly derived from the proposed result. But it requires more work, and unnecessarily disadvantages non-string formatting targets. There should not be a need here to create a single concatenated string, calling the stringifiers of all its elements (in other words, this would add many `[object Object]` to the world), and to then require the consumer to reconstruct the useful representation by splitting the string and replacing specific elements of it with their values.

Put another way, it's easier to transform parts to ranges rather than vice versa. In terms of code, [this](https://gist.github.com/eemeli/e081c442b524878313725943a45dff58) is what those operations look like. With `partsToRanges()`, I'm relatively confident that the implementation will behave as intended. For `rangesToParts()`, I'd need to write a pretty extensive test suite to be sure, and in real-world code account for all sorts of additional error conditions.

exploration/0003-formatted-parts.md
eemeli: Consider this message, intended for consumption by a text-to-speech system:
```
In French, the number {98 :number} is commonly expressed as {98 :number @locale=fr},
but in Belgium it's {98 :number @locale=fr-BE}.
```

How would you propose that the locale information is transmitted, if not as a field on the formatted parts?

```js
[
  { type: 'text', value: 'In French, the number ' },
  { type: 'number', source: '|98|', parts: [{ type: 'integer', value: '98' }] },
  { type: 'text', value: ' is commonly expressed as ' },
  { type: 'number', source: '|98|', locale: 'fr', parts: [{ type: 'integer', value: '98' }] },
  { type: 'text', value: ", but in Belgium it's " },
  { type: 'number', source: '|98|', locale: 'fr-BE', parts: [{ type: 'integer', value: '98' }] },
  { type: 'text', value: '.' }
]
```

As context, the `fr` number would be "quatre-vingt-dix-huit", while in `fr-BE` it's "nonante-huit".

eemeli:
> I don't see a way to represent the result of "markdown" in the parts. Which would map naturally to a range of text with attribute(s).

Could you share an example message for this? I have a hard time imagining just what that would look like, and my suspicion is that such handling would be taken care of by a consumer of MF2 formatting.

> Since `MessageBiDiIsolationPart` is not created by an expression, it means it was in the pattern (translated or not). Which means it was in plain text somehow. So it is something that was actually plain text. [...]

This doesn't quite follow. The bidi isolation parts are added between parts that have different directionalities, very much like we're adding LRI/RLI/FSI/PDI characters in the formatted string. They are not a part of the plain text in the source message.

Our current spec pretty much requires something like this:
https://github.com/unicode-org/message-format-wg/blob/705924846e1793d574e6c05202b525ae5e124a7f/spec/formatting.md?plain=1#L678-L680

exploration/0003-formatted-parts.md
aphillips: Language and direction are needed for placeholders because they represent values being inserted into the overall string. The language (locale) is used to ensure proper rendering and processing (such as line breaking, text transforms, or spell-checking). The direction is used to enable bidi isolation and get the direction of the substring correct.

The language and direction of a formatted part might not match that of the message due to resource fallback when looking up the message. Or because values passed in have different language or direction. (And we want bidi isolation even if the directions match!!!!)

Providing the fields in the formatted part structure allows the user to easily access the values, e.g. it makes it easy to do something like this, resulting in proper isolation of the formatted parts (not shown is decorating the parts separately):

```javascript
var message = // whatever the host node is for the string
for (let part of formattedMessage.parts) {
    var span = document.createElement('span');
    span.lang = part.lang;
    span.dir = part.dir;
    span.appendChild(document.createTextNode(part.value));
    message.appendChild(span);
}
```

That is done on portions of the message. The whole message also has a language (locale) and base paragraph direction.

mihnita:
> > I don't see a way to represent the result of "markdown" in the parts. Which would map naturally to a range of text with attribute(s).
> 
> Could you share an example message for this? I have a hard time imagining just what that would look like, and my suspicion is that such handling would be taken care of by a consumer of MF2 formatting.

I am not sure what's unclear here.
Example:
```
Click {+link src=|...|}here{-link} for more info.
```
Where are the open / close links in the formatted to parts result?

mihnita:
> This doesn't quite follow. The bidi isolation parts are added between parts that have different directionalities, very much like we're adding LRI/RLI/FSI/PDI characters in the formatted string. They are not a part of the plain text in the source message.

If they are not a part of the plain text in the source message, where are they coming from?
There is no equivalent structure in the ebnf. The `pattern` is a sequence of `text`  and `expression`s.
The `text` results in a `MessageTextPart`, and an `expression` results in a `MessageExpressionPart`.
What generates a `MessageBiDiIsolationPart`?

In most cases translators will insert bidi control characters when needed.
And they will be in the `text` part of a `pattern`.
That's what I meant that they are part of the "plain text"

mihnita:
> Following the draft mentioned https://github.com/unicode-org/message-format-wg/pull/463#discussion_r1311879464, the result of formatting that message to parts would look like this:

That is not part of the document, it is a comment.
But let's assume it is.
Then I would expect the type to be `link`, not `open` / `close`.

Everything else is seems to be based on what that part represents:
* `MessageDateTimePart` => `"datetime"`
* `MessageNumberPart` => `"number"`
* message reference => `"message"`

---

> The selected bidirectional isolation strategy. The spec section on [Handling Bidirectional Text](https://github.com/unicode-org/message-format-wg/blob/main/spec/formatting.md#handling-bidirectional-text) currently defines their addition only in terms of formatting to a string, but it'll need to be expanded to handle formatting to parts as well.

It looks to me like the `dir` property in `MessageExpressionPart` does the same thing as `MessageBiDiIsolationPart`.
We don't need both.

The problems usually show up "at the edge" of `MessageExpressionPart`s.
So the `dir` info can be used to wrap the expression region in bidi control characters, or (in html) in `<span dir` or `<bdi>` or `<bdo>`.
That is in fact recommended by W2C (https://www.w3.org/International/articles/inline-bidi-markup/)

The other place where bidi control characters might show up is if a translator really inserted a RLE / RLI / ...
Are these going to be converted to `MessageBiDiIsolationPart`?

TLDR: I think `MessageBiDiIsolationPart` tries to solve the same problem that `dir` in `MessageExpressionPart` solves, and it is not needed.

---

Looks to me like both `MessageBiDiIsolationPart` and `MessageMarkupPart` would be more naturally represented with ranges: from 10 to 20 is a link, and from 12 to 18 is a RTI section.
Similar for sections of text in a different language (from 0 to 40 the text is Arabic)

mihnita:
> Let's discuss this further when we have an agreement on open/close. It's left out of this doc for now because we've not been able to achieve that consensus yet.

I think we do have an agreement: we will have open / close, very likely standalone.
We might not have agreement on the sigils, but that does not affect this parts design.

And I'm unhappy with "let's land this" approach piece-meal, because we ignore the big picture.
If when we get to "agreement" on open / close and I argue that annotated strings are a better representation for that, the push back will be "but that means a big change on landed stuff that you already agreed on"

---

HTML and bidi: I don't think we disadvantage anything.

My main argument is:
- we only expect problems "at the edge" between text and expressions
- the expression parts already have `dir` attribute
- the plain text part might contain bidi characters introduced "as is" by translators

Non-html can use the `dir` info in expression parts to can be used to generate bidi control character for HTML, no problem.
It is about duplicated functionality, nothing to do with the format.

---

> As I've presented here previously, that representation is rather easy to generate from the proposed model. The reverse transformation for users who are not ultimately formatting to a string is much more cumbersome.

Of course.
The annotated strings option is richer, and has more functionality.
It is easier to move from a "poor format" to "richer" one, but not the other way around.

mihnita:
Example of one thing that can be handed with annotated ranges but not with a flat(ish) structure (2 levels only):
`"We are closed {$closedRange :dateinterval}"`
=>
`"We are closed November 27-December 3, 2023"`

With ranges I can make the difference between 27 and 3, because 27 is annotated with `[ closedRange, start, day ]`
and 3 is annotated with `[ closedRange, end, day ]`
With sequential parts (proposed, and the JS style of formatting to parts) all you get is `[ text month day month day year ]`, there is no way to tell what's what.

Because ranges can overlap one can also have different text for tts and to be rendered on screen.

eemeli:
> Example of one thing that can be handed with annotated ranges but not with a flat(ish) structure (2 levels only): `"We are closed {$closedRange :dateinterval}"` => `"We are closed November 27-December 3, 2023"`
> 
> With ranges I can make the difference between 27 and 3, because 27 is annotated with `[ closedRange, start, day ]` and 3 is annotated with `[ closedRange, end, day ]`
> 
> With sequential parts (proposed, and the JS style of formatting to parts) all you get is `[ text month day month day year ]`, there is no way to tell what's what.

Would your concerns about the proposed API be satisfied if it could be shown how it could represent the above information you claim it cannot represent?

exploration/formatted-parts.md
```diff
@@ -0,0 +1,212 @@
...
+or overridden by an expression attribute or formatting function,
+or otherwise set by the implementation.
+Each part should have at most one of `value` or `parts` defined;
+some may have none.
+
+```ts
+interface MessageExpressionPart {
+  type: string;
+  source: string;
+  parts?: Iterable<{ type: string; value: unknown }>;
```
eemeli, line 103:
```suggestion
  parts?: Iterable<{ type: string; value: unknown; source?: string }>;
```

exploration/formatted-parts.md
```diff
@@ -0,0 +1,212 @@
...
+
+Formatting functions defined in the registry
+Each function defined in the registry MUST define its "formatted-parts" representation.
+A function can define either a unitary string `value` or a `parts` representation.
+Where possible, a function SHOULD provide a `parts` representation
+if its output might reasonably consist of multiple fields.
+Where available, such a formatted value should itself be represented by `parts`
+rather than a unitary string `value`.
+These sub-parts should not need fields beyond their `type` and `value`,
+and in most cases it's presumed that the sub-part `value` would be a string.
```
eemeli, line 171:
```suggestion
In most cases, these sub-parts should not need fields beyond their `type` and a string `value`,
Where necessary, other `value` types may be used and other fields such as a `source` included.
```

exploration/formatted-parts.md
```diff
@@ -0,0 +1,212 @@
...
+the part's representation would be `'{' + source + '}'`.
+
+```ts
+interface MessageFallbackPart {
+  type: "fallback";
+  source: string;
+}
+```
+
+Formatting functions defined in the registry
```
stasm, line 163:
```suggestion
### Formatting functions defined in the registry
```

stasm: Or, perhaps even better:

```suggestion
### Registry definition of formatted parts
```

exploration/formatted-parts.md
```diff
@@ -0,0 +1,212 @@
...
+}
+```
+
+For MessageExpressionPart, the `source` corresponds to the expression's fallback value.
+The `dir` and `locale` attributes of a part may be inherited from the message
+or from the operand (if present),
+or overridden by an expression attribute or formatting function,
+or otherwise set by the implementation.
+Each part should have at most one of `value` or `parts` defined;
+some may have none.
```
stasm, line 97: While this may be how this gets implemented in certain implementations, I think our definition should be stricter. Something like:

```ts
type MessageExpressionPart = MessageSingleValuePart | MessageMultiValuePart;

interface MessageSingleValuePart {
  type: string;
  source: string;
  value?: unknown;
  dir?: "ltr" | "rtl" | "auto";
  locale?: string;
}

interface MessageMultiValuePart {
  type: string;
  source: string;
  parts?: Iterable<{ type: string; value: unknown; source?: string }>;
  dir?: "ltr" | "rtl" | "auto";
  locale?: string;
}
```

exploration/formatted-parts.md
```diff
@@ -0,0 +1,212 @@
...
+}
+```
+
+Formatting functions defined in the registry
+Each function defined in the registry MUST define its "formatted-parts" representation.
+A function can define either a unitary string `value` or a `parts` representation.
+Where possible, a function SHOULD provide a `parts` representation
+if its output might reasonably consist of multiple fields.
+Where available, such a formatted value should itself be represented by `parts`
+rather than a unitary string `value`.
```
stasm, line 169: What does this mean, given the sentence above it? The design doesn't allow both `value` and `parts` to be defined on a single part, does it?

exploration/formatted-parts.md
```diff
@@ -0,0 +1,212 @@
...
+or overridden by an expression attribute or formatting function,
+or otherwise set by the implementation.
+Each part should have at most one of `value` or `parts` defined;
+some may have none.
+
+```ts
+interface MessageExpressionPart {
+  type: string;
+  source: string;
+  parts?: Iterable<{ type: string; value: unknown; source?: string }>;
```
stasm, line 103: Let's call this `subparts`, `items`, `values`, `chunks`, `fragments`... anything but `parts`, which we already use to describe the things returned by the `formatToParts` API.

exploration/formatted-parts.md
```diff
@@ -0,0 +1,212 @@
...
+  value?: unknown;
```
stasm, line 104: I think it would be more robust to describe the polymorphism of parts through generics rather than `unknown`. Something like the following:

```ts
interface MessageSingleValuePart<T> {
  type: string;
  source: string;
  value: T;
  dir?: "ltr" | "rtl" | "auto";
  locale?: string;
}

type MessageStringPart = MessageSingleValuePart<string>;
```

stasm (changes requested):

exploration/0003-formatted-parts.md
stasm: In https://github.com/unicode-org/message-format-wg/pull/463#discussion_r1406427143 I'm suggesting to define separate interfaces for single-valued and multi-valued parts. This could extend to fallback parts and markup, as well.

mihnita:
Sorry, in vacation last week.

> Would your concerns about the proposed API be satisfied if it could be shown how it could represent the above information you claim it cannot represent?

Sorry, but no.

I listed several concerns, so satisfying one is not enough.

Recap:
 1. overlapping ranges
 2. alternate representations on the same range (for example "27/12/2023" as text, and "December 27, 2023" as TTS)
 3. moving objects between server / client or across programming languages (for example PHP, or Go, calling a C++ layer implementation)
 4. "impedance" with the already established way the existing OS / tech stack does things

The "impedance" is a problem even if 1-3 can be represented with the above proposal, because the representation does not feel "natural" to a system that already uses attribute-style APIs.

Such a system would have several options:
- implement the format to parts as this spec proposal says, which will not feel "natural" to such system
- implement both this proposal, and something that matches theirs style => duplicate effort for the implementer, and "noise" for the user (_"Why two APIs? Which one should I use?"_)
- implement only something that feels natural, not this proposal => then being accused or not implementing the full MF2 standard (_"Yes, they have something, but it is not the standard, why can't they follow standards? Evil company X!"_)

aphillips:
@mihnita Thanks for your comment. Do you have an alternate design approach to suggest?

exploration/formatted-parts.md
eemeli: This sounds like a sensible way to specify this, should this design doc be accepted. Thought with a required `parts` to explicitly differentiate the two.

exploration/formatted-parts.md
eemeli: I'm pretty sure this is a remnant of some editorial work, and it's saying the same as the previous sentence but more complicatedly.

```suggestion
```

eemeli:
> I listed several concerns, so satisfying one is not enough.

Addressing each below.

> 1. overlapping ranges

The example message you provided could be represented like [this](https://gist.github.com/eemeli/ba0e1f43771a75e3fe5befb97b131914). If there is an example of a real-world message with improperly nesting ranges that you could point at, I'd be happy to show how such could be represented as well.

> 2. alternate representations on the same range (for example "27/12/2023" as text, and "December 27, 2023" as TTS)

Should the target not be an input for the date formatter? I mean, if an environment sometimes has its messages formatted for display on a screen and sometimes for TTS, would it not be reasonable to expect that this target is an input to the date formatter as well? If this distinction is made only after formatting, then I would presume that the MF2-internal date formatter that it's using is a custom one, which emits parts that include a Date object of some sort, along with a basket of formatting options. That's certainly doable with the proposed approach.

> 3. moving objects between server / client or across programming languages (for example PHP, or Go, calling a C++ layer implementation)

Could you clarify how this is a concern?

> 4. "impedance" with the already established way the existing OS / tech stack does things

In case you missed it earlier, I've [here](https://gist.github.com/eemeli/e081c442b524878313725943a45dff58) provided an example `partsToRanges()` function that can bridge this sort of impedance. Is that not sufficiently simple to solve the impedance issues?

exploration/formatted-parts.md
stasm: I mean to use it as a language for describing the concepts introduced by this doc; `MessageStringPart` is one example.

eemeli:
> One use case is that `formatToParts` is executed in one environment (server, or C++ layer), and the result is passed to another layer (client, or JS / PHP / Go / Dart / whatever). ( I will use server / client for all cases, to simplify the discussion, but it is intended to cover everything)
> 
> If the a result has objects (`value?: unknown`), and to get the string representation of that object means I need to invoke a "stringify" operation on it, this means that I have to somehow copy and convert that object between layers. Potentially with executable code in it.

Another use case for MF2 is formatting in front-end environments that do not have such restrictions. For example, a seemingly simple message like
```
Click to continue: {$button}
```
could have its `$button` be a reference to a React component which must retain a connection to the same React instance with which it was created. In other words, it is an opaque value that is not stringifiable, and it is not transmissible over any boundaries.

To format that message, we cannot format it to a string because then we'd end up with `[object Object]`. So we format it to parts. And to allow that representation to represent this React component, the spec definition _must_ include something like `value?: unknown`.

Would you agree that the above is a reasonable use case for MF2? If so, do you agree that this unfortunately leaves `value?: unknown` as the common denominator?

In the scenario you present, there are obviously further constraints that must be accounted for. These constraints may then be applied by the specific MF2 implementation, the functions you're using, or by checks on the post-processing for formatted-parts output. These restrictions _should not_ limit what may be done by a different user of MF2.

> In one of the example above we had
> 
> ```
>   { type: 'number', source: '|98|', locale: 'fr', parts: [{ type: 'integer', value: '98' }] },
> ```
> 
> What if the locale was Arabic?

Then the corresponding formatted-parts results would be:
```
{ type: 'number', source: '|98|', locale: 'ar', parts: [{ type: 'integer', value: '٩٨' }] }
```

The formatted parts are fully formatted parts.

eemeli:
@mihnita Are you satisfied by how each of your four concerns has been addressed?

exploration/0003-formatted-parts.md
```diff
@@ -0,0 +1,187 @@
...
+that are not necessarily best represented by a plain concatenated string.
+
+It would be useful to define a formatted-parts target for MessageFormat 2.
+
+## Background
+
+Past examples have shown us that if we don't provide a formatter to parts,
+the string output will be re-parsed and re-processed by users.
+
+## Use-Cases
```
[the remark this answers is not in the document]
eemeli: @aphillips This section has been expanded since this comment. Sufficiently?

exploration/formatted-parts.md
eemeli: Added the interfaces.

exploration/formatted-parts.md
eemeli: I added generic interfaces in line with your suggestion, but they're now referred to in the doc via comments, so that each part can retain its full definition in-place and a reader doesn't need to (but may) look up how they relate to each other. This also allows us to continue expressing how e.g. MessageFallbackPart explicitly does not have `locale` and `dir` fields, even if it does otherwise extend `MessageSingleValuePart<"fallback", never>`.

mihnita:
> @mihnita Are you satisfied by how each of your four concerns has been addressed?

I don't think I see any of my concerns addressed.

The only change I see is ["Add MessageSingleValuePart & MessageMultiValuePart definitions; other…"](https://github.com/unicode-org/message-format-wg/pull/463/commits/1645aeafeebb0e46b076499f988bd6347d288777)

Which splits the previous `MessageExpressionPart` (which had an `Iterable`) into 
`MessageSingleValuePart` and `MessageMultiValuePart` (which has an `Iterable`, and kind of corresponds to the old `MessageExpressionPart`).

Then `value: unknown` changes to `value: V`, which is as unknown as before.
Has all the limitations I described.
I still don't know how to send it over the wire, and I still don't know how to stringify it.
Unless we say that one of the special cases is that `unknown` is in fact a string (or something that extends string, to be more precise)?

It does not solve the "one level tree" problem, unless we say that `V` can be a `MessageMultiValuePart`
Then the whole thing becomes a deeper tree.
Which would address the deeper nesting (placeholder => range of dates with start / end date, each with fields)

---

>> In one of the example above we had
>>   { type: 'number', source: '|98|', locale: 'fr', parts: [{ type: 'integer', value: '98' }] },
>> What if the locale was Arabic?

> Then the corresponding formatted-parts results would be:
> { type: 'number', source: '|98|', locale: 'ar', parts: [{ type: 'integer', value: '٩٨' }] }
> The formatted parts are fully formatted parts.

So that means that in this case value is a string (or something that extends string).
Which is in fact what the parts returned by `Intl.DateTimeFormat` are, strings.

But in general "type" should not really be a type, is semantic, no?
Would be something like "year", or "day" (which happen to be numbers)
So the in example above "number" is really coming from the function that was used (`:number`), not a type.

Then maybe we need a name better than `type`.
It does not describe the type of `value` (which in some programming language is available at runtime anyway, with something like `typeof`)

---

You bring in React. And I get it.

> Would you agree that the above is a reasonable use case for MF2? If so, do you agree that this unfortunately leaves value?: unknown as the common denominator?

Absolutely.

---

To summarize the current status, as I understand it:

> 1. overlapping ranges

Because we have a bunch of `unknown` (ok `V` now), we can say "oh, but the value can be a `MessageTextPart`"
So one can imagine a placeholder that is a list formatter, where each element is a date range, each with start / end, and each with fields (day, month, year)
You get deeper nesting, a tree (because we "abuse" the value to be anything we want, so we put a full tree there)

> 2. alternate representations on the same range (for example "27/12/2023" as text, and "December 27, 2023" as TTS)

Still not sure how to represent alternate streams of text (for screen rendering vs TTS)
Of course, by the magic of "unknown" we can put any object there. So you can say that I can put a map with alternate text values...

In general most things can be "solved" just because there are unknown types that can store anything.

But that way things get soo flexible that there is no point to have this in the spec anymore.
If I can write a JS implementation, all in JS, following the spec 100%, and you can write a JS implementation, also all in JS, also following the spec 100%, but one can't swap the two implementations, it means the whole thing is kind of pointless.

> 3. moving objects between server / client or across programming languages (for example PHP, or Go, calling a C++ layer implementation)

Still unclear. Except by saying "hey, the value V can be anything, even JSON or protobuf (or string).
Again, the magic of "unknown"

> 4. "impedance" with the already established way the existing OS / tech stack does things

There is no solution for this, except for MF2 format to parts to return an annotated string.

mihnita:
> @mihnita Thanks for your comment. Do you have an alternate design approach to suggest?

Yes, an annotated string.
Which is conceptually what every existing framework seems to use.

And if this is a design document (as the title says) should at least document that option, compare with the current tree of unknown nodes, and explain why not do what everything else does.

mihnita:
Still open: we don't need the `MessageBiDiIsolationPart`

Let me explain again, in a different way, maybe I can be more clear:

1. how are these kind of parts generated?

Somehow MF takes the parse tree, with input arguments, and generates parts.

And the `MessageBiDiIsolationPart` are:

a. not generated from a bidi control character typed by a translator (see Eemeli's answer in a thread above)

b. not an from an explicit placeholder

Example:
"You have a message from {$user :person_name level=casual}"

Result, something like this:
```
{
   { type: "text" value: "You have a message from" } // MessageTextPart
  { type: "person_name " source: "$user" dir: "rtl", locale: "ar" value: "فلانة"  // MessageSingleValuePart
}
```

We already have info about the `MessageSingleValuePart`, that is Arabic, and direction is rtl.

The only place where `MessageBiDiIsolationPart` might be handy is before / after the `MessageSingleValuePart`
But we have all the info we need in the result above, without `MessageBiDiIsolationPart`.
Enough info that when we take that result we can add real bidi control characters, or HTML attributes, or elements):
That is stuff I do when I take the format to parts and generate the DOM, or spanned text.
It is a kind of post-processing on the result, similar to changing case, or making a section bold.

mihnita:
> making a section bold.

Speaking of bold: we have no parts for markdown.

eemeli:
> You bring in React. And I get it.
> 
> > Would you agree that the above is a reasonable use case for MF2? If so, do you agree that this unfortunately leaves value?: unknown as the common denominator?
> 
> Absolutely.

Good! So we agree that at the most generic level, a formatted part returned by an arbitrary MF2 implementation should have a shape conforming to `MessageSingleValuePart<string, unknown>`. The `unknown` value there corresponds to e.g. the `Object what` in Android Spannables.

And that's what this spec is specifying; the most generic possible definition within which _**all**_ parts for _**all**_ placeholders in _**all**_ implementations must fit. So we need to leave a lot of space for the unknown, because we need to support all possible formatting targets (not just strings), while still explicitly defining what structure we can, and common fields such as `source` and `dir`.

Any implementation is of course allowed to be stricter than what's defined here in its outputs, and processors of formatted parts are allowed to specify that they require e.g. only string values, or assign other restrictions to what they're capable of dealing with. But these limitations are not universal, and do not apply to all cases where formatted parts are emitted or consumed.

> > > In one of the example above we had
> > > { type: 'number', source: '|98|', locale: 'fr', parts: [{ type: 'integer', value: '98' }] },
> > > What if the locale was Arabic?
> 
> > Then the corresponding formatted-parts results would be:
> > { type: 'number', source: '|98|', locale: 'ar', parts: [{ type: 'integer', value: '٩٨' }] }
> > The formatted parts are fully formatted parts.
> 
> So that means that in this case value is a string (or something that extends string). Which is in fact what the parts returned by `Intl.DateTimeFormat` are, strings.
> 
> But in general "type" should not really be a type, is semantic, no? Would be something like "year", or "day" (which happen to be numbers) So the in example above "number" is really coming from the function that was used (`:number`), not a type.
> 
> Then maybe we need a name better than `type`. It does not describe the type of `value` (which in some programming language is available at runtime anyway, with something like `typeof`)

The `type` here is specifying the type of the part, not of the value. The current proposal matches the formatted-parts structure used by the JS Intl formatters, which indeed always have string values. And for example the JS implementation's `'number'` formatted parts will always have `string` values, which is allowed because it's a subset of `unknown`.

> > 2. alternate representations on the same range (for example "27/12/2023" as text, and "December 27, 2023" as TTS)
> 
> Still not sure how to represent alternate streams of text (for screen rendering vs TTS) Of course, by the magic of "unknown" we can put any object there. So you can say that I can put a map with alternate text values...
> 
> In general most things can be "solved" just because there are unknown types that can store anything.

Yes. Because we must allow for arbitrary values like React components to show up as values, we must allow for the possibility that those values are not primitive, but contain some internal structure.

> But that way things get soo flexible that there is no point to have this in the spec anymore. If I can write a JS implementation, all in JS, following the spec 100%, and you can write a JS implementation, also all in JS, also following the spec 100%, but one can't swap the two implementations, it means the whole thing is kind of pointless.

This thing is not pointless. An emitter or consumer of formatted parts can be _more specific_ than this spec in what it emits or consumes, allowing for interoperability.

It also has a rather important point in being a prerequisite for our markup definition.

> > @mihnita Thanks for your comment. Do you have an alternate design approach to suggest?
> 
> Yes, an annotated string. Which is conceptually what every existing framework seems to use.
> 
> And if this is a design document (as the title says) should at least document that option, compare with the current tree of unknown nodes, and explain why not do what everything else does.

Could you expand that alternative from its current definition? Right now it just defines it as "Format to a string, but separately define metadata or other values." It seems like you think this should be expanded?

> > making a section bold.
> 
> Speaking of bold: we have no parts for markdown.

Presuming that you mean "markup", then you'll find those included in the proposed design of the doc linked to from #537. Unless you really do mean "markdown", in which case I don't know what you mean.

eemeli:
> What about this: have both an `unknown` value, and a string one? (let's call it `as_string` for now)

This sounds like something that an implementation could choose to do, but I don't really see why it should be enforced in all cases for all users?

As far as I know, practically all reasonable programming languages allow for arbitrary objects to define a way for their own stringification when they're coerced to a string. In both JS and Java, that can be done by implementing a custom `.toString()` method.

What's the benefit or requiring all formatters to always perform such a stringification?

> Since `{$button}` is a placeholder, there is a function associated with it (explicit, `{$button :react}`, or implied from the type) That function understands React objects. Potentially how to get the string representation from one. The part it produces might contain the react ref as a `value` AND the string form.

At least in the JS Intl implementation, a value like that without an explicitly recognised type will be formatted to a string by calling `String(button)`, and formatted to parts as
```js
{ type: 'unknown', source: '$button', value: button }
```

This way, arbitrary input values can be accepted and "formatted" without the core implementation needing to understand them, or being passed functions supporting their handling.

mihnita:
> This sounds like something that an implementation could choose to do

But then it is not following the spec.

> As far as I know, practically all reasonable programming languages allow for arbitrary objects to define a way for their own stringification

Do you know about C and C++? There is not standard way for an object to stringify itself.
Maybe they are not "reasonable programming languages" by your judgement, but they are still high ranking.
One of the ICU implementations in C/C++, and the Windows APIs are C.
So that means that ICU4C and Windows are not reasonable, and we should deny them the right to implement this standard.

Second, even programming languages that have standard ways to objects to stringify themselves, that is intended for debugging and for developer consumption, not for end user.
They are not necessarily readable, often intentionally so, and are locale independent.
So a React button with a label saying "Subscribe" will not have a toString implementation returning "Subscribe" in French.
Which is what you need to do linguistic fixes ("La [button]" to "L'[button]")

Third, saying that objects can stringify themselves ignores one of my requirements to be able to easily move "across borders" between programming languages, or server/client.

And if possible be consistent?
First (in https://github.com/unicode-org/message-format-wg/pull/463#discussion_r1311507386) you give an implementation of `stringifyParts` that require objects to be stringifiable.
Then you give example of React and you say "we cannot format it to a string because then we'd end up with [object Object]."
Then you say that objects that can stringify themselves are are everywhere (except for React, I suppose?)
What is it?

I really tried to see if we can make this proposal work your way, with "a tree of nodes."
I didn't push too much  for the attributed strings. At least not in the last few days of interactions.
But answering piece-meal, with one answer contradicting previous answers, or solving one issue, but not ignoring other, will not move the discutient forward.

Do you think that all 4 bullets listed https://github.com/unicode-org/message-format-wg/pull/463#issuecomment-1828254194 are reasonable?
4 can't be solved, with "tree of parts" not matching the attributed strings already used by iOS, Android, and ICU.
At least the other 1-3 are reasonable?

exploration/formatted-parts.md
```diff
@@ -0,0 +1,230 @@
...
+```
+
+### Registry definition of formatted parts
+
+Each function defined in the registry MUST define its "formatted-parts" representation.
+A function can define either a unitary string `value` or a `parts` representation.
+Where possible, a function SHOULD provide a `parts` representation
+if its output might reasonably consist of multiple fields.
+In most cases, these sub-parts should not need fields beyond their `type` and a string `value`,
+Where necessary, other `value` types may be used and other fields such as a `source` included.
```
eemeli, line 184:
```suggestion
In most cases, these sub-parts should not need fields beyond their `type` and a string `value`.
Where necessary, other `value` types may be used
and other fields such as a `source` included in the sub-parts,
and additional fields included in the `MessageExpressionPart`.
```

exploration/formatted-parts.md
```diff
@@ -0,0 +1,232 @@
...
+### Registry definition of formatted parts
+
+Each function defined in the registry MUST define its "formatted-parts" representation.
+A function can define either a unitary string `value` or a `parts` representation.
+Where possible, a function SHOULD provide a `parts` representation
+if its output might reasonably consist of multiple fields.
+In most cases, these sub-parts should not need fields beyond their `type` and a string `value`.
+Where necessary, other `value` types may be used
+and other fields such as a `source` included in the sub-parts,
+and additional fields included in the `MessageExpressionPart`.
```
eemeli, line 186:
```suggestion
and additional fields may be included in the `MessageExpressionPart`.
```

eemeli:
> > This sounds like something that an implementation could choose to do
> 
> But then it is not following the spec.

Yes, it would be. In case that was not implicitly clear (we're using TypeScript syntax, after all), I've added an explicit mention of the liberty to define additional fields.

> > As far as I know, practically all reasonable programming languages allow for arbitrary objects to define a way for their own stringification
> 
> Do you know about C and C++? There is not standard way for an object to stringify itself. Maybe they are not "reasonable programming languages" by your judgement, but they are still high ranking. One of the ICU implementations in C/C++, and the Windows APIs are C. So that means that ICU4C and Windows are not reasonable, and we should deny them the right to implement this standard.

My most sincere apologies. I thought `operator<<` would be considered as the standard way for C++, and honestly didn't think of C users who did want to format "arbitrary object" values without a framework like GObject that does provide for a standard stringification interface.

Now, leaving the snark aside (which tbh I'd also appreciate from you), do you see how _requiring_ an `as_string` property would mean that a bare C implementation would not be able to support truly arbitrary `void *` values as input arguments for a format-to-parts MF2 function? Adding such a required property on all parts severely restricts such an implementation.

> Second, even programming languages that have standard ways to objects to stringify themselves, that is intended for debugging and for developer consumption, not for end user. They are not necessarily readable, often intentionally so, and are locale independent. So a React button with a label saying "Subscribe" will not have a toString implementation returning "Subscribe" in French. Which is what you need to do linguistic fixes ("La [button]" to "L'[button]")

To reflect, we are here talking about formatting a message containing a placeholder `{$button}` (note, no annotation) in JavaScript, where the MF2 implementation will be powered by the ECMA-402 defined `Intl.MessageFormat` interface. Its current spec, and that proposed in the current design here, allow for input values of unknown type to pass through the parts formatter and come out the other side as
```js
{ type: 'unknown', source: '$button', value: button }
```
where the `button` is the exact unmodified value that was passed as an argument to the `formatToParts()` method.

My understanding of what you're asking for is that even in this case, that formatted part should include something like `as_string`, yes? Please do correct me if I've misunderstood your position. Now, given that the core Intl.MF implementation is not going to have any React dependency hard-coded within it, any such `as_string` value will be generated by calling `String(button)`, which will mean that it'll be `"[object Object]"`.

Using such an implementation (ignoring any `as_string` fields, as they'll be useless), it's certainly possible for a user to take the formatted parts of a message, and _if the user code can handle React components_, to render the component part and extract localized text from the ultimately resolving HTML before feeding it into a system capable of applying linguistic fixes.

We are here defining the most specific and minimal data types that still enable for something like `Intl.MessageFormat` to be able to support behaviour as above. Enforcing an `as_string` field won't help with any of it.

> Third, saying that objects can stringify themselves ignores one of my requirements to be able to easily move "across borders" between programming languages, or server/client.

Wait, I thought we'd agreed that something like `value: unknown` is required to support arbitrary objects, which cannot be transmitted across realms like that? There is literally no representation that allows for their transmission as you require, but we still want to support them in formatted parts.

What you're asking for is possible with further restrictions on what's enabled by this proposal. It is perfectly valid for e.g. a server-side formatter to have stricter requirements for what it supports as input data than just "anything goes", and for it to make stricter guarantees about its outputs, as required for transmission across the wire, for instance. On the other hand, it is also perfectly valid for a formatter to _not_ impose such restrictions on its inputs, and _not_ to promise anything stricter than what's in this proposal.

If we were to restrict this proposal to the requirements of "formatted parts must be transmissible as data", we could not support React components. And you said this earlier:

> > Would you agree that the above is a reasonable use case for MF2? If so, do you agree that this unfortunately leaves value?: unknown as the common denominator?
>
> Absolutely.

Could you now clarify what you meant by "Absolutely", because I took that to be agreement with my preceding statement?

> And if possible be consistent? First (in [#463 (comment)](https://github.com/unicode-org/message-format-wg/pull/463#discussion_r1311507386)) you give an implementation of `stringifyParts` that require objects to be stringifiable. Then you give example of React and you say "we cannot format it to a string because then we'd end up with [object Object]." Then you say that objects that can stringify themselves are are everywhere (except for React, I suppose?) What is it?

My statements above do not contradict each other. In JS, _anything_ can be coerced to a string with `String()`, it's just that for most object the resulting `"[object Object]"` is useless. In the comment that you link to, that `stringifyParts()` will indeed match exactly the `formatToString()` output. If the input contained e.g. a React component, it would end up in the string as `"[object Object]"` either way.

> I really tried to see if we can make this proposal work your way, with "a tree of nodes." I didn't push too much for the attributed strings. At least not in the last few days of interactions. But answering piece-meal, with one answer contradicting previous answers, or solving one issue, but not ignoring other, will not move the discutient forward.

I do not think I will be able to convince you. I think you've decided that the approach proposed here isn't good, and you refuse to ever admit being wrong. I find discussions with you challenging, and they have been literally (not figuratively) the most draining part of my work for the past few years.

You refuse to discuss any one item at a time, pivoting to others when it seems like continuing on a track will lead to a conclusion you don't like, and then when returning you discard any previous developments not to your liking. Sure, this might be just me, but I much prefer having this dialogue written down, where it's recorded.

> Do you think that all 4 bullets listed [#463 (comment)](https://github.com/unicode-org/message-format-wg/pull/463#issuecomment-1828254194) are reasonable? 4 can't be solved, with "tree of parts" not matching the attributed strings already used by iOS, Android, and ICU. At least the other 1-3 are reasonable?

All your concerns are reasonable, but they should also be considered resolved by my preceding answers. Specifically, I've answered 1, 2, and 4 in https://github.com/unicode-org/message-format-wg/pull/463#issuecomment-1828303960 and 3 in https://github.com/unicode-org/message-format-wg/pull/463#issuecomment-1828591656. If those answers are not satisfactory, then please re-read them, and tell me specifically why or how they do not address the concerns you've previously expressed, or if they contain any parts that you do not understand. Our discussion since then seems to have been going around in circles.

Also, I reiterate my earlier request to expand on the attributed string alternative in this proposal, in case you find it lacking, and should you wish for it to be seriously considered as an alternative.

stasm (changes requested):
Observing the ongoing discussion, I'd like to propose to descope formatted parts from 2.0.

I think we should still recommend that implementations provide an API returning parts, but re-reading the doc, I don't think there's a compelling case for why to define such API in any more detail than what's currently presented in the "Requirements" section.

In fact, it occurs to me that our written requirements *are* the recommendation for implementers:

> - Define an iterable sequence of formatted part objects.
> - Include metadata for each part, such as type, source, direction, and locale.
> - Allow the representation of non-string values.
> - Allow the representation of values that consist of an iterable sequence of formatted parts.
> - Be able to represent each resolved value of a pattern with any number of formatted parts, including none.
> - *[Removed the last one, because it's the only one that's not a recommendation to implementers.]*

We struggle because we want to be super-agnostic on one side, and on the other, we want to specify *something concrete* so that compliant implementations can be swapped and can communicate with each other. However, I note that runtime fungibility and IPC are not listed in the use-cases of this design. Are we solving the right thing?

If runtime interoperability is not a goal, what do we actually gain from defining this *optional* agnostic interface as part of the spec?

PR #465, opened 2023-09-02
Explain why we use `|` and not `"` and/or `'`

eemeli:
Adds an explanation for why the _literal_ syntax uses `|` delimiters. I think the explanation is pretty flimsy, but we don't have a better one and I've not been able to convince the WG to change it with #414, which should be closed if this is merged.

### DIFF
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -596,6 +596,15 @@ A **_<dfn>quoted</dfn>_** literal begins and ends with U+005E VERTICAL BAR `|`.
 The characters `\` and `|` within a _quoted_ literal MUST be
 escaped as `\\` and `\|`.
 
+The `|` is used as a quote delimiter
+because it is not commonly used as a quote delimiter.
+Using something more ordinary like `'` or `"`
+would mean that they would need to be escaped
+if MessageFormat 2 syntax needed to be included in a resource format
+that used the same quote delimiter.
+Supporting both `'` and `"` as quote delimiters would potentially require
+the delimiters to be manually escaped when copying them between resource formats.
+
 An **_<dfn>unquoted</dfn>_** literal is a _literal_ that does not require the `|`
 quotes around it to be distinct from the rest of the _message_ syntax.
 An _unquoted_ MAY be used when the content of the _literal_


spec/syntax.md
```diff
@@ -596,6 +596,15 @@ A **_<dfn>quoted</dfn>_** literal begins and ends with U+005E VERTICAL BAR `|`.
...
 escaped as `\\` and `\|`.
 
+The `|` is used as a quote delimiter
+because it is not commonly used as a quote delimiter.
+Using something more ordinary like `'` or `"`
+would mean that they would need to be escaped
+if MessageFormat 2 syntax needed to be included in a resource format
+that used the same quote delimiter.
+Supporting both `'` and `"` as quote delimiters would potentially require
+the delimiters to be manually escaped when copying them between resource formats.
```
aphillips, line 606: `'` and `"` are not special to us, so we can lump these in with other characters that we also rejected while searching for characters that are not common in natural language or programming syntax to use as a quote character.

```suggestion
This is important because the quote delimiting character
has to be escaped whenever it appears in _text_ in a _message_
or in the body of a literal. 
Characters such as `'` and `"` are known to be common in ordinary
text, as well as having meaning in various syntaxes.
In particular, our format will be embedded into various resource formats,
many of which use `"` characters to delimit message values. 
When our escape and quote characters conflict with those of the host syntax,
this would require the user to escape our quotes
and to double-escape our quoted quotes.
These levels of escaping will make the _message_ difficult to read
in various resource formats.
Experience with special meaning for the apostrophe (`'`) in MF1
also informs this choice.
```

> **Example**
>
>```
>{"Quoted \"literal\""}
>```
> when embedded into (say) a JSON file might require the developer to type:
>```json
>"message": "{\"Quoted \\\"literal\\\"\"}"
>```

Make sense?

aphillips (changes requested):
It is worth writing this down in the spec. Thanks for proposing this.

See if my proposal below helps draw the sting of not adopting quote marks 😉

spec/syntax.md
eemeli: > When our escape and quote characters conflict with those of the host syntax,
> this would require the user to escape our quotes
> and to double-escape our quoted quotes.

This is why #414 suggests allowing either `'` or `"` as _quoted_ delimiters. That way, your example could be expressed as
```
{'Quoted "literal"'}
```
and be embedded in JSON as
```json
"message": "{'Quoted \"literal\"'}"
```
so only the `"` that are actual characters of the literal need to be quoted, exactly as is the case with `|`.

Hence our explanation ought to say why the above is a worse option than picking a novel quote delimiter `|`. This is missing from your suggestion.

aphillips: That works in JSON, but not in many syntaxes (which are more rigid about quotes). And the quote alternation can depend on the exterior quoting regime not visible to e.g. translation tooling. Changes to the quote characters look like differences to processes such as translation memory.

gibson042: It might be helpful to just embed something like that example, e.g.

> For example, `message = '{|Eat At Joe\'s|}'` (using `|` as the MessageFormat quote delimiter and using the embedding language's syntax to include a literal `'` in a `'`-delimited string containing the message) is more readable than the alternative `message = '{\'Eat At Joe\\\'s\'}'` (using `'` as the MessageFormat quote delimiter, which must itself be escaped in the embedding language along with both characters of the `\'` MessageFormat-level sequence for including a raw `'` in a `'`-quoted literal).

eemeli: > That works in JSON, but not in many syntaxes (which are more rigid about quotes).

Could you give an example? I'm not sure how a resource syntax could be more rigid about quotes than JSON.

> And the quote alternation can depend on the exterior quoting regime not visible to e.g. translation tooling. Changes to the quote characters look like differences to processes such as translation memory.

Sure, if you're comparing the exact string representations of messages. However, a reasonable MF2 message comparator would either compare the data model representations of messages, or their understanding of each message's canonical source strings, which would normalise away any whitespace or quotation style differences.

spec/syntax.md
```diff
@@ -596,6 +596,15 @@ A **_<dfn>quoted</dfn>_** literal begins and ends with U+005E VERTICAL BAR `|`.
 The characters `\` and `|` within a _quoted_ literal MUST be
 escaped as `\\` and `\|`.
 
+The `|` is used as a quote delimiter
+because it is not commonly used as a quote delimiter.
```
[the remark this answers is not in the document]
stasm: A comment on @aphillips's suggestion:

> or in syntaxes used by various programming and scripting languages.

It isn't just that vertical pipes are rare in other syntaxes, but specifically that they are rare as quote delimiters.

spec/syntax.md
stasm: There are 2 conveniences that `|` give us:

1. "inwards", it's easy to include `"` and `'` in literals, without escaping them via `\`,
2. "outwards", it's easy to embed messages with string literals in code or container formats.

The combination of these two is particularly interesting: it's when a message both contains quotes in literals and is embedded in code. If we delimited literals with `"` or `'`, then we'd often need to solve two problems at the same time:

* escape quotes inside literals via `\`, and
* escape quotes delimiting literals in code or container formats **and** escape the MF2's `\` from the bullet above, because it's a common escape char in many languages.

This is how we get the gnarly `\\\`. By using `|`, we sidestep both problems.

eemeli:
Closing in favour of #477; a design doc is a better way to rationalise the decision.

PR #466, opened 2023-09-04
Change open/close from function to expression

mihnita:
Replacing https://github.com/unicode-org/message-format-wg/pull/447

Merging was too messy, and my change was in main, not in a branch.

### DIFF
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -412,32 +412,32 @@ what _options_ and _option_ values are valid,
 and what outputs might result.
 See [function registry](./) for more information.
 
-_Functions_ can be _standalone_, or can be an _opening element_ or _closing element_.
+_Expressions_ can be _standalone_, or can be an _opening element_ or _closing element_.
 
-A **_<dfn>standalone</dfn>_** _function_ is not expected to be paired with another _function_.
-An **_<dfn>opening element</dfn>_** is a _function_ that SHOULD be paired with a _closing element_.
-A **_<dfn>closing element</dfn>_** is a _function_ that SHOULD be paired with an _opening element_.
+A **_<dfn>standalone</dfn>_** _expression_ is not expected to be paired with another _expression_.
+An **_<dfn>opening element</dfn>_** is a _expression_ that SHOULD be paired with a _closing element_.
+A **_<dfn>closing element</dfn>_** is a _expression_ that SHOULD be paired with an _opening element_.
 
 An _opening element_ MAY be present in a message without a corresponding _closing element_,
 and vice versa.
 
-> A _message_ with a _standalone_ _function_ operating on the _variable_ `$now`:
+> A _message_ with a _standalone_ _expression_ operating on the _variable_ `$now`:
 >
 > ```
 > {{$now :datetime}}
 > ```
 >
-> A _message_ with two markup-like _functions_, `button` and `link`,
+> A _message_ with two markup-like _expressions_, `button` and `link`,
 > which the runtime can use to construct a document tree structure for a UI framework:
 >
 > ```
-> {{+button}Submit{-button} or {+link}cancel{-link}.}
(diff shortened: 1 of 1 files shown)


eemeli (changes requested):
As proposed, this PR is a pretty direct `s/function/expression/` replacement in the middle half of the Expression » Annotation » Function section of the syntax spec. [Here's](https://github.com/mihnita/message-format-wg/blob/mihai_sync_spec_to_abnf/spec/syntax.md#function) how it currently looks like.

This kinda doesn't make sense, with half of the _function_ explanation suddenly talking about _expression_ instead. If we do want to describe open/close/standalone with reference to the expression as a whole, these paragraphs should be extracted into a separate section.

I'm also not convinced that this change is beneficial as:
1. the description of open/close with respect to _function_ is valid,
2. our formatting and registry is completely unaffected by the proposed changes and still considers the open/close sigils only during function lookup, and
3. I don't see how this is "Bringing the spec description in sync with the abnf".

spec/syntax.md
```diff
@@ -412,32 +412,32 @@ what _options_ and _option_ values are valid,
 and what outputs might result.
 See [function registry](./) for more information.
 
-_Functions_ can be _standalone_, or can be an _opening element_ or _closing element_.
+_Expressions_ can be _standalone_, or can be an _opening element_ or _closing element_.
```
aphillips, line 415: If I were to make this change, I would make "element" into expression too:

```suggestion
An _Expression_ can be _standalone_, or can be an _opening expression_ or _closing expression_.
```

spec/syntax.md
```diff
@@ -412,32 +412,32 @@ what _options_ and _option_ values are valid,
...
 
-A **_<dfn>standalone</dfn>_** _function_ is not expected to be paired with another _function_.
-An **_<dfn>opening element</dfn>_** is a _function_ that SHOULD be paired with a _closing element_.
-A **_<dfn>closing element</dfn>_** is a _function_ that SHOULD be paired with an _opening element_.
+A **_<dfn>standalone</dfn>_** _expression_ is not expected to be paired with another _expression_.
+An **_<dfn>opening element</dfn>_** is a _expression_ that SHOULD be paired with a _closing element_.
+A **_<dfn>closing element</dfn>_** is a _expression_ that SHOULD be paired with an _opening element_.
```
aphillips, line 419: ... and I would do that globally...

spec/syntax.md
```diff
@@ -412,32 +412,32 @@ what _options_ and _option_ values are valid,
...
+> {{button +html}Submit{button -html} or {link +html}cancel{link -html}.}
 > ```
 
 A _function_ consists of a prefix sigil followed by a _name_.
 The following sigils are used for _functions_:
 
-- `:` for a _standalone_ function
+- `:` for a _standalone_ expression
 - `+` for an _opening element_
 - `-` for a _closing element_
```
aphillips, line 442: expressions here too

aphillips (changes requested):
I agree with @eemeli that the title is probably wrong. The ABNF clearly makes the +- sigils part of function names and is quiet about open/close expressions. If we were to truly bring the spec into alignment with the ABNF, this would effectively say that `+foo`, `-foo` and `$foo` are not distinct from one another, or, if they are distinct, that the distinction has no effect on the parsing of the message.

We need to decide about what open and close mean and what, if any, restrictions go on them, how open and close "communicate" with each other or get paired, and so on. Or we need to reserve these differences to future versions or to implementations.

stasm (commented):
Looking at the current ABNF, `expressions` are what goes into the placeholder, including the case when there's no function. Consequently, I think this PR should instead talk about _function annotations_, which is our way of saying "function invocations" or "function application".

In other words, open/close are not a property of the function (an abstract entity that needs to be called to do anything), but instead are a property (in the current design) of the function's application/invocation/call, i.e. _annotation_.

PR #469, opened 2023-09-04
Design document for variable mutability and namespacing

aphillips:
Per today's call (2023-09-04) here is the start of the design doc to address the need to decide how to handle:

- variable mutability
- local vs. external variable "masking"

This is based in large part on the discussion in #310 coupled with some discussion elsewhere. I incorporated commentary liberally. 

The design presented here is for limited mutability (only via a separate keyword) and for separate namespaces.

_Let the games begin._

### DIFF
--- a/exploration/variable-mutability.md
+++ b/exploration/variable-mutability.md
@@ -174,8 +174,6 @@ The _expression_ rule can't be used directly in _input-declaration_ because the 
 
 With this approach, variables are immutable,
 so each may be defined by only one _declaration_.
-The order of declarations does not matter,
-to allow for e.g. `input` annotations to refer to `local` variables.
 
 References to later declarations are not allowed,
 so this is considered an error:
@@ -184,6 +182,13 @@ so this is considered an error:
 local $foo = {$bar :number}
 local $bar = {42 :number}
 {The answer is {$foo}}
+```
+
+Note that this means that `input` declarations can (and sometimes _must_)
+follow `local` ones, such as when an `input` is annotated using a `local` value:
+```
+local $foo = {|2| :number}
+input $bar :number maxFractionDigits={$foo}
 ```
 
 An _input-declaration_ is not required for each external variable.
--- a/exploration/variable-mutability.md
+++ b/exploration/variable-mutability.md
@@ -188,7 +188,7 @@ local $bar = {42 :number}
 
 An _input-declaration_ is not required for each external variable.
 A _local-declaration_ takes precedence and does not cause an error
-if a similarly named external variable is passed to the formatter
+if an identically named external variable is passed to the formatter
 _without_ a corresponding _input-declaration_ in the message.
 
 The use case of chaining operations on a variable with a single name is not supported here,
--- a/exploration/variable-mutability.md
+++ b/exploration/variable-mutability.md
@@ -267,3 +267,22 @@ this problem goes away. However, a local variable cannot be used to augment or a
 let #arg1 = {$arg1 :number maxFractionDigits=2}
 {Now I have to change {$arg} to {#arg}...}
 ```
+
+### All Variables are Immutable; Externals are Annotatable; Shared Namespace
+
+_This is @eemelie's proposal from the PR thread_
+
+If we add a new reserved keyword (the proposal says `input`) to
+allow annotation of external variables, and allow `let` to mask
+external values.
+
+```
+[ {"arg1": "37"},
+  {"arg2": "42"}
+]
+...
+input {$arg1 :number minFractionDigits=2}
+let $arg2 = {|wildebeest|}
+{{$arg1} prints 37.00 and {$arg2} prints wildebeest, no errors}
+...
+```
--- a/exploration/variable-mutability.md
+++ b/exploration/variable-mutability.md
@@ -39,6 +39,7 @@ _What use-cases do we see? Ideally, quote concrete examples._
 - Users want to reference external variables in expressions.
 - Users can modify external variables using declarations.
   For example, they can perform a text transformation or assign reusable formatting options.
+
   > ```
   > let $foo = {$bar :uppercase}
   > let $baz = {$someNumber :number groupingUsed=false}
@@ -51,6 +52,7 @@ _What use-cases do we see? Ideally, quote concrete examples._
   in the various pattern strings, as well as issues that could arise from
   (for example) translation memory systems recalling the old expression.
   For example:
+
   > ```
   > let $foo = {$foo :transform}
   > match {$a :plural} {$b :plural}
@@ -67,6 +69,7 @@ _What use-cases do we see? Ideally, quote concrete examples._
 
 - Users want to perform multiple transforms on a value.
   Since our syntax does not permit embedding or chaining, this requires multiple declarations.
+
   > ```
   > let $foo = {$foo :text-transform transform=uppercase}
   > let $foo = {$foo :trim}
@@ -82,12 +85,14 @@ _What use-cases do we see? Ideally, quote concrete examples._
   > ```
 
 - Users want to annotate external variables or literals:
+
   > ```
   > let $fooAsNumber = {$foo :number}
   > let $anotherNumber = {42 :number}
   > ```
 
 - Users may wish to provide complex annotations which are reused across mulitple patterns
+
   > ```
   > let $count = {$count :number}
   > let $date = {$date :datetime dateStyle=long}
@@ -111,6 +116,7 @@ _What use-cases do we see? Ideally, quote concrete examples._
 _What properties does the solution have to manifest to enable the use-cases above?_
 
 These were taken from a comment by @stasm in #310:
+
 - Be able to re-annotate variables without having to rename them in the message body
 - Allow static analysis to detect mistakes when referencing an undefined local variable
 - Be able to re-annotate variables multiple times (because we do not allow nesting)
@@ -198,7 +204,7 @@ Note: if we have separate namespaces then local variables don't
 require Unicode names because their namespace is not subject
 to external data requirements.
 
-A different option is to say: it is up to the user to avoid using 
+A different option is to say: it is up to the user to avoid using
 declared names that would confuse translators and others.
 This would mean that we provide no defense on the syntax level.
 
--- a/exploration/variable-mutability.md
+++ b/exploration/variable-mutability.md
@@ -21,7 +21,6 @@ _What is this proposal trying to achieve?_
 Describe how variables are named and how externally passed variables
 and internally defined variables interact.
 
-
 ## Background
 
 _What context is helpful to understand this proposal?_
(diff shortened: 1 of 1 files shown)


exploration/variable-mutability.md
```diff
@@ -0,0 +1,244 @@
...
+  > ```
+
+- Implementers need to know what value is associated with a named variable, see #299.
+
+## Requirements
+
+_What properties does the solution have to manifest to enable the use-cases above?_
+
+- Be able to re-annotate variables without having to rename them in the message body
+- Allow static analysis to detect mistakes when referencing an undefined local variable
```
eemeli, line 100:
```suggestion
```

I don't think this should be considered as a requirement. At best it's a nice-to-have feature improving static analysis when a single translated message needs to be considered in isolation, and not with respect to its form in the source locale.

exploration/variable-mutability.md
```diff
@@ -0,0 +1,244 @@
...
+
+_Describe the proposed solution. Consider syntax, formatting, errors, registry, tooling, interchange._
+
+Separate local variables from externally passed values by altering the sigil
+and by using a visually distinctive pattern for local names
+(in an effort to prevent `$foo`/`@foo` confusion).
+
+```abnf
+variable     = local_var / external_var
+local_var    = "@_" name
```
eemeli, line 125: Yeah, no. We categorically do not need any two-character prefixes.

exploration/variable-mutability.md
```diff
@@ -0,0 +1,244 @@
...
+
+It is a variable resolution error to call `modify`
+on an external variable that does not exist.
+
+> _Example_
+>
+> ```
+> let @_local = {$external :transform}
+> modify @_local = {@_local :modification with=options}
+> modify $external = {$external :transform adding=annotation}
```
eemeli, line 157: This looks an awful lot like we're doing pass-by-reference for formatting function arguments and we're modifying the `$external` value in the calling context. Is that intentional?

exploration/variable-mutability.md
```diff
@@ -0,0 +1,244 @@
...
+
+The choice here of `@_` as the local variable sigil is probably not distinctive enough.
+It is probably okay to be a little inconvenient with local variable naming
+as these are less common than external variables.
+Alternatives to consider:
+
+- `@@foo`
+- `@foo@`
+- `@!foo`
+- `@ONLY_UPPER_ASCII_SNAKE`
```
eemeli, line 180: These are all pretty ugly, but I presume that's intentional. Do we not want users to use local variables? Because that's what we're saying by making them ugly and clumsy.

exploration/variable-mutability.md
```diff
@@ -0,0 +1,244 @@
...
+There are no warnings or errors produced when this occurs, even when it is unintentional.
+
+If variables are mutable and namespaces are shared, it's easy to write a message that never
+fails but does produce unintended or unexpected results (from the caller's point of view).
+
+```
+{"arg1": "10000"}
+...
+let $arg1 = {42}
+{This always says {$arg1} == 42}
```
eemeli, line 209: The proposed design has this same flaw:
```
let @_arg1 = {42}
{This always says {@_arg1} == 42}
```

Anyone not working with MF2 daily will easily forget when looking at that message which prefix is for external and which for internal ones, leading to the exact same failure mode.

Is this the only argument against this alternative?

exploration/variable-mutability.md
```diff
@@ -0,0 +1,244 @@
...
+
+### All Variables are Immutable; Shared Namespace
+
+If we make all variables immutable and external and local vars share a namespace,
+passing an argument that shares a name with a local declaration can cause a message to fail.
+
+```
+{ "arg1": "37"}
+...
+let $arg1 = {|42| :number maxFractionDigits=2}  // error
```
eemeli, line 231: Not throwing an error for this should be considered a requirement.

eemeli (commented):
A first pass.

exploration/variable-mutability.md
aphillips: Would you prefer a use case for this? This specific list was taken from @stam's comment in #310 near the end (after I had written a longer and more detailed set of requirements). The "static analysis" requirement is kind of "boiling down" the question of whether people will make mistakes with overwriting.

exploration/variable-mutability.md
aphillips: See later discussion. This is horrifying while we agree on something...

exploration/variable-mutability.md
aphillips: Is there a beautiful way to make locale variables and not have them be confused with external variables "of the same name"? 

An alternative not presented is the "rope merchant" option: do nothing (ugly or otherwise) and let people adopt whatever convention prevents mistakes, e.g. `$foo` and `#foo` are separate values (external vs. local).

Note: I just recalled that we wanted to use `@` for annotation, so will change that to `#` globally.

exploration/variable-mutability.md
aphillips: In this option, this is an error, because one is attempting to overwrite `$arg1` with a new declaration. Presumably if we chose this option, it would be to **_ensure_** that this was an error.

I happen to agree that there is a use case that needs to be accounted for here (I have made the argument elsewhere many times about declarations not knowing what all the values are externally)

exploration/variable-mutability.md
aphillips: It intentionally **_looks like_** it is modifying the value. In practice, it (probably) does not have write access to the external value and is only masking it for the duration of this message. But the idea is that the value is mutable only if you are consciously mutating it.

Note that we could disallow `modify` to externals, that is, you have to either `let` or `modify` a local variable in order to annotate an external var.

exploration/variable-mutability.md
aphillips: The proposed design has the matching example in the next block, in which I call out the similarity error. But it isn't "the same flaw". In this case, the declaration completely blocks the original value.

This may not be super terrible. Not having separate namespaces is conceptually simpler and the ability to just annotate via declaration is much easier than doing multiple assignments and teaching developers/translators/etc. about local vs. external vars. Maybe `modify` (however we choose to spell it in the end) solves the "accidental overwrite" problem and is enough?

eemeli:
Here's an alternative proposal with immutable variables that I think could work. I think someone else may have previously proposed something like this, but I can't find that source right away.

Let us introduce a new keyword `input` that allows for the annotation of external variables. It works like this:
```
input {$count :number}
input {$date :datetime dateStyle=long}
match {$count}
when 1 {You received one message on {$date}}
when * {You received {$count} messages on {$date}}
```

This effectively replaces the current "hack" of saying `let $foo = {$foo ...}`, and provides a way to explicitly declare that a variable is non-local: `input {$bar}` doesn't require an annotation.

At the syntax/data model level, `input` or `let` declarations would not be required for all selectors and placeholders, but a user-configured validator could of course be stricter.

In the ABNF the change would look like this:
```abnf
message = [s] *(input-annotation [s]) *(declaration [s]) body [s]

input-annotation = input s "{" [s] operand [s annotation] [s] "}"
input = %x69.6E.70.75.74  ; "input"
```
The _expression_ rule can't be used here directly because the _operand_ is required.

With this in place, I would support prohibiting re-assignment/shadowing, as long as a local `let` declaration would take precedence and not cause an error if a similarly named external variable is passed to the formatter _without_ a corresponding `input` declaration in the message.

The use case of chaining operations on a variable with a single name is not supported here, and the `$foo1`, `$foo2` `$foo3` sorts of names would be required for that.

One observation I do have is that if our variables are immutable, then we should be able to reorder local variable declarations exactly like we allow for `when` variants to be reordered. We do still need to prohibit referential loops, but that doesn't require the declarations to be in any specific order.

aphillips:
> Here's an alternative proposal with immutable variables that I think could work. I think someone else may have previously proposed something like this, but I can't find that source right away.

I think this is an interesting proposal. I am not sure that I'm excited about it, although it does solve annotation of external values.

> 
> One observation I do have is that if our variables are immutable, then we should be able to reorder local variable declarations exactly like we allow for `when` variants to be reordered. We do still need to prohibit referential loops, but that doesn't require the declarations to be in any specific order.

This would bring back shadowing concerns, though, wouldn't it?

For inputs `a=1;b=2;c=3` what do these print:

```
let $c = {$b}
let $b = {$a}
{a={$a} b={$b} c={$c}}

let $b = {$a}
let $c = {$b}
{a={$a} b={$b} c={$c}}
```

exploration/variable-mutability.md
```diff
@@ -0,0 +1,288 @@
...
+```
+{ "arg1": "42" }
+...
+let #arg1 = {$arg1 :number maxFractionDigits=2}
+{Now I have to change {$arg} to {#arg}...}
+```
+
+### All Variables are Immutable; Externals are Annotatable; Shared Namespace
+
+_This is @eemelie's proposal from the PR thread_
```
eemeli, line 273:
```suggestion
_This is @eemeli's proposal from the PR thread_
```

stasm:
> > Here's an alternative proposal with immutable variables that I think could work. I think someone else may have previously proposed something like this, but I can't find that source right away.

We briefly discussed defining input variables in https://github.com/unicode-org/message-format-wg/issues/403#issue-1775350157. The, in https://github.com/unicode-org/message-format-wg/issues/310#issuecomment-1648115042 @aphillips suggested the `modify` keyword which would avoid the need of reassignment.

I quite like the idea of `input {$count :number}` because it cleverly address one of the most common existing use-cases for local variables (decorating input variables) _without_ introducing local variables.

Earlier in the thread, @eemeli commented on proposed syntax for local variables:

> These are all pretty ugly, but I presume that's intentional. Do we not want users to use local variables? Because that's what we're saying by making them ugly and clumsy.

Maybe we actually don't want local variables to be common, especially if there's a convenient and dedicated solution to annotate input variables (like the `input ...` syntax). What are the use-cases for local variables **that are not transformed input variables**? I can think of a few:

* Transforming input variables in a way that prevents access to the original value. Most formatters can be implemented in a "lazy" way that makes it possible to access the raw original value. However, I can imagine some formatters don't, although it can be argued that the following example is overly complex and contrived.

    > ```
    > let $firstUser = {$guests :list.take count=1}
    > let $tailCount = {$guests :list.length offset=1}
    >
    > {{$firstUser} and ${tailCount} other people like this.}
    > ```

* Factoring out literals with their formatting options to avoid repetition across multiple variants. A synthetic example: 

    > ```
    > let $epoch = {0 :datetime weekday=long}
    > match {$count}
    > when 1 {{$epoch} one.}
    > when * {{$epoch} many.}
    > ```

    I note, however, that perhaps we should consider how badly we want to avoid repetition in such cases. After all, top-level variants do cause repetition of translations anyways, by design. 

* Factoring out markup to avoid repetition across multiple variants. 

    > ```
    > let $openBold = {bold +html opt=val}
    > match {$count}
    > when 1 {{$openBold}strong{bold -html}}
    > when * {{$openBold}strong{bold -html}}
    > ```

    I think this is underspecified in the current design of open/close elements (#448), and I'm not sure if it should be allowed. Same as above, perhaps repetition isn't actually that bad? We're OK with repeating text content in all variants — why not markup?

eemeli:
> > One observation I do have is that if our variables are immutable, then we should be able to reorder local variable declarations exactly like we allow for `when` variants to be reordered. We do still need to prohibit referential loops, but that doesn't require the declarations to be in any specific order.
> 
> This would bring back shadowing concerns, though, wouldn't it?

Not really, because each `$var` would get looked up by a well defined method:
1. If there's an `input` statement for `$var`, find the corresponding `var` value in the inputs and apply the annotation to it.
2. Else, if there's a `let $var` statement, use its value.
3. Else, find the corresponding `var` value in the inputs.
4. Else, complain with Unresolved Variable.

So if the value gets determined in step 2, the corresponding input is ignored.

> For inputs `a=1;b=2;c=3` what do these print:
> 
> ```
> let $c = {$b}
> let $b = {$a}
> {a={$a} b={$b} c={$c}}
> 
> let $b = {$a}
> let $c = {$b}
> {a={$a} b={$b} c={$c}}
> ```

With logic as defined above, both of those would end up formatting as `a=1 b=1 c=1`.

aphillips:
@eemeli noted:

> Not really, because each $var would get looked up by a well defined method:

I think I'm still worried about the conceptual simplicity element of this. For example, would this be technically allowed?

```
input {$a :whatever}
let $a = {$b}
...
```

I put the reordered examples because people might make mistakes based on thinking of the lines as being executed in order.

@stasm noted:

> I quite like the idea of input {$count :number} because it cleverly address one of the most common existing use-cases for local variables (decorating input variables) without introducing local variables

I wasn't keen on `input` previously because the way it was originally suggested was to require it before using the variable (that is, variables had to be declared before use). This seems onerous when all I really want is `{Hello {$user}}`. Our syntax should be easy to use (at least in the common cases). But I agree with you that it helps give us an out for modifying values without allocating local variables.

Stepping back: does everyone agree with the matrix of options in the document? Are there other options we want to consider? Comparing the options and weighting our reasons behind those choices seems useful

exploration/variable-mutability.md
```diff
@@ -135,7 +135,108 @@ _What prior decisions and existing conditions limit the possible design?_
...
+
+The _expression_ rule can't be used directly in _input-declaration_ because the _variable_ is required.
+
+With this approach, variables are immutable,
+so each may be defined by only one _declaration_.
+The order of declarations does not matter,
+to allow for e.g. `input` annotations to refer to `local` variables.
+
+References to later declarations are not allowed,
+so this is considered an error:
```
aphillips, line 181: This confuses me. If the order doesn't matter then forward definitions should work, but then you say that they don't (so the order would matter). I think the first paragraph might be a holdover from when you had input followed by local (the new ABNF allows intermixing: intentional?)

exploration/variable-mutability.md
```diff
@@ -135,7 +135,108 @@ _What prior decisions and existing conditions limit the possible design?_
...
+
+```
+local $foo = {$bar :number}
+local $bar = {42 :number}
+{The answer is {$foo}}
+```
+
+An _input-declaration_ is not required for each external variable.
+A _local-declaration_ takes precedence and does not cause an error
+if a similarly named external variable is passed to the formatter
```
aphillips, line 191:
```suggestion
if an identically named external variable is passed to the formatter
```

similarly is too vague?

exploration/variable-mutability.md
eemeli: We ended up determining that forward-referencing should not be allowed, because if we do allow it, someone not aware of that fact looking at the example message could easily presume that the first `$bar` reference is to an external variable that's shadowed by the `local $bar` declaration.

exploration/variable-mutability.md
```diff
@@ -0,0 +1,370 @@
+# Design Proposal Template
+
+Status: **Proposed**
+
+<details>
+	<summary>Variable Namespacing and Mutability</summary>
+	<dl>
+		<dt>Contributors</dt>
+		<dd>@aphillips</dd>
```
eemeli, line 9:
```suggestion
		<dd>@aphillips</dd>
		<dd>@eemeli</dd>
```

exploration/variable-mutability.md
```diff
@@ -0,0 +1,370 @@
...
+		<dt>First proposed</dt>
+		<dd>2023-09-04</dd>
+		<dt>Pull Request</dt>
+		<dd>#000</dd>
```
eemeli, line 13:
```suggestion
		<dd><a href="https://github.com/unicode-org/message-format-wg/pull/469">#469</a></dd>
```

exploration/variable-mutability.md
```diff
@@ -0,0 +1,370 @@
+# Design Proposal Template
+
+Status: **Proposed**
```
eemeli, line 3:
```suggestion
Status: **Accepted**
```

eemeli (approved):
We're clearly already presuming this to be our syntax in our further work, so it'd be good to note this as "Accepted".

PR #470, opened 2023-09-05
Design document for Open/Close Expressions

eemeli:
This is submitted as a draft, as it currently only considers the use case (HTML-ish markup) that I care about, its requirements, and proposes a minimal design that fulfills my needs.

I invite everyone to contribute here, as discussed on yesterday's call. Especially when adding new items or paragraphs to the doc, please do directly commit to the [document](https://github.com/unicode-org/message-format-wg/blob/design-open-close/exploration/open-close-expressions.md) rather than supplying suggestions.

Having contributed my part, I don't consider myself as the owner here, so have added a couple of assignees instead.


exploration/0005-open-close-expressions.md
```diff
@@ -0,0 +1,152 @@
...
+- Representing markup, such as HTML, as expressions in a pattern
+  rather than as plain character sequences embedded in the _text_
+  of a pattern. This also allows parts of a markup sequence to be
+  programmable.
+
+**_Non-markup use cases to consider (which are not addressed by the design)_**
+
+- Represent templating language constructs so that they can pass through
+  the translation process and are not exposed to translation. For example,
+  Mustache templates.
```
eemeli, line 41: @aphillips I'm pretty sure that this is covered by the current "proposed design". Or is there some aspect of this that I'm missing? A concrete example would help.

```suggestion
- Represent templating language constructs so that they can pass through
  the translation process and are not exposed to translation.
  For example, Mustache templates.

**_Non-markup use cases to consider (which are not addressed by the design)_**
```

exploration/0005-open-close-expressions.md
```diff
@@ -0,0 +1,152 @@
...
+## Requirements
+
+Be able to indicate that some identified markup applies to
+a non-empty sequence of pattern parts (text, expressions).
+
+Markup spans may be nested,
+as in `<b>Bold and <i>also italic</i></b>`.
+
+Markup spans may be improperly nested,
+as in `<b>Bold <i>and</b> italic</i>`.
```
gibson042, line 56: Why is it a requirement to support this? I can see not caring at all, but rejecting designs outright that fail to accommodate improper nesting seems inappropriate.

exploration/0005-open-close-expressions.md
```diff
@@ -0,0 +1,152 @@
...
+Markup may have localisable options,
+such as `Click <a title="Link tooltip">here</a> to continue`.
+
+As with function options,
+markup options should be defined in a registry so that they can be validated.
+
+## Constraints
+
+Due to segmentation,
+markup may be split across messages so that it opens in one and closes in another.
```
gibson042, line 67: I don't think I understand this claim, can you elaborate? Messages are intended to be self-contained entities AFAIK.

exploration/0005-open-close-expressions.md
eemeli: That certainly works for me. Getting an example or two included in the doc would help explain what this is about.

exploration/0005-open-close-expressions.md
eemeli: That's actually a fair point. I'd be happy to drop this. Anyone disagree?

```suggestion
```

exploration/0005-open-close-expressions.md
eemeli: It is likely that in some cases a user will want to translate something like an entire HTML page via MF2, and rely on automated tooling to separate that into translatable segments, such as sentences, and consider each of them a "message". Such messages could indeed be self-contained from a localization PoV, but not necessarily from a markup PoV.

exploration/0005-open-close-expressions.md
aphillips: Your proposal focuses on HTML. I imagine that @mihnita might bring back such nesting for "attributed string" support. But I'd reserve that for when it is needed?

exploration/0005-open-close-expressions.md
aphillips: Meh. I don't think this type of segmentation is a good example. A better example might be something that captures some markup that is being used to generate page structure. For example:

```json
"msg1": "{{+section id=$sectionId}{+h2}Hello {$user}{-h2}}",
"msg2": "{{+p}Here is your data:{-p}}",
"msg3": "{+p}Footer{-p}{{-section}}"
```

This would be used to generate the top of a section, with the section close being in a separate resource:

```html
<section id="sectionId"><h2>Hello user</h2>
<p>Here is your data:</p>
<ul>
   <li>... various bits of data generated in code...
</ul>
<p>Footer</p>
</section>
```

The above could be generated in a for-loop over a list of users, for example.

gibson042: Thanks. That is plausible, although I do question whether or not it is _advisable_ because it hinders validation of well-formedness. But at least the motivation for the constraint now makes sense.

aphillips: It is not advisable, since the better course would be to leave the markup in the template. 

Generally, you'd only want inline markup in strings (`span`, `b`, `em`, `a`, etc.). It's harder to think of cases where you'd want to split those up **_and_** retain the markup in the message.

eemeli:
@aphillips @stasm Would you be ok with this landing as "Proposed"?

exploration/0470-open-close-expressions.md
```diff
@@ -0,0 +1,222 @@
...
+pattern = "{" *(text / placeholder) "}"
+placeholder = expression / markup
+
+markup       = "{" [s] markup-body [s] "}"
+markup-body  = (markup-standalone *(s option))
+             / (markup-open *(s option))
+             / markup-close
+markup-standalone = "#" name
+markup-open       = "+" name
+markup-close      = "-" name
```
eemeli, line 164: These sigils will need further consideration before we eventually accept this proposed design. As we evolve the syntax, `#` may get used for statements, and some other alternative open/close pair may end up being preferred.

stasm:
I’ve been thinking about an alternative design for this. I’m going to take note of it here briefly and add a new section in the doc later this week. I'm also fine with this PR landing as "Proposed" before I edit it -- I'll open a new PR then. (Actually, that may be preferable.)

Also, please be gentle, it’s a work-in-progress which I’m sharing early to see if there’s any support for it.

My goal is, as usual, to avoid adding new sigils to the syntax. Instead, I’d like to leverage the fact that we already have the concept of *keywords* in the syntax. Furthermore, with #469, we were able to find a good solution by adding a new keyword (`input`). I think we can try to do the same here.

I’d like to propose we add two new declaration types called `openclose` and `standalone`, along with new placeholder syntax: `{foo}`, `{/foo}`, and `{foo/}`. I've come to think that the only good syntax for spannables is one that looks like a markup language. There's also vast prior art in using some sort of `[foo]` and `[/foo]` in various web forums.  Note that this would require dropping unquoted literals as operands, which I think is an acceptable trade-off.

```
{{
    openclose {strong :html class=foo onclick=bar}
    standalone {img :html src=http://…}
    {{This is {strong}bold{/strong} and this is {img alt=|an image|/}.}}
}}
```

The declarations are optional (similar to input). If they’re absent, then any `{foo}` is interpreted as a span-open, any `{/foo}` as a span-close, and `{foo/}` as a standalone element. If declarations are present, they specify the kinds of markup precisely, making `{strong/}` and `{img}` and `{/img}` invalid in the example above. Additionally, they allow to add non-localizable attributes to makrup.

I’m not yet sure how to allow more than one element of the same name but with different attributes. Probably something like:

```
standalone img1 = {img :html src=http://a}
standalone img2 = {img :html src=http://b}
```

stasm:
I opened #513 which adds the above idea as an alternative to the design doc.

PR #471, opened 2023-09-06
(Design) Number Selection

eemeli:
See the [document](https://github.com/eemeli/message-format-wg/blob/design-number-selection/exploration/0006-number-selection.md) for details.

I started to draft a PR for the proposed design directly, but then admitted to myself that my PR description text started to look so much like a design doc that I should write it as one.

Ping @mihnita and @ryzokuken, whom I name-drop in the doc.


exploration/0006-number-selection.md
```diff
@@ -0,0 +1,127 @@
...
+Additional options such as `minimumFractionDigits` and others already supported by `:number`
+should also be supported.
+
+## Alternatives Considered
+
+An alternative approach to this problem could be to leave the `:number` `<matchSignature>` undefined,
+and to define three further functions, each with a `<matchSignature>`:
+
+- `:plural`
+- `:ordinal`
```
aphillips, line 88: Note that `ordinal` is both a selector and a formatter, e.g. 

| Value | Selector | Output |
|---|---|---|
| 1 | one | 1st |
| 2 | two | 2nd |
| 3 | few | 3rd |
| 4 | other | 4th |
|97 | other | 97th |
| 101 | one | 101st |

exploration/0006-number-selection.md
```diff
@@ -0,0 +1,127 @@
...
+- Encourage developers to use the same options for formatting and selection.
+
+## Constraints
+
+ICU MF1 messages using `plural` and `selectordinal` should be representable in MF2.
+
+## Proposed Design
+
+Given that we already have a `:number`,
+it makes sense to add a `<matchSignature>` to it with an option
```
aphillips, line 66: Actually... we don't yet have a `:number`. I kind of expected to find it in this doc! A complete definition would help figure out if this is a good design.

My first reaction to your proposal to use `:number` for both was that it was conceptually complicated (functions do multiple _different_ things depending on context), but I can see how it could be potentially both a simplifier and thus more robust for users who don't understand plurals. 

At the same time, I think that some common shorthands might be useful, just as MF1 has `{count,number,integer}`.

I also am concerned about the need to do multiple annotation (this has to do with the mutability question we're working), e.g.:

```
let $count = {$count :number groupingUsed=true 
               minimumFractionDigits=0 minimumDecimalDigits=4 
               maximumFractionDigits=2} <- formatting
match {$count :number type=plural} <- get the right selector
when ...
```

I recognize that my example uses the default selector `plural`, and so wouldn't need `type=`, but how do I get the `exact` selector if I want it without hosing all of that formatting?

aphillips (changes requested):
Thank you for starting on this. Much appreciated!

exploration/0006-number-selection.md
```diff
@@ -0,0 +1,127 @@
...
+## Background
+
+As discussed on numerous past occasions and explicitly identified in
+<a href="https://github.com/unicode-org/message-format-wg/pull/457">#457</a>,
+it would be good for both plural and non-plural match selection to be supported in MF2.
+
+MF1 supported selection on either cardinal plurals or ordinal numbers,
+via the `plural` and `selectordinal` selectors.
+It also customized this selection beyond the capabilities of `com.ibm.icu.text.PluralRules`
+by allowing for an `offset` parameter.
```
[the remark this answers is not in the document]
eemeli:
```suggestion
by allowing for explicit value matching and an `offset` parameter.
```

exploration/0006-number-selection.md
```diff
@@ -0,0 +1,127 @@
...
+As pointed out by <a href="https://github.com/mihnita">@mihnita</a> in particular,
+category selection is not always appropriate for selection on a number:
+the number may be representing some completely other quantity,
+such as a four-digit year or the integer value of an enumerator.
+
+Furthermore, as pointed out by <a href="https://github.com/ryzokuken">@ryzokuken</a>
+in <a href="https://github.com/unicode-org/message-format-wg/pull/457#discussion_r1307443288">#457 (comment)</a>,
+ordinal values are not truly "plural",
+even if they use categories with similar names and are determined by similar rules.
+
```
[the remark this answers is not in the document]
eemeli:
```suggestion

Additionally,
MF1 provides `ChoiceFormat` selection based on whether a number is within a range,
and both JS and ICU PluralRules implementations provide for determining the plural category
of a range based on its start and end values.
These range-based selectors are not initially considered here.

```

exploration/0006-number-selection.md
```diff
@@ -0,0 +1,127 @@
...
+## Use-Cases
+
+Put together, we have thus identified three different types of selection
+we would like to support on numeric values:
+
+- cardinal plural selection
+- ordinal selection
+- exact match selection
```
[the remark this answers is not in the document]
eemeli:
```suggestion
- exact value match selection
```

exploration/0006-number-selection.md
eemeli: > Actually... we don't yet have a `:number`. I kind of expected to find it in this doc! A complete definition would help figure out if this is a good design.

We have its definition in [`spec/registry.xml`](https://github.com/unicode-org/message-format-wg/blob/main/spec/registry.xml), which includes a `<formatSignature>` for it. Follow-up work from #457 should probably add a document describing that and `:datetime` to start with.

> At the same time, I think that some common shorthands might be useful, just as MF1 has `{count,number,integer}`.

Possibly yes, but this sounds like a separate follow-on spec PR or design doc. :)

> I also am concerned about the need to do multiple annotation (this has to do with the mutability question we're working), e.g.:
> 
> ```
> let $count = {$count :number groupingUsed=true 
>                minimumFractionDigits=0 minimumDecimalDigits=4 
>                maximumFractionDigits=2} <- formatting
> match {$count :number type=plural} <- get the right selector
> when ...
> ```
> 
> I recognize that my example uses the default selector `plural`, and so wouldn't need `type=`, but how do I get the `exact` selector if I want it without hosing all of that formatting?

First of all, the `:number` in your selector would be getting the resolved rather than formatted value of the first expression, so it could extend rather than clobber the set of options defined in the variable declaration. So just using `type=exact` where you have `type=plural` should do the trick.

Alternatively, you should be able to include it in the first bag of options:
```
let $count = {$count :number type=exact groupingUsed=true 
               minimumFractionDigits=0 minimumDecimalDigits=4 
               maximumFractionDigits=2}
match {$count}
when ...
```

If/as we'll want to keep the formatting and selection options compatible, you should be able to assign a bag of options covering both to a variable.

exploration/0006-number-selection.md
```diff
@@ -0,0 +1,127 @@
...
+Given that we already have a `:number`,
+it makes sense to add a `<matchSignature>` to it with an option
+
+```xml
+<option name="type" values="plural ordinal exact" default="plural" />
+```
+
+The default `plural` value is presuming that to be the most common use case,
+and it affords the least bad fallback when used incorrectly:
+Using "plural" for "exact" still selects exactly matching cases,
```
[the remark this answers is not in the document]
eemeli: I think they may end up subtly different, in that `:number type=exact` would be doing comparison based on numerical values, whereas `:string` or `:equals` would need to cast its input to a string.

So for example with:
```
let $x = {1 :number minimumFractionDigits=1}
match {$x :number type=exact}
when 1 {one}
when * {not one}
```
we'd clearly end up with "one", whereas with
```
let $x = {1 :number minimumFractionDigits=1}
match {$x :string}
when 1 {one}
when 1.0 {one dot zero}
when |1,0| {one comma zero}
when * {not one}
```
my guess would be that either "one dot zero" or "one comma zero" gets selected, _depending on the locale_.

exploration/0006-number-selection.md
```diff
@@ -0,0 +1,127 @@
...
+whereas using "exact" for "plural" will not select LDML category matches.
+This might not be noticeable in the source language,
+but may cause problems in target languages that the original developer is not considering.
+
+Additional options such as `minimumFractionDigits` and others already supported by `:number`
+should also be supported.
```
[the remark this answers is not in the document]
eemeli: I think we _have_ specified an initial set of formatting options for `:number`, and so this is referring to them directly.

exploration/0006-number-selection.md
eemeli: If an implementation does support `:ordinal` as a formatter, then it should probably also support it as a selector. But I'm not convinced that we'll want to include that in the required set.

exploration/0006-number-selection.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Alternatives Considered
+
+An alternative approach to this problem could be to leave the `:number` `<matchSignature>` undefined,
+and to define three further functions, each with a `<matchSignature>`:
+
+- `:plural`
+- `:ordinal`
+- `:exact`
```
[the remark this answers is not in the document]
eemeli: This too sounds like a possible implementation or user extension.

exploration/0006-number-selection.md
```diff
@@ -0,0 +1,127 @@
...
+
+```
+let $count = {$count :number minimumFractionDigits=1}
+match {$count}
+when 0 {You have no apples}
+when 1 {You have exactly one apple}
+when * {You have {$count} apples}
+```
+
+The proposed design is more robust than this alternative,
```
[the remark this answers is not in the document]
eemeli: So where in this document should this discussion be placed if not here?

exploration/0006-number-selection.md
```diff
@@ -0,0 +1,127 @@
...
+match {$count :plural minimumFractionDigits=1}
+when 0 {You have no apples}
+when 1 {You have exactly one apple}
+when * {You have {$count :number minimumFractionDigits=1} apples}
+```
+
+Here, because selection on `:number` is not allowed,
+it's easy to duplicate the options because _some_ annotation is required on the selector.
+It would also be relatively easy to leave out the `minimumFractionDigits=1` option from the selector,
+as it's not required for the English source.
```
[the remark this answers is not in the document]
aphillips: It does and that is good from the point of view that "one apple" is incorrectly internationalized with keyword `one`. You'll notice that elsewhere I did both 😉 

However, we should be careful to model good internationalization in our docs because others will copy us. In my example above I put the formatted count in the "one" example for that reason.

exploration/0006-number-selection.md
aphillips: Ordinal would be needed for MF1 compatibility, if by MF1 compatibility we mean "ICU". But the callout is, I think, important, as you only mentioned ordinal as a selector.

aphillips:
I think this might need a revision in light of our Slack conversation yesterday. Having number shorthands for `integer` and such present in the proposal seems like a good way to work through it.

I like the idea of using `:number` as a model for other additions to the default registry, which we really really need to do next 😉

eemeli:
I refreshed the design doc a bunch, please take another look? This now relies somewhat on #532.

exploration/number-selection.md
```diff
@@ -0,0 +1,165 @@
...
+also add the following `<alias>` definitions to `<function name="number">`:
+
+```xml
+<alias name="plural" supports="match">
+	<setOption name="select" value="plural"/>
+</alias>
+<alias name="ordinal" supports="match">
+	<setOption name="select" value="ordinal"/>
+</alias>
+```
```
aphillips, line 98: Is this list exhaustive? #532 includes `:integer`. Presumably `:integer` is an integer plural selector by default (in addition to being a formatter).

Other candidates for shorthands might be `:percent`, `:currency` and (maybe??) `:scientific`. The first two are shorthands in ICU4J MF1. `scientific` is a numeric formatting option (but not exposed by MF1). 

There is also the potentially lamentable `:spellout` from ICU4J MF1.

I'm guessing that what we want to promote is single-annotation messages, e.g.:

```
.input {$var :spellout}
.match {$var}                               // it's already a plural
[0]   {{You have no more bone dragons}}
[one] {{You have {$var} more bone dragon}}  // "You have one more..."
[*]   {{You have {$var} more bone dragons}}
```

aphillips (commented):
A few comments, but generally looks good.

exploration/number-selection.md
eemeli: I think the ones you mention are all formatters? The list here is purely about `.match` selectors, and for that I think just `:plural` and `:ordinal` could be enough. If you need exact matching, then the relatively rare `:number select=exact` is pretty good?

eemeli: To clarify a bit, a proposed alias like `:integer` or `:spellout` still ends up calling the `:number` function, and so if the expression doesn't have an explicit `select` attribute and the alias does not set `supports="format"`, it'll end up inheriting the default `select="plural"` for selection, together with the rest of the expression & alias options.

eemeli: Here's how I would classify the different aliases that have been mentioned so far:
- `:integer` -- The most obvious one, and therefore included already in [#532](https://github.com/unicode-org/message-format-wg/pull/532).
- `:plural` and `:ordinal` -- Functions for selection only with `supports="match"`. Included in this design doc.
- `:percent`, `:currency`,`:scientific`, `:spellout` -- Possible aliases for later/separate consideration. As with `:integer`, they might be used also for selection, but I don't see how any of them would specify a `<setOption name="select">`. So I don't think they belong in a design doc on Number Selection.

Once #532 lands, would a PR adding that third set be the right focal point for discussion of their merits, rather than this PR?

exploration/number-selection.md
```diff
@@ -0,0 +1,179 @@
...
+
+## Objective
+
+Define how selection on numbers happens.
+
+## Background
+
+As discussed on numerous past occasions and explicitly identified in
+<a href="https://github.com/unicode-org/message-format-wg/pull/457">#457</a>,
+it would be good for both plural and non-plural match selection to be supported in MF2.
```
aphillips, line 25: This is slightly misleading--it implies a binary plural/non-plural set of selection, whereas we go on to point out that there are a variety of number selection needs. Perhaps:

```suggestion
As discussed by the working group and explicitly identified in
<a href="https://github.com/unicode-org/message-format-wg/pull/457">discussion of #457</a>,
there is a need to support multiple types of selection on numeric values in MF2.
```

exploration/number-selection.md
```diff
@@ -0,0 +1,179 @@
...
+by allowing for explicit value matching and an `offset` parameter.
+
+As pointed out by <a href="https://github.com/mihnita">@mihnita</a> in particular,
+category selection is not always appropriate for selection on a number:
+the number may be representing some completely other quantity,
+such as a four-digit year or the integer value of an enumerator.
+
+Furthermore, as pointed out by <a href="https://github.com/ryzokuken">@ryzokuken</a>
+in <a href="https://github.com/unicode-org/message-format-wg/pull/457#discussion_r1307443288">#457 (comment)</a>,
+ordinal selection is similar to plural selection, but not identical.
```
aphillips, line 39:
```suggestion
ordinal selection works similarly to plural selection, but uses a separate set of rules
for each locale.
This is visible in English, for example, where plural rules use only `one` and `other`
but ordinal rules use `one` (_1st_, _21st_, etc.), `few` (_2nd_, _22nd_, etc.), 
`many` (_3rd_, _23rd_, etc.), and `other` (_4th_, _5th_, etc.).
```

exploration/number-selection.md
```diff
@@ -0,0 +1,179 @@
...
+
+Additionally,
+MF1 provides `ChoiceFormat` selection based on whether a number is within a range,
+and both JS and ICU PluralRules implementations provide for determining the plural category
+of a range based on its start and end values.
+These range-based selectors are not initially considered here.
```
aphillips, line 45: These are separate things; they should be kept separate.

```suggestion
Additionally,
MF1 provides `ChoiceFormat` selection based on a complex rule set
(and which allows determining if a number falls into a specific range).

Both JS and ICU PluralRules implementations provide for determining the plural category
of a range based on its start and end values.
Range-based selectors are not initially considered here.
```

exploration/number-selection.md
eemeli: I agree that `:ordinal` could also be a formatter, but I am not here proposing that it be one of the core formatters that we expect all MF2 implementations to provide. That could be considered as a separate further step, should this design on number _selection_ first introduce it as a selector.

exploration/number-selection.md
```diff
@@ -0,0 +1,179 @@
...
+as it's not required for the English source.
+
+With the proposed design, this message would much more naturally be written as:
+
+```
+.input {$count :number minimumFractionDigits=1}
+.match {$count}
+0 {{You have no apples}}
+1 {{You have exactly one apple}}
+* {{You have {$count} apples}}
```
aphillips, line 155: As noted before we should model good behavior. This message should also have a `one` rule

```suggestion
one {{You have {$count} apple}}
* {{You have {$count} apples}}
```

aphillips (commented):
A bunch of housekeeping suggestions in this round.

Note the dependency on #433 that we need to resolve.

exploration/number-selection.md
```diff
@@ -0,0 +1,179 @@
...
+## Alternatives Considered
+
+### Completely Separate Functions
+
+An alternative approach to this problem could be to leave the `:number` `<matchSignature>` undefined,
+and to define three further functions, each with a `<matchSignature>`:
+
+- `:plural`
+- `:ordinal`
+- `:exact`
```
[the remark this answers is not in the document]
eemeli:
```suggestion
- `:exact` (actual name TBD, pending the resolution of [#433](https://github.com/unicode-org/message-format-wg/issues/433)
```

exploration/number-selection.md
eemeli: I would very strongly prefer taking this smaller step first, and not expanding the scope of this design document.

exploration/number-selection.md
```diff
@@ -0,0 +1,200 @@
...
+- Encourage developers to provide the formatting options used in patterns to the selector
+  so that proper selection can be done.
+
+## Constraints
+
+ICU MF1 messages using `plural` and `selectordinal` should be representable in MF2.
+
+## Proposed Design
+
+Given that we already have a `:number`,
```
mihnita, line 95: > Given that we already have a `:number`,
> it makes sense to add a `<matchSignature>` to it with an option


Sorry, but it is not that obvious.
`:number` is not a question, so one can't decide on it.

You have to ask a question about the number.

"Hey, `:number 43` , true or false?
If you answer correctly you win 1000$"
That means nothing. It is not a good question.

`is :number 43 even`" is a question, you can answer.
`is :number 43 prime`" is a question, you can answer.
`is :number 43 plural`" is a question, you can answer.

macchiati:
One note: we might not want to treat :integer as simply an alias for :number. There is one difference in handling:

When you have a match, for translation the **when** clauses are expanded (or contracted) depending on the locale. There are some locales that have a plural category that is only present when the number is fractional, and in those locales you can limit the expansion by not including that category if the numeric value is known to be integral.

exploration/number-selection.md
aphillips: There are two questions here.

The first is: should the `:number` function used for formatting also do selection? @eemeli has already pointed out that we want users to use the same configuration for selection (e.g. with `:plural`) as will ultimately be used to format the number. This message produces grammatically inferior results:

```
.local $tenths = {$num :number minFractionDigits=1}
.match {$num :plural}
one {{You have {$tenths} value}}  // won't be seen
* {{You have {$tenths} values}}
```

The second question is: if `:number` is a selector, what sort of selector is it by default? @eemeli answers that `:plural` is the most common usage. Alternatives would be `:ordinal` and whatever we call exact match. I agree with @eemeli that `:plural` is the far-and-away most common case and thus should be the default.

So... @mihnita are you objecting to the phrasing here or the conclusion? If the phrasing, do you want to suggest alternate text?

aphillips:
@macchiati `:integer` is not "just an alias for `:number`". It includes the settings that force numbers to be formatted as integers (and thus guides selection also). What this design doc says, in effect, is that `:integer` as a function/selector inherits all of its setup from the `:number` entry in the registry except for it's overrides. That saves us from having to recreate every detail of `:number` in a separate function description.

macchiati:
Ah, my mistake; I thought I had read that :integer was just an alias; glad that it isn't. And having it inherit from :number except when overridden for given locales makes sense.

exploration/number-selection.md
eemeli: @aphillips Your "won't be seen" comment is a bit misleading, as the problem with that message is that the `one` variant _may_ indeed be seen, formatted as "You have 1.0 value".

aphillips: You're right.. that comment goes against what I'm trying to say, so I fixed it.

mihnita: Since we already defined `$tenths` and use it in the body of the options, then I expect `:plural` on it:
```
.local $tenths = {$num :number minFractionDigits=1}
.match {$tenths :plural}
one {{You have {$tenths} value}}  // not desirable
* {{You have {$tenths} values}}
```

And not doing that is easy to do at lint time.

PR #472, opened 2023-09-11
Add stability policy

eemeli:
As discussed face-to-face yesterday.

The first two paragraphs state that future changes won't break your old MF2 messages, and that future messages will not use syntax that an older parser will consider as errors -- this is why we've _reserved_ some parts of the spec.

The last paragraph is about soft-reserving namespace for future spec versions.

### DIFF
--- a/spec/README.md
+++ b/spec/README.md
@@ -76,6 +76,12 @@ of valid messages written for earlier versions of this specification
 that only use functions and expression attributes defined in this specification.
 Later specification versions MAY make previously invalid messages valid.
 
+> [!NOTE]
+> This does not guarantee that the results of formatting will never change.
+> Even when the specification doesn't change,
+> the functions for date formatting, number formatting and so on
+> will change their results over time.
+
 Updates to this specification MUST NOT introduce message syntax that,
 when parsed according to earlier versions of this specification,
 would produce syntax or data model errors.
--- a/spec/README.md
+++ b/spec/README.md
@@ -82,7 +82,12 @@ would produce syntax or data model errors.
 Such messages MAY produce errors when formatted
 according to an earlier version of this specification.
 
-Later versions of this specification MAY define new function names,
+From version 2.0, MessageFormat will only reserve, define, or require
+function names,
+function option names,
+or expression attribute names 
+consisting of characters in the ranges a-z, A-Z, and 0-9.
+All other names in these categories are reserved for the use of implementations or users.
 function option names, and expression attribute names with new meanings.
 Such names defined in this specification MUST consist only of
 characters in the ranges a-z, A-Z, and 0-9.
--- a/spec/README.md
+++ b/spec/README.md
@@ -71,7 +71,7 @@ A reference to a _term_ looks like this.
 ### Stability Policy
 
 Updates to this specification MUST NOT change
-the syntactical meaning, the formatting, or other behaviour
+the syntactical meaning, the runtime output, or other behaviour
 of valid messages written for earlier versions of this specification
 that only use functions and expression attributes defined in this specification.
 Later specification versions MAY make previously invalid messages valid.
--- a/spec/README.md
+++ b/spec/README.md
@@ -79,7 +79,7 @@ in this version. Future versions MAY add additional structure or meaning
 to existing syntax.
 
 Updates to this specification will not remove any reserved keywords
-or sigils. 
+or sigils.
 
 > Note: future versions may defined new keywords
 
--- a/spec/README.md
+++ b/spec/README.md
@@ -74,7 +74,22 @@ Updates to this specification MUST NOT change
 the syntactical meaning, the runtime output, or other behaviour
 of valid messages written for earlier versions of this specification
 that only use functions and expression attributes defined in this specification.
-Later specification versions MAY make previously invalid messages valid.
+Updates to this specification will not remove any syntax provided
+in this version. Future versions MAY add additional structure or meaning
+to existing syntax.
+
+Updates to this specification will not remove any reserved keywords
+or sigils. 
+
+> Note: future versions may defined new keywords
+
+Updates to this specification will not reserve or assign meaning to any
+character "sigils" except for those in the `reserved` production.
+
+Updates to this specification will not remove any functions
+defined in the default registry nor will they remove any options
+or option values. Additional options or option values MAY be
+defined.
 
 > [!NOTE]
 > This does not guarantee that the results of formatting will never change.
--- a/spec/README.md
+++ b/spec/README.md
@@ -83,14 +83,9 @@ Such messages MAY produce errors when formatted
 according to an earlier version of this specification.
 
 From version 2.0, MessageFormat will only reserve, define, or require
-function names,
-function option names,
-or expression attribute names
+function names or function option names
 consisting of characters in the ranges a-z, A-Z, and 0-9.
 All other names in these categories are reserved for the use of implementations or users.
-function option names, and expression attribute names with new meanings.
-Such names defined in this specification MUST consist only of
-characters in the ranges a-z, A-Z, and 0-9.
 
 > [!NOTE]
 > Users defining custom names SHOULD include at least one character outside these ranges
--- a/spec/README.md
+++ b/spec/README.md
@@ -85,7 +85,7 @@ according to an earlier version of this specification.
 From version 2.0, MessageFormat will only reserve, define, or require
 function names,
 function option names,
-or expression attribute names 
+or expression attribute names
 consisting of characters in the ranges a-z, A-Z, and 0-9.
 All other names in these categories are reserved for the use of implementations or users.
 function option names, and expression attribute names with new meanings.
--- a/spec/README.md
+++ b/spec/README.md
@@ -85,5 +85,7 @@ Later versions of this specification MAY define new function names,
 function option names, and expression attribute names with new meanings.
 Such names defined in this specification MUST consist only of
 characters in the ranges a-z, A-Z, and 0-9.
-Users defining custom names SHOULD include at least one character outside these ranges
-to ensure that they will be compatible with future versions of this specification.
+
+> [!NOTE]
+> Users defining custom names SHOULD include at least one character outside these ranges
+> to ensure that they will be compatible with future versions of this specification.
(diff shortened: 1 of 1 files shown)


spec/README.md
```diff
@@ -66,3 +67,23 @@ A **_term_** looks like this when it is defined in this specification.
...
+when parsed according to earlier versions of this specification,
+would produce syntax or data model errors.
+Such messages MAY produce errors when formatted
+according to an earlier version of this specification.
+
+Later versions of this specification MAY define new function names,
+function option names, and expression attribute names with new meanings.
+Such names defined in this specification MUST consist only of
+characters in the ranges a-z, A-Z, and 0-9.
+Users defining custom names SHOULD include at least one character outside these ranges
```
alerque, line 88: Why not specify/suggest that custom names (i.e. private names) be prefixed with '`_`' or some other traditional indicator of intent? The "some character outside of [a-zA-Z0-9] somewhere in the name" seems a bit squishy if technically verifiable.

eemeli: I would be open to including a specific suggestion like that, but it probably ought to allow for namespacing like `moz-platform` or `com.google.number`. If we ever define some more structure within the _name_ rule then that ought to be obviously followed.

The rule should not be hard, though, as that would prevent e.g. customizing existing formatters with custom options.

spec/README.md
```diff
@@ -66,3 +67,23 @@ A **_term_** looks like this when it is defined in this specification.
 A reference to a _term_ looks like this.
 
 > Examples are non-normative and styled like this.
+
+### Stability Policy
+
```
aphillips, line 72: I would probably put syntax before "formatting or other behavior", since the syntax has less implementation specificity and since it forms the basis for everything that comes after.

spec/README.md
```diff
@@ -66,3 +67,23 @@ A **_term_** looks like this when it is defined in this specification.
...
+that only use functions and expression attributes defined in this specification.
+Later specification versions MAY make previously invalid messages valid.
+
+Updates to this specification MUST NOT introduce message syntax that,
+when parsed according to earlier versions of this specification,
+would produce syntax or data model errors.
+Such messages MAY produce errors when formatted
+according to an earlier version of this specification.
+
+Later versions of this specification MAY define new function names,
```
aphillips, line 84: I think we want to start separating the specification and registry now, since they are likely to have different life-cycles. Perhaps s/specification/registry/ along the lines of:

```suggestion
Unicode will only register or consider for registration function names,
function option names,
or expression attribute names 
consisting of characters in the ranges a-z, A-Z, and 0-9.
All other names are reserved for the use of implementations or users.
```

... and note that this policy begins now.

spec/README.md
aphillips: This guidance should be a note, not part of the stability guarantee directly... or be in the specification body or user guide. That is, this probably should be mentioned in [syntax#function](https://github.com/unicode-org/message-format-wg/blob/main/spec/syntax.md#function) etc.

aphillips (changes requested):
Good stuff. Some minor quibbles.

spec/README.md
eemeli: The intent with the current order is to first define backward compatibility, then forward compatibility. As the first one extends to formatting, its extension to covering also parsing is implicit.

spec/README.md
eemeli: I would prefer to keep the reference to "specification", as it currently includes the "registry". As we've identified in the discussions on markup and expression attributes as well as the registry itself, we'll need to have space somewhere describing the behaviour associated with some of the specified names.

If we do end up with a spec that does not by name refer to _any_ of these but relegates that to a separate document and we have a settled name for that document such as "schema", then this could be changed.

spec/README.md
```diff
@@ -66,3 +67,23 @@ A **_term_** looks like this when it is defined in this specification.
...
+function option names, and expression attribute names with new meanings.
+Such names defined in this specification MUST consist only of
+characters in the ranges a-z, A-Z, and 0-9.
+Users defining custom names SHOULD include at least one character outside these ranges
+to ensure that they will be compatible with future versions of this specification.
```
eemeli, line 89:
```suggestion

> [!NOTE]
> Users defining custom names SHOULD include at least one character outside these ranges
> to ensure that they will be compatible with future versions of this specification.
```

spec/README.md
eemeli: Put the last sentence into a note, but did not at least yet add a specific suggestion.

spec/README.md
aphillips: Okay. We need to be careful with "implicit" though. Stability policies are generally exercises in explicit promises 😉

spec/README.md
aphillips: Notice that my suggested text doesn't mention the spec or the registry. Maybe give it a second look? Also, it gives a durable promise of what we'll consider for names.

eemeli: It does, however, use the terms "register" and "registration", which might not be what we'll want to use if we end up calling the document defining functions and other things something else than "registry".

Also, looking at it again I notice that it's binding "Unicode" and not "this specification". I think that's too large a scope, as it could be seen as a restriction on what ICU can do, or what other Unicode specifications can do.

spec/README.md
eemeli: Added "syntactical meaning" to the first paragraph to make it explicit.

spec/README.md
aphillips: By using Unicode and not "this specification" I can encompass all aspects of prescribed naming, regardless of form: we are promising not to use different terms in MF2. I agree that we need to narrow it to MF2 and not prohibit ICU from making an implementation choice (especially if that choice were to create a local-to-ICU definition consistent with the beat practice found later in this section).

Perhaps then:

```suggestion
From version 2.0, MessageFormat will only reserve, define, or register 
function names,
function option names,
or expression attribute names 
consisting of characters in the ranges a-z, A-Z, and 0-9.
All other names in these categories are reserved for the use of implementations or users.
```

alerque (approved):

spec/README.md
```diff
@@ -66,3 +67,31 @@ A **_term_** looks like this when it is defined in this specification.
 A reference to a _term_ looks like this.
 
 > Examples are non-normative and styled like this.
+
+### Stability Policy
+
+Updates to this specification MUST NOT change
+the syntactical meaning, the formatting, or other behaviour
```
stasm, line 74: When I first read this, for some reason I thought "formatting referred to the layout of the code, not the runtime behavior. Perhaps be specific here?

```suggestion
the syntactical meaning, the runtime output, or other behaviour
```

spec/README.md
```diff
@@ -66,3 +67,31 @@ A **_term_** looks like this when it is defined in this specification.
...
+Updates to this specification MUST NOT introduce message syntax that,
+when parsed according to earlier versions of this specification,
+would produce syntax or data model errors.
+Such messages MAY produce errors when formatted
+according to an earlier version of this specification.
+
+From version 2.0, MessageFormat will only reserve, define, or require
+function names,
+function option names,
+or expression attribute names
```
stasm, line 88: We don't have expression attributes yet.

```suggestion
```

spec/README.md
```diff
@@ -66,3 +67,31 @@ A **_term_** looks like this when it is defined in this specification.
...
+consisting of characters in the ranges a-z, A-Z, and 0-9.
+All other names in these categories are reserved for the use of implementations or users.
+function option names, and expression attribute names with new meanings.
+Such names defined in this specification MUST consist only of
+characters in the ranges a-z, A-Z, and 0-9.
```
stasm, line 93:
```suggestion
```

stasm (changes requested):

spec/README.md
```diff
@@ -66,3 +67,31 @@ A **_term_** looks like this when it is defined in this specification.
...
+
+Updates to this specification MUST NOT introduce message syntax that,
+when parsed according to earlier versions of this specification,
+would produce syntax or data model errors.
+Such messages MAY produce errors when formatted
+according to an earlier version of this specification.
+
+From version 2.0, MessageFormat will only reserve, define, or require
+function names,
+function option names,
```
eemeli, line 87:
```suggestion
function names or function option names
```

stasm (approved):

macchiati:
> +the syntactical meaning, the formatting, or other behaviour

People may read this wrong. If we add a comma between "Wednesday" and " January 12" in Australian English in CLDR, and someone uses the stock Unicode date formatting algorithm with CLDR, that will cause the end result of formatting to not be character for character identical. So I think it is worth adding a note, something like:

This does not guarantee that the results of formatting will never change. Even when the specification doesn't change, the functions for date formatting, number formatting and so on will change their results over time, as new information about customary formats for locales surfaces — and sometimes the customary format in a locale itself will change over time: for example, customary Swedish sort order behavior changed over time, and unit preferences for particular usages can change over time.

spec/README.md
```diff
@@ -76,6 +76,12 @@ of valid messages written for earlier versions of this specification
 that only use functions and expression attributes defined in this specification.
 Later specification versions MAY make previously invalid messages valid.
```
aphillips, line 77: > Later specification versions MAY make previously invalid messages valid.

This should be a separate paragraph as it is a separate guarantee.

>Updates to this specification MUST NOT change
>the syntactical meaning, the runtime output, or other behaviour
>of valid messages written for earlier versions of this specification

To Mark's point, I think this isn't right yet. Not changing the runtime output is a *very* difficult guarantee and going beyond that to behavior? We need to be more specific about what we're exactly stabilizing.

Perhaps:

```suggestion
Updates to this specification will not remove any syntax provided
in this version. Future versions MAY add additional structure or meaning
to existing syntax.

Updates to this specification will not remove any reserved keywords
or sigils. 

> Note: future versions may defined new keywords (??)

Updates to this specification will not reserve or assign meaning to any
character "sigils" except for those in the `reserved` production.

Updates to this specification will not remove any functions
defined in the default registry nor will they remove any options
or option values. Additional options or option values MAY be
defined.
```

aphillips (changes requested):
I think more work is needed to address Mark's comment. My suggestions are probably incomplete and perhaps not thoroughly thought through. We need to review as a WG.

aphillips:
Merging per 2023-09-18 call. Note this is a work in progress.

PR #473, opened 2023-09-12
Expand expression attributes proposal

eemeli:
Adds examples for use cases, relaxes namespacing to a suggestion, and documents function chaining as an alternative.

Full doc: https://github.com/unicode-org/message-format-wg/blob/exprattr/exploration/expression-attributes.md


exploration/0002-expression-attributes.md
```diff
@@ -38,18 +38,59 @@ At least the following expression attributes should be considered:
...
+    > {Hello {$place @fallback=world}}
+    > ```
+
   - `locale` — An override for the locale used to format the expression.
     Should be expressed as a non-empty sequence of BCP 47 language codes.
+
+    > Example embedding a French literal in an English message:
+    >
+    > ```
+    > {In French, "{|bonjour| @locale=fr}" is a greeting}
```
stasm, line 54: Is this use-case also solved by spans?

```
{In French, {+span locale=fr}bonjour{-span} is a greeting}
```

`+span`/`-span` doesn't even have to mean HTML `span` here. It's just a start and an end of a span of text that should have a different locale property. The runtime can then decide how to express it such that the locale is isolated properly.

exploration/0002-expression-attributes.md
```diff
@@ -38,18 +38,59 @@ At least the following expression attributes should be considered:
...
+    > ```
+
   - `dir` — An override for the LTR/RTL/auto directionality of the expression.
 
+    > Example explicitly isolating the directionality of a placeholder:
+    >
+    > ```
+    > {Welcome, {$username @dir=auto}}
```
stasm, line 62: Same as above, why not encourage solving this via spans?

```
{Welcome, {+span dir=auto}{$username}{-span}.}
```

exploration/0002-expression-attributes.md
```diff
@@ -38,18 +38,59 @@ At least the following expression attributes should be considered:
...
 - Attributes relevant for translators, tools, and other message operations,
   but with no runtime impact:
 
   - `example` — A literal value representing
     what the expression's formatted value will look like.
+
+    > Example providing a translator with an example for `$place`:
+    >
+    > ```
+    > {Hello {$place @example=world}}
```
stasm, line 74: This is treading the semantic comments territory. Perhaps it would be best to leave this out of scope?

stasm (commented):
I'm not sure about the runtime uses of attributes. I primarily see them as aid for tooling. For the runtime, I think we can leverage the open/close syntax to create spans decorated with some extra data, as needed.

> Attributes can be assigned to any expression, including expressions without an annotation.

I'm not sure if attributes should be allowed on all expressions, or perhaps just placeholders? For example, if the `@locale` attribute impacts the runtime, would the following example produce `1.0 apple`? 

```
let $myCount = {$count :number minFracDigits=2}
match {$myCount @locale=fr}
when 1 {{$myCount} apple.}
when * {{$myCount} apples.}
```

exploration/0002-expression-attributes.md
```diff
@@ -38,18 +38,70 @@ At least the following expression attributes should be considered:
 
   - `fallback` — A value to use instead of the default fallback,
     should the expression's primary formatting fail in some way.
+
+    > Example that would format as "Hello world" if `$place` is not defined:
+    >
+    > ```
+    > {Hello {$place @fallback=world}}
```
aphillips, line 45: This example is evil? It's basically a string concatenation that requires interior localization?

exploration/0002-expression-attributes.md
```diff
@@ -38,18 +38,70 @@ At least the following expression attributes should be considered:
...
+    > ```
+
+  - `id` — An identifier for the expression.
+    This is included in the formatted part,
+    and allows the parts of an expression to be explicitly addressed.
+
+    > Example identifying two formatted numbers
+    > in a message with badly named input variables:
```
aphillips, line 53: Perhaps use better names for the input variables? Or are you thinking of MF1 ported values with auto-numbered replacements?

I think this use case is more compelling when the variable names and ID names align differently, such as when the formatted value needs to be placed withing a DOM element with a given ID but the input variable names cannot be changed.

exploration/0002-expression-attributes.md
aphillips: @stasm I don't think so. The spans don't affect their interior text? Recall that in formatToParts the `MessageValue` has direction and language properties for this. Although, your formulation has a certain attraction, as "bonjour" is just some text. One can imagine this:

```
In French {+g @locale=fr @translate=no}bonjour{-g} is a greeting
```

(using `g` for `xliff:g`)

exploration/0002-expression-attributes.md
aphillips: In this case, where the thing affected is not a literal but actually a real placeholder, using a direct attribute is better I think.

exploration/0002-expression-attributes.md
```diff
@@ -38,18 +38,70 @@ At least the following expression attributes should be considered:
...
+    > {Welcome to {$place @note=|The user's current location|}}
+    > ```
+
   - `translate` — A boolean `yes`/`no` indicator communicating to translators
     whether the expression should or should not be localised.
+
+    > Example embedding a non-translatable French literal in an English message:
+    >
+    > ```
+    > {In French, "{|bonjour| @locale=fr @translate=no}" is a greeting}
```
aphillips, line 102: See above: I like `bonjour` being part of the string better than as a placeholder operand. Placeholders often look like a little colored box and the translator has to expand the thing or carefully read the attributes to learn what it says.

aphillips (commented):
Not a full review yet, but sharing comments.

exploration/0002-expression-attributes.md
```diff
@@ -1,6 +1,6 @@
 # Expression Attributes
 
-Status: **Proposed**
+Status: **Accepted**
```
aphillips, line 3: This needs to stay proposed until the WG says otherwise 😉 

```suggestion
Status: **Proposed**
```

exploration/0002-expression-attributes.md
eemeli: With semantic comments we may reasonably describe values such as `$place`, but not specific expressions, which may even be repeated in variant patterns. So expression attributes such as `@example` and `@note` should be considered.

exploration/0002-expression-attributes.md
eemeli: Dropped `@fallback` from the list.

exploration/0002-expression-attributes.md
eemeli: Replaced the example to show how `@id` can make placeholders formatting literal values identifiable.

exploration/0002-expression-attributes.md
eemeli: In this case, would that not be an appropriate representation of a non-translatable literal?

eemeli:
@aphillips @stasm Please take another look?

exploration/expression-attributes.md
```diff
@@ -36,20 +36,62 @@ At least the following expression attributes should be considered:
...
   - `note` — A comment on the expression for translators.
+
+    > Example providing a translator with an example for `$place`:
+    >
+    > ```
+    > Welcome to {$place @note=|The user's current location|}
+    > ```
+
   - `translate` — A boolean `yes`/`no` indicator communicating to translators
     whether the expression should or should not be localised.
```
aphillips, line 87: Add a note about the relationship to e.g. ITS/HTML's `translate` attribute, which also uses `yes`/`no` rather than `true`/`false`?

exploration/expression-attributes.md
```diff
@@ -102,7 +144,7 @@ Define the meaning and supported values of some expression attributes in the spe
 including at least `@dir` and `@locale`.
 
 To support later extension of the specified set of attributes while allowing user extensibility,
-require custom attribute names to include a U+002D Hyphen-Minus `-`.
+suggest custom attribute names to include a U+002D Hyphen-Minus `-`.
```
aphillips, line 147: This might be related to #475 ?

exploration/expression-attributes.md
```diff
@@ -133,7 +175,38 @@ esp. if a similar expression is used in multiple variants.
...
+the locale may be set separately from other options,
+and a custom attribute applied during a conversion to XLIFF.
+
+This approach may produce the same result as the proposed design,
+but has a few caveats:
 
-Later spec versions will not be able to define _any_ new attributes
-without a danger of breaking implementations or users already using those names.
+- The order of operations matters,
+  and an attribute such as the locale must be set before others that are influenced by it.
```
aphillips, line 208: There is the order of evaluation to consider. I was asking myself if the above example evaluated from right-to-left (since setting the locale after formatting a number would be backwards)

aphillips (changes requested):
I think maybe having a source standard for additional expression attributes makes some kind of sense. While XLIFF and ITS and such have been with us a while, evolution is always a possibility here. If there were a source standard, that might suggest namespacing the attributes and making them a plug-in rather than requiring them for MF2's release. Some systems don't have meaning/support for some of these attributes and some systems provide external support for e.g. notes and examples.

That is, can we modularize this out of MF2? Or do we need this now? Or (my preference) can we reserve space for future work?

exploration/expression-attributes.md
eemeli: Yeah, this is ultimately dependent on resolving namespacing.

exploration/expression-attributes.md
eemeli:
```suggestion
  - `translate` — A boolean `yes`/`no` indicator communicating to translators
    whether the expression should or should not be localised.
    The values here correspond to those used for this property in HTML and elsewhere.
```

exploration/expression-attributes.md
eemeli:
```suggestion
- The order of operations matters,
  and an attribute such as the locale must be set before others that are influenced by it.
- It's not clear whether the functions will be evaluated
  left-to-right or right-to-left.
```

stasm:
This PR improves the proposal doc, and keeps its status as "proposed", so I think it's good to merge.

That said, I have a number of comments about the overall attributes proposal. I opened #513 to continue the discussion.

aphillips:
I think we should merge this and make time imminently to discuss the inclusion of expression attributes into the syntax. Currently this looks like it could miss the `LMDL45` train?

macchiati:
Merging is fine, but this should be prioritized below the LDML45 release; that is, not spending time time on it until all LDML45 MVP issues are settled.

eemeli:
This PR adds examples and alternatives to the design doc, and does not change its status. Merging it would make the design doc more cohesive and comprehensive, making it easier for us to consider the proposition as a whole.

Given @stasm's latest verbal approval above and the lack of comments from others on this for the past few months, I would suggest that an approval from @aphillips could be considered sufficient for merging this, which would make it easier to file a follow-up PR updating the proposition's status, for consideration before and at a January meeting.

aphillips (approved):

aphillips:
I am merging this now and will schedule a topic in January to discuss inclusion in `LDML45`. While I am personally a supporter of the inclusion of this feature generally, it is uniquely positioned as something that can be developed during the technical preview with very little to no collateral effect. As such, the discussion will be timeboxed and focused on whether we can live without this feature in the 45 time frame (as always, WG consensus governs).

PR #475, opened 2023-09-13
(Design) Extending functions/options/namespacing

aphillips:
Here is the design document for extending functions and options and the namespacing thereof. 

This describes the namespacing elements and some of the normative text needed, along with early ABNF changes.

Note that I extended the reach of namespaces to:
- functions
- options
- expression attributes


eemeli:
Am confused. Could you clarify what sort of review you're looking for here, given that the doc is "NOT READY FOR LOOKING AT"?

aphillips:
Not looking for a review yet. Just created the CR to hold space until I can finish the work... hence description

eemeli:
✅ Removed the review requests for now; that's what was confusing me.

exploration/0475-overriding-extending-namespacing.md
```diff
@@ -0,0 +1,240 @@
...
+There is no requirement that an implementation read the document at the end of the URL,
+nor validate the contents in general or against the actual functionality installed.
+
+> [!NOTE]
+> It is a good idea to use ASCII strings for namespace identifiers.
+> Remember that translators (and others) in many different languages and
+> with many different keyboards need to be able to enter the prefix.
+
+The namespace prefix is part of the `name` production.
+The prefix is limited to eight characters in length and MUST be at least two characters
```
stasm, line 146: What's the purpose of the 8 character limit?

exploration/0475-overriding-extending-namespacing.md
```diff
@@ -0,0 +1,240 @@
...
+in length.
+The prefix is separated from the name by a colon (U+003A COLON).
+The namespace prefix `std` is reserved and refers to the default registry.
+
+```abnf
+name      = [namespace] name-body
+namespace = name-start *7(name-char) namespace-sep
```
stasm, line 153: By using `name-char` here we're allowing namespace names themselves to contain colons. I think we need to split `name-char` and `nmtoken-char`.

exploration/0475-overriding-extending-namespacing.md
```diff
@@ -0,0 +1,240 @@
...
+namespace-sep = ":"
+name-body = name-start *name-char
```
stasm, line 155: Ditto: this allows names to contain colons which are not namespace separators.

exploration/0475-overriding-extending-namespacing.md
```diff
@@ -0,0 +1,240 @@
...
+or use it for any particular purpose.
+
+> For example, a Java implementation might use the `ServiceProvider` interface to load
+> functionality, while a Node application might use `requires`.
+
+If an implementation supports user-installed formatters, selectors, function options,
+or expression annotations, it must also support providing "namespace" prefixes for
+each installed set of functionality.
+
+In this design, each namespace prefix is a short string and is associated with a URL.
```
stasm, line 133: Is there a URL for the `std` namespace?

exploration/0475-overriding-extending-namespacing.md
```diff
@@ -0,0 +1,240 @@
...
+  and option values for add-on packages.
+
+## Requirements
+
+_What properties does the solution have to manifest to enable the use-cases above?_
+
+- Developers must be able to write functions that do not later collide with items in the default registry.
+- Developers must be able to write function add-ons that do not later collide with items in the default registry.
+- Users should be able to tell visually when an add-on feature has been used vs. a built-in
+- Users should be able to resolve conflicts between add-on packages that use the same
```
stasm, line 108: All of these requirements combined mean that developer are in control of choosing the namespace name. I think that's by design and desirable, but we should address in the doc the negative impact this can have on portability of messages.

stasm (commented):
A quick first pass, but this is looking very much in line with what I had imagined. Thanks for writing it!

It would be great to also include a list of changes to the spec that we'll need to make as a result of this design, so that we can track them and make sure the PR that follows the design is comprehensive. For example:

* From the use-case about tooling, I gather we'll want to design changes to the [data model](https://github.com/unicode-org/message-format-wg/tree/main/spec/data-model), too?
* The *function resolution* section of `formatting.md`. How are implementations supposed to resolve namespaced functions?
* The *error handling* section of `formatting.md`. What happens when a namespace is not known to the runtime?
* Are any changes to the registry required? (It doesn't seem so.)

exploration/0475-overriding-extending-namespacing.md
aphillips: brevity. I meant to include something about the limit, but we should probably make it unlimited.

exploration/0475-overriding-extending-namespacing.md
aphillips: There should be. I can put a placeholder for it.

exploration/0475-overriding-extending-namespacing.md
```diff
@@ -0,0 +1,240 @@
...
+- Developers need to add options to the base functions to suit local needs.
+  For example, ICU's skeletons as part of the `:datetime` function
+  Support for this option needs to be specified for local implemented versions.
+
+- Developers want to write a function and access it from messages.
+
+- Developers want to import 3rd party formatting packages and use the package's
+  features from within messages.
+
+- Developers wish to import two or more formatting packages
```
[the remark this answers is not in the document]
aphillips: check my next commit and see if I captured this.

exploration/0475-overriding-extending-namespacing.md
```diff
@@ -0,0 +1,240 @@
...
+- Users should be able to tell visually when an add-on feature has been used vs. a built-in
+- Users should be able to resolve conflicts between add-on packages that use the same
+  function names without altering add-on packages
+-
+
+## Constraints
+
+_What prior decisions and existing conditions limit the possible design?_
+
+- A syntactical prefix must not collide with `nmtoken`, to avoid parsing ambiguities with unquoted literals...
```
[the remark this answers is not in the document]
aphillips: Added notes.

exploration/0475-overriding-extending-namespacing.md
aphillips: Removed limit and changed wording.

exploration/0475-overriding-extending-namespacing.md
aphillips: Added a note. We still have the Nmtoken problem to solve, so called that out in text.

exploration/0475-overriding-extending-namespacing.md
```diff
@@ -0,0 +1,240 @@
...
+The actual addition and provisioning of features to an implementation is implementation specific.
+Implementations are not required to read the registry format defined by MFv2
+or use it for any particular purpose.
+
+> For example, a Java implementation might use the `ServiceProvider` interface to load
+> functionality, while a Node application might use `import`.
+
+If an implementation supports user-installed formatters, selectors, function options,
+or expression annotations, it must also support providing "namespace" prefixes for
+each installed set of functionality.
```
eemeli, line 137: Do I understand right that with this approach, an implementation could choose to use pretty much _any_ pattern for establishing namespacing, and have it internally work the same as the proposed `ns:func`?

I'm not really seeing the benefit of explicitly defining a separator sigil, as opposed to providing functionality that would let a `name` prefix like `"ns:"` associate itself with a URL defining its functionality.

exploration/0475-overriding-extending-namespacing.md
```diff
@@ -0,0 +1,240 @@
...
+
+> [!NOTE]
+> It is a good idea to use ASCII strings for namespace identifiers.
+> Remember that translators (and others) in many different languages and
+> with many different keyboards need to be able to enter the prefix.
+
+The namespace prefix is part of the `name` production.
+The prefix is limited to eight characters in length and MUST be at least two characters
+in length.
+The prefix is separated from the name by a colon (U+003A COLON).
```
eemeli, line 155: Could some commentary be included explaining why exactly we're proposing `:`, which already has a different meaning in our syntax? Valid alternatives include at least `.`, `/` and `-`.

exploration/0475-overriding-extending-namespacing.md
```diff
@@ -0,0 +1,240 @@
...
+The namespace prefix `std` is reserved and refers to the default registry.
```
eemeli, line 163: I find this a bit confusing. Does this mean that by default these messages should both be valid, and use the same formatter?
```
{{1234 :number}}
```

```
{{1234 :std:number}}
```

exploration/0475-overriding-extending-namespacing.md
```diff
@@ -0,0 +1,240 @@
...
+> following messages might be acceptable alternatives:
+>
+> ```
+> Today is {$today :datetime skeleton=EEEEMd}
+> Today is {$today :datetime icu:skeleton=EEEEMd}
+> Today is {$today :icu:datetime skeleton=EEEEMd}
+> Today is {$today :icu:datetime icu:skeleton=EEEEMd}
+> ```
+
+## Alternatives Considered
```
eemeli, line 264: An additional alternative that was discussed in Seville and should be included here is registry support for prefixes during import that may include a sigil-like trailing character. So for example the ICU skeleton support could be provided by importing it with a `"icu:"` prefix, rather than a `"icu"` prefix that's automatically followed by a `:`.

This would avoid our need to further distance the relevant rules from Nmtoken, and could achieve the same goals with a softer guidance towards some recommended separator character.

eemeli (changes requested):

exploration/0475-overriding-extending-namespacing.md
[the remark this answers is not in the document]
stasm: Yes, thanks.

exploration/0475-overriding-extending-namespacing.md
stasm: Thanks, the note looks good to me.

exploration/0475-overriding-extending-namespacing.md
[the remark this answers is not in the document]
stasm: Ah, I think I understand why we you included this constraint; sorry I missed this earlier. We're effectively saying here that we can't get rid of the `:` in front of the annotation, even if it's namespaced.

Sounds like something worth documenting in #478.

exploration/0475-overriding-extending-namespacing.md
stasm: I think this means that if implementations want to go beyond `std:`, they need to allow users to namespace addons. The pattern for doing so is `ns:func`, and that wouldn't change.

exploration/0475-overriding-extending-namespacing.md
stasm: The way I understood the design is that `{{1234 :std:number}}` is guaranteed to call the implementation's builtin number formatter (which conforms to the requirements stated by the default registry), while `{{1234 :number}}` may call a different one, depending on (a) whether the implementation allows registering packages of custom functions under the anonymous namespace, and (b) such custom package has been somehow made available to the runtime.

If that's the intent, perhaps having some explicit mention of the `anonymous namespace` would be helpful?

exploration/0475-overriding-extending-namespacing.md
aphillips: Added this option.

exploration/0475-overriding-extending-namespacing.md
aphillips: Added.

exploration/0475-overriding-extending-namespacing.md
```diff
@@ -0,0 +1,334 @@
...
+
+> ```
+> Today is {$today :com.example.foo.datetime dateStyle=short}
+> Today is {$today :datetime com.example.foo.skeleton=EEEEMd}
+> ```
+
+- **+** Familiarity. This is a familiar structure for developers.
+- **-** Verbose. The resulting names are long and difficult to parse visually
+
+### Namespacing sigils are customary not normative
```
eemeli, line 322: This would be my preferred solution. Are there significant negatives to it compared to the proposed solution?

stasm: Why do you prefer this one?

IIUC, it doesn't address the requirement about machine-readability: _Translators and tools would like a machine-readable way to find out the names and option values for add-on packages_.

Additionally, messages can end up using different namespacing conventions, in particular in big projects. Coupled with the first issue, unifying them automatically/programmatically can be problematic.

exploration/0475-overriding-extending-namespacing.md
```diff
@@ -0,0 +1,334 @@
...
+- Developers want to import 3rd party formatting packages and use the package's
+  features from within messages.
+
+- Users want to import two or more formatting packages
+  and these might have the same-named functions.
+  For example, there might be both an HTML `p` and TTS `p`
+  function.
+
+- Users want to control how extensions are referenced in their messages.
+  For example, they might wish to make a long namespace name shorter.
```
stasm, line 99: I've been thinking about these 3 use-cases, and I think there's a portability risk that we should call out.

I see two groups of users who could be in charge of controlling the namespace name:

* (a) Authors of custom functions bundled in "packages"; this is similar to how C/C++ or Java handle naming.
* (b) Developers who import those packages into their MFv2 implementation of choice; this is in line with how modern languages handle imports and namespaces, via some sort of `import ... as` mechanism.

The current solution proposed by this PR is to go with (b), and allow developers to adjust the namespace if necessary. The risk this brings is that these custom monikers chosen by developers are then hardcoded inside messages, hurting the portability of messages.

Example: If I choose `myFoo` as a namespace, this naming choice propagates down to all messages (`myFoo:number` etc.). These messages then require edits when they're migrated to another project, when codebases merge, or even when I change my mind and rename the namespace to `myBar`.

We can attempt to avoid or mitigate this risk (or accept it).

* Avoid: Go with (a), i.e. authors choose and control the namespace name. Since this is still likely to produce conflicts, a naming scheme similar to Java's fully qualified names would be recommended. Unfortunately, it means that namespaced names become domain.name.like long.

* Mitigate: Go with (b), i.e. developers importing packages can give them custom names. Embrace the fact that those custom names are then hardcoded in mesasges and make it easy for tooling to work with them and modify them.

aphillips: I agree with this callout. The challenge with (a) is that then you need a fairly sizeable namespace that can be made unique by function authors, e.g. it suggests reverse domain naming. 

There is a third path, which is basically (b) plus a registry (think: media types). I am not wild about creating a registry, but that would help with the portability and toe-stepping problem (while allowing users to control the prefix for their own space).

Since there is no library ecosystem currently, we can also create the mechanism but not flesh it out for 2.0. That is, we can reserve the syntax but come back later to define its usage. In some ways that's like @eemeli's proposal (it's just a convention), except that it is formally reserved (if you name a function with an interior `:`, it might not work later)

exploration/0475-overriding-extending-namespacing.md
eemeli: > Why do you prefer this one?

Because I find it simpler and lighter, and achieving the same results without explicitly adding a new sigil into the syntax.

> IIUC, it doesn't address the requirement about machine-readability: _Translators and tools would like a machine-readable way to find out the names and option values for add-on packages_.

As I understand it, any such machine-readability will need to rely on something like a function registry. My argument with respect to that is that equivalent results may be achieved by declaring in such a registry that a set of functionality uses the `moz-` prefix, rather than the `moz` namespace.

> Additionally, messages can end up using different namespacing conventions, in particular in big projects. Coupled with the first issue, unifying them automatically/programmatically can be problematic.

To reduce such, we could well include a SHOULD statement with a recommended sigil; this would almost certainly be followed by the vast majority of users.

stasm: > As I understand it, any such machine-readability will need to rely on something like a function registry.

I don't think the registry should be relevant here. Instead, I'm imagining the AST for an annotation to look something like `{type: "function", ns: "moz", name: "plural", operand: ..., options: ...}`, which makes the namespace (`ns`) a first-class concept in the data model.

> To reduce such, we could well include a SHOULD statement with a recommended sigil; this would almost certainly be followed by the vast majority of users.

We may be putting a different amount of trust in the SHOULDs :)

aphillips: What @stasm said. There is no requirement that implementations be able to read the default registry file or implement the concept of a registry. They aren't required to provide an extension mechanism either. What this change does is provide a convenient syntactic route by which an implementation can have an extension mechanism and users can write messages that use it. This includes facilitating a translation toolchain downstream of the message.

We don't have to make the namespace/namespace prefix part of the syntax. If it were "just a convention", though, that would prevent us from ever formalizing it as a mechanism (since users might have created `moz-foo` functions that are not actually namespaced). Another option would be to reserve the prefix mechanism for future standardization.

eemeli: Ok, but what do you then do with a `ns: "moz"` without something like a registry?

stasm: > Ok, but what do you then do with a `ns: "moz"` without something like a registry?

I'm looking at it from the PoV of the requirement: _Translators and tools would like a machine-readable way to find out the names and option values for add-on packages._ Having `ns: "moz"` in the data model means that it can be reliably identified by the tooling, e.g. for highlighting, "find all references", mass rename, etc.

eemeli: Reconsidered my stance here. Happy now for us to make the separator explicitly `:` in the syntax.

exploration/0475-overriding-extending-namespacing.md
eemeli: I think we should drop the `std` namespace, and just assert that all non-namespaced names may be extended and used by future spec versions.

If an implementation chooses to e.g. extend `:number` with a custom option like `icu:skeleton`, it's just confusing if there's a `:std:number` as well which might or might not be the same implementation as `:number`.

eemeli (changes requested):

exploration/0475-overriding-extending-namespacing.md
```diff
@@ -0,0 +1,334 @@
...
+In addition, we expect to provide support for markup and templating regimes.
+These need to be implemented using values not found in the default registry.
+
+An additional hope is that a robust ecosystem of function libraries will be created.
+A successful ecosystem will allow users to pick-and-choose or cherry-pick fuctions or
+options to use in a given development environment.
+Each function, option, or option value extension needs to work as seamlessly as possible
+with other add-ons and with the built-in functionality.
+
+To that end, we need to define how externally-authored functions appear in a _message_;
```
catamorphism, line 37: I think this paragraph should come immediately after the heading. As it is, the first sentence after `_What is this proposal trying to achieve?_` doesn't answer the question.

exploration/0475-overriding-extending-namespacing.md
```diff
@@ -0,0 +1,334 @@
...
+> naming concerns.
+> This design document does not show the naming changes because there are
+> other issues in play for these names.
+> For now, just consider that `name-char` will have no colon.
+
+The `name` production as defined here applies to:
+
+- function (selector/formatting) names
+- option names
+- spanable names
```
catamorphism, line 186:
```suggestion
- spannable names
```

exploration/0475-overriding-extending-namespacing.md
```diff
@@ -0,0 +1,334 @@
...
+
+Examples:
+
+> Add-on function:
+>
+> ```
+> Today is {$today :icu:datetime skeleton=EEEEyMdjm}
+> ```
+>
+> Add on option:
```
catamorphism, line 197:
```suggestion
> Add-on option:
```

exploration/0475-overriding-extending-namespacing.md
```diff
@@ -0,0 +1,334 @@
...
+> ```
+>
+> Add-on expression annotation:
+>
+> ```
+> Today is {$today :datetime @my:annotation}
+> ```
+>
+> Everything altogether all at once. This probably does not work
+> correctly, since `std:datetime` may not understand `icu:skeleton`:
```
catamorphism, line 216: So can the example use `icu:datetime` instead?

aphillips:
We agreed to merge this today (2023-11-06) as proposed, but asking for one more review before I do that.

exploration/overriding-extending-namespacing.md
```diff
@@ -0,0 +1,337 @@
+# Design Proposal Template
```
stasm, line 1:
```suggestion
# Extending functions and their options
```

stasm: Or more simply:

```suggestion
# Namespacing
```

exploration/overriding-extending-namespacing.md
```diff
@@ -0,0 +1,337 @@
...
+## Requirements
+
+_What properties does the solution have to manifest to enable the use-cases above?_
+
+- Developers must be able to write functions that do not later collide with items in the default registry.
+- Developers must be able to write function add-ons that do not later collide with items in the default registry.
+- Users should be able to tell visually when an add-on feature has been used vs. a built-in
+- Users should be able to resolve conflicts between add-on packages that use the same
+  function names without altering add-on packages
+-
```
stasm, line 113:
```suggestion
```

stasm (approved):

eemeli (approved):

exploration/0475-overriding-extending-namespacing.md
aphillips: Good point. Moved to start and slightly reworded for context.

exploration/0475-overriding-extending-namespacing.md
aphillips: Done

exploration/0475-overriding-extending-namespacing.md
aphillips: Done

exploration/0475-overriding-extending-namespacing.md
aphillips: Good callout. I made a different change: I made this example use a built-in option and then did `icu:skeleton` for options after.

exploration/overriding-extending-namespacing.md
aphillips: I went with:

> Design Proposal: Namespaces for Extending Functions, Options, etc.

exploration/overriding-extending-namespacing.md
aphillips: Done

exploration/overriding-extending-namespacing.md
```diff
@@ -0,0 +1,336 @@
...
+Much of this functionality will be mandated by the default registry.
+We expect that default registry entries will serve the core needs for MF2 users.
+However, there are many capabilities available in platform, library,
+or operating environment APIs that could be useful to developers and translators
+or which might be expected on a specific platform or in a specific programming language.
+In addition, we expect to provide support for markup and templating regimes.
+These need to be implemented using values not found in the default registry.
+
+An additional hope is that a robust ecosystem of function libraries will be created.
+A successful ecosystem will allow users to pick-and-choose or cherry-pick fuctions or
```
catamorphism, line 36:
```suggestion
A successful ecosystem will allow users to pick-and-choose or cherry-pick functions or
```

exploration/overriding-extending-namespacing.md
```diff
@@ -0,0 +1,336 @@
...
+Skeletons are not proposed for inclusion in the default registry
+because they are not universally available in all datetime formatting
+libraries.
+
+## Use-Cases
+
+_What use-cases do we see? Ideally, quote concrete examples._
+
+- Developers need to add options to the base functions to suit local needs.
+  For example, ICU's skeletons as part of the `:datetime` function
```
catamorphism, line 85:
```suggestion
- Developers need to add options to the base functions to suit local needs:
  for example, ICU's skeletons as part of the `:datetime` function.
```

catamorphism (approved):

mihnita (approved):

exploration/overriding-extending-namespacing.md
```diff
@@ -0,0 +1,336 @@
+# Design Proposal: Namespaces for Extending Functions, Options, etc.
+
+Status: **Proposed**
```
aphillips, line 3:
```suggestion
Status: **Accepted**
```

PR #477, opened 2023-09-17
Document the design of quoted literals

stasm:
An alternative to #465 — an explainer doc on why we need quoted literals and why we settled on delimiting them with vertical bars.

[Rendered.](https://github.com/unicode-org/message-format-wg/blob/design-quoted-literals/exploration/quoted-literals.md)


exploration/0477-quoted-literals.md
```diff
@@ -0,0 +1,172 @@
...
+
+- [r1 POOR] Writing `"` and `'` in literals requires escaping them via `\`, which then needs to be escaped itself in code which uses `\` as the escape character (which is common).
+- [r2 FAIR] Embedding messages in certain programming languages and containers requires escaping the literal delimiters. Most notably, storing MF2 messages in JSON suffers from this. In many programming languages, however, alternatives to quotation marks exist, which could be used to allow unescaped quotes in messages. See [comment on #263](https://github.com/unicode-org/message-format-wg/issues/263#issuecomment-1430929542).
+- [r3 FAIR] One of the suggestions proposed to allow for both single and double quotation marks, and make them interchangeable in case one set was used by the inner content or surrounding code. This, however, requires directed modification of the message's body.
+- [r4 GOOD] Quotation marks are universally recognized as string delimiters.
+- [r5 POOR] Quotation marks cannot be paired by parsers nor IDEs.
```
eemeli, line 159: In general, IDEs commonly assume by default that double quotes`"` are paired, and handle them as pairs. Given sufficient context, this also applies to single quotes `'`.

Vim provides built-in functionality for selecting within quotes `vi'` or around them `va"`.

This behaviour is not "POOR", it's either "FAIR" or "GOOD".

exploration/0477-quoted-literals.md
```diff
@@ -0,0 +1,172 @@
...
+
+### Use round or angle brackets
+
+- Round parentheses are very uncommon as string delimiters [r2 GOOD], and thus may be surprising [r4 POOR]. Furthermore, they are relatively common in text, where they'd require escaping [r1 POOR].
```
eemeli, line 163: If this is r4 POOR, then the same must hold for vertical bars `|`.

Prior art does exist for parentheses `()` as literal delimiters: https://en.wikipedia.org/wiki/PostScript#%22Hello_world%22

eemeli (changes requested):

exploration/0477-quoted-literals.md
```diff
@@ -0,0 +1,172 @@
...
+  > }
+  > ```
+
+## Requirements
+
+_What properties does the solution have to manifest to enable the use-cases above?_
+
+- **[r1; high priority]** Minimize the need to escape characters inside literals. In particular, choose a delimiter that isn't frequently used in translation content. Having to escape characters inside literals is inconvenient and error-prone when done by hand, and it also introduces the backslash into the message, `\`, which is the escape introducer. The backslash then needs to be escaped too, when the message is embedded in code or containers. (This is how some syntaxes produce the gnarly `\\\`.)
+- **[r2; high priority]** Minimize the need to escape characters when embedding messages in code or containers. In particular, choose a delimiter that isn't frequently used as a string delimiter in programming languages and container formats. However, note that many programming languages also provide alternative ways of delimiting strings, e.g. _raw strings_ or triple-quoted literals.
+- **[r3; high priority]** Minimize the need to change the message in other ways than to escape some of its characters (e.g. rephrase content or change syntax).
```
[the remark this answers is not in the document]
stasm: I meant to include here the need to change straight quotes to curly quotes, or to switch from single quotes to double quote, or the other way around.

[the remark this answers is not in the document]
stasm: I'm OK with dropping this to medium priority.

exploration/0477-quoted-literals.md
```diff
@@ -0,0 +1,172 @@
...
+- **[r4; medium priority]** Don't surprise users with syntax that's too exotic. We expect quoted literals to be rare, which means fewer opportunities to get used to their syntax and remember it.
```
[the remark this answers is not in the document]
stasm: I chose "high" for things that can break translations or make working with them objectively harder. The exoticness of the syntax is in the eye of the beholder, and thus is chose "medium" — it's difficult to rate a solution objectively here.

exploration/0477-quoted-literals.md
```diff
@@ -0,0 +1,172 @@
...
+              / %xE000-10FFFF
+quoted-escape = backslash ( backslash / "|" )
+```
+
+By being both uncommon in text content and uncommon as a string delimiter in other programming languages, the vertical line sidesteps the "inwards" and "outwards" problems of escaping.
+
+- [r1 GOOD] Writing `"` and `'` in literals doesn't require escaping them via `\`. This means no extra `\` that need escaping.
+- [r2 GOOD] Embedding messages in most code or containers doesn't require escaping the literal delimiters.
+- [r3 GOOD] Message don't have to be modified otherwise before embedding them.
+- [r4 FAIR] Vertical lines are not commonly used as string delimiters and thus can be harder to learn for beginners. There's prior art in a practice of using vertical lines as delimiters for inline code literals.
```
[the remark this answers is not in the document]
stasm: I don't have a concrete reference other than my own experience from NNTP and IRC, where `|foo|` was a common way to delimit code before backticks became popular.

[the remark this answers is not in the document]
stasm: https://en.wikipedia.org/wiki/Posting_style#Quoted_line_prefix mentions the vertical bar, but for quoting replies, not inline code.

https://en.wikipedia.org/wiki/Vertical_bar#Delimiter mentions it as a delimiter for strings, but in all fairness, in rare and obscure cases.

exploration/0477-quoted-literals.md
```diff
@@ -0,0 +1,172 @@
...
+- [r5 POOR] Vertical lines cannot be paired by parsers nor IDEs.
+
+## Alternatives Considered
+
+_What other solutions are available?_
+_How do they compare against the requirements?_
+_What other properties they have?_
+
+### Use quotation marks
```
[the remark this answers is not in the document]
stasm: Would you mind adding a section on dual quoting to the branch?

exploration/0477-quoted-literals.md
stasm: IDEs can assume quotes are paired, but will not know which direction the pair extends into. Consider: `aaa"bbb"ccc"ddd`; in vim you can use `vi"` to delete both `bbb` and `ccc`, depending on where your caret is positioned.

Let's switch this to FAIR. There is *some* support here, but not as good as for brackets. (I'll add a point about r5 in the section about using brackets.)

exploration/0477-quoted-literals.md
stasm: I didn't realize there was prior art for `()` as string delimiters.

I can still see a difference between `||` and `()`: round parens have a well-established meaning in prose. In my mind, this justifies the POOR rating for r4.

exploration/0477-quoted-literals.md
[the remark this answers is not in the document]
eemeli: I would appreciate this priority being dropped a peg, and including a bit more explanation in the text. All the other requirements include a bit of reasoning and commentary to explain themselves.

exploration/0477-quoted-literals.md
stasm: Switched to FAIR and mentioned that most text editors have features to make working with quotes easier.

exploration/0477-quoted-literals.md
stasm: I mentioned prior art and added the "well-established meaning in prose" as one of the drawbacks. I kept the rating at POOR.

exploration/0477-quoted-literals.md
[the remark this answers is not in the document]
eemeli: While I agree that some comparisons here are difficult or impossible to rank objectively: "Is `||` more or less exotic than `()` as text delimiters?" is for instance a near-impossible question to answer.

I would, however, posit that comparisons like "Is `||` more or less exotic than `""` as text delimiters?" may be objectively answered. Witness, for instance, their use in each of our comments on this thread.

exploration/0477-quoted-literals.md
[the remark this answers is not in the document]
eemeli: I would prefer some reference being included as a link, if the assertion is retained.

I think the applicability of the prior art is debatable, but as long as it's qualified in some way (e.g. via a link, or by describing it in the text as "rare and obscure") I don't think it's worth quibbling over.

exploration/0477-quoted-literals.md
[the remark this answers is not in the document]
stasm: Agreed. That's why it's medium, not low :)

exploration/0477-quoted-literals.md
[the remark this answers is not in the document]
eemeli: Thank you for the update. This is still missing the sort of rationalisation that is provided for the other requirements. It is not immediately obvious (at least to me) _why_ this is a requirement. What sort of operation on a message source hits this issue?

eemeli:
I split out "Dual quoting" into its own alternative, and updated "Use quotation marks" to correspond.

I left the r3 assessment as `???` for these, as I'm still not sure what it's measuring, and it wasn't clear to me if the previous "FAIR" would apply to both alternatives.

@stasm Have a look at [semantic line breaks](https://sembr.org/) at some point? It's the markdown style we've been using for most content here.

exploration/0477-quoted-literals.md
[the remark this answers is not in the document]
stasm: OK, I see that this requirement isn't well worded. I'll try to rephrase it (possibly tomorrow).

As a discussion point for right now, in your commit on dual quoting, you wrote:

> If the container format does not itself support dual quoting, the embedded message's quotes may be adjusted to avoid their escaping.

This is an example of "needing to change the message in other ways than to escape some of its characters" that I wanted to capture here.

Ideally, it should be possible to drop the message verbatim into code or a container. The next best thing (in my mind) is to only escape some characters, do it rarely, and also do it without thinking about it too much. OTOH, H\having to modify the body in any other way, e.g. switch from straight quotes to curly quotes in translation content, or switch between single and double quotes around literals, is more effort (again, in my mind). I think we should prioritize solutions that don't require such edits, but I also acknowledge that this isn't as high on the priority list as r1 and r2.

stasm:
> @stasm Have a look at [semantic line breaks](https://sembr.org/) at some point? It's the markdown style we've been using for most content here.

Oh, sure! I thought we were only using it for spec text. I'l admit that it was easier for me to write this design doc without worrying about line breaks. But, happy to refactor this PR to use them.

exploration/0477-quoted-literals.md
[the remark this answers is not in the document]
eemeli: It might be useful to identify a few specific exemplars of likely container formats in which MF2 messages are expected to often be embedded by developers, such as Java, JS, .properties, YAML, JSON, our own resource format. Are there others we should explicitly consider?

eemeli:
Eh, it makes markdown sources a bit easier to read and edit. I would not consider a hard requirement of any sort, more of a good practice in general. 😇

exploration/0477-quoted-literals.md
```diff
@@ -0,0 +1,192 @@
...
+1. encoding literals containing whitespace, like literals consisting of multiple words,
+1. encoding literals containing exotic characters that do not conform to the `unquoted` production in ABNF.
+
+More specifically:
+
+- Message authors and translators need to be able to use the apostrophe in the message content, and may want to use the single quote character to represent it instead of the typograhic (curly) apostrophe.
+
+  > ```
+  > {…{|New Year's Eve|}…}
+  > ```
```
aphillips, line 40: This is the wrong example. The problem isn't the occasional use of apostrophes in literals but the unavailability of apostrophe (and double-quote) in the message body:

```suggestion
>```
>{You can't have more than {$num :number} {'bone dragons'}.}
>```
```
```

File: exploration/0477-quoted-literals.md [hunk 8]
```diff
@@ -0,0 +1,192 @@
...
+  > {The Unix epoch is defined as {|Thu, 01 Jan 1970 00:00:00 GMT| :datetime}.}
+  > ```
+
+- Message authors may want to use multiple words as values of certain options passed to custom functions and markup elements:
+
+  > ```
+  > {{+button title=|Click here!|}Submit{-button}}
+  > ```
+  >
+  > Note that quoted literals cannot contain placeholders, making interpolating data into them impossible.
```
aphillips, line 54: This is a note and should go last

exploration/0477-quoted-literals.md
```diff
@@ -0,0 +1,192 @@
...
+  >
+  > ```
+  > -- This is impossible in MessageFormat 2.0.
+  > {{+button title=|Goodbye, {$userName}!|}Sign out{-button}}
+  > ```
+
+- Selector function implementers may want to support exotic characters in variant keys to effectively create "mini-DSLs" for the matching logic:
```
aphillips, line 61:
```suggestion
A selector's potential `when` values might need to include a wide range of characters,
such as needed for a "mini-DSL" or for comparison of string values defined
externally:
```

I'd probably do a couple of examples here, Note the differences in mine:

>```
>match ($count :choice}
> when |<10| {A handful.}
> when * {Lots.}
>
>match {$arbitraryString}
>when |can't resolve| {Can't resolve!}
>when |11'233.44| {Locale formatted number}
>when * {Imagine more...}
>```
```

File: exploration/0477-quoted-literals.md [hunk 10]
```diff
@@ -0,0 +1,192 @@
...
+quoted-char   = %x0-5B         ; omit \
+              / %x5D-7B        ; omit |
+              / %x7D-D7FF      ; omit surrogates
+              / %xE000-10FFFF
+quoted-escape = backslash ( backslash / "|" )
+```
+
+By being both uncommon in text content and uncommon as a string delimiter in other programming languages, the vertical line sidesteps the "inwards" and "outwards" problems of escaping.
+
+- [r1 GOOD] Writing `"` and `'` in literals doesn't require escaping them via `\`. This means no extra `\` that need escaping.
```
aphillips, line 139: I like the scoring thing, but when writing documents like this, I tend to prefer a table for "all up" comparison somewhere in the document, perhaps at the end, e.g.

| Option | Writing | Embedding | Modification | Commonality |
|---|---|---|---|---|
| '|' | + | + | + | 0 | - |
| `'` `"` | - | - | - | + | 0 |

exploration/0477-quoted-literals.md
```diff
@@ -0,0 +1,192 @@
...
+- [r2 GOOD] Embedding messages in most code or containers doesn't require escaping the literal delimiters.
+- [r3 GOOD] Message don't have to be modified otherwise before embedding them.
+- [r4 FAIR] Vertical lines are not commonly used as string delimiters and thus can be harder to learn for beginners. There's prior art in a practice of using vertical lines as delimiters for inline code literals in email and chat. Vertical bars can also be used to delimit [regular expressions in `sed`](https://en.wikipedia.org/wiki/Vertical_bar#Delimiter), and as a separator in [delimiter-separated data formats](http://www.catb.org/~esr/writings/taoup/html/ch05s02.html).
+- [r5 POOR] Vertical lines cannot be paired by parsers nor IDEs.
```
aphillips, line 143: cannot => are not automatically

They could be autopaired (apostrophe and double-quotes are paired by IDEs but are not normally considered open/close punctuation)

exploration/0477-quoted-literals.md
stasm: It’s legal to have quotes inside patterns. That’s by design, and the consequence of our using curly braces as pattern delimiters. This use-case was meant to be about the occasional use of apostrophe inside literals :)

exploration/0477-quoted-literals.md
```diff
@@ -0,0 +1,192 @@
...
+
+- **[c1]** MessageFormat uses the backslash, `\`, as the escape sequence introducer.
+- **[c2]** Straight quotation marks, `'` and `"`, are common in content across many languages, even if other Unicode codepoints should be used in well-formatted text.
+- **[c3]** Straight quotation marks, `'` and `"`, are common as string delimiters in many programming languages.
+
+## Proposed Design
+
+_Describe the proposed solution. Consider syntax, formatting, errors, registry, tooling, interchange._
+
+Use the vertical line character, `|`, to delimit quoted strings. For example:
```
aphillips, line 121: Somewhere there needs to be a bit about why we specifically chose `|` and not some other character. We have some specific cases (apostrophe, double quote, parens) that we rubbish. I think we are not as direct about saying: this character is rare in normal text.

exploration/0477-quoted-literals.md
```diff
@@ -0,0 +1,192 @@
...
+  > {The official native name of the Republic of Poland is {|Rzeczpospolita Polska| @lang=pl}.}
+  > ```
+  >
+  > See [design proposal 0002](https://github.com/unicode-org/message-format-wg/blob/main/exploration/0002-expression-attributes.md).
+
+- Developers may want to embed messages with quoted literals in code written in another programming language which uses single or double quotes to delimit strings.
+
+  > ```js
+  > let message = new MessageFormat("en", "{A message with {|a literal|}.}");
+  > ```
```
gibson042, line 89: I think this should be better distinguished from the following JSON example, which necessarily uses double quotes.
```suggestion
- Developers may want to embed messages with quoted literals in code written in another programming language which uses single or double quotes to delimit strings.

  > ```js
  > let message = new MessageFormat('en', '{A message with {|a literal|}.}');
  > ```
```

exploration/0477-quoted-literals.md
```diff
@@ -0,0 +1,192 @@
...
+
+- **[r1; high priority]** Minimize the need to escape characters inside literals. In particular, choose a delimiter that isn't frequently used in translation content. Having to escape characters inside literals is inconvenient and error-prone when done by hand, and it also introduces the backslash into the message, `\`, which is the escape introducer. The backslash then needs to be escaped too, when the message is embedded in code or containers. (This is how some syntaxes produce the gnarly `\\\`.)
+- **[r2; high priority]** Minimize the need to escape characters when embedding messages in code or containers. In particular, choose a delimiter that isn't frequently used as a string delimiter in programming languages and container formats. However, note that many programming languages also provide alternative ways of delimiting strings, e.g. _raw strings_ or triple-quoted literals.
+- **[r3; medium priority]** Minimize the need to change the message in other ways than to escape some of its characters (e.g. rephrase content, use typographic apostrophes, or switch to using a second set of delimtiers).
```
gibson042, line 105:
```suggestion
- **[r3; medium priority]** Minimize the incentive to avoid escaping by changing messages (e.g. rephrasing content, using typographic apostrophes, or switching outer delimiters).
```

exploration/0477-quoted-literals.md
stasm: I agree, but unfortunately, it looks like our Prettier bot disagrees: https://github.com/unicode-org/message-format-wg/pull/477/commits/22a4cb2874e1a3ed19747229b72356a55bd9c312. @eemeli, help? :)

eemeli: Filed #481 for this.

exploration/0477-quoted-literals.md
```diff
@@ -0,0 +1,192 @@
...
+
+More specifically:
+
+- Message authors and translators need to be able to use the apostrophe in the message content, and may want to use the single quote character to represent it instead of the typograhic (curly) apostrophe.
+
+  > ```
+  > {…{|New Year's Eve|}…}
+  > ```
+
+- Message authors may want to use literals to define locale-aware dates as literals in the RFC 7231 format:
```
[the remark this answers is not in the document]
stasm: Do you mean something like this?

```suggestion
- Message authors may want to use literals to define locale-aware dates as literals in a modified RFC 3339 format:

    > ```
    > {The Unix epoch is defined as {|1970-01-01 00:00:00Z| :datetime}.}
    > ```
```

exploration/0477-quoted-literals.md
stasm: Not sure if I understand. Would you prefer me to use the `> [!NOTE]` syntax here?

exploration/0477-quoted-literals.md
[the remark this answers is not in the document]
aphillips: Yep.

exploration/0477-quoted-literals.md
```diff
@@ -0,0 +1,268 @@
...
+  > ```
+  >
+  > See [design proposal 0002](https://github.com/unicode-org/message-format-wg/blob/main/exploration/0002-expression-attributes.md).
+
+- Developers may want to embed messages with quoted literals in code written in another programming language
+  which uses single or double quotes to delimit strings.
+
+  > ```js
+  > let message = new MessageFormat("en", "{A message with {|a literal|}.}");
+  > ```
```
eemeli, line 115:
```suggestion
  <!-- prettier-ignore -->
  > ```js
  > let message = new MessageFormat('en', '{A message with {|a literal|}.}');
  > ```
```

exploration/0477-quoted-literals.md
```diff
@@ -0,0 +1,268 @@
...
+  The backslash then needs to be escaped too,
+  when the message is embedded in code or containers.
+  (This is how some syntaxes produce the gnarly `\\\`.)
+
+- **[r2; high priority]** Minimize the need to escape characters when embedding messages in code or containers.
+  In particular, choose a delimiter that isn't frequently used as a string delimiter in programming languages and container formats.
+  However, note that many programming languages also provide alternative ways of delimiting strings, e.g. _raw strings_ or triple-quoted literals.
+
+- **[r3; medium priority]** Minimize the incentive to avoid escaping by changing messages
+  (e.g. rephrasing content, using typographic apostrophes, or switching outer delimiters).
```
eemeli, line 143: Do I understand right that the first two examples are referring to changes in the actual message contents, while the third one is a resource format concern? If so, that seems like a separate concern from the first two that ought to be expressed as a separate requirement.

exploration/0477-quoted-literals.md
```diff
@@ -0,0 +1,268 @@
...
+
+- **[r4; medium priority]** Don't surprise users with syntax that's too exotic.
+  We expect quoted literals to be rare,
+  which means fewer opportunities to get used to their syntax and remember it.
```
eemeli, line 147: I think the priority of this requirement should be ranked higher. I also realise that I ought to look into the data I have available to establish how rare the need for quoting will be, as that might be better than relying just on our expectations.

```suggestion
- **[r4; high priority]** Don't surprise users with syntax that's too exotic.
  We expect quoted literals to be rare,
  which means fewer opportunities to get used to their syntax and remember it.
```

exploration/0477-quoted-literals.md
```diff
@@ -0,0 +1,268 @@
...
+
+- **[r5; low priority]** Be able to pair the opening and the closing delimiter,
+  to aid parsers recover from syntax errors,
```
eemeli, line 150: I don't see how paired delimiters would help parsers in recovery. A missing quote delimiter of any sort is pretty fatal for the whole message, even for a parser that can do some recovery.

If we did want to actually help here, then requiring `{}` to be escaped in quoted literals would let the damage be limited in most cases to a single expression.

exploration/0477-quoted-literals.md
```diff
@@ -0,0 +1,268 @@
...
+By being both uncommon in text content
+and uncommon as a string delimiter in other programming languages,
+the vertical line sidesteps the "inwards" and "outwards" problems of escaping.
+
+- [r1 GOOD] Writing `"` and `'` in literals doesn't require escaping them via `\`.
+  This means no extra `\` that need escaping.
+- [r2 GOOD] Embedding messages in most code or containers doesn't require escaping the literal delimiters.
+- [r3 GOOD] Message don't have to be modified otherwise before embedding them.
+- [r4 FAIR] Vertical lines are not commonly used as string delimiters
+  and thus can be harder to learn for beginners.
```
eemeli, line 199: This really is "poor". Using `|` as a delimiter is both exotic and surprising.

```suggestion
- [r4 POOR] Vertical lines are not commonly used as string delimiters
  and thus can be harder to learn for beginners.
```

exploration/0477-quoted-literals.md
```diff
@@ -0,0 +1,268 @@
...
+  There's prior art in a practice of using vertical lines as delimiters for inline code literals in email and chat.
```
eemeli, line 200: As far as I'm aware, in these use cases `|` is used as a line prefix, much like `>` for email replies and markdown blockquotes.

I think it's quite indicative that this very different usage is considered the prime example of prior art.

exploration/0477-quoted-literals.md
```diff
@@ -0,0 +1,268 @@
...
+  Vertical bars can also be used to delimit [regular expressions in `sed`](https://en.wikipedia.org/wiki/Vertical_bar#Delimiter),
+  and as a separator in [delimiter-separated data formats](http://www.catb.org/~esr/writings/taoup/html/ch05s02.html).
```
eemeli, line 202: For sed, this argument can be made about [almost any character](https://pubs.opengroup.org/onlinepubs/009604599/utilities/sed.html):
> Any character other than backslash or \<newline> can be used instead of a slash to delimit the BRE and the replacement.

I, for one, tend to use `#` when the regexp includes `/` characters.

This is not really valid prior art for `|`.

You may want to also consider markdown tables as a "delimiter" use of `|`, though I'd say that's about as much a stretch as its use as a DSV separator.

```suggestion
  Vertical bars can also be used as a separator in [delimiter-separated data formats](http://www.catb.org/~esr/writings/taoup/html/ch05s02.html).
```

eemeli (changes requested):

exploration/0477-quoted-literals.md
aphillips: Paired delimiters don't help runtime parsers in recovery (as there is effectively no recovery in many cases), but they do help a different sort of parser: editors and other tools often auto-pair enclosing punctuation. So I'd probably phrase this differently and I might boost the priority a little.

The problem is that there are not many ASCII paired characters: `()`, `{}`, and `[]`

exploration/0477-quoted-literals.md
aphillips: This (and later comments) about differing priorities is why I suggested a table. We might not agree about the specific priorities of this or that item, but it's hard to compare options when they are discrete bits of text spread over many paragraphs. It's easier to "rescore" when everything is in one place.

aphillips:
I'm going to recommend that we merge this design document at the "proposed" maturity level.

exploration/0477-quoted-literals.md
stasm: I added a comparison matrix in 9d53b4d6ad4c1932d7e3e142065391844606e294.

exploration/0477-quoted-literals.md
stasm: @eemeli ended up fixing this via a `prettier-ignore` comment, thanks.

eemeli:
> I'm going to recommend that we merge this design document at the "proposed" maturity level.

My line comment concerns and suggestions above are still valid and largely unaddressed, and I'd really prefer not needing to reformulate them as a separate PR.

stasm:
We discussed this at the teleconference and agreed not to merge before I address all comments. Sorry for not having updated the status here.

aphillips:
@stasm ping to address comments so we can merge

stasm:
Thanks for the ping. I'll do it before the Monday meeting.

exploration/0477-quoted-literals.md
stasm: I'll change this to `medium/high` to capture the difference in priority scores.

exploration/0477-quoted-literals.md
stasm: > I don't see how paired delimiters would help parsers in recovery. A missing quote delimiter of any sort is pretty fatal for the whole message, even for a parser that can do some recovery.

I admit that the technique that I had in mind here is rather obscure. Namely, if both paired delimiters are made special in the literal, then the start of another literal can be an anchor point for a parser to resume parsing.

```
    There {:is a=|broken literal=|here|}
                        ^         ^
                        The closing delimiter is missing here.
                                  The syntax error occurs here.

    There {:is a=[broken literal=[here]}
                        ^^       ^ 
                        The closing delimiter is missing here.
                         |       The parser can recognize a new literal here...
                         and rewind to here.
```

I'll add a `<details>` block to explain this.

exploration/0477-quoted-literals.md
stasm: I agree that when used in the same context as `>`, it would not be a good prior art. What I meant here, instead, was using vertical bars the way markdown uses backticks.

I remember seeing such use and personally using it myself in irc.mozilla.org. I'm not able to find any evidence of it, however. I'll remove this point.

exploration/0477-quoted-literals.md
stasm: Your comment made me realize that vertical lines are indeed more often used as delimiters for _separating_ rather than for _enclosing_.

exploration/0477-quoted-literals.md
stasm: Changed it to `POOR/FAIR`.

eemeli:
I added a new alternative to the doc that effectively merges the proposed design with `[a2]` by allowing for _either_ `|vertical pipes|`, `'single quotes'` or `"double quotes"`.

It's an attempt to combine the best of both approaches by allowing for "normal" quotes where they're available, while also supporting something a bit more exotic when embedding in formats like JSON.

exploration/quoted-literals.md
```diff
@@ -0,0 +1,390 @@
...
+  > Quoted literals cannot contain placeholders, making interpolating data into them impossible.
+  >
+  > ```
+  > -- This is impossible in MessageFormat 2.0.
+  > {+button title=|Goodbye, {$userName}!|}Sign out{-button}
+  > ```
+
+- Selector function implementers may want to support multi-word variant keys
+  or exotic characters in variant keys
+  to effectively create "mini-DSLs" for the matching logic:
```
aphillips, line 77:
```suggestion
- Selector function implementers might need to match different string values
  such as those present in data values.
  These might include keys containing arbitrary text, multiple words,
  or other sequences not otherwise permitted in the syntax.
```

exploration/quoted-literals.md
```diff
@@ -0,0 +1,390 @@
...
+  and it also introduces the backslash into the message, `\`,
+  which is the escape introducer.
+  The backslash then needs to be escaped too,
+  when the message is embedded in code or containers.
+  (This is how some syntaxes produce the gnarly `\\\`.)
+
+- **[r2; high priority]** Minimize the need to escape characters when embedding messages in code or containers.
+  In particular, choose a delimiter that isn't frequently used as a string delimiter in programming languages and container formats.
+  However, note that many programming languages also provide alternative ways of delimiting strings, e.g. _raw strings_ or triple-quoted literals.
+
```
aphillips, line 143: I think we're missing a requirement? We have minimization of escaping inside the literal and minimization of escaping for the host format. We also don't want to choose a character/sequence that requires normal patterns to have lots of escapes.

```suggestion
- **[r2a; high priority]** Minimize the need to escape characters in patterns when writing normal messages.
```

What I mean here is:
```
{{This has \|an escape\| because it needs it}}
{{This has \'an escape\' because it needs it}}
```

aphillips (changes requested):

exploration/0477-quoted-literals.md
stasm: Not quite. All three examples in parentheses relate to literals:

* Rephrasing content: `Tom's pen` → `A pen of Tom`.
* Using typographic quotes: `Tom's pen` → `Tom’s pen`.
* Switching delimiters: `'Tom's pen'` → `"Tom's pen"`.

I'll clarify by `s/outer/literal`.

stasm: Actually, re-reading the previous discussion about r3, let me also add r6 about the use-case of dropping messages into host formats.

exploration/quoted-literals.md
```diff
@@ -0,0 +1,390 @@
...
+  > ```
+  > {+button title=|Click here!|}Submit{-button}
+  > ```
+
+  > [!NOTE]
+  > Quoted literals cannot contain placeholders, making interpolating data into them impossible.
+  >
+  > ```
+  > -- This is impossible in MessageFormat 2.0.
+  > {+button title=|Goodbye, {$userName}!|}Sign out{-button}
```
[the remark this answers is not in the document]
stasm: It's not surprising once you know that the value of `title=` *can only be a literal*, and nothing else. 

What I attempted to capture here is that the use-case in question (*Message authors may want to use multiple words as values of certain options passed to custom functions and markup elements*) can only be satisfied partially anyways by MF2.

exploration/quoted-literals.md
stasm: Why would these patterns require any escaping?

exploration/quoted-literals.md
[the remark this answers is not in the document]
aphillips: > What I attempted to capture here is that the use-case in question (Message authors may want to use multiple words as values of certain options passed to custom functions and markup elements) can only be satisfied partially anyways by MF2.

The above could be solved by using a `local`:

```
{{
   local $title = {|Goodbye, {$username}!}
   {{{+button title=$title}Sign out{-button}}}
}}
```

That's complicated to localize, though.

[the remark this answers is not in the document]
stasm: `$title` is still a literal, and so `$username` would not be resolved. Or did you mean a possible extension of the syntax that would allow locals to be bound to _patterns_?

```
{{
   local $title = {{Goodbye, {$username}!}}
   {{{+button title=$title}Sign out{-button}}}
}}
```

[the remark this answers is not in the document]
aphillips: No, you're right. Not enough ☕ yet this morning. Actually fixing this would require something like a string-format function. Actually _internationalizing_ it would require stuff like gender selection. So I'm busy proving your point 😉

exploration/quoted-literals.md
```diff
@@ -155,6 +156,16 @@ _What properties does the solution have to manifest to enable the use-cases abov
...
+- **[r6; low priority]** Be able to drop messages into host formats without changing them.
+  For example, if the host format uses `""` to delimit strings literals,
+  then all occurences of `"` inside messages must be escaped.
+  While we can't prevent `"` from appearing in translation content,
+  we can minimize the overhead of the MessageFormat 2 syntax.
+  This requirement is scored as _low_, because many storage formats don't use delimiters at all (`.properties`, YAML),
+  or they are meant to be primarily used by machines (JSON),
+  and because many programming languages provide a way to delimit _raw strings_,
+  e.g. via backticks in JavaScript and `"""` in Python.
+
```
eemeli, line 168:
```suggestion
  Note that messages including e.g. newlines or `\` escapes in their source
  will likely need those characters accounted for when dropping them into new host formats.

```

exploration/quoted-literals.md
```diff
@@ -0,0 +1,409 @@
+# Quoted Literals
+
+Status: **Accepted**
```
eemeli, line 3: As discussed during the call, this should land first as "proposed".

```suggestion
Status: **Proposed**
```

exploration/quoted-literals.md
stasm: Eemeli and I met, and we included this suggestion in the PR.

exploration/quoted-literals.md
```diff
@@ -0,0 +1,403 @@
...
+  Message don't have to be modified otherwise before embedding them,
+  unless they happen to contain conflicting quote delimiters.
+- [r2 GOOD] Embedding messages in most code or containers doesn't require escaping the literal delimiters.
+- [r3 GOOD] Quotation marks are universally recognized as string delimiters.
+- [r4 FAIR] Using the same marks for quote-start and quote-end cannot be paired by parsers nor IDEs,
+  but many text editors provide features to make working with and around quotes easier.
+
+## Comparison table
+
+<table>
```
aphillips, line 335: This is a job for markdown tables.

Don't write it by hand. Use a [tool](https://www.tablesgenerator.com/markdown_tables)

exploration/quoted-literals.md
```diff
@@ -0,0 +1,403 @@
...
+              / %x5D-7B        ; omit |
+              / %x7D-D7FF      ; omit surrogates
+              / %xE000-10FFFF
+quoted-escape = backslash ( backslash / "|" / "'" / DQUOTE )
+```
+
+- [r1 GOOD] Writing any two of `|`, `"` and `'` in literals doesn't require escaping them via `\`.
+  This means no extra `\` that need escaping.
+  Message don't have to be modified otherwise before embedding them,
+  unless they happen to contain conflicting quote delimiters.
```
aphillips, line 327: This is actually a con "in a trenchcoat" :-)

Messages MUST be modified if they "happen to contain conflicting quote delimiters", e.g. those that conflict with the syntax.

My big concern here is: how to do tools up and down the translation stack decide what quotes to use?

There should be a "con" somewhere for having multiple ways of doing the same thing.

exploration/quoted-literals.md
```diff
@@ -0,0 +1,403 @@
...
+
+Allow any of the following as literal delimiters:
+
+- the vertical line character `|`
+- single quotes `'`
+- double quotes `"`
+
+This approach supports multiple different quoting styles to be used for literals.
+This flexibility allows for using a familiar and common style such as `'single'` or `"double"` quotes,
+while also allowing for `|pipes|` when the message's contents or embedding would otherwise require additional escaping.
```
aphillips, line 308: define "additional escaping". Do we mean due to the containing syntax? Or do we mean the literal's content? Or both?

aphillips (changes requested):

exploration/quoted-literals.md
eemeli:
```suggestion
while also allowing for `|pipes|` when the message's contents or embedding would otherwise require additional escaping.
This means that literals could for example prefer `'single quotes'`,
but use `"double 'em"` if necessary,
or `|'pipe' characters|` if the whole message is wrapped in `"quotes"` due to the host format
or if the literal value contains both `'` and `"` quotes.
```

exploration/quoted-literals.md
eemeli: That's `r6`, and it's included in the table.

eemeli: Oops, sorry, typo above: I meant `r5`, the "one way" requirement.

PR #480, opened 2023-09-19
Fix stability policy

eemeli:
When #472 was merged, it ended up including changes suggested by @aphillips in https://github.com/unicode-org/message-format-wg/pull/472#discussion_r1327996351 that had not been fully reviewed. This PR is about fixing issues related to that late addition.

Right now this is mostly editorial, and re-introducing a phrase that was dropped on merge:
> Later specification versions MAY make previously invalid messages valid.

However, we should also decide here what to do about this indented note that was included in the change:
> > Note: future versions may define new keywords.

For this to be valid, we need to find a way to make it play nice with this other bit of policy:
> Updates to this specification will not introduce message syntax that,
> when parsed according to earlier versions of this specification,
> would produce syntax or data model errors.

In other words, we need to come up with a syntax that allows for space for keywords to be _reserved_, so that a current parser won't syntax-error with them. At least thus far, each of our keywords associates with one of more subsequent _expressions_ or _patterns_.

This is problematic with a syntax like our current one, which uses `{` at the top level as both an _expression_ and _pattern_ start, depending on the context. Essentially, it's not possible to determine if `keyword {foo|bar}` is valid without knowing if it's meant to be processed as an expression or a pattern.

With a syntax as proposed in [this alternative](https://github.com/unicode-org/message-format-wg/blob/text-vs-code/exploration/0474-text-vs-code.md#start-in-text-encapsulate-code), this problem becomes at least tractable.

This note should not be left as is; either it needs to be dropped, or it needs to be made explicitly normative.

### DIFF
--- a/spec/README.md
+++ b/spec/README.md
@@ -95,6 +95,8 @@ Additional options or option values MAY be defined.
 > the functions for date formatting, number formatting and so on
 > will change their results over time.
 
+Later specification versions MAY make previously invalid messages valid.
+
 Updates to this specification will not introduce message syntax that,
 when parsed according to earlier versions of this specification,
 would produce syntax or data model errors.
--- a/spec/README.md
+++ b/spec/README.md
@@ -70,26 +70,24 @@ A reference to a _term_ looks like this.
 
 ### Stability Policy
 
-Updates to this specification MUST NOT change
+Updates to this specification will not change
 the syntactical meaning, the runtime output, or other behaviour
 of valid messages written for earlier versions of this specification
 that only use functions and expression attributes defined in this specification.
-Updates to this specification will not remove any syntax provided
-in this version. Future versions MAY add additional structure or meaning
-to existing syntax.
+Updates to this specification will not remove any syntax provided in this version.
+Future versions MAY add additional structure or meaning to existing syntax.
 
-Updates to this specification will not remove any reserved keywords
-or sigils.
+Updates to this specification will not remove any reserved keywords or sigils.
 
-> Note: future versions may defined new keywords
+> Note: future versions may define new keywords.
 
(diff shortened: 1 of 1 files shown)


aphillips (approved):
LGTM. Thanks for getting back to this before I did.

eemeli:
This bit from my earlier comment should be addressed:
> This note should not be left as is; either it needs to be dropped, or it needs to be made explicitly normative.

aphillips:
@eemeli Agreed. Note that #482 exists now for the syntax part.

It is possible to define a keyword syntax that is forward compatible parsable but whose meaning is unknown. Is it worth the effort? I think it might be, since "never" is a long time and many languages have created new keywords later (sometimes much later). However, this will require greater work on the syntax, since that usually involves the concept of a "statement". 

In the discussion of text-first syntaxes, to choose an example, starting a statement with `#` and ending the expression close might do that for declarations, with `match` being an outlier or potential outlier:

```
#input {something}
#local (something}
#unknown {something}
#match {expr} {expr}
when foo foo {still in the match...}
when bar bar {still in the match...}
when * * {still in the match...}
// implicit close
```

(If I had to guess what we might add, a likely candidate would probably be a loop construct or iterator, such as when building a `ul` or `ol` in HTML. This suggests a more complex syntax closer to `match` than to a declaration.)

So, net net...

I think the stability guarantee could apply only to syntax or at least that we should split this into two policy items:

>* Updates to this specification will not introduce message syntax that would be 
>   considered syntax errors in earlier versions of this specification
>* Updates to this specification will not introduce message syntax that makes
>   existing data model items invalid.

I don't think we can guarantee stability of the data model as the current text has it: that would prevent us from ever adding something that isn't in the current data model. That's different from "existing syntax items won't produce something different".

PR #481, opened 2023-09-19
style: Prefer single quotes

eemeli:
Update our Prettier styles to prefer single quotes, as requested in https://github.com/unicode-org/message-format-wg/pull/477#discussion_r1328885090.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,9 @@
   "dependencies": {
     "prettier": "^3.0.1"
   },
+  "prettier": {
+    "singleQuote": true
+  },
   "scripts": {
     "prettier": "prettier --write docs/ exploration/ spec/"
   }
--- a/spec/data-model/README.md
+++ b/spec/data-model/README.md
@@ -29,13 +29,13 @@ A message without _selectors_ and with a single _pattern_ is represented by a `P
 type Message = PatternMessage | SelectMessage;
 
 interface PatternMessage {
-  type: "message";
+  type: 'message';
   declarations: Declaration[];
   pattern: Pattern;
 }
 
 interface SelectMessage {
-  type: "select";
+  type: 'select';
   declarations: Declaration[];
   selectors: Expression[];
   variants: Variant[];
@@ -66,7 +66,7 @@ interface Variant {
 }
 
 interface CatchallKey {
-  type: "*";
+  type: '*';
   value?: string;
 }
 ```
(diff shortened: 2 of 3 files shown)


eemeli:
Requesting fast-tracking.

stasm (changes requested):
Can we configure Prettier to instead not touch code blocks at all? We may want to use them to describe examples that are explicitly not well formatted. I'd like Prettier to not get in the way.

eemeli:
> Can we configure Prettier to instead not touch code blocks at all? We may want to use them to describe examples that are explicitly not well formatted. I'd like Prettier to not get in the way.

You can do so in your markdown file by prepending a comment before the code block: https://prettier.io/docs/en/ignore#markdown

aphillips:
> You can do so in your markdown file by prepending a comment before the code block

This is about our repo's configuration. I'm with @stasm: can we make it just ignore the code blocks? Switching to single-quotes is great for JS examples---not so much for other languages (which we also quote sometimes)

eemeli:
There isn't a way to ignore all markdown code blocks, but the link I mention above shows how to skip prettifying specific cases.

Including a minimal config (as added here) will also make Prettier use that explicitly, and not e.g. whatever a user has configured as their default in an editor plugin.

stasm:
I know you care about this, @eemeli, and that you put effort to configure the bot, but I continue to have doubts about it, and this issue showcases another friction point that I think we could simply not have.

I understand using Prettier for code, but in my mind there's less utility for prose, even when written in Markdown. Prose is free-form and sometimes will not look exactly like the linter wants it to look, and that's OK. Humans are good at parsing ambiguous input :)

What problems does the bot solve? I browsed through some of its recent commits, and it mostly removes accidental trailing spaces, and converts my bullet points from `*` to `-`. Are these actually harmful?

OTOH, the bot adds noise to the PR, and breaks the flow of pushing to the branch while still working on it via local edits. I now have to stash, push, pull, pop stash, for arguably little benefit.

And now, as we've seen in https://github.com/unicode-org/message-format-wg/pull/477#discussion_r1328885090, the bot also overwrites an example whose purpose to use a different style.

I can learn to live with this, but I guess I'm going to also be a bit grumpy every time I hit a merge conflict between my pushes.

aphillips:
I'm with @stasm. This is generating a lot of work/noise/turbulence for not much noticeable benefit. I'm a big believer in standardized formatting, but this tool is consuming a lot of cycles that could better be applied to the many work items we have. If I promise to practice using `-` for bullets, can we live without it?!? 😿

eemeli:
Closing in favour of #491.

PR #485, opened 2023-09-25
Beauty Contest for Various Syntaxes

aphillips:
This was discussed in the 2023-09-25 call.

This is creating a scratch pad for different syntax options for starting in text mode, with an eye towards choosing one of them to proceed. The goal here is to choose an "operating theory" of what kinds of messages users want to write, so the choice of sigils and keywords and such are mutable. Bring what you've learned in that past few years of working on this problem.


exploration/0483-beauty-contest.md
```diff
@@ -0,0 +1,129 @@
...
+uses we have now (the other use is for patterns). This requires escaping whitespace or using
+a placeholder for it.
+
+The sigil `#` was chosen because `#define` type constructs are fairly common. Introduces `[`/`]` for keys.
+
+```
+#{input $var :function option=value}
+Hello {$var}
+
+#{input $var :function option=value}
```
stasm, line 83: For consistency with `#local` and `#match`, this could be:

```suggestion
#input {$var :function option=value}
Hello {$var}

#input {$var :function option=value}
```

stasm: (Same for the single-line version.)

exploration/0483-beauty-contest.md
```diff
@@ -0,0 +1,282 @@
...
+
+This is the current syntax.
+The list of messages in this section also serves as the basis for all other examples.
+
+```
+{Hello world!}
+
+{Hello {$user}}
+
+input $var :function option=value
```
eemeli, line 16:
```suggestion
input {$var :function option=value}
```

exploration/0483-beauty-contest.md
```diff
@@ -0,0 +1,282 @@
...
+{Hello {$var}}
+
+input $var :function option=value
```
eemeli, line 19:
```suggestion
input {$var :function option=value}
```

exploration/0483-beauty-contest.md
```diff
@@ -0,0 +1,282 @@
...
+match {$foo} {$bar}
+when foo bar {Hello {$foo} you have a {$var}}
+when * * {{$foo} hello you have a {$var}}
+
+match {$foo :function option=value} {$bar :function option=value}
+when a b {  {$foo} is {$bar}  }
+when x y {  {$foo} is {$bar}  }
+when * * {  {$foo} is {$bar}  }
+
+input $var :function option=value local $foo = {$bar :function option=value}{Hello {$var}, you have a {$foo}}
```
eemeli, line 32:
```suggestion
input {$var :function option=value} local $foo = {$bar :function option=value}{Hello {$var}, you have a {$foo}}
```

exploration/0483-beauty-contest.md
```diff
@@ -0,0 +1,282 @@
...
+Hello {$var}, you have a {$foo}
+
+{match {$foo} {$bar}}
+{when foo bar} Hello {$foo} you have a {$var}
+{when * *} {$foo} hello you have a {$var}
+
+{match {$foo :function option=value} {$bar :function option=value}}
+{when a b} {  {$foo} is {$bar}  }
+{when x y} {  {$foo} is {$bar}  }
+{when * *} {|  |}{$foo} is {$bar}{|  |}
```
eemeli, line 62:
```suggestion
{when a b} {||}  {$foo} is {$bar}  {||}
{when x y} {||}  {$foo} is {$bar}  {||}
{when * *} {||}  {$foo} is {$bar}  {||}
```

exploration/0483-beauty-contest.md
```diff
@@ -0,0 +1,282 @@
...
+
+{input $var :function option=value}{local $foo = $bar :function option=value}Hello {$var}, you have a {$foo}
+
+{match {$foo} {$bar}}{when foo bar} Hello {$foo} you have a {$var}{when * *} {$foo} hello you have a {$var}
+
+{match {$foo :function option=value}{$bar :function option=value}}{when a b} {  {$foo} is {$bar}  }{when x y} {  {$foo} is {$bar}  }{when * *} {|  |}{$foo} is {$bar}{|  |}
```
eemeli, line 68:
```suggestion
{match {$foo :function option=value}{$bar :function option=value}}{when a b}{||}  {$foo} is {$bar}  {||}{when x y}{||}  {$foo} is {$bar}  {||}{when * *}{||}  {$foo} is {$bar}  {||}
```

exploration/0483-beauty-contest.md
```diff
@@ -0,0 +1,282 @@
...
+{match {$foo} {$bar}}{when foo bar} Hello {$foo} you have a {$var}{when * *}{{$foo} hello you have a {$var}}
+
+{match {$foo :function option=value}{$bar :function option=value}}{when a b} {  {$foo} is {$bar}  }{when x y} {  {$foo} is {$bar}  }
+```
+
+## 3. Use sigils for code mode
+
+Try to redues the use of `{`/`}` to just expressions and placeholders instead of the three
+uses we have now (the other use is for patterns). This requires escaping whitespace or using
+a placeholder for it. See #486 for a discussion of whitespace options.
```
eemeli, line 108:
```suggestion
Try to reduce the use of `{`/`}` to just expressions and placeholders
instead of the three uses we have now (the other use is for patterns).
This requires escaping whitespace or using a placeholder for it.
See [#487](https://github.com/unicode-org/message-format-wg/pull/487)
for a discussion of whitespace options.
```

exploration/0483-beauty-contest.md
```diff
@@ -0,0 +1,282 @@
...
+input $var :function option=value local $foo = {$bar :function option=value}{Hello {$var}, you have a {$foo}}
+
+match {$foo} {$bar} when foo bar {Hello {$foo} you have a {$var}} when * * {{$foo} hello you have a {$var}}
+
+match {$foo :function option=value} {$bar :function option=value}when a b {  {$foo} is {$bar}  }when x y {  {$foo} is {$bar}  }when * * {  {$foo} is {$bar}  }
+```
+
+## 1. Invert for Text Mode
+
+Consumes exterior whitespace.
```
eemeli, line 41:
```suggestion
Consumes exterior whitespace,
requiring leading/trailing whitespace in a pattern to use `{||}` as a terminal marker,
or to explicitly escape the whitespace as `{| |}`.
Drops the use of `{}` as pattern delimiters.
```

exploration/0483-beauty-contest.md
```diff
@@ -0,0 +1,282 @@
...
+#input {$var :function option=value}
+Hello {$var}
+
+#input {$var :function option=value}
+#local $foo = {$bar :function option=value}
+Hello {$var}, you have a {$foo}
+
+#match {$foo} {$bar}
+#when[foo bar] Hello {$foo} you have a {$var}
+#when[  *   *] {$foo} hello you have a {$var}
```
eemeli, line 123: This would require escaping `#` when used as a pattern character.

exploration/0483-beauty-contest.md
```diff
@@ -0,0 +1,282 @@
...
+#$var :function option=value
+Hello {$var}
+
+#$var :function option=value
+#$foo = {$bar :function option=value}
+Hello {$var}, you have a {$foo}
+
+?? {$foo} {$bar}
+::[ foo bar] Hello {$foo} you have a {$var}
+::[ *     *] {$foo} hello you have a {$var}
```
eemeli, line 155: This would require escaping `::[` when used as a pattern sequence.

exploration/0483-beauty-contest.md
```diff
@@ -0,0 +1,282 @@
...
+   input {$var :function option=value}
+   local $foo = {$bar :function option=value}
+}
+Hello {$var}, you have a {$foo}
+
+{
+  match {$foo} {$bar}
+  [ foo bar] Hello {$foo} you have a {$var}
+  [ *     *] {$foo} hello you have a {$var}
+}
```
eemeli, line 199: This would require escaping `[` when used as a pattern character.

exploration/0483-beauty-contest.md
```diff
@@ -0,0 +1,282 @@
...
+
+{# input $var :function option=value
+   local $foo = {$bar :function option=value}
+}
+Hello {$var}, you have a {$foo}
+
+#match {$foo :function option=value} {$bar :function option=value}
+[a b] {$foo} is {$bar}
+[x y] {$foo} is {$bar}
+[* *] {$foo} is {$bar}
```
eemeli, line 237: This would require escaping `[` when used as a pattern character.

exploration/0483-beauty-contest.md
```diff
@@ -0,0 +1,282 @@
...
+Hello {$var}
+
+#input {$var :function option=value};
+#local $foo = {$bar :function option=value};
+Hello {$var}, you have a {$foo}
+
+#match {$foo} {$bar}
+when [ foo bar] Hello {$foo} you have a {$var}
+when [ *     *] {$foo} hello you have a {$var}
+;
```
eemeli, line 268: This would require escaping `[` when used as a pattern character.

eemeli (changes requested):
Not sure if this is still relevant, as I missed the last call, but added some fixes and clarifications at least for posterity's sake.

I think 3, 5, 5a, and 6 are all unworkable as currently presented, as they would require escaping `#` or `[` in addition to `{}` in patterns. If that's an additional cost we're willing to consider, or if there's a way to mitigate that cost, it should be explicitly noted here.

Only the line comments with suggestions are intended as blocking here.

exploration/0483-beauty-contest.md
aphillips: Any additions to the sigil set would require attendant changes to the escaping rules.

exploration/0483-beauty-contest.md
aphillips: That's correct. Doubling the `:` was intentional for that reason. Note that `::` and `[` might be separate constructs.

eemeli (changes requested):
@aphillips I added Option 1a directly on the branch, and fixed the Option 0 syntax.

I've also ported and expanded on my previous line comments that got obsoleted by the file name change. All of the syntax proposals except for 1a have requirements that need to be taken into account when deciding between them. For example, needing to escape `[` or `;` in any pattern is a pretty significant requirement.

I also spotted a couple of inconsistencies in the examples; suggested fixes for those are also included.

exploration/beauty-contest.md
```diff
@@ -0,0 +1,316 @@
...
+
+match {$foo} {$bar} when foo bar {Hello {$foo} you have a {$var}} when * * {{$foo} hello you have a {$var}}
+
+match {$foo :function option=value} {$bar :function option=value}when a b {  {$foo} is {$bar}  }when x y {  {$foo} is {$bar}  }when * * {  {$foo} is {$bar}  }
+```
+
+## 1. Invert for Text Mode
+
+Consumes exterior whitespace.
+
```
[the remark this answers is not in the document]
aphillips: I don't think this is a correct statement. In fact, you can see unquoted literal strings in all of the `when` clause examples. Can you explain what you mean?

exploration/beauty-contest.md
```diff
@@ -0,0 +1,316 @@
...
+{#match {$foo :function option=value}{$bar :function option=value}} {#when a b} {{  {$foo} is {$bar}  }} {#when x y} {{  {$foo} is {$bar}  }} {#when * *} {|  |}{$foo} is {$bar}{|  |}
+```
+
+## 2. Text First, but Always Code After Code-Mode
+
+This is @mihnita's proposal, mentioned in the 2023-10-02 call.
+Start in text mode, but stay in code mode once you get there.
+Adapted @eemeli's start-in-text-mode syntax for these examples.
+
```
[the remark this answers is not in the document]
aphillips: ibid.

exploration/beauty-contest.md
```diff
@@ -0,0 +1,316 @@
...
+```
+
+## 4. Reducing keywords
+
+Avoids keywords in favor of sigil based parsing.
+The theory here is that the syntactic sugar of `match` and `when` are nice the first time you use them
+but the benefit is lost or reduced after that.
+
+Uses double or paired sigils to reduce the need for escaping common characters (such as `?`)
+
```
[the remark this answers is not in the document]
aphillips:
```suggestion

Requires escaping `#` inside a pattern. 
Requires escaping `??` at the start of a pattern.
Requires escaping `::[` inside a pattern.
```

exploration/beauty-contest.md
```diff
@@ -0,0 +1,316 @@
...
+?? {$foo} {$bar}
+::[ foo bar] Hello {$foo} you have a {$var}
+::[ *     *] {$foo} hello you have a {$var}
+
+?? {$foo :function option=value} {$bar :function option=value}
+::[a b] {{  {$foo} is {$bar}  }}
+::[x y] {{  {$foo} is {$bar}  }}
+::[* *] {|  |}{$foo} is {$bar}{|  |}
+
+#$var :function option=value}#$foo = {$bar :function option=value}Hello {$var}, you have a {$foo}
```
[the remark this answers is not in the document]
aphillips: This is upside down. The example can't work without `{`/`}` because there would otherwise be no divider between an `input` and the pattern.

exploration/beauty-contest.md
```diff
@@ -0,0 +1,316 @@
...
+
+{#
+  match {$foo} {$bar}
+  [ foo bar] Hello {$foo} you have a {$var}
+  [ *     *] {$foo} hello you have a {$var}
+}
+
+{#
+   input $foo :function option=value
+}{
```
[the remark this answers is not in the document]
aphillips: The original theory was: "declaration block" and "body block". The body block only needs to be bracketed when it contains code (i.e. a `match`). There's no reason to have blocks if they aren't separate (that would just be option 2 then)

exploration/beauty-contest.md
```diff
@@ -0,0 +1,316 @@
...
+## 6. Use "statements"
+
+Many languages delimit statements using a terminator, such as `;`.
+In some languages, the terminator is the newline and lines can be extended using an escape like `\`.
+Here we use `;` as a terminator.
+Note that the closing `;` on `match` might be optional.
+
+Defining a statement syntax might allow us to introduce different types of selectors in future versions,
+such as looping constructs.
+
```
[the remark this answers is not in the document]
aphillips: This is the least thought out of the options. Escaping `;` is not a good idea, but I wanted to at least put a foot in the door for the idea of statements. A lot of the examples above need `}` to find the end of the statement (to separate it from an unquoted pattern). If we reduced the number of uses for `{`/`}`, what would that look like?

exploration/beauty-contest.md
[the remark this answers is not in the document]
eemeli: Ah, good point. I fixed the suggestion above to specify unquoted literal _operands_, which we can't have because they would conflict with e.g. `{input` and `{when`.

exploration/beauty-contest.md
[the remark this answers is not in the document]
eemeli: Also fixed as above.

exploration/beauty-contest.md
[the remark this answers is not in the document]
eemeli: Ah, ok. Then I think lines 257 and 259 may need fixing, because together with this we've got all of:
- `{ match`
- `{#match`
- `match`

I can't tell which of these three is the proposed syntax here.

exploration/beauty-contest.md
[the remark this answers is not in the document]
eemeli: The `;` escape is only required for the trailing one at the end of a `#match` message; dropping that would simplify this quite a bit.

I'm not much of a fan of this one, tbh. To see the value of the semicolons, I think the `#input` and `#local` should drop their {braces}, and it maybe should be used as a variant key list terminator as well. So something like this:

```
#match {$foo} {$bar};
#when foo bar; Hello {$foo} you have a {$var}
#when * *; {$foo} hello you have a {$var}
```

exploration/beauty-contest.md
```diff
@@ -0,0 +1,378 @@
...
+| 3      | Use sigils for code mode                                       | +               | --                        | +                              | +                                | +                | --             |
+| 3a     | Use sigils for code mode, use {/} for keys                     | +               | -                         | +                              | +                                | +                | -              |
+| 4      | Reducing keywords                                              | +               | ---                       | +                              | +                                | +                | ---            |
+| 5      | Use blocks for declarations and body                           | --              | +                         | +                              | -                                | -                | -              |
+| 5a     | Use blocks for declarations but sigils for selectors           | +               | -                         | +                              | -                                | -                | -              |
```
eemeli, line 20: This option requires escaping `[` when used as a pattern character, and the sequence `#match` when used at the start of a pattern.

```suggestion
| 5a     | Use blocks for declarations but sigils for selectors           | +               | --                        | +                              | -                                | -                | -              |
```

exploration/beauty-contest.md
```diff
@@ -0,0 +1,378 @@
...
+| 2      | Text first, but always code mode after code                    | -               | +                         | -                              | -                                | +                | +              |
+| 3      | Use sigils for code mode                                       | +               | --                        | +                              | +                                | +                | --             |
+| 3a     | Use sigils for code mode, use {/} for keys                     | +               | -                         | +                              | +                                | +                | -              |
+| 4      | Reducing keywords                                              | +               | ---                       | +                              | +                                | +                | ---            |
+| 5      | Use blocks for declarations and body                           | --              | +                         | +                              | -                                | -                | -              |
```
eemeli, line 19: Requires escaping `[` when used as a pattern character in an unquoted pattern.

```suggestion
| 5      | Use blocks for declarations and body                           | --              | -                         | +                              | -                                | -                | -              |
```

exploration/beauty-contest.md
```diff
@@ -0,0 +1,378 @@
...
+| 0      | Current syntax, starts in code mode                            | -               | +                         | --                             | +                                | -                | +              |
+| 1      | Invert for text mode                                           | -               | +                         | +                              | -                                | +                | +              |
+| 1a     | Invert for text mode, distinguish statements from placeholders | -               | +                         | +                              | +                                | +                | -              |
+| 2      | Text first, but always code mode after code                    | -               | +                         | -                              | -                                | +                | +              |
+| 3      | Use sigils for code mode                                       | +               | --                        | +                              | +                                | +                | --             |
```
eemeli, line 16: The escapes for 3 should be the same as 3a: `#` in patterns.

```suggestion
| 3      | Use sigils for code mode                                       | +               | -                         | +                              | +                                | +                | --             |
```

aphillips: 3 also introduces a special use for `[`/`]`. Presumably we could work around unquoted literals by excluding that character from unquoted, so I'll merge this

eemeli (approved):
I think this could be merged now, so that it can have better visibility.

aphillips:
Merging this in preparation for the stackranking exercise.

PR #491, opened 2023-10-10
Remove the Prettier push action

stasm:
Following the (short) discussion in https://github.com/unicode-org/message-format-wg/pull/481#issuecomment-1726101082 and then after the 2023-10-09 teleconference, we agreed to remove the Prettier push action, originally introduced in #451.


stasm:
@eemeli Is there anything else to remove or de-configure?

eemeli (approved):
Nah, that should do it.

alerque (approved):

PR #494, opened 2023-10-12
Add 3 more beauty contestants

stasm:
After some more thinking and offline discussions, I'd like to extend the beauty contest (#485) with the following 3 approaches:

* <del>`0a`  is a hybrid mode which allows for simple patterns to be unquoted (...)</del> Renamed to `2a`.

* `1b` is a variant of the text-first `1`, but with the syntax directly inspired by existing templating languages.

    ```
    {% match {$foo} {$bar} %}
    {% when foo bar %} Hello {{ $foo }} you have a {{ $var }}
    {% when * * %} {{ $foo }} hello you have a {{ $var }}
    ```

* `2a` is a hybrid mode which allows for simple patterns to be unquoted (`Hello!`), and switches to the "complex" mode when the message is wrapped in `{{ }}`. The complex mode then uses the current syntax verbatim. 

    ```
    {{ match {$foo} {$bar}
       when foo bar {Hello {$foo} you have a {$var}}
       when * * {{$foo} hello you have a {$var}}
    }}
    ```

* `3b` is a variant of keyword-based `3`, but it uses `#[...]` for all control-flow-related syntax.

    ```
    #[match {$foo} {$bar}]
    #[foo bar] Hello {$foo} you have a {$var}
    #[  *   *] {$foo} hello you have a {$var}
    ```


exploration/beauty-contest.md
```diff
@@ -10,11 +10,14 @@ Please dispute factual errors in this. The table uses multiple `-` when an optio
 | Option | Description                                                    | Doesn’t Nest {} | Doesn’t Need More Escapes | Doesn’t Require Quoted Pattern | Allows Unquoted Literal Operands | Counted {} works | Adds no sigils |
 | :----- | :------------------------------------------------------------- | :-------------- | :------------------------ | :----------------------------- | :------------------------------- | :--------------- | :------------- |
 | 0      | Current syntax, starts in code mode                            | -               | +                         | --                             | +                                | -                | +              |
+| 0a     | Hybrid syntax, starts in text mode                             | -               | +                         | -                              | +                                | -                | +              |
```
aphillips, line 13: As mentioned in slack, I think this should be 2a instead. 0 is start in code mode. This is a start-in-text-mode option

exploration/beauty-contest.md
```diff
@@ -121,6 +162,40 @@ Hello {$var}, you have a {$foo}
...
+{% local $foo = $bar :function option=value %}
+Hello {{ $var }}, you have a {{ $foo }}
+
+{% match {$foo} {$bar} %}
+{% when foo bar %} Hello {{ $foo }} you have a {{ $var }}
+{% when * * %} {{ $foo }} hello you have a {{ $var }}
+
+{% match {$foo :function option=value} {$bar :function option=value} %}
+{% when a b %} {|  {{$foo}} is {{$bar}}  |}
+{% when x y %} {|  {{$foo}} is {{$bar}}  |}
```
aphillips, line 189: These aren't right. They should be:

```suggestion
{% when x y % }}{||}  {{$foo}} is {{$bar}}  {||}
```

(I'm not sure how to type the `a b` example here, as you're using double-`{{` for variables. Quoting the whole pattern would have to be revisited as a topic. Conversely, I'm not sure we need to double the brackets for variables. I understand that you're being super-consistent with templating and so shouldn't read too much into it.

I'd probably add a comment about what I just said above to the description.

aphillips (changes requested):

exploration/beauty-contest.md
stasm: I chose to call it `0a` because in my mind it's our current syntax extended with the "simple message" mode. But I agree it's closely related to `2` in terms of how they transition between modes.

exploration/beauty-contest.md
stasm: No, this is intended (but we should discuss exact sigils). In this approach, all sigils are double characters: `{% %}` for statements, `{{ }}` for placeholders, and `{| |}` for quoted patterns. OTOH, an empty string literal in a placeholder would be `{{ || }}`.

aphillips (approved):

PR #495, opened 2023-10-15
Data-driven tests proposal

mradbourne:
A design document to gain consensus on the MF2 test suite, both content and runner.

### DIFF
--- a/exploration/data-driven-tests.md
+++ b/exploration/data-driven-tests.md
@@ -1,0 +1,234 @@
+# Data-driven tests
+
+Status: **Proposed**
+
+<details>
+	<summary>Metadata</summary>
+	<dl>
+		<dt>Contributors</dt>
+		<dd>@mradbourne</dd>
+		<dt>First proposed</dt>
+		<dd>2023-09-28</dd>
+		<dt>Pull Request</dt>
+		<dd>#000</dd>
+	</dl>
+</details>
+
+## Objective
+
+Establish a suitable approach for testing that different implmentations of MessageFormat 2 align with the specification.
+
+This approach includes _how_ tests are written: They should be captured in a single platform-agnostic format that can be utilized by all MF2 implementations. There should be no need to rewrite individual test cases for each platform.
+
+This approach also includes _what_ kind of tests are written. We need to identify which parts of MF2 should be covered by different types of test as a minimum.
+
+## Background
+
+Several pre-existing test files have been considered before forming this proposal:
+
+- [Unicode's Data Driven Test framework](https://github.com/unicode-org/conformance)
+- [message-format-wg XML test format](https://github.com/unicode-org/message-format-wg/tree/514758923abac13a2c5eb71b6b6cdef4a181280e/test)
+- [Intl.MessageFormat polyfill tests](https://github.com/messageformat/messageformat/tree/main/packages/mf2-messageformat/src)
+- [ICU4J test examples](https://github.com/unicode-org/icu/tree/main/icu4j/main/core/src/test/java/com/ibm/icu/dev/test/message2)
+- [Tim Chevalier's draft ICU4C tests](https://github.com/catamorphism/icu/blob/parser-plus-data-model-plus-full-api/icu4c/source/test/intltest/messageformat2test.cpp)
+
+## Use-Cases
+
+The main platforms for which the tests should initially run are:
+- Node.js
+- ICU4J (Java)
+- ICU4C (C++)
+
+Other platforms, such as ICU4X (Rust) may be added later.
+
+## Requirements
+
+- Test framework
+  - The test cases and assertions must be easy to read.
+  - The test cases and assertions must be completely platform-agnostic.
+  - The framework must include the platform-specific test executors as part of the solution.
+  - The framework must be extendable with new executors (e.g. ICU4X) and it should be clear how to do this. 
+
+- Test content
+  - Syntax tests: Testing which standard registry functions are invoked when different arguments are passed to a given pattern.
+  - Formatter tests: Testing the output of each of the standard registry's formatting functions.
+  - Selector tests: Testing which case of a `match` statement is selected, based on what follows the `when` keyword.
+
+## Constraints
+
+### External dependencies
+
+Due to the platform-agnostic nature of the tests, we must ensure that they are easily adaptable to new technologies. This means that any external dependency must be introduced with a great deal of caution because it may impact the portability of the test framework.
+
+### Data model is not part of the specification
+
+Although a standard data model is included in this repository, there is no requirement for all MF2 implementation to use it. This means that any data model tests included in the test suite may fail for standard-compliant implementations. If any tests of this type are included, they must be optional.
+
+### Errors and evaluation strategy
+
+It is important to test error cases for each of the test types mentioned above but, because variable evaluation is not captured within the standard, we cannot guarantee what kind of error will be raised in all cases.
+
+For example, the pattern below may or may not result in an error depending on how lazily the expression is evaluated. This presents a challenge for testing.
+
+```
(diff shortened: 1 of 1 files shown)


exploration/data-driven-tests.md
```diff
@@ -0,0 +1,243 @@
+# Data-driven tests
+
+Status: **Proposed**
```
eemeli, line 3:
```suggestion
Status: **Accepted**
```

eemeli (approved):
Looks good. I don't see any reason why not proceed with this directly.

exploration/data-driven-tests.md
```diff
@@ -0,0 +1,243 @@
...
+
+  Scenario: One item - male
+    When the message is resolved with params:
+      | key        | value     |
+      | photoCount |         1 |
+      | userGender | masculine |
+    Then the string output is "Matt added a new photo to his album."
+```
+
+The [Cucumber framework](https://cucumber.io/) was considered because of its integration with the Gherkin syntax. Cucumber's approach of using platform-specific step definitions for Gherkin scenarios aligns with our goal of having a data-only representation of the test content. Unfortunately, the MF2 test runner may need to target a wide range of platforms in future. In this case, integrating with Cucumber would incur additional overhead.
```
ryzokuken, line 241: At the risk of repeating @stasm from above, another challenge could be integrating cucumber with existing, possibly even legacy platforms.

exploration/data-driven-tests.md
```diff
@@ -0,0 +1,243 @@
...
+		<dt>First proposed</dt>
+		<dd>2023-09-28</dd>
+		<dt>Pull Request</dt>
+		<dd>#495</dd>
+	</dl>
+</details>
+
+## Objective
+
+Establish a suitable approach for testing that different implmentations of MessageFormat 2 align with the specification.
```
catamorphism, line 19:
```suggestion
Establish a suitable approach for testing that different implementations of MessageFormat 2 align with the specification.
```

exploration/data-driven-tests.md
```diff
@@ -0,0 +1,243 @@
...
+
+- Test framework
+
+  - The test cases and assertions must be easy to read.
+  - The test cases and assertions must be completely platform-agnostic.
+  - The framework must include the platform-specific test executors as part of the solution.
+  - The framework must be extendable with new executors (e.g. ICU4X) and it should be clear how to do this.
+
+- Test content
+  - Syntax tests: Testing which standard registry functions are invoked when different arguments are passed to a given pattern.
```
catamorphism, line 55: I'm not sure this is the right description of "syntax tests"? Patterns might not invoke any functions, but we would still want to test that correct patterns don't generate syntax errors and that incorrect patterns generate the right syntax errors.

exploration/data-driven-tests.md
```diff
@@ -0,0 +1,243 @@
...
+
+```jsonc
+{
+  "label": "Skeleton affects datetime format",
+  "locale": "en-US",
+  "pattern": "{$givenDateTime :datetime skeleton=yMMMdE}",
+  "inputs": {
+    "givenDateTime": { "type": "datetime", "value": "2000-12-31T00:00:00.000Z" }
+  }
+  // "verify":  "Sun, 31 Dec 2000"
```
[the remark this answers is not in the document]
mradbourne: Makes sense. I'll mark as optional.
If we assert on something more abstract than the output string, I don't know if there's enough value-add over the "syntax" tests though. It would be good to discuss.

exploration/data-driven-tests.md
```diff
@@ -0,0 +1,243 @@
...
+
+These are extensive tests of the cases within a `match` statement. Testing of multiple selectors is included.
+
+Example:
+
+```jsonc
+{
+  "label": "Matches numbers other than one",
+  "locale": "en-US",
+  "pattern": "match {$arg :number} when one {one} when * {*}",
```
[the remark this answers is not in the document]
mradbourne: Yes, the output string value are matching purely as a convention. Now you've mentioned it, I think this approach could be misleading - I'll change them to avoid the need for more explanation. I'll also add a multi-selector example 👍

exploration/data-driven-tests.md
mradbourne: Agreed. I'll reword accordingly.

exploration/data-driven-tests.md
```diff
@@ -0,0 +1,243 @@
...
+
+## Objective
+
+Establish a suitable approach for testing that different implmentations of MessageFormat 2 align with the specification.
+
+This approach includes _how_ tests are written: They should be captured in a single platform-agnostic format that can be utilized by all MF2 implementations. There should be no need to rewrite individual test cases for each platform.
+
+This approach also includes _what_ kind of tests are written. We need to identify which parts of MF2 should be covered by different types of test as a minimum.
+
+## Background
```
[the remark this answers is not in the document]
mradbourne: Added to the 'Objective' section 👍

exploration/data-driven-tests.md
```diff
@@ -0,0 +1,243 @@
...
+
+Several pre-existing test files have been considered before forming this proposal:
+
+- [Unicode's Data Driven Test framework](https://github.com/unicode-org/conformance)
+- [message-format-wg XML test format](https://github.com/unicode-org/message-format-wg/tree/514758923abac13a2c5eb71b6b6cdef4a181280e/test)
+- [Intl.MessageFormat polyfill tests](https://github.com/messageformat/messageformat/tree/main/packages/mf2-messageformat/src)
+- [ICU4J test examples](https://github.com/unicode-org/icu/tree/main/icu4j/main/core/src/test/java/com/ibm/icu/dev/test/message2)
+- [Tim Chevalier's draft ICU4C tests](https://github.com/catamorphism/icu/blob/parser-plus-data-model-plus-full-api/icu4c/source/test/intltest/messageformat2test.cpp)
```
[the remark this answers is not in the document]
mradbourne: Good idea. I've added an explanatory sentence to each bullet explaining how it influences the proposal.

exploration/data-driven-tests.md
mradbourne: Good point. I've updated accordingly.

exploration/data-driven-tests.md
```diff
@@ -0,0 +1,301 @@
...
+```jsonc
+{
+  "label": "Renders multiple inputs in formatted string",
+  "locale": "en-US",
+  "pattern": "{{$strArg :string} and {$numArg :number minimumFractionDigits=2}}",
+  "inputs": {
+    "strArg": { "type": "string", "value": "foo" },
+    "numArg": { "type": "number", "value": 123 }
+  }
+  // "verify":  "{ formatter: "string", value: "foo" } and { formatter: number, value: 123, minimumFractionDigits: 2 }"
```
stasm, line 193: Option values are strings, even if they look like numbers.

```suggestion
  // "verify":  "{ formatter: "string", value: "foo" } and { formatter: "number", value: 123, minimumFractionDigits: "2" }"
```

exploration/data-driven-tests.md
```diff
@@ -0,0 +1,243 @@
...
+
+These tests focus on the standard registry's formatters (e.g. `:number`, `:datetime`). They cover the different options that can be passed to each formatter (e.g. `offset`, `skeleton`).
+
+Example:
+
+```jsonc
+{
+  "label": "Skeleton affects datetime format",
+  "locale": "en-US",
+  "pattern": "{$givenDateTime :datetime skeleton=yMMMdE}",
```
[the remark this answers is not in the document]
stasm: With the latest changes to the syntax, which now starts in text mode, the pattern defined in this test is now valid syntax, making my review comment obsolete.

stasm (approved):

PR #500, opened 2023-10-23
Implement text-mode-first syntax

aphillips:
Following the 2023-10-23 teleconference, implement "option 2a with additional ugliness".

### DIFF
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -11,10 +11,10 @@ body = quoted-pattern
      / (selectors 1*([s] variant))
 
 quoted-pattern = "{{" pattern "}}"
-pattern        = *(text / expression)
-selectors      = match 1*([s] expression)
-variant        = when 1*(s key) [s] quoted-pattern
-key            = literal / "*"
+pattern = *(text / expression)
+selectors = match 1*([s] expression)
+variant = when 1*(s key) [s] quoted-pattern
+key = literal / "*"
 
 expression = literal-expression / variable-expression / function-expression
 literal-expression = "{" [s] literal [s annotation] [s] "}"
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -1,13 +1,16 @@
-message = [s] *(declaration [s]) body [s]
+message = pattern / complex-message
+
+complex-message = "{{" [s] *(declaration [s]) body [s] "}}"
 
 declaration = let s variable [s] "=" [s] expression
-body = pattern
+body = quoted-pattern
      / (selectors 1*([s] variant))
 
-pattern = "{" *(text / expression) "}"
-selectors = match 1*([s] expression)
-variant = when 1*(s key) [s] pattern
-key = literal / "*"
+quoted-pattern = "{{" pattern "}}"
+pattern        = *(text / expression)
+selectors      = match 1*([s] expression)
+variant        = when 1*(s key) [s] quoted-pattern
+key            = literal / "*"
 
 expression = "{" [s] ((operand [s annotation]) / annotation) [s] "}"
 operand    = literal / variable
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -178,7 +178,7 @@ Unless there is an error, resolving a _message_ always results in the formatting
 of a single _pattern_.
 
 ```abnf
-pattern = "{" *(text / expression) "}"
+pattern = *(text / expression)
 ```
 A _pattern_ MAY be empty.
 
@@ -190,6 +190,10 @@ during the formatting process.
 A **_<dfn>quoted pattern</dfn>_** is a pattern that is enclosed by
 starting with a sequence of two U+007B LEFT CURLY BRACKET `{{` 
 and ending with a sequence of two U+007D RIGHT CURLY BRACKET `}}`.
+
+```abnf
+quoted-pattern = "{{" pattern "}}"
+```
 
 A _quoted pattern_ MAY be empty.
 
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -163,7 +163,7 @@ local-declaration = local s variable [s] "=" [s] expression
 
 ### Body
 
-The **_<dfn>body</dfn>_** of a _message_ is the part that will be formatted.
+The **_<dfn>body</dfn>_** of a _complex message_ is the part that will be formatted.
 The _body_ consists of either a _quoted pattern_ or a _matcher_.
 
 ```abnf
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -123,7 +123,13 @@ A **_<dfn>message</dfn>_** is the complete template for a specific message forma
 > >
 > > An exception to this is: whitespace inside a _pattern_ is **always** significant.
 
-A _message_ consists of two parts:
+A _message_ can be a _pattern_ or it can be a _complex message_.
+
+A **_<dfn>complex message</dfn>_** is any _message_ that contains _declarations_,
+a _matcher_, or both.
+A _complex message_ always begins with the the sequence `{{`
+and is terminated by the sequence `}}`
+and consists of:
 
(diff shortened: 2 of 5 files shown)


spec/message.abnf
```diff
@@ -1,13 +1,16 @@
-message = [s] *(declaration [s]) body [s]
+message = pattern / complex-message
+
+complex-message = "{{" [s] *(declaration [s]) body [s] "}}"
 
 declaration = let s variable [s] "=" [s] expression
-body = pattern
+body = quoted-pattern
      / (selectors 1*([s] variant))
```
aphillips, line 8: Note that this makes the following a valid message:

```
{{{{I am a message}}}}
```

We could change `body` to be just:

```suggestion
body = selectors 1*([s] variant)
```

spec/message.abnf
```diff
@@ -1,13 +1,16 @@
...
 
-pattern = "{" *(text / expression) "}"
-selectors = match 1*([s] expression)
-variant = when 1*(s key) [s] pattern
-key = literal / "*"
+quoted-pattern = "{{" pattern "}}"
+pattern        = *(text / expression)
+selectors      = match 1*([s] expression)
+variant        = when 1*(s key) [s] quoted-pattern
+key            = literal / "*"
```
eemeli, line 13: It would probably be easier to not align the `=` on adjacent lines, especially as we're looking to make further changes to the syntax.

spec/syntax.md
```diff
@@ -397,9 +394,9 @@ Additionally, an _input-declaration_ can contain a _variable-expression_.
 > Placeholders:
 >
 > ```
-> {This placeholder contains an {|expression with a literal|}}
-> {This placeholder references a {$variable}}
-> {This placeholder references a function on a variable: {$variable :function with=options}}
+> {{This placeholder contains an {|expression with a literal|}}}
+> {{This placeholder references a {$variable}}}
+> {{This placeholder references a function on a variable: {$variable :function with=options}}}
```
eemeli, line 399: No need to quote these patterns; better keep them as valid messages.

```suggestion
> This placeholder contains an {|expression with a literal|}
> This placeholder references a {$variable}
> This placeholder references a function on a variable: {$variable :function with=options}
```

aphillips: Ugh. I thought I got this one...

spec/syntax.md
```diff
@@ -156,56 +162,41 @@ local-declaration = local s variable [s] "=" [s] expression
 ### Body
 
 The **_<dfn>body</dfn>_** of a _message_ is the part that will be formatted.
```
eemeli, line 164:
```suggestion
The **_<dfn>body</dfn>_** of a _complex message_ is the part that will be formatted.
```

spec/syntax.md
```diff
@@ -156,56 +162,41 @@ local-declaration = local s variable [s] "=" [s] expression
...
-
 ## Pattern
 
 A **_<dfn>pattern</dfn>_** contains a sequence of _text_ and _placeholders_ to be formatted as a unit.
-All _patterns_ begin with U+007B LEFT CURLY BRACKET `{` and end with U+007D RIGHT CURLY BRACKET `}`.
 Unless there is an error, resolving a _message_ always results in the formatting
 of a single _pattern_.
 
 ```abnf
 pattern = "{" *(text / expression) "}"
```
eemeli, line 179:
```suggestion
pattern = *(text / expression)
```

spec/syntax.md
```diff
@@ -156,56 +162,41 @@ local-declaration = local s variable [s] "=" [s] expression
...
-> An empty _pattern_:
+A _pattern_ MAY contain an arbitrary number of _placeholders_ to be evaluated
+during the formatting process.
+
+### Quoted Pattern
+
+A **_<dfn>quoted pattern</dfn>_** is a pattern that is enclosed by
+starting with a sequence of two U+007B LEFT CURLY BRACKET `{{` 
+and ending with a sequence of two U+007D RIGHT CURLY BRACKET `}}`.
+
```
eemeli, line 193:
````suggestion

```abnf
quoted-pattern = "{{" pattern "}}"
```

````

**Review by eemeli — APPROVED** [review]
I think that might be everything? So looks good.

For anyone coming to this without context and wondering what the **** we're thinking with a syntax that sometimes ends with `}}}}}`:

We agreed on today's call to resolve the syntax blockage we'd gotten ourselves in by picking a suitably ugly version of one of the variants under consideration, so that we'd have a baseline that at least nominally fulfills our preceding consensus decisions on not delimiting simple patterns and using `{{...}}` for complex patterns.

This is **_not_** intended as the final MF2 syntax, but as a step towards that. It **_will_** be iterated on.

With this to work from, we're now focusing on revisiting our [consensus](https://github.com/unicode-org/message-format-wg/blob/main/exploration/pattern-exterior-whitespace.md) on whether to require non-simple patterns to always be quoted, or to make that optional.

**Review by stasm — APPROVED** [review]

**Review by echeran — COMMENTED** [review]
@mihnita Can you double check the `.properties` file examples for correctness?

Apart from that, everything else LGTM. Hopefully, we're all satisfied in the ugliness that has been achieved.

File: spec/syntax.md [hunk 7]
```diff
@@ -216,17 +214,20 @@ respectively.
...
 
 > **Example**
 > In a Java `.properties` file, the values `hello` and `hello2` both contain
 > an identical _message_ which consists of a single _pattern_.
 > This _pattern_ consists of _text_ with exactly three spaces before and after the word "Hello":
 >
 > ```properties
-> hello = {   Hello   }
-> hello2={   Hello   }
+> hello = {{{{   Hello   }}}}
```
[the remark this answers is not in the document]
aphillips: To get the spaces in a properties file, the spaces need to be quoted.

It is possible to do this in the `.properties` syntax:

```
hello = \   Hello  \ 
```

My change illustrates using one method to quote the spaces using MF2's syntax. A different example would be:

```
hello  = {||}   Hello   {||}
hello2 = {|   |}Hello{|   |}
```

spec/syntax.md
```diff
@@ -216,17 +214,20 @@ respectively.
...
+> hello2=\   Hello  \
```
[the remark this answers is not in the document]
aphillips: LOL. I already had a hello2 I guess...

The space trimming here is an artifact of .properties, not of MF2. The problem is to get intentional spaces to the MF2 formatter. The `hello2` example shows how to do that in a properties file (a problem that already exists for users of this format--look for it on stackoverflow, for example).

From an MF2 perspective, `   Hello   ` is perfectly valid and clear about space handling. We don't need to solve every format's way of protecting the spaces in order for MF2 to be done. As you point out, the old syntax did guard against that... at the cost of making the message `Hello` invalid.

spec/syntax.md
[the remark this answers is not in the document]
echeran: I had to stare at [examples of .properties entries](https://en.wikipedia.org/wiki/.properties) to make sense of what you did. Makes sense. Wow. Who would have thought a popular file format without a spec would be so tricky?

spec/syntax.md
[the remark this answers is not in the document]
echeran: This is a good point. I'll leave it unresolved so that I have a better chance of finding it again later. It SGTM. (And `.properties` is still weird (and still popular).)

spec/syntax.md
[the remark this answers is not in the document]
mihnita: If we want to avoid any confusion for people not familiar with properties we can use json: `{ "msg" : "    Hello    " }`
(and they will be confused by all the extra "curlies" :-)

Or make it code (`String msg = "    Hello    "`), readable even if you are not a Java programmer.

But just kidding / nitpicking.
I think it is correct as changed.

echeran (approved):

mihnita (approved):

PR #502, opened 2023-10-24
Make option values optional, defaulting to true

### LINKED ISSUE #386: The case for options without values
TL;DR: Supporting implicit-true unvalued options would provide better UX for developers and translators and simplify validation.

Consider a function that accepts a keyword as its primary argument, say `:brand` for looking up the appropriate localized form of a brand name. Right now, this is achievable in two ways:
```
{|foo-thing| :brand}
```

```
{:brand key=foo-thing}
```

While both of these kinda work, neither is particularly pleasant to use. If #364 is accepted, the first one improves a bit:
```
{foo-thing :brand}
```

However, this order makes autocompletion impossible. Let's presume that the function registry is providing a known set of valid brands that may be localised, and it's only `foo-thing`. Then, by the time you've entered `{fog-thing`, there's no way to know if this is valid or not because the function hasn't been identified yet. To get there, you need to type all of `{fog-thing :brand}`, and then go back and fix your mistake.

Using the `key=` approach is a little bit better in this regard, but really those four characters are rather superfluous, and as with the positional argument, the value _might be a variable_. This means that it could be valid to write `{:brand key=$foo}`, and we wouldn't know until formatting time which brand we're localising. That's pretty powerful (much like dynamic message references, actually), but in many cases perhaps _too_ powerful.

Instead, I think we ought to support unvalued options, so that this becomes possible:
```
{:brand foo-thing}
```
This has three benefits over the current alternatives:
1. Autocompletion can "just work", so writing MF2 feels more natural.
2. Option names cannot be variable, so we can validate the expression without consideration of runtime values.
3. Option basket validation becomes easier, because it depends on the presence of specific options, rather than the value assigned to them.

If we don't support unvalued options, we probably ought to introduce a way that the function registry can specify that a positional argument or an option value cannot be set by a variable. Though tbh that would probably make sense independently of this.

eemeli:
Closes #386

This allows for options without a value, which here default to a literal `true`. This replicates similar behaviour as used in HTML.

The default value is a string, as this explicitly makes `{:foo bar}`, `{:foo bar=true}` and `{:foo bar=|true|}` all indistinguishable from each other, and avoids adding boolean values as a concept.

In addition to function options, the utility of this syntax has been raised in discussions of expression attributes (#450, #458), which would also be able to use the same syntax.

### DIFF
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -21,7 +21,8 @@ annotation = (function *(s option)) / reserved / private-use
 literal = quoted / unquoted
 variable = "$" name
 function = (":" / "+" / "-") name
-option = name [[s] "=" [s] (literal / variable)]
+option = name [option-value]
+option-value = [s] "=" [s] (literal / variable)
 
 ; reserved keywords are always lowercase
 input = %x69.6E.70.75.74  ; "input"
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -21,7 +21,7 @@ annotation = (function *(s option)) / reserved / private-use
 literal = quoted / unquoted
 variable = "$" name
 function = (":" / "+" / "-") name
-option = name [s] "=" [s] (literal / variable)
+option = name [[s] "=" [s] (literal / variable)]
 
 ; reserved keywords are always lowercase
 input = %x69.6E.70.75.74  ; "input"
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -470,16 +470,19 @@ _Options_ are not required.
 An **_<dfn>option</dfn>_** is a key-value pair
 containing a named argument that is passed to a _function_.
 
-An _option_ has a _name_ and a _value_.
-The _name_ is separated from the _value_ by an U+003D EQUALS SIGN `=` along with
+An _option_ has a _name_ and optionally a _value_.
+An _option_ without a _value_ is taken to have the literal value `true`.
+
+If an _option_ has a _value_,
+it is separated from the _name_ by an U+003D EQUALS SIGN `=` along with
 optional whitespace.
-The value of an _option_ can be either a _literal_ or a _variable_.
+The _value_ of an _option_ can be either a _literal_ or a _variable_.
 
 Multiple _options_ are permitted in an _annotation_.
 Each _option_ is separated by whitespace.
 
 ```abnf
-option = name [s] "=" [s] (literal / variable)
+option = name [[s] "=" [s] (literal / variable)]
 ```
 
 > Examples of _functions_ with _options_
@@ -492,10 +495,11 @@ option = name [s] "=" [s] (literal / variable)
 
 > A _message_ with a `$userName` _variable_ formatted with
 > the custom `:person` _function_ capable of
-> declension (using either a fixed dictionary, algorithmic declension, ML, etc.):
+> declension (using either a fixed dictionary, algorithmic declension, ML, etc.),
+> as well as capitalization:
 >
 > ```
-> {Hello, {$userName :person case=vocative}!}
+> Hello, {$userName :person case=vocative capitalized}!
 > ```
 
 > A _message_ with a `$userObj` _variable_ formatted with
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -471,7 +471,7 @@ An **_<dfn>option</dfn>_** is a key-value pair
 containing a named argument that is passed to a _function_.
 
 An _option_ has a _name_ and optionally a _value_.
-An _option_ without a _value_ is taken to have the literal value `true`.
+An _option_ without a _value_ has the literal value `true`.
 
 If an _option_ has a _value_,
 it is separated from the _name_ by an U+003D EQUALS SIGN `=` along with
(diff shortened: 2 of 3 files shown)


stasm:
Did anything change in #386? I don't think there was an agreement to go ahead with this. Let's revive the discussion over there rather than jump to an implementation.

spec/message.abnf
```diff
@@ -21,7 +21,7 @@ annotation = (function *(s option)) / reserved / private-use
 literal = quoted / unquoted
 variable = "$" name
 function = (":" / "+" / "-") name
-option = name [s] "=" [s] (literal / variable)
+option = name [[s] "=" [s] (literal / variable)]
 
 ; reserved keywords are always lowercase
```
aphillips, line 26: We should reserve the keyword `true` if this goes in?

spec/message.abnf
```diff
@@ -21,7 +21,7 @@ annotation = (function *(s option)) / reserved / private-use
 literal = quoted / unquoted
 variable = "$" name
 function = (":" / "+" / "-") name
-option = name [s] "=" [s] (literal / variable)
+option = name [[s] "=" [s] (literal / variable)]
```
aphillips, line 24: Perhaps it would be clearer to separate the value part (CLDR does this with Unicode Locale Identifiers, which have a similar feature)

```suggestion
option = name [option-value]
option-value = [s] "=" [s] (literal / variable)
```

spec/syntax.md
```diff
@@ -470,16 +470,19 @@ _Options_ are not required.
 An **_<dfn>option</dfn>_** is a key-value pair
 containing a named argument that is passed to a _function_.
 
-An _option_ has a _name_ and a _value_.
-The _name_ is separated from the _value_ by an U+003D EQUALS SIGN `=` along with
+An _option_ has a _name_ and optionally a _value_.
+An _option_ without a _value_ is taken to have the literal value `true`.
```
aphillips, line 474: "Is taken" is a little bit elliptical.

```suggestion
An _option_ without a _value_ has the literal value `true`.
```

aphillips (changes requested):
I can't decide if I like this or not. It doesn't _actually_ bring in boolean options, except that it does.

spec/message.abnf
eemeli: I don't think that would be required? The resolved value of a un-valued option doesn't actually show up in the syntax.

ray007:
I do not really like APIs where missing options enable something.
Depending on type, the default for missing should be falsy, `""`, `0`, `false`, `null`, ...

eemeli:
> I do not really like APIs where missing options enable something. Depending on type, the default for missing should be falsy, `""`, `0`, `false`, `null`, ...

This would not be a missing option, it would be an option that is defined. The clearest existing parallel this is connecting with is HTML, which supports e.g. `<input readonly>`, which is rendered with the `readonly` attribute enabled.

aphillips:
@eemeli There is a parallel that is "closer to home": boolean/flag type locale properties in a Unicode Locale Identifier canonicalize to just a key (e.g. `en-US-u-kk` not `en-US-u-kk-true`). I can't find the text in UTR#35 this morning but might come back and put a link here later.

stasm (changes requested):
I'm opposed to this change. It unnecessarily complicates the language. Spelling out option=value is simple, and in fact, I think it's also clearer.

Additionally, I don't understand why this has been suddenly revived as a PR. There's #386 and the discussion in it gives reasons for not implementing this change. Those concerns have not been addressed.

aphillips:
@stasm's comment suggests that a design doc go first. I agree that #386 has some unaddressed bits.

One small insight: MF1 has a number of these "subformat options" that are flags. There might be something useful to things like:

```
{$num :number integer}
{$num :number percent}
```

vs. longer/more verbose:

```
{$num :number maxFractionDigits=0}
{$num :number type=percent}
```

At the same time, imputed values that consist of the string "true" feels a little wacky.

mihnita:
I don't like this.
And never liked it in HTML.

---

It is unclear for example if the value of such a parameter is an empty string, or boolean.

Are "{$foo attr}"  "{$foo attr=true}"   "{$foo attr=yes}"  "{$foo attr=YES}"  "{$foo attr=|yes|}"  "{$foo attr=||}"  the same?
How do I know?
I need to go to the spec :-(

---

> There is a parallel that is "closer to home": boolean/flag type locale properties in a Unicode Locale Identifier canonicalize to just a key (e.g. en-US-u-kk not en-US-u-kk-true). I can't find the text in UTR#35 this morning but might come back and put a link here later.

And there are several bugs in ICU caused by this:
* https://unicode-org.atlassian.net/browse/ICU-20477
* https://unicode-org.atlassian.net/browse/ICU-20310
* https://unicode-org.atlassian.net/browse/ICU-11269

So it is a parallel, and we also kind of know it's error-prone, even for those who implement ICU.

aphillips:
This proposal was rejected in the 2024-01-08 teleconference. Please note that this decision does **_not_** include whether expression attributes can omit the value. That is a separate consideration. In the ldml45 release, all _options_ must include a value.

PR #503, opened 2023-10-27
Update design doc for message pattern quoting

echeran:
An update to the design doc, specifically on the topic of: "Do we allow unquoted variant patterns?"

Very much a WIP while in draft mode (not suitable for "drive-by reviews" until officially ready for review).

@mihnita please take an initial look and provide suggestions.

### DIFF
--- a/exploration/text-vs-code.md
+++ b/exploration/text-vs-code.md
@@ -1,4 +1,4 @@
-# Message Parse Mode
+# Message Pattern Quoting
 
 Status: **Proposed**
 
@@ -17,7 +17,9 @@ Status: **Proposed**
 
 ## Objective
 
-Decide whether text patterns or code statements should be enclosed in MF2.
+Decide whether text patterns must always be quoted,
+or whether we allow them to be optionally quoted,
+for non-simple messages in MF2.
 
 ## Background
 
--- a/exploration/text-vs-code.md
+++ b/exploration/text-vs-code.md
@@ -76,13 +76,26 @@ The [whitespace behavior for Freemarker](https://freemarker.apache.org/docs/dgui
 Other formats supporting multiple message variants tend to rely on a surrounding resource format to define variants,
 such as [Rails internationalization](https://guides.rubyonrails.org/i18n.html#pluralization) in Ruby or YAML
 and [Android String Resources](https://developer.android.com/guide/topics/resources/string-resource.html#Plurals) in XML.
-These formats rely on the resource format providing clear delineation of the beginning and end of a pattern.
+A message author will need to resolve the combination of the rules of these formats and the rules of the containing resource formats in order to achieve a clear delineation of the beginning and end of a pattern.
+For example, an Android resource string that includes leading whitespace in the message might look like
+```
+<string xml:space="preserve">"   Section 7.a. Attribute Types"</string>
+```
+In this example above, the containing XML format will collapse consecutive whitespace characters into a single space unless you provide the attribute `xml:space="preserve"`.
+After the resource file gets parsed as XML, the Android string resource format 
+[does additional whitespace collapsing and Android escaping](https://developer.android.com/guide/topics/resources/string-resource#escaping_quotes),
+requiring the entire text node string to be wrapped in double quotation marks `"..."` to preserve the initial whitespace, or the inital whitespace to use Android escaping (`\u0032 \u0032 ...`).
 
 Based on available data,
 no more than 0.3% of all messages and no more than 0.1% of messages with variants
 contain leading or trailing whitespace.
-No more than one third of this whitespace is localizable,
-and most commonly it's due to improper segmentation or other internationalization bugs.
+However, frequency of occurrence is not an indicator of the importance of leading or trailing whitespace to those authoring such messages.
+For example, sometimes such messages are authored in order to achieve a semblance of formatting in contexts that lack rich text presentation styles,
+such as operating system widgets.
+Even though such messages are usually infrequent relative to the size of all user-facing / transalatable messages,
+that is not an indicator of their significance.
+Also importantly, we cannot make assumptions about the validity of leading or trailing whitespace in a message,
+especially since their usage may be entirely unrelated to internationalization issues (ex: sentence agreement disruption by concatenation).
 
 ## Use-Cases
 
--- a/exploration/text-vs-code.md
+++ b/exploration/text-vs-code.md
@@ -171,27 +171,40 @@ the extent that we make it easy to get into trouble we fail.
 Developers and translators should be able to read and write the syntax easily in a text editor.
 
 Translators (and their tools) are not software engineers, so we want our syntax
-to be as simple, robust, and non-fussy as possible.
-Multiple levels of complex nesting should be avoided,
-along with any constructs that require an excessive
-level of precision on the part of non-technical authors.
+to be as simple and robust as possible.
+
+Nesting level is not a requirement.
+People are not parsers, and don't care about nesting.
+What matters to them is their ability to recognize where a message pattern starts and where it ends.
+In the following example, localizable text is easily recognizable (especially with syntax highlighting),
+even if it occurs 3 level deep.
+
+```java
+print "{{{This is translatable}}}"
+if (foo) print "{{{This is translatable}}}" else print "{{{This is NOT translatable}}}"
+if (foo) if (bar) switch (baz) case 1: print "{{{This is translatable, deep}}}" break; default: print "{{{This is NOT translatable, deep}}}"
+```
 
 As MessageFormat 2 will be at best a secondary language to all its authors and editors,
 it should conform to user expectations and require as little learning as possible.
 
-The syntax should avoid footguns,
-in particular as it's passed through various tools during formatting.
+The syntax should avoid footguns, in particular as it's passed through various tools during formatting or stored existing file formats, databases, etc.
+Very importantly in this regard,
+we should minimize the range of characters that need to be escaped in patterns.
+
 
 ASCII-compatible syntax. While support for non-ASCII characters for variable names,
 values, literals, options, and the like are important, the syntax itself should
 be restricted to ASCII characters. This allows the message to be parsed
 visually by humans even when embedded in a syntax that requires escaping.
 
-Whitespace is forgiving.
-We _require_ the minimum amount of whitespace and allow
-authors to format or change unimportant whitespace as much as they want.
+Whitespace is forgiving, so we should be flexible with its use in the code area of message.
 This avoids the need for translators or tools to be super pedantic about
 formatting.
+However, we want WYSIWYG behavior as much as possible in patterns, meaning that there is minimal visual difference
+between the pattern and its interpolated output,
+and that there is minimal ambiguity.
+This avoids chances for unwanted surprises between the message authoring time expectations and the actual runtime formatted results.
 
 ## Constraints
 
--- a/exploration/text-vs-code.md
+++ b/exploration/text-vs-code.md
@@ -8,6 +8,8 @@ Status: **Proposed**
 		<dt>Contributors</dt>
 		<dd>@eemeli</dd>
 		<dd>@aphillips</dd><!-- Seville and other inserted edits -->
+		<dd>@mihnita</dd>
+		<dd>@echeran</dd>
 		<dt>First proposed</dt>
 		<dd>2023-09-13</dd>
 		<dt>Pull Request</dt>
--- a/exploration/text-vs-code.md
+++ b/exploration/text-vs-code.md
@@ -282,7 +282,7 @@ directly after code:
 >
 > This has a newline and a space at the start and a space-newline at the end
 >
->>{when *}{|
+>{when *}{|
 >|} You can quote the newlines and spaces should you desire {|
 >|}
 >```
--- a/exploration/text-vs-code.md
+++ b/exploration/text-vs-code.md
@@ -66,12 +66,14 @@ Some templating libraries support both styles.
 
 When considering string formatting and templating libraries,
 it is important to keep the rules of pattern or template handling separate from and uninfluenced by the output format's rules.
-For example, many templating languages are designed around producing HTML output, for which consecutive whitespace characters within the output are collapsed into a single ASCII space.
-However, if the templating language is not strict on preserving whitespace,
+For example, many templating languages are designed around producing HTML output, for which consecutive whitespace characters within the output are collapsed into a single ASCII space by HTML renderers.
+However, if the templating language is similarly not strict on preserving whitespace,
 then it would be incapable of generating Python source code,
 for which whitespace is significant in determining block scope via the indentation (leading whitespace on a line).
 
-In fact, some HTML-oriented templating libraries preserve whitespace by default in a what-you-see-is-what-you-get (WYSIWYG) manner (Mustache, [Jinja](https://jinja.palletsprojects.com/en/3.1.x/templates/#whitespace-control)),
+In fact, many HTML-oriented templating libraries preserve whitespace by default in a what-you-see-is-what-you-get (WYSIWYG) manner
+([Mustache](https://mustache.github.io/mustache.5.html#Sections),
+[Jinja](https://jinja.palletsprojects.com/en/3.1.x/templates/#whitespace-control)),
 and some perform whitespace trimming in unspecified ways ([Handlebars](https://handlebarsjs.com/guide/expressions.html#whitespace-control)).
 The [whitespace behavior for Freemarker](https://freemarker.apache.org/docs/dgui_misc_whitespace.html), a general purpose templating library for multiple formats, is also WYSIWYG by default while allowing several optional trimming controls.
 
@@ -165,8 +167,8 @@ Translators (and their tools) are not software engineers, so we want our syntax
 to be as simple and robust as possible.
 
 Nesting level is not a requirement.
-People are not parsers, and don't care about nesting.
-What matters to them is their ability to recognize where a message pattern starts and where it ends.
+People are not parsers, and don't care about the nesting level as a primary concern when reading a message.
+What matters to them is their ability to recognize where a message's pattern starts and where it ends.
 In the following example, localizable text is easily recognizable (especially with syntax highlighting),
 even if it occurs 3 level deep.
 
@@ -205,7 +207,7 @@ The current syntax includes some plain-ascii keywords:
 `input`, `local`, `match`, and `when`.
 
 The current syntax and active proposals include some sigil + name combinations,
-such as :number, $var, |literal|, +bold, -bold, and posibly @attr.
+such as `:number`, `$var`, `|literal|`, `+bold`, `-bold`, and posibly `@attr`.
 
 The current syntax supports unquoted literal values as operands.
 
--- a/exploration/text-vs-code.md
+++ b/exploration/text-vs-code.md
@@ -214,13 +214,25 @@ The current syntax includes some plain-ascii keywords:
 `input`, `local`, `match`, and `when`.
 
 The current syntax and active proposals include some sigil + name combinations,
-such as `:number`, `$var`, `|literal|`, `+bold`, and `@attr`.
+such as :number, $var, |literal|, +bold, -bold, and posibly @attr.
 
 The current syntax supports unquoted literal values as operands.
 
-Messages themselves are "simple strings" and must be considered to be a single
-line of text. In many containing formats, newlines will be represented as the local
-equivalent of `\n`.
+Messages themselves are "simple strings" and must be considered to be WYSIWYG.
+The WYSIWYG nature of representing a message pattern is independent of whether the message is a single line or contains multiple lines.
+
+There is no restriction that a message must only contain a single line (that is, not contain any newline characters),
+nor are there constraints about how newlines must be represented. As our [`spec/syntax.md`](../spec/syntax.md) states:
+
+> Any Unicode code point is allowed, except for surrogate code points U+D800 through U+DFFF inclusive.
+
+> Whitespace in _text_, including tabs, spaces, and newlines is significant and MUST be preserved during formatting.
+
+> ... Instead, we tolerate direct use of nearly all
+characters (including line breaks, control characters, etc.) and rely upon escaping
+in those outer formats to aid human comprehension (e.g., depending upon container
+format, a U+000A LINE FEED might be represented as `\n`, `\012`, `\x0A`, `\u000A`,
+`\U0000000A`, `&#xA;`, `&NewLine;`, `%0A`, `<LF>`, or something else entirely).
 
 ## Proposed Design
 
--- a/exploration/text-vs-code.md
+++ b/exploration/text-vs-code.md
@@ -100,7 +100,6 @@ especially since their usage may be entirely unrelated to internationalization i
 ## Use-Cases
 
 Most messages in any localization system do not contain any expressions, statements or variants.
-These should be expressible as easily as possible.
 
 Many messages include expressions that are meant to be replaced during formatting.
 For example, a greeting like "Hello, {$username}!" would be formatted with the variable
@@ -126,9 +125,20 @@ according to its plural category
 So, in American English, the formatter might need to choose between formatting
 `You have 1 kilometer to go` and `You have 2 kilometers to go`.
 
-Rarely, messages needs to include leading or trailing whitespace due to
-e.g. how they will be concatenated with other text,
+Rarely do messages that need to include leading or trailing whitespace do so due to
+how they will be concatenated with other text,
 or as a result of being segmented from some larger volume of text.
+Based on available data,
+no more than 0.3% of all messages and no more than 0.1% of messages with variants
+contain leading or trailing whitespace.
+
+However, frequency of occurrence is not an indicator of the importance of leading or trailing whitespace to those authoring such messages.
+For example, sometimes such messages are authored in order to achieve a semblance of formatting in contexts that lack rich text presentation styles,
+such as operating system widgets.
+Even though such messages are usually infrequent relative to the size of all user-facing / transalatable messages,
+that is not an indicator of their significance.
+Also importantly, we cannot make assumptions about the validity of leading or trailing whitespace in a message,
+especially since their usage may be entirely unrelated to internationalization issues (ex: sentence agreement disruption by concatenation).
 
 ---
 
--- a/exploration/text-vs-code.md
+++ b/exploration/text-vs-code.md
@@ -28,18 +28,38 @@ ICU MessageFormat and Fluent both support inline selectors
 separated from the text using `{…}` for multi-variant messages.
 ICU MessageFormat is the only known format that uses `{…}` to also delimit text.
 
-[Mustache templates](https://mustache.github.io/mustache.5.html)
-and related languages wrap "code" in `{{…}}`.
-In addition to placeholders that are replaced by their interpolated value during formatting,
-this also includes conditional blocks using `{{#…}}`/`{{/…}}` wrappers.
+Formatting and templating are distinct operations with similarities.
+Both interpolate strings by using input values,
+provided as inputs alongisde the formatting pattern string or template,
+to produce a new string.
+Formatting usually refers to smaller strings, usually no larger than a sentence,
+whereas templating are used to produce larger strings, usually for text files of various file formats, often for HTML documents.
 
-[Handlebars](https://handlebarsjs.com/guide/) extends Mustache expressions
-with operators such as `{{#if …}}` and `{{#each …}}`,
-as well as custom formatting functions that become available as e.g. `{{bold …}}`.
-
-[Jinja templates](https://jinja.palletsprojects.com/en/3.1.x/templates/) separate
-`{% statements %}` and `{{ expressions }}` from the base text.
-The former may define tests that determine the inclusion of subsequent text blocks in the output.
+There are two different styles of templating library design.
+Some languages/libraries enable the interopolation of the template substrings through programmatic expressions in "code mode" that print expressions to the output stream 
+(ex: [PHP](https://www.php.net/),
+[Freemarker](https://freemarker.apache.org/index.html)):
+```php
+<html>
+...
+	<?php
+		if (true) {
+			echo '<p>Hello World</p>';
+		}
+	?>
+...
+</html>
+```
+Some libraries separate string literal values from the programmatic expressions in "code mode" by defining a set of control flow constructs within delimiters,
+and all text outside the delimiters is printed to the output stream,
+and subject to control flow rules of their containing constructs.
+(ex: [Mustache templates](https://mustache.github.io/mustache.5.html),
+[Freemarker](https://freemarker.apache.org/index.html)).
+```
+{{#repo}}
+  <b>{{name}}</b>
+{{/repo}}
+```
 
 A cost that the message formatting and templating languages mentioned above need to rely on
 is some rule or behaviour that governs how to deal with whitespace at the beginning and end of a pattern,
--- a/exploration/text-vs-code.md
+++ b/exploration/text-vs-code.md
@@ -142,9 +142,6 @@ especially since their usage may be entirely unrelated to internationalization i
 
 ---
 
-Developers editing a simple message and who wish to add an `input` or `local` annotiation
-to the message do not wish to reformat the message extensively.
-
 Developers who have messages that include leading or trailing whitespace
 want to ensure that this whitespace is included in the translatable
 text portion of the message.
--- a/exploration/text-vs-code.md
+++ b/exploration/text-vs-code.md
@@ -223,7 +223,72 @@ in those outer formats to aid human comprehension (e.g., depending upon containe
 format, a U+000A LINE FEED might be represented as `\n`, `\012`, `\x0A`, `\u000A`,
 `\U0000000A`, `&#xA;`, `&NewLine;`, `%0A`, `<LF>`, or something else entirely).
 
+## Simple Messages
+
+In the Subcommittee meetings following Github discussions on Issues #493 and #499,
+the general consensus that formed for simple messages
+is that we allow them to be unquoted.
+
+("Simple messages" refers to messages consisting solely of a pattern, and thus are not complex messages.)
+
+Because the simple message pattern consists of the entire message,
+the pattern includes any leading or trailing whitespace.
+
+Given simple messages already being decided at a high level,
+the design decisions below for the proposed and alternative designs pertain specifically to complex messages.
+
 ## Proposed Design
+
+### Start in text, encapsulate message, always quote patterns
+
+Description:
+
+Since simple messages are unquoted (starting in text mode),
+complex messages must also start in text mode.
+
+Within a complex message, patterns are always quoted with `{{...}}` or other choice of delimiter.
+
+The entire complex message is also wrapped with `{{...}}` or other choice of delimiter.
+This allows interior "code mode" of message to have flexible whitespace in between tokens
+and _around_ quoted patterns.
+
+Pros
+
+* The rule about the whether leading and trailing whitespace is included is simple and unambiguous.
+* This matches the WYSIWIG behavior that simple messages preserve.
+* The patterns can be detected within the pattern more easily due to the delimiters serving as a visual anchor.
+* Requiring all patterns to be quoted minimizes the number of characters that need to be escaped within a pattern to 3:
+the 2 pattern delimiter characters and the escape character itself.
+* Because the sum of counts of declarations + `match` statement + `when` statements is always
+greater than or equal to the number of patterns,
+wrapping the entire message once yields less visual noise of repetitive code mode introducer symbols
+when there is 1+ declarations in a `match` (selection) message,
+or when there are 2+ declarations in a non-`match` complex message.
+
+Cons:
+
+* This comes at the cost of an inconsistency in the WYSIWYG patterns are quoted between simple and complex messages.
+In the case of simple messages, the containing format itself implicitly defines the beginning and end of the pattern (example: `"..."`), which is not visible at the level of MF2,
+while complex messages use the aforementioned delimiter to quote patterns (ex: `{{...}}`).
+* Another potential drawback, specifically in the case of non-`match` complex messages with exactly 1 declaration,
+is that this option adds 2 extra delimiters compared to an alternative syntax that doesn't require quoted patterns
+and is designed to minimize delimiter usage only to code mode introducers.
+
+Evaluation:
+
+The pros outweigh the cons, not just in cardinality, but far more importantly, according to the relative weight
+our value system places to the requirements met by the pro aspects compared to the con aspects. Namely:
+
+* [high] Unsurprising WYSIWYG behavior from patterns
+* [high] Easy recognition of patterns, even for non-developers
+* [high] A minimal number of characters requiring escaping
+* [high] No limitations on users with valid non-i18n concerns
+* [med] Flexible whitespace outside of patterns
+* [low] Number of characters typed (probably comparable with alternatives anyways)
+* [low] Number of "mode levels" from a parser perspective
+
+
+## Alternatives Considered
 
 ### Start in text, encapsulate code, trim around statements
 
@@ -244,8 +309,6 @@ Allow for a pattern to be `{{…}}` quoted
 such that it preserves its leading and/or trailing whitespace
 even when preceded or followed by statements.
 
-## Alternatives Considered
-
 ### Start in code, encapsulate text
 
 This approach treats messages as something like a resource format for pattern values.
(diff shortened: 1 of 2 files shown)


exploration/text-vs-code.md
```diff
@@ -17,7 +19,9 @@ Status: **Proposed**
 
 ## Objective
 
-Decide whether text patterns or code statements should be enclosed in MF2.
+Decide whether text patterns must always be quoted,
+or whether we allow them to be optionally quoted,
+for non-simple messages in MF2.
```
aphillips, line 24: Perhaps:

```suggestion
Decide whether to permit _patterns_ (message text) to be unquoted
when embedded in code. 
Currently all _patterns_ must be quoted in non-simple messages.
```

exploration/text-vs-code.md
```diff
@@ -28,38 +32,69 @@ ICU MessageFormat and Fluent both support inline selectors
...
-`{% statements %}` and `{{ expressions }}` from the base text.
-The former may define tests that determine the inclusion of subsequent text blocks in the output.
-
-A cost that the message formatting and templating languages mentioned above need to rely on
-is some rule or behaviour that governs how to deal with whitespace at the beginning and end of a pattern,
-as statements may be separated from each other by newlines or other constructs for legibility.
+Formatting and templating are distinct operations with similarities.
+Both interpolate strings by using input values,
+provided as inputs alongisde the formatting pattern string or template,
+to produce a new string.
```
aphillips, line 38: I've been trying to back away from the use of "interpolate" and it strikes me that this could be said more simply. What do you think of:

```suggestion
Both take input values to replace portions of the pattern string or template, producing a new, formatted, string.
```

exploration/text-vs-code.md
```diff
@@ -28,38 +32,69 @@ ICU MessageFormat and Fluent both support inline selectors
...
+Formatting usually refers to smaller strings, usually no larger than a sentence,
+whereas templating are used to produce larger strings, usually for text files of various file formats, often for HTML documents.
```
aphillips, line 40: Grammar and clarity suggestions:

```suggestion
whereas templating is typically used to produce larger strings (generally whole documents, such as an HTML file)
```

exploration/text-vs-code.md
```diff
@@ -85,14 +120,22 @@ according to its plural category
 So, in American English, the formatter might need to choose between formatting
 `You have 1 kilometer to go` and `You have 2 kilometers to go`.
 
-Rarely, messages needs to include leading or trailing whitespace due to
-e.g. how they will be concatenated with other text,
+Rarely do messages that need to include leading or trailing whitespace do so due to
+how they will be concatenated with other text,
```
aphillips, line 124: I think this change made the text far less clear. 

To be honest, I would have guessed that you would have replaced the somewhat biased word "Rarely" with something more neutral such as "Some messages need..."

exploration/text-vs-code.md
```diff
@@ -85,14 +120,22 @@ according to its plural category
...
+contain leading or trailing whitespace.
 
----
+However, frequency of occurrence is not an indicator of the importance of leading or trailing whitespace to those authoring such messages.
+For example, sometimes such messages are authored in order to achieve a semblance of formatting in contexts that lack rich text presentation styles,
+such as operating system widgets.
+Even though such messages are usually infrequent relative to the size of all user-facing / transalatable messages,
+that is not an indicator of their significance.
+Also importantly, we cannot make assumptions about the validity of leading or trailing whitespace in a message,
+especially since their usage may be entirely unrelated to internationalization issues (ex: sentence agreement disruption by concatenation).
```
aphillips, line 136: This doesn't seem clear to me?

I think what you might be trying to say is:

```suggestion
Also, importantly, whether the intentional inclusion of whitespace by a 
message author might be considered "desirable" or might be interpreted 
as "an internationalization bug",
we need to provide the ability of an author to control the content of a given pattern without ambiguity.
```

exploration/text-vs-code.md
```diff
@@ -101,32 +144,62 @@ Which whitespace characters are displayed at runtime should not be surprising.
...
 
-Common things should be easy, uncommon things should be possible.
+It should be easy to do simple things; possible to do complex things; and impossible, or at least difficult, to do wrong things.
+
+<details>
+<blockquote>
+APIs should be easy to use and hard to misuse. It should be easy to do simple things; possible to do complex things; and impossible, or at least difficult, to do wrong things.
+
+—<a href="https://www.infoq.com/articles/API-Design-Joshua-Bloch/">Joshua Bloch, 2008</a>, author of <i>Effective Java</i>, etc.
+</blockquote>
```
aphillips, line 154: Maybe just quote Bloch?

exploration/text-vs-code.md
```diff
@@ -101,32 +144,62 @@ Which whitespace characters are displayed at runtime should not be surprising.
...
 Translators (and their tools) are not software engineers, so we want our syntax
-to be as simple, robust, and non-fussy as possible.
-Multiple levels of complex nesting should be avoided,
-along with any constructs that require an excessive
-level of precision on the part of non-technical authors.
+to be as simple and robust as possible.
+
+Nesting level is not a requirement.
+People are not parsers, and don't care about the nesting level as a primary concern when reading a message.
+What matters to them is their ability to recognize where a message's pattern starts and where it ends.
```
aphillips, line 173: I see what you're trying to do, but the original requirement didn't say "avoid all nesting". It used "should" and was concerned with depth of nesting and message complexity. A well-known shortcoming of MF1 are the plural/select constructs with their complex use of nested `{...}` sequences. The example below is argument and this is (and claims to be!) "not a requirement" in the requirements section. If you want to talk about nesting like this, talk about it in design considerations?

If you want a requirement, then perhaps:

```suggestion
Nesting of text or code should be as simple as possible, but no simpler.
```

exploration/text-vs-code.md
```diff
@@ -101,32 +144,62 @@ Which whitespace characters are displayed at runtime should not be surprising.
...
+
 
 ASCII-compatible syntax. While support for non-ASCII characters for variable names,
 values, literals, options, and the like are important, the syntax itself should
 be restricted to ASCII characters. This allows the message to be parsed
 visually by humans even when embedded in a syntax that requires escaping.
 
-Whitespace is forgiving.
-We _require_ the minimum amount of whitespace and allow
-authors to format or change unimportant whitespace as much as they want.
```
aphillips, line 127: Why did you remove this? It seems like a good observation about all of our designs.

exploration/text-vs-code.md
```diff
@@ -101,32 +144,62 @@ Which whitespace characters are displayed at runtime should not be surprising.
...
+Whitespace is forgiving, so we should be flexible with its use in the code area of message.
 This avoids the need for translators or tools to be super pedantic about
 formatting.
+However, we want WYSIWYG behavior as much as possible in patterns, meaning that there is minimal visual difference
+between the pattern and its interpolated output,
+and that there is minimal ambiguity.
+This avoids chances for unwanted surprises between the message authoring time expectations and the actual runtime formatted results.
```
aphillips, line 202: This seems like a separate requirement. Note that the core of this conversation is not actually about whether **_patterns_** are WYSIWYG but with the problem of determining pattern _boundaries_.

Requirements language might be more like:

```suggestion

Pattern contents MUST be WYSIWYG. Once a pattern is determined or selected, there should be no surprise about the sequence of characters included in the pattern, including which portions are replaced during formatting.
```

exploration/text-vs-code.md
```diff
@@ -136,16 +209,93 @@ The current syntax includes some plain-ascii keywords:
 `input`, `local`, `match`, and `when`.
 
 The current syntax and active proposals include some sigil + name combinations,
-such as `:number`, `$var`, `|literal|`, `+bold`, and `@attr`.
+such as `:number`, `$var`, `|literal|`, `+bold`, `-bold`, and posibly `@attr`.
```
aphillips, line 212: Typo

```suggestion
such as `:number`, `$var`, `|literal|`, `+bold`, `-bold`, and possibly `@attr`.
```

exploration/text-vs-code.md
```diff
@@ -136,16 +209,93 @@ The current syntax includes some plain-ascii keywords:
...
 
 The current syntax supports unquoted literal values as operands.
 
-Messages themselves are "simple strings" and must be considered to be a single
-line of text. In many containing formats, newlines will be represented as the local
-equivalent of `\n`.
+Messages themselves are "simple strings" and must be considered to be WYSIWYG.
```
aphillips, line 216: This is incorrect and the change greatly reduces the impact of the document IMO. The thing that is WYSIWYG is the pattern. In the case of simple messages, this is the whole message. But in the case of complex messages, what you see (`{#input $foo :number minimumFractionDigits=11}`) is not exactly what you get 😉 

The most important part of the original statement here is removed, which reminds readers that this message:

```
myMessage = {{
   {#input $var :number}
   {{You have {$var} message(s)}}
}}
```

Is actually this message in many storage formats:

```
myMessage = {{\n   {#input $var :number}\n   {{You have {$var} message(s)}}\n}}
```

exploration/text-vs-code.md
```diff
@@ -136,16 +209,93 @@ The current syntax includes some plain-ascii keywords:
...
+The WYSIWYG nature of representing a message pattern is independent of whether the message is a single line or contains multiple lines.
+
+There is no restriction that a message must only contain a single line (that is, not contain any newline characters),
```
aphillips, line 219: I would s/message/pattern/g this section (carefully because my suggestion is not always true). Only talk about the message when you intend to include the code.

exploration/text-vs-code.md
```diff
@@ -136,16 +209,93 @@ The current syntax includes some plain-ascii keywords:
...
+* The patterns can be detected within the pattern more easily due to the delimiters serving as a visual anchor.
+* Requiring all patterns to be quoted minimizes the number of characters that need to be escaped within a pattern to 3:
+the 2 pattern delimiter characters and the escape character itself.
+* Because the sum of counts of declarations + `match` statement + `when` statements is always
+greater than or equal to the number of patterns,
+wrapping the entire message once yields less visual noise of repetitive code mode introducer symbols
+when there is 1+ declarations in a `match` (selection) message,
+or when there are 2+ declarations in a non-`match` complex message.
+
+Cons:
```
aphillips, line 274: Consider adding that the message closing pattern characters add no value.

spec/syntax.md
```diff
@@ -64,7 +64,7 @@ The syntax specification takes into account the following design restrictions:
 
 ## Messages and their Syntax
 
-The purpose of MessageFormat is the allow content to vary at runtime.
+The purpose of MessageFormat is to allow content to vary at runtime.
```
aphillips, line 67: Good catch. Should we fix this separately? I can fast track it.

aphillips (changes requested):
Looking good. Thank you both for working on this (especially during "maximum crunch time". I've made a number of suggestions below. Please have a look. They are mostly editorial in nature.

exploration/text-vs-code.md
```diff
@@ -101,32 +144,62 @@ Which whitespace characters are displayed at runtime should not be surprising.
...
+our platform and frameworks. To
+the extent that we make it easy to get into trouble we fail.
+          
+—Rico Mariani, MS Research MindSwap Oct 2003. (<a href="https://learn.microsoft.com/en-us/archive/blogs/brada/the-pit-of-success">restated by Brad Adams</a>, MS CLR and .Net team cofounder)
+</blockquote>
+</details>
 
 Developers and translators should be able to read and write the syntax easily in a text editor.
 
 Translators (and their tools) are not software engineers, so we want our syntax
```
stasm, line 168: I think it would be good to prioritize the requirements. For example, while I agree that we should, in general terms, make the syntax simple and robust, I would also suggest that the primary consumer of the _message_ syntax are developers. Translators will oftentimes work with just the _pattern_ syntax, through CAT tools.

exploration/text-vs-code.md
```diff
@@ -136,16 +209,93 @@ The current syntax includes some plain-ascii keywords:
...
+
+Description:
+
+Since simple messages are unquoted (starting in text mode),
+complex messages must also start in text mode.
+
+Within a complex message, patterns are always quoted with `{{...}}` or other choice of delimiter.
+
+The entire complex message is also wrapped with `{{...}}` or other choice of delimiter.
+This allows interior "code mode" of message to have flexible whitespace in between tokens
```
stasm, line 258: What do you think should happen to whitespace outside the entire code block? Should we specify it here as well?

exploration/text-vs-code.md
```diff
@@ -136,16 +209,93 @@ The current syntax includes some plain-ascii keywords:
...
+and _around_ quoted patterns.
+
+Pros
+
+* The rule about the whether leading and trailing whitespace is included is simple and unambiguous.
+* This matches the WYSIWIG behavior that simple messages preserve.
+* The patterns can be detected within the pattern more easily due to the delimiters serving as a visual anchor.
+* Requiring all patterns to be quoted minimizes the number of characters that need to be escaped within a pattern to 3:
```
stasm, line 266: I think this also holds for at least some of the other proposals.

In fact, if we always quote variant patterns, we must make the _closing_ delimiter special. Theoretically, in other proposals, we could only special-case the opening delimiter, because both code and placeholders would be wrapped in the same delimiter. (Although that would require agreeing to a different way of preserving whitespace than the currently agreed `{{ ... }}`.)

exploration/text-vs-code.md
```diff
@@ -136,16 +209,93 @@ The current syntax includes some plain-ascii keywords:
...
+or when there are 2+ declarations in a non-`match` complex message.
+
+Cons:
+
+* This comes at the cost of an inconsistency in the WYSIWYG patterns are quoted between simple and complex messages.
+In the case of simple messages, the containing format itself implicitly defines the beginning and end of the pattern (example: `"..."`), which is not visible at the level of MF2,
+while complex messages use the aforementioned delimiter to quote patterns (ex: `{{...}}`).
+* Another potential drawback, specifically in the case of non-`match` complex messages with exactly 1 declaration,
+is that this option adds 2 extra delimiters compared to an alternative syntax that doesn't require quoted patterns
+and is designed to minimize delimiter usage only to code mode introducers.
```
stasm, line 281: We should add some of the other previously discussed drawbacks:

* If the code block is delimited from both sides, users may be tempted to insert text around it.
* If the code block is delimited from both sides, it may be easy to forget the `}}` closing the entire block.
* If we use curlies for patterns and for placeholders, then they serve double duty, which may make the syntax harder to understand, and also harder to make the pattern out visually.

exploration/text-vs-code.md
```diff
@@ -136,16 +209,93 @@ The current syntax includes some plain-ascii keywords:
...
+the 2 pattern delimiter characters and the escape character itself.
+* Because the sum of counts of declarations + `match` statement + `when` statements is always
+greater than or equal to the number of patterns,
+wrapping the entire message once yields less visual noise of repetitive code mode introducer symbols
+when there is 1+ declarations in a `match` (selection) message,
+or when there are 2+ declarations in a non-`match` complex message.
+
+Cons:
+
+* This comes at the cost of an inconsistency in the WYSIWYG patterns are quoted between simple and complex messages.
```
stasm, line 276: We should discuss here the risk of "two syntaxes in a trench coat." For someone who's only ever seen simple messages, the only syntax rule they can infer is that `{}` is used for placeholders. A whole separate complex-message mode cannot be easily "guessed".

(But see also my comment about considering developers the primary audience for the complex message syntax, so perhaps this is acceptable.)

exploration/text-vs-code.md
```diff
@@ -136,16 +209,93 @@ The current syntax includes some plain-ascii keywords:
...
+is that this option adds 2 extra delimiters compared to an alternative syntax that doesn't require quoted patterns
+and is designed to minimize delimiter usage only to code mode introducers.
+
+Evaluation:
+
+The pros outweigh the cons, not just in cardinality, but far more importantly, according to the relative weight
+our value system places to the requirements met by the pro aspects compared to the con aspects. Namely:
+
+* [high] Unsurprising WYSIWYG behavior from patterns
+* [high] Easy recognition of patterns, even for non-developers
```
stasm, line 289: To be fair, I've never found ICU MF patterns easy to spot—specifically because they use `{}`.

exploration/text-vs-code.md
```diff
@@ -28,38 +32,69 @@ ICU MessageFormat and Fluent both support inline selectors
...
+There are two different styles of templating library design.
+Some languages/libraries enable the interopolation of the template substrings through programmatic expressions in "code mode" that print expressions to the output stream 
+(ex: [PHP](https://www.php.net/),
+[Freemarker](https://freemarker.apache.org/index.html)):
+```php
+<html>
+...
+	<?php
+		if (true) {
+			echo '<p>Hello World</p>';
```
stasm, line 51: Just a nit, but I think both PHP and Freemarker would actually prefer the second method:

```php
<?php if (true): ?>
    <p>Hello, world!</p>
<?php endif ?>
```

stasm (commented):
I know you said this wasn't ready for reviews yet, so please consider this my early feedback on an early draft, which I'm sharing now because the day here is soon over :)

exploration/text-vs-code.md
mihnita: A bit worse: as a translator you often get strings with no context.
There is no way to know if  through what kind of API the string will go.

So if I get this string "Bill Gates did something" and I have to translate, and I want to put the "honorific space" in front of the name, I don't know how to do it.
Is the message going through MF2 with trimming? Then I have to wrap it in `{  Bill ....}`
But maybe it is not going through MF2.
Then if I wrap it the `{` ... `}` will render on screen "as is", not what you want.

We've seen this:
"The bee" => "L'abbeile" : if the message goes through MF1, the apostrophe needs to be escaped
"I'm 1 in 100" => "Eu sunt 1%" : if the string goes through a printf-like API then I need to escape the `%`

And as a translator I have no why to know what API the dev uses, and I'm not familiar with the escape rules for myriads of APIs.
I am in fact faced with more APIs than a developer.
A dev might do "java + html + js".
A translator often works for many projects, from many companies, so it is exposed to strings consumed by native Windows apps, PHP, some Ruby stuff, C#, SQL, others. And even switch several times per day.

These days you don't pay the bills as a translator if you only handle one single format for one produce from one company.

spec/syntax.md
echeran: Fast-tracking this typo fix separately SGTM.

exploration/text-vs-code.md
mihnita: We think they should be preserved "as is". If it is not specified, we should.

* consistency with complex (no trimming)
* many of the templating systems don't trim by default, some have it opt-in
* clear what you get
   If you see "   Foo" you know you will see spaces. If we trim ascii spaces, the leading spaces might have nbsp, em-space, en-space, ideographic spaces, etc. So if you read or even edit the string, you can't tell what the result is without a way to inspect those spaces that look all the same.
* as a translator, if I want "  Bill Gates ..." (honorific spaces in Chinese) I don't know if I need to wrap / escape or not.
   Because as a translator I don't know what API is used on that string. Maybe it is MF2, or maybe not.

exploration/text-vs-code.md
mihnita: > In fact, if we always quote variant patterns, we must make the closing delimiter special.

But we have to do it for all the other proposals too.
Because all allow for optional wrapping.

exploration/text-vs-code.md
mihnita: adding text after closing is not necessarily a drawback (I have a comment somewhere)

> If the code block is delimited from both sides, it may be easy to forget the }} closing the entire block.

Fair enough.
Maybe make it optional?

I think that `{{` is to make this intentionally ugly. I would probably go with `{#` to enter code mode, and `{` to wrap the patterns.
And yes, it is double douty. But choosing anything else  means that we escape one extra thing.
If we use `<<<` and I want (for some reason `<<< Hello world` as simple text, then we need to escape it.

So we add another escaping rule.
Pros and cons :-)

exploration/text-vs-code.md
mihnita: >  "two syntaxes in a trench coat."

Sounds ugly, but there are already templating systems doing this, and didn't prevent adoption, or trigger many complaints.
Heck, when I write HTML with stylesheets and code I have 3 syntaxes in trench coat :-)

exploration/text-vs-code.md
mihnita: Fair point.

But I've never seen any questions asked for single selections.
(on StackOverflow or internally sites, here and in previous companies)

It gets ugly, even for developers, when you have multiple selections (plural-in-plural, select in plural in ???)

exploration/text-vs-code.md
mihnita: I didn't even know that is possible.

If I check https://www.w3schools.com/php/php_if_else.asp and https://www.php.net/manual/en/control-structures.if.php the option is not even mentioned.

It is mentioned in the "User Contributed Notes" of the manual.

But I kind of doubt that something that is not even mentioned in the official manual is the preferred way.

If anything this makes the point that having more than one way to do things, some more recommended than others, is not a good thing.

One will argue that PHP is not the best designed language. And it is not.
But do we want MF2 to become that? :-)

eemeli:
On my part, I've spent much more time than I'd wish over the past few days and weeks thinking about and looking into messages with external whitespace. In the absence of any better place to put down some of my thoughts on this, these are the aspects and arguments that I find important to account for:

### Localizable external whitespace is really rare, while mistakes are common

Sometimes a leading or trailing space could be localizable, but you can't necessarily tell without looking through the code which is using the message. So I did that. This is what I'd mentioned previously via email:

> In total, [Mozilla's Pontoon] system has so far handled about 175k translatable messages, of which 568 had exterior whitespace in the source locale. These I've manually categorised as follows, starting from the most common:
> - 180 incorrect segmentation / string concatenation, e.g. " for this version." or "Download the "
> - 122 contents wrapped in a `<tag>`, so the whitespace is almost certainly an insignificant segmentation artifact.
> - 108 ends in colon+space, such as "Description: "
> - 106 unlocalized markup, such as leading or trailing newlines
> - 52 potentially localizable space between clauses
>
> So overall that's perhaps 0.3% of all messages, of which (generously) up to a third might be localizable. If you'd like to perform your own analysis, please reach out to me separately for the source data.

As a next step, I filtered all of the above to the 41 potentially localizable strings which are currently in production (looks like a sentence, has one leading or trailing space), and found where they're coming from in code, and how they're used: [0](https://github.com/mozilla/addons-server/blob/57c21f099f037f7246e724ba1b055cfdaeec6a53/src/olympia/devhub/templates/devhub/addons/invitation.html#L21) [1](https://github.com/mozilla/addons-server/blob/66952b999cbbf824d7618f9106a5a44c6e18fe0f/src/olympia/devhub/templates/devhub/includes/addon_details.html#L25-L27) [2](https://github.com/mozilla/addons-server/blob/66952b999cbbf824d7618f9106a5a44c6e18fe0f/src/olympia/devhub/templates/devhub/addons/edit/describe.html#L158-L160) [3](https://github.com/mozilla/contain-facebook/blob/f6bcec97a32d007687ba3babdbdd366bef836757/src/_locales/en/messages.json#L213) [4](https://github.com/mozilla/fxa/blob/40cd252a2052c07a095cfc21a13b6c314a70b83c/packages/fxa-content-server/app/scripts/templates/pair/auth_allow.mustache#L20) [5](https://github.com/mozilla/kitsune/blob/f03155bfbf64b8e211e1de544ad3e256a4c2c45e/kitsune/sumo/db_strings.py#L422) [6](https://github.com/mozilla/kitsune/blob/f03155bfbf64b8e211e1de544ad3e256a4c2c45e/kitsune/sumo/db_strings.py#L471) [7](https://github.com/mozilla/kitsune/blob/f03155bfbf64b8e211e1de544ad3e256a4c2c45e/kitsune/sumo/db_strings.py#L479) [8](https://github.com/mozilla/kitsune/blob/f03155bfbf64b8e211e1de544ad3e256a4c2c45e/kitsune/sumo/db_strings.py#L569) [9](https://github.com/mozilla/kitsune/blob/f03155bfbf64b8e211e1de544ad3e256a4c2c45e/kitsune/sumo/db_strings.py#L585) [10](https://github.com/mozilla/kitsune/blob/f03155bfbf64b8e211e1de544ad3e256a4c2c45e/kitsune/wiki/jinja2/wiki/includes/document_macros.html#L353) [11](https://github.com/mozilla-mobile/focus-android/blob/7fb21e5800cf9fd0d52954b85f7b24d26b2e0c34/app/src/main/res/values/strings.xml#L814) [12](https://github.com/mozilla-mobile/focus-ios/blob/6a1d56f5321c1140f0d93d89fb658f76774fb45c/Blockzilla/UIComponents/UIConstants.swift#L346) [13](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/7e81e3771b3fa669236c01bf9b59be82a47c4ae7/network-api/networkapi/templates/pages/buyersguide/about/methodology.html#L333) [14](https://github.com/thundernest/thunderbird-website/blob/6b9e9354c1753a61b19932a41805fd44f9aa9cb6/website/download/beta/index.html#L64) [15](https://github.com/thundernest/thunderbird-website/blob/6b9e9354c1753a61b19932a41805fd44f9aa9cb6/website/download/beta/index.html#L120) [16](https://github.com/thundernest/thunderbird-website/blob/6b9e9354c1753a61b19932a41805fd44f9aa9cb6/website/download/beta/index.html#L143) [17](https://searchfox.org/comm-central/search?q=ListDescription.label) [18](https://searchfox.org/comm-central/search?q=ListName.label) [19](https://searchfox.org/comm-central/search?q=ListNickName.label) [20](https://searchfox.org/comm-central/search?q=addToAddressBook.label) [21](https://searchfox.org/comm-central/search?q=authMethod.label) [22](https://searchfox.org/comm-central/search?q=clearTimeDuration.label) [23](https://searchfox.org/comm-central/search?q=connectionSecurity.label) [24](https://searchfox.org/comm-central/search?q=directoryBaseDN.label) [25](https://searchfox.org/comm-central/search?q=directoryHostname.label) [26](https://searchfox.org/comm-central/search?q=directoryLogin.label) [27](https://searchfox.org/comm-central/search?q=directoryName.label) [28](https://searchfox.org/comm-central/search?q=folderSelectionCaption.label) [29](https://searchfox.org/comm-central/search?q=ispHeaders.label) [30](https://searchfox.org/comm-central/search?q=portNumber.label) [31](https://searchfox.org/comm-central/search?q=saslMechanism.label) [32](https://searchfox.org/comm-central/search?q=scope.label) [33](https://searchfox.org/comm-central/search?q=searchFilter.label) [34](https://searchfox.org/comm-central/search?q=searchTermCaption.label) [35](https://searchfox.org/comm-central/search?q=sendingPrefix.label) [36](https://searchfox.org/comm-central/search?q=serverDescription.label) [37](https://searchfox.org/comm-central/search?q=serverName.label) [38](https://searchfox.org/comm-central/search?q=serverPort.label) [39](https://searchfox.org/comm-central/search?q=userName.label) [40](https://searchfox.org/comm-central/search?q=whitelistHeader.label).

Of the above, only the 0th message actually contains a localizable space; the other 40 are all bugs. So that's exactly one localizable external space in about 66k messages currently in production. This space was incorrectly dropped in 15 of the 35 locales to which it's translated; I've now submitted corrections for all of them.

Real localizable external whitespace is so rare that it gets drowned by the noise.

All of the above bugs are in formats that explicitly delimit patterns, and thereby make it too easy to include leading or trailing whitespace. About 28% of the messages currently in production use Fluent, which does not quote patterns. None of these were similarly buggy.

From this I would conclude that using a syntax which requires external whitespace to be explicitly intentional would make it much less likely for it to be ignored, and would lead to better localizations.


### We're not actually talking about quoting patterns

To be precise, we are talking about _delimiting_ patterns with {braces}, which rather explicitly are not 'quotes' or "quotes". The distinction here matters, because we are looking to assign a novel meaning to a pair of characters that no other syntax than ICU MessageFormat uses to delimit localizable text. Every other syntax which uses braces in text uses them to delimit _code_. Which MF2 is also doing.

When we talk of the syntax being "WYSIWYG", we are asking for its readers to _not see_ the {braces}, a symbol so prevalent in our syntax that we've jokingly incorporated it into our logo {�}. Humans are not trained for that the way they are with "quotes", or with empty spaces acting as content separators. In other words, if I see the braces but I don't see the empty space, how is the syntax "WYSIWYG"?

We are also asking for MF2 authors and editors to somehow know that the spaces within the { braces } are significant -- but only if they're delimiting patterns. Within expressions, MF2 syntax ignores whitespace, so `{{ Hello {world }}}` formats to contain a leading space, but no trailing space. I mean, go back to the first sentence of this paragraph and consider how you saw the " braces ": Was it truly obvious that its padding spaces were a part of the string it represents the same way they are with the double quotes?


### The overlap of external whitespace and variants is really truly tiny

In a very real sense, the discussion of whether messages with variants should always be delimited is a discussion asking if [this string](https://searchfox.org/comm-central/source/mail/locales/en-US/messenger/messengercompose/messengercompose.ftl#85) is ok being represented as

```
{#match :platform}
{#when macos} ⇧ ⌘{| |}
{#when *} Ctrl+Shift+
```

or if that is sufficiently problematic that we need every pattern of every message with variants to be {{delimited}}.

I have been actively seeking for examples of messages with leading or trailing whitespace for the last month, and the above is the _one_ actual, current message with variants and external whitespace that has been identified. We are talking about such a rare situation it should not be driving our whole syntax.

We should choose to do with patterns what we're doing with literals, where for common values we allow them to be delimited by whitespace, but also permit `|vertical pipes|` to be used as "quotes" in the rare cases where they're required.

spec/syntax.md
aphillips: Done

exploration/text-vs-code.md
aphillips: If there is anything "outside" the complex message-wide delimiters, then:

1. It's whitespace that we trim
   a. Or produces an error (worth discussion)
3. or the message is actually a simple message that produces a lot of errors

```
This {{ match {$var} when * {{{$var}}}}} is an interesting message.
```

This evaluates, I think, as:

>This {&#xfffd;} is an interesting message.

Or possibly (with `$var`==123) as:

> This {&#xfffd;}{&#xfffd;} match 123 when * 123{&#xfffd;}{&#xfffd;} is an interesting message.

(both emit a syntax error)

exploration/text-vs-code.md
aphillips: >> If the code block is delimited from both sides, it may be easy to forget the }} closing the entire block.
> Forgetting to type syntax can happen in any alternative option, so this feels like a weaker argument than the previous.

I disagree a bit. Any enclosing syntax is prone to errors, but the argument here is that the more levels of nesting (and the further apart the enclosure bits), the more opportunities for error exist because the user is keeping track of more things.

Note that one of the proposals for "2a" was to have just a starter sigil. This would eliminate the enclosure for the message.

aphillips:
@eemeli Thanks for the long comment.

> To be precise, we are talking about delimiting patterns with {braces}, which rather explicitly are not 'quotes' or "quotes".

I agree that we are talking about _delimiting_ patterns and that this is what our technical decision is about. _Quoting_ would be one mechanism for delimiting patterns, but is not the only one. I would call out that _quoting_ does not require the use of "quote" characters. I think that referring to `{{` and `}}` as pattern quotes is fine (although we could be pedantic and call them "pattern delimiters" if you prefer)

> The overlap of external whitespace and variants is really truly tiny

I think this isn't quite on the nose. The real problem we're dealing with here is intentionality. There has to be a way for users to intentionally include various kinds of character sequence into their _pattern_. This includes invisible _Unicode_ whitespace that is not MF2 whitespace. For example, non-breaking space or NNBSP or ZWNJ or what have you. There are a lot of characters that have no "ink" but which a user might intend to be part of the message. Some of those characters will be MF2 whitespace.

MF2 _intentionally_ does not include a general purpose character escaping mechanism (because we expect the host environment or file format to include one and we are avoiding the double-escaping mess). If the boundary between "pattern" and "not pattern" is between invisible characters, that's pretty difficult to work with.

---

I think there are **four** audiences that need to be served:

1. Developers/message authors. The creator of the source message needs to understand how the pattern is delimited and be able to clearly include any character sequence into a pattern.
2. Translators. Translators, like developers, need to know whether whitespace is part of the pattern or not and to intentionally include whitespace into patterns where necessary.
3. MF2 parsers. The parser obviously need to be able to determine the boundary for the pattern if the message is to be rendered.
4. Tools. CAT tools _could_ treat the entire message as a single segment. But they could also produce separate segments for each `when` case (including generating necessary additional segments for the target locale). Either way, tools generally protect syntax to allow translators to focus on content. Non-CAT tools also need to preserve intentional whitespace and not interfere with pattern delimiters.

What I think is interesting is that pattern delimiters are probably syntax. If pattern delimiters are optional, it might be unclear to translators whether a given pattern is already quoted (delimited) or to tools as to whether delimiters are needed. It's hard for machines to guess people's intentions. I agree that PEWS is rare, which is why we need to be especially clear about how to handle the rare cases.

eemeli:
> [...] I think that referring to `{{` and `}}` as pattern quotes is fine (although we could be pedantic and call them "pattern delimiters" if you prefer)

I don't want to insist on pedantry, as long as we recognise that the pattern "quote" characters we're considering are rather explicitly not generally used as quote characters. For anyone coming to MF2 not via MF1, this will be an additional weird thing to learn. Very much comparable to `|quoted literals|`, which I've heard argued not to be so bad because they'll be so rare.

> [...] There has to be a way for users to intentionally include various kinds of character sequence into their _pattern_. This includes invisible _Unicode_ whitespace that is not MF2 whitespace. For example, non-breaking space or NNBSP or ZWNJ or what have you. There are a lot of characters that have no "ink" but which a user might intend to be part of the message. Some of those characters will be MF2 whitespace.

Yes, and this will be supported no matter which way we decide to go, by the _optional_ `{{pattern quoting}}` ability. Which unfortunately will mean that the closing brace `}` may need to be considered a syntax character, and require special escaping. Pontoon has seen 17 messages with `}` as a pattern character, an order of magnitude more than messages with localizable external whitespace.

echeran:
Ready for review now. I made edits mostly based on #504, with some additions, and also responding comments that came prior to being ready anyways. Previous info from the doc (background, use cases, stats, examples) is kept pre-hidden at the bottom, like appendices.

exploration/text-vs-code.md
```diff
@@ -17,49 +19,361 @@ Status: **Proposed**
...
+>   {when *} This pattern has a space in front (it's between \} and This)
+>   {when other}
+>      This pattern has a newline and six spaces in front of it
+>   {when moo}This pattern has no spaces in front of it, but an invisible space at the end
+>}}
+>```
+
+Pros:
+- WYSIWYG (on steroids)
+
```
eemeli, line 188:
```suggestion
- Avoids as many escape sequences as possible,
  as `}` does not need escaping in patterns.

```

exploration/text-vs-code.md
```diff
@@ -17,49 +19,361 @@ Status: **Proposed**
...
+Cons:
+- Probably not a serious alternative: the example
+  includes any number of obvious footguns that have to be addressed
```
eemeli, line 191: This seems really rather opinionated. In many ways, this is the same as the "Always quote" solution, except that the pattern delimiters are `}…{` instead of `{{…}}`. So the unnamed footguns probably apply to that alternative as well.

exploration/text-vs-code.md
```diff
@@ -17,49 +19,361 @@ Status: **Proposed**
...
+>```
+
+Pros:
+- Code is special instead of text.
+- Easy to use (best of both worlds?)
+
+Cons:
+- Requires one of the alternate syntaxes
+- Has two ways to represent a pattern.
+- May be difficult for translators to add quotes when needed.
```
eemeli, line 215: As far as I've been able to determine, there are exactly three scenarios in which a translator may need to add leading or trailing spaces to a pattern that starts out without them:
1. When translating a whole-sentence message from a CJK script to a non-CJK script, such that the sentences are concatenated into a single paragraph and need spaces between them. As with all other string concatenations, I would expect for this to be explicitly called out to the translator, so that they may know whether to add the space at the start or end of the pattern.
2. When translating a pattern to Chinese which ends up requiring a leading honorific space. As far as I can tell, this is really rare in dynamic message strings.
3. When the message is expected to be output using a monospace font and fakes either centering or right-alignment by using in-message spaces for indentation, and the first line of the pattern happens to be exactly the maximum length, and so does not need leading spaces. This is sufficiently rare that I'm pretty sure this is only a theoretical possibility, and in any case I'd expect it to be rather clearly called out to the translator.

Given that each of the above only has an impact on the pattern delimiting if the message _also_ has multiple variants _and_ if the translator is not using any tooling that'd take care of the delimiting _and_ if the developer has not pre-emptively delimited the pattern, I would be ok accepting this negative, especially as the downside would be a single missing space in the translation.

exploration/text-vs-code.md
```diff
@@ -17,49 +19,361 @@ Status: **Proposed**
...
+>{when many}   {{ This also has one space start and end. }}
+>{when *}{{You can quote patterns even without whitespace.}}
+>```
+
+Pros:
+- Code is special instead of text.
+- Easy to use (best of both worlds?)
+
+Cons:
+- Requires one of the alternate syntaxes
```
eemeli, line 213: Given that we just ran a "beauty contest" in which these "alternate syntaxes" were preferred over the current `main` syntax by an absolute majority of the participants, this could also be listed as one of the "Pros".

exploration/text-vs-code.md
```diff
@@ -17,49 +19,361 @@ Status: **Proposed**
...
+>       No amount of whitespace matters before this pattern
+>       but all of the whitespace at the end does.
+>
+>       {||}
+>```
+
+Pros:
+- Code is special, whitespace is not.
+- Makes PEWS into a "special event", alerting developers to the non-I18N aspects of it?
+
```
eemeli, line 248:
```suggestion
- Avoids as many escape sequences as possible,
  as `}` does not need escaping in patterns.
  
```

exploration/text-vs-code.md
```diff
@@ -17,49 +19,361 @@ Status: **Proposed**
...
+> This has no space or newline in front of it.
+>{when 1}
+>  This has no newline but does have one space in front of it.
+>{when few}
+>  This has no space or newline in front of it or at the end {when many}This has no spaces or newlines.
+>{when 11}
+>
+> This has a newline and a space at the start and a space-newline at the end
+>
+>>{when *}{|
```
eemeli, line 272:
```suggestion
>{when *}{|
```

exploration/text-vs-code.md
aphillips: It is opinionated. When I wrote it, I was being lazy by not enumerating the issues.

> this is the same as the "Always quote" solution, except that the pattern delimiters are }…{ instead of {{…}}. So the unnamed footguns probably apply to that alternative as well.

No, this is incorrect.

`{{...}}` encloses all and only the whitespace that is intentional in the pattern, with `{{` and `}}` forming the pattern boundary. These boundary characters are visible.

`}...{` makes all whitespace in the variant block meaningful. It effectively prohibits a multiline representation of a message, because the newlines are always meaningful. It also means that trailing spaces (which are invisible) have meaning. 

To make a message multiline, you have to put the whitespace inside the key.:

```
{{
match {$var}
{when 0
}This has no newline or space.{
when one}
This has a newline at the start.{
when *} This has a space at the start and six spaces and a newline at the end.      
}}
```

echeran:
> The distinction here matters, because we are looking to assign a novel meaning to a pair of characters that no other syntax than ICU MessageFormat uses to delimit localizable text. Every other syntax which uses braces in text uses them to delimit _code_. Which MF2 is also doing.

Remember our previous discussions from last year about this. This MF2 group chose curly braces _precisely because_ they are the least likely to occur in other syntaxes and in message patterns themselves.

exploration/text-vs-code.md
eemeli: If we take your example above, and after the `match` replace each `}` with `{{` and each `{` with `}}`, we get this:

```
{{
match {$var}
}}when 0
{{This has no newline or space.}}
when one{{
This has a newline at the start.}}
when *{{ This has a space at the start and six spaces and a newline at the end.      
}}
```

Ignoring the specifics of what's happening with the preamble, that seems pretty similar to me. It's just that we're conditioned to look at the `{` and `}` a certain way.

aphillips:
For the purposes of reading the document in our 2023-10-30 call, I'm merging this work now. This does not make the comments above right/wrong, relevant/irrelevant, or anything else. It's just to enable "easy reading".

exploration/text-vs-code.md
mihnita: It needs escaping if it is used as ending of the "wrapping" of the string

PR #504, opened 2023-10-27
Should we allow unquoted variant patterns?

aphillips:
Here's a different take (per Slack conversation) on the "Question 3" alternatives. The Pros/Cons might be better fleshed out. I didn't spend enough time on the case for unquoted variants. However, I think this might be clearer than trying to fit `text-vs-code`


aphillips:
This text was copied into #503, so closing this without merging it.

PR #508, opened 2023-10-31
Link to history in syntax-exploration-2.md

gibson042:
Syntax Exploration v2 refines options from the Beauty Contest, and should therefore link to that page.


aphillips (approved):

aphillips:
Thanks for the improvement.

PR #511, opened 2023-11-01
Clarify declaration immutability

aphillips:
Having reviewed @catamorphism's previous text and looking at our current text, these are proposed **_editorial_** changes to make variable immutability clearer in the spec. The previous text's "a _messages_ is not considered valid" doesn't read as proper English. We also lack an error for this (we have one for duplicate options), so I added one.

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -758,7 +758,7 @@ These are divided into the following categories:
     contains a _selector_ that does not have an _annotation_,
     or contains a _variable_ that does not directly or indirectly reference a _declaration_ with an _annotation_.
 
-    > Example invalid messages resulting in a _Missing Selector Annotation error_:
+    > Examples of invalid messages resulting in a _Missing Selector Annotation error_:
     >
     > ```
     > {{
@@ -786,11 +786,32 @@ These are divided into the following categories:
     > }}
     > ```
 
+  - **Duplicate Declaration errors** occur when a _variable_ appears in two _declarations_.
+    This includes when an _input-declaration_ binds a _variable_ that appears in a previous _declaration_,
+    when a _local-declaration_ binds a _variable_ that appears in a previous _declaration_,
+    or when a _local-declaration_ refers to its _variable_ in its _expression_.
+
+    > Examples of invalid messages resulting in a Duplicate Declaration error:
+    >
+    > ```
+    > {{
+    >    input {$var :number maxFractionDigits=0}
+    >    input {$var :number minFractionDigits=0}
+    >    {{Redeclaration of the same variable}}
+    > }}
+    > {{
+    >    local $var = {$ext :someFunction}
+    >    local $var = {$error}
+    >    local $var2 = {$var2 :error}
+    >    {{{$var} cannot be redefined. {$var2} cannot refer to itself}}
+    > }}
+    > ```
+
   - **Duplicate Option Name errors** occur when the same _name_
     appears on the left-hand side
     of more than one _option_ in the same _expression_.
 
-    > Example invalid messages resulting in a Duplicate Option Name error:
+    > Examples of invalid messages resulting in a Duplicate Option Name error:
     >
     > ```
     > Value is {42 :number style=percent style=decimal}
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -800,6 +800,16 @@ These are divided into the following categories:
     >    {{Redeclaration of the same variable}}
     > }}
     > {{
+    >    local $var = {$ext :number maxFractionDigits=0}
+    >    input {$var :number minFractionDigits=0}
+    >    {{Redeclaration of a local variable}}
+    > }}
+    > {{
+    >    input {$var :number minFractionDigits=0}
+    >    local $var = {$ext :number maxFractionDigits=0}
+    >    {{Redeclaration of an input variable}}
+    > }}
+    > {{
     >    local $var = {$ext :someFunction}
     >    local $var = {$error}
     >    local $var2 = {$var2 :error}
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -786,7 +786,7 @@ These are divided into the following categories:
     > }}
     > ```
 
-  - **Duplicate Declaration errors** occur when a _variable_ appears in two _declarations_.
+  - A **Duplicate Declaration error** occurs when a _variable_ appears in two _declarations_.
     This includes when an _input-declaration_ binds a _variable_ that appears in a previous _declaration_,
     when a _local-declaration_ binds a _variable_ that appears in a previous _declaration_,
     or when a _local-declaration_ refers to its _variable_ in its _expression_.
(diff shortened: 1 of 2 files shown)


spec/syntax.md
```diff
@@ -148,19 +148,35 @@ MAY include an _annotation_ that is applied to the external value.
...
 local-declaration = local s variable [s] "=" [s] expression
 ```
 
+_Variables_, once declared, MUST NOT be redeclared. 
+A _message_ that does any of the following is not _valid_ and will produce a 
+Duplicate Declaration error during formatting:
+- An _input-declaration_ MUST NOT bind a _variable_ that appears as a _variable_ in a previous 
+  _declaration_.
+- A _local-declaration_ MUST NOT bind a _variable_ that appears as a _variable_ in a previous
+  _declaration_.
```
catamorphism, line 163: I thought local declarations were allowed to override external variables? Or was that a rejected proposal?

spec/syntax.md
```diff
@@ -148,19 +148,35 @@ MAY include an _annotation_ that is applied to the external value.
...
+- A _local-declaration_ MUST NOT bind a _variable_ that appears in its _expression_.
+
+> [!Note]
+> These restrictions only apply to _declarations_.
+> A _placeholder_ or _selector_ MAY override the annotation provided in a _declaration_.
```
catamorphism, line 168: I think it's very confusing to call this "overriding", since no new name is being introduced. It's just constructing a new expression.

catamorphism (changes requested):

spec/syntax.md
aphillips: It's a little complicated and I should probably add text about the corner case here.

Currently you can annotate an external value using `input`:
```
#input {$var :function opt=val}
```

... and you can assign it to a `local` variable:
```
#local $foo = {$var :function opt=val}
```

... and you can overwrite an external variable **_so long as it is not declared_**:

```
#local $var = {$foo :function opt=val}  // overwrites any `$var` passed in
```

But not this:

```
#input {$var :function opt=val}
#local $var = {$foo :something opt=val}
```

spec/syntax.md
aphillips: Fair enough. Try the wording changes in the e9908c8.

spec/syntax.md
```diff
@@ -148,19 +148,39 @@ MAY include an _annotation_ that is applied to the external value.
...
+> These restrictions only apply to _declarations_.
+> A _placeholder_ or _selector_ MAY supply a different annotation than one
+> provided in a _declaration_ for the same _variable_.
+> For example, this message is _valid_:
+> ```
+> {{
+>    input {$var :number maxFractionDigits=0}
+>    match {$var :plural maxFractionDigits=2}
+>    when 0 {{The selector can re-annotate {$var}}}
+>    when * {{This pattern can re-annotate {$var :number maxFractionDigits=3}}}
```
catamorphism, line 179: The exposition is better, but I'm still confused about the semantics. In the scope of the `input` declaration (the scope where `$var` is defined), does `$var` refer to the result of applying the `:number` formatter to `$var` with the option `maxFractionDigits` bound to 0? Or does it just refer to whatever value was passed in for `$var`? If the latter is true, what's the point of applying the `:number` annotation in the `input` declaration? And if the former is true, what does it mean to apply the `:number` formatter to an already-formatted number (or is `:number` expected to check whether its argument is a raw number or a formatted number, and handle those cases differently?)

catamorphism (changes requested):

spec/syntax.md
aphillips: Our specification is explicit in that it doesn't say whether the declaration is early or late binding, which I think means that it doesn't matter. Another way to say that is that the `input` declaration means that any reference to `$var` should be interpreted through the function `:number` and the passed options.

Maybe the word `re-annotate` is not right? Perhaps:

```suggestion
>    input {$var :number maxFractionDigits=0}
>    match {$var :plural maxFractionDigits=2}
>    when 0 {{The selector can apply a different annotation to {$var} for the purposes of selection}}
>    when * {{A placeholder in a pattern can apply a different annotation to {$var :number maxFractionDigits=3}}}
```

ryzokuken (approved):

spec/formatting.md
```diff
@@ -786,11 +786,32 @@ These are divided into the following categories:
     > }}
     > ```
 
-  - **Duplicate Option Name errors** occur when the same _name_
+  - A **Duplicate Declaration error** occurs when a _variable_ appears in two _declarations_.
+    This includes when an _input-declaration_ binds a _variable_ that appears in a previous _declaration_,
+    when a _local-declaration_ binds a _variable_ that appears in a previous _declaration_,
+    or when a _local-declaration_ refers to its _variable_ in its _expression_.
```
stasm, line 792:
```suggestion
    or when a _local-declaration_ refers to its bound _variable_ in its _expression_.
```

spec/syntax.md
stasm: > And if the former is true, what does it mean to apply the :number formatter to an already-formatted number (or is :number expected to check whether its argument is a raw number or a formatted number, and handle those cases differently?)

We don't specify what type exactly the argument to `:number` should be, but we do say that (a) it should be the *resolved* value, and that (b) the shape of the resolved value is implementation-specific. If a "resolved value" is allowed by the implementation to be a raw value, then functions must accept such raw values as arguments, in addition to richer "formattable" data structures. 

Also, note that the same situation can be reproduced without `input`, using multiple `local` declarations:

```
local $a = {1 :number minFractionDigits=2} // formats as 1.00
local $b = {$a :number minFractionDigits=3} // formats as 1.000
```

stasm: Aside: I'm not sure if the spec is currently clear enough about the requirement to allow functions to inspect the raw value and the formatting options of their arguments. Specifically, while a function may eagerly format its operand to a string, it should (must?) also return the raw underlying value and the formatting options used for formatting, in case another function wants to extend them or use them for other logic. Consider:

```
local $a = {1 :number minIntegerDigits=3} // formats as 001.
local $b = {$a :number minFractionDigits=3} // formats as 001.000
// min integer digits are preserved from the previous call.
```

```
input {$item :noun case=accusative count=1}
local $colorMatchingGrammaticalNumberGenderCase = {$color :adjective accord=$item}
// :adjective inspects $item to learn about case=accusative, count=1
```

Should I file an issue about this?

spec/formatting.md
```diff
@@ -786,11 +786,32 @@ These are divided into the following categories:
...
+
+    > Examples of invalid messages resulting in a Duplicate Declaration error:
+    >
+    > ```
+    > {{
+    >    input {$var :number maxFractionDigits=0}
+    >    input {$var :number minFractionDigits=0}
+    >    {{Redeclaration of the same variable}}
+    > }}
```
stasm, line 801: Perhaps add examples of all possible error scenarios?

```suggestion
    > }}
    > {{
    >    local $var = {$ext :number maxFractionDigits=0}
    >    input {$var :number minFractionDigits=0}
    >    {{Redeclaration of a local variable}}
    > }}
    > {{
    >    input {$var :number minFractionDigits=0}
    >    local $var = {$ext :number maxFractionDigits=0}
    >    {{Redeclaration of an input variable}}
    > }}
```

stasm (approved):

PR #514, opened 2023-11-06
Change syntax to use RFC7405 for ABNF

aphillips:


spec/message.abnf
```diff
@@ -27,10 +27,10 @@ function = (":" / "+" / "-") name
...
 
 ; reserved keywords are always lowercase
-input = %x69.6E.70.75.74  ; "input"
-local = %x6C.6F.63.61.6C  ; "local"
-match = %x6D.61.74.63.68  ; "match"
-when  = %x77.68.65.6E     ; "when"
+input = "input"
+local = "local"
+match = "match"
+when  = "when"
```
stasm, line 33: Unfortunately, RFC 7405 states that:

> If no prefix is used, then case-insensitive matching is used (as is consistent with previous behavior).

```suggestion
input = %s"input"
local = %s"local"
match = %s"match"
when  = %s"when"
```

stasm (changes requested):

spec/message.abnf
stasm: (This is why tools which don't support 7405 can't parse it.)

spec/syntax.md
```diff
@@ -733,4 +733,18 @@ s = 1*( SP / HTAB / CR / LF )
 
 The grammar is formally defined in [`message.abnf`](./message.abnf)
 using the ABNF notation,
-as specified by [RFC 5234](https://datatracker.ietf.org/doc/html/rfc5234).
```
stasm, line 736: Nit: since 7405 is "structured as a set of changes to the full ABNF
   specification [RFC5234]", should we technically keep both references here? Or is 7405 enough?

stasm (approved):

spec/message.abnf
```diff
@@ -27,10 +27,10 @@ function = (":" / "+" / "-") name
...
 
 ; reserved keywords are always lowercase
-input = %x69.6E.70.75.74  ; "input"
-local = %x6C.6F.63.61.6C  ; "local"
-match = %x6D.61.74.63.68  ; "match"
-when  = %x77.68.65.6E     ; "when"
+input = %s"input"
+local = %s"local"
+match = %s"match"
+when  = %s"when"
```
eemeli, line 33: Why are these variables now needed? Couldn't their string values be used directly in the rules now referring to these variables?

aphillips: The `%s` is what makes the value case-sensitive. Plain strings are, for back-compat, case insensitive by default.

stasm: I think @eemeli was asking why not use `%s"input"` inline instead of indirectly through a named production:

```diff
-input-declaration = input [s] variable-expression
+input-declaration = %s"input" [s] variable-expression
```

One reason to not do it this way is that it would make it harder to apply the workaround mentioned in syntax.md for tools that dont' support 7405.

aphillips: Yes, that's why I did it that way. 

We might go back and make the change later, after we decide what to do with the code-mode sigil.

PR #517, opened 2023-11-07
Open/close design: Add two more alternatives using exact HTML syntax

stasm:
We should document that we considered using the exact HTML syntax, either completely not supported or parsed as part of the message.

I'm filing this PR as a draft, please feel free to extend it.


aphillips:
> We should document that we considered using the exact HTML syntax, either completely not supported or parsed as part of the message.

We do support using the exact HTML syntax as "just a literal" because we support literal text. Nothing we do will prevent (or _should_ prevent) people from writing messages like:

```
Hello <b>{$username}</b>!
```

or even:

```
Click <a href={$target}>here</a> to get {10 :percent} savings!
```

This has obvious problems for translators, for which a lot of tooling exists (but which could be considered less effective than something supported by the syntax). And it means different handling if you're running in JS. But it's not "wrong".

exploration/open-close-expressions.md
```diff
@@ -220,3 +220,68 @@ Rendered as React this could become:
...
 _How do they compare against the requirements?_
 _What other properties they have?_
+
+### Do Nothing
+
+We could not provide any special support for spans and markup elements,
+and instead delegate the problem completely to tools and downstream processing layers.
+
+```
+This is <html:strong>bold</html:strong> and this is <html:img alt="an image">.
```
aphillips, line 230: Why the `html:` prefix? 

```suggestion
This is <strong>bold</strong> and this is <img alt="an image" src="{$imgsrc}">.
```

exploration/open-close-expressions.md
```diff
@@ -220,3 +220,68 @@ Rendered as React this could become:
...
+```
+
+#### Pros:
+
+* No work required from us right now.
+* It's HTML.
+* The least surprising syntax for developers and translators.
+* Some CAT tools already support HTML in translations.
```
aphillips, line 238:
```suggestion
* Some CAT tools already support HTML and other markup in translations.
* Supports other markup or templating syntaxes with no additional effort (it's just strings)
```

exploration/open-close-expressions.md
```diff
@@ -220,3 +220,68 @@ Rendered as React this could become:
...
+
+#### Cons:
+
+* Requires quoting in XML-based containers.
```
aphillips, line 242: But this isn't any different from a string that happens to contain any of the syntax characters.

exploration/open-close-expressions.md
```diff
@@ -220,3 +220,68 @@ Rendered as React this could become:
...
+* Relies on a best-effort convention; is not a standard.
+* Markup becomes a completely alient concept in MessageFormat:
```
aphillips, line 244: Typo

```suggestion
Markup becomes a completely alien concept in MessageFormat:
```

exploration/open-close-expressions.md
```diff
@@ -220,3 +220,68 @@ Rendered as React this could become:
...
+  * It cannot be validated via the AST nor the reigstry.
+  * It cannot be protected, unless put inside literal expressions.
+  * It is not supported by `formatToParts`, which in turn makes double-parsing difficult.
+
```
aphillips, line 248:
```suggestion
* It requires special handling when inserting messages into the DOM.
* It requires "sniffing" the message to detect embedded markup. XSS prevention becomes much more complicated.

File: exploration/open-close-expressions.md [hunk 6]
```diff
@@ -220,3 +220,68 @@ Rendered as React this could become:
...
+### HTML Syntax
+
+We could parse the HTML syntax as part of MessageFormat parsing,
+and represent markup as first-class data-model concepts of MessageFormat.
+
+```
+This is <html:strong>bold</html:strong> and this is <html:img alt="an image">.
```
aphillips, line 255: The HTML prefix is still inappropriate here?

exploration/open-close-expressions.md
```diff
@@ -220,3 +220,68 @@ Rendered as React this could become:
...
+#### Pros:
+
+* Looks like HTML.
+* The least surprising syntax for developers and translators.
+
+#### Cons:
+
+* Looks like HTML, but isn't *exactly* HTML, unless we go to great lengths to make it so.
+  See the differences between HTML and React's JSX as a case-study of consequences.
+* Requires quoting in XML-based containers.
```
aphillips, line 287:
```suggestion
* Requires quoting in XML-based containers.
* It only supports HTML.
```

exploration/open-close-expressions.md
```diff
@@ -220,3 +220,68 @@ Rendered as React this could become:
 _What other solutions are available?_
 _How do they compare against the requirements?_
 _What other properties they have?_
+
+### Do Nothing
+
+We could not provide any special support for spans and markup elements,
+and instead delegate the problem completely to tools and downstream processing layers.
```
aphillips, line 227: "could not" is hard to parse?

```suggestion
Do not provide any special support for spannables or markup.
This would delegate the problem to tools and downstream processing layers.
```

exploration/open-close-expressions.md
```diff
@@ -220,3 +220,68 @@ Rendered as React this could become:
...
+
+```
+This is <html:strong>bold</html:strong> and this is <html:img alt="an image">.
+```
+
+#### Pros:
+
+* No work required from us right now.
```
aphillips, line 235:
```suggestion
* No work required from us right now. We can always add support later.
* We already support (and are required to support) in-line literal markup in messages, since they are just character sequences.

**Review by aphillips — CHANGES_REQUESTED** [review]
I think it is useful to document these options.

**Comment by stasm** [comment]
> This has obvious problems for translators, for which a lot of tooling exists (but which could be considered less effective than something supported by the syntax). And it means different handling if you're running in JS. But it's not "wrong".

Agreed. I wasn't trying to say that it was wrong, but instead I wanted to capture the discussion about why -- even if it's possible -- it's also not enough for some of our use-cases. The really interesting thing for me, however, is that a lot of use-cases that we listed in the doc can in fact be satisfied by the "do nothing" approach. This gives me pause.

Things start to get messy once we consider formatting to parts: `Hello <b>`, `$username`, `</b>!`. Even more so, when placeholders appear inside the HTML: `Click <a href=`, `$url`, `>here</a>` is really hard to parse.

Perhaps we're missing a few requirements?

**Comment by aphillips** [comment]
@stasm 
> Things start to get messy once we consider formatting to parts: Hello <b>, $username, </b>!. Even more so, when placeholders appear inside the HTML: Click <a href=, $url, >here</a> is really hard to parse.

Actually, format-to-parts (FTP) and the "click" example is really interesting. FTP would make that into three parts:

```
Click <a href="
{$url}  // as evaluated at runtime
">here</a>
```

The two text fragments cannot be turned into HTML cleanly on their own--only when the whole `<a>` tag is reassembled does it make sense. This is a case where spannables **_might_** provide something better than "do nothing"

> Perhaps we're missing a few requirements?

Maybe, but maybe not.

The compromise in Seville was basically: no operands, no to-string interpretation _by default_ (implementations may provide one). Not everyone agrees about the need for or importance of spannables nor about how to handle spannables. Implementations that don't support spannables could evaporate them in FTP or, as Eemeli suggests, return a "part" that the consumer can do whatever they want with. 

One thing I kind of like about the proposal is that it can support straightforward conversion to HTML/DOM without resorting to a registry of tags (there is no need to list the supported tags): the "function" is the HTML tag and the options are the attributes. You can create a DOM node out of it and fill it with content ("parts") until you see the close. You can call `getParent` when you see that close to restore state--the DOM provides the nesting for free. At the same time, it can also support Mihai's desire for attributed strings. MF2 doesn't care about validity.

File: exploration/open-close-expressions.md (hunk 6 above)
Comment (eemeli) [reply exploration/open-close-expressions.md:255]:
```suggestion
This is <strong>bold</strong> and this is <img alt="an image">.
```

exploration/open-close-expressions.md
stasm: I like this suggestion, but one comment: `supports` is saying too much. This approach *doesn't impede* markup expressed as regular text, but also does nothing to *support* it. It doesn't exist in the AST, in the data model, nor doesn't affect the result of `formatToString` not `formatToParts`.

exploration/open-close-expressions.md
stasm: Thanks for adding these. I agree with your last point. Re. the first two:

* "add support later" -- we'd need to carefully reserve syntax for the future addition and we'd need to do it now.
* "we already support" -- see my other comment from today; I wouldn't call it "support"; instead, perhaps "not impede" is a better phrasing.

exploration/open-close-expressions.md
stasm: Yes, but I think this is worth calling out explicitly because:

* HTML markup in translations is common, and
* storing translations in XML-based containers is common.

The combination of these two points means that if the syntax to get HTML markup in MF2 is the HTML syntax (because there's no alternative), then you'll get a lot of `&lt;` in XML-based containers.

exploration/open-close-expressions.md
stasm: I'm going to commit this suggestion and propose a slightly different wording in a follow-up comment.

exploration/open-close-expressions.md
stasm: I'm going to commit this suggestion and propose a slightly different wording in a follow-up comment.

eemeli (changes requested):

eemeli (approved):

eemeli:
I'm merging this, as it finally reflects the starting point for last week's deliberations. Then I'll file another PR on top of this that reflects the results of those deliberations, and the preliminary consensus on `{#foo}...{/foo}` that we reached last week.

aphillips:
@eemeli Done

PR #521, opened 2023-11-10
(Design) Code Mode Introducer choice

aphillips:
This design doc attempts to capture the options for beautifying the code mode introducer.

### DIFF
--- a/exploration/code-mode-introducer.md
+++ b/exploration/code-mode-introducer.md
@@ -1,0 +1,194 @@
+# Design Proposal: Choosing a Code Mode Introducer
+
+Status: **Proposed**
+
+<details>
+	<summary>Metadata</summary>
+	<dl>
+		<dt>Contributors</dt>
+		<dd>@aphillips</dd>
+		<dt>First proposed</dt>
+		<dd>2023-11-10</dd>
+		<dt>Pull Requests</dt>
+		<dd>#000</dd>
+	</dl>
+</details>
+
+## Objective
+
+_What is this proposal trying to achieve?_
+
+It must be possible to reliably parse messages.
+
+Our current syntax features unquoted patterns for simple messages
+and unquoted code tokens with quoted patterns for complex messages.
+Determining whether a message will have code tokens requires some
+special character sequence, either part of the code itself or
+prepended to the message.
+This proposal examines the options for determining code mode.
+
+## Background
+
+_What context is helpful to understand this proposal?_
+
+## Use-Cases
+
+_What use-cases do we see? Ideally, quote concrete examples._
+
+As a developer, I want to create messages with the minimal amount of special syntax.
+I don't want to have to type additional characters that add no value.
+I want the syntax to be logical and as consistent as possible.
+
+As a translator, I don't want to have to learn special syntax to support features such as declarations.
+
+As a user, I want my messages to be robust.
+
+As a user, I want to be able to see which messages are complex at a glance
+and to parse messages into their component parts visually as easily as possible.
+
+## Requirements
+
+_What properties does the solution have to manifest to enable the use-cases above?_
+
+## Constraints
+
+_What prior decisions and existing conditions limit the possible design?_
+
+## Proposed Design
+
+_Describe the proposed solution. Consider syntax, formatting, errors, registry, tooling, interchange._
+
+We need to choose one of these (or another option not yet considered).
+Presentation at UTW did not produce any opinions.
+
+Based on the pro/cons below, I would suggest Option D is possibly the best option?
+
+## Alternatives Considered
+
+_What other solutions are available?_
+_How do they compare against the requirements?_
+_What other properties they have?_
+
+There are the following designs being considered:
+
+### Option A. Use Pattern Quotes for Messages (current design)
+
+Complex messages are quoted with double curly brackets.
+The closing curly brackets might be optional.
+
+Sample pattern:
+```
+{{
+input {$var}
+match {$var}
+when * {{Pattern}}
+}}
+```
+Sample quoted pattern with no declarations or match:
+```
+{{{{Pattern}}}}
+```
+
+Pros:
+- Uses a sigil `{` already present in the syntax
+- No additional escapes
+- Consistent with other parts of the syntax?
+Cons:
+- Somewhat verbose
+- Closing portion of the syntax adds no value;
+  could be a source of unintentional syntax errors
+- Messages commonly end with four `}}}}`
+
+### Option B. Use a Sigil
+
+Complex messages start with a special sigil character.
+
+```
+#input {$var}
+match {$var}
+when * {{Pattern}}
+```
+Sample quoted pattern with no declarations or match:
+```
+#{{Pattern}}
+```
+
+Pros:
+- Requires minimum additional typing
+Cons:
+- Requires an additional sigil
+- Requires an additional escape for simple pattern start
+- Has no other purpose in the syntax
+
+### Option C. Use a Double Sigil
+
+Like Option B, except the sigil is doubled.
+
+```
+##input {$var}
+match {$var}
+when * {{Pattern}}
+```
+Sample quoted pattern with no declarations or match:
+```
+##{{Pattern}}
+```
+
+Pros:
+- Less likely to conflict with a simple pattern
+Cons:
+- Requires an additional sigil
+- Requires an additional escape for simple pattern start
+- Has no other purpose in the syntax
+
+### Option D. Sigilized Keywords
+
+Instead of quoting the message, adds a sigil to keywords that
+start statements, that is, `#input`, `#local` and `#match`.
+The keyword `when` might be considered separately.
+
+```
+#input {$var}
+#local $foo = {$bar}
+#match {$var}
+when * {{Pattern}}
+```
+Sample quoted pattern with no declarations or match:
+```
+{{Pattern}}
+```
+
+Pros:
+- Sigil is part of the keyword, not something separate
+- Requires minimum additional typing
+- Adds no characters to messages that consist of only a quoted pattern;
+  that is, quoting the pattern consists only of adding the `{{`/`}}` quotes
+- Maybe makes single-line messages easier to parse visually???
+Cons:
+- Requires an additional sigil
+- Requires an additional escape for simple pattern start
+
+### Option E. Special Sequence
+
+Like Option A except the sequence is closed locally (not at the end of the message).
+The suggested sequence is `{#}` but might be `{}` or `{{}}` also.
+
+```
+{#}input {$var}
+match {$var}
+when * {{Pattern}}
+```
+Sample quoted pattern with no declarations or match:
+```
+{#}{{Pattern}}
+```
+
+Pros:
+- Less likely to conflict with a simple pattern
+- Requires no additional sigil
+- Requires no additional escape
+Cons:
+- Has no other purpose in the syntax
+- Looks like something should happen inside it
+- Most additional typing
+  
--- a/exploration/code-mode-introducer.md
+++ b/exploration/code-mode-introducer.md
@@ -93,6 +93,7 @@ Pros:
 - Uses a sigil `{` already present in the syntax
 - No additional escapes
 - Consistent with other parts of the syntax?
+
 Cons:
 - Somewhat verbose
 - Closing portion of the syntax adds no value;
--- a/exploration/code-mode-introducer.md
+++ b/exploration/code-mode-introducer.md
@@ -119,6 +119,9 @@ Cons:
   could be a source of unintentional syntax errors
 - Messages commonly end with four `}}}}`
 
+> [!NOTE] Other enclosing sequences are also an option, notably `{%...%}` (or similar). 
+> This does reduce the number of curly brackets in a row.
+
 ### Option B. Use a Sigil
 
 Complex messages start with a special sigil character.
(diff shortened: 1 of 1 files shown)


exploration/code-mode-introducer.md
```diff
@@ -0,0 +1,194 @@
...
+Sample quoted pattern with no declarations or match:
+```
+{{{{Pattern}}}}
+```
+
+Pros:
+- Uses a sigil `{` already present in the syntax
+- No additional escapes
+- Consistent with other parts of the syntax?
+Cons:
```
gibson042, line 96: Markdown formatting correction (for all options):
```suggestion
- Consistent with other parts of the syntax?

Cons:
```

gibson042, line 96: Markdown formatting correction (the Pros list for each option must be followed by a blank line):
```suggestion
- Consistent with other parts of the syntax?

Cons:
```

aphillips: thanks!!

exploration/code-mode-introducer.md
```diff
@@ -0,0 +1,199 @@
...
+
+_What use-cases do we see? Ideally, quote concrete examples._
+
+As a developer, I want to create messages with the minimal amount of special syntax.
+I don't want to have to type additional characters that add no value.
+I want the syntax to be logical and as consistent as possible.
+
+As a translator, I don't want to have to learn special syntax to support features such as declarations.
+
+As a user, I want my messages to be robust.
```
echeran, line 44: nit: Define what "robust" means. Does it mean that the visual diff between the simple quoted pattern & a non-simple pattern is minimal? Or does it mean that there is an unambiguous 1:1 correspondence between a message (including its patterns and behavior logic) and its syntactical text representation? Or something else?

echeran (approved):
LGTM

eemeli (approved):
I agree with Option D being the overall best choice of those available, but the document should not present `#` as the sigil of choice. As discussed previously, it's such a common comment-start character that using it like this would be problematic.

I'd recommend either `!` or `%` as alternatives.

eemeli:
I'm not sure if it was captured earlier, but we probably shouldn't use `^` for anything, given that it's a [dead key](https://en.wikipedia.org/wiki/Dead_key) on a number of non-American keyboards.

exploration/code-mode-introducer.md
```diff
@@ -0,0 +1,224 @@
...
+		<dt>Pull Requests</dt>
+		<dd>#000</dd>
+	</dl>
+</details>
+
+## Objective
+
+_What is this proposal trying to achieve?_
+
+It must be possible to reliably parse messages.
```
catamorphism, line 21: "Reliable" is vague. Can this be framed as a property of the grammar? (For example: "The grammar for messages must be LL(k), so that it can be parsed without arbitrary lookahead.")

exploration/code-mode-introducer.md
```diff
@@ -0,0 +1,224 @@
...
+
+Our current syntax features unquoted patterns for simple messages
+and unquoted code tokens with quoted patterns for complex messages.
+Determining whether a message will have code tokens requires some
+special character sequence, either part of the code itself or
+prepended to the message.
```
catamorphism, line 27:
```suggestion
For ease of parsing, a distinguished character sequence
should denote the presence of code tokens in a message
(either part of the code itself or prepended to the message).
```

(Strictly speaking, it doesn't require that, since you could just scan for the first code token -- but that's undesirable since it requires arbitrary lookahead.)

exploration/code-mode-introducer.md
```diff
@@ -0,0 +1,224 @@
...
+
+## Constraints
+
+_What prior decisions and existing conditions limit the possible design?_
+
+Some of the options use a new sigil as part of the introducer.
+For various reasons, `#` has been used recently as a placeholder for this sigil.
+There are concerns that this character is not suitable, since it is used as a comment
+introducer in a number of formats. 
+See for example #520.
```
catamorphism, line 62:
```suggestion
See for example [#520](https://github.com/unicode-org/message-format-wg/issues/520).
```

exploration/code-mode-introducer.md
```diff
@@ -0,0 +1,224 @@
...
+The actual sigil used needs to be an ASCII character in the reserved or private use
+set (with syntax adjustments if we use up a private-use one).
+Most of the options below have been changed to use `^`, using 
+Apple's experimental syntax as a model for sigil choice.
```
catamorphism, line 66: Not sure what Apple's experimental syntax is -- can you add a link?

exploration/code-mode-introducer.md
```diff
@@ -0,0 +1,224 @@
...
+when * {{Pattern}}
+```
+Sample quoted pattern with no declarations or match:
+```
+{{Pattern}}
+```
+
+Pros:
+- Sigil is part of the keyword, not something separate
+- Requires minimum additional typing
```
catamorphism, line 185: I wouldn't say "minimum" here. Option B requires the absolute minimum additional typing, since it's just adding one character per message.

exploration/code-mode-introducer.md
```diff
@@ -0,0 +1,224 @@
...
+```
+
+Pros:
+- Less likely to conflict with a simple pattern
+- Requires no additional sigil
+- Requires no additional escape
+
+Cons:
+- Has no other purpose in the syntax
+- Looks like something should happen inside it
```
catamorphism, line 222: Not sure what that means?

catamorphism (approved):

exploration/code-mode-introducer.md
aphillips: I will clarify. This has to do with determining whether a message is simple or complex (and only that). Any given message must produce a consistent result for this.

exploration/code-mode-introducer.md
aphillips: I got this from @grhoten's presentation at the Unicode Technical Workshop this past week. I don't have a link handy, but one will be in the offing I think.

exploration/code-mode-introducer.md
aphillips: Actually, Option D adds zero characters. That's actually the minimum ;-).

I could say "Requires **no** additional typing". Also, note that this option avoids some of the iteration hazard of other options.

exploration/code-mode-introducer.md
aphillips: `{^}` looks like some sort of placeholder. So does `{}`. The sequence doesn't _do_ anything, but it looks like it might do something. If we chose this option, we would have to tell users why it's there.

aphillips:
@eemeli noted:
> I'm not sure if it was captured earlier, but we probably shouldn't use ^ for anything, given that it's a [dead key](https://en.wikipedia.org/wiki/Dead_key) on a number of non-American keyboards.

Ugh. You're right. Wave-dash/tilde anyone? `~`?

eemeli:
> Ugh. You're right. Wave-dash/tilde anyone? `~`?

Not available in Italian: https://superuser.com/a/667694

exploration/code-mode-introducer.md
stasm: I remember watching https://developer.apple.com/videos/play/wwdc2023/10153/ which demoed a syntax which used `^`: `^[Bienvenido](inflect: true) a tu iPhone`.

exploration/code-mode-introducer.md
```diff
@@ -0,0 +1,224 @@
...
+- Uses a sigil `{` already present in the syntax
+- No additional escapes
+- Consistent with other parts of the syntax?
+
+Cons:
+- Somewhat verbose
+- Closing portion of the syntax adds no value;
+  could be a source of unintentional syntax errors
+- Messages commonly end with four `}}}}`
+
```
stasm, line 121:
```suggestion
- The whole complex message looks similar to a placeholder, which may mislead users into thinking that it's OK to put text before and after it, or even to do things like `You have {{match {$count} when 1 {{one thingy}} when * {{{$count} thingies}}}} and {{match {$count} when 1 {{one gizmo}} when * {{{$count} gizmos}}}}`

```

exploration/code-mode-introducer.md
```diff
@@ -0,0 +1,224 @@
...
+## Alternatives Considered
+
+_What other solutions are available?_
+_How do they compare against the requirements?_
+_What other properties they have?_
+
+There are the following designs being considered:
+
+### Option A. Use Pattern Quotes for Messages (current design)
```
stasm, line 93: We could also consider slight variations of Option A which aren't exactly the current `main` syntax. For example: `{# ... #}`, `{% ... %}`, or `{[ ... ]}`. They could help mitigate at least the `}}}}` problem.

exploration/code-mode-introducer.md
```diff
@@ -0,0 +1,224 @@
...
+Complex messages start with a special sigil character.
+
+```
+^input {$var}
+match {$var}
+when * {{Pattern}}
+```
+Sample quoted pattern with no declarations or match:
+```
+^{{Pattern}}
```
stasm, line 133: In all options not using `{{` for code-mode, we could allow `{{Pattern}}` to be a valid shorthand for quoted patterns.

exploration/code-mode-introducer.md
```diff
@@ -0,0 +1,224 @@
...
+- Requires minimum additional typing
+- Adds no characters to messages that consist of only a quoted pattern;
+  that is, quoting the pattern consists only of adding the `{{`/`}}` quotes
+- Maybe makes single-line messages easier to parse visually???
+
+Cons:
+- Requires an additional sigil
+- Requires an additional escape for simple pattern start
+
+> [!Note] Unlike the other options, Option D is presented with `#` as the sigil.
```
stasm, line 194:
```suggestion
> [!NOTE]
> Unlike the other options, Option D is presented with `#` as the sigil.
```

exploration/code-mode-introducer.md
```diff
@@ -0,0 +1,224 @@
...
+Pros:
+- Less likely to conflict with a simple pattern
+- Requires no additional sigil
+- Requires no additional escape
+
+Cons:
+- Has no other purpose in the syntax
+- Looks like something should happen inside it
+- Most additional typing
+
```
stasm, line 224:
```suggestion

### Option F. Preamble

In this option, all declarations are placed in a dedicated block at the beginning of the message.
The preamble is the "front-matter" of the message, containing the message's logic.
`when` clauses are not part of the preamble.

The preamble can be delimited with `{% ... %}`:

    {%input {$var} match {$var}%} when * {{Pattern}}

Alternatively, it can be delimited with a new kind of delimiter, to make it visually distinct from placeholders and patterns:

    [[input {$var} match {$var}]] when * {{Pattern}}
    
We could also consider dropping the `when` keywords:

    [[input {$var} match {$var}]] * {{Pattern}}

Pros:
- Provides a clear conceptual distinction between declarations and variants.
- Visually, all code is grouped together.
- Unnests variant patterns. 

Cons:
- If `[[ ... ]]` is used to delimit the preamble, it will require `[[` to be escaped in patterns.

```

stasm (approved):
I've added another option, F, which proposes to place all declarations in a dedicated block for code, a preamble.

I'm OK with merging this PR as "proposed". I don't know yet if I agree that Option D is the best. Let's discuss on Monday.

exploration/code-mode-introducer.md
aphillips: We could, but then we'd be into having multiple equivalent representations?

exploration/code-mode-introducer.md
aphillips: Good callout, although it's still a lot of closing syntax `}}%}`

exploration/code-mode-introducer.md
stasm: Well, we already have two anyways: `Pattern` and `{{{{Pattern}}}}`. I'm merely describing options that we have. We could, for instance, also consider forbidding complex message syntax without declarations, so that only `Pattern` is valid.

I think the question to ask first is: given that `Pattern` is valid syntax and that we preserve whitespace, do we expect use-cases for quoting simple patterns?

It would be another way of ensuring leading and trailing whitespace is preserved, using only the MF2 syntax, regardless of what the host format offers. Consider:

```xml
<string xml:space="preserve">"   Hello   "</string>
```
```xml
<string>{{{{   Hello   }}}}</string>
```

eemeli: Depending on how you look at this, it would either mean supporting for a pattern within a pattern, or adding a third syntax under the trenchcoat. I'd really rather not. Either we leave the external spaces for the surrounding syntax, or we don't. Let's not do both if we don't need to.

aphillips:
I am merging this document as **_proposed_** and then creating a ballot around it.

macchiati:
It seems to me that there is a higher-level breakdown, where 'special sequence' below is a sequence of one or more ASCII punctuation/symbol characters.

Alpha: mark the start of a complex message with a special sequence, and either the end of the declarations or the the end of the complex message with a special sequence (Option A, F)

Beta: mark the start of a complex message with a special sequence (Option B, C, E)

Gamma: mark start of each keyword that starts statements with a special sequence (Option D)

Personally, I don't think either Alpha or Gamma add much value; and the key to Beta is to pick a sequence that is distinctive, doesn't collide with any other sequence, and very unlikely to start a simple message (thus probably a multi character sequence of symbols instead of a single symbol.
* By the way, we'd have a lot more choices for these if we considered non-ASCII — §§ would be a nice separator.

aphillips:
> By the way, we'd have a lot more choices for these if we considered non-ASCII — §§ would be a nice separator.

Sure, but non-ASCII is a non-starter for syntax. I can think of *lots* of characters that would be nice markers.......

macchiati:
I think the only real impediment for non-ASCII in syntax would be that the standard keyboards in some OS's don't yet provide easy access to even all the Latin 1 characters.

duerst:
> I think the only real impediment for non-ASCII in syntax would be that the standard keyboards in some OS's don't yet provide easy access to even all the Latin 1 characters.

"Easy access" is relative. The first level (easiest) of access is what comes 'engraved' on a physical keyboard. There are software keyboards, but I'd guess MF2 will mostly be written on PCs, not smartphones. Some keyboards have an AltGr key, which (as far as I understand) allows access to another 'shift' level of characters per key. But some keyboards don't have an AltGr key. Mine doesn't. I wouldn't know how to type most Latin-1 characters, except for some by switching to the Japanese IME, and entering the Japanese name of the character. It's a pity, because using a non-ASCII character such as § (we probably wouldn't even need two of them) would be really nice.

aphillips:
(as chair)

We're discussing code mode introducers in #526 now, as this PR is closed. Consider moving this discussion there.

PR #524, opened 2023-11-13
Implementing namespacing

aphillips:
Initial implementation of namespacing design into the spec.

This version makes the following pair of options parallel to one another (i.e. they are not duplicates):

```
{$foo :function option=value ns:option=value2}
```

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -202,15 +202,16 @@ the following steps are taken:
    If the registry does not define an implementation for this _name_,
    emit an Unknown Function error
    and use a _fallback value_ for the _expression_.
-4. If the _expression_ includes _options_, resolve the _options_ to a mapping
+3. If the _expression_ includes _options_, resolve the _options_ to a mapping
    of string identifiers to values.
    For each _option_:
-   - _Resolve the name_ of the option.
-   - If its right-hand side successfully resolves to a value,
-     bind the _name_ of the _option_ to the resolved value in the mapping.
+   - _Resolve the name_ of the _option_.
+   - If the _option_'s _name-part_ already exists in the resolved mapping of _options_,
+     emit a Duplicate Option Name error.
+   - If the _option_'s right-hand side successfully resolves to a value,
+     bind the _name-part_ of the _option_'s to the resolved value in the mapping.
    - Otherwise, do not bind the _name_ of the _option_ to any value in the mapping.
-6. Call the function implementation with the following arguments:
-
+4. Call the function implementation with the following arguments:
    - The current _locale_.
    - The resolved mapping of _options_.
    - If the _expression_ includes an _operand_, its resolved value.
@@ -220,18 +221,20 @@ the following steps are taken:
    An implementation MAY pass additional arguments to the function,
    as long as reasonable precautions are taken to keep the function interface
    simple and minimal, and avoid introducing potential security vulnerabilities.
-   Implementations SHOULD use an implementation-defined _namesapce_ for
-   any additional arguments exposed to users as _options_.
 
-An implementation MAY define its own functions.
-An implementation MAY allow custom functions to be defined by users.
+   Implementations MAY expose _options_ in an implementation-defined _namespace_
+   in addition to or superceding those found in the default registry.
+
+   An implementation MAY define its own functions.
+   An implementation MAY allow custom functions to be defined by users.
+
    Function access to the _formatting context_ MUST be minimal and read-only,
    and execution time SHOULD be limited.
    Implementation-defined _functions_ SHOULD use an implementation-defined _namespace_.
    User-defined _functions_ SHOULD be permitted to define the _namespace_ to use
    in name resolution.
 
-8. If the call succeeds,
+5. If the call succeeds,
    resolve the value of the _expression_ as the result of that function call.
    If the call fails or does not return a valid value,
    emit a Resolution error and use a _fallback value_ for the _expression_.
@@ -239,10 +242,10 @@ An implementation MAY allow custom functions to be defined by users.
 ### Name Resolution
 
 <dfn title="resolve the name">**_Name resolution_**</dfn> is the determination of
-the _name_ and the _namespace_ of the name for a given _function_
+the _name-part_ and, where present, the _namespace_ of the _name_ for a given _function_
 or _option_.
 
-To determine the _name_ of an item, the following steps are taken:
+To resolve the _name_ of an item, the following steps are taken:
 
 1. Let the _name_ be the string to be processed, less any preceeding
    sigils or markers.
@@ -251,10 +254,11 @@ To determine the _name_ of an item, the following steps are taken:
 2. Determine if _name_ contains the character U+003A COLON `:`:
    - If _name_ contains a colon, let _namespace_ be the
       substring of _name_ up to (but not including)
-      the colon and _name_ be the substring of _name_ following the colon;
-   - Else let _namespace_ be an empty value (null, void, or empty string).
-4. If the _name_ is empty, return a Syntax error.
-5. Return _namespace_ and _name_.
+      the colon and _name-part_ be the substring of _name_ following the colon;
+   - Else let _namespace_ be an empty value (null, void, or empty string)
+     and _name-part_ be the _name_.
+3. If the _name-part_ is empty, return a Syntax error.
+4. Return _namespace_ and _name-part_.
 
 Implementations are not required to support the installation or resolution
 of different namespaces.
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -197,14 +197,14 @@ the following steps are taken:
 
 1. If the _expression_ includes an _operand_, resolve its value.
    If this fails, use a _fallback value_ for the _expression_.
-2. _Resolve the name_ of the _function_ and, based on the starting sigil,
+2. Resolve the _identifier_ of the _function_ and, based on the starting sigil,
    find the appropriate function implementation to call.
    If the implementation cannot find the function, emit an Unknown Function error
    and use a _fallback value_ for the _expression_.
 
    An implementation MAY emit an Unknown Function error if the implementation    
-   does not support the _namespace_ of the _name_
-   or the _namespace_ does not support the _name-part_ named _function_.
+   does not support the _namespace_ of the _identifier_
+   or the _namespace_ does not support the named _function_.
 
    Implementations are not required to implement _namespaces_ or installable
    _function registries_.
@@ -212,12 +212,12 @@ the following steps are taken:
 4. If the _expression_ includes _options_, resolve the _options_ to a mapping
    of string identifiers to values.
    For each _option_:
-   - _Resolve the name_ of the _option_.
-   - If the _option_'s _name_ already exists in the resolved mapping of _options_,
+   - Resolve the _identifier_ of the _option_.
+   - If the _option_'s _identifier_ already exists in the resolved mapping of _options_,
      emit a Duplicate Option Name error.
    - If the _option_'s right-hand side successfully resolves to a value,
-     bind the _name_ of the _option_ to the resolved value in the mapping.
-   - Otherwise, do not bind the _name_ of the _option_ to any value in the mapping.
+     bind the _identifier_ of the _option_ to the resolved value in the mapping.
+   - Otherwise, do not bind the _identifier_ of the _option_ to any value in the mapping.
 4. Call the function implementation with the following arguments:
 
    - The current _locale_.
@@ -271,13 +271,13 @@ The _fallback value_ depends on the contents of the _expression_:
   > Example: `$user`
 
 - _expression_ with no _operand_:
-  the _function_ starting sigil followed by its _name_
+  the _function_ starting sigil followed by its _identifier_
 
   > Examples: `:platform`, `+tag`, `-tag`
 
 - Otherwise: The U+FFFD REPLACEMENT CHARACTER `�` character
 
-_Option_ names and values are not included in the _fallback value_.
+_Option_ identifiers and values are not included in the _fallback value_.
 
 When an error occurs in an _expression_ with a _variable_ _operand_
 and the _variable_ refers to a local _declaration_,
@@ -833,7 +833,7 @@ These are divided into the following categories:
     > }}
     > ```
 
-  - A **Duplicate Option Name error** occurs when the same _name_
+  - A **Duplicate Option Name error** occurs when the same _identifier_
     appears on the left-hand side
     of more than one _option_ in the same _expression_.
 
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -199,7 +199,9 @@ the following steps are taken:
    If this fails, use a _fallback value_ for the _expression_.
 2. Resolve the _identifier_ of the _function_ and, based on the starting sigil,
    find the appropriate function implementation to call.
-   If the implementation cannot find the function, emit an Unknown Function error
+   If the implementation cannot find the function,
+   or if the _identifier_ includes a _namespace_ that the implementation does not support,
+   emit an Unknown Function error
    and use a _fallback value_ for the _expression_.
 
    An implementation MUST emit an Unknown Function error
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -198,11 +198,17 @@ the following steps are taken:
 1. If the _expression_ includes an _operand_, resolve its value.
    If this fails, use a _fallback value_ for the _expression_.
 2. _Resolve the name_ of the _function_ and, based on the starting sigil,
-   find the appropriate function implementation from the _function registry_.
-   If the implementation does not support the _namespace_ of the _name_
-   or the _namespace_ does not support the _name-part_ named _function_,
-   emit an Unknown Function error
+   find the appropriate function implementation to call.
+   If the implementation cannot find the function, emit an Unknown Function error
    and use a _fallback value_ for the _expression_.
+
+   An implementation MAY emit an Unknown Function error if the implementation    
+   does not support the _namespace_ of the _name_
+   or the _namespace_ does not support the _name-part_ named _function_.
+
+   Implementations are not required to implement _namespaces_ or installable
+   _function registries_.
+
 4. If the _expression_ includes _options_, resolve the _options_ to a mapping
    of string identifiers to values.
    For each _option_:
@@ -226,58 +232,19 @@ the following steps are taken:
    An implementation MAY define its own functions.
    An implementation MAY allow custom functions to be defined by users.
 
+   An implementation MAY perform additional processing before calling the function.
+   For example, it can modify the _name_ of an _option_ to match the function's
+   input expectations (such as by removing the _namespace_ prefix from the _name-part_).
+
    Function access to the _formatting context_ MUST be minimal and read-only,
    and execution time SHOULD be limited.
+   
    Implementation-defined _functions_ SHOULD use an implementation-defined _namespace_.
 
-6. If the call succeeds,
+7. If the call succeeds,
    resolve the value of the _expression_ as the result of that function call.
    If the call fails or does not return a valid value,
    emit a Resolution error and use a _fallback value_ for the _expression_.
-
-### Name Resolution
-
-<dfn title="resolve the name">**_Name resolution_**</dfn> is the determination of
-the _name-part_ and, where present, the _namespace_ of the _name_ for a given _function_
-or _option_.
-
-To resolve the _name_ of an item, the following steps are taken:
-
-1. Let the _name_ be the string to be processed, less any preceeding
-   sigils or markers.
-   For example, in the _expression_ `{:function}`, the string `function`
-   is the _name_ string.
-2. Determine if _name_ contains the character U+003A COLON `:`:
-   - If _name_ contains a colon, let _namespace_ be the
-      substring of _name_ up to (but not including)
-      the colon and _name-part_ be the substring of _name_ following the colon;
-   - Else let _namespace_ be an empty value (null, void, or empty string)
-     and _name-part_ be the _name_.
-3. If the _name-part_ is empty, return a Syntax error.
-4. Return _namespace_ and _name-part_.
-
-Implementations are not required to support the installation or resolution
-of different namespaces.
-How namespaces are defined, provisioned, or handled is also implementation defined.
-The _namespace_ MAY affect which function is called.
-
-Example:
-> Suppose the ICU implementation of MessageFormat added a custom option `skeleton`
-> to the `:datetime` function.
-> Users of that implementation could invoke the datetime formatter with an
-> pattern similar to:
-> ```
-> {{Today is {$date :datetime icu:skeleton=yMMMd}.}}
-> ```
-> The `name` of the `datetime` function is resolved as:
-> * `name`     : `datetime`
-> * `namespace`: (blank)
-> * `name-part`: `datetime`
-> 
-> The `name` of the option in the expression is resolved as:
-> * `name`     : `icu:skeleton`
-> * `namespace`: `icu`
-> * `name-part`: `skeleton`
 
 
 ### Fallback Resolution
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -218,8 +218,8 @@ the following steps are taken:
      emit a Duplicate Option Name error.
    - If the _option_'s right-hand side successfully resolves to a value,
      bind the _identifier_ of the _option_ to the resolved value in the mapping.
-   - Otherwise, emit an appropriate error (e.g., an Unresolved Variable error) and
-     bind the _identifier_ of the _option_ to an unresolved value in the mapping.
+   - Otherwise, bind the _identifier_ of the _option_ to an unresolved value in the mapping.
+     (Note that an Unresolved Variable error will have been emitted.)
 4. Remove from the resolved mapping of _options_ any binding for which the value is an unresolved value.
 5. Call the function implementation with the following arguments:
 
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -197,17 +197,19 @@ the following steps are taken:
 
 1. If the _expression_ includes an _operand_, resolve its value.
    If this fails, use a _fallback value_ for the _expression_.
-2. Based on the _function_ starting sigil and _name_,
+2. _Resolve the name_ of the _function_ and, based on the starting sigil,
    find the appropriate function implementation from the _function registry_.
    If the registry does not define an implementation for this _name_,
    emit an Unknown Function error
    and use a _fallback value_ for the _expression_.
-3. Resolve the _option_ values to a mapping of string identifiers to values.
+4. If the _expression_ includes _options_, resolve the _options_ to a mapping
+   of string identifiers to values.
    For each _option_:
+   - _Resolve the name_ of the option.
    - If its right-hand side successfully resolves to a value,
      bind the _name_ of the _option_ to the resolved value in the mapping.
    - Otherwise, do not bind the _name_ of the _option_ to any value in the mapping.
-4. Call the function implementation with the following arguments:
+6. Call the function implementation with the following arguments:
 
    - The current _locale_.
    - The resolved mapping of _options_.
@@ -218,15 +220,63 @@ the following steps are taken:
    An implementation MAY pass additional arguments to the function,
    as long as reasonable precautions are taken to keep the function interface
    simple and minimal, and avoid introducing potential security vulnerabilities.
+   Implementations SHOULD use an implementation-defined _namesapce_ for
+   any additional arguments exposed to users as _options_.
 
-   As implementations MAY allow custom functions to be defined by users,
-   their access to the _formatting context_ SHOULD be minimal and read-only,
-   and their execution time SHOULD be limited.
+   An implementations MAY define its own functions or allow custom functions
+   to be defined by users.
+   Function access to the _formatting context_ SHOULD be minimal and read-only,
+   and execution time SHOULD be limited.
+   Implementation-defined _functions_ SHOULD use an implementation-defined _namespace_.
+   User-defined _functions_ SHOULD be permitted to define the _namespace_ to use
+   in name resolution.
 
-5. If the call succeeds,
+8. If the call succeeds,
    resolve the value of the _expression_ as the result of that function call.
    If the call fails or does not return a valid value,
    emit a Resolution error and use a _fallback value_ for the _expression_.
+
+### Name Resolution
+
+<dfn title="resolve the name">**_Name resolution_**</dfn> is the determination of
+the _name_ and the _namespace_ of the name for a given _function_
+or _option_.
+
+To determine the _name_ of an item, the following steps are taken:
+
+1. Let the _name_ be the string to be processed, less any preceeding
+   sigils or markers.
+   For example, in the _expression_ `{:function}`, the string `function`
+   is the _name_ string.
+2. Determine if _name_ contains the character U+003A COLON `:`:
+   - If _name_ contains a colon, let _namespace_ be the
+      substring of _name_ up to (but not including)
+      the colon and _name_ be the substring of _name_ following the colon;
+   - Else let _namespace_ be an empty value (null, void, or empty string).
+4. If the _name_ is empty, return a Syntax error.
+5. Return _namespace_ and _name_.
+
+Implementations are not required to support the installation or resolution
+of different namespaces.
+How namespaces are defined, provisioned, or handled is also implementation defined.
+It is important to note that the _namespace_ is not considered part of the _name_
+after resolution. 
+The _namespace_ MAY affect which function is called.
+
+Example:
+> Suppose the ICU implementation of MessageFormat added a custom option `skeleton`
+> to the `:datetime` function.
+> Users of that implementation could invoke the datetime formatter with an
+> expression similar to:
+> ```
+> {{Today is {$date :datetime icu:skeleton=yMMMd}.}}
+> ```
+> The `name` of the `datetime` function is resolved as:
+> > `namespace`: (blank)
+> > `name`: `datetime`
+> The `name` of the option in the expression is resolved as:
+> > `namespace`: `icu`
+> > `name`: `skelelton`
 
 ### Fallback Resolution
 
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -271,7 +271,7 @@ Example:
 > Suppose the ICU implementation of MessageFormat added a custom option `skeleton`
 > to the `:datetime` function.
 > Users of that implementation could invoke the datetime formatter with an
-> expression similar to:
+> pattern similar to:
 > ```
 > {{Today is {$date :datetime icu:skeleton=yMMMd}.}}
 > ```
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -223,17 +223,12 @@ the following steps are taken:
    as long as reasonable precautions are taken to keep the function interface
    simple and minimal, and avoid introducing potential security vulnerabilities.
 
-   Implementations MAY expose _options_ in an implementation-defined _namespace_
-   in addition to or superseding those found in the default registry.
-
    An implementation MAY define its own functions.
    An implementation MAY allow custom functions to be defined by users.
 
    Function access to the _formatting context_ MUST be minimal and read-only,
    and execution time SHOULD be limited.
    Implementation-defined _functions_ SHOULD use an implementation-defined _namespace_.
-   User-defined _functions_ SHOULD be permitted to define the _namespace_ to use
-   in name resolution.
 
 6. If the call succeeds,
    resolve the value of the _expression_ as the result of that function call.
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -217,8 +217,10 @@ the following steps are taken:
      emit a Duplicate Option Name error.
    - If the _option_'s right-hand side successfully resolves to a value,
      bind the _identifier_ of the _option_ to the resolved value in the mapping.
-   - Otherwise, do not bind the _identifier_ of the _option_ to any value in the mapping.
-4. Call the function implementation with the following arguments:
+   - Otherwise, emit an appropriate error (e.g., an Unresolved Variable error) and
+     bind the _identifier_ of the _option_ to an unresolved value in the mapping.
+4. Remove from the resolved mapping of _options_ any binding for which the value is an unresolved value.
+5. Call the function implementation with the following arguments:
 
    - The current _locale_.
    - The resolved mapping of _options_.
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -223,7 +223,7 @@ the following steps are taken:
    simple and minimal, and avoid introducing potential security vulnerabilities.
 
    Implementations MAY expose _options_ in an implementation-defined _namespace_
-   in addition to or superceding those found in the default registry.
+   in addition to or superseding those found in the default registry.
 
    An implementation MAY define its own functions.
    An implementation MAY allow custom functions to be defined by users.
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -202,10 +202,10 @@ the following steps are taken:
    If the implementation cannot find the function, emit an Unknown Function error
    and use a _fallback value_ for the _expression_.
 
-An implementation MUST emit an Unknown Function error
-if the implementation does not support _namespaces_
-or if the _namespace_ of the _identifier_ is not recognized by the implementation
-or the _namespace_ does not support the named _function_.
+   An implementation MUST emit an Unknown Function error
+   if the implementation does not support _namespaces_
+   or if the _namespace_ of the _identifier_ is not recognized by the implementation
+   or the _namespace_ does not support the named _function_.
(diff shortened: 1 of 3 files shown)


eemeli (changes requested):
A first pass, with some comments. Also, the list numbers are all messed up.

spec/formatting.md
```diff
@@ -218,16 +220,64 @@ the following steps are taken:
...
-   As implementations MAY allow custom functions to be defined by users,
-   their access to the _formatting context_ SHOULD be minimal and read-only,
-   and their execution time SHOULD be limited.
-
-5. If the call succeeds,
+   Implementations SHOULD use an implementation-defined _namesapce_ for
+   any additional arguments exposed to users as _options_.
+
+   An implementations MAY define its own functions or allow custom functions
+   to be defined by users.
```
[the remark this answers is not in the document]
aphillips: No. They should be separate statements.

```suggestion
An implementation MAY define its own functions.
An implementation MAY allow custom functions to be defined by users.
```

spec/formatting.md
```diff
@@ -218,16 +220,64 @@ the following steps are taken:
...
+   Function access to the _formatting context_ SHOULD be minimal and read-only,
+   and execution time SHOULD be limited.
```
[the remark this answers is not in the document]
aphillips: I didn't really change the requirement, but agree that we should.

spec/formatting.md
```diff
@@ -218,16 +220,64 @@ the following steps are taken:
...
    as long as reasonable precautions are taken to keep the function interface
    simple and minimal, and avoid introducing potential security vulnerabilities.
-
-   As implementations MAY allow custom functions to be defined by users,
-   their access to the _formatting context_ SHOULD be minimal and read-only,
-   and their execution time SHOULD be limited.
-
-5. If the call succeeds,
+   Implementations SHOULD use an implementation-defined _namesapce_ for
+   any additional arguments exposed to users as _options_.
```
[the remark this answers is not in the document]
aphillips: Yeah, this needed more work. Have another look.

spec/formatting.md
```diff
@@ -218,16 +220,64 @@ the following steps are taken:
...
    If the call fails or does not return a valid value,
    emit a Resolution error and use a _fallback value_ for the _expression_.
 
+### Name Resolution
+
+<dfn title="resolve the name">**_Name resolution_**</dfn> is the determination of
+the _name_ and the _namespace_ of the name for a given _function_
+or _option_.
+
+To determine the _name_ of an item, the following steps are taken:
```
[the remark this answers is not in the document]
aphillips: The colon is not part of the namespace or the name-part. It's just a separator.

spec/formatting.md
```diff
@@ -197,18 +197,21 @@ the following steps are taken:
...
+3. If the _expression_ includes _options_, resolve the _options_ to a mapping
+   of string identifiers to values.
    For each _option_:
-   - If its right-hand side successfully resolves to a value,
-     bind the _name_ of the _option_ to the resolved value in the mapping.
+   - _Resolve the name_ of the _option_.
+   - If the _option_'s _name-part_ already exists in the resolved mapping of _options_,
+     emit a Duplicate Option Name error.
+   - If the _option_'s right-hand side successfully resolves to a value,
+     bind the _name-part_ of the _option_'s to the resolved value in the mapping.
```
eemeli, line 212: Is this saying that formatting an expression `{:func one:foo=42 two:foo=13}` would generate an error? That would be very surprising, as we'd effectively be saying that the `ns:` part of an option name is completely meaningless.

spec/formatting.md
```diff
@@ -219,15 +222,66 @@ the following steps are taken:
    as long as reasonable precautions are taken to keep the function interface
    simple and minimal, and avoid introducing potential security vulnerabilities.
 
-   As implementations MAY allow custom functions to be defined by users,
-   their access to the _formatting context_ SHOULD be minimal and read-only,
-   and their execution time SHOULD be limited.
+   Implementations MAY expose _options_ in an implementation-defined _namespace_
+   in addition to or superceding those found in the default registry.
```
catamorphism, line 226:
```suggestion
   in addition to or superseding those found in the default registry.
```

spec/formatting.md
```diff
@@ -219,15 +222,66 @@ the following steps are taken:
...
+
+   An implementation MAY define its own functions.
+   An implementation MAY allow custom functions to be defined by users.
+
+   Function access to the _formatting context_ MUST be minimal and read-only,
+   and execution time SHOULD be limited.
+   Implementation-defined _functions_ SHOULD use an implementation-defined _namespace_.
+   User-defined _functions_ SHOULD be permitted to define the _namespace_ to use
+   in name resolution.
```
catamorphism, line 235: Does this mean that a user can define their own namespaces (e.g. introduce new namespaces with arbitrary names), or that they can choose a namespace from a set of existing namespaces?

spec/formatting.md
```diff
@@ -219,15 +222,66 @@ the following steps are taken:
...
+> expression similar to:
+> ```
+> {{Today is {$date :datetime icu:skeleton=yMMMd}.}}
+> ```
+> The `name` of the `datetime` function is resolved as:
+> > `namespace`: (blank)
+> > `name`: `datetime`
+> The `name` of the option in the expression is resolved as:
+> > `namespace`: `icu`
+> > `name`: `skelelton`
```
catamorphism, line 283:
```suggestion
> > `name`: `skeleton`
```

spec/formatting.md
```diff
@@ -219,15 +222,66 @@ the following steps are taken:
...
+How namespaces are defined, provisioned, or handled is also implementation defined.
+It is important to note that the _namespace_ is not considered part of the _name_
+after resolution. 
+The _namespace_ MAY affect which function is called.
+
+Example:
+> Suppose the ICU implementation of MessageFormat added a custom option `skeleton`
+> to the `:datetime` function.
+> Users of that implementation could invoke the datetime formatter with an
+> expression similar to:
```
catamorphism, line 274:
```suggestion
> pattern similar to:
```
(What follows is a pattern; alternately, you could leave this as-is and change the example to `{$date :datetime icu:skeleton=yMMMd}`)

spec/syntax.md
```diff
@@ -677,16 +677,44 @@ unquoted-start = name-start / DIGIT / "."
...
-as that would conflict with the _function_ start character.
+The namespace for _names_ is based on <cite>Namespaces in XML 1.0</cite>'s 
+[NCName](https://www.w3.org/TR/xml-names/#NT-NCName).
+This is different from XML's [Name](https://www.w3.org/TR/xml/#NT-Name)
+in that it MUST NOT start with `:`.
 Otherwise, the set of characters allowed in names is large.
 
+_Functions_ and _options_ MAY be preceded by a _namespace_ identifier
+which is separated from the body of the _name_ by a U+003A COLON `:`.
+Built-in _functions_ and _options_ do not have a _namespace_ identifier.
```
catamorphism, line 688: What if an implementation extends a built-in function with additional options -- could those implementation-specific options be namespaced?

spec/message.abnf
```diff
@@ -70,9 +70,10 @@ reserved-char  = %x00-08        ; omit HTAB and LF
...
 
-; based on https://www.w3.org/TR/xml/#NT-Name,
-; but cannot start with U+003A COLON ":"
-name = name-start *name-char
+; based on https://www.w3.org/TR/REC-xml-names/#NT-QName
+name       = [namespace ":"] name-body
+namespace  = name-part
+name-part  = name-start *name-char
 name-start = ALPHA / "_"
            / %xC0-D6 / %xD8-F6 / %xF8-2FF
```
gibson042, line 78: There is no definition for `name-body`, but this grammar is ambiguous if that is intended to be `name-part`—which ends in `*name-char` and therefore can itself include colons. For example, would parsing "foo:bar:baz" as a `name` be `{ namespace: "foo", name-part: "bar:baz" }` or `{ namespace: "foo:bar", name-part: "baz" }` or `{ name-part: "foo:bar:baz" }`?

spec/syntax.md
aphillips: Yes. In fact, we show this in the examples.

spec/formatting.md
```diff
@@ -219,15 +222,66 @@ the following steps are taken:
...
+> ```
+> {{Today is {$date :datetime icu:skeleton=yMMMd}.}}
+> ```
+> The `name` of the `datetime` function is resolved as:
+> > `namespace`: (blank)
+> > `name`: `datetime`
+> The `name` of the option in the expression is resolved as:
+> > `namespace`: `icu`
+> > `name`: `skeleton`
```
gibson042, line 283: There's some Markdown mangling here: https://github.com/unicode-org/message-format-wg/blob/1dc97bfcd95494e80de5a2b1e5ee62681f4c92e4/spec/formatting.md#name-resolution
```suggestion
> The `name` of the `datetime` function is resolved as:
> * `namespace`: (blank)
> * `name`: `datetime`
> 
> The `name` of the option in the expression is resolved as:
> * `namespace`: `icu`
> * `name`: `skeleton`
```

gibson042 (changes requested):

spec/formatting.md
aphillips: See: the description of this PR 😀. It does currently mean exactly this. 

I don't think it makes it meaningless, but much depends on how namespaced things (currently only functions and options) are implemented. If the `ns:` says, effectively, where to find the registry entry for validation but not, for example, the invocation, then it probably *is* an error to have `one:foo`/`two:foo` (you can only call one function and that one either expects the `foo` from `one:` or the `foo` from `two:`, but not both. This also alleviates the function from having to figure out that `one:foo` means "my version of `foo`". If the namespace is programmable, the called function has no way of knowing what it is.

Is there a function that would expect two separately-namespaced options of the otherwise-same name?

spec/message.abnf
stasm: I agree that we need to fix the ambiguity here. I think #519 is the right place to do it, so that we have a chance to holistically look at the syntax and consider other productions. We don't need to block the PR.

My preference, however, would be to make this PR complete, i.e. fix the ambiguity *somehow* before merging. Two options for temporary solutions:

* Use `/` as a temporary namespace separator.
* Make `namespace` consist only of `name-start` characters.

spec/formatting.md
stasm: > If the namespace is programmable, the called function has no way of knowing what it is.

I think that *function* namespaces and *option* namespaces should be controlled by different actors, because they serve different purposes.

* Function namespaces are defined **by i18n engineers** who integrate an MF2 implementation into a codebase. They help disambiguate between different packages providing formatters of the same name.
* Option namespaces are defined **by implementors of add-ons** to already existing functions. For example, `:datetime ns:skeleton=YYY` means that the implementation of `datetime` conforms to the default registry, and on top of that, it offers an extra option, `ns:skeleton`.

Note that functions themselves are a kind of a namespace for their options: In both `ns:func foo=bar`, `foo` isn't spelled as `ns:foo` because it's already effectively namespaced by the fact of being an option to `ns:func`. It's the add-on options that need namespacing, and that namespacing should not be programmable.

> Is there a function that would expect two separately-namespaced options of the otherwise-same name? (and, yes, I know this is different from the normal behavior of "namespaces")

As you note, two separately namespaced options are unlikely, because ultimately, there's only one implementation being invoked anyways.

However, we should consider how namespacing impacts other *unnamespaced* options. The namespacing of add-on options guarantees forward compatibility, in case the function being extended adds a new option in the future.

This is why functions should be able to see the namespaces of all their options.

spec/message.abnf
stasm: I see that the changes to `syntax.md` mention `NCName`. @aphillips do you intend to factor `NCName` into `message.abnf` in this PR?

spec/formatting.md
stasm: In other words: we probably don't need `one:foo=value` and `two:foo=value` to be different, but we do need `foo=value` and `one:foo=value` to be different because o forward compatibility of the add-on mechanism.

The easiest way to achieve it is to pass options together with their namespaces to the function during invocation. This in turn makes `one:foo=value` and `two:foo=value` different options.

eemeli: I agree with @stasm. I would expect that the `:func` implementation to get called with an options bag
```js
{ 'one:foo': '42', 'two:foo': '13' }
```

spec/formatting.md
```diff
@@ -219,15 +222,67 @@ the following steps are taken:
...
-   and their execution time SHOULD be limited.
+   Implementations MAY expose _options_ in an implementation-defined _namespace_
+   in addition to or superseding those found in the default registry.
+
+   An implementation MAY define its own functions.
+   An implementation MAY allow custom functions to be defined by users.
+
+   Function access to the _formatting context_ MUST be minimal and read-only,
+   and execution time SHOULD be limited.
+   Implementation-defined _functions_ SHOULD use an implementation-defined _namespace_.
```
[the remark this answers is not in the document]
aphillips: I'm thinking of going back to the design document and adding some use cases to illustrate these. In this case, I mean "non-default registry functions". I will add text to clarify.

---

Suppose you're Mihai and you're creating the ICU4J implementation. The `:datetime` function in the default registry might be backed by `com.ibm.icu.text.DateFormat` as an implementing class. There's no need to prefix `:datetime` for that. That's just implementation.

Mihai might want to expose `skeleton` though. That's not (currently) a default registry option. So he'd prefix it as `icu:skeleton` in his implementation.

Suppose he then goes on to implement `com.ibm.icu.text.DateIntervalFormat` as a function. That's not in the default registry, so it gets `:icu:dateinterval` and its options might or might not be namespaced.

Suppose Mihai goes on to implement an SPI for custom functions. Suppose I write my own date formatter. It doesn't replace the one in `:dateformat`, so it gets a prefix like `:app:dateformat`. The standard options might not be prefixed, but my custom options would be. ICU's options would probably be unrecognized to me. So you could see expressions like:

```
{$now :datetime icu:skeleton=yMMMd}
{$now :icu:dateinterval end=$then skeleton=yMMM}
{$now :app:datetime dateStyle=short :app:wonderfulness=high}
```

Suppose you're a tool implementer working with messages written for Mihai's implementation with my add-on. When you see a namespace prefix, that tells you that the function or the option is non-standard. The prefix might tell you where to look for the add-on registry or at least be something that you can key locally. Tool consumption is a key reason for namespacing.

aphillips:
@stasm noted:
> This is why functions should be able to see the namespaces of all their options.

I think implementation should make it possible for functions to see the namespaces. More specifically, there is probably glue code between the actual function and MessageFormat. That code can find out what its namespace is and select options based on that.

Is the argument here that I should change the algorithm to pack up the namespace on the options? We'd lose the duplicate checking between `one:foo` and `two:foo` (and `foo`) and make it the problem of the implementation to solve?

spec/formatting.md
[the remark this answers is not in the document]
stasm: Adding these use-cases in form of such user journeys sounds great. I'll be happy to contribute. Open a new PR?

stasm:
> Is the argument here that I should change the algorithm to pack up the namespace on the options? We'd lose the duplicate checking between `one:foo` and `two:foo` (and `foo`) and make it the problem of the implementation to solve?

I think so, yes. `foo` and `one:foo` should be different at the moment of the function's invocation, just in case `one:foo` comes first time-wise, and then `foo` is added to the upstream implementation of the function. I think generalizing this to all options means that `one:foo` and `two:foo` should also be different.

spec/formatting.md
```diff
@@ -219,15 +222,66 @@ the following steps are taken:
...
+To resolve the _name_ of an item, the following steps are taken:
+
+1. Let the _name_ be the string to be processed, less any preceeding
+   sigils or markers.
+   For example, in the _expression_ `{:function}`, the string `function`
+   is the _name_ string.
+2. Determine if _name_ contains the character U+003A COLON `:`:
+   - If _name_ contains a colon, let _namespace_ be the
+      substring of _name_ up to (but not including)
+      the colon and _name-part_ be the substring of _name_ following the colon;
```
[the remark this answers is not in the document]
aphillips: Removed the ability of colon to appear in `name`. Note that this makes our names match NCName (which I cite in this revision). This helps reduce the "sorta like an" XML name thing.

spec/message.abnf
aphillips: Removed colon (which I had intended in the first place 🤦🤷‍♂️🙈). As noted above, this makes us match XML names.

spec/message.abnf
```diff
@@ -70,15 +70,16 @@ reserved-char  = %x00-08        ; omit HTAB and LF
...
            / %x370-37D / %x37F-1FFF / %x200C-200D
            / %x2070-218F / %x2C00-2FEF / %x3001-D7FF
            / %xF900-FDCF / %xFDF0-FFFD / %x10000-EFFFF
-name-char  = name-start / DIGIT / "-" / "." / ":"
+name-char  = name-start / DIGIT / "-" / "."
            / %xB7 / %x300-36F / %x203F-2040
```
gibson042, line 83: This also affects `unquoted`, invalidating e.g. `{2023-11-15T17:23:42Z :datetime}` (which has been cited as desirable to some IIRC). But we're expecting unquoted literals to include strings that won't be valid variable/function/option names anyway (such as `-40`). So probably _this_ PR should get to a place like
```diff
-variable = "$" name
+variable = "$" name-part
 function = (":" / "+" / "-") name
 option = name [s] "=" [s] (literal / variable)
 …
 ; based on https://www.w3.org/TR/xml/#NT-Nmtoken,
 ; but cannot start with U+002D HYPHEN-MINUS or U+003A COLON ":"
-unquoted = unquoted-start *name-char
+unquoted = unquoted-start *unquoted-char
-unquoted-start = name-start / DIGIT / "."
+unquoted-start = name-part-start / DIGIT / "."
                / %xB7 / %x300-36F / %x203F-2040
+unquoted-char  = name-part-char / ":"
 …
-; based on https://www.w3.org/TR/xml/#NT-Name,
-; but cannot start with U+003A COLON ":"
-name = name-start *name-char
-name-start = ALPHA / "_"
-           / %xC0-D6 / %xD8-F6 / %xF8-2FF
-           / %x370-37D / %x37F-1FFF / %x200C-200D
-           / %x2070-218F / %x2C00-2FEF / %x3001-D7FF
-           / %xF900-FDCF / %xFDF0-FFFD / %x10000-EFFFF
-name-char  = name-start / DIGIT / "-" / "." / ":"
-           / %xB7 / %x300-36F / %x203F-2040
+; matches https://www.w3.org/TR/REC-xml-names/#NT-QName
+name            = [namespace ":"] name-part
+namespace       = name-part
+; matches https://www.w3.org/TR/REC-xml-names/#NT-NCName
+name-part       = name-part-start *name-part-char
+name-part-start = ALPHA / "_"
+                / %xC0-D6 / %xD8-F6 / %xF8-2FF
+                / %x370-37D / %x37F-1FFF / %x200C-200D
+                / %x2070-218F / %x2C00-2FEF / %x3001-D7FF
+                / %xF900-FDCF / %xFDF0-FFFD / %x10000-EFFFF
+name-part-char  = name-part-start / DIGIT / "-" / "."
+                / %xB7 / %x300-36F / %x203F-2040
```

spec/syntax.md
catamorphism: I think it might be clearer for this sentence to say something like "Built-in _functions_ and standard _options_ to built-in functions..."

I'm not sure if "standard" is the right word there, but I mean the options specified in [the standard registry](https://github.com/unicode-org/message-format-wg/blob/main/spec/registry.xml) (if it's correct to call that the standard registry).

spec/formatting.md
```diff
@@ -197,18 +197,23 @@ the following steps are taken:
 
 1. If the _expression_ includes an _operand_, resolve its value.
    If this fails, use a _fallback value_ for the _expression_.
-2. Based on the _function_ starting sigil and _name_,
+2. _Resolve the name_ of the _function_ and, based on the starting sigil,
    find the appropriate function implementation from the _function registry_.
-   If the registry does not define an implementation for this _name_,
+   If the implementation does not support the _namespace_ of the _name_
+   or the _namespace_ does not support the _name-part_ named _function_
+   or the registry does not otherwise support the _name_,
```
eemeli, line 204:
```suggestion
   or the _namespace_ does not support the _name-part_ named _function_,
```

Referring to the registry here would require it to concretely exist during the formatting.

spec/formatting.md
```diff
@@ -219,15 +224,68 @@ the following steps are taken:
    as long as reasonable precautions are taken to keep the function interface
    simple and minimal, and avoid introducing potential security vulnerabilities.
 
-   As implementations MAY allow custom functions to be defined by users,
-   their access to the _formatting context_ SHOULD be minimal and read-only,
-   and their execution time SHOULD be limited.
+   Implementations MAY expose _options_ in an implementation-defined _namespace_
+   in addition to or superseding those found in the default registry.
```
eemeli, line 228: I don't understand how including or not including this sentence changes what implementations and users are able or allowed to do.

spec/formatting.md
```diff
@@ -218,16 +220,64 @@ the following steps are taken:
...
+   Implementations SHOULD use an implementation-defined _namesapce_ for
+   any additional arguments exposed to users as _options_.
+
+   An implementations MAY define its own functions or allow custom functions
+   to be defined by users.
+   Function access to the _formatting context_ SHOULD be minimal and read-only,
+   and execution time SHOULD be limited.
+   Implementation-defined _functions_ SHOULD use an implementation-defined _namespace_.
+   User-defined _functions_ SHOULD be permitted to define the _namespace_ to use
+   in name resolution.
```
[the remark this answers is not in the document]
eemeli: I still don't understand what this means.

eemeli (changes requested):

spec/formatting.md
aphillips: This is the part that I *didn't* change from the original.

eemeli: Lol, you're right. So let's fix an earlier mistake.

spec/formatting.md
aphillips: Rmoeved

eemeli:
Re-reviewing the [design doc](https://github.com/unicode-org/message-format-wg/blob/main/exploration/overriding-extending-namespacing.md) on this, I now find the changes it's proposing to include in formatting and the data model a bit excessive. As discussed above (https://github.com/unicode-org/message-format-wg/pull/524#discussion_r1393277858), I think `foo`, `one:foo` and `two:foo` should always be distinct.

Reflecting back to the requirements presented for the design doc, could we just leave out the changes to formatting and the data model, retaining only the syntax and registry changes? As far as I can tell, that would fulfill all of the requirements, while leaving messy bits like name resolution as implementation-internal. I think we could continue consider `name` as just a string, which now might or might not include a `:` separated namespace at its start.

aphillips:
@eemeli Did you review the changes I made in the current revision? They are effectively what you're suggesting (`foo` and `*:foo` are distinct and `name` is that the formatting cares about). The main difference is that we do have a "resolve the name" step that checks if the name has a namespace (one might phrase it as "exactly zero or one colons"). This would allow a smart implementation to access the namespace when looking up the function. The spec is silent about how or whether that happens, though. Ditto options.

eemeli:
> [...] The main difference is that we do have a "resolve the name" step that checks if the name has a namespace (one might phrase it as "exactly zero or one colons"). This would allow a smart implementation to access the namespace when looking up the function. The spec is silent about how or whether that happens, though. Ditto options.

That's the bit that confuses me: Why are we defining parsing in the formatting section, when that's already in the syntax?

spec/formatting.md
aphillips: Done, although I added text permitting an implementation to do processing of the options.

aphillips:
> That's the bit that confuses me: Why are we defining parsing in the formatting section, when that's already in the syntax?

I put it in formatting because it's part of determining which function to call and what to call it with. The name can be syntactically correct, but the implementation needs to know about the parts if-and-only-if it supports namespacing. I followed your comment, though, and moved "resolving the name" to the syntax part of the spec and made changes to make all this namespacey stuff implementation defined in formatting.

spec/formatting.md
```diff
@@ -197,18 +197,28 @@ the following steps are taken:
...
-   emit an Unknown Function error
+2. _Resolve the name_ of the _function_ and, based on the starting sigil,
+   find the appropriate function implementation to call.
+   If the implementation cannot find the function, emit an Unknown Function error
    and use a _fallback value_ for the _expression_.
-3. Resolve the _option_ values to a mapping of string identifiers to values.
+
+   An implementation MAY emit an Unknown Function error if the implementation    
+   does not support the _namespace_ of the _name_
+   or the _namespace_ does not support the _name-part_ named _function_.
```
eemeli, line 207: This sentence does not add anything to the previous one. If either of the conditions here are true, then it's also true that the implementation "cannot find the function", and will therefore emit the error.

```suggestion
```

spec/formatting.md
```diff
@@ -197,18 +197,28 @@ the following steps are taken:
...
+
+   Implementations are not required to implement _namespaces_ or installable
+   _function registries_.
```
eemeli, line 210: This seems out of place. If we do want to mention something like this, it should not be here, in the middle of an algorithmic sequence of steps.

```suggestion
```

spec/formatting.md
```diff
@@ -197,18 +197,28 @@ the following steps are taken:
...
+
+4. If the _expression_ includes _options_, resolve the _options_ to a mapping
+   of string identifiers to values.
```
eemeli, line 213: Including the conditional here means that if there are no options, no (empty) mapping is defined. We should drop the conditional and rely on the implicit skipping of the "for each" loop as previously.

Also, the numbering got confused again.

```suggestion
3. Resolve the _options_ to a mapping of string identifiers to values.
```

spec/formatting.md
```diff
@@ -197,18 +197,28 @@ the following steps are taken:
...
-   - If its right-hand side successfully resolves to a value,
+   - _Resolve the name_ of the _option_.
+   - If the _option_'s _name_ already exists in the resolved mapping of _options_,
+     emit a Duplicate Option Name error.
+   - If the _option_'s right-hand side successfully resolves to a value,
      bind the _name_ of the _option_ to the resolved value in the mapping.
    - Otherwise, do not bind the _name_ of the _option_ to any value in the mapping.
-4. Call the function implementation with the following arguments:
-
+5. Call the function implementation with the following arguments:
```
eemeli, line 221:
```suggestion
4. Call the function implementation with the following arguments:

```

spec/formatting.md
```diff
@@ -219,15 +229,24 @@ the following steps are taken:
    as long as reasonable precautions are taken to keep the function interface
    simple and minimal, and avoid introducing potential security vulnerabilities.
 
-   As implementations MAY allow custom functions to be defined by users,
-   their access to the _formatting context_ SHOULD be minimal and read-only,
-   and their execution time SHOULD be limited.
+   An implementation MAY define its own functions.
+   An implementation MAY allow custom functions to be defined by users.
```
eemeli, line 234: This too is out of place, and should be stated elsewhere.

```suggestion
```

spec/formatting.md
```diff
@@ -219,15 +229,24 @@ the following steps are taken:
...
 
-5. If the call succeeds,
+   An implementation MAY perform additional processing before calling the function.
+   For example, it can modify the _name_ of an _option_ to match the function's
+   input expectations (such as by removing the _namespace_ prefix from the _name-part_).
+
```
eemeli, line 238: This feels a bit weird, and I don't think it should be allowed. The functions that an implementation will end up calling should be considered rather specific to that implementation, and therefore fully capable of dealing with their options without preprocessing.

For example, let's say an implementation provides an `icu:number` function that's mostly the same except for the core `number` but it adds a few options. That function may internally support both `icu:skeleton` and `skeleton` as synonyms for each other, but figuring that out ought to be on the function, rather than the implementation. E.g. detecting that both should not be defined at the same time ought to be a function concern.

```suggestion
```

spec/formatting.md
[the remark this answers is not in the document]
eemeli: Might be best to add a subsection to "Function Resolution" that collects these directions about what's possible and recommended for implementations and users to do with functions?

```suggestion
```

spec/formatting.md
```diff
@@ -219,15 +229,24 @@ the following steps are taken:
...
+   Function access to the _formatting context_ MUST be minimal and read-only,
+   and execution time SHOULD be limited.
+   
+   Implementation-defined _functions_ SHOULD use an implementation-defined _namespace_.
+
+7. If the call succeeds,
```
eemeli, line 244:
```suggestion
5. If the call succeeds,
```

eemeli (changes requested):
Getting better, but it's still very wordy. See line comments; probably best to look at the syntax ones before formatting.

spec/formatting.md
aphillips: I think of the function and the implementation of the function interchangeably. This permits an implementation to say to plug-in functionality that the namespacing will be managed for the function. But we could leave it out.

spec/formatting.md
aphillips: Where?

spec/syntax.md
```diff
@@ -677,16 +677,49 @@ unquoted-start = name-start / DIGIT / "."
 A **_<dfn>name</dfn>_** is an identifier for a _variable_ (prefixed with `$`),
 for a _function_ (prefixed with `:`, `+` or `-`),
 or for an _option_ (these have no prefix).
```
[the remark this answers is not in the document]
eemeli: So far, the syntax and its _definitions_ have kept the similarly named ones matching. With the text that you propose, _name_ and `name` no longer match exactly: The `variable` rule explicitly uses `name-part`, not `name`.

If there is a need for a common denominator for these (which I'm not convinced by, given that they don't share a namespace), then I'd much prefer _identifier_ or some other term that is not used in the ABNF.

eemeli:
> > Might be best to add a subsection to "Function Resolution" that collects these directions about what's possible and recommended for implementations and users to do with functions?
> 
> Probably this material goes in the registry specification?

That sounds like it could work. Looking at it now, `registry.md` does feel like it's missing a section on "what do you use this for".

spec/syntax.md
```diff
@@ -672,21 +672,65 @@ unquoted-start = name-start / DIGIT / "."
...
+in this release.
 
 ```abnf
 variable = "$" name
-function = (":" / "+" / "-") name
+function = (":" / "+" / "-") identifier
+option = identifier [s] "=" [s] (literal / variable)
 
-name = name-start *name-char
+identifier = [namespace ":"] name-part
```
eemeli, line 731:
```suggestion
identifier = [namespace ":"] name
```

eemeli (approved):
My significant prior concerns seem addressed now. There's still stuff included that I think is a bit out of place, but it's less important, and we can fix it later.

It would be good to get clarification/resolution on [this comment](https://github.com/unicode-org/message-format-wg/pull/524/files#r1394576082) by @gibson042 on `unquoted`, but I'm not sure I have strong feelings about how it ought to be resolved.

Also, an inline typo fix as a comment.

spec/message.abnf
```diff
@@ -45,12 +45,8 @@ quoted-char = %x0-5B         ; omit \
             / %x7D-D7FF      ; omit surrogates
             / %xE000-10FFFF
 
-; based on https://www.w3.org/TR/xml/#NT-Nmtoken,
-; but cannot start with U+002D HYPHEN-MINUS or U+003A COLON ":"
-unquoted = unquoted-start *name-char
+unquoted       = unquoted-start *name-char
 unquoted-start = name-start / DIGIT / "."
-               / %xB7 / %x300-36F / %x203F-2040
```
eemeli, line 52: Hang on, what happened here? Why are we further restricting `unquoted-start` from what it was before? On the call we talked about _expanding_ `unquoted` to something like:

```abnf
unquoted = unquoted-start *unquoted-char
unquoted-char = name-char / ":"
```

There is no reason to leave out any of these characters from `unquoted-start`, and if there is, that's beyond the scope of this PR.

aphillips: Unquoted cannot start with `:` or it would be a function identifier. The character we want to add is `-`.

The bottom part is my mistake.

aphillips: One other note: adding `-` would make `unquoted-start` identical to `name-char`, so `unquoted` becomes just `*name-char`

eemeli: > Unquoted cannot start with `:` or it would be a function identifier. The character we want to add is `-`.

Agreed, `unquoted` cannot start with `:`. But the `unquoted-char` that I propose introducing is for characters after the first, which before this PR _could_ include `:` because it was a permitted `name-char` character. As it no longer is, `unquoted-char` adds it back in so that this PR does not change anything about `unquoted`.

Adding `-` to `unquoted-start` is beyond the scope of this PR, and depends on resolving open-close. At earliest it should be done in the PR implementing open-close in the spec.

> The bottom part is my mistake.

Ah! Thank you for fixing it already.

gibson042: I think we should leave leading `-` in the other ticket and just ensure here that unquoted literals can start with decimal digits while names cannot.

eemeli (approved):
I think it's all groovy now.

aphillips:
@gibson042 ... over to you 😉

spec/message.abnf
```diff
@@ -70,15 +67,16 @@ reserved-char  = %x00-08        ; omit HTAB and LF
                / %x7E-D7FF      ; omit surrogates
                / %xE000-10FFFF
 
-; based on https://www.w3.org/TR/xml/#NT-Name,
-; but cannot start with U+003A COLON ":"
-name = name-start *name-char
+; based on https://www.w3.org/TR/REC-xml-names/#NT-QName
```
gibson042, line 70:
```suggestion
; identifier matches https://www.w3.org/TR/REC-xml-names/#NT-QName
; name matches https://www.w3.org/TR/REC-xml-names/#NT-NCName
```

gibson042 (approved):
I object to the optionality of an Unknown Function error for a function with unrecognized namespace or a function that does not exist in the claimed namespace. Everything else looks good, modulo a few editorial suggestions.

spec/formatting.md
```diff
@@ -197,16 +197,27 @@ the following steps are taken:
...
-   emit an Unknown Function error
+2. Resolve the _identifier_ of the _function_ and, based on the starting sigil,
+   find the appropriate function implementation to call.
+   If the implementation cannot find the function, emit an Unknown Function error
    and use a _fallback value_ for the _expression_.
-3. Resolve the _option_ values to a mapping of string identifiers to values.
+
+   An implementation MAY emit an Unknown Function error if the implementation    
+   does not support the _namespace_ of the _identifier_
+   or the _namespace_ does not support the named _function_.
```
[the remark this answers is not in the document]
aphillips: You're correct. This should be `MUST`. I think the idea was that implementations are not required to support installable namespaces, but in that case it should say something like:

```suggestion
An implementation MUST emit an Unknown Function error
if the implementation does not support _namespaces_
or if the _namespace_ of the _identifier_ is not recognized by the implementation
or the _namespace_ does not support the named _function_.
```

spec/formatting.md
```diff
@@ -197,16 +197,27 @@ the following steps are taken:
...
-   - If its right-hand side successfully resolves to a value,
-     bind the _name_ of the _option_ to the resolved value in the mapping.
-   - Otherwise, do not bind the _name_ of the _option_ to any value in the mapping.
+   - Resolve the _identifier_ of the _option_.
+   - If the _option_'s _identifier_ already exists in the resolved mapping of _options_,
+     emit a Duplicate Option Name error.
+   - If the _option_'s right-hand side successfully resolves to a value,
+     bind the _identifier_ of the _option_ to the resolved value in the mapping.
+   - Otherwise, do not bind the _identifier_ of the _option_ to any value in the mapping.
 4. Call the function implementation with the following arguments:
```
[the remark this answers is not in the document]
aphillips: Good callout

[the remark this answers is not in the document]
eemeli: The appropriate-error emission is already taken care of during the resolution of the right-hand value, in Variable Resolution. So if it's noted here, we should not say something like "emit an error" here because that would mean that it needs to be emitted twice.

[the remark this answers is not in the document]
eemeli: @aphillips Our messages crossed paths; see my comment above for a problem this language introduces.

spec/formatting.md
```diff
@@ -197,17 +197,31 @@ the following steps are taken:
...
+   find the appropriate function implementation to call.
+   If the implementation cannot find the function, emit an Unknown Function error
    and use a _fallback value_ for the _expression_.
-3. Resolve the _option_ values to a mapping of string identifiers to values.
+
+   An implementation MUST emit an Unknown Function error
+   if the implementation does not support _namespaces_
+   or if the _namespace_ of the _identifier_ is not recognized by the implementation
+   or the _namespace_ does not support the named _function_.
+
```
eemeli, line 209: I still don't see what this is adding to the preceding "If the implementation cannot find the function, emit an Unknown Function error". What sort of a situation would lead to that sentence not being sufficient to also cover everything added here?

Also, the unconditionality of its start now says that an implementation MUST "implement _namespaces_", or always emit an error at this point.

```suggestion
```

spec/formatting.md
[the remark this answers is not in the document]
aphillips: See change below.

spec/formatting.md
aphillips: it should say "if a namespace is present and the implementation does not support namespaces". Is that a sufficiently separate consideration from "can't find the function"?

eemeli: The question I'm trying to ask is, if a namespace is present and the implementation does not support namespaces, how can it find the corresponding function?

aphillips: It can't, which is they the MUSTard says "emit an error if you see a namespace and don't support namespaces"

spec/formatting.md
```diff
@@ -197,17 +197,31 @@ the following steps are taken:
 
 1. If the _expression_ includes an _operand_, resolve its value.
    If this fails, use a _fallback value_ for the _expression_.
-2. Based on the _function_ starting sigil and _name_,
-   find the appropriate function implementation from the _function registry_.
-   If the registry does not define an implementation for this _name_,
-   emit an Unknown Function error
+2. Resolve the _identifier_ of the _function_ and, based on the starting sigil,
+   find the appropriate function implementation to call.
+   If the implementation cannot find the function, emit an Unknown Function error
```
eemeli, line 202: @aphillips Would this change (together with removing the next paragraph) be minimally sufficient to covering all cases? Not supporting _any_ namespaces and not supporting _this_ namespace are pretty much the same at this point.

```suggestion
   If the implementation cannot find the function,
   or if the _identifier_ includes a _namespace_ that the implementation does not support,
   emit an Unknown Function error
```

gibson042 (approved):
LGTM!

spec/formatting.md
eemeli: Now that we have 023b5a8, could this be removed?

eemeli (changes requested):
Marking as "Request changes" just to make sure this doesn't get merged without https://github.com/unicode-org/message-format-wg/pull/524#discussion_r1408389224 being resolved; otherwise this is still fine.

eemeli (approved):
🎉

spec/formatting.md
```diff
@@ -197,17 +197,28 @@ the following steps are taken:
...
-   - Otherwise, do not bind the _name_ of the _option_ to any value in the mapping.
-4. Call the function implementation with the following arguments:
+   - Resolve the _identifier_ of the _option_.
+   - If the _option_'s _identifier_ already exists in the resolved mapping of _options_,
+     emit a Duplicate Option Name error.
+   - If the _option_'s right-hand side successfully resolves to a value,
+     bind the _identifier_ of the _option_ to the resolved value in the mapping.
+   - Otherwise, bind the _identifier_ of the _option_ to an unresolved value in the mapping.
+     (Note that an Unresolved Variable error will have been emitted.)
+4. Remove from the resolved mapping of _options_ any binding for which the value is an unresolved value.
```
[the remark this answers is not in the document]
eemeli: This is a small fix unrelated to namespacing that's bundled in here to ensure that if
```
{:foo opt=$none opt=bar}
```

is formatted with `$none` not resolving, not only will that cause an Unresolved Variable error to be emitted, but that a Duplicate Option Name error will also be emitted -- the latter was previously skipped.

The behaviour around options with errors is unchanged from the previous; it's just expressed a bit differently. If you think that ought to be changed, then that should go in a different PR as this is scope-creepy enough already.

eemeli:
> Do we want namespaces to be a separate concept in the [data model](https://github.com/unicode-org/message-format-wg/tree/main/spec/data-model)? Or should the data model just store the stringified _identifier_, i.e. `ns:name`?

I would prefer not adding namespaces to the data model, and continuing there with a single-string `name`.

spec/formatting.md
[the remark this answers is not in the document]
stasm: I filed #545 to continue this.

PR #528, opened 2023-11-20
Copy user story from #524 to design doc

aphillips:
Per a comment from @stasm, saving a user story from a [comment in #524](https://github.com/unicode-org/message-format-wg/pull/524#discussion_r1394423131) into the design doc.

### DIFF
--- a/exploration/overriding-extending-namespacing.md
+++ b/exploration/overriding-extending-namespacing.md
@@ -101,6 +101,45 @@ _What use-cases do we see? Ideally, quote concrete examples._
 - Translators and tools would like a machine-readable way to find out the names
   and option values for add-on packages.
 
+---
+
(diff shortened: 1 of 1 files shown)


eemeli (approved):

stasm:
A kind request for the future: please wait at least one full day between opening a PR and merging it. This way everyone can review it during their daytime.

duerst:
@stasm wrote:
> A kind request for the future: please wait at least one full day between opening a PR and merging it. This way everyone can review it during their daytime.

Very strongly agree. Even more, please consider the fact that some countries may have holidays when others don't, and that for some people, it may already be the weekend even if it's not yet for you, and that some people have day jobs or family emergencies even if you don't,.... Many thanks in advance!

PR #529, opened 2023-11-21
Implement code more introducer

eemeli:
This starts implementing the changes for the [code mode introducer](https://github.com/unicode-org/message-format-wg/blob/main/exploration/code-mode-introducer.md) (#521, #526), including dropping `when`.

Before I get any further, it might be best for us to align on the exact ABNF changes. See line comments for a few specific questions.

### DIFF
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -1,18 +1,30 @@
-message = pattern / complex-message
+message = simple-message / complex-message
 
-complex-message = "{{" [s] *(declaration [s]) body [s] "}}"
+simple-message = [almost-keyword]
+               / ((almost-keyword not-keyword / text-start / expression) pattern)
+complex-message = *(declaration [s]) body
 
-declaration = input-declaration / local-declaration
+declaration = input-declaration / local-declaration / reserved-declaration
 input-declaration = input [s] variable-expression
 local-declaration = local s variable [s] "=" [s] expression
+reserved-declaration = "." 4*(%x61-7A) reserved-body 1*expression
+
+almost-keyword = "." *3(%x61-7A)
+not-keyword    = %x0-5B        ; omit \
+               / %x5D-60       ; omit a-z and {
+               / %x7C          ; omit }
+               / %x7E-D7FF     ; omit surrogates
+               / %xE000-10FFFF
+               / expression
 
 body = quoted-pattern
      / (selectors 1*([s] variant))
 
 quoted-pattern = "{{" pattern "}}"
 pattern = *(text / expression)
+
 selectors = match 1*([s] expression)
-variant = when 1*(s key) [s] quoted-pattern
+variant = 1*(s key) [s] quoted-pattern
 key = literal / "*"
 
 expression = literal-expression / variable-expression / function-expression
@@ -26,18 +38,19 @@ variable = "$" name
 function = (":" / "+" / "-") name
 option = name [s] "=" [s] (literal / variable)
 
-; reserved keywords are always lowercase
-input = %s"input"
-local = %s"local"
-match = %s"match"
-when  = %s"when"
+; keywords always consist of a . followed by at least four lower-case a-z letters
+input = %s".input"
+local = %s".local"
+match = %s".match"
 
 text = 1*(text-char / text-escape)
-text-char = %x0-5B         ; omit \
-          / %x5D-7A        ; omit {
-          / %x7C           ; omit }
-          / %x7E-D7FF      ; omit surrogates
-          / %xE000-10FFFF
+text-start = %x0-2D         ; omit .
+           / %x2F-5B        ; omit \
+           / %x5D-7A        ; omit {
+           / %x7C           ; omit }
+           / %x7E-D7FF      ; omit surrogates
+           / %xE000-10FFFF
+text-char  = text-start / "."
 
 quoted      = "|" *(quoted-char / quoted-escape) "|"
 quoted-char = %x0-5B         ; omit \
@@ -60,7 +73,7 @@ private-start  = "^" / "&"
 ; future versions of this specification
 reserved       = reserved-start reserved-body
 reserved-start = "!" / "@" / "#" / "%" / "*" / "<" / ">" / "/" / "?" / "~"
-reserved-body  = *( [s] 1*(reserved-char / reserved-escape / quoted))
+reserved-body  = *([s] 1*(reserved-char / reserved-escape / quoted))
 
 reserved-char  = %x00-08        ; omit HTAB and LF
                / %x0B-0C        ; omit CR
@@ -81,7 +94,7 @@ name-start = ALPHA / "_"
 name-char  = name-start / DIGIT / "-" / "." / ":"
            / %xB7 / %x300-36F / %x203F-2040
 
-text-escape     = backslash ( backslash / "{" / "}" )
+text-escape     = backslash ( backslash / "." / "{" / "}" )
 quoted-escape   = backslash ( backslash / "|" )
 reserved-escape = backslash ( backslash / "{" / "|" / "}" )
 backslash       = %x5C ; U+005C REVERSE SOLIDUS "\"
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -40,7 +40,7 @@ simple-start-char = %x0-2D         ; omit .
                   / %x7C           ; omit }
                   / %x7E-D7FF      ; omit surrogates
                   / %xE000-10FFFF
-text-char = simple-start-char-start / "."
+text-char = simple-start-char / "."
 
 quoted      = "|" *(quoted-char / quoted-escape) "|"
 quoted-char = %x0-5B         ; omit \
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -56,10 +56,11 @@ unquoted-start = name-start / DIGIT / "."
 
 
 ; Reserve additional .keywords for use by future versions of this specification.
+reserved-statement = reserved-keyword [s [reserved-body [s]]] 1*expression
 ; Note that the following expression is a simplification,
-; as this rule MUST NOT be considered to match existing keywords:
-;   (input / local / match) (s / "{")
-reserved-statement = "." reserved-body 1*expression
+; as this rule MUST NOT be considered to match existing keywords.
+; As regexp, this would be: /\.(?!input|local|match)[a-z]+/
+reserved-keyword = "." 1*(%x61-7A)
 
 ; Reserve additional sigils for use by future versions of this specification.
 reserved-annotation = reserved-annotation-start reserved-body
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -56,11 +56,11 @@ unquoted-start = name-start / DIGIT / "."
 
 
 ; Reserve additional .keywords for use by future versions of this specification.
-reserved-statement = reserved-keyword [s [reserved-body [s]]] 1*expression
+reserved-statement = reserved-keyword [s reserved-body] 1*([s] expression)
 ; Note that the following expression is a simplification,
-; as this rule MUST NOT be considered to match existing keywords.
-; As regexp, this would be: /\.(?!input|local|match)[a-z]+/
-reserved-keyword = "." 1*(%x61-7A)
+; as this rule MUST NOT be considered to match existing keywords
+; (`.input`, `.local`, and `.match`).
+reserved-keyword = "." name
 
 ; Reserve additional sigils for use by future versions of this specification.
 reserved-annotation = reserved-annotation-start reserved-body
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -5,10 +5,9 @@ simple-start = simple-start-char / text-escape / expression
 pattern = *(text-char / text-escape / expression)
 
 complex-message = *(declaration [s]) complex-body [s]
-declaration = input-declaration / local-declaration / reserved-syntax
+declaration = input-declaration / local-declaration / reserved-keyword
 input-declaration = input [s] variable-expression
 local-declaration = local s variable [s] "=" [s] expression
-reserved-syntax = "." reserved-body 1*expression
 
 complex-body = quoted-pattern
              / (selectors 1*([s] variant))
@@ -22,7 +21,9 @@ expression = literal-expression / variable-expression / function-expression
 literal-expression = "{" [s] literal [s annotation] [s] "}"
 variable-expression = "{" [s] variable [s annotation] [s] "}"
 function-expression = "{" [s] annotation [s] "}"
-annotation = (function *(s option)) / reserved / private-use
+annotation = (function *(s option))
+           / reserved-annotation
+           / private-use-annotation
 
 literal = quoted / unquoted
 variable = "$" name
@@ -54,16 +55,18 @@ unquoted-start = name-start / DIGIT / "."
                / %xB7 / %x300-36F / %x203F-2040
 
 
-; reserve sigils for private-use by implementations
-private-use    = private-start reserved-body
-private-start  = "^" / "&"
+; Reserve additional .keywords for use by future versions of this specification.
+; Note that the following expression is a simplification,
+; as this rule MUST NOT be considered to match existing keywords:
+;   (input / local / match) (s / "{")
+reserved-keyword = "." reserved-body 1*expression
 
-; reserve additional sigils for use by 
-; future versions of this specification
-reserved       = reserved-start reserved-body
-reserved-start = "!" / "@" / "#" / "%" / "*" / "<" / ">" / "/" / "?" / "~"
+; Reserve additional sigils for use by future versions of this specification.
+reserved-annotation = reserved-annotation-start reserved-body
+reserved-annotation-start = "!" / "@" / "#" / "%" / "*"
+                          / "<" / ">" / "/" / "?" / "~"
+
 reserved-body  = *([s] 1*(reserved-char / reserved-escape / quoted))
-
 reserved-char  = %x00-08        ; omit HTAB and LF
                / %x0B-0C        ; omit CR
                / %x0E-19        ; omit SP
@@ -71,6 +74,10 @@ reserved-char  = %x00-08        ; omit HTAB and LF
                / %x5D-7A        ; omit { | }
                / %x7E-D7FF      ; omit surrogates
                / %xE000-10FFFF
+
+; Reserve sigils for private-use by implementations.
+private-use-annotation = private-start reserved-body
+private-start  = "^" / "&"
 
 ; based on https://www.w3.org/TR/xml/#NT-Name,
 ; but cannot start with U+003A COLON ":"
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -5,7 +5,7 @@ simple-start = simple-start-char / text-escape / expression
 pattern = *(text-char / text-escape / expression)
 
 complex-message = *(declaration [s]) complex-body [s]
-declaration = input-declaration / local-declaration / reserved-keyword
+declaration = input-declaration / local-declaration / reserved-statement
 input-declaration = input [s] variable-expression
 local-declaration = local s variable [s] "=" [s] expression
 
@@ -59,7 +59,7 @@ unquoted-start = name-start / DIGIT / "."
 ; Note that the following expression is a simplification,
 ; as this rule MUST NOT be considered to match existing keywords:
 ;   (input / local / match) (s / "{")
-reserved-keyword = "." reserved-body 1*expression
+reserved-statement = "." reserved-body 1*expression
 
 ; Reserve additional sigils for use by future versions of this specification.
 reserved-annotation = reserved-annotation-start reserved-body
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -4,13 +4,13 @@ simple-message = [simple-start pattern]
 simple-start = simple-start-char / text-escape / expression
 pattern = *(text-char / text-escape / expression)
 
-complex-message = *(declaration [s]) complex-body [s]
+complex-message = *(declaration [s]) complex-body
 declaration = input-declaration / local-declaration / reserved-statement
 input-declaration = input [s] variable-expression
 local-declaration = local s variable [s] "=" [s] expression
 
 complex-body = quoted-pattern
-             / (selectors 1*([s] variant))
+             / ((selectors / reserved-statement) 1*([s] variant))
 quoted-pattern = "{{" pattern "}}"
 
 selectors = match 1*([s] expression)
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -1,27 +1,19 @@
 message = simple-message / complex-message
 
-simple-message = [almost-keyword]
-               / ((almost-keyword not-keyword / text-start / expression) pattern)
-complex-message = *(declaration [s]) body
+simple-message = [simple-start pattern]
+simple-start = text-start / text-escape / expression
 
+pattern = *(text / expression)
+
+complex-message = *(declaration [s]) complex-body
 declaration = input-declaration / local-declaration / reserved-declaration
 input-declaration = input [s] variable-expression
 local-declaration = local s variable [s] "=" [s] expression
-reserved-declaration = "." 4*(%x61-7A) reserved-body 1*expression
+reserved-declaration = "." reserved-body 1*expression
 
-almost-keyword = "." *3(%x61-7A)
-not-keyword    = %x0-5B        ; omit \
-               / %x5D-60       ; omit a-z and {
-               / %x7C          ; omit }
-               / %x7E-D7FF     ; omit surrogates
-               / %xE000-10FFFF
-               / expression
-
-body = quoted-pattern
-     / (selectors 1*([s] variant))
-
+complex-body = quoted-pattern
+             / (selectors 1*([s] variant))
 quoted-pattern = "{{" pattern "}}"
-pattern = *(text / expression)
 
 selectors = match 1*([s] expression)
 variant = 1*(s key) [s] quoted-pattern
@@ -38,7 +30,7 @@ variable = "$" name
 function = (":" / "+" / "-") name
 option = name [s] "=" [s] (literal / variable)
 
-; keywords always consist of a . followed by at least four lower-case a-z letters
+; keywords always consist of a . followed by lower-case a-z letters
 input = %s".input"
 local = %s".local"
 match = %s".match"
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -90,7 +90,7 @@ name-start = ALPHA / "_"
 name-char  = name-start / DIGIT / "-" / "." / ":"
            / %xB7 / %x300-36F / %x203F-2040
 
-text-escape     = backslash ( backslash / "." / "{" / "}" )
+text-escape     = backslash ( backslash / "{" / "}" )
 quoted-escape   = backslash ( backslash / "|" )
 reserved-escape = backslash ( backslash / "{" / "|" / "}" )
 backslash       = %x5C ; U+005C REVERSE SOLIDUS "\"
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -1,11 +1,10 @@
 message = simple-message / complex-message
 
 simple-message = [simple-start pattern]
-simple-start = text-start / text-escape / expression
+simple-start = simple-start-char / text-escape / expression
+pattern = *(text-char / text-escape / expression)
 
-pattern = *(text / expression)
-
-complex-message = *(declaration [s]) complex-body
+complex-message = *(declaration [s]) complex-body [s]
 declaration = input-declaration / local-declaration / reserved-declaration
 input-declaration = input [s] variable-expression
 local-declaration = local s variable [s] "=" [s] expression
@@ -16,7 +15,7 @@ complex-body = quoted-pattern
 quoted-pattern = "{{" pattern "}}"
 
 selectors = match 1*([s] expression)
-variant = 1*(s key) [s] quoted-pattern
+variant = key *(s key) [s] quoted-pattern
 key = literal / "*"
 
 expression = literal-expression / variable-expression / function-expression
@@ -30,19 +29,17 @@ variable = "$" name
 function = (":" / "+" / "-") name
 option = name [s] "=" [s] (literal / variable)
 
-; keywords always consist of a . followed by lower-case a-z letters
 input = %s".input"
 local = %s".local"
 match = %s".match"
 
-text = 1*(text-char / text-escape)
-text-start = %x0-2D         ; omit .
-           / %x2F-5B        ; omit \
-           / %x5D-7A        ; omit {
-           / %x7C           ; omit }
-           / %x7E-D7FF      ; omit surrogates
-           / %xE000-10FFFF
-text-char  = text-start / "."
+simple-start-char = %x0-2D         ; omit .
+                  / %x2F-5B        ; omit \
+                  / %x5D-7A        ; omit {
+                  / %x7C           ; omit }
+                  / %x7E-D7FF      ; omit surrogates
+                  / %xE000-10FFFF
+text-char = simple-start-char-start / "."
 
 quoted      = "|" *(quoted-char / quoted-escape) "|"
 quoted-char = %x0-5B         ; omit \
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -5,10 +5,10 @@ simple-start = simple-start-char / text-escape / expression
 pattern = *(text-char / text-escape / expression)
 
 complex-message = *(declaration [s]) complex-body [s]
-declaration = input-declaration / local-declaration / reserved-declaration
+declaration = input-declaration / local-declaration / reserved-syntax
 input-declaration = input [s] variable-expression
 local-declaration = local s variable [s] "=" [s] expression
-reserved-declaration = "." reserved-body 1*expression
+reserved-syntax = "." reserved-body 1*expression
 
 complex-body = quoted-pattern
              / (selectors 1*([s] variant))


spec/message.abnf
```diff
@@ -1,18 +1,30 @@
...
-complex-message = "{{" [s] *(declaration [s]) body [s] "}}"
+simple-message = [almost-keyword]
+               / ((almost-keyword not-keyword / text-start / expression) pattern)
+complex-message = *(declaration [s]) body
 
-declaration = input-declaration / local-declaration
+declaration = input-declaration / local-declaration / reserved-declaration
 input-declaration = input [s] variable-expression
 local-declaration = local s variable [s] "=" [s] expression
+reserved-declaration = "." 4*(%x61-7A) reserved-body 1*expression
```
eemeli, line 10: This has two potential problems:
1. It conflicts with the `input-declaration` and `local-declaration` rules. Is that ok, or do we need the ABNF to be that meticulous?
2. This is requiring every declaration to end with at least one expression. Is that too restricting? If so, can ABNF somehow do a non-capturing lookahead for .keywords as a rule terminator?

spec/message.abnf
```diff
@@ -1,18 +1,30 @@
-message = pattern / complex-message
+message = simple-message / complex-message
 
-complex-message = "{{" [s] *(declaration [s]) body [s] "}}"
+simple-message = [almost-keyword]
+               / ((almost-keyword not-keyword / text-start / expression) pattern)
+complex-message = *(declaration [s]) body
```
eemeli, line 5: Should we still allow for non-space-separated declarations? As in, should this still be valid?
```
.input{$foo}.input{$bar}.match{$foo}one{{o-n-e}}*{{e-l-s-e}}
```

For the sake of legibility, I'd really like to require at least a space separating each declaration, `.match` and variant:
```
.input{$foo} .input{$bar} .match{$foo} one{{o-n-e}} *{{e-l-s-e}}
```

spec/message.abnf
```diff
@@ -1,18 +1,30 @@
...
 local-declaration = local s variable [s] "=" [s] expression
+reserved-declaration = "." 4*(%x61-7A) reserved-body 1*expression
+
+almost-keyword = "." *3(%x61-7A)
+not-keyword    = %x0-5B        ; omit \
+               / %x5D-60       ; omit a-z and {
+               / %x7C          ; omit }
+               / %x7E-D7FF     ; omit surrogates
+               / %xE000-10FFFF
+               / expression
```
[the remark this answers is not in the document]
eemeli: Ach, `not-keyword` needs a `/ text-escape` as well.

aphillips:
I was hoping to land the namespace changes before embarking on this, which is why I didn't do it yesterday. But worth resolving stuff before we change the ABNF for reals.

spec/message.abnf
aphillips, line 5: This is too awful to contemplate.

Really it should be:

```abnf
message = unquoted-pattern / complex-message
unquoted-pattern = (expression / unquoted-start) pattern
```

spec/message.abnf
aphillips: This is too opinionated about future keyword use. You don't need to define a reserved-declaration. Just let a future keyword define what it needs.

spec/message.abnf
```diff
@@ -26,18 +38,19 @@ variable = "$" name
...
-          / %x7C           ; omit }
-          / %x7E-D7FF      ; omit surrogates
-          / %xE000-10FFFF
+text-start = %x0-2D         ; omit .
+           / %x2F-5B        ; omit \
+           / %x5D-7A        ; omit {
+           / %x7C           ; omit }
+           / %x7E-D7FF      ; omit surrogates
+           / %xE000-10FFFF
+text-char  = text-start / "."
```
aphillips, line 53: pretty sure you don't want to do this, since quoted patterns exist.

spec/message.abnf
gibson042: I don't think those spaces should be required if the grammar is unambiguous without them. Legibility is for people, not machines.

spec/message.abnf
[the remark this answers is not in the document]
gibson042: Extending https://github.com/unicode-org/message-format-wg/pull/529/files#r1400798653 , simple messages should just not be allowed to start with the keyword-prefix sigil. If `.input{$s} {{"{$s}"}}` is not interpreted as a simple message, then `.foo{$s} {{"{$s}"}}` and `.x` and `.` should not be either.

spec/message.abnf
gibson042: Omitting `\` (or more specifically, `text-escape`) is also a problem in itself, because it means that ` \{\} starts with <U+0020, U+007B, U+007D>` is a valid simple message but `\{\} starts with <U+007B, U+007D>` is not.

spec/message.abnf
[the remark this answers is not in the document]
stasm: +1 to @gibson042. Let's at least start with making all dots in column 0 special, so that the PR is easier to review. Then open a follow-up.

spec/message.abnf
eemeli: Doing so would break our self-imposed stability policy. Specifically, this part: https://github.com/unicode-org/message-format-wg/blob/85e16398b07a91d8b1a16eccdec474af4e6198cc/spec/README.md?plain=1#L100-L104

If we don't introduce a `reserved-declaration`, then we won't be able to define any further declarations or keyword operations without calling it MF3 or something.

spec/message.abnf
```diff
@@ -1,18 +1,30 @@
-message = pattern / complex-message
+message = simple-message / complex-message
```
stasm, line 1: This makes `{{Pattern}}` a valid complex message, in addition to `Pattern` being a valid simple message.

I think this is good. It's a degenerate case of a complex message with no declarations and one implicit variant. I also think it's consistent with how many storage format that allow unquoted content also allow quoting it.

I'm calling it out to make sure we all agree about this. I recall  `.{{Pattern}}` being brought up earlier, which I think would be worse. By choosing dotted keywords as code introducers we effectively freed up `{{`/`}}`, which now are only used unambiguously for quoting patterns.

eemeli:
Based on feedback from @aphillips and @gibson042, I changed it so a leading `.` in a message always needs to be escaped, lest it be introducing a complex-message keyword.

spec/message.abnf
eemeli: That was the case before as well, yes? Only with the previous style it would've needed to be quoted as `{{{{Pattern}}}}`.

eemeli: Also noting that this (as before) makes the syntax at least LL2, as we need to look at the second character after the first `{` to determine if it's a placeholder, or a complex-mode pattern quote.

spec/message.abnf
eemeli: I'm pretty sure the updated syntax resolves any issues here. Please check?

spec/message.abnf
```diff
@@ -1,18 +1,22 @@
...
 
-complex-message = "{{" [s] *(declaration [s]) body [s] "}}"
+simple-message = [simple-start pattern]
+simple-start = text-start / text-escape / expression
 
-declaration = input-declaration / local-declaration
+pattern = *(text / expression)
```
gibson042, line 6: Suggestion: make more clear that `simple-message` differs from `pattern` only in leading character constraints stop by removing `text`.
```suggestion
pattern = *(text-char / text-escape / expression)
```
or maybe keep `text` and better align nonterminals with the [data model](https://github.com/unicode-org/message-format-wg/blob/85e16398b07a91d8b1a16eccdec474af4e6198cc/spec/syntax.md#pattern) by defining both `simple-message` and `pattern` as a sequence of "parts":
```abnf
simple-message    = [simple-start-part *pattern-part] ; a pattern that doesn't start with "."
simple-start-part = simple-start-text / expression
simple-start-text = (simple-start-char / text-escape) [text] ; text that doesn't start with "."
simple-start-char = %x0-2D         ; omit .
                  / %x2F-5B        ; omit \
                  / %x5D-7A        ; omit {
                  / %x7C           ; omit }
                  / %xE000-10FFFF

text      = 1*(text-char / text-escape)
text-char = simple-start-char / "."

pattern      = *pattern-part
pattern-part = text / expression
```

spec/message.abnf
```diff
@@ -1,18 +1,22 @@
...
-     / (selectors 1*([s] variant))
-
+complex-body = quoted-pattern
+             / (selectors 1*([s] variant))
 quoted-pattern = "{{" pattern "}}"
-pattern = *(text / expression)
+
 selectors = match 1*([s] expression)
-variant = when 1*(s key) [s] quoted-pattern
+variant = 1*(s key) [s] quoted-pattern
```
gibson042, line 19: Whitespace handling seems a bit off here; I think it's only mandatory _between_ keys.
```suggestion
variant = key *(s key) [s] quoted-pattern
```

spec/message.abnf
```diff
@@ -1,18 +1,22 @@
...
+message = simple-message / complex-message
 
-complex-message = "{{" [s] *(declaration [s]) body [s] "}}"
+simple-message = [simple-start pattern]
+simple-start = text-start / text-escape / expression
 
-declaration = input-declaration / local-declaration
+pattern = *(text / expression)
+
+complex-message = *(declaration [s]) complex-body
```
gibson042, line 8: This is a preexisting constraint, but it's worth noting that even a complex message is prohibited to have any trailing whitespace by the grammar—including a terminal newline that would generally be expected if represented as the contents of a text file.

gibson042 (commented):
Once the grammar is stable, don't forget to update [syntax.md](https://github.com/unicode-org/message-format-wg/blob/main/spec/syntax.md) as well.

spec/message.abnf
gibson042: Confirmed: https://github.com/unicode-org/message-format-wg/pull/529#pullrequestreview-1743003445

spec/message.abnf
aphillips: @stasm noted:
> I'm calling it out to make sure we all agree about this. I recall .{{Pattern}} being brought up earlier, which I think would be worse. By choosing dotted keywords as code introducers we effectively freed up {{/}}, which now are only used unambiguously for quoting patterns.

The design/choice doc explicitly calls out that the `.keyword` syntax makes `{{Pattern}}` a quoted pattern "complex" message because the dot isn't special--it is part of the keyword. This is a _feature_ of the design 😸

spec/message.abnf
aphillips: I think I prefer what I suggested before: making dot special for unquoted patterns. I like the conceptual simplicity and having fewer overall productions:

```suggestion
message = simple-message / complex-message
; simple-messages are any unquoted pattern 
; unquoted patterns cannot start with dot
; the empty string is a valid message
simple-message = [(unquoted-start / expression) pattern]
unquoted-start = %x0-2D ; omit . 
               / %x2F-5B ; omit \ 
               / %x5D-7A ; omit { 
               / %x7C ; omit } 
               / %x7E-D7FF ; omit surrogates 
               / %xE000-10FFFF
```

aphillips (changes requested):
Making progress. Some of my suggestions take a different route to the same destination for reasons I call out. Pay attention to the whitespace issue.

spec/message.abnf
aphillips: This is a good callout, but I would remove the `-declaration` from the name because it might not be a declaration but instead some other kind of syntax. I would also not do `4*`. I would suggest:

```suggestion
reserved-syntax = "." %s*[a-z] [(s reserved-body)] 1*expression
```

spec/message.abnf
```diff
@@ -81,7 +86,7 @@ name-start = ALPHA / "_"
 name-char  = name-start / DIGIT / "-" / "." / ":"
            / %xB7 / %x300-36F / %x203F-2040
 
-text-escape     = backslash ( backslash / "{" / "}" )
+text-escape     = backslash ( backslash / "." / "{" / "}" )
```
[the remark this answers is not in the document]
gibson042: > Do we need `\.` all the time? Or can we make it a special case of starting a simple message?

Let's _please_ avoid such special cases; they're unnecessary complexity for people and machines alike. If `\.` is a valid simple message representing ".", then `period\.` and `\.\.\.` should be valid as well (respectively representing "period." and "...").

> I'm almost willing to suggest that we require patterns starting with `.` to be quoted and not provide any escape

I had the same thought, and that approach I could live with.

spec/message.abnf
gibson042: That fails to support a leading escape.
```diff
-simple-message = [(unquoted-start / expression) pattern]
+simple-message = [(unquoted-start / text-escape / expression) pattern]
```

spec/message.abnf
gibson042: No, that's already covered by the containing `selectors 1*([s] variant)` production.

spec/message.abnf
gibson042: I agree that it _should_ be supported, but neither the current `complex-message = "{{" [s] *(declaration [s]) body [s] "}}"` nor this `complex-message = *(declaration [s]) complex-body` production accommodate that.

spec/message.abnf
duerst: @gibson042 wrote:
> I don't think those spaces should be required if the grammar is unambiguous without them. Legibility is for people, not machines.

@eemeli wrote:
> For the sake of legibility, I'd really like to require at least a space separating each declaration, .match and variant.

I'm strongly on @eemeli's side. Legibility is for people. But requiring the spaces will make sure that even machine-generated messages are legible to people. And it will make sure that there's a consistent "custom" of writing messages with spaces. (You'd be surprised how Japanese, in particular unexperienced students, tend to omit the spaces because they are not used to spaces in Japanese text.)

gibson042: I disagree that messages like `.input{$foo} .input{$bar} .match{$foo} one{{o-n-e}} … *{{e-l-s-e}}` are sufficiently legible to people, and in my own use would use additional separating line feeds and spaces:
```
.input {$foo}
.input {$bar}
.match {$foo}
one {{o-n-e}}
…
* {{e-l-s-e}}
```
Insisting on unnecessary spaces at a syntactic level is a half-measure that doesn't actually address any legibility concern IMO.

spec/message.abnf
gibson042: So altogether, we're looking at ABNF like one of the following:
**Data Model Analog**
```abnf
message = simple-message / complex-message

; simple-message is a pattern that doesn't start with dot
; the empty string is a valid message
simple-message    = [simple-start-part *pattern-part]
simple-start-part = simple-start-text / expression
simple-start-text = (simple-start-char / text-escape) [text]
simple-start-char = %x0-2D         ; omit .
                  / %x2F-5B        ; omit \
                  / %x5D-7A        ; omit {
                  / %x7C           ; omit }
                  / %xE000-10FFFF

; text is a sequence of characters, some represented using escapes
text      = 1*(text-char / text-escape)
text-char = simple-start-char / "."

; pattern is a possibly-empty sequence of text and expression parts
pattern      = *pattern-part
pattern-part = text / expression

complex-message = *(declaration [s]) complex-body
…
```

**Fewer Productions**
```abnf
message = simple-message / complex-message

; simple-message is a pattern that doesn't start with dot
; the empty string is a valid message
simple-message = [(simple-start-char / text-escape / expression) pattern]
simple-start-char = %x0-2D         ; omit .
                  / %x2F-5B        ; omit \
                  / %x5D-7A        ; omit {
                  / %x7C           ; omit }
                  / %xE000-10FFFF

; pattern is a possibly-empty sequence of characters, escapes, and expressions
pattern   = *(text-char / text-escape / expression)
text-char = simple-start-char / "."

complex-message = *(declaration [s]) complex-body
…
```

spec/message.abnf
duerst: @gibson042 I'd definitely also use line feeds in some places. I'd be fine if we require line feeds in these positions. Let's do this as a full measure rather than a half-measure.

spec/message.abnf
eemeli: The latest draft has https://github.com/unicode-org/message-format-wg/blob/915cbb7f096815d4c6d908fb8bfac162dfe86b1b/spec/message.abnf#L12

as the keyword also matches `reserved-body`. I'll update the name and then resolve this comment.

spec/message.abnf
eemeli: Renamed `text-start` as `simple-start-char` and mostly followed Richard's last suggestion.

spec/message.abnf
eemeli: Added a `[s]` to the end of `complex-message` to fix this.

spec/message.abnf
```diff
@@ -26,18 +30,19 @@ variable = "$" name
 function = (":" / "+" / "-") name
 option = name [s] "=" [s] (literal / variable)
 
-; reserved keywords are always lowercase
-input = %s"input"
-local = %s"local"
-match = %s"match"
-when  = %s"when"
+; keywords always consist of a . followed by lower-case a-z letters
```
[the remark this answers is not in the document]
eemeli: I removed the comment as it was distracting.

spec/message.abnf
```diff
@@ -1,18 +1,22 @@
-message = pattern / complex-message
+message = simple-message / complex-message
 
-complex-message = "{{" [s] *(declaration [s]) body [s] "}}"
+simple-message = [simple-start pattern]
```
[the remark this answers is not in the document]
eemeli: I agree, let's make sure this is called out in `syntax.md`.

eemeli:
Updated again. I dropped `text`, as with different handling for the first character there's now no way for a single ABNF rule to correspond with what we'd like to consider _text_ in the syntax. So including such a rule would be misleading.

I'd also strongly prefer not to add comments to the `message.abnf` file, as we already have its annotated version in `syntax.md`. OTOH, I added one for `reserved-keyword`, as its ABNF rule is only an approximation of what it should be (see https://github.com/unicode-org/message-format-wg/pull/529#pullrequestreview-1741641928 above), and it will cause automated parser generators to fail with this syntax.

spec/message.abnf
stasm: Ah, you're right, I must have missed it. Or rather, https://github.com/unicode-org/message-format-wg/pull/521#discussion_r1390250275 made me believe that it wasn't the case. Thanks for confirming!

spec/message.abnf
stasm: We expect many messages to only exist in a single line form, e.g. in spreadsheets. We should not require line breaks. As for spaces -- we can recommend a style, but I strongly believe we should be lenient in what we parse, and be as whitespace-agnostic as possible.

spec/message.abnf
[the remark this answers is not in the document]
stasm: > I'm almost willing to suggest that we require patterns starting with . to be quoted and not provide any escape

+1. We need to support `{{.com}}` anyways as a valid pattern, so we could reduce the scope of this PR, and not introduce any new escape sequences. Then disccuss separately.

spec/message.abnf
```diff
@@ -1,43 +1,46 @@
...
+simple-start = simple-start-char / text-escape / expression
+pattern = *(text-char / text-escape / expression)
 
-declaration = input-declaration / local-declaration
+complex-message = *(declaration [s]) complex-body [s]
```
stasm, line 7: Why do we need `[s]` at the end here?

spec/message.abnf
```diff
@@ -1,43 +1,46 @@
-message = pattern / complex-message
+message = simple-message / complex-message
 
-complex-message = "{{" [s] *(declaration [s]) body [s] "}}"
+simple-message = [simple-start pattern]
+simple-start = simple-start-char / text-escape / expression
+pattern = *(text-char / text-escape / expression)
```
stasm, line 5: I'm unhappy about losing the `text` production. Conceptually, we think of patterns as sequences of contiguous text runs or placeholders, not a sequence of characters or placeholders.

Perhaps we shoud have `quoted-text` and `unquoted-text`? The latter would replace `simple-message`.

spec/message.abnf
aphillips: This is the discussion about message trailing whitespace not being invalid, I think?

spec/message.abnf
aphillips: @duerst 

We long ago decided about non-essential whitespace. Resource formats and message authors can impose additional formatting and structure onto messages, but the syntax itself does not impose it. Messages can often appear in a single, simple string (think Java properties file), where line breaks would be a problem or be reduced to illegible escape sequences (reducing rather than adding value).

spec/message.abnf
aphillips: I agree. This was better as `*(text / expression)`

spec/message.abnf
```diff
@@ -1,43 +1,46 @@
...
 
-declaration = input-declaration / local-declaration
+complex-message = *(declaration [s]) complex-body [s]
+declaration = input-declaration / local-declaration / reserved-keyword
```
aphillips, line 8: `reserved-keyword` doesn't sound right (it suggests that it is only the keyword). Perhaps `reserved-statement` or even just `reserved`. The corresponding (and different) `reserved-annotation` isn't in conflict with it.

spec/message.abnf
eemeli: Yeah; this makes is so that `{{Pattern}}\n` is valid. It makes e.g. embedding complex messages in YAML a bit easier, as their trailing newlines don't need to be trimmed.

Note that ` {{Pattern}}` is still invalid, as entering complex mode depends on the first character of the message, and we're not trimming spaces. The difference between the two is that trailing whitespace is harder to see, and doesn't affect the parser.

spec/message.abnf
```diff
@@ -1,43 +1,46 @@
...
 input-declaration = input [s] variable-expression
 local-declaration = local s variable [s] "=" [s] expression
 
-body = quoted-pattern
-     / (selectors 1*([s] variant))
-
+complex-body = quoted-pattern
+             / (selectors 1*([s] variant))
```
aphillips, line 13: Just name this `body` like it was before.

I think this also needs `reserved-keyword`? So perhaps:

```suggestion
body = quoted-pattern
     / (selectors 1*([s] variant))
     / reserved
```

spec/message.abnf
eemeli: Note that `simple-start` is only applied to the first character, and it needs to be a part of the eventual _text_. Furthermore, a simple message pattern may include multiple _text_ spans separated by expressions.

To see this, consider the valid message `one{$x}.two`. How could you express a `text` that captures both `one` and `.two`, but would not if their order was reversed as `.two{$x}one`, which is invalid?

spec/message.abnf
stasm: Wait, in `{{Pattern}}\n` the trailing newline is a storage format's concern. Why do you think solving this at the syntax level is appropriate?

Can we discuss this separately from this PR? Why is it in scope?

spec/message.abnf
gibson042: This seems like a request for what I labeled "Data Model Analog" in https://github.com/unicode-org/message-format-wg/pull/529#discussion_r1401328223 .
> ```abnf
> simple-message    = [simple-start-part *pattern-part]
> simple-start-part = simple-start-text / expression
> …
> ; pattern is a possibly-empty sequence of text and expression parts
> pattern      = *pattern-part
> pattern-part = text / expression
> ```

spec/message.abnf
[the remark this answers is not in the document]
eemeli: Not supporting `\.` feels to me like it would introduce an odd disparity in how the two code-y characters `{` and `.` are handled when they show up in a pattern. With `{`, the answer is to escape it as `\{`. But with `.` I need to escape it as `{{.}}`? Why not support `\.` as well?

spec/message.abnf
eemeli: I'd consider this in scope because before this PR this rule was:
```abnf
complex-message = "{{" [s] *(declaration [s]) body [s] "}}"
```
and now it's
```abnf
complex-message = *(declaration [s]) complex-body [s]
```

So removing that terminal `[s]` is a greater change than what's proposed here.

spec/message.abnf
eemeli: I'll rename as `reserved-statement`.

spec/message.abnf
eemeli: If we're taking the "body" term exclusively for complex messages, then we should rename `reserved-body` to something else. My preference would be to instead specify it as `complex-body`, because that's where it's used.

As we're not allowing any `{}` nesting in reserved syntax, I don't see how we'd later allow for it replacing the whole of a complex message's body. Could you describe what sort of future extension you had in mind here?

spec/message.abnf
gibson042: > So removing that terminal `[s]` is a greater change than what's proposed here.

I don't think I agree... the current `complex-message = "{{" [s] *(declaration [s]) body [s] "}}"` already does not permit anything after the closing "}}" so e.g. `{{ input {$x} {{foo}} }}` is valid but `{{ input {$x} {{foo}} }}    ` is not. If the scope of this PR is limited to [code-mode-introducer.md](https://github.com/unicode-org/message-format-wg/blob/main/exploration/code-mode-introducer.md), then making `.input {$x} {{foo}}` valid applies but IMO making `.input {$x} {{foo}}    ` valid does not, **especially** since it's making `    .input {$x} {{foo}}    ` (with both leading and trailing whitespace) a valid declaration-free _simple_ message.

I'm not comfortable with the trailing whitespace being invalid, but I'm also not comfortable with whitespace being significant only when it is leading or in an already-established pattern context—and at any rate, changing either more than is necessary to differentiate simple vs. complex messages by how they start would be scope creep.

spec/message.abnf
```diff
@@ -52,16 +55,18 @@ unquoted-start = name-start / DIGIT / "."
...
 
 
-; reserve sigils for private-use by implementations
-private-use    = private-start reserved-body
-private-start  = "^" / "&"
+; Reserve additional .keywords for use by future versions of this specification.
+; Note that the following expression is a simplification,
+; as this rule MUST NOT be considered to match existing keywords:
+;   (input / local / match) (s / "{")
+reserved-statement = "." reserved-body 1*expression
```
gibson042, line 62: I think requiring `reserved-statement` to end with at least one `expression` matches too little.
```suggestion
; A reserved statement continues until the next statement or quoted-pattern,
; and may include any number of expressions and/or quoted literals
reserved-statement      = "." *([s] reserved-statement-part)
reserved-statement-part   = 1*(reserved-statement-char / reserved-statement-escape)
                          / expression
                          / quoted
reserved-statement-char   = %x00-08        ; omit `s`: HTAB and LF
                          / %x0B-0C        ; omit `s`: CR
                          / %x0E-19        ; omit `s`: SP
                          / %x21-2D        ; omit .
                          / %x2F-5B        ; omit \
                          / %x5D-7A        ; omit { | }
                          / %x7E-D7FF      ; omit surrogates
                          / %xE000-10FFFF
reserved-statement-escape = backslash ( %x00-D7FF / %xE000-10FFFF )
```
Note that this locks in the `expression` syntax for effectively forever because a left brace _could_ start a `{{…}}` quoted expression which would terminate the reserved statement—though possible to work around that too; it's rather impractical. And likewise  for `quoted`.

spec/message.abnf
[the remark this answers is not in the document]
stasm: OTOH, one could also argue that *supporting* `\.` introduces an odd disparity in how the period in column 0 needs to be quoted but any other period doesn't. Or that it's an odd diparity in how `\.` produces a `.` in column 0 but it isn't allowed anywhere else (depending on the choice we make in this thread).

This is a good illustration of the accidental comprexity brought about by choosing D in #525. Note that neither A, E, nor F suffer from this problem.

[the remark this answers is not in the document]
aphillips: That's not correct, though. Every option has column 0 quoting issues. They just differ in frequency and such. Options C and E were about trying to minimize the frequency.

spec/message.abnf
aphillips: > Could you describe what sort of future extension you had in mind here?

Basically, a different type of selector from `.match`

spec/message.abnf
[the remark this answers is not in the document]
stasm: In the options using `{` as the first character of the code mode introducer sequence, the quoting issues are not specific to column 0, and do not create inconsistencies in columns 1 and onwards. They leverage the exact same escaping mechanism: `\{`.

I know this may seem like a banal observation, but I don't think the consequences of having separate escaping requirements were thoroughly discussed. What's your take on what should happen to `foo\.`?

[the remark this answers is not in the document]
gibson042: No, @stasm is right. A, E, and F all used prefixes that start with `{`, which already needs escaping _everywhere_ in pattern text. So,
* In option A, a simple message producing `{{Use pattern quotes for messages}}` would look like e.g. `\{\{Use pattern quotes for messages\}\}`.
* In option E, a simple message producing `{^} Use a magic sequence at the start` would look like e.g. `\{^\} Use a magic sequence at the start`.
* In option F, a simple message producing `{% Use a preamble with %} {{Patterns to follow}}` would look like e.g. `\{% Use a preamble with %\} \{\{Patterns to follow\}\}`.

And in none of those is any special treatment necessary to escape the first character of a simple message that is not also necessary everywhere else the same character appears in it.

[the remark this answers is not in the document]
eemeli: Given that we _have_ selected option D, could we focus in this thread on whether to allow for `\.` to represent a literal `.` in a pattern, rather than rehashing what the situation would have been had we chosen something else?

My position is that given that we already allow for escaping `{` as `\{` and `}` as `\}`, some users will have an expectation that they may also escape `.` as `\.`, and that with `.` as a keyword introducer, there is no sensible other meaning for `\.`. Therefore it behooves us to allow for it.

And if we don't allow for it, then by consistency we probably should also not allow for escaping `}` in a simple pattern, given that doing so is not required.

spec/message.abnf
eemeli: For that, we need something like this:

```suggestion
complex-body = quoted-pattern
             / ((selectors / reserved-statement) 1*([s] variant))
```

Feel free to commit that suggestion to this PR if you agree.

spec/message.abnf
eemeli: The `.` exclusion also disallows using e.g. `name` within a future statement, so e.g. introducing `local-declaration` would not be possible if we didn't include it already.

On the other hand, if each `declaration` had to end in a required whitespace `s`, we would only need to bar the sequence `(s ".")` from what's reserved.

spec/message.abnf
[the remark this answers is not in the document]
aphillips: Yes, but that's still an escape of `{` (the fact that we've accounted for it already doesn't make it "not an escape").

Anyway, one of the problems here is the sigil choice. Using a rarer sigil than `.` would help.

Some solutions might include:

- Reserve a bunch of keywords now. The problem with the current situation isn't the keywords we have, but rather future compatibility needs with keywords that we don't know. We could reserve a list of words for future standardization. (This solution is not a serious contender)

- Fix the sigils or "spelling". The problem here is that `.` is not rare and `..` is also not rare. If the starter sigil/sigil sequence is rare enough, while it still has to be escaped when in plain text, it isn't a serious problem for anyone. This could be done for existing keywords or only for future keywords.

Examples:

```
.=input {$var :foo}
.=local $bar = {$foo :function}
.?match {$var} {$bar}
[*] {{.= and .? are rare so users rarely have to escape them}}
[*] {{== and ?? or some other sigil choice instead of . suggests itself}}
[*] {{note that we started with #}}

.input {$var}
.local $bar = {$foo :function}
.=future 
{{future keywords start with a special sequence--sorry future me--
  and the well-known sequences .input/.local/.match are reserved now}}
```

- Enclosing instead of doubling. Make the keyword distinct in another way:

```
.input. {$var}
.local. $bar = {$foo :function}
~input~
!local!
```

- Adjust the syntax so there is no optional whitespace and the bracket is part of the word

```
.input{$var :function}
.local{$bar} = {$foo :function}
.match{$var}
[*] {{This is not a serious contender either}}
```

- Use the character we've already abused

```
{input} {$var :foo}
{local} $bar = {$foo :function}
{match} {$var} {$bar}
[*] {{So many \{ }}

File: spec/message.abnf (hunk 15 above)
Comment (eemeli) [reply spec/message.abnf:13]: Actually, let's not change this line. The `reserved-statement` alternative should really go within `selectors` instead.

File: spec/message.abnf (hunk 9 above)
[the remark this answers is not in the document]
Comment (eemeli) [reply spec/message.abnf:93]: @aphillips I've hidden your preceding comment to keep this thread somewhat on-topic on whether and where to allow for `\` escaping the keyword introducer `.`.

File: spec/message.abnf (hunk 12 above)
Comment (eemeli) [reply spec/message.abnf:7]: I don't feel like blocking on this will help us, so let's just get rid of the optional trailing whitespace for now and argue about it separately. 

```suggestion
complex-message = *(declaration [s]) complex-body
```

eemeli:
@aphillips I'd like to discuss the open threads here on today's call, if possible:
1. Confirm that we don't want to require any extra whitespace. [(thread)](#discussion_r1400369771)
2. Do we allow `\.` generally in text, or only in very specific places, or not at all? [(thread)](#discussion_r1401245593)
3. Should we include a `text` rule, even if _text_ will mean something different? [(thread)](https://github.com/unicode-org/message-format-wg/pull/529#discussion_r1401700188)
4. Is it enough to reserve statements that end with an `expression`? [(thread)](https://github.com/unicode-org/message-format-wg/pull/529#discussion_r1402811850)

spec/message.abnf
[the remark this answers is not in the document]
aphillips: I think that we should not allow any `.` to be escaped except in the initial position. My previous comment was about making the escaping less necessary by making a clearer keyword syntax, since I think we want to allow future keyword additions. This design depends on code-mode-introducing keywords being something the parser can pick out. Modifying keywords to include a second sigil might help. 

I'm super nervous about escapes in general because of the interaction with host syntax escaping. In many languages `\.` becomes `\\.` in storage (this is the same conversation about quotes). The double `\\` can be consumed by tools and turned into plain text that is further escaped (`\\\\`). I'm not as worried about `\{` and `\|` because these will be rather rare (and `\|` only inside quoted literals which are mainly inside placeholders). But `\.` is our apostrophe nightmare waiting to happen?

spec/message.abnf
eemeli: During today's call we resolved to not add any new whitespace requirements.

spec/message.abnf
eemeli: During today's call we resolved to be satisfied with _text_ being well defined in `syntax.md`, and not having a single `text` ABNF rule.

spec/message.abnf
eemeli: @gibson042 Please correct me if I misremember, but my understanding from today is that we're fine with reserving only statements that end in an expression?

spec/message.abnf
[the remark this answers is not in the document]
eemeli: As discussed on today's call, the consensus resolution was to not include any `\.` escapes, so a message starting with a literal `.` needs to be quoted.

eemeli:
I've now resolved the discussions as above, and dropped the `\.` escapes. Please review?

During the call, we planned to merge this provided that it receives some approvals, and for a separate PR to apply the further spec changes.

aphillips (approved):

spec/message.abnf
eemeli, line 62: @aphillips While working on the follow-up changes to this PR, I recalled that we'd discussed being more specific and limited about the syntax space that we reserve for statements, so that an actual keyword would always be required. Would this be an improvement in that regard?

```suggestion
reserved-statement = reserved-keyword [s reserved-body] 1*expression
; Note that the following expression is a simplification,
; as this rule MUST NOT be considered to match existing keywords.
reserved-keyword = "." 2*(%x61-7A)  ; (a-z)
```

aphillips: Yes, this looks good (with the `[s]` edit). I probably would not do `2*` even though we're unlikely to make a one letter keyword. I would use our new-found case sensitive ABNF as well and pull the regex comment above (keeping the other comment parts):

```abnf
; as regexp: /\.(?!input|local|match)[a-z]*
reserved-keyword = "." *(%s(a-z))
```

eemeli: Edited the suggestion to use `1*` instead of `2*`, and moved the comment.

Alas, `%s` syntax doesn't support ranges, so need to stick with the `%x` here.

aphillips (approved):
```
૮꒰ ˶• ༝ •˶꒱ა
```

Sheep it.

spec/message.abnf
```diff
@@ -52,16 +55,19 @@ unquoted-start = name-start / DIGIT / "."
...
-reserved-body  = *( [s] 1*(reserved-char / reserved-escape / quoted))
+; Reserve additional sigils for use by future versions of this specification.
+reserved-annotation = reserved-annotation-start reserved-body
+reserved-annotation-start = "!" / "@" / "#" / "%" / "*"
+                          / "<" / ">" / "/" / "?" / "~"
 
+reserved-body  = *([s] 1*(reserved-char / reserved-escape / quoted))
 reserved-char  = %x00-08        ; omit HTAB and LF
                / %x0B-0C        ; omit CR
                / %x0E-19        ; omit SP
```
gibson042, line 73: `reserved-char` should omit `.` so e.g. `.new {$a} .hotness {$b}` is recognized as two distinct reserved statements and `.prefix {$foo} .local ${var} = ${|val|}` is recognized as a reserved statement followed by a local declaration, while still allowing reserved statements to include non-consecutive expressions (like `local-declaration = local s variable [s] "=" [s] expression` already does).
```suggestion
reserved-char  = %x00-08        ; omit HTAB and LF
               / %x0B-0C        ; omit CR
               / %x0E-19        ; omit SP
               / %x21-2D        ; omit .
               / %x2F-5B        ; omit \
```

And accordingly, `reserved-escape` should be generalized to e.g. `reserved-statement-escape = backslash ( %x00-D7FF / %xE000-10FFFF )`.

gibson042 (changes requested):
This PR still has a few issues that IMO should be fixed before it lands.

spec/message.abnf
```diff
@@ -1,43 +1,46 @@
...
+complex-message = *(declaration [s]) complex-body
+declaration = input-declaration / local-declaration / reserved-statement
 input-declaration = input [s] variable-expression
 local-declaration = local s variable [s] "=" [s] expression
 
-body = quoted-pattern
-     / (selectors 1*([s] variant))
-
+complex-body = quoted-pattern
+             / ((selectors / reserved-statement) 1*([s] variant))
```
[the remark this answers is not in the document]
eemeli: Could you say why? The two expressions are wholly synonymous, and I don't really see why we should prefer one over the other.

spec/message.abnf
```diff
@@ -52,16 +55,19 @@ unquoted-start = name-start / DIGIT / "."
...
 
-; reserve additional sigils for use by 
-; future versions of this specification
-reserved       = reserved-start reserved-body
-reserved-start = "!" / "@" / "#" / "%" / "*" / "<" / ">" / "/" / "?" / "~"
-reserved-body  = *( [s] 1*(reserved-char / reserved-escape / quoted))
+; Reserve additional sigils for use by future versions of this specification.
+reserved-annotation = reserved-annotation-start reserved-body
+reserved-annotation-start = "!" / "@" / "#" / "%" / "*"
+                          / "<" / ">" / "/" / "?" / "~"
```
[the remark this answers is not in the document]
eemeli: I think that's pretty well out of scope here.

spec/message.abnf
eemeli: Note that this precludes syntax like that used now in `local-declaration`, where we have a `variable` `name` at the top level, and `name` _can_ include a `.`. Is that intentional?

spec/message.abnf
```diff
@@ -52,16 +55,19 @@ unquoted-start = name-start / DIGIT / "."
                / %xB7 / %x300-36F / %x203F-2040
 
 
-; reserve sigils for private-use by implementations
-private-use    = private-start reserved-body
-private-start  = "^" / "&"
+; Reserve additional .keywords for use by future versions of this specification.
+reserved-statement = reserved-keyword [s [reserved-body [s]]] 1*expression
```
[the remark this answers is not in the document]
eemeli: Is there any particular benefit to parsing out the spaces within reserved content? Originally in https://github.com/unicode-org/message-format-wg/pull/374#discussion_r1169011236 we took them out of `reserved-char` to effectively trim the reserved expression contents within the ABNF.

Should we perhaps revisit that decision, and simplify some of this by not splitting the reserved text into parts?

In the meantime, good point about the missing `[s]` between expressions. If we leave the reserved cleanup to later, this should work for the current line:

```suggestion
reserved-statement = reserved-keyword [s reserved-body] 1*([s] expression)
```

spec/message.abnf
```diff
@@ -52,16 +55,19 @@ unquoted-start = name-start / DIGIT / "."
                / %xB7 / %x300-36F / %x203F-2040
 
 
-; reserve sigils for private-use by implementations
-private-use    = private-start reserved-body
-private-start  = "^" / "&"
+; Reserve additional .keywords for use by future versions of this specification.
+reserved-statement = reserved-keyword [s reserved-body] 1*([s] expression)
```
eemeli, line 59: Replying to @gibson042 from https://github.com/unicode-org/message-format-wg/pull/529#discussion_r1408383628:
> > Please correct me if I misremember, but my understanding from today is that we're fine with reserving only statements that end in an expression?
> 
> That is not my recollection, and would preclude possibilities like `.strict true`. However, I'm fine with addressing it in a followup.

Could we merge this PR with this note in the ABNF, in expectation of getting a PR from @gibson042 that could update all reserved content holistically?

```suggestion
; NOTE: Rules for reserved statements and annotations
; should be considered provisional, pending a near-future PR updating them.
reserved-statement = reserved-keyword [s reserved-body] 1*([s] expression)
```

ps. I had not considered statements like `.strict true`. That does sound like something we ought to reserve space for.

aphillips: The reason statements end with `expression` is that the closing `}` of the expression is how we know the statement ended. Otherwise we're in `reserved-body`... forever. It is a hidden detail of our syntax that statements end with a close bracket.

eemeli: While that's true with the current state of this PR, I believe @gibson042 would like us to drop `.` as a valid character in `reserved-body`, which would allow determining its end by encountering it at the start of the next keyword.

My request here is for us to find a way to not block this PR on that discussion, and to have it instead separately so that the syntax of both reserved statements and annotations could be considered in one package.

aphillips: I agree with finalizing this PR. I don't think the reserved-statement discussion (were we to change something) needs to be in the way of that. It should be a separate conversation. If we want to pursue a change, it should be at least an issue if not a design document.

That said, I would not include the proposed note in this comment thread. Any changes we consider or make will be done in the near future. If we include text about it, put it in syntax.md.

spec/message.abnf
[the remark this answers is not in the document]
gibson042: Because I expect the latter to diverge—I don't think we should treat `.match {…}…` and `.<reserved> …` as interchangeable _prefixes_ before a list of variants, but rather `.match {…}… <variant>…` as one thing complete in itself and `.<reserved> …` as something independent.

spec/message.abnf
gibson042: No, it will need further tweaking to match `.` only after `name-start` or `name-char`, as inside `name`.

spec/message.abnf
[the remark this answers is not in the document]
gibson042: > Is there any particular benefit to parsing out the spaces within reserved content? Originally in [#374 (comment)](https://github.com/unicode-org/message-format-wg/pull/374#discussion_r1169011236) we took them out of `reserved-char` to effectively trim the reserved expression contents within the ABNF.

I think there is; it allows us to more accurately capture adjacency, especially when it comes to things like consecutive declarations (ok but should remain separate) and directly adjacent quoted literals (should be rejected IMO) and—as you mentioned in another comment—dots, which should be accepted as part of a name but not when they appear to start a new statement.

spec/message.abnf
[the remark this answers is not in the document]
gibson042: #546

gibson042 (approved):
I have issues with syntactic reservations in this PR:
* use of `reserved-statement` as [interchangeable with a `.match` _head_](https://github.com/unicode-org/message-format-wg/pull/529#discussion_r1409803560) rather than a full substitute for `.match` with its variants
* requiring `reserved-statement` to end with a list of expressions, precluding e.g. [`.strict true`](https://github.com/unicode-org/message-format-wg/pull/529#discussion_r1408383628)
  * ...and [greedy consumption](https://github.com/unicode-org/message-format-wg/pull/529#discussion_r1408491369) fails to capture the intuition of multiple statements in e.g. `.strict true .local $var = {|val|}`, instead parsing that as a single reserved statement subsuming the `.local`
* [reusing the `reserved-annotation` `reserved-body` for `reserved-statement`](https://github.com/unicode-org/message-format-wg/pull/529#discussion_r1408440410), precluding e.g. `.something opt1={$var} opt2={$var}` because of the intent to prohibit e.g. nested expressions in annotations like `{@reserved {|…|}}`

But that said, it is coherent and can be used as a foundation for further iteration.

spec/message.abnf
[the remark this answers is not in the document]
stasm: What's the motiviation for putting `reserved-statement` inside `complex-body`? Do we expect other multivariant constructs than `match`? I think this may be building too much flexibility into the spec. Could we instead agree that any future keywords would go with other *declarations*?

spec/message.abnf
```diff
@@ -1,43 +1,46 @@
...
+local = %s".local"
+match = %s".match"
 
-text = 1*(text-char / text-escape)
-text-char = %x0-5B         ; omit \
-          / %x5D-7A        ; omit {
-          / %x7C           ; omit }
-          / %x7E-D7FF      ; omit surrogates
-          / %xE000-10FFFF
+simple-start-char = %x0-2D         ; omit .
```
stasm, line 37: Nit: it looks like we don't append `-char` to other `-start` productions in our grammar.

```suggestion
simple-start = %x0-2D         ; omit .
```

stasm (approved):

aphillips:
@stasm asked:
> What's the motiviation for putting reserved-statement inside complex-body? Do we expect other multivariant constructs than match? I think this may be building too much flexibility into the spec. Could we instead agree that any future keywords would go with other declarations?

I think it was me who asked for `reserved-statement` to potentially be _also_ a replacement for `selectors`. This would allow for a different type of selector in future syntax.

Perhaps this should be:

```abnf
complex-body = quoted-pattern
             / (selectors 1*([s] variant))
             / reserved-statement ; which matches some-newly-unreserved-statement 1*([s] variant) 
                                  ; for some future statement
```

spec/message.abnf
eemeli: Note that there already is another rule that's called `simple-start`. If necessary, could we take care of any such editorial rule renames in a separate follow-up PR?

spec/message.abnf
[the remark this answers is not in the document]
eemeli: Tbh, I would find it easier to include it, show what impact it has across the stack, and then evaluate whether to drop it.

We have already agreed on having two follow up PRs to this that will both re-tread this ground. Could we agree that this will be considered more in both of those, and accept having it in the ABNF for now?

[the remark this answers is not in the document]
stasm: > Tbh, I would find it easier to include it, show what impact it has across the stack, and then evaluate whether to drop it.

We should do things exactly the other way.

I also don't appreciate pushing hard for merging. We should merge when things look good to merge.

stasm:
I discussed with @eemeli in Slack and we agreed to merge this as-is, and iterate further starting from @gibson042's feedback in his reivew above: https://github.com/unicode-org/message-format-wg/pull/529#pullrequestreview-1758656138.

spec/message.abnf
stasm: Ah, I see. I'll open a new PR for this.

PR #531, opened 2023-11-27
Reformat `registry.dtd` to not repeat `<!ATTLIST>` & align attribute definitions

eemeli:
This is a purely editorial change, so asking for fast-tracking.

This makes the `registry.dtd` align with commonly used conventions on its formatting.


aphillips:
Fast tracked.

PR #532, opened 2023-11-27
Define function aliases in registry, initially including `:integer`

eemeli:
Adds `<alias>` definitions to the registry, so that shorthands like `:integer` can be made available.


aphillips (approved):
Seems like a useful addition.

PR #533, opened 2023-11-27
Include input & option descriptions as data in registry

eemeli:
One rather important consumer of the registry is syntax highlighting. For that use, we should include the descriptions of the inputs and options in the actual registry data, rather than just in the comments. And the element bodies are a rather excellent place for that.

### DIFF
--- a/spec/registry.dtd
+++ b/spec/registry.dtd
@@ -3,7 +3,7 @@
     xml:lang       NMTOKEN  #IMPLIED
 >
 
-<!ELEMENT function (description+,(formatSignature|matchSignature)+)>
+<!ELEMENT function (description+,(formatSignature|matchSignature|alias)+)>
 <!ATTLIST function
     name           NMTOKEN  #REQUIRED
 >
@@ -19,16 +19,12 @@
     regex          CDATA  #REQUIRED
 >
 
-<!ELEMENT formatSignature (input?,option*)>
+<!ELEMENT formatSignature (input?,option*,override*)>
 <!ATTLIST formatSignature
     position       (open|close|standalone)  "standalone"
-    locales        NMTOKENS                 #IMPLIED
 >
 
-<!ELEMENT matchSignature (input?,option*,match*)>
-<!ATTLIST matchSignature
-    locales        NMTOKENS  #IMPLIED
->
+<!ELEMENT matchSignature (input?,option*,match*,override*)>
 
 <!ELEMENT input (description*)>
 <!ATTLIST input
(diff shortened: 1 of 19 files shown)


spec/registry.xml
```diff
@@ -15,21 +15,25 @@
...
-         "America/New_York".
-      -->
-      <option name="timeZone" validationRule="timeZoneId"/>
+      <option name="dateStyle" values="full long medium short">
+        The predefined date formatting style to use.
+      </option>
+      <option name="timeStyle" values="full long medium short">
+        The predefined time formatting style to use.
+      </option>
+      <option name="calendar" values="buddhist chinese coptic dangi ethioaa ethiopic gregory hebrew indian islamic islamic-umalqura islamic-tbla islamic-civil islamic-rgsa iso8601 japanese persian roc">
```
aphillips, line 24: Which source did you use for these? We should have a pointer as this list is not likely to be stabilized.

spec/registry.xml
```diff
@@ -15,21 +15,25 @@
...
+        Calendar to use.
+      </option>
+      <option name="numberingSystem" values="arab arabext bali beng deva fullwide gujr guru hanidec khmr knda laoo latn limb mlym mong mymr orya tamldec telu thai tibt">
```
aphillips, line 27: This list is incomplete? Can we point to CLDR data here?

aphillips (approved):
I like the general direction. Some comments (mainly editorial).

spec/registry.xml
eemeli: I only moved the description into the body here; the values were added by @mihnita in #420; I presume that their source is the ECMA-402 spec, which may itself refer to another source.

spec/registry.xml
aphillips: ibid

spec/registry.xml
```diff
@@ -83,63 +102,71 @@
...
+      <option name="compactDisplay" values="short long" default="short">
+          Only used when notation is "compact".
+      </option>
+      <option name="currency" validationRule="currencyCode">
+        The currency to use in currency formatting.
         Possible values are the ISO 4217 currency codes, such as "USD" for the US dollar,
         "EUR" for the euro, or "CNY" for the Chinese RMB — see the
-        Current currency & funds code list
+        Current currency &amp; funds code list
         (https://www.six-group.com/en/products-services/financial-information/data-standards.html#scrollTo=currency-codes).
```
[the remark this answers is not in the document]
aphillips: CLDR would almost certainly be a better reference for a CLDR-TC subcommittee/working group to use?

Actually, probably we should point to supplmentalData.xml in the "latest" CLDR release for this. 
For numbering systems we should point to numberingSystems.xml in the latest CLDR release.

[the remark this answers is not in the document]
eemeli: Replaced with the suggested CLDR link, which is generated from the latest `supplementalData.xml`.

spec/registry.dtd
```diff
@@ -1,4 +1,7 @@
 <!ELEMENT registry (function|validationRule)*>
+<!ATTLIST registry
+    xml:lang       NMTOKEN  #IMPLIED
+>
```
eemeli, line 4: I added `xml:lang` as an optional attribute on the top-level `<registry>`, as well as directly on `<input>` and `<option>`, to indicate the language of the included descriptions.

Should this be required, though? Or are there use cases where a `<registry>` could be e.g. embedded in another XML document, where the locale is defined somewhere further up the tree?

aphillips (approved):
sheep it.

🐑

PR #534, opened 2023-11-27
Refactor registry locale overrides

### LINKED ISSUE #410: Define locale-specific variants of function signatures in the registry
_A follow-up to #368._
---
_Originally posted by @aphillips in https://github.com/unicode-org/message-format-wg/pull/368#discussion_r1187631141_:

> I'm very curious about the `locales` attribute--curious in this case is a euphemism for "nervous about". I would prefer if functions were mostly locale neutral (I can declare, for example, a number format in any message in any locale).
---
_Originally posted by @eemeli in https://github.com/unicode-org/message-format-wg/pull/368#discussion_r1199066428_:

> We should have some way to define at least locale-dependent option and match values. For example, the registry should allow for a way to note that while the whole set of CLDR categories is `zero one two few many other`, a specific locale such as `en` only uses `one other`.

> One alternative could be for `<matchSignature>` and `<formatSignature>` to be able to contain an `<override locales="en fr it">` section.
---
_Originally posted by @stasm in https://github.com/unicode-org/message-format-wg/pull/368#discussion_r1199075920_

> For completeness of the example, my plan was to allow the overrides directly on the level of the `matchSignature` and `formatSignature`:

    <function name="plural">
        <matchSignature locales="en">
            <input pattern="anyNumber"/>
            <option name="type" values="cardinal ordinal"/>
            <option .../>
            <match values="one other"/>                    ← English plurals
            <match pattern="anyNumber"/>
        </matchSignature>

        <matchSignature locales="pl">
            <input pattern="anyNumber"/>
            <option name="type" values="cardinal ordinal"/>
            <option .../>
            <match values="one few many other"/>           ← Polish plurals
            <match pattern="anyNumber"/>
        </formatSignature>
    </function>
---
_Originally posted by @aphillips in https://github.com/unicode-org/message-format-wg/pull/368#discussion_r1199092977_:

> I am not a fan of this. I think that a plural selector should recognize all of the keywords and not depend on the registry to "allow" or "disallow" the enumerated values for a given locale (note that `locales` ***must*** be a "language range", otherwise you would need to list every possible language tag and is probably an _extended_ language range, e.g. so that `pl-PL` matches `pl-Latn-PL` -- not that anyone would use that tag)

> I think it is reasonable that (for example) the `root` locale message can contain keywords that do not fire for a specific locale but which fire in others, in case that `message` is used in one of those other locales. This is also why the `*` defaulting key exists (such as the case where an English-language resource with only `one` and `*` gets used in the `pl` locale). 

> In the case of plurals, CLDR provides the data about which keywords apply to which locale. In the case of some other formatter or selector that is not supplied by CLDR, the implementation should know what applies to each locale. If we want the registry to describe that relationship (for example to support exploding the matrix of keys in localization tools), I think I would prefer that it be *separate* from the signature, e.g.:

```xml
<function name="customPluralLikeSelector">
   <matchSignature>
      <input pattern="anyNumber"/>
      <option name="type" value="foo bar"/>
      <match value="zero one two few many other"/>
      <match pattern="anyNumber"/>
   </matchSignature>
   <validate comment="the naming is terrible and we wouldn't structure it exactly like this">
      <match type="values">
         <value lang="">one other</value>
         <value lang="pl">one few many other</value>
         <value lang="ja">other</value>
         <!-- etc. --->
      </match>
   </validate>
```

eemeli:
Closes #410 

At the moment, locale-specific overrides work via the `locales` attribute list of identifiers on `<formatSignature>` and `<matchSignature>`. This is a bit clumsy, because it's mixing together two different concerns:
1. Some functions may have options that are only available in specific locales.
2. Some selectors may have their matching depend on the locale.

So I propose separating the two. Let's remove the `locales` attribute from the signatures, and rather:
1. Allow for a `<override locales="...">` within a signature that may include rules for inputs and options that override any set generally for the formatter.
2. Allow for both `<match>` and `<match locales="...">`, and specify that only one such rule (first matching one) is used at a time.

This should make it easier to implement and manage locale-specific overrides, as there's much less need for repetition.

If #532 is accepted, `<alias>` elements should also be allowed to include `<override>`s.

### DIFF
--- a/spec/registry.dtd
+++ b/spec/registry.dtd
@@ -9,12 +9,10 @@
 <!ATTLIST validationRule id ID #REQUIRED>
 <!ATTLIST validationRule regex CDATA #REQUIRED>
 
-<!ELEMENT formatSignature (input?,option*)>
+<!ELEMENT formatSignature (input?,option*,override*)>
 <!ATTLIST formatSignature position (open|close|standalone) "standalone">
-<!ATTLIST formatSignature locales NMTOKENS #IMPLIED>
 
-<!ELEMENT matchSignature (input?,option*,match*)>
-<!ATTLIST matchSignature locales NMTOKENS #IMPLIED>
+<!ELEMENT matchSignature (input?,option*,match*,override*)>
 
 <!ELEMENT input EMPTY>
 <!ATTLIST input values NMTOKENS #IMPLIED>
@@ -30,5 +28,13 @@
 <!ATTLIST option readonly (true|false) "false">
 
 <!ELEMENT match EMPTY>
-<!ATTLIST match values NMTOKENS #IMPLIED>
-<!ATTLIST match validationRule IDREF #IMPLIED>
+<!ATTLIST match
+    locales        NMTOKENS  #IMPLIED
+    values         NMTOKENS  #IMPLIED
+    validationRule IDREF     #IMPLIED
+>
+
+<!ELEMENT override (input?,option*)>
+<!ATTLIST override
+    locales        NMTOKENS  #REQUIRED
+>
--- a/spec/registry.dtd
+++ b/spec/registry.dtd
@@ -49,7 +49,7 @@
     locales        NMTOKENS  #REQUIRED
 >
 
-<!ELEMENT alias (description,setOption*)>
+<!ELEMENT alias (description,setOption*,override*)>
 <!ATTLIST alias
     name           NMTOKEN  #REQUIRED
     supports       (format|match|all)  "all"
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -141,7 +141,6 @@ The element whose `locales` best matches the current locale
 using resource item [lookup](https://unicode.org/reports/tr35/#Lookup) from LDML is used.
 An element with no `locales` attribute is the default
 (and is considered equivalent to the `root` locale).
-(or an element with no `locales`) is used:
 
 - `<match locales="en" values="one two few other" .../>` can be used in locales like `en` and `en-GB`
   to validate the `when other` variant by verifying that the `other` key is present
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -137,7 +137,10 @@ when * {{{$count :number} new messages}}
 Furthermore,
 `:number`'s `<matchSignature>` contains two `<match>` elements
 which allow the validation of variant keys.
-The first element with `locales` matching the current locale
+The element whose `locales` best matches the current locale
+using resource item [lookup](https://unicode.org/reports/tr35/#Lookup) from LDML is used.
+An element with no `locales` attribute is the default
+(and is considered equivalent to the `root` locale).
 (or an element with no `locales`) is used:
 
 - `<match locales="en" values="one two few other" .../>` can be used in locales like `en` and `en-GB`
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -50,8 +50,6 @@ A _signature_ defines one particular set of at most one argument and any number 
 that can be used together in a single call to the function.
 `<formatSignature>` corresponds to a function call inside a placeholder inside translatable text.
 `<matchSignature>` corresponds to a function call inside a selector.
-Signatures with a non-empty `locales` attribute are locale-specific
-and only available in translations in the given languages.
 
 A signature may define the positional argument of the function with the `<input>` element.
 If the `<input>` element is not present, the function is defined as a nullary function.
@@ -61,6 +59,12 @@ unless the `required` attribute is present.
 They accept either a finite enumeration of values (the `values` attribute)
 or validate their input with a regular expression (the `validationRule` attribute).
 Read-only options (the `readonly` attribute) can be displayed to translators in CAT tools, but may not be edited.
+
+As the `<input>` and `<option>` rules may be locale-dependent,
+each signature can include an `<override locales="...">` that extends and overrides
+the corresponding input and options rules.
+If multiple `<override>` elements would match the current locale,
+only the first one is used.
 
 Matching-function signatures additionally include one or more `<match>` elements
 to define the keys against which they can match when used as selectors.
@@ -93,7 +97,7 @@ For the sake of brevity, only `locales="en"` is considered.
             Match a **formatted** numerical value against CLDR plural categories or against a number literal.
         </description>
 
-        <matchSignature locales="en">
+        <matchSignature>
             <input validationRule="anyNumber"/>
             <option name="type" values="cardinal ordinal"/>
             <option name="minimumIntegerDigits" validationRule="positiveInteger"/>
@@ -102,11 +106,11 @@ For the sake of brevity, only `locales="en"` is considered.
             <option name="minimumSignificantDigits" validationRule="positiveInteger"/>
             <option name="maximumSignificantDigits" validationRule="positiveInteger"/>
             <!-- Since this applies to both cardinal and ordinal, all plural options are valid. -->
-            <match values="zero one two few many"/>
-            <match validationRule="anyNumber"/>
+            <match locales="en" values="one two few other" validationRule="anyNumber"/>
+            <match values="zero one two few many other" validationRule="anyNumber"/>
         </matchSignature>
 
-        <formatSignature locales="en">
+        <formatSignature>
             <input validationRule="anyNumber"/>
             <option name="minimumIntegerDigits" validationRule="positiveInteger"/>
             <option name="minimumFractionDigits" validationRule="positiveInteger"/>
@@ -132,14 +136,15 @@ when * {{{$count :number} new messages}}
 
 Furthermore,
 `:number`'s `<matchSignature>` contains two `<match>` elements
-which allow to validate the variant keys.
-If at least one `<match>` validation rules passes,
-a variant key is considered valid.
+which allow the validation of variant keys.
+The first element with `locales` matching the current locale
+(or an alement with no `locales`) is used:
 
-- `<match validationRule="anyNumber"/>` can be used to valide the `when 1` variant
+- `<match locales="en" values="one two few other" .../>` can be used in locales like `en` and `en-GB`
+  to valide the `when other` variant by verifying that the `other` key is present
+  in the list of enumarated values: `one other`.
+- `<match ... validationRule="anyNumber"/>` can be used to valide the `when 1` variant
   by testing the `1` key against the `anyNumber` regular expression defined in the registry file.
-- `<match values="one other"/>` can be used to valide the `when other` variant
-  by verifying that the `other` key is present in the list of enumarated values: `one other`.
 
 ---
 
@@ -152,26 +157,32 @@ A localization engineer can then extend the registry by defining the following `
 <registry>
     <function name="noun">
         <description>Handle the grammar of a noun.</description>
-        <formatSignature locales="en">
-            <input/>
-            <option name="article" values="definite indefinite"/>
-            <option name="plural" values="one other"/>
-            <option name="case" values="nominative genitive" default="nominative"/>
+        <formatSignature>
+            <override locales="en">
+                <input/>
+                <option name="article" values="definite indefinite"/>
+                <option name="plural" values="one other"/>
+                <option name="case" values="nominative genitive" default="nominative"/>
+            </override>
         </formatSignature>
     </function>
 
     <function name="adjective">
         <description>Handle the grammar of an adjective.</description>
-        <formatSignature locales="en">
-            <input/>
-            <option name="article" values="definite indefinite"/>
-            <option name="plural" values="one other"/>
-            <option name="case" values="nominative genitive" default="nominative"/>
+        <formatSignature>
+            <override locales="en">
+                <input/>
+                <option name="article" values="definite indefinite"/>
+                <option name="plural" values="one other"/>
+                <option name="case" values="nominative genitive" default="nominative"/>
+            </override>
         </formatSignature>
-        <formatSignature locales="en">
-            <input/>
-            <option name="article" values="definite indefinite"/>
-            <option name="accord"/>
+        <formatSignature>
+            <override locales="en">
+                <input/>
+                <option name="article" values="definite indefinite"/>
+                <option name="accord"/>
+            </override>
         </formatSignature>
     </function>
 </registry>
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -141,7 +141,7 @@ The first element with `locales` matching the current locale
 (or an alement with no `locales`) is used:
 
 - `<match locales="en" values="one two few other" .../>` can be used in locales like `en` and `en-GB`
-  to valide the `when other` variant by verifying that the `other` key is present
+  to validate the `when other` variant by verifying that the `other` key is present
   in the list of enumarated values: `one other`.
 - `<match ... validationRule="anyNumber"/>` can be used to valide the `when 1` variant
   by testing the `1` key against the `anyNumber` regular expression defined in the registry file.
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -138,7 +138,7 @@ Furthermore,
 `:number`'s `<matchSignature>` contains two `<match>` elements
 which allow the validation of variant keys.
 The first element with `locales` matching the current locale
-(or an alement with no `locales`) is used:
+(or an element with no `locales`) is used:
 
 - `<match locales="en" values="one two few other" .../>` can be used in locales like `en` and `en-GB`
   to validate the `when other` variant by verifying that the `other` key is present


spec/registry.md
```diff
@@ -102,11 +106,11 @@ For the sake of brevity, only `locales="en"` is considered.
             <option name="minimumSignificantDigits" validationRule="positiveInteger"/>
             <option name="maximumSignificantDigits" validationRule="positiveInteger"/>
             <!-- Since this applies to both cardinal and ordinal, all plural options are valid. -->
-            <match values="zero one two few many"/>
-            <match validationRule="anyNumber"/>
+            <match locales="en" values="one two few other" validationRule="anyNumber"/>
+            <match values="zero one two few many other" validationRule="anyNumber"/>
```
aphillips, line 110: `other` or `*`?

---

`locales="en"` looks odd. Can `locales` contain multiple values? Space separated, I guess? 

This has always struck me as a wobbly mechanism. Many locales have identical plural rules to English. Are we going to list them? Why do we need to repeat CLDR data here? Perhaps we should have a referencing mechanism to CLDR instead of replicating data. Note that other things besides plurals depend on CLDR data, such as date patterns and such.

spec/registry.md
```diff
@@ -132,14 +136,15 @@ when * {{{$count :number} new messages}}
...
-If at least one `<match>` validation rules passes,
-a variant key is considered valid.
-
-- `<match validationRule="anyNumber"/>` can be used to valide the `when 1` variant
+which allow the validation of variant keys.
+The first element with `locales` matching the current locale
+(or an alement with no `locales`) is used:
+
+- `<match locales="en" values="one two few other" .../>` can be used in locales like `en` and `en-GB`
+  to valide the `when other` variant by verifying that the `other` key is present
```
aphillips, line 144: typo

```suggestion
to validate the `when other` variant by verifying that the `other` key is present
```

spec/registry.md
```diff
@@ -132,14 +136,15 @@ when * {{{$count :number} new messages}}
 
 Furthermore,
 `:number`'s `<matchSignature>` contains two `<match>` elements
-which allow to validate the variant keys.
-If at least one `<match>` validation rules passes,
-a variant key is considered valid.
-
-- `<match validationRule="anyNumber"/>` can be used to valide the `when 1` variant
+which allow the validation of variant keys.
+The first element with `locales` matching the current locale
```
aphillips, line 140: The bullet item below says "can be used in locales like" and this says "matching" in a pretty non-specific way. That's probably not helpful to registry authors. We should specify the matching scheme (lookup is mechanical, best match is what some will want)

aphillips (changes requested):
Some thoughts below. We should discuss some of these concepts in the group.

spec/registry.md
eemeli: Probably lookup? Is there an explicit reference here we can use for that?

macchiati: BCP47 lookup is mechanical, but often produces very bad results. (Eg mixture of sr_Latn with sr_Cyrl in the same message.

Also, the matching is really bad, due to cramming values that are both for ordinal and cardinal. The valid values are going to vary depending on the options.
```
            <match locales="en" values="one two few other" validationRule="anyNumber"/>
            <match values="zero one two few many other" validationRule="anyNumber"/>
```

Moreover, it would be repeating data in CLDR, a really bad idea. It should be possible in the registry to reference external sources.

eemeli:
```suggestion
The first element with `locales` matching the current locale
according to the [BCP 47 Lookup](https://datatracker.ietf.org/doc/html/rfc4647#section-3.4) scheme
```

eemeli: For context, I'm not at all concerned about JS `Intl` here, but about much more limited environments. For a real-world example, Firefox's crash reporter is a separate executable that does not have access to the ICU that Firefox comes with, or CLDR data. Its localization will need to be wholly self-contained. It would be Very Useful to be able to still use MF2 for its localization.

As it's currently expressed, MF2 fits that bill. But if we add here a requirement to use a better matching algorithm than Lookup that e.g. knows that `sr` and `sr-Cyrl` are synonymous, we're effectively introducing a dependency on the CLDR that I'd rather we didn't.

And that's especially as I very strongly suspect that all real-world `<match>` sets will be rather reasonable ones, and so will let a mechanical first-match scheme perform just as well as any other scheme would.

aphillips: I hear you. FWIW, I'm a fan of BCP47 Lookup's mechanical approach for situations like this (resource files and the like) where you want predictable results across many implementations. I called out JS `Intl`, though, because it offers both options. The "best fit" option there is described by MDN as "provid[ing] a locale that's at least, but possibly more, suited for the request than the result of the Lookup algorithm". Allowing for both might be a good idea, if some implementations end up using ICU-based functionality under the hood.

I'm engaged in publishing some more material about this following today's W3C call. [This text](https://w3c.github.io/localizable-manifests/#locale-fallback) might also eventually appear in [LTLI](https://www.w3.org/TR/ltli) as well as other places.

For this text, I would suggest the following changes. Note that this also omits "first" from "first element...matching" since matching does not stop with the first potential match.

```suggestion
The element whose `locales` best matches the current locale is used. 
The "best match" is determined according to [BCP47 Lookup](https://datatracker.ietf.org/doc/html/rfc4647#section-3.4) 
or an appropriately-compatible implementation-specific "best fit" scheme
such as that [found in Javascript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl#locale_identification_and_negotiation)
or used by CLDR.
An element with no `locales` attribute is considered the default.
```

(Also, note that the next line in your current text misspells `element`)

spec/registry.md
```diff
@@ -132,14 +136,15 @@ when * {{{$count :number} new messages}}
...
+(or an alement with no `locales`) is used:
```
eemeli, line 141:
```suggestion
(or an element with no `locales`) is used:
```

spec/registry.md
macchiati: I disagree. It would be better to have no algorithm, and have people explicitly add the locales they want to match, then it would be to get the wrong matches. And flipping scripts can have really bad ransom-note effects.

Example:
```
 // if you really don't care which script the output is in
<match ... locales="sr sr-Cyrl sr-Latn">

// otherwise either 
<match ... locales="sr sr-Cyrl">
// or
<match ... locales="sr sr-Latn">
```

eemeli: Allowing for optionality here would mean that two implementations could end up selecting different rules when given the same registry and locale. Could we not rely on registry definitions being reasonable, and accept the first match for which Lookup succeeds?

A validator/linter for the registry XML itself could wholly separately point out to its author if it contains any mis-ordering that would be problematic.

aphillips: @macchiati noted:
>  It would be better to have no algorithm, and have people explicitly add the locales they want to match, then it would be to get the wrong matches.

This would effectively require listing *every* locale in the registry (even if we allow the current locale to have its extensions trimmed off), since there would be no prefix matching. That doesn't seem reasonable?

`plural` isn't a good example from the point of view that we want to point at the CLDR data rather than mirroring it into the registry. However, it does provide a reasonable example of what an arbitrary function would need. The CLDR data for plural rules does not list every locale. It inherits down (usually from the primary language subtag) in most cases (including, I would point out, the `sr` one). Obviously cross-script cases and others require greater care (although it doesn't affect plural rules in general).

An alternative here would be to not provide for the registry to specify locale-based variations. Tools that want to do things such as matrix explosion will then need out-of-band information. Using `plural` as an example, the variation list would be:

_anyNumber_, `zero`, `one`, `two`, `few`, `many`, `other`, (implied `*`)

It would be valid to write any locale _message_ using those keys (and, in fact, I have done this for I18N demos, particularly in the root locale). A developer writing a message and assuming the `en` locale will use keys `0`/`one`/`*`; they would want translators to expand the key set according to the target locale (so Polish ends up with `0`/`one`/`few`/`many`/`*`). The registry, though, would not help them.

For a developer writing a custom function, they won't have CLDR data to help them and so would potentially use @eemeli's feature to create this information. That way machines (CAT tools, not runtimes) could read it to generate the target language keyset. If the locale I'm creating a resource for is `sr-Latn-BS` and the registry has root and `sr`, I probably want the `sr` set and not the root.

---
@eemeli 
> accept the first match for which Lookup succeeds?

It should not be the *first* match. It should always be the *best* match. Note well that "best" is overloaded here.

Consider this example:

```
"localized-resource": {
	"en":     {"value": "This is English"},
	"en-GB":  {"value": "This is UK English"},
	"en-US":  {"value": "This is US English"}
}
```

For the locale `en-GB-oxendict`, `en` is the *first* match, but it is not the *best* match. `en-GB` is the *best* match. (This is pure Lookup, please note). A "best fit" matcher could decide that a request for `br` (Breton) matches `fr` before falling back to root (to use one of Mark's favorite examples) and that would not be BCP47 Lookup.

eemeli: > Consider this example:
> 
> ```
> "localized-resource": {
> 	"en":     {"value": "This is English"},
> 	"en-GB":  {"value": "This is UK English"},
> 	"en-US":  {"value": "This is US English"}
> }
> ```
> 
> For the locale `en-GB-oxendict`, `en` is the _first_ match, but it is not the _best_ match. `en-GB` is the _best_ match. (This is pure Lookup, please note). A "best fit" matcher could decide that a request for `br` (Breton) matches `fr` before falling back to root (to use one of Mark's favorite examples) and that would not be BCP47 Lookup.

Ah! I was not previously clear in what I meant by e.g. "reasonable" when describing a registry. Apologies.

In the XML, we don't have a mapping as above, but an ordered sequence of e.g. `<match>` statements. If the Lookup is done by selecting first-match, then the "reasonable" order for the ones you list would be:

```xml
<match locales="en-GB"/>
<match locales="en-US"/>
<match locales="en"/>
```

With that order, a first-match selection would produce the same results as best-match, because the prioritisation ends up being encoded positionally in the XML.

aphillips: At the cost of requiring the tags to be ordered. Moreover, there are sets of tags that might be difficult to order (e.g. `zh-CN` vs. `zh-Hans` vs. `zh-Hans-CN`). There are other ways in XML to perform selection which might be more appropriate here.

---

Part of me is still puzzled by the need to represent locales in the registry in this way. I think that a `match` element represents _potential_ output values of a selector (in a specific locale set when it has `locales`). It doesn't represent the specific outputs (which depend on evaluating the inputs). So it isn't consumed by the runtime (which would just call the function anyway). Instead it describes the potential outputs for cases such as I describe above (CAT tools exploding or contracting the variant matrix). 

Is that what you're thinking this is for? If that's the case, is that "nice to have" for release? Or do you think there will be a lot of cases where there are selectors whose potential output varies outside a closed set for different locales, e.g.

(default): `a`, `b`, `c`
en: `b`, `c`, `d`
fr: `e`, `f`, `g`

Such that matrices look different for a given match statement in a message. 

```
.match {$var :function}
[a] {{Default locale can get here}}
[b] {{default or en}}
[c] {{default or en}}
[d] {{en only}}
[e] {{fr only}}
[f] {{fr only}}
[g] {{fr only}}
[*] {{everyone can get here}}
```

eemeli: > At the cost of requiring the tags to be ordered. Moreover, there are sets of tags that might be difficult to order (e.g. `zh-CN` vs. `zh-Hans` vs. `zh-Hans-CN`). There are other ways in XML to perform selection which might be more appropriate here.

The tricky bit here is of course what to do if we have
```xml
<match locales="zh-CN"/>
<match locales="zh-Hans"/>
```
and our current locale is `zh-Hans-CN`. And I'd argue that with first-match it's actually easier to resolve this than with best-match, as the ordering can define for this selector in particular whether to prioritise the script or the region.

In practice, I would be surprised if any real-world selectors turn out to depend not only on the primary subtag, but also the region _and_ script. For example, with plurals the _only_ one with any variance beyond the primary is `pt-PT` vs. `pt`.

> Part of me is still puzzled by the need to represent locales in the registry in this way. [...] Instead it describes the potential outputs for cases such as I describe above (CAT tools exploding or contracting the variant matrix).
> 
> Is that what you're thinking this is for? If that's the case, is that "nice to have" for release? [...]

This has no impact on formatting, correct, but the improvement we're offering for tooling is significant enough that if we get this right, it may by itself end up as a reason for promoting migration to MF2. At the moment, no data definition like this exists for any format. So if e.g. a CAT tool wants to be able to map plural categories correctly when translating, it needs to figure out a custom way of getting the data for that, and applying it.

Instead, here we're defining a regular shape to such data, and not just for plurals but any other selector as well. And not just to proprietary CAT tools, but _all_ such tools so that _any_ validator based on MF2 registry data can validate the result coming from _any_ CAT tool.

I think of what's proposed here as a minimum we'll need for the release. There are further aspects that I think we'll want to follow up with possibly post-release that are "nice to have", such as being able to encode data for the intersection of "integer value" and "plural selector", or how/when leaving out the `many` plural case for Romance languages is appropriate. On that note: we ought to ensure that we can version the registry definition separately from the spec, so that it can experience breaking changes.

aphillips: > Instead, here we're defining a regular shape to such data, and not just for plurals but any other selector as well. And not just to proprietary CAT tools, but all such tools so that any validator based on MF2 registry data can validate the result coming from any CAT tool.

Okay, so we're in agreement about the objective here, I think.

Yes, I agree that MF2 processors such as CAT tools or resource editors or other parts of the ecosystem should be able to have a machine-readable way of figuring out how to manage variant sets that vary by locale. 

For cases in MF2's default registry, I believe that the source data for such information will come from existing CLDR data (such as the plural rules!) and that we should provide an interface to that data.

If the existing CLDR data cannot be referenced in-situ and a transform is required, we should include that transform in our release and ensure that each new CLDR release automagically updates the MF2 registry. **This means the MF2 default registry will be versioned with CLDR.** (This is not necessarily a bad thing)

What you're defining here (to go back on topic) is the registry structure for selectors (including custom ones written by users of a given implementation). The test of this would be our ability to transform CLDR data for built-in selectors to deliver that format.

Assuming that this is correct, the locale matching we specify probably needs to exactly match that used by CLDR. As you note, plural rules are not exactly a deep and exacting test of this. Ditto ordinals. Maybe that's a good thing too. In any case, we can can specify how `<match locales=...>` is matched by referencing LDML [roughly here](https://unicode.org/reports/tr35/#Bundle_vs_Item_Lookup). This is a higher bar than BCP47 Lookup, but in practice it is extremely similar. The main problem is `addLikelySubtags`, which can be implemented with a small amount of code and about 14k of (uncompressed) textual data in my implementation (I didn't spend any effort on optimizing for size). Most locales won't be affected. In fact, **_we_** won't need it for the default registry values we produce. There can be a test or tests for "registry consumers", by which we do not mean MF2 formatter implementations. Those will just call the selector.

eemeli: Okay, you're right.

Thinking this through again, I'd not accounted for the registry being completely irrelevant during formatting, where the resource constraints are tight. Let us indeed use CLDR matching exactly. That'll also obviate the need for any plural data transforms to re-order their results, which could be quite hard to express in XSLT.

Could you suggest language here to that effect?

aphillips:
```suggestion
The element whose `locales` best matches the current locale 
using resource item [lookup](https://unicode.org/reports/tr35/#Lookup) from LDML is used.
An element with no `locales` attribute is the default 
(and is considered equivalent to the `root` locale).
```

spec/registry.md
```diff
@@ -132,14 +136,18 @@ when * {{{$count :number} new messages}}
...
-If at least one `<match>` validation rules passes,
-a variant key is considered valid.
-
-- `<match validationRule="anyNumber"/>` can be used to valide the `when 1` variant
+which allow the validation of variant keys.
+The element whose `locales` best matches the current locale
+using resource item [lookup](https://unicode.org/reports/tr35/#Lookup) from LDML is used.
+An element with no `locales` attribute is the default
+(and is considered equivalent to the `root` locale).
+(or an element with no `locales`) is used:
```
aphillips, line 144:
```suggestion
```

This line should go away since the preceding sentences cover it.

aphillips (approved):
Minor change needed. Otherwise good.

aphillips (approved):

aphillips (approved):
Merge looks clean.

spec/registry.dtd
```diff
@@ -43,10 +39,16 @@
...
     values         NMTOKENS  #IMPLIED
     validationRule IDREF     #IMPLIED
 >
 
+<!ELEMENT override (input?,option*)>
+<!ATTLIST override
+    locales        NMTOKENS  #REQUIRED
+>
+
 <!ELEMENT alias (description,setOption*)>
```
eemeli, line 52: As mentioned in the description:
```suggestion
<!ELEMENT alias (description,setOption*,override*)>
```

stasm:
I'm sorry that I missed this conversation when it happened. I hope we can still reconsider the design choices made in this PR. Or at least better substantiate it.

> At the moment, locale-specific overrides work via the `locales` attribute list of identifiers on `<formatSignature>` and `<matchSignature>`. This is a bit clumsy, because it's mixing together two different concerns:
> 
> 1. Some functions may have options that are only available in specific locales.
> 2. Some selectors may have their matching depend on the locale.

@eemeli Can you elaborate why you think these are two different concerns? In my mind, they are a single concern -- the selector function will be called in the context of a specific locale which impacts the function's input and output types, i.e. the function's signature.

I'd much prefer to build as flat a registry as possible, and then separately define how the data modelled by it should be used. So rather than nest overrides inside `matchSignatures`, I'd propose that multiple `matchSignature` be used and the rules of specificity be applied to choose the right ones.

```xml
<function name="number">
  <!-- A generic signature, common for all locales. -->
  <matchSignature>
    <input validationRule="anyNumber"/>
    <option name="type" values="cardinal ordinal"/>
    <option name="minimumIntegerDigits" validationRule="positiveInteger"/>
    <option name="minimumSignificantDigits" validationRule="positiveInteger"/>
    <option name="maximumSignificantDigits" validationRule="positiveInteger"/>
    <!-- Two match elements, because it's a union. -->
    <match validationRule="anyNumber"/>
    <match values="zero one two few many other"/>
  </matchSignature>

  <matchSignature locales="en">
    <!-- Narrow down the possible keys for the en locale. -->
    <match values="one other"/>
  </matchSignature>

  <matchSignature locales="pl">
    <!-- Narrow down the possible keys for the pl locale. -->
    <match values="one few many other"/>
  </matchSignature>
</function>
```

PR #535, opened 2023-11-27
Update open-close design doc to accept the hash & slash alternative

eemeli:
This reflects the consensus reached during the 2023-11-20 call, to accept `{#foo}...{/foo}` as the open-close syntax.

The previously proposed `{#foo/}` syntax for standalone elements did not reach consensus, and so is left out here.

Examples are updated to use the new syntax, and the `{+foo}...{-foo}` alternative is documented as such.

### DIFF
--- a/exploration/open-close-expressions.md
+++ b/exploration/open-close-expressions.md
@@ -1,6 +1,6 @@
 # Open/Close Expressions
 
-Status: **Proposed**
+Status: **Accepted**
 
 <details>
 	<summary>Metadata</summary>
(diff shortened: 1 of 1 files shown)


exploration/open-close-expressions.md
```diff
@@ -1,6 +1,6 @@
 # Open/Close Expressions
 
-Status: **Proposed**
+Status: **Accepted**
```
eemeli, line 3: @aphillips Didn't we agree on the call to merge this PR as proposed, re: #537?

```suggestion
Status: **Proposed**
```

stasm:
During the teleconference on Monday I supoprted merging this PR as proposed. However, I had forgotten about this part:

> The previously proposed `{#foo/}` syntax for standalone elements did not reach consensus, and so is left out here.

I think we should phrase this differently. Right now, the doc recognizes that standalone elements are still an open question, but the documented proposal goes one step further: it suggests to use the registry to encode the standalone aspect. I filed #542 with my suggestion for how to improve the doc.

PR #539, opened 2023-11-28
Clarify fallback resolution

gibson042:
* Consistently phrase "fail(s) to resolve" and italicize "fallback value".
* Improve examples by showing both input and output.
* Include previously missing cases:
  * non-function annotations (e.g., `{@reserved …}`, `{|…| @reserved}`, `{$var @reserved}`)
  * literals that need escaping (e.g., `{|C:\\| :func}`)
  * non-annotated variables (e.g., `{$noSuchVar}`)
  * transitive fallback (e.g., `.local $bad = {:noSuchFunc} {{{$bad}}}`)
* Remove unused "otherwise" case.

Also includes a syntax change to require a function-like identifier for every annotation, in support of defining the fallback value for expressions like `{@reserved}` and `{^private}`.

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -309,17 +309,24 @@ The _fallback value_ depends on the contents of the _expression_:
   > In a context where `:func` fails to resolve, `{:func}` resolves to the _fallback value_ `:func`.
   > In a context where `:ns:func` fails to resolve, `{:ns:func}` resolves to the _fallback value_ `:ns:func`.
 
-- non-_function_ _annotation_ (i.e., _private-use_ or _reserved_) consisting of a _sigil_-prefixed _identifier_
-  optionally followed by a whitespace-prefixed tail:
-  the _annotation_ starting sigil followed by that _identifier_
+- unsupported _private-use annotation_ or _reserved annotation_ with no _operand_:
+  the _annotation_ starting sigil
 
   > Examples:
-  > In any context, `{@reserved}` and `{@reserved |...|}` both resolve to the _fallback value_ `@reserved`.
+  > In any context, `{@reserved}` and `{@reserved |...|}` both resolve to the _fallback value_ `@`.
+
+- supported _private-use annotation_ with no _operand_:
+  the _annotation_ starting sigil, optionally followed by implementation-defined details
+  conforming with patterns in the other cases (such as quoting literals).
+  If details are provided, they SHOULD NOT leak potentially private information.
+
+  > Examples:
+  > In a context where `^` expressions are used for comments, `{^▽^}` might resolve to the _fallback value_ `^`.
+  > In a context where `&` expressions are _function_-like macro invocations, `{&foo |...|}` might resolve to the _fallback value_ `&foo`.
 
 - Otherwise: the U+FFFD REPLACEMENT CHARACTER `�`
 
-  > Examples:
-  > In any context, `{@reserved\\}` and `{? reserved ?}` both resolve to the _fallback value_ `�`.
+  This is not currently used by any expression, but may apply in future revisions.
 
 _Option_ _identifiers_ and values are not included in the _fallback value_.
 


spec/message.abnf
```diff
@@ -60,7 +60,7 @@ private-start  = "^" / "&"
 ; future versions of this specification
 reserved       = reserved-start reserved-body
 reserved-start = "!" / "@" / "#" / "%" / "*" / "<" / ">" / "/" / "?" / "~"
-reserved-body  = *( [s] 1*(reserved-char / reserved-escape / quoted))
+reserved-body  = name *(s 1*(reserved-char / reserved-escape / quoted))
```
gibson042, line 63: I hope it is non-controversial to reject expressions like `{@}` and `{^|foo|}` and `{& bar}` as syntactically invalid.

eemeli: This would preclude the possibility of using reserved placeholders as inline comments. For example:
```
The meeting repeats fortnightly {? every second week ?} on Thursdays
```
Or as protection markers:
```
This text is {<}<b>{>}bold{<}</b>{>}
```

I'm not suggesting that we should ever do either of the above, but that restricting `reserved-body` like this would need to be explicitly discussed.

eemeli (changes requested):
I like the fallback resolution changes, but not the `reserved-body` changes. See inline comment reply for more.

spec/message.abnf
gibson042: Absolutely! If we do want to preserve that possibility, I think I would suggest a slightly different refactor:
```diff
-reserved-body  = *( [s] 1*(reserved-char / reserved-escape / quoted))
+reserved-body       = [s] [(1*name-char / reserved-body-start) *(s reserved-body-part) [s]]
+reserved-body-start = *name-char (non-name-char / reserved-escape / quoted) [reserved-body-part]
+reserved-body-part  = 1*(reserved-char / reserved-escape / quoted)
+non-name-char       = %x21-2C        ; omit - .
+                    / %x2F           ; omit digits and :
+                    / %x3B-40        ; omit uppercase letters
+                    / %x5B           ; omit \
+                    / %x5D-5E        ; omit _
+                    / %x60           ; omit lowercase letters and { | }
+                    / %x7E-D7FF      ; omit surrogates
+                    / %xE000-10FFFF
+reserved-char       = non-name-char / ALPHA / DIGIT / "-" / "." / ":" / "_"
```
And updating the fallback behavior to again emit `{�}` when the first `1*name-char` doesn't match.

aphillips:
> Also includes a syntax change to require a function-like identifier for every annotation, in support of defining the fallback value for expressions like {@reserved} and {^private}.

We explicitly did not add this the first time and, in particular, allow a lot of freedom for private use. If people want to use private use for comments, it's their funeral...

spec/formatting.md
```diff
@@ -175,7 +176,7 @@ whether its value was originally _quoted_ or _unquoted_.
 > For example,
 > the _option_ `foo=42` and the _option_ `foo=|42|` are treated as identical.
 
-The resolution of a _text_ or _literal_ token MUST always succeed.
+The resolution of a _text_ or _literal_ token MUST never fail to resolve.
```
aphillips, line 179: "MUST never" is not proper 2119

```
The resolution of a _text_ or _literal_ token MUST NOT fail to resolve.
```

... which is still a double negative, so the "suggestion" would be:

```suggestion
The resolution of a _text_ or _literal_ MUST resolve to a string.
```

spec/formatting.md
```diff
@@ -230,65 +231,54 @@ the following steps are taken:
 
 ### Fallback Resolution
 
-A **_fallback value_** is the resolved value emitted when an _expression_ cannot be resolved.
+A **_fallback value_** is the resolved value for an _expression_ that fails to resolve.
 
 An _expression_ fails to resolve when:
 
-- A _variable_ _operand_ fails to resolve.
+- A _variable_ used as complete _expression_ contents or as an _operand_ fails to resolve.
```
aphillips, line 238: "complete expression contents" is the same thing as an operand.

```suggestion
A _variable_ used as an _operand_ fails to resolve.
(This includes _expressions_ containing only an _operand_.)
```

spec/formatting.md
```diff
@@ -230,65 +231,54 @@ the following steps are taken:
...
-
-  > Examples: `|your horse|`, `|42|`
+- _expression_ with _annotation_ and no _operand_:
+  the _annotation_ starting sigil followed by its _identifier_
 
-- _expression_ with _variable_ _operand_: U+0024 DOLLAR SIGN `$`
-  followed by the _variable_ _name_ of the _operand_
+  > Examples:
+  > In a context where `:func` fails to resolve, `{:func}` resolves to the _fallback value_ `:func`.
+  > `{@reserved |...|}` always resolves to the _fallback value_ `@reserved`.
```
aphillips, line 252: Be consistent in phrasing

```suggestion
`{@reserved |...|}` resolves to the _fallback value_ `@reserved`.
```

spec/formatting.md
```diff
@@ -230,65 +231,54 @@ the following steps are taken:
...
+  and then by U+007C VERTICAL LINE `|`.
+  The same representation is used for both _quoted_ and _unquoted_ values.
 
-- _expression_ with no _operand_:
-  the _function_ starting sigil followed by its _name_
+  > Examples:
+  > In a context where `:func` fails to resolve,
+  > `{42 :func}` resolves to the _fallback value_ `|42|` and
+  > `{|C:\\| :func}` resolves to the _fallback value_ `|C:\\|`.
+  > `{|| @reserved}` always resolves to the _fallback value_ `||`.
```
aphillips, line 265: The fallback loses the `{`/`}`

spec/formatting.md
```diff
@@ -230,65 +231,54 @@ the following steps are taken:
...
-_Option_ names and values are not included in the _fallback value_.
+  - If the _variable_ resolves to a _fallback value_: the _fallback value_ to which it resolves
 
-When an error occurs in an _expression_ with a _variable_ _operand_
-and the _variable_ refers to a local _declaration_,
-the _fallback value_ is formatted based on the _expression_
-on the right-hand side of the _declaration_,
-rather than the _expression_ in the _selector_ or _pattern_.
+    > In a context where `:now` fails to resolve but `:datetime` does not,
+    > `.local $t = {:now format=iso8601} .local $pretty_t = {$t :datetime} {{{$pretty_t}}}` resolves to the _fallback value_ `:now`.
```
aphillips, line 279: Two comments:

1. This is tricky. If `$t` is evaluated eagerly (which we allow), then `$pretty_t` might contain a different sort of error (since `$t` contains a string that isn't a representative of a date/time value). Is this what we want here?

2. I find the examples hard to read vs. what we had before. I think the whitespace and multiline presentation of the original example (which is fine to delete) could be beneficial to replicate in the new examples.

spec/formatting.md
```diff
@@ -974,7 +964,7 @@ SHOULD prioritise Syntax and Data Model errors over others.
 
 When an error occurs in the resolution of an _option_,
 the surrounding _expression_ MUST be processed as if the _option_ were not present.
-This MAY allow the _expression_ to resolve to a non-fallback _value_,
+This MAY allow the _expression_ to resolve to a value that is not a _fallback value_,
```
aphillips, line 967: It's incorrect to have normative MAY here. We mean to say:

```suggestion
This can result in the _expression_ resolving to a value that is not a _fallback value_.
```

aphillips (changes requested):
Some comment. Good start.

spec/message.abnf
eemeli: Would it be possible to leave the `reserved-*` syntax updates out of this PR, and handle them separately?

spec/formatting.md
gibson042: The braces are added outside of these steps, cf. [Formatting Fallback Values](https://github.com/unicode-org/message-format-wg/blob/main/spec/formatting.md#formatting-fallback-values). I've updated the relevant examples to clarify.

spec/formatting.md
gibson042: The second point is now handled, but the first seems like a consequence of logic we already have—I'm just highlighting its transitive nature:
1. `:now` fails to resolve, so `.local $t = {:now format=iso8601}` resolves to the fallback value ":now".
2. Therefore `.local $pretty_t = {$t :datetime}` resolves to the same fallback value ":now", even in the [current text](https://github.com/unicode-org/message-format-wg/blob/c0abd4f793d0a6f10807b5f8f2d44fe016860db2/spec/formatting.md#fallback-resolution)—"When an error occurs in an <em>expression</em> with a <em>variable</em> <em>operand</em> and the <em>variable</em> refers to a local <em>declaration</em>, the <em>fallback value</em> is **formatted based on the <em>expression</em> on the right-hand side of the <em>declaration</em>**…", followed by an explicit example, and also "If a <em>variable</em> would resolve to a <em>fallback value</em>,
this MUST also be considered a failure" text in [Variable Resolution](https://github.com/unicode-org/message-format-wg/blob/c0abd4f793d0a6f10807b5f8f2d44fe016860db2/spec/formatting.md#variable-resolution)
3. Therefore `{$pretty_t}` resolves to the same fallback value ":now" for the same reason.

spec/message.abnf
gibson042: Done.

spec/formatting.md
eemeli: The main intent with the current transitive logic is to prevent the name of a local variable from ever showing up in the output, either during fallback or as the `source` of a formatted part.

Local variables should be considered message-internal implementation details that should never leak outside it.

gibson042: ...which I believe is demonstrated in the new example.

spec/formatting.md
```diff
@@ -230,65 +231,77 @@ the following steps are taken:
...
+  followed by the value of the _literal_
+  with escaping applied to U+005C REVERSE SOLIDUS `\` and U+007C VERTICAL LINE `|`,
+  and then by U+007C VERTICAL LINE `|`.
+  The same representation is used for both _quoted_ and _unquoted_ values.
 
-_Option_ names and values are not included in the _fallback value_.
+  > Examples:
+  > In a context where `:func` fails to resolve,
+  > `{42 :func}` resolves to the _fallback value_ `|42|` and
+  > `{|C:\\| :func}` resolves to the _fallback value_ `|C:\\|`.
```
aphillips, line 277: We preserve the escaped value (such as `\\`, `\|` or `\{`) here, but presumably these should be _unescaped_ at the end of processing (as they would be if the literal were processed normally). The `\` escape is syntax to allow the literal's actual value to contain the escaped character.

spec/formatting.md
```diff
@@ -230,65 +231,77 @@ the following steps are taken:
...
+  > In any context, `{|| @reserved}` resolves to the _fallback value_ `||`.
 
-When an error occurs in an _expression_ with a _variable_ _operand_
-and the _variable_ refers to a local _declaration_,
-the _fallback value_ is formatted based on the _expression_
-on the right-hand side of the _declaration_,
-rather than the _expression_ in the _selector_ or _pattern_.
+- _expression_ with _variable_ _operand_ (with or without an _annotation_):
+  - If the _variable_ fails to resolve, or resolves to a value that is not a _fallback value_:
```
aphillips, line 281: I found this confusing. You have to read everything here, including the next case, to understand how a fallback value could get into the variable. Unless you know about the transitive case, this appears to say "all variables". I would suggest putting the variable fails to resolve case first, then the transitive case (starting line 291) and end with an "otherwise" case for variables that resolve correctly but whose expressions fail.

Yes, that will be slightly repetitive in the last case.

spec/formatting.md
```diff
@@ -230,65 +231,77 @@ the following steps are taken:
...
+  > Examples:
+  > In a context where `:func` fails to resolve, `{:func}` resolves to the _fallback value_ `:func`.
+  > In a context where `:ns:func` fails to resolve, `{:ns:func}` resolves to the _fallback value_ `:ns:func`.
 
-- _expression_ with _variable_ _operand_: U+0024 DOLLAR SIGN `$`
-  followed by the _variable_ _name_ of the _operand_
+- _expression_ with non-_function_ _annotation_ and no _operand_:
+  - If the _annotation_ consists of a _sigil_-prefixed _identifier_
+    optionally followed by a whitespace-prefixed tail:
+    the _annotation_ starting sigil followed by that _identifier_
```
eemeli, line 257: This sounds an awful lot like someone setting up a follow-up change to reserved syntax. 😇

spec/formatting.md
eemeli: Shouldn't this be as follows, given the text above on escaping `\`?
```suggestion
  > `{|C:\\| :func}` resolves to the _fallback value_ `|C:\\\\|`.
```

eemeli (approved):

spec/formatting.md
eemeli: Aah, you're right. Sorry for the noise.

spec/formatting.md
```diff
@@ -246,65 +247,81 @@ the following steps are taken:
...
+    the _annotation_ starting sigil followed by its _identifier_
 
-  > Examples: `|your horse|`, `|42|`
+    > Examples:
+    > In a context where `:func` fails to resolve, `{:func}` resolves to the _fallback value_ `:func`.
+    > In a context where `:ns:func` fails to resolve, `{:ns:func}` resolves to the _fallback value_ `:ns:func`.
 
-- _expression_ with _variable_ _operand_: U+0024 DOLLAR SIGN `$`
-  followed by the _variable_ _name_ of the _operand_
+  - If the non-_function_ _annotation_ consists of a _sigil_-prefixed _identifier_
```
aphillips, line 271: A "non-_function_ _annotation_" is the same as `reserved-annotation`+`private-use-annotation`. Is there some reason to be elliptical about it?

```suggestion
If the _annotatation_ is a _reserved annotation_ or a _private-use annotation_
and the `reserved` or `private-use` sigil is immediately followed by a character
sequence that is an _identifier_, 
make the _fallback value_ the sigil followed by that _identifier_
```

spec/formatting.md
```diff
@@ -246,65 +247,81 @@ the following steps are taken:
...
+
+  - Otherwise: U+0024 DOLLAR SIGN `$` followed by the _name_ of the _variable_
+
+    > Examples:
+    > In a context where `$var` fails to resolve, `{$var}` and `{$var :number}` and `{$var @reserved}`
+    > all resolve to the _fallback value_ `$var`.
+    > In a context where `:func` fails to resolve,
+    > the _pattern_'s _expression_ in `.input $arg {{{$arg :func}}}`
+    > resolves to the _fallback value_ `$arg` and
+    > the message formats to `{$arg}`.
```
aphillips, line 322: I think this whole block should be moved to about line 262 so that the block is effectively structured:

> The fallback value depends on the contents of the expression:
> - expression with literal operand
>   - emit the literal
> - expression with variable operand
>   - emit the variable
> - expression with no operand but with a function
>   - emit the function name
> - expression with no operand but with reserved/private
>   - if sigil+identifier, emit that, else fall through to:
> - otherwise emit the logo

aphillips (changes requested):
Close, but I think some further structuring will make this easier to read/implement

spec/formatting.md
gibson042: I figure it's pretty much "function" vs. "other", because there's no differentiation here between private-use vs. reserved annotations. But your suggestion does more than just rephrase, because the instances of the latter do not necessarily start with a sigil-prefixed _identifier_ (e.g., consider `{@ reserved}` and `{>}` and `{&|quoted|}`).

spec/formatting.md
gibson042: That also would be a behavior change, because not every variable operand is represented as a variable (specifically, a variable referencing a `.local` declaration is transparently treated as its resolved value).

aphillips: Fair enough, but the above is a short representation of the whole. Do you disagree that reordering the text would make it clearer?

> The fallback value depends on the contents of the expression:
> - expression with literal operand
>   - emit the literal
> - expression with variable operand
>   - if the variable references a `.local` emit the resolved value
>   - else emit the variable
> - expression with no operand but with a function
>   - emit the function name
> - expression with no operand but with reserved/private
>   - if sigil+identifier, emit that, else fall through to:
> - otherwise emit the logo

gibson042: I think a single-level list of cases with no nested branches _is_ clearer, so I've updated accordingly (basically just promoting "variable referencing `.local`" vs. "variable not referencing `.local`" into distinct cases).

spec/formatting.md
```diff
@@ -115,7 +115,8 @@ and a _function_, _private-use_, or _reserved_ _annotation_,
 the resolved value of the _expression_ is determined as follows:
 
 If the _expression_ contains a _reserved_ _annotation_,
-an `Unsupported Expression` error is emitted and a fallback value is used as its value.
+an Unsupported Expression error is emitted and
```
aphillips, line 118:
```suggestion
an `Unsupported Expression` error is emitted and
```

(missing the quoting of the error type?)

gibson042: No, updating it to match the prevailing convention—all of the other error types ("Unresolved Variable error", "Unknown Function error", etc.) in https://github.com/unicode-org/message-format-wg/blob/main/spec/formatting.md are unformatted.

spec/formatting.md
aphillips: Those instances would not match my text and they would emit the logo. What's our intention here? Do we want to pass reserved/private garbage unmolested? Or turn them into the logo? My reading of the current text is that we make them into the logo unless they look like `{^identifier something something}` where we emit `{^identifier}` for some reason.

gibson042: [Current text](https://github.com/unicode-org/message-format-wg/blob/main/spec/formatting.md#fallback-resolution) is incoherent—the predicate of "_expression_ with no _operand_: the _function_ starting sigil followed by its _identifier_" covers non-function expressions, but the consequent cannot be applied to them. That can be fixed by always having such expressions fall back to the logo, or (as I have done here) by applying function-like treatment to those that use function-like syntax. But I consider it important for "function-like" to be defined narrowly enough to exclude e.g. `{@spaceless|quoted|}`.

aphillips: It's almost coherent except for the possibility of reserved/private... which is what we're fixing here. Because `reserved-annotation` and `private-use-annotation` are permissive about content and because we don't have a construct for parsing _inside_ them in the ABNF (nor, I think, should we...) I think emitting the logo makes some sense.

Putting myself in the shoes of a user, this is the rendered output of a message containing some placeholders that my implementation doesn't understand. If we emit their contents, yes, that might assist in debugging. But if shown to users it might contain code internal stuff. The logo basically is an admission that your implementation has no idea what this placeholder means. Does that make sense?

Note: it's probably important that we emit _something_ as a way of preventing sneak attacks such as `{{<scr{@omit me}ipt href=...>}}`

eemeli: There's a bit further down in the spec that probably ought to be highlighted here as well: https://github.com/unicode-org/message-format-wg/blob/3f35bee287671f5162d8599fd0ee85bbf7682759/spec/formatting.md?plain=1#L591-L595

The intent with that is that even though our default fallback must guard against e.g. `{$user :name}` having a failure mode where a custom stringifier of the `$user` value gets called, dumping personal data into the output, there are still plenty of places where a better fallback than the default can and should be defined.

But it does mean that the space of fallback values cannot be considered exhaustive, and so a reserved expression producing `{�}` fallback now should not be considered a restriction on what might happen if we introduce a new meaning for its sigil later. Also, we should add language somewhere allowing an implementation to define separately the fallback behaviour for private-use sigils.

Put together, I think we _could_ default to `{�}` fallback for reserved and private-use, but make it clear(er) that later spec versions or implementations supporting private-use can and should define something better.

macchiati: I also think that it is important to indicate to the user that a fallback happened : that is, what is displayed to the user is **_not_** the intended result. One possibility is prefixing and/or suffixing with �.

aphillips: @macchiati Perhaps? Presumably the presence of curly-bracketed garbage is an indicator. But having &#xfffd; present consistently would make it easier to find and easier to check for. If we did this, I'd make it a prefix so that we'd have fallbacks like:

```
{�}
{�$var}
{�|42.7 some literal|}
{�:function}
{�^unsupported}
```

aphillips:
In the 2023-12-04 teleconference we resolved to *not* use the U+FFFD character except as the ultimate fallback and to use only the private-use/reserved sigil (no groping) as fallback for those. This makes the patterns:

```
{�}
{$var}                 // some variable
{|42.7 some literal|}  // some literal
{:function}            // some function
{^}                    // reserved/private-use exposes the sigil used, ^ is only an example
```

gibson042:
> In the 2023-12-04 teleconference we resolved to _not_ use the U+FFFD character except as the ultimate fallback and to use only the private-use/reserved sigil (no groping) as fallback for those. This makes the patterns:
> 
> ```
> {�}
> {$var}                 // some variable
> {|42.7 some literal|}  // some literal
> {:function}            // some function
> {^}                    // reserved/private-use exposes the sigil used, ^ is only an example
> ```

These changes are complete, along with a case for supported private-use expression falling back to the sigil plus optional imlementation-specific details and a note that `�` is not used for expression fallback values in this revision (since the cases are exhaustive).

aphillips (approved):
Looks good. Thanks!

aphillips:
As agreed in teleconference, merging these changes.

PR #540, opened 2023-11-29
Clarify that open-close elements are placeholders, not expressions

eemeli:
I noticed that the open-close design doc was in places (like its title...) talking about _expressions_ when it's really about _placeholders_. In our current syntax, those are synonymous, but the change we're proposing here would make them different, and while _markup_ would be a _placeholder_, it would not be an _expression_.

Requesting fast-tracking, as we're currently discussing this in #537.

### DIFF
--- a/exploration/open-close-placeholders.md
+++ b/exploration/open-close-placeholders.md
@@ -16,6 +16,7 @@ Status: **Proposed**
 		<dd><a href="https://github.com/unicode-org/message-format-wg/pull/516">#516</a></dd>
 		<dd><a href="https://github.com/unicode-org/message-format-wg/pull/517">#517</a></dd>
 		<dd><a href="https://github.com/unicode-org/message-format-wg/pull/535">#535</a></dd>
+		<dd><a href="https://github.com/unicode-org/message-format-wg/pull/540">#540</a></dd>
 	</dl>
 </details>
 
--- a/exploration/open-close-placeholders.md
+++ b/exploration/open-close-placeholders.md
@@ -1,4 +1,4 @@
-# Open/Close Expressions
+# Open/Close Placeholders
 
 Status: **Proposed**
 
@@ -23,7 +23,7 @@ Status: **Proposed**
 
 _What is this proposal trying to achieve?_
 
-Describe the use cases and requirements for _expressions_
+Describe the use cases and requirements for _placeholders_
 that enclose parts of a pattern
 and develop a design that satisfies these needs.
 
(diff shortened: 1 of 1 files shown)


aphillips:
I think that's not quite true. All placeholders are expressions and they would still be expressions after all the various changes. However, the converse would not hold true. Not all expressions are placeholders. So the current text isn't wrong (with current spec wording) and wouldn't technically be wrong post all the changes. But I do think we should clarify the doc, hence giving an approval.

I will note that @stasm has proposed alternatives that do not use placeholders to solve spannable. I don't personally care for these options, but they would use functions to implement spannable and those would be expressions in all senses of the term.

aphillips (approved):

exploration/open-close-placeholders.md
```diff
@@ -1,4 +1,4 @@
-# Open/Close Expressions
+# Open/Close Placeholders
```
aphillips, line 1: I think we would be better off calling this design "spannable" but don't care *that* much about it.

eemeli: All of the options currently being considered are `placeholders` in a pattern; hence my suggested name for this.

eemeli:
> I think that's not quite true. All placeholders are expressions and they would still be expressions after all the various changes. However, the converse would not hold true. Not all expressions are placeholders. So the current text isn't wrong (with current spec wording) and wouldn't technically be wrong post all the changes. But I do think we should clarify the doc, hence giving an approval.

Do note that the proposed design includes this change to the ABNF: https://github.com/unicode-org/message-format-wg/blob/2bd3e6e828c231e294eb98b2deb4ec8067a8d6e9/exploration/open-close-expressions.md?plain=1#L163-L165

The explicit definition of the rule as _not_ being an expression is needed to prevent it from showing up in e.g. local variable declarations. So after the changes here, not all placeholders will be expressions.

exploration/open-close-placeholders.md
stasm: +1 to titling this design doc "Spannables". If you agree, please also rename the file.

stasm (approved):
I think it's good enough to use _placeholders_ here to mean _non-text pattern parts_.

aphillips:
@eemeli can you respond to the titling comment yay or nay so I can merge this?

exploration/open-close-placeholders.md
eemeli: I would prefer "placeholders".

PR #541, opened 2023-11-29
Accept open-close design doc

### LINKED ISSUE #240: Markup: can we use argument-less functions to represent standalone display elements?

### LINKED ISSUE #262: Do markup elements map to registered functions?

> About the markup syntax: Are these free-form, implementation-dependent tokens? Do you expect them to be HTML or TTS hints or accessibility hints? Without any distinctions, this seems like [Unicode private use code points](https://www.unicode.org/faq/private_use.html) with all of their problems.

> Or are the markup tokens registered functions? If so, then why not use the `{:function}` syntax, and maybe with a literal string before the function name (which could then be something like `:html`).

> If we do need markup with special syntax, then it should fit better into the overall syntax (e.g., always an ASCII symbol after the opening `{`) and into the framework as a whole (e.g., registered entities).

### LINKED ISSUE #375: What's left to discuss on markup?
As we merged #371, we didn't explicitly discuss if this allows us to close all or some of the following issues:
- #238 
- #241 
- #262 
- #356 
- #357

Rather than commenting on each of those, I thought it might be appropriate to reflect as a whole what parts of markup we ought to still discuss at this time, and then close or highlight those issues.

My own sense is that all of the above could and should be closed.

### LINKED ISSUE #424: Clarify `standalone` vs. `open` vs. `close` function/expression
See: https://github.com/unicode-org/message-format-wg/pull/419/files#r1262730738

In adding definitions for open and close expressions (and their friend standalone), we found that we are underspecifying them. This is a placeholder to solve that.

### LINKED ISSUE #537: [Discussion] {{Spannables}}
Per the 2023-11-27 teleconference, this issue is for discussing the design of spannables (also known as open/close/standalone markup).

The design document lives [here](https://github.com/unicode-org/message-format-wg/blob/main/exploration/open-close-placeholders.md) and should be used as a reference in this discussion.

eemeli:
This PR updates the design doc to include `{#standalone/}` markup as discussed on today's call, and should reflect the consensus reached during the meeting.

~This is a **draft** PR filed in support of the discussion in #537, and its eventual conclusion of the discussion on spannables/markup. We may want to amend this as an outcome of the discussion, but for now it's here to allow for the collection of the various historical issues that we'll be able to resolve with this.~

~You are not expected to review this PR until we've settled on the matter in #537.~

Closes #240 
Closes #262
Closes #375
Closes #424 
Closes #537

### DIFF
--- a/exploration/open-close-expressions.md
+++ b/exploration/open-close-expressions.md
@@ -1,6 +1,6 @@
 # Open/Close Expressions
 
-Status: **Proposed**
+Status: **Accepted**
 
 <details>
 	<summary>Metadata</summary>
--- a/exploration/open-close-placeholders.md
+++ b/exploration/open-close-placeholders.md
@@ -17,6 +17,7 @@ Status: **Accepted**
 		<dd><a href="https://github.com/unicode-org/message-format-wg/pull/517">#517</a></dd>
 		<dd><a href="https://github.com/unicode-org/message-format-wg/pull/535">#535</a></dd>
 		<dd><a href="https://github.com/unicode-org/message-format-wg/pull/540">#540</a></dd>
+		<dd><a href="https://github.com/unicode-org/message-format-wg/pull/541">#541</a></dd>
 	</dl>
 </details>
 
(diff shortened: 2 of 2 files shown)


aphillips:
(chair 🎩 on)

I presume that this is a proposal to accept the mooted design in the Monday call. If we choose something other than hash-and-slash, other changes would be merged here. Added `Agenda+` pending discussion.

mihnita:
I don't consider this closed unless we resolve the standalone.
We should not split things into tiny bits and approve them separately.

Decomposing complex problems helps up to a point, but below a certain level of granularity is bad.

aphillips (approved):
This is what we discussed in the 2023-12-18 call. I think the wording is a bit weighted but we're due to discuss again in the 2024-01-08 call and @stasm will contribute further as well. Let's commit this so we can fix the ABNF.

PR #544, opened 2023-11-29
Update affiliations

eemeli:
I happened to notice that we maintain a list of named "delegates" and their affiliations, and that it's at least somewhat out of date. I updated the Mozilla entries to start, as I joined a few years ago and @stasm is now at Google.

There are probably other out-of-date entries here as well, and as we rely on the mailing list + GitHub tracking for the places where "membership" of the group is relevant, we could also just remove this whole file as misleading.

This could also be fast-tracked.

### DIFF
--- a/delegates.md
+++ b/delegates.md
@@ -21,7 +21,7 @@ Please include your primary affiliation (e.g., the company you represent or wher
 
 ### Current
 
-- Addison Phillips - Amazon.com (APS)
+- Addison Phillips - Unicode (APP)
 - Dan Chiba - Oracle (DCA)
 - Eemeli Aro - Mozilla & OpenJSF (EAO)
 - Elango Cheran - Google (ECH)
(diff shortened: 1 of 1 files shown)


delegates.md
```diff
@@ -23,7 +23,7 @@ Please include your primary affiliation (e.g., the company you represent or wher
 
 - Addison Phillips - Amazon.com (APS)
```
aphillips, line 24:
```suggestion
- Addison Phillips - Unicode (APP)
```

aphillips:
I was considering killing this file, but wanted to ensure it didn't mess up anything first. It's not only out of date, but not useful for anything given how we actually operate membership etc.

aphillips (approved):

aphillips:
I'll update this now and deal with additional changes in a minute

PR #548, opened 2023-12-01
Update the rest of the spec to match the ABNF after adding `.keywords`

eemeli:
This applies the changes from #529 to the rest of the specification (syntax, data model, formatting).

The ABNF is also updated editorially to include the `placeholder` and `matcher` described in the syntax.

A new Unsupported Statement error is added.

The data model changes also account for the change from `let` to `local` and `input`, which was not previously done.

### DIFF
--- a/spec/data-model/README.md
+++ b/spec/data-model/README.md
@@ -25,8 +25,22 @@ Two equivalent definitions of the data model are also provided:
 A `SelectMessage` corresponds to a syntax message that includes _selectors_.
 A message without _selectors_ and with a single _pattern_ is represented by a `PatternMessage`.
 
+In the syntax,
+a `PatternMessage` may be represented either as a _simple message_ or as a _complex message_,
+depending on whether it has declarations and if its `pattern` is allowed in a _simple message_.
+
+While this specification currently only supports `"match"` selectors,
+the _matcher_ syntax may include a _reserved statement_,
+allowing for future expansion.
+In the data model, this is represented as a `SelectMessage` with a string `type`
+that is not `"input"`, `"local"`, `"match"`, or `"message"`
+(corresponding to a concatenation of the `UnsupportedStatement` `keyword` and `body`),
+and with `selectors` matching the _reserved statement_ _expressions_.
+As TypeScript does not support [negated types](https://github.com/microsoft/TypeScript/issues/4196#issuecomment-1601404332),
+that is left out from here but included in the JSON Schema and DTD definitions.
+
 ```ts
-type Message = PatternMessage | SelectMessage;
+type Message = PatternMessage | SelectMessage<"match">;
 
 interface PatternMessage {
   type: "message";
@@ -34,8 +48,8 @@ interface PatternMessage {
   pattern: Pattern;
 }
 
-interface SelectMessage {
-  type: "select";
+interface SelectMessage<T extends string> {
+  type: T;
   declarations: Declaration[];
   selectors: Expression[];
   variants: Variant[];
@@ -43,14 +57,46 @@ interface SelectMessage {
 ```
 
 Each message _declaration_ is represented by a `Declaration`,
-which connects the `name` of the _variable_
+which connects the `name` of a _variable_
 with its _expression_ `value`.
 The `name` does not include the initial `$` of the _variable_.
 
+The `name` of an `InputDeclaration` MUST be the same
+as the `name` in the `VariableRef` of its `VariableExpression` `value`.
+
+An `UnsupportedStatement` represents a statement not supported by the implementation.
+Its `keyword` is a non-empty string of a-z contents (i.e. not including the initial `.`).
+If not empty, the `body` is the "raw" value (i.e. escape sequences are not processed)
+starting after the keyword and up to the first _expression_,
+not including leading or trailing whitespace.
+The non-empty `expressions` correspond to the trailing _expressions_ of the _reserved statement_.
+
+> **Note**
+> Be aware that future versions of this specification
+> might assign meaning to _reserved statement_ values.
+> This would result in new interfaces being added to
+> this data model.
+
 ```ts
-interface Declaration {
+type Declaration = InputDeclaration | LocalDeclaration | UnsupportedStatement;
+
+interface InputDeclaration {
+  type: "input";
+  name: string;
+  value: VariableExpression;
+}
+
+interface LocalDeclaration {
+  type: "local";
   name: string;
   value: Expression;
+}
+
+interface UnsupportedStatement {
+  type: "unsupported-statement";
+  keyword: string;
+  body?: string;
+  expressions: Expression[];
 }
 ```
 
@@ -74,28 +120,35 @@ interface CatchallKey {
 ## Patterns
 
 Each `Pattern` represents a linear sequence, without selectors.
-Each element of the sequence MUST have either a `Text` or an `Expression` shape.
-`Text` represents literal _text_,
+Each element of the `body` array MUST either be a string or an `Expression` object.
+String values represent literal _text_,
 while `Expression` wraps each of the potential _expression_ shapes.
-The `value` of `Text` is the "cooked" value (i.e. escape sequences are processed).
+The `body` strings are the "cooked" _text_ values, i.e. escape sequences are processed.
 
-Implementations MUST NOT rely on the set of `Expression` `body` values being exhaustive,
+Implementations MUST NOT rely on the set of `Expression` interfaces being exhaustive,
 as future versions of this specification MAY define additional expressions.
-A `body` with an unrecognized value SHOULD be treated as an `Unsupported` value.
+An `Expression` `func` with an unrecognized value SHOULD be treated as an `UnsupportedExpression` value.
 
 ```ts
 interface Pattern {
-  body: Array<Text | Expression>;
+  body: Array<string | Expression>;
 }
 
-interface Text {
-  type: "text";
-  value: string;
+type Expression = LiteralExpression | VariableExpression | FunctionExpression;
+
+interface LiteralExpression {
+  arg: Literal;
+  func?: FunctionRef | UnsupportedExpression;
 }
 
-interface Expression {
-  type: "expression";
-  body: Literal | VariableRef | FunctionRef | Unsupported;
+interface VariableExpression {
+  arg: VariableRef;
+  func?: FunctionRef | UnsupportedExpression;
+}
+
+interface FunctionExpression {
+  arg?: never;
+  func: FunctionRef | UnsupportedExpression;
 }
 ```
 
@@ -148,8 +201,8 @@ interface Option {
 }
 ```
 
-An `Unsupported` represents an _expression_ with a
-_reserved_ _annotation_ or a _private-use_ _annotation_ not supported
+An `UnsupportedExpression` represents an _expression_ with a
+_reserved annotation_ or a _private-use annotation_ not supported
 by the implementation.
 The `sigil` corresponds to the starting sigil of the _annotation_.
 The `source` is the "raw" value (i.e. escape sequences are not processed)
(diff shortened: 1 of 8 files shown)


spec/data-model/README.md
```diff
@@ -25,33 +25,79 @@ Two equivalent definitions of the data model are also provided:
...
-which connects the `name` of the _variable_
+which connects the `name` of a _variable_
 with its _expression_ `value`.
 The `name` does not include the initial `$` of the _variable_.
 
+The `name` of an `InputDeclaration` MUST be the same
+as the `name` in the `VariableRef` of its `VariableExpression` `value`.
+
+An `UnsupportedStatement` represents a statement not supported by the implementation.
+Its `keyword` is a non-empty string of a-z contents (i.e. not including the initial `.`).
```
gibson042, line 68: Grammar nonterminal `name` is not limited to letters.
```suggestion
Its `keyword` is a non-empty string name (i.e. not including the initial `.`).
```

spec/data-model/README.md
```diff
@@ -74,28 +120,35 @@ interface CatchallKey {
 ## Patterns
 
 Each `Pattern` represents a linear sequence, without selectors.
-Each element of the sequence MUST have either a `Text` or an `Expression` shape.
-`Text` represents literal _text_,
+Each element of the `body` array MUST either be a string or an `Expression` object.
```
gibson042, line 123:
```suggestion
Each element of the `body` array MUST either be a non-empty string or an `Expression` object.
```
(assuming adherence to [ECMA-402 precedent](https://tc39.es/ecma402/#sec-partitionpattern))

spec/data-model/README.md
```diff
@@ -74,28 +120,35 @@ interface CatchallKey {
...
 
-interface Text {
-  type: "text";
-  value: string;
+type Expression = LiteralExpression | VariableExpression | FunctionExpression;
+
+interface LiteralExpression {
+  arg: Literal;
+  func?: FunctionRef | UnsupportedExpression;
+}
```
gibson042, line 142: For consistency with the data model implied by the grammar, `func` fields should probably be renamed (and possibly `FunctionExpression` as well, in both the grammar and here).
```suggestion
type Expression = LiteralExpression | VariableExpression | AnnotationExpression;

interface LiteralExpression {
  arg: Literal;
  annotation?: FunctionRef | UnsupportedAnnotation;
}
```
etc.

spec/syntax.md
```diff
@@ -111,30 +111,44 @@ A **_<dfn>message</dfn>_** is the complete template for a specific message forma
...
+```abnf
+simple-message = [simple-start pattern]
+simple-start = simple-start-char / text-escape / placeholder
+```
 
 A **_<dfn>complex message</dfn>_** is any _message_ that contains _declarations_,
 a _matcher_, or both.
-A _complex message_ always begins with the the sequence `{{`
-and is terminated by the sequence `}}`
+A _complex message_ always begins with a keyword that has a `.` prefix
```
gibson042, line 143:
```suggestion
A _complex message_ always begins with either a keyword that has a `.` prefix or a _quoted pattern_
```

spec/syntax.md
```diff
@@ -575,39 +624,40 @@ private-start = "&" / "^"
...
+A _reserved annotation_ does not include trailing whitespace.
 
 Implementations MUST NOT assign meaning or semantics to
 an _annotation_ starting with `reserved-start`:
 these are reserved for future standardization.
-Implementations MUST NOT remove or alter the contents of a _reserved_ _annotation_.
+Implementations MUST NOT remove or alter the contents of a _reserved annotation_.
 
 While a reserved sequence is technically "well-formed",
 unrecognized reserved sequences have no meaning and MAY result in errors during formatting.
```
gibson042, line 646:
```suggestion
unrecognized reserved sequences have no meaning and MUST result in errors during formatting.
```

spec/syntax.md
```diff
@@ -618,14 +668,13 @@ This section defines common elements used to construct _messages_.
 
 A **_<dfn>keyword</dfn>_** is a reserved token that has a unique meaning in the _message_ syntax.
 
-The following four keywords are reserved: `input`, `local`, `match`, and `when`.
-Reserved keywords are always lowercase.
+The following three keywords are reserved: `.input`, `.local`, and `.match`.
```
gibson042, line 671:
```suggestion
The following three keywords are defined: `.input`, `.local`, and `.match`.
```

spec/syntax.md
```diff
@@ -618,14 +668,13 @@ This section defines common elements used to construct _messages_.
...
+Reserved keywords are always lowercase and start with U+002E FULL STOP `.`.
```
gibson042, line 672: The reservations support future addition of non-lowercase keywords.
```suggestion
Keywords are always lowercase and start with U+002E FULL STOP `.`.
```

spec/data-model/README.md
```diff
@@ -25,33 +25,79 @@ Two equivalent definitions of the data model are also provided:
...
+a `PatternMessage` may be represented either as a _simple message_ or as a _complex message_,
+depending on whether it has declarations and if its `pattern` is allowed in a _simple message_.
+
+While this specification currently only supports `"match"` selectors,
+the _matcher_ syntax may include a _reserved statement_,
+allowing for future expansion.
+In the data model, this is represented as a `SelectMessage` with a string `type`
+that is not `"input"`, `"local"`, `"match"`, or `"message"`
+(corresponding to a concatenation of the `UnsupportedStatement` `keyword` and `body`),
+and with `selectors` matching the _reserved statement_ _expressions_.
```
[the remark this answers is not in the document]
eemeli: Hang on, let's rewind a bit. I think there's an issue in the current ABNF this is pointing at. Consider this message:
```
.foo {$var}
.foo {$var}
* {{text}}
```

I believe the _intent_ of our current ABNF would have the first `.foo {$var}` parse as a `declaration`, but to have the second one end up in the `matcher`. But that can only happen if we allow for infinite lookahead past the `1*([s] expression)` and for the ambiguity to be resolved by only one overall interpretation of the message providing a valid parse.

And I don't think we'd be happy with that.

I can imagine two different approaches to resolve this:

1. Rather than having something like `declarations`, have `statements` (including `.match`) at the start of a complex message. Then make having `match-statement` or `reserved-statement` be the last one (if present) be a requirement for it to be _valid_ rather than _well-formed_.
2. Make the selector syntax distinct from what's allowed in `declarations`, e.g. with `.match [{$foo}]`.

Option 1 also makes it much harder to allow for statements like `.strict true`, because we need to detect the end of the selectors.

Option 2 adds syntactical noise, and would still require lookup past the `reserved-statement` keyword to disambiguate the syntax.

And what do we get out of it? An ability to later add something like `.select` as an alternative to `.match`. Given that we've already delegated _a lot_ to custom selectors, and we're not attempting to remove the validity requirements of selector/variant-keys count matching or having an all-catch-all case, how could this `.select` differ from `.match`? The only remaining aspect that I can imagine is using first-match rather than best-match for the selection.

@aphillips @stasm Would you concur that the cost/benefit ratio here is not very good, and that we should drop `reserved-statement` from the `matcher`? I think first-match vs best-match is something that we could assess post-tech-preview and, if necessary, apply as a behaviour change on `.match` rather than via a later keyword introduction.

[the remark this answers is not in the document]
stasm: My thinking here was that building too much flexibility via the reserved syntax was powerful but also dangerous -- we may end up with a very different syntax than right now. Designating some keywords as invariant, like `.match`, seems like a good idea.

The first-match `.select` use-case gives me pause. Maybe it would be best to explicitly reserve `.select` as the 4th keyword?

The entire problem of forward compatibility and reserved syntax is complex enough that perhaps it would make sense to document it in a design doc? What sort of extensions do we want to make possible? Is there anything that we can rule out today? How do build all of that into the ABNF?

@gibson042 Would you be interested in writing such a design?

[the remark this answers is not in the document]
eemeli: @aphillips What do you think of the idea of including `.select` as a reserved alternative to `.match`, otherwise using the same syntax? In the ABNF it would probably look like this:
```abnf
matcher = match-statement 1*([s] variant)
match-statement = (match / select) 1*([s] selector)
match = %s".match"
select = %s".select"
```

As with other reserved syntax, this would generate an error during formatting, and not in the data model.

The intent here would be to allow for one future opportunity to iterate on `.match`, should we identify any issues with it after the 2.0 release. In terms of process steps, its [selection method](https://github.com/unicode-org/message-format-wg/blob/main/spec/formatting.md#pattern-selection) is by far the most complex part of the spec.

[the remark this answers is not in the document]
aphillips: I think I'd be against it, particularly if we also had a `:select` matcher. 

I know I brought this up originally, but I think having multiple selection mechanisms isn't helpful for users (if `.select` were a real keyword there would be a constant question of which matcher to use whether the results were "correct" for a given one). 

I agree that the best-match selection is the most complex part of the spec, but it is consistent (or should be) with what MF1 already does. We have a lot of experience with that. If we need to adjust the algorithm because of something we find in testing, I'm at peace with that. But I think it will not be helpful to have more kinds of selection.

Clarification of my earlier comment:
> The extension mechanism for .match could use one of the forest of reserved sigils rather than different keywords.

What I meant here was replacing functions, not a different sigil for the keyword, e.g. `.match {$var ^extended}`

So, in summary, I favor:

1. Remove reserved-statement from the selectors
2. Stick with the current matcher
3. Be done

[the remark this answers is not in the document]
eemeli: Let's drop it then.

spec/data-model/README.md
eemeli: One challenge with "annotation" is that when it's in an expression without an operand, what is it "annotating"?

eemeli:
Dropped the `matcher` `reserved-statement`, as discussed in https://github.com/unicode-org/message-format-wg/pull/548#discussion_r1412642672.

spec/syntax.md
aphillips: The normative language MAY not be appropriate here 😁. Also, "unrecognized reserved sequence" is an oxymoron. If it were recognized, it would not be reserved. 

Since we eliminated reserved from complex-body, a reserved sequence is not in the way of selecting the pattern. The question we have to ask ourselves (which is what @gibson042's comment is really about...) is: is a reserved sequence an error or default ignorable? If it is an error, it not only "MUST result in errors during formatting", it needs to result in a *specific* error (Unsupported Statement?). If it is not an error, it "MUST NOT result in an error during formatting". 

Which behavior do we prefer?

spec/data-model/README.md
aphillips: The expression.

spec/syntax.md
aphillips: We would not error on unselected patterns (per your quote). Any MUST here would only apply inside a pattern being formatted. Note that reserved-statement and reserved-annotation are different things. @gibson042's suggestion should be specific about which it means. I think the right choice here is actually to stop sooner. Note that the front of the sentence says: 
> While a reserved sequence is technically "well-formed",

So we should just finish up with:

```suggestion
unrecognized _reserved-annotations_ or _private-use-annotations_ have no meaning.
```

This is the syntax spec and the normative text is really a "forward reference" to error handling in formatting.md. We should just let formatting do the work. Or we could say that unrecognized placeholders turn into fallbacks during formatting.

spec/data-model/README.md
eemeli: For the cases where the implementation does support a specific private-use annotation, we have this: https://github.com/unicode-org/message-format-wg/blob/f35985614c66eb513954f342fcebc1532530ae02/spec/data-model/README.md?plain=1#L167-L171

And then in the [Extensions](https://github.com/unicode-org/message-format-wg/blob/main/spec/data-model/README.md#extensions) section, we specify that an implementation is allowed to extend the data model for that purpose. Or at least it's intended to do so.

So we _can_ at this point lump all of the unsupported reserved & private-use into this one basket of the data model, rather than needing to split it as in the syntax.

spec/syntax.md
gibson042: Works for me too.

spec/data-model/README.md
aphillips: If the data-model is meant to be interchanged, does that mean that the "private-use-ness" might be a relevant detail to the consumer? I mean, maybe it's `FunctionRef` but if it weren't, it wouldn't be `UnsupportedAnnotation` necessarily either? Or am I barking up the wrong tree? 🌳

eemeli: If an implementation does define and support a private-use annotation, I think there are two options for its data model representation:
1. It's mapped to some representation within the base data model, e.g. as some `func` value or maybe as an operand of some sort. But then why need the new syntax in the first place?
2. The annotation is doing something truly different, and it doesn't fit within the base data model. In this case, the implementation extends the base, e.g. adding a new field in the `Expression` interfaces.

If going the first route, then interchange is of course easier, but that syntax will also have some representation without private use. If going the second route, interchange will require both ends to agree about the private-use.

PR #553, opened 2023-12-04
Unquoted literals as name or number

### LINKED ISSUE #518: Forbid unquoted non-numeric literals as expression operands
I propose that we restrict the literal syntax and disallow unquoted non-numeric literals as expression operands.

Today, literals (quoted and unquoted) can appear in 3 positions:

* as variant keys,
* as option values,
* as expression operands.

The first two positions are common and well justified. OTOH, I posit that the third position is very rare, except numeric values. In fact, I'm not aware of strong use-cases for unquoted non-numeric literals as expression operands. Furthermore, I'm concerned about the risk of confusing such unquoted literals for keywords.

---

Forbidding unquoted non-numeric literals as expression operands will provide two benefits:

* Since our syntax uses keywords positioned at the front of a declaration, unquoted literals as expression operands may be confused for novel keywords. Consider:

      {now :datetime} vs. {|now| :datetime}
      {error :msgref} vs. {|error| :msgref}

    `|now|` and `|error|` are clearly recognizable as literal *values*.

* It would open new opportunities for introducing new placeholder or expression syntax: `{foo}`. See #516 for an example use-case.

As a reminder, `{5 :number}` would still be valid under this proposal, as well as `:func opt=unquoted` and `when unquoted`.

---

ABNF-wise, I think we'd need the following two changes:

* Split `unquoted` into `unquoted-string` and `unquoted-numeric`.
* Change `literal-expression` to `"{" [s] (quoted / unquoted-numeric) [s annotation] [s] "}"`

aphillips:
Implements the changes discussed in the 2023-12-04 teleconference.

This makes an unquoted literal be a name or a number.

This PR does **NOT** include changes to `+` and `-` sigils in the `function` production pending discussion/implementation of hash-and-slash for spannables. Merging this PR without those changes will make our syntax inconsistent/ambiguous.

Fixes #518 #519

### DIFF
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -52,7 +52,7 @@ quoted-char = %x0-5B         ; omit \
 
 unquoted       = name
                / number-literal
-number-literal = ["-"] (0 / ([1-9]) *DIGIT [ "." 1*DIGIT] [ (%i"e") ("+"/"-") 1*DIGIT]
+number-literal = ["-"] (0 / ([1-9] *DIGIT) [ "." 1*DIGIT] [ (%i"e") ["+" / "-"] 1*DIGIT]
 
 ; Reserve additional .keywords for use by future versions of this specification.
 reserved-statement = reserved-keyword [s reserved-body] 1*([s] expression)
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -50,9 +50,9 @@ quoted-char = %x0-5B         ; omit \
             / %x7D-D7FF      ; omit surrogates
             / %xE000-10FFFF
 
-unquoted       = unquoted-start *(name-char / ":")
-unquoted-start = name-start / DIGIT / "."
-               / %xB7 / %x300-36F / %x203F-2040
+unquoted       = name
+               / number-literal
+number-literal = ["-"] (0 / ([1-9]) *DIGIT [ "." 1*DIGIT] [ (%i"e") ("+"/"-") 1*DIGIT]
 
 ; Reserve additional .keywords for use by future versions of this specification.
 reserved-statement = reserved-keyword [s reserved-body] 1*([s] expression)
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -52,7 +52,9 @@ quoted-char = %x0-5B         ; omit \
 
 unquoted       = name
                / number-literal
-number-literal = ["-"] (0 / ([1-9] *DIGIT) [ "." 1*DIGIT] [ (%i"e") ["+" / "-"] 1*DIGIT]
+; number-literal matches JSON number
+; https://www.rfc-editor.org/rfc/rfc8259#section-6
+number-literal = ["-"] (0 / ([1-9] *DIGIT)) ["." 1*DIGIT] [%i"e" ["-" / "+"] 1*DIGIT]
 
 ; Reserve additional .keywords for use by future versions of this specification.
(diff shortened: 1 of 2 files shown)


spec/message.abnf
```diff
@@ -50,9 +50,9 @@ quoted-char = %x0-5B         ; omit \
             / %x7D-D7FF      ; omit surrogates
             / %xE000-10FFFF
 
-unquoted       = unquoted-start *(name-char / ":")
-unquoted-start = name-start / DIGIT / "."
-               / %xB7 / %x300-36F / %x203F-2040
+unquoted       = name
+               / number-literal
+number-literal = ["-"] (0 / ([1-9]) *DIGIT [ "." 1*DIGIT] [ (%i"e") ("+"/"-") 1*DIGIT]
```
eemeli, line 55: A few fixes to make this actually match JSON:

```suggestion
number-literal = ["-"] (0 / ([1-9] *DIGIT) [ "." 1*DIGIT] [ (%i"e") ["+" / "-"] 1*DIGIT]
```

spec/syntax.md
```diff
@@ -713,9 +712,9 @@ quoted-char    = %x0-5B         ; omit \
                / %x7D-D7FF      ; omit surrogates
                / %xE000-10FFFF
 
-unquoted       = unquoted-start *name-char
-unquoted-start = name-start / DIGIT / "."
-               / %xB7 / %x300-36F / %x203F-2040
+unquoted       = name
+               / number-literal
+number-literal = ["-"] (0 / ([1-9]) *DIGIT [ "." 1*DIGIT] [ (%i"e") ("+"/"-") 1*DIGIT]
```
eemeli, line 717:
```suggestion
number-literal = ["-"] (0 / ([1-9] *DIGIT) [ "." 1*DIGIT] [ (%i"e") ["+" / "-"] 1*DIGIT]
```

spec/syntax.md
```diff
@@ -732,10 +731,11 @@ _Function_ _identifiers_ are prefixed with `:`, `+`, or `-`.
 _Option_ _identifiers_ have no prefix.
 
 A **_<dfn>name</dfn>_** is a character sequence used in an _identifier_ 
-or as the name for for a _variable_.
+or as the name for for a _variable_
+or the value of an _unquoted_ _literal_.
 
 _Variable_ names are prefixed with `$`.
-for a _function_ (prefixed with `:`, `+` or `-`),
+for a _function_ (prefixed with `:`, `+` or `-`).
```
eemeli, line 738: This line is missing the rest of itself. What's this trying to say?

eemeli (changes requested):
Looks good overall, but see line comments.

spec/syntax.md
aphillips: It was missing when I got here. Fixed this to be consistent with the previous sentence.

spec/syntax.md
```diff
@@ -732,10 +731,11 @@ _Function_ _identifiers_ are prefixed with `:`, `+`, or `-`.
 _Option_ _identifiers_ have no prefix.
 
 A **_<dfn>name</dfn>_** is a character sequence used in an _identifier_ 
-or as the name for for a _variable_.
+or as the name for for a _variable_
```
gibson042, line 734:
```suggestion
or as the name for a _variable_
```

gibson042 (approved):

PR #554, opened 2023-12-05
data-model: Better represent annotations in the data model, particularly when unsupported

### LINKED ISSUE #552: Create a PR for renaming `FunctionExpression`
This is an action item from the 2023-12-04 teleconference.

---

Discussion thread:

See this [comment thread](https://github.com/unicode-org/message-format-wg/pull/548#discussion_r1413182404) for context. 

EAO: We currently represent all annotations as “func”, including those that are private-use and reserved. A _supported_ private-use annotation should show up in data model extension, but a reserved annotation or _unsupported_ private-use annotation by definition could not. The names are not especially visible, but they do appear.

APP: We have two kinds of reserved things—statements and annotations. We’re talking here about annotations, which appear in expressions with sigils other than `:`. EAO is proposing renaming interfaces in the data model to correspond with the ABNF.

EAO: Clarification, this is RGN’s suggestions.

APP: I suggested later adding private-use annotations separately.

STA: What is reserved again?

APP: Anything we’ve reserved. As opposed to private-use, which is open for implementation innovation.

STA: An implementation that advertises support for private-use would need to expose such information in the data model, but error out for unsupported private-use or reserved annotations.

APP: Back on topic, do we want to rename FunctionExpression to AnnotationExpression?

STA: This may be related to markup, but I don’t object to the change now.

EAO: How do change advocates imagine an implementation supporting private-use to represent instances thereof? What we currently expect is extension of the data model, but if you don’t like that then what shape would you imagine?

APP: If we renamed as suggested, that resolves some of the challenge. That is less true with the FunctionExpression name, because a private-use annotation may not be function-like.

STA: To answer EAO, it should be possible to use a discriminated union type. I like renaming to AnnotationExpression. Is there a reason why that would not be a good idea?

EAO: I think the question is whether supported private use goes in what is now called FunctionExpression, or whether it belongs one level up.

STA: I have strong feelings in favor of a discriminated union, one way or the other.

APP: RGN, want to PR this?

RGN: Sure, I can do a PR.

EAO: What you’re describing might already be there, but please take a look.

STA: Is there a specific use case to help me understand?

EAO: An inline comment.

STA: Do we have a design doc on reservations and private use?

APP: We’re sure spending a lot of time on things no one can use. 🙂

gibson042:
More accurately captures annotations in the data model, particularly when they are unsupported, and also just generally increases alignment between data model representations and the ABNF.

Fixes #552

### DIFF
--- a/spec/data-model/message.json
+++ b/spec/data-model/message.json
@@ -22,8 +22,15 @@
       },
       "required": ["type", "name"]
     },
-    "value": {
-      "oneOf": [{ "$ref": "#/$defs/literal" }, { "$ref": "#/$defs/variable" }]
+    "option": {
+      "type": "object",
+      "properties": {
+        "name": { "type": "string" },
+        "value": {
+          "oneOf": [{ "$ref": "#/$defs/literal" }, { "$ref": "#/$defs/variable" }]
+        }
+      },
+      "required": ["name", "value"]
     },
 
     "function": {
@@ -32,16 +39,12 @@
         "type": { "const": "function" },
         "kind": { "enum": ["open", "close", "value"] },
         "name": { "type": "string" },
+        "operand": {
+          "oneOf": [{ "$ref": "#/$defs/literal" }, { "$ref": "#/$defs/variable" }]
+        },
         "options": {
           "type": "array",
-          "items": {
-            "type": "object",
-            "properties": {
-              "name": { "type": "string" },
-              "value": { "$ref": "#/$defs/value" }
-            },
-            "required": ["name", "value"]
-          }
+          "items": { "$ref": "#/$defs/option" }
         }
       },
       "required": ["type", "kind", "name"]
@@ -51,24 +54,26 @@
       "properties": {
         "type": { "const": "unsupported-expression" },
         "sigil": {
-          "enum": ["!", "@", "#", "%", "^", "&", "*", "<", ">", "?", "~"]
+          "enum": ["!", "@", "#", "%", "^", "&", "*", "<", ">", "/", "?", "~"]
         },
-        "source": { "type": "string" }
+        "source": { "type": "string" },
+        "operand": {
+          "oneOf": [{ "$ref": "#/$defs/literal" }, { "$ref": "#/$defs/variable" }]
+        }
       },
       "required": ["type", "sigil", "source"]
-    },
-    "annotation": {
-      "oneOf": [
-        { "$ref": "#/$defs/function" },
-        { "$ref": "#/$defs/unsupported-expression" }
-      ]
     },
 
     "literal-expression": {
       "type": "object",
       "properties": {
         "arg": { "$ref": "#/$defs/literal" },
-        "func": { "$ref": "#/$defs/annotation" }
+        "func": {
+          "oneOf": [
+            { "$ref": "#/$defs/function" },
+            { "$ref": "#/$defs/unsupported-expression" }
+          ]
+        }
       },
       "required": ["arg"]
     },
@@ -76,14 +81,24 @@
       "type": "object",
       "properties": {
         "arg": { "$ref": "#/$defs/variable" },
-        "func": { "$ref": "#/$defs/annotation" }
+        "func": {
+          "oneOf": [
+            { "$ref": "#/$defs/function" },
+            { "$ref": "#/$defs/unsupported-expression" }
+          ]
+        }
       },
       "required": ["arg"]
     },
     "function-expression": {
       "type": "object",
       "properties": {
-        "func": { "$ref": "#/$defs/annotation" }
+        "func": {
+          "oneOf": [
+            { "$ref": "#/$defs/function" },
+            { "$ref": "#/$defs/unsupported-expression" }
+          ]
+        }
       },
       "required": ["func"]
     },
--- a/spec/data-model/message.json
+++ b/spec/data-model/message.json
@@ -33,10 +33,10 @@
       "required": ["name", "value"]
     },
 
-    "function": {
+    "function-annotation": {
       "type": "object",
       "properties": {
-        "type": { "const": "function" },
+        "type": { "const": "function-annotation" },
         "kind": { "enum": ["open", "close", "value"] },
         "name": { "type": "string" },
         "operand": {
@@ -49,10 +49,10 @@
       },
       "required": ["type", "kind", "name"]
     },
-    "unsupported-expression": {
+    "unsupported-annotation": {
       "type": "object",
       "properties": {
-        "type": { "const": "unsupported-expression" },
+        "type": { "const": "unsupported-annotation" },
         "sigil": {
           "enum": ["!", "@", "#", "%", "^", "&", "*", "<", ">", "/", "?", "~"]
         },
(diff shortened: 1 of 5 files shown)


spec/message.abnf
```diff
@@ -20,12 +20,14 @@ variant = key *(s key) [s] quoted-pattern
...
 variable-expression = "{" [s] variable [s annotation] [s] "}"
-function-expression = "{" [s] annotation [s] "}"
-annotation = (function *(s option))
-           / reserved-annotation
-           / private-use-annotation
+function-expression = "{" [s] function-annotation [s] "}"
+private-use-expression = "{" [s] private-use-annotation [s] "}"
+reserved-expression = "{" [s] reserved-annotation [s] "}"
+annotation = function-annotation / private-use-annotation / reserved-annotation
+function-annotation = function *(s option)
```
aphillips, line 30: Where did this come from?

I like the existing tight definition of `annotation` and am not very thrilled by lots of different expression types. The syntax doesn't need the proliferation of expressions. The data model (above) captures the deeper split between functions, private-use (maybe supported), and reserved (unsupported) and that is an outcome of parsing the syntax. But we don't, I guess, need to go all-in with every kind of placeholder. THoughts?

spec/syntax.md
```diff
@@ -431,13 +431,21 @@ optionally followed by an _annotation_.
 A **_<dfn>variable-expression</dfn>_** contains a _variable_,
 optionally followed by an _annotation_.
 
-A **_<dfn>function-expression</dfn>_** contains only an _annotation_.
+A **_<dfn>function-expression</dfn>_** contains only a _function_ _annotation_.
+
+A **_<dfn>private-use-expression</dfn>_** contains only a _private-use annotation_.
+
+A **_<dfn>reserved-expression</dfn>_** contains only a _reserved annotation_.
```
aphillips, line 438: See above. I think I'd rather have it be:

```suggestion
A **_<dfn>function-expression</dfn>_** contains an _annotation_ without an _operand_.
The _annotation_ can be a _function_ _annotation_, 
a _private-use annotation_,
or a _reserved annotation_.
```

aphillips (changes requested):
Not sure I agree with the ABNF changes. The data model stuff looks good tho.

spec/data-model/message.json
```diff
@@ -22,76 +22,92 @@
...
-      "oneOf": [
-        { "$ref": "#/$defs/function" },
-        { "$ref": "#/$defs/unsupported-expression" }
-      ]
-    },
 
     "literal-expression": {
       "type": "object",
       "properties": {
+        "type": { "const": "literal-expression" },
```
eemeli, line 64: I think adding a `type` at this level would be unnecessary and noisy. It's effectively duplicating information that's already available in the model as
```js
arg ? arg.type : annotation.type
```

spec/message.abnf
gibson042: We don't _need_ to make this change, but if we don't then the less direct translation from ABNF to data model can be unnecessarily confusing. Is that worth having fewer productions in the grammar?
```abnf
expression = literal-expression / variable-expression / nullary-expression
literal-expression = "{" [s] literal [s annotation] [s] "}"
variable-expression = "{" [s] variable [s annotation] [s] "}"
nullary-expression = "{" [s] annotation [s] "}"
annotation = (function *(s option))
           / reserved-annotation
           / private-use-annotation
```

spec/data-model/message.json
gibson042: There are _many_ levels where `type` is unnecessary (e.g., SelectMessage has `selectors` and `variants` while PatternMessage has `pattern`; VariableRef has `name` while Literal has `value`), and in fact the sole exception that I see (InputDeclaration has `value: VariableExpression` rather than `value: VariableRef`) appears to be a bug. So what guidelines should be used? What I have done here is include `type` in the constituents of any union—e.g., `Expression = LiteralExpression | VariableExpression | …`, and all the subtypes can be trivially discriminated by `type`.

> ```js
> arg ? arg.type : annotation.type
> ```

I want to avoid the need for consumers to invent such cleverness, especially since it is not robust in the face of adding private-use expressions.

spec/message.abnf
aphillips: The (normative!) grammar is **_way_** more important to me than the (informative) data-model. It's important that the grammar be easily understood and as conceptually clear as possible (using Einstein's simplicity maxim--_as simple as possible but no simpler_). 

Reserved annotations don't really matter--no one can use them. The private use ones are where any additional heat comes into play. The nature of how we set up private-use is that, while they might be "function-like" they don't have to be.

I don't think `nullary-expression` is as clean as `function-expression` or maybe `annotation-expression`?

spec/data-model/message.json
eemeli: The way I'm looking at this, an expression is a container, i.e. the `{…}` part of the syntax where the `…` might contain an operand, and it might contain an annotation, but in any case the surrounding `{` and `}` stay the same. So the data model should not change or change as little as possible throughout such changes as well.

As it does make sense to encode in the data model the "expression must not be empty" bit, we do need at least two different shapes for the expression contents, i.e. `{ arg: …, func?: … } | { func: … }`. But why go further than that?

I do get your argument about private-use, but if we're renaming its slot in the expression from `func` to `annotation`, then I don't see how it needs to impact the expression-level syntax. OTOH, if we were to continue with a dedicated slot for functions, then I could see something like this making sense:

```ts
type Expression = ValueExpression | FunctionExpression | UnsupportedExpression;

interface ValueExpression {
  type: "expression";
  arg: Literal | VariableRef;
  func?: FunctionRef;
}

interface FunctionExpression {
  type: "expression";
  arg?: null;
  func: FunctionRef;
}

interface UnsupportedExpression {
  type: "unsupported";
  arg?: Literal | VariableRef | null;
  annotation: UnsupportedAnnotation;
}
```
with `UnsupportedAnnotation` as you currently have it.

One previously unstated optimization target I have here is the developer experience of working with MF2 data models, which shows up here in two different ways:
1. If I need to write a script that'll e.g. add a function to all unannotated placeholders matching some variable name, needing to change the expression type in addition to the `func` values feels pretty clumsy.
2. Needing to work with string identifiers that are over 20 characters long is just asking for typos.

spec/message.abnf
```diff
@@ -19,15 +19,13 @@ selector = expression
...
 variable-expression = "{" [s] variable [s annotation] [s] "}"
-function-expression = "{" [s] function-annotation [s] "}"
-private-use-expression = "{" [s] private-use-annotation [s] "}"
-reserved-expression = "{" [s] reserved-annotation [s] "}"
-annotation = function-annotation / private-use-annotation / reserved-annotation
-function-annotation = function *(s option)
+annotation-expression = "{" [s] annotation [s] "}"
+annotation = (function *(s option))
+           / reserved-annotation
+           / private-use-annotation
```
aphillips: super-minor editorial nit... I would probably flop the order of these, since private-use is functional vs. reserved being, well, reserved:

```suggestion
           / private-use-annotation
           / reserved-annotation
```

aphillips (approved):
... or ship it...

```
~~~~𖠳~~~~
```

spec/message.abnf
gibson042: I agree, and had thought about doing this anyway. Thanks for the push.

aphillips:
@eemeli If you approve, I can merge

eemeli (approved):

PR #555, opened 2023-12-05
Add normative language about expression nesting

### LINKED ISSUE #353: Document: "no nesting function calls"
Add normative language and example of no nesting to spec

See conversation in [#347](https://github.com/unicode-org/message-format-wg/pull/347#pullrequestreview-1302494429).

aphillips:
Fixes #353 

We previously agreed not to document not allowing function/expression nesting. This is a minor fix to add that. The ABNF already doesn't permit nesting, so it's really just documentation cleanup. 

In addition, I added examples of variables as options in two places as an oversight in our set of examples.

### DIFF
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -424,6 +424,7 @@ during the _message_'s formatting.
 An _expression_ MUST begin with U+007B LEFT CURLY BRACKET `{`
 and end with U+007D RIGHT CURLY BRACKET `}`.
 An _expression_ MUST NOT be empty.
+An _expression_ cannot contain another _expression_.
 
 A **_<dfn>literal-expression</dfn>_** contains a _literal_,
 optionally followed by an _annotation_.
@@ -470,6 +471,7 @@ Additionally, an _input-declaration_ can contain a _variable-expression_.
 > This placeholder contains an {|expression with a literal|}
 > This placeholder references a {$variable}
 > This placeholder references a function on a variable: {$variable :function with=options}
+> This placeholder has an option with a variable: {:function option=$variable}
 > ```
 
 ### Annotation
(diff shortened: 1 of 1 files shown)


eemeli (approved):

spec/syntax.md
```diff
@@ -470,6 +471,7 @@ Additionally, an _input-declaration_ can contain a _variable-expression_.
 > This placeholder contains an {|expression with a literal|}
 > This placeholder references a {$variable}
 > This placeholder references a function on a variable: {$variable :function with=options}
+> This placeholder has an option with a variable: {:function option=$variable}
```
gibson042, line 474: I think the first two examples here are overly cute.
```suggestion
> This placeholder contains a literal expression: {|literal|}
> This placeholder contains a variable expression: {$variable}
> This placeholder references a function on a variable: {$variable :function with=options}
> This placeholder contains a function expression with a variable-valued option: {:function option=$variable}
```

spec/syntax.md
```diff
@@ -578,6 +580,12 @@ option = identifier [s] "=" [s] (literal / variable)
 > Hello, {$userObj :person firstName=long}!
 > ```
 
+> A _message_ with a _variable_ as an option:
```
gibson042, line 583:
```suggestion
> A _message_ using the `:list` _function_ with a _variable_-valued `maxEntries` option:
```

gibson042 (approved):

spec/syntax.md
aphillips: Fixed with slightly different wording

spec/syntax.md
aphillips: ... but I _like_ cute examples. 🐱  🐈 😺 

Oh, all right... no fun in the specification...

PR #557, opened 2023-12-08
(Design) Exact Match Selector Options

aphillips:
In response to #433 we need to choose a name for the "exact match" selector.

This isn't done yet but will serve as a foundation for choosing.

### DIFF
--- a/exploration/exact-match-selector-options.md
+++ b/exploration/exact-match-selector-options.md
@@ -98,6 +98,11 @@ The exact match selector matches the string literal value of its operand against
 any sets of keys in a case-sensitive, encoding-sensitive manner.
 No Unicode normalization is performed.
 
+> [!NOTE]
+> In this context "encoding-sensitive" does not mean "character encoding"
+> but rather "sensitive to the sequence of code points used by the string".
+> See [Character Model for the World Wide Web: String Matching](https://www.w3.org/TR/charmod-norm)
+
 The `:string` selector is also a verbatim formatting function.
 That is, the message "{$var :string}" when formatted contains the resolved string
 value of the variable `$var`.
--- a/exploration/exact-match-selector-options.md
+++ b/exploration/exact-match-selector-options.md
@@ -99,7 +99,9 @@ _What other properties do they have?_
 ### Option A. `:select`
 
 Pros:
-- The same as ICU MF1's keyword. We have implementation experience.
+- The same as ICU MF1's keyword. Users have implementation experience.
+  Might fascilitate easier migration from MF1?
+- Not overly specific, so can conceptually fit with "non-exact" matching.
 
 Cons:
 - Somewhat generic name.
@@ -110,7 +112,9 @@ Pros:
 - Says what the match does.
 
 Cons:
-- Not super specific?
+- Unclear whether it refers to the serialization form or the imputed value.
+  There may be some cases where the match depends on options,
+  such as with `{|42.3| :integer}` and a key `42`
 
 ### Option C. `:equals`
 
@@ -129,7 +133,10 @@ Pros:
   it's the string value?
 
 Cons:
-- Not clear about 
+- Might confuse users who are comparing e.g. two numbers or two dates.
+  In typed languages, the values being compared might not actually be strings.
+- Depends on the serialization.
+
 
 ### Option E. Something else
 
--- a/exploration/exact-match-selector-options.md
+++ b/exploration/exact-match-selector-options.md
@@ -27,10 +27,12 @@ and issues raised in various design documents, notably
 <a href="https://github.com/unicode-org/message-format-wg/pull/471">#471</a>
 (about number selection).
 
-The exact match selector function is the function used when an implementation
-cannot find another selector.
-It is also an attribute of selectors, such as `:number` or perhaps `:date`
-that need to match specific values.
+The exact match selector function is an "all-purpose" built-in selector
+for use in messages when a more specific selector (such as for numbers or
+dates) is not available.
+Other selectors, such as `:number` or `:datetime`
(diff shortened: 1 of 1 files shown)


eemeli:
I'm not sure how to best express this in the design doc, but please feel to add this as appropriate: One yardstick that I think is rather relevant here is what happens when the value we're selecting on is not a string, but e.g. a number.

Put another way, substituting each of the options for `:xxx`, which pattern gets selected here, and what would a reader expect based on the function name?

```
.local $num = {1 :number}
.match {$local :xxx}
1.0 {{Very one}}
1   {{Integral one}}
one {{Category one}}
*   {{Not one}}
```

With `:select`, `:exact` and `:equals`, no hint is given that anything like a type cast might be happening here, so a reader could very easily presume that the comparison is numeric, and that therefore `Very one` would get selected, as it's before the `Integral one`.

My preference here would be for `:string`, which more clearly communicates that its input value is stringified and then compared, and that the `Integral one` pattern is therefore selected.

aphillips:
In the 2023-12-11 teleconference we decided to use `:string`. I have updated this PR to reflect that choice.

exploration/exact-match-selector-options.md
```diff
@@ -0,0 +1,161 @@
...
+		<dt>First proposed</dt>
+		<dd>2023-12-08</dd>
+	</dl>
+</details>
+
+## Objective
+
+_What is this proposal trying to achieve?_
+
+We need to choose a name for the selector function used when no other selector is specified.
```
eemeli, line 19:
```suggestion
We need to choose a name for the selector function used to match keys according to their string value.
```

exploration/exact-match-selector-options.md
```diff
@@ -0,0 +1,161 @@
...
+
+This addresses issue <a href="https://github.com/unicode-org/message-format-wg/issues/433">#433</a>
+and issues raised in various design documents, notably
+<a href="https://github.com/unicode-org/message-format-wg/pull/471">#471</a>
+(about number selection).
+
+The exact match selector function is the function used when an implementation
+cannot find another selector.
+It is also an attribute of selectors, such as `:number` or perhaps `:date`
+that need to match specific values.
```
eemeli, line 33:
```suggestion
The exact match selector function is the simplest selection function,
which compares its input to the variant keys using string equality.

Other selectors, such as `:number` or perhaps `:date`
may provide similar functionality for their supported value types.
```

exploration/exact-match-selector-options.md
```diff
@@ -0,0 +1,161 @@
...
+
+> [!NOTE]
+> In this context "encoding-sensitive" does not mean "character encoding"
+> but rather "sensitive to the sequence of code points used by the string".
+> See [Character Model for the World Wide Web: String Matching](https://www.w3.org/TR/charmod-norm)
+
+The `:string` selector is also a verbatim formatting function.
+That is, the message "{$var :string}" when formatted contains the resolved string
+value of the variable `$var`.
+
```
eemeli, line 109:
```suggestion

Other selectors may follow the example of `:number`,
which includes an option `select` that may take a value `exact`.

```

eemeli (changes requested):
Overall looks good, but see inline comments.

exploration/exact-match-selector-options.md
aphillips: This is not quite right.

```suggestion
We need to choose a name for the selector function used to match keys to the string value of an operand.
```

exploration/exact-match-selector-options.md
aphillips: Minor tweaks to this suggestion.

```suggestion
The exact match selector function is the simplest selector function,
which compares its operand to the variant keys using string equality.

Other selectors, such as `:number` and `:date` usually
provide similar functionality for their supported value types.
```

Note that both your (@eemeli) suggestion and mine lose the "defaultness" of this selector. Do you disagree with saying that this is the selector used when no other selector's signature matches?

eemeli: Yeah. It looks like we accidentally dropped a line in #441 that was added in #431: https://github.com/unicode-org/message-format-wg/blob/812c6dde8e8834341aae42198256f76e399f9a8c/spec/syntax.md?plain=1#L313-L314

In other words, we've resolved not to have a fallback or default selector. Instead, one must always be defined. I'll add a PR re-introducing that line.

eemeli: Otherwise, your suggestion is good, and with that I'm happy to approve this.

eemeli (approved):

aphillips:
@eemeli I fixed the text, so check it again? You were right that we lost the normative text about annotations--and I wrote this not thinking of that.

eemeli (approved):

mihnita (approved):

exploration/exact-match-selector-options.md
```diff
@@ -0,0 +1,166 @@
...
+ICU MF1 has a selector called `com.ibm.icu.text.SelectFormat` 
+associated with the keyword `select` in the MF1 syntax. 
+Our design should be compatible with this selector when used for the same use case,
+but it not constrained to use the same name or terminology.
+
+## Proposed Design
+
+_Describe the proposed solution. Consider syntax, formatting, errors, registry, tooling, interchange._
+
+The exact match selector is named `:string`.
```
markusicu, line 98: For selection, I think this is confusing. The alternative below of `exact` seems much clearer. I could also live with `equals`.

aphillips:
In the 2023-12-18 call there was a consensus to merge this as "proposed" for further consideration in the 2024-01-08 call. If you have additional proposals, please make a new PR.

PR #559, opened 2023-12-11
Update status of Delimiting of Patterns in Complex Messages design doc

echeran:
Updating doc after balloting finished with a conclusion.


aphillips:
Per 2023-12-11 call, accepted

PR #560, opened 2023-12-12
Add a `<matchSignature>` for `:number`, together with `:ordinal` & `:plural` aliases

eemeli:
This implements the [number selection design](https://github.com/unicode-org/message-format-wg/blob/main/exploration/number-selection.md).

Separately from this, we should add a "default functions" section to `registry.md`.

The only `<match>` included in this proposal is the fallback one; further changes could follow if/once #558 is accepted.

### DIFF
--- a/spec/registry.xml
+++ b/spec/registry.xml
@@ -141,7 +141,7 @@
   <function name="number">
     <!-- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat -->
     <description>
-      Locale-sensitive number formatting and selection
+      Number selection and formatting
     </description>
 
     <formatSignature>
@@ -249,12 +249,13 @@
         <description>
           The number selection type.
 
-          With `exact` selection, only a variant with a key matching
-          the JSON representation of the input number (without an exponent) will be selected.
-          With `plural` and `ordinal` selection, if no such exactly matching key is present,
-          a variant with the matching CLDR plural category (`zero`/`one`/`two`/`few`/`many`/`other`)
-          for the corresponding selection type will be selected,
-          according to the rules of the current locale
+          One of three selection models for numeric values, defaulting to "plural".
+
+          Each selector type matches numeric values to keys using the "number-literal" 
+          production from the ABNF.
+          The "plural" and "ordinal" selectors also use plural or ordinal rules from CLDR
+          to match categories "zero", "one", "two", "few", "many" and "other" with a lower
+          match quality than exact numeric match.
         </description>
       </option>
       <option name="minimumIntegerDigits" values="positiveInteger" default="1">
@@ -295,7 +296,7 @@
     </matchSignature>
 
     <alias name="integer">
-      <description>Locale-sensitive integral number formatting and selection</description>
+      <description>Integer selection and formatting</description>
       <setOption name="maximumFractionDigits" value="0" />
       <setOption name="style" value="decimal" />
     </alias>
--- a/spec/registry.xml
+++ b/spec/registry.xml
@@ -141,7 +141,7 @@
   <function name="number">
     <!-- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat -->
     <description>
-      Locale-sensitive number formatting
+      Locale-sensitive number formatting and selection
     </description>
 
     <formatSignature>
@@ -243,10 +243,71 @@
       </option>
     </formatSignature>
 
+    <matchSignature>
+      <input validationRule="anyNumber"/>
+      <option name="select" values="plural ordinal exact" default="plural">
+        <description>
+          The number selection type.
+
+          With `exact` selection, only a variant with a key matching
+          the JSON representation of the input number (without an exponent) will be selected.
+          With `plural` and `ordinal` selection, if no such exactly matching key is present,
+          a variant with the matching CLDR plural category (`zero`/`one`/`two`/`few`/`many`/`other`)
+          for the corresponding selection type will be selected,
+          according to the rules of the current locale
+        </description>
+      </option>
+      <option name="minimumIntegerDigits" values="positiveInteger" default="1">
+        <description>
+          The minimum number of integer digits to use.
+          A value with a smaller number of integer digits than this number will be
+          left-padded with zeros (to the specified length) when formatted.
+        </description>
+      </option>
+      <option name="minimumFractionDigits" values="positiveInteger">
+        <description>
+          The minimum number of fraction digits to use.
+          The default for plain number and percent formatting is 0;
+          the default for currency formatting is the number of minor unit digits provided by
+          the ISO 4217 currency code list (2 if the list doesn't provide that information).
+        </description>
+      </option>
+      <option name="maximumFractionDigits" values="positiveInteger">
+        <description>
+          The maximum number of fraction digits to use.
+          The default for plain number formatting is the larger of minimumFractionDigits and 3;
+          the default for currency formatting is the larger of minimumFractionDigits and the number of minor
+          unit digits provided by the ISO 4217 currency code list (2 if the list doesn't provide that information);
+          the default for percent formatting is the larger of minimumFractionDigits and 0.
+        </description>
+      </option>
+      <option name="minimumSignificantDigits" values="positiveInteger" default="1">
+        <description>
+          The minimum number of significant digits to use.
+        </description>
+      </option>
+      <option name="maximumSignificantDigits" values="positiveInteger" default="21">
+        <description>
+          The maximum number of significant digits to use.
+        </description>
+      </option>
+      <match validationRule="anyNumber" values="zero one two few many other" />
+    </matchSignature>
+
     <alias name="integer">
-      <description>Locale-sensitive integral number formatting</description>
+      <description>Locale-sensitive integral number formatting and selection</description>
       <setOption name="maximumFractionDigits" value="0" />
       <setOption name="style" value="decimal" />
+    </alias>
+
+    <alias name="ordinal" supports="match">
+      <description>Ordinal number selection</description>
+      <setOption name="select" value="ordinal" />
+    </alias>
+
+    <alias name="plural" supports="match">
+      <description>Plural number selection</description>
+      <setOption name="select" value="plural" />
     </alias>
   </function>
 


spec/registry.xml
```diff
@@ -243,11 +243,72 @@
...
+          A value with a smaller number of integer digits than this number will be
+          left-padded with zeros (to the specified length) when formatted.
+        </description>
+      </option>
+      <option name="minimumFractionDigits" values="positiveInteger">
+        <description>
+          The minimum number of fraction digits to use.
+          The default for plain number and percent formatting is 0;
+          the default for currency formatting is the number of minor unit digits provided by
+          the ISO 4217 currency code list (2 if the list doesn't provide that information).
```
aphillips, line 272: This is more complicated than it looks. There are several currencies with variable presentation (official vs. customary). In most cases, the currency overrides the fraction digits when it is set and we don't have to be overly specific, so perhaps:

```suggestion
the default for currency formatting depends on the currency code selected.
```

spec/registry.xml
```diff
@@ -243,11 +243,72 @@
...
+        </description>
+      </option>
+      <option name="maximumFractionDigits" values="positiveInteger">
+        <description>
+          The maximum number of fraction digits to use.
+          The default for plain number formatting is the larger of minimumFractionDigits and 3;
+          the default for currency formatting is the larger of minimumFractionDigits and the number of minor
+          unit digits provided by the ISO 4217 currency code list (2 if the list doesn't provide that information);
```
aphillips, line 280: see above

spec/registry.xml
```diff
@@ -243,11 +243,72 @@
...
+          The minimum number of significant digits to use.
+        </description>
+      </option>
+      <option name="maximumSignificantDigits" values="positiveInteger" default="21">
+        <description>
+          The maximum number of significant digits to use.
+        </description>
+      </option>
+      <match validationRule="anyNumber" values="zero one two few many other" />
+    </matchSignature>
```
aphillips, line 295: Are we missing some options? I'd expect to see a control for "groupingUsed" and "numberingSystem" and possibly some of the other common options (rounding, etc.)

spec/registry.xml
```diff
@@ -243,11 +243,72 @@
...
+
     <alias name="integer">
-      <description>Locale-sensitive integral number formatting</description>
+      <description>Locale-sensitive integral number formatting and selection</description>
```
aphillips, line 298: We should use less-jargony language, also, see above

```suggestion
  <description>Integer selection and formatting</description>
```

spec/registry.xml
```diff
@@ -243,11 +243,72 @@
...
       <setOption name="maximumFractionDigits" value="0" />
       <setOption name="style" value="decimal" />
     </alias>
+
+    <alias name="ordinal" supports="match">
+      <description>Ordinal number selection</description>
+      <setOption name="select" value="ordinal" />
```
aphillips, line 305: This also needs to be a formatter.

spec/registry.xml
```diff
@@ -243,11 +243,72 @@
...
+      <option name="select" values="plural ordinal exact" default="plural">
+        <description>
+          The number selection type.
+
+          With `exact` selection, only a variant with a key matching
+          the JSON representation of the input number (without an exponent) will be selected.
+          With `plural` and `ordinal` selection, if no such exactly matching key is present,
+          a variant with the matching CLDR plural category (`zero`/`one`/`two`/`few`/`many`/`other`)
+          for the corresponding selection type will be selected,
+          according to the rules of the current locale
```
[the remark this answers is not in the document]
eemeli: The choice of markdown is intentional, because:
1. It's remarkably human-readable even in syntax form.
2. It's a relatively well-defined formatting language that doesn't require any `<>` that would conflict with the surrounding XML.
3. It's rather well supported as a syntax for code comments.
4. What's the alternative here?

spec/registry.xml
eemeli: This is a copy of the text used for the corresponding formatting option. Could we take on as a separate task a review of these option descriptions?

spec/registry.xml
eemeli: The grouping & numbering systems don't affect the plural categorization, and the rounding options are not included for formatting either.

The options added here are the subset of currently defined formatting options that affect plural category selection.

spec/registry.xml
eemeli: This PR is strictly about implementing the consensus achieved on number selection.

As previously discussed in https://github.com/unicode-org/message-format-wg/pull/471#discussion_r1408109534, making ordinal also a formatter was not a part of the selected design. We can and should discuss it separately.

spec/registry.xml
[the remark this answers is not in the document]
eemeli: I merged the suggestion here, but kept the thread open to account for the markdown discussion. Also: not mentioning e.g. "match quality" was intentional, as this selector only has two quality levels: exact and category, with the former always pre-empting the latter.

The `*` is not mentioned here because it is not a concern for a selector function; its behaviour is defined directly in the selection algorithm, and the selector is never consulted about it.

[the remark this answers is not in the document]
aphillips: I have no objection to markdown. I just didn't want to let it pass by unremarked?

> The * is not mentioned here because it is not a concern for a selector function; its behaviour is defined directly in the selection algorithm, and the selector is never consulted about it.

I'm not sure that's correct. In the case of `plural` I would be unsurprised to see the keyword `other` in CLDR be treated as a synonym for `*` in MF2 (to prevent having two defaults). If that's not the case, some mention of the relationship is probably called for??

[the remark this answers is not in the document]
eemeli: `other` is not special in the syntax the way `*` is. In [Pattern Selection](https://github.com/unicode-org/message-format-wg/blob/main/spec/formatting.md#pattern-selection), it gets special treatment in:
- Resolve Preferences 2.ii.b. (this is where it's left out of the `keys` sent to the selector)
- Filter Variants 2.i.b.
- Sort Variants 5.iii.c.

To make it possible to use e.g. only `one` and `other` as English plural variant keys, leaving out the `*` variant, we would need to:
1. Remove the syntax [validity requirement](https://github.com/unicode-org/message-format-wg/blob/main/spec/syntax.md#matcher) to always have all-`*` variant.
2. Refactor the [Missing Fallback Variant error](https://github.com/unicode-org/message-format-wg/blob/main/spec/formatting.md#error-handling) into a formatting error for when we fail to select any variant.
3. Add a way for the [registry](https://github.com/unicode-org/message-format-wg/blob/main/spec/registry.md#data-model) `<match>` element to state that its `values` attribute exhaustively covers the input space, so that a build-time validator could still verify that the variant set is complete even if it does not include a `*` variant.

I would enthusiastically support such a set of changes.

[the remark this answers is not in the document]
aphillips: I'm not talking about leaving out `*` but about documenting that `*` and `other` are equivalent. What we don't want are tools expanding a variant set from `0`/`one`/`*` to `0`/`one`/`other`/`*`. I wouldn't make the latter set invalid (`other` is just a string) but it is pointless to write.

spec/registry.xml
aphillips: I saw that. I suppose we could review separately, but why not get it out of the way? But I could live with it.

spec/registry.xml
aphillips: We need a list of functions and options. I think maybe I'll pull one together so we can agree to it... Ignore this for now.

spec/registry.xml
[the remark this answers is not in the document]
eemeli: I think defining that here would require us to define the behaviour of e.g. selection on `{horse :plural}`. Would that silently match an `other` variant, or would it emit an error and end up selecting the `*` variant? And in e.g. Polish, could that match a `many` variant?

Do we really need to define in the spec such internal behaviour of a selector in error cases?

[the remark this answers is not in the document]
aphillips: The documentation of ICU's PluralFormat says:

> You always have to define a message text for the default plural case "other" which is contained in every rule set. If you do not specify a message text for a particular plural case, the message text of the plural case "other" gets assigned to this plural case. 

This is equivalent functionally to `*`. Do you disagree that tools should be told about this equivalence? Do you think we should promote developers having to write `0`/`one`/`other`/`*` messages? Note that the error being handled here isn't "horse" but rather "missing translation" in which the runtime locale would prefer to match `few` but the message doesn't provide `few`.

This says nothing about error handling. Note that we _do_ specify that `horse` is an error by making `:plural` and `:number`'s operand `anyNumber`, but that's wholly separate from describing what the plural cases are.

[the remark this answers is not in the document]
eemeli: > Do you disagree that tools should be told about this equivalence?

No, I agree that we should communicate clearly how `*` works, and that it's very similar to `other` in MF1. But I don't think this is the right place for that, because the rules about `*` apply to _all_ selectors, not just `:number`.

> Do you think we should promote developers having to write `0`/`one`/`other`/`*` messages?

No, that's unlikely to be helpful in most cases. But we should let them know that it's at least _possible_, for the rarer cases when distinguishing `other` from `*` is important.

> This says nothing about error handling. Note that we _do_ specify that `horse` is an error by making `:plural` and `:number`'s operand `anyNumber`, but that's wholly separate from describing what the plural cases are.

This sounds like a detail we ought to talk about some more, later. My position here would be that we can define in the spec a minimum of what must work, but that we should not seek to define behaviour for all possible inputs. Also, we ought to have normative text separately from what's in the `<description>` saying what a `:number` ought to be doing.

mihnita:
> This PR is strictly about implementing the consensus achieved on number selection.

I don't think there is agreement that the selection function name for plural is `:number`
I think is it worse than `:plural`, and I explained why.

Can I live with `:number` instead? Yes.
But I don't think there was an agreement on it `:number`.

spec/registry.xml
eemeli: Let's follow up on this under #596.

eemeli:
I believe that consensus on this was reached during the call this week.

aphillips:
We have a consensus on what functions should appear in the registry. There are open questions about the registry structure (that this PR cannot address). There are also questions about the options for the aliases, which I think can be decided separately. I don't think that merging this will have an impact on what we ultimately choose to do with these details, so I'm giving my approval.

aphillips (approved):

aphillips:
Without prejudice in favor of any specific functions, merging to clear.

PR #562, opened 2023-12-12
Add content-char as common root of simple-start-char, text-char, quoted-char & reserved-char

eemeli:
This change does not change the syntax in any way, and only introduces a `content-char` rule as a common base for other character classes.

At the moment, we have multiple explicit character range definitions in the ABNF, and it's not easy to see that they are very nearly the same as each other, and how exactly they differ. By introducing their overlap as `content-char`, this becomes more obvious to a reader.

### DIFF
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -28,6 +28,8 @@ annotation = (function *(s option))
            / reserved-annotation
 
 literal = quoted / unquoted
+quoted = "|" *(quoted-char / quoted-escape) "|"
+unquoted = name / number-literal
 variable = "$" name
 function = (":" / "+" / "-") identifier
 option = identifier [s] "=" [s] (literal / variable)
@@ -35,26 +37,6 @@ option = identifier [s] "=" [s] (literal / variable)
 input = %s".input"
 local = %s".local"
 match = %s".match"
-
-simple-start-char = %x0-2D         ; omit .
-                  / %x2F-5B        ; omit \
-                  / %x5D-7A        ; omit {
-                  / %x7C           ; omit }
-                  / %x7E-D7FF      ; omit surrogates
-                  / %xE000-10FFFF
-text-char = simple-start-char / "."
-
-quoted      = "|" *(quoted-char / quoted-escape) "|"
-quoted-char = %x0-5B         ; omit \
-            / %x5D-7B        ; omit |
-            / %x7D-D7FF      ; omit surrogates
-            / %xE000-10FFFF
-
-unquoted       = name
-               / number-literal
-; number-literal matches JSON number
-; https://www.rfc-editor.org/rfc/rfc8259#section-6
-number-literal = ["-"] (0 / ([1-9] *DIGIT)) ["." 1*DIGIT] [%i"e" ["-" / "+"] 1*DIGIT]
 
 ; Reserve additional .keywords for use by future versions of this specification.
 reserved-statement = reserved-keyword [s reserved-body] 1*([s] expression)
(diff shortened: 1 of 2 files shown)


spec/message.abnf
```diff
@@ -28,6 +28,8 @@ annotation = (function *(s option))
            / reserved-annotation
 
 literal = quoted / unquoted
+quoted = "|" *(quoted-char / quoted-escape) "|"
```
aphillips, line 31: Let's start at least some cleanup. Keep the `=` aligned with the other items in this block (yes, I see that `option` is not aligned... yet)

spec/message.abnf
```diff
@@ -94,6 +73,19 @@ name-start = ALPHA / "_"
...
            / %xB7 / %x300-36F / %x203F-2040
 
+content-char = %x00-08        ; omit HTAB (%x09) and LF (%x0A)
+             / %x0B-0C        ; omit CR (%x0D)
+             / %x0E-19        ; omit SP (%x20)
+             / %x21-2D        ; omit . (%x2E)
+             / %x2F-5B        ; omit \ (%x5C)
+             / %x5D-7A        ; omit { | } (%x7B-7D)
+             / %x7E-D7FF      ; omit surrogates
+             / %xE000-10FFFF
```
aphillips, line 83: As above, move the `=` and `/` out to match the *-char productions under this.

I think maybe a comment is well advised here, along the lines of:

```abnf
; base repertoire of characters for various text productions
```

Side note: we haven't done a great job of this, but a common practice in ABNF grammars is to put constituent parts _under_ where they are used (that is, forward definitions are the norm).

In this case, that suggests putting the *-char definitions first and then this item that they all consume.

spec/syntax.md
```diff
@@ -285,13 +285,16 @@ Whitespace in _text_, including tabs, spaces, and newlines is significant and MU
...
+content-char = %x00-08        ; omit HTAB (%x09) and LF (%x0A)
+             / %x0B-0C        ; omit CR (%x0D)
+             / %x0E-19        ; omit SP (%x20)
+             / %x21-2D        ; omit . (%x2E)
+             / %x2F-5B        ; omit \ (%x5C)
+             / %x5D-7A        ; omit { | } (%x7B-7D)
+             / %x7E-D7FF      ; omit surrogates
+             / %xE000-10FFFF
+simple-start-char = content-char / s / "|"
+text-char = content-char / s / "." / "|"
```
aphillips, line 297: Fix the alignment

spec/syntax.md
```diff
@@ -285,13 +285,16 @@ Whitespace in _text_, including tabs, spaces, and newlines is significant and MU
 be preserved during formatting.
```
aphillips, line 285: We should add some text about the grammar to make clear our intentions

aphillips (commented):
I kind of like it. I was semi-depressed about all of the lengthy char productions.

spec/message.abnf
eemeli: Could we take care of alignment in a single pass later, kinda like we resolved to not need Prettier to be automatically styling everything? Here, for instance, we may well later rename this rule as `quoted-literal` given that we also have `quoted-pattern`.

spec/syntax.md
eemeli: By "the grammar", do you here mean the added `content-char`, or something else?

aphillips: That's what I mean: some text about the ABNF productions. Just a quick sentence or two. Probably not on this specific line, but somewhere

spec/message.abnf
aphillips: 🤷 

That's fine, but I'd like to start locking the ABNF down to some degree.

spec/syntax.md
eemeli: Done.

spec/message.abnf
eemeli: Moved the `content-char` definition below its users.

aphillips (approved):

eemeli:
@aphillips Conflicts resolved here as well.

aphillips:
Merging per 2024-01-15 telecon

PR #564, opened 2023-12-15
(Design) Default registry and MF1 compatibility matrix

aphillips:
Create a list of selectors and formatters required by the default registry as well as a matrix showing how MF1 and MF2 map functionality. 

This is an attempt to capture the current (and proposed) registry entries for the spring release. If we're missing something, it needs to get added.

🎵 _...he's making a list and checking it twice..._  🎵

### DIFF
--- a/exploration/default-registry-and-mf1-compatibility.md
+++ b/exploration/default-registry-and-mf1-compatibility.md
@@ -112,7 +112,7 @@ How to write an MF1 format or selector in MF2:
 | Currency | {num,number,currency} | {$num :number currency=$code} {$num :currency}              |         |
 | Plural (selector)  | {num,plural, ...}    | .match {$num :number} {$num :plural}               |         |
 | Ordinal (selector) | {num,selectordinal, ...} | .match {$num :ordinal}                         |         |
-| Ordinal (format)   | {num,ordinal} | {$num :number type=ordinal} {$num :ordinal}               |         |
+| Ordinal (format)   | {num,ordinal} |                                                           | missing |
 | Date     | {date,date}          | {$date :datetime}                                            | short date is default |
 | Date     | {date,date,short}    | {$date :datetime dateStyle=short}                            | also medium,long,full |
 | Time     | {date,time}          | {$date :datetime timeStyle=short}                            | timeStyle required    |
--- a/exploration/default-registry-and-mf1-compatibility.md
+++ b/exploration/default-registry-and-mf1-compatibility.md
@@ -97,18 +97,20 @@ Options:
 
 ## Compatibility Matrix
 
+How to write an MF1 format or selector in MF2:
+
 | MF1      | Syntax               | MF2                                                          | Comment |
 |----------|----------------------|--------------------------------------------------------------|---------|
 | Number   | {num,number}         | {$num :number}                                               |         |
 | Integer  | {num,number,integer} | {$num :number maximumFractionDigits=0}  {$num :integer}      |         |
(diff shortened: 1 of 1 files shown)


exploration/default-registry-and-mf1-compatibility.md
```diff
@@ -0,0 +1,131 @@
...
+		<dt>First proposed</dt>
+		<dd>2023-12-15</dd>
+	</dl>
+</details>
+
+## Objective
+
+_What is this proposal trying to achieve?_
+
+This section contains the list of functions (both _selectors_ and _formatters_)
```
mihnita, line 19: Looks like there is quite a bit of duplication between this and what we already have in the `registry.xml`

What I thing is hight priority to address:
 - selectors, including decision on names: plural (cardinal & ordinal, with offset), select (or `string`?), maybe gender (maybe not now)
 - do we really want all the aliases?

I am afraid that we are reopening things that are already decided (formatters), and nobody complained about.

aphillips: This is, in fact, copied from `registry.xml` on purpose (thank you for making the original registry file). The only differences here should be the things proposed by @eemeli in the number selector PR. I wanted to have a way to discuss what goes into 2.0 without the clutter and to illustrate where we are vs. MF1.

eemeli (changes requested):
This seems to be mixing up two closely related but ultimately different things:
1. Defining the initial full set of functions and their aliases.
2. Documenting an MF1 → MF2 mapping.

I would find it easier to consider these separately, but I'm fine to also consider them together. If going the latter route, could we start with the functions and aliases on which we've previously reached (or are reaching) consensus, and leave out at least `:currency`, `:percent`, and `:ordinal` as a formatter? Any such additions ought to be argued for separately, as we've done for everything so far.

aphillips:
>  If going the latter route, could we start with the functions and aliases on which we've previously reached (or are reaching) consensus, and leave out at least :currency, :percent, and :ordinal as a formatter?

`:currency` and `:percent` are already in the registry, no?

`:ordinal` as a formatter is called out as missing from current consensus. It's weird to have the selector but not the formatter for that. I did, in fact, file an issue #565 for missing functionality (I also call for us _not_ to include some functionality by consensus in the spring release)

> This seems to be mixing up two closely related but ultimately different things:

Yes, that's true. However, I think there is a relationship between our minimum for launch and MF1 compatibility. Having a table that shows how we stack up turned out to be informative (to me). The real question is how we want to modify the aliases and options or if we want to include functions not listed here in the list that implementations are **_required_** to provide.

This item is, in fact, in the agenda for Monday.

eemeli:
> `:currency` and `:percent` are already in the registry, no?

No, they're not.

aphillips:
>>     :currency and :percent are already in the registry, no?
> No, they're not.

So what is this then? 

```
      <option name="style" values="decimal currency percent unit" default="decimal">
```

I mean, yes, that's not a function named `:currency` and a function names `:percent`. But that is currency and percent formatting.

eemeli:
> I mean, yes, that's not a function named `:currency` and a function names `:percent`. But that is currency and percent formatting.

Correct. We support currency, percent, and unit formatting via the `:number` formatter's `style` option. We do not have a consensus decision on including aliases such as `:currency` and `:percent`, and I don't think we necessarily need them.

Compared to MF1, this is rather similar to having `{x, number, percent}` instead of `{x, percent}`.

PR #566, opened 2023-12-15
Improve bidi isolation requirements

aphillips:
Addresses #549 

- Formally names the bidi isolation strategy as "Plain Text Bidirectional Isolation Strategy". 
- Refactors some text, fixing some problems with referring to bidi.
- Adds a better example from `ar-AE` locale

The algorithm isn't quite right yet.

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -686,23 +686,24 @@ and a U+007D RIGHT CURLY BRACKET `}`.
 ### Handling Bidirectional Text
 
 _Messages_ contain text. Any text can be 
-<a href="https://www.w3.org/TR/i18n-glossary/#dfn-bidirectional-text">bidirectional text</a>.
+[bidirectional text](https://www.w3.org/TR/i18n-glossary/#dfn-bidirectional-text).
 That is, the text can can consist of a mixture of left-to-right and right-to-left spans of text.
-The [Unicode Bidirectional Algorithm](http://www.unicode.org/reports/tr9/) [UAX9]
-is used to determine how bidirectional text is displayed.
+The display of bidirectional text is defined by the
+[Unicode Bidirectional Algorithm](http://www.unicode.org/reports/tr9/) [UAX9].
 
 The directionality of the message as a whole is provided by the _formatting context_.
 
 When a _message_ is formatted, _placeholders_ are replaced
 with their formatted representation.
-When the Unicode Bidirectional Algorithm is applied to the text of a formatted _message_ (including its formatted parts)
-there can be unexpected or undesirable 
-<a href="https://www.w3.org/TR/i18n-glossary/#dfn-spillover-effects">spillover effects</a>.
-Applying <a href="https://www.w3.org/TR/i18n-glossary/#dfn-bidi-isolation">bidi isolation</a>
-to the formatted values helps avoid spillover in a formatted _message_.
+Applying the Unicode Bidirectional Algorithm to the text of a formatted _message_ 
+(including its formatted parts)
+can result in unexpected or undesirable 
+[spillover effects](https://www.w3.org/TR/i18n-glossary/#dfn-spillover-effects).
+Applying [bidi isolation](https://www.w3.org/TR/i18n-glossary/#dfn-bidi-isolation)
+to each affected formatted value helps avoid this spillover in a formatted _message_.
 
-The reference data model includes direction for both the _message_ and, separately,
-for each _placeholder_.
+Note that both the _message_ and, separately, each _placeholder_ need to have
+direction metadata for this to work.
 If an implementation supports formatting to something other than a string
 (such as a sequence of parts),
 the directionality of each formatted _placeholder_ needs to be available to the caller.
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -692,9 +692,8 @@ isolating such parts to ensure that the formatted value displays correctly in a 
 > The code point sequence for this string, as produced by the ICU4J `NumberFormat` function,
 > includes **U+200F U+200E** at the start and **U+200F** at the end of the string.
 > If it did not do this, the same string would appear like this instead:
-> ```
+> 
 > ![image](https://github.com/unicode-org/message-format-wg/assets/69082/6cc7f16f-8d9b-400b-a333-ae2ddb316edb)
->```
 
 A **_bidirectional isolation strategy_** is functionality in the formatter's
 processing of a _message_ that produces bidirectional output text that is ready for display.
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -731,12 +731,12 @@ processing of a _message_ that produces bidirectional output text that is ready 
 
 The **_Default Bidi Strategy_** is a _bidirectional isolation strategy_ that uses
 isolating Unicode control characters around _placeholder_'s formatted values.
-It is intended for use in plain-text strings, where markup or other mechanisms
+It is primarily intended for use in plain-text strings, where markup or other mechanisms
 are not available.
 Implementations MUST provide the _Default Bidi Strategy_ as one of the 
 _bidirectional isolation strategies_.
 
-Implementations MAY provide others _bidirectional isolation strategies_.
+Implementations MAY provide other _bidirectional isolation strategies_.
 
 Implementations MAY supply a _bidirectional isolation strategy_ that performs no processing.
 
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -683,8 +683,10 @@ isolating such parts to ensure that the formatted value displays correctly in a 
 > ```
 > ‎-1,234.56 د.إ.‏
 > ```
-> The code point sequence used by the ICU4J `NumberFormat` function for this is:
-> U+200F U+200E U+002D U+0031 U+002C U+0032 U+0033 U+0034 U+002E U+0035 U+0036 U+00A0 U+062F U+002E U+0625 U+002E U+200F
+> The code point sequence produced by the ICU4J `NumberFormat` function for this is:
+> **U+200F U+200E** U+002D U+0031 U+002C U+0032 U+0033 U+0034 U+002E U+0035 U+0036
+> U+00A0 U+062F U+002E U+0625 U+002E **U+200F**
+> Note that U+200F is RIGHT-TO-LEFT MARK and U+200E is LEFT-TO-RIGHT MARK.
 
 Implementations formatting messages as a concatenated string or a sequence of strings
 MUST provide one or more strategies for bidirectional isolation.
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -676,22 +676,28 @@ its formatter SHOULD perform any necessary processing, such as inserting control
 isolating such parts to ensure that the formatted value displays correctly in a plain text context.
 
 > For example, an implementation could provide a `:currency` formatting function
-> which inserts RLM, ALM, or LRM characters to properly display a currency symbol
-> next to a formatted number.
-> An example of this is formatting the value `1234.56` as the currency `AED`
+> which inserts strongly directional characters, such as U+200F RIGHT-TO-LEFT MARK (RLM),
+> U+200E LEFT-TO-RIGHT MARK (LRM), or U+061C ARABIC LETTER MARKER (ALM), 
+> to coerce proper display of the sign and currency symbol next to a formatted number.
+> An example of this is formatting the value `-1234.56` as the currency `AED`
 > in the `ar-AE` locale. The formatted value appears like this:
 > ```
 > ‎-1,234.56 د.إ.‏
 > ```
-> The code point sequence produced by the ICU4J `NumberFormat` function for this is:
-> **U+200F U+200E** U+002D U+0031 U+002C U+0032 U+0033 U+0034 U+002E U+0035 U+0036
-> U+00A0 U+062F U+002E U+0625 U+002E **U+200F**
-> Note that U+200F is RIGHT-TO-LEFT MARK and U+200E is LEFT-TO-RIGHT MARK.
+> The code point sequence for this string, as produced by the ICU4J `NumberFormat` function,
+> includes **U+200F U+200E** at the start and **U+200F** at the end of the string.
 
 Implementations formatting messages as a concatenated string or a sequence of strings
 MUST provide one or more strategies for bidirectional isolation.
-One such strategy MUST be the **_Plain Text Bidirectional Isolation Strategy_**, which behaves
-as follows:
+
+The **_Default Bidi Strategy_** is a bidirectional isolation strategy that uses
+isolating Unicode control characters around placeholders.
+It is intended for use in plain-text strings, where markup or other mechanisms
+are not available.
+Implementations MUST provide the _Default Bidi Strategy_ as one of the 
+bidirectional isolation strategies.
+
+The _Default Bidi Strategy_ is defined as follows:
 
 1. Let `msgdir` be the directionality of the whole message,
    one of « `'LTR'`, `'RTL'`, `'unknown'` ».
@@ -701,17 +707,17 @@ as follows:
    1. Let `fmt` be the formatted string representation of the resolved value of `exp`.
    1. Let `dir` be the directionality of `fmt`,
       one of « `'LTR'`, `'RTL'`, `'unknown'` », with the same meanings as for `msgdir`.
-   1. If `dir` is `'unknown'`:
+   1. If `dir` is `'RTL'`:
       1. In the formatted output,
-         prefix `fmt` with U+2068 FIRST STRONG ISOLATE
+         prefix `fmt` with U+2067 RIGHT-TO-LEFT ISOLATE
          and postfix it with U+2069 POP DIRECTIONAL ISOLATE.
-   1. Else, if `dir` is `'LTR'` and `msgdir` is not `'LTR'`:
+   1. Else, if `dir` is `'LTR'`:
       1. In the formatted output,
          prefix `fmt` with U+2066 LEFT-TO-RIGHT ISOLATE
          and postfix it with U+2069 POP DIRECTIONAL ISOLATE.
-   1. Else, if `dir` is `'RTL'` and `msgdir` is not `'RTL'`:
+   1. Else:
       1. In the formatted output,
-         prefix `fmt` with U+2067 RIGHT-TO-LEFT ISOLATE
+         prefix `fmt` with U+2068 FIRST STRONG ISOLATE
          and postfix it with U+2069 POP DIRECTIONAL ISOLATE.
 
 Alternatives to this strategy MAY be provided by implementations,
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -705,7 +705,7 @@ The reference data model includes direction for both the _message_ and, separate
 for each _placeholder_.
 If an implementation supports formatting to something other than a string
 (such as a sequence of parts),
-the directionality of each _placeholder_ needs to be available to the caller.
+the directionality of each formatted _placeholder_ needs to be available to the caller.
 
 If a formatted _expression_ itself contains spans with differing directionality,
 its formatter SHOULD perform any necessary processing, such as inserting controls or
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -703,25 +703,32 @@ The _Default Bidi Strategy_ is defined as follows:
    one of « `'LTR'`, `'RTL'`, `'unknown'` ».
    These correspond to the message having left-to-right directionality,
    right-to-left directionality, and to the message's directionality not being known.
+1. Let `localedir` be the default directionality of the current locale,
+   one of « `'LTR'`, `'RTL'`, `'unknown'` ».
+   These correspond to the locale having left-to-right directionality,
+   right-to-left directionality, and to the locale's default directionality not being known.
 1. For each _expression_ `exp` in _pattern_:
    1. Let `fmt` be the formatted string representation of the resolved value of `exp`.
    1. Let `dir` be the directionality of `fmt`,
       one of « `'LTR'`, `'RTL'`, `'unknown'` », with the same meanings as for `msgdir`.
-   1. If `dir` is `'RTL'`:
+   1. If `dir` is `'LTR'`:
+      1. If `msgdir` and `localedir` are both `'LTR'`
+         in the formatted output, let `fmt` be itself
+      1. Else, in the formatted output,
+         prefix `fmt` with U+2066 LEFT-TO-RIGHT ISOLATE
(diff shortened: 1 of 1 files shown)


spec/formatting.md
```diff
@@ -671,16 +672,26 @@ MAY rely on the introspection of its contents, or on other means.
...
-negatively affecting the presentation of the overall formatted result.
-
-> For example, an implementation could provide a `:number` formatting function
-> which would always produce output matching the message's locale,
-> allowing for its formatted string representation to never need isolation.
+its formatter SHOULD perform any necessary processing, such as inserting controls or
+isolating such parts to ensure that the formatted value displays correctly in a plain text context.
+
+> For example, an implementation could provide a `:currency` formatting function
+> which inserts RLM, ALM, or LRM characters to properly display a currency symbol
```
eemeli, line 679: These abbreviations will not be familiar to every reader, and should be avoided esp. as this is the only place where they're used.

spec/formatting.md
```diff
@@ -671,16 +672,26 @@ MAY rely on the introspection of its contents, or on other means.
...
+> next to a formatted number.
+> An example of this is formatting the value `1234.56` as the currency `AED`
+> in the `ar-AE` locale. The formatted value appears like this:
+> ```
+> ‎-1,234.56 د.إ.‏
+> ```
```
eemeli, line 685: The signs on the values don't match.

spec/formatting.md
```diff
@@ -671,16 +672,26 @@ MAY rely on the introspection of its contents, or on other means.
...
+> The code point sequence produced by the ICU4J `NumberFormat` function for this is:
+> **U+200F U+200E** U+002D U+0031 U+002C U+0032 U+0033 U+0034 U+002E U+0035 U+0036
+> U+00A0 U+062F U+002E U+0625 U+002E **U+200F**
```
eemeli, line 688: This ought to be fully annotated or expressed somehow differently, as we should not presume that the reader will know what all of these code points are.

spec/formatting.md
```diff
@@ -671,16 +672,26 @@ MAY rely on the introspection of its contents, or on other means.
...
+> Note that U+200F is RIGHT-TO-LEFT MARK and U+200E is LEFT-TO-RIGHT MARK.
 
 Implementations formatting messages as a concatenated string or a sequence of strings
 MUST provide one or more strategies for bidirectional isolation.
-One such strategy MUST behave as follows:
+One such strategy MUST be the **_Plain Text Bidirectional Isolation Strategy_**, which behaves
+as follows:
```
eemeli, line 694: This needs a snappier name; a single word (maybe max two) that can be used as an option value. I'm also not sure about "plain text", given that other strategies that apply more introspection could apply equivalently well to plain text.

eemeli (changes requested):

spec/formatting.md
aphillips: Fixed. Thanks.

spec/formatting.md
aphillips: Fixed. Thanks.

spec/formatting.md
aphillips: Changed for clarity. Check it out.

spec/formatting.md
aphillips: How about the new text's version: _Default Bidi Strategy_?

eemeli: `default` is a good name.

spec/formatting.md
```diff
@@ -690,20 +707,20 @@ One such strategy MUST behave as follows:
    1. Let `fmt` be the formatted string representation of the resolved value of `exp`.
    1. Let `dir` be the directionality of `fmt`,
       one of « `'LTR'`, `'RTL'`, `'unknown'` », with the same meanings as for `msgdir`.
-   1. If `dir` is `'unknown'`:
+   1. If `dir` is `'RTL'`:
       1. In the formatted output,
-         prefix `fmt` with U+2068 FIRST STRONG ISOLATE
+         prefix `fmt` with U+2067 RIGHT-TO-LEFT ISOLATE
          and postfix it with U+2069 POP DIRECTIONAL ISOLATE.
-   1. Else, if `dir` is `'LTR'` and `msgdir` is not `'LTR'`:
```
eemeli, line 697: So just to clarify, you would like us to change the default bidi isolation strategy to _always_ add isolates around placeholders, even if they match the message's base direction? So for example when formatting the message `Hello {42 :number}!` in English, the expected output would be:
```
Hello \u{2066}42\u{2069}!
```

Could you give an example for what sort of situation and message this change is necessary for?

aphillips: The challenge here is to get isolation when it is needed without introspecting the characters in the pattern and placeholder string. Spillover can occur for a number of reasons, but it requires a string that has mixed direction and it most frequently involves cases where the placeholder and adjacent text have the same direction (so that there is no directional boundary) or where neutral or weakly directional characters are involved.

Numbers (especially currency values) and dates often have weakly directional or opposite directional characters in them and they are common in placeholders. An example of this is [Example 13](https://w3c.github.io/i18n-discuss/explainers/bidi-html-demo.html?item=13), where RTL placeholders for two currency values have a neutral between them:

image


We could require RTL patterns to do isolation all the time (because, excepting Adlam, numbers are weakly LTR in most such languages) and exempt LTR patterns for LTR placeholders. Or we could allow/require implementations to introspect. If there is no mixing of directional runs, there can't be spillover. Let me make some additional edits...

eemeli: How would that look like in an MF2 message pattern? I'm struggling to figure out which parts of the example are the nominally RTL parts that need isolation.

aphillips: The Arabic means "The price is {$price :currency} + {$shipping :currency} shipping".

The first value is "1234.56 USD". The second is "12.99 AED" (which currency is used doesn't matter). Both are formatted for the ar-AE locale in the example and thus have a base direction of RTL (the "USD" should appear to the left of the number because it is in a trailing position). The appearance of LTR characters in a formatted RTL string is not unusual.

The problem here is that `+` is a neutral, so you end up with the weak `12.99` being attached to the end of the LTR run. If a `dir=auto` or `dir=rtl` were added to the `<span>` elements, the message presents correctly. FWIW, message patterns need isolation to be legible in RTL:

image

aphillips: Check out the change to the algorithm. Basically I allow implementations to omit isolates when the message and locale are both LTR (which is the most common case).

spec/formatting.md
```diff
@@ -671,40 +672,63 @@ MAY rely on the introspection of its contents, or on other means.
...
+The _Default Bidi Strategy_ is defined as follows:
 
 1. Let `msgdir` be the directionality of the whole message,
    one of « `'LTR'`, `'RTL'`, `'unknown'` ».
    These correspond to the message having left-to-right directionality,
    right-to-left directionality, and to the message's directionality not being known.
+1. Let `localedir` be the default directionality of the current locale,
+   one of « `'LTR'`, `'RTL'`, `'unknown'` ».
+   These correspond to the locale having left-to-right directionality,
+   right-to-left directionality, and to the locale's default directionality not being known.
```
eemeli, line 709: Do we really need this `localedir`? The term "current locale" is also rather ambiguous here.

At the message level, `msgdir` should already be determined from the locale if it's not explicitly set. The only way I can see this having any effect is if a user is formatting a string in an RTL locale but explicitly asks for LTR directionality, and has a custom formatter function that's resolving with LTR directionality. This seems like such an edge case that I can't tell if it ever happens, or what would be the right thing to do here.

aphillips: You're forgetting locale fallback on the resource level? Suppose my code looks like this:

```javascript
var locale = new Intl.Locale("ar-AE"); // an Arabic locale
// the bundle 'foo' isn't localized yet and is in English
var pattern = Intl.MessageResource.getBundle("foo", locale).getMessage("bar");
var mf = new Intl.MessageFormat(pattern, locale);
var output = mf.format({"arg":someNumber, "arg2": someDate});
```

Here my pattern is in locale `und` and has LTR directionality, but my formatters are in Arabic (and have RTL directionality). This is a super common situation, particularly in development. In this case, `localedir` is not the same as `msgdir` and you **_really_** want isolation for that number and date.

I'll admit that in most cases the base direction of a specific message generally matches that of the _bundle_ locale. Users can have LTR strings or occasionally auto-directional strings. For example, if I'm assembling a part number, I might have: `PREFIX-{$partId}-SUFFIX` (where prefix and suffix are ID parts themselves--think VIN number or things like that).

Anyway, the reason I'm using `localedir` is to get the direction of each expression (I could put it in the expression loop, I suppose) as a proxy for telling the implementation to decide by groping the character sequence. We want to know if there is likely to be a counter-directional run anywhere in the string and, for most formatters, the locale is close enough?

eemeli: > You're forgetting locale fallback on the resource level? Suppose my code looks like this:
> 
> ```js
> var locale = new Intl.Locale("ar-AE"); // an Arabic locale
> // the bundle 'foo' isn't localized yet and is in English
> var pattern = Intl.MessageResource.getBundle("foo", locale).getMessage("bar");
> var mf = new Intl.MessageFormat(pattern, locale);
> var output = mf.format({"arg":someNumber, "arg2": someDate});
> ```
> 
> Here my pattern is in locale `und` and has LTR directionality, but my formatters are in Arabic (and have RTL directionality). This is a super common situation, particularly in development. In this case, `localedir` is not the same as `msgdir` and you **_really_** want isolation for that number and date.

According to the current JS spec for `Intl.MessageFormat`, the `pattern` value above must be an MF2 message source string or an object matching the MF2 message data model. Neither that string or object carries any explicit information about its locale or direction. So when `mf` is constructed, an assumption is made that the `pattern` locale is `locale`. Subsequently, `msgdir` and `localedir` will always be the same, determined from `locale`.

The constructor does support an explicit `dir` option, which explicitly sets `msgdir` rather than letting it be determined from `locale`:
```js
const mf = new Intl.MessageFormat(pattern, 'ar-AE', { dir: 'LTR' });
```
With that, `msgdir` would be `'LTR'` and `localedir` would be `'RTL'`. While formatter functions would get called with the `msgdir` direction, at least `:number` would only consider the ar-AE locale, and would produce RTL output, which would therefore get isolated irrespective of `localedir`.

> I'll admit that in most cases the base direction of a specific message generally matches that of the _bundle_ locale. Users can have LTR strings or occasionally auto-directional strings. [...]

We don't have bundles, though. As we've left everything beyond a single message out of scope, I don't think we really have a way of expressing a mismatch between the pattern's inherent direction and the direction a user claims the pattern has.

> Anyway, the reason I'm using `localedir` is to get the direction of each expression (I could put it in the expression loop, I suppose) as a proxy for telling the implementation to decide by groping the character sequence. We want to know if there is likely to be a counter-directional run anywhere in the string and, for most formatters, the locale is close enough?

Ah, I think we have different mental models of how we're expecting placeholders to be formatted? Continuing with the previous example where the message is being formatted as `{ locale: 'ar-AE', dir: 'LTR' }`, my expectation is that a placeholder formatter will in general care about the locale, but _not_ about the message's overall direction. So a formatted number would have `RTL` directionality, and a variable reference to a random string would have `auto` directionality. Are you presuming that one or both of those would inherit the nominal `LTR` direction of the message?

aphillips: > We don't have bundles, though. As we've left everything beyond a single message out of scope, I don't think we really have a way of expressing a mismatch between the pattern's inherent direction and the direction a user claims the pattern has.

**_We_** don't have bundles, but most applications do. And the resources working group will want to define things in a modern way. Expressing (and, where needed, overriding) the direction of a message should be expected to happen externally.

> Ah, I think we have different mental models of how we're expecting placeholders to be formatted? Continuing with the previous example where the message is being formatted as { locale: 'ar-AE', dir: 'LTR' }, my expectation is that a placeholder formatter will in general care about the locale, but not about the message's overall direction. So a formatted number would have RTL directionality, and a variable reference to a random string would have auto directionality. Are you presuming that one or both of those would inherit the nominal LTR direction of the message?

No, I think our mental models are similar. The placeholder formatter in general cares about the locale, since this tells it how to do whatever locale-affected operation it does (like formatting a number). It will use its locale to compute its output's direction (in cases like number). The formatter code won't really consume input `dir`, but it might expose it if it is a formatter like `:string` that is just doing pass-through or as a hint (suppose the formatter is `:person`). And, in these cases, the formatter only cares about the base direction of the _variable_. The direction of the message is opaque.

What my paragraph about `localedir` was trying to say is this: 

You need bidi isolation (and thus isolating controls) if the string contains mixed direction runs. To know this, you need to know the entire character sequence of the formatted message and then look at character properties. That's too great a burden for implementations, so, to avoid over-inserting bidi controls, I am trying to use the message's direction and the locale to estimate whether it is likely that the message needs isolation without looking at the characters. 

This estimation will be broken if the pattern contains any anti-directional runs, but this isn't that common. And I put `localedir` outside the expression loop because it's pretty likely that all of the formatters have the same locale (technically I shouldn't in case we do `@locale`, so I guess I'll go fix that.......).

The result is that, in LTR messages in LTR locales, most placeholders won't have isolating controls. Otherwise we'll isolate by default.

eemeli: > **_We_** don't have bundles, but most applications do. And the resources working group will want to define things in a modern way. Expressing (and, where needed, overriding) the direction of a message should be expected to happen externally.

Agreed. Could you add a comment on eemeli/message-resource-wg#19 as I believe we may want to include a tag like `@dir` there?

> What my paragraph about `localedir` was trying to say is this:
> 
> You need bidi isolation (and thus isolating controls) if the string contains mixed direction runs. To know this, you need to know the entire character sequence of the formatted message and then look at character properties. That's too great a burden for implementations, so, to avoid over-inserting bidi controls, I am trying to use the message's direction and the locale to estimate whether it is likely that the message needs isolation without looking at the characters.
> 
> This estimation will be broken if the pattern contains any anti-directional runs, but this isn't that common. And I put `localedir` outside the expression loop because it's pretty likely that all of the formatters have the same locale (technically I shouldn't in case we do `@locale`, so I guess I'll go fix that.......).

And my argument here is that it's more surprising than useful for the locale's direction to affect the algorithm as proposed. To recap, we have:
- `localedir` determined from the language, region and script of the current locale
- `msgdir` either explicitly set, or the same as `localedir`
- `dir` as the direction of the formatted result of the placeholder.

In order for `localedir` to have an effect on the end results, we need a situation where `localedir` is RTL while `msgdir` and `dir` are both LTR. This means that:
- The formatting locale is an RTL locale, such as the ar-AE we've been using as an example.
- The formatter is called with an explicit LTR override for the message's direction.
- The formatted placeholder value has LTR directionality, even though it's been called with an RTL locale.

When, and how, can this set of circumstances happen? Based on our discussions above, I believe we're agreed that in this case both `:datetime` and `:number` would format to RTL, and `:string` & bare input variables would format to "auto" direction. I believe that exhaustively covers the scope of what we're currently supporting in 2.0, so how could we even test for this?

> The result is that, in LTR messages in LTR locales, most placeholders won't have isolating controls. Otherwise we'll isolate by default.

Sidetracking, but I would guess that with `default` most placeholders will still end up with isolation. For example, we'll include isolation for `Hello {$name}!` because we can't know without introspection the direction of `$name`.

aphillips: > When, and how, can this set of circumstances happen?

It might not be in 2.0, but:

```
{{My locale is ar-EG and I'm not (yet) localized and I have a {$now :date @locale=en} example.}}
```

> Sidetracking, but I would guess that with default most placeholders will still end up with isolation. For example, we'll include isolation for Hello {$name}! because we can't know without introspection the direction of $name.

That's true for a plain string, although my W3C WG encourages all strings to come with language and direction metadata. Many placeholders, though, will be number or dates and the formatter will know or generate the direction.

On some level, I am not fond of the algorithm versus using normative prose. If we wrote `default` as prose, we could allow an implementation to determine unidirectional situations. That could read something like:

> The **_default bidi isolation strategy_** is one in which placeholders are prefixed
> with the Unicode isolating control that matches their base direction and are
> suffixed with U+2069 POP DIRECTIONAL ISOLATE.
> Implementations MAY omit the isolates when they determine that the direction
> of the _message_, formatted _locale_, and the direction of
> the value of the _placeholder_ all match; 
> or that the resulting string would not be bidirectional;
> or some other implementation-specific criteria are met.

eemeli: > > When, and how, can this set of circumstances happen?
> 
> It might not be in 2.0, but:
> 
> ```
> {{My locale is ar-EG and I'm not (yet) localized and I have a {$now :date @locale=en} example.}}
> ```

Ok, that's plausible with expression attributes. But why in that message would the `:date` need isolation? How is this message being formatted with an LTR override on an RTL locale different from this same message being formatted with an LTR locale?

> On some level, I am not fond of the algorithm versus using normative prose. If we wrote `default` as prose, we could allow an implementation to determine unidirectional situations. That could read something like: [...]

I think we do need an explicit algorithm definition that is based purely on data, and does not include any optional behaviour. The specific use case that I have in mind is [hydration](https://en.wikipedia.org/wiki/Hydration_(web_development)), where the server and the client need to be able to render the same content exactly the same. For that to work with MF2 messages, we need a way for the two different implementations to produce the same output, and that includes bidi isolation.

There are certainly ways to improve on the default algorithm, and implementations are rather likely to provide them, but we also need one that behaves the same in all implementations.

macchiati: I think we need to focus on the results, not the exact method of achieving those results. Here is a rough statement for discussion.

Suppose that have a message X intended for locale L, and we create from that a message Y, where bidi isolation is applied around each of the placeholders's curly brackets. For external parameters p1, p2, ... we produce 
FX = format(X, L, p1, p2, ...) 
FY = format(X, L, p1, p2, ...). 

We would expect that after filtering out BIDI controls, the characters in FX are the same as FY (ie, in the same order).

spec/formatting.md
```diff
@@ -685,60 +685,88 @@ and a U+007D RIGHT CURLY BRACKET `}`.
...
+When a _message_ is formatted, _placeholders_ are replaced
+with their formatted representation.
+When UBA is applied to the text of a formatted _message_ (including its formatted parts)
+there can be unexpected or undesirable 
+<a href="https://www.w3.org/TR/i18n-glossary/#dfn-spillover-effects">spillover effects</a>.
+Applying <a href="https://www.w3.org/TR/i18n-glossary/#dfn-bidi-isolation">bidi isolation</a>
+to the formatted values helps avoid spillover in a formatted _message_.
+
+The reference data model includes direction for both the _message_ and, separately,
+for each _placeholder_.
```
eemeli, line 705: Actually, it doesn't. The data model is about input, and direction is an attribute of the output.

spec/formatting.md
```diff
@@ -685,60 +685,88 @@ and a U+007D RIGHT CURLY BRACKET `}`.
...
+A **_bidirectional isolation strategy_** is functionality in the formatter's
+processing of a _message_ that produces bidirectional output text that is ready for display.
 
-Implementations formatting messages as a concatenated string or a sequence of strings
-MUST provide one or more strategies for bidirectional isolation.
-One such strategy MUST behave as follows:
+The **_Default Bidi Strategy_** is a _bidirectional isolation strategy_ that uses
+isolating Unicode control characters around _placeholder_'s formatted values.
+It is intended for use in plain-text strings, where markup or other mechanisms
+are not available.
```
eemeli, line 735: We should not leave out the possibility of using it for e.g. formatted parts.
```suggestion
It is primarily intended for use in plain-text strings, where markup or other mechanisms
are not available.
```

spec/formatting.md
```diff
@@ -685,60 +685,88 @@ and a U+007D RIGHT CURLY BRACKET `}`.
...
+Implementations MUST provide the _Default Bidi Strategy_ as one of the 
+_bidirectional isolation strategies_.
+
+Implementations MAY provide others _bidirectional isolation strategies_.
```
eemeli, line 739:
```suggestion
Implementations MAY provide other _bidirectional isolation strategies_.
```

spec/formatting.md
```diff
@@ -685,60 +685,88 @@ and a U+007D RIGHT CURLY BRACKET `}`.
...
       one of « `'LTR'`, `'RTL'`, `'unknown'` », with the same meanings as for `msgdir`.
-   1. If `dir` is `'unknown'`:
-      1. In the formatted output,
-         prefix `fmt` with U+2068 FIRST STRONG ISOLATE
-         and postfix it with U+2069 POP DIRECTIONAL ISOLATE.
-   1. Else, if `dir` is `'LTR'` and `msgdir` is not `'LTR'`:
-      1. In the formatted output,
+   1. If `dir` is `'LTR'`:
+      1. If `msgdir` and `localedir` are both `'LTR'`
+         in the formatted output, let `fmt` be itself
```
eemeli, line 759: I still contend that the reference here to `localedir` and its definition above should be left out.

```suggestion
      1. If `msgdir` is `'LTR'`, let `fmt` be itself.
```

spec/formatting.md
```diff
@@ -685,60 +685,88 @@ and a U+007D RIGHT CURLY BRACKET `}`.
 
 ### Handling Bidirectional Text
 
-_Messages_ contain text which can be bidirectional,
-consisting of a mixture of left-to-right and right-to-left spans of text.
+_Messages_ contain text. Any text can be 
+<a href="https://www.w3.org/TR/i18n-glossary/#dfn-bidirectional-text">bidirectional text</a>.
+That is, the text can can consist of a mixture of left-to-right and right-to-left spans of text.
+The [Unicode Bidirectional Algorithm](http://www.unicode.org/reports/tr9/) [UAX9], aka "UBA",
+is used to determine how bidirectional text is displayed.
```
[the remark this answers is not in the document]
aphillips: This text wasn't clear. I rewrote to make clear that Unicode defines UBA for this purpose.

spec/formatting.md
aphillips: Ack. See changes.

spec/formatting.md
```diff
@@ -685,60 +685,88 @@ and a U+007D RIGHT CURLY BRACKET `}`.
 
 ### Handling Bidirectional Text
 
-_Messages_ contain text which can be bidirectional,
-consisting of a mixture of left-to-right and right-to-left spans of text.
+_Messages_ contain text. Any text can be 
+<a href="https://www.w3.org/TR/i18n-glossary/#dfn-bidirectional-text">bidirectional text</a>.
```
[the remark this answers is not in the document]
aphillips: Context switch from W3C docs left me momentarily disoriented. Fixed.

spec/formatting.md
aphillips: I don't quite agree for the reasons I stated earlier. But I made the change anyway. "Your" way will result in a some LTR values being unwrapped when maybe they should be.

aphillips:
Per telecon discussion, merging

PR #567, opened 2023-12-16
Re-add accidentally dropped message validity condition

eemeli:
As discovered in https://github.com/unicode-org/message-format-wg/pull/557#discussion_r1428727301, this was added in #431 but accidentally dropped in the #441 refactor.


aphillips (approved):

aphillips:
Need a second ship-it for fast track. This is group consensus and should be merged.

stasm (approved):

PR #571, opened 2023-12-18
Relax stability for tech preview

### LINKED ISSUE #568: We should relax the stability policy for the tech preview release
One particularly powerful commitment that we're currently making is this part of our [Stability Policy](https://github.com/unicode-org/message-format-wg/tree/main/spec#stability-policy):

> Updates to this specification will not introduce message syntax that, when parsed according to earlier versions of this specification, would produce syntax or data model errors.

The idea here is that you could write a tool that works with 2.0 syntax, and as long as it didn't touch the internals of any reserved syntax, that tool would also work for all future 2.1, 2.2, 2.n syntaxes because they would only ever modify the parts that we had explicitly reserved.

I'm starting to think that for next spring's tech preview release we should not make this commitment, and to allow for later releases to introduce syntax that a tech preview parser would not understand. This would allow us to leave out for now _all_ the reserved rules, and to not tie our hands quite so tightly regarding later changes.

Doing this would allow us to, for example:
- Leave out standalone markup placeholders from the tech preview, but add them in later.
- Add expression attributes to the syntax after the tech preview.
- Table discussion on literal and pattern delimiters until later.

To be clear, I'm not proposing that we necessarily do any of the above, but we _could_ if we first relaxed the stability policy. And perhaps most significantly, we would not need to enumerate already now all the possible changes we could consider later.

I still think that the current commitment is one we ought to make, but maybe a bit later than this coming Spring?

I would also be interested in explicitly reserving the possibility to make some breaking changes to the syntax after the tech preview release. Specifically, I remain concerned about our literal and pattern delimiter choices, but I think further discussion on them would benefit greatly from wider review comments from outside our group. Is it technically possible for us to reserve the right to make such specific changes, so that we don't need to seek full consensus on delimiters before the release?

aphillips:
Fixes #568 

Consensus in 2023-12-18 call

### DIFF
--- a/spec/README.md
+++ b/spec/README.md
@@ -97,11 +97,19 @@ Additional options or option values MAY be defined.
 
 Later specification versions MAY make previously invalid messages valid.
 
-Updates to this specification will not introduce message syntax that,
-when parsed according to earlier versions of this specification,
-would produce syntax or data model errors.
-Such messages MAY produce errors when formatted
-according to an earlier version of this specification.
+> [!NOTE]
+> For the purposes of a technical preview in LDML v45, the following stability
+> policy does not apply.
+> The working group expects that this policy will be enforced once
+> MessageFormat has completed its preview:
+> 
+> ---
+> 
+>> Updates to this specification will not introduce message syntax that,
+>> when parsed according to earlier versions of this specification,
+>> would produce syntax or data model errors.
+>> Such messages MAY produce errors when formatted
+>> according to an earlier version of this specification.
 
 From version 2.0, MessageFormat will only reserve, define, or require
 function names or function option names


aphillips:
> In addition to this exclusion, we should include something about reserving the right to make any changes to the tech preview release, should there be sufficient cause.

That's reasonable. I made a change to make the stability policy inoperative until after the tech preview ends (and the spec is stabilized at 2.0). In effect, the stability policy itself is a "tech preview" 😁

spec/README.md
```diff
@@ -70,6 +70,14 @@ A reference to a _term_ looks like this.
 
 ### Stability Policy
 
+> [!IMPORTANT]
+> The provisions of the stability policy are not in effect until
+> the conclusion of the technical preview and adoption of this specification.
+> Implementers should expect that changes to the specification during
+> and following the preview period will minimal and will require compelling evidence.
```
eemeli:
```suggestion
> and following the preview period will be minimal and will require compelling evidence.
```

eemeli (approved):

macchiati:
I think we should drop the following; this is just vague enough to scare people, and we don't want to spend time now arguing over the circumstances that would require changes, and what "minimal" means. All we have to say is that the stability policy and its exact text are not in effect for the technical preview.

> Implementers should expect that changes to the specification during
> and following the preview period will minimal and will require compelling evidence.

aphillips:
@macchiati 
> I think we should drop the following;

Fair enough. See follow up PR.

PR #572, opened 2023-12-18
Implement ideographic space

### LINKED ISSUE #498: Consider U+3000 as foldable whitespace?
@eemeli noted in [this comment](https://github.com/unicode-org/message-format-wg/pull/497#discussion_r1363345174) that we might consider U+3000 as ignorable whitespace from a conversation in our 2023-10-16 call. 

> Conversation during this Monday's call suggested we should consider trimming also at least U+3000 when encountered on the pattern exterior.
> 
> This raises the possibility that we could have two separate whitespace rules and categories for the places where it's optional and effectively dropped, and another for where it's required.

This issue tracks the idea.

Note that U+3000 can be produced from CJK IMEs instead of space without the user noticing.

aphillips:
Fixes #498

### DIFF
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -74,7 +74,8 @@ reserved-char  = %x00-08        ; omit HTAB and LF
                / %x0E-19        ; omit SP
                / %x21-5B        ; omit \
                / %x5D-7A        ; omit { | }
-               / %x7E-D7FF      ; omit surrogates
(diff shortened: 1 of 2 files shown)


eemeli (approved):

stasm (approved):

aphillips:
Fast tracking this small change.

PR #574, opened 2023-12-20
Add markup as designed

eemeli:
This implements markup [as designed](https://github.com/unicode-org/message-format-wg/blob/main/exploration/open-close-placeholders.md), including syntax, data model, and formatting changes.

The open/close/value "kind" concept is removed from functions, which now only ever use `:` as a prefix. The `+` is added to `reserved-annotation-start`, but `-` is not.

Resolution and formatting to a string is defined for markup, but formatting to parts is not, given that we don't define parts.

Noting for @gibson042 in particular: In the data model, I did need to add a `type: "expression"` on the expressions, so that they could be differentiated from the `type: "markup"` elements that a pattern may now also hold.

### DIFF
--- a/spec/data-model/README.md
+++ b/spec/data-model/README.md
@@ -110,13 +110,16 @@ interface CatchallKey {
 ## Patterns
 
 Each `Pattern` represents a linear sequence, without selectors.
-Each element of the `body` array MUST either be a non-empty string or an `Expression` or `Markup` object.
-String values represent literal _text_,
-while `Expression` and `Markup` wrap each of the potential _expression_ and _markup_ shapes.
+Each element of the `body` array MUST be a non-empty string, an `Expression` or a `Markup` object.
+String values represent literal _text_.
+`Expression` wraps each of the potential _expression_ shapes.
+`Markup` wraps each of the potential _markup_ shapes.
 The `body` strings are the "cooked" _text_ values, i.e. escape sequences are processed.
 
-Implementations MUST NOT rely on the set of `Expression` and `Markup` interfaces being exhaustive,
-as future versions of this specification MAY define additional expressions.
+Implementations MUST NOT rely on the set of `Expression` and 
+`Markup` interfaces defined in this document being exhaustive.
+Future versions of this specification might define additional
+types of expression or markup.
 
 ```ts
 interface Pattern {
@@ -224,8 +227,9 @@ interface UnsupportedAnnotation {
 
 ## Markup
 
-Each of _markup-open_, _markup-close_, and _markup-standalone_ are represented
-by the corresponding `MarkupOpen`, `MarkupStandalone`, and `MarkupClose`.
+Each of the markup interfaces `MarkupOpen`, `MarkupStandalone`,
+and `MarkupClose` represent the _markup_ `type` of the corresponding
+`kind`.
 The `name` in these does not include the starting sigils `#` and `/`.
 The optional `options` for open and standalone markup use the same `Option`
 as `FunctionAnnotation`.
--- a/spec/data-model/README.md
+++ b/spec/data-model/README.md
@@ -227,9 +227,8 @@ interface UnsupportedAnnotation {
 
 ## Markup
 
-Each of the markup interfaces `MarkupOpen`, `MarkupStandalone`,
-and `MarkupClose` represent the _markup_ `type` of the corresponding
-`kind`.
+A `Markup` object is either `MarkupOpen`, `MarkupStandalone`, or `MarkupClose`,
+which are differentiated by `kind`.
 The `name` in these does not include the starting sigils `#` and `/`.
 The optional `options` for open and standalone markup use the same `Option`
 as `FunctionAnnotation`.
--- a/spec/data-model/README.md
+++ b/spec/data-model/README.md
@@ -110,38 +110,45 @@ interface CatchallKey {
 ## Patterns
 
 Each `Pattern` represents a linear sequence, without selectors.
-Each element of the `body` array MUST either be a non-empty string or an `Expression` object.
+Each element of the `body` array MUST either be a non-empty string or an `Expression` or `Markup` object.
 String values represent literal _text_,
-while `Expression` wraps each of the potential _expression_ shapes.
+while `Expression` and `Markup` wrap each of the potential _expression_ and _markup_ shapes.
 The `body` strings are the "cooked" _text_ values, i.e. escape sequences are processed.
 
-Implementations MUST NOT rely on the set of `Expression` interfaces being exhaustive,
+Implementations MUST NOT rely on the set of `Expression` and `Markup` interfaces being exhaustive,
 as future versions of this specification MAY define additional expressions.
 
 ```ts
 interface Pattern {
-  body: Array<string | Expression>;
+  body: Array<string | Expression | Markup>;
 }
 
-type Expression = LiteralExpression | VariableExpression | FunctionExpression |
-                  UnsupportedExpression;
+type Expression =
+  | LiteralExpression
+  | VariableExpression
+  | FunctionExpression
+  | UnsupportedExpression;
 
 interface LiteralExpression {
+  type: "expression";
   arg: Literal;
   annotation?: FunctionAnnotation | UnsupportedAnnotation;
 }
 
 interface VariableExpression {
+  type: "expression";
   arg: VariableRef;
   annotation?: FunctionAnnotation | UnsupportedAnnotation;
 }
 
 interface FunctionExpression {
+  type: "expression";
   arg?: never;
   annotation: FunctionAnnotation;
 }
 
 interface UnsupportedExpression {
+  type: "expression";
   arg?: never;
   annotation: UnsupportedAnnotation;
 }
@@ -172,17 +179,13 @@ interface VariableRef {
 ```
 
 A `FunctionAnnotation` represents a _function_ _annotation_.
-In a `FunctionAnnotation`,
-the `kind` corresponds to the starting sigil of a _function_:
-`'open'` for `+`, `'close'` for `-`, and `'value'` for `:`.
-The `name` does not include this starting sigil.
+The `name` does not include the `:` starting sigil.
 
 Each _option_ is represented by an `Option`.
 
 ```ts
 interface FunctionAnnotation {
   type: "function";
-  kind: "open" | "close" | "value";
   name: string;
   options?: Option[];
 }
@@ -214,8 +217,40 @@ that the implementation attaches to that _annotation_.
 ```ts
 interface UnsupportedAnnotation {
   type: "unsupported-annotation";
-  sigil: "!" | "@" | "#" | "%" | "^" | "&" | "*" | "<" | ">" | "/" | "?" | "~";
+  sigil: "!" | "@" | "%" | "^" | "&" | "*" | "+" | "<" | ">" | "?" | "~";
   source: string;
+}
+```
+
+## Markup
+
+Each of _markup-open_, _markup-close_, and _markup-standalone_ are represented
+by the corresponding `MarkupOpen`, `MarkupStandalone`, and `MarkupClose`.
+The `name` in these does not include the starting sigils `#` and `/`.
+The optional `options` for open and standalone markup use the same `Option`
+as `FunctionAnnotation`.
+
+```ts
+type Markup = MarkupOpen | MarkupStandalone | MarkupClose;
+
+interface MarkupOpen {
+  type: "markup";
+  kind: "open";
+  name: string;
+  options?: Option[];
+}
+
+interface MarkupStandalone {
+  type: "markup";
+  kind: "standalone";
+  name: string;
+  options?: Option[];
+}
+
+interface MarkupClose {
+  type: "markup";
+  kind: "close";
+  name: string;
 }
 ```
 
--- a/spec/data-model/README.md
+++ b/spec/data-model/README.md
@@ -110,7 +110,7 @@ interface CatchallKey {
 ## Patterns
 
 Each `Pattern` represents a linear sequence, without selectors.
-Each element of the `body` array MUST be a non-empty string, an `Expression` or a `Markup` object.
+Each element of the `body` array MUST be a non-empty string, an `Expression`, or a `Markup` object.
 String values represent literal _text_.
 `Expression` wraps each of the potential _expression_ shapes.
 `Markup` wraps each of the potential _markup_ shapes.
@@ -119,7 +119,7 @@ The `body` strings are the "cooked" _text_ values, i.e. escape sequences are pro
 Implementations MUST NOT rely on the set of `Expression` and 
 `Markup` interfaces defined in this document being exhaustive.
 Future versions of this specification might define additional
-types of expression or markup.
+expressions or markup.
 
 ```ts
 interface Pattern {
@@ -229,7 +229,8 @@ interface UnsupportedAnnotation {
 
 A `Markup` object is either `MarkupOpen`, `MarkupStandalone`, or `MarkupClose`,
 which are differentiated by `kind`.
-The `name` in these does not include the starting sigils `#` and `/`.
+The `name` in these does not include the starting sigils `#` and `/` 
+or the ending sigil `/`.
 The optional `options` for open and standalone markup use the same `Option`
 as `FunctionAnnotation`.
 
--- a/spec/data-model/message.json
+++ b/spec/data-model/message.json
@@ -113,7 +113,17 @@
       "type": "object",
       "properties": {
         "type": { "const": "markup" },
-        "kind": { "enum": ["open", "standalone"] },
+        "kind": { "const": "open" },
+        "name": { "type": "string" },
+        "options": { "$ref": "#/$defs/options" }
+      },
+      "required": ["type", "kind", "name"]
+    },
+    "markup-standalone": {
+      "type": "object",
+      "properties": {
+        "type": { "const": "markup" },
+        "kind": { "const": "standalone" },
         "name": { "type": "string" },
         "options": { "$ref": "#/$defs/options" }
       },
@@ -131,6 +141,7 @@
     "markup": {
       "oneOf": [
         { "$ref": "#/$defs/markup-open" },
+        { "$ref": "#/$defs/markup-standalone" },
         { "$ref": "#/$defs/markup-close" }
       ]
     },
--- a/spec/data-model/message.json
+++ b/spec/data-model/message.json
@@ -22,27 +22,29 @@
       },
       "required": ["type", "name"]
     },
-    "option": {
-      "type": "object",
-      "properties": {
-        "name": { "type": "string" },
-        "value": {
-          "oneOf": [{ "$ref": "#/$defs/literal" }, { "$ref": "#/$defs/variable" }]
-        }
-      },
-      "required": ["name", "value"]
+    "options": {
+      "type": "array",
+      "items": {
+        "type": "object",
+        "properties": {
+          "name": { "type": "string" },
+          "value": {
+            "oneOf": [
+              { "$ref": "#/$defs/literal" },
+              { "$ref": "#/$defs/variable" }
+            ]
+          }
+        },
+        "required": ["name", "value"]
+      }
     },
 
     "function-annotation": {
       "type": "object",
       "properties": {
         "type": { "const": "function" },
-        "kind": { "enum": ["open", "close", "value"] },
         "name": { "type": "string" },
-        "options": {
-          "type": "array",
-          "items": { "$ref": "#/$defs/option" }
-        }
+        "options": { "$ref": "#/$defs/options" }
       },
       "required": ["type", "kind", "name"]
     },
@@ -67,32 +69,36 @@
     "literal-expression": {
       "type": "object",
       "properties": {
+        "type": { "const": "expression" },
         "arg": { "$ref": "#/$defs/literal" },
         "annotation": { "$ref": "#/$defs/annotation" }
       },
-      "required": ["arg"]
+      "required": ["type", "arg"]
     },
     "variable-expression": {
       "type": "object",
       "properties": {
+        "type": { "const": "expression" },
         "arg": { "$ref": "#/$defs/variable" },
         "annotation": { "$ref": "#/$defs/annotation" }
       },
-      "required": ["arg"]
+      "required": ["type", "arg"]
     },
     "function-expression": {
       "type": "object",
       "properties": {
+        "type": { "const": "expression" },
         "annotation": { "$ref": "#/$defs/function-annotation" }
       },
-      "required": ["annotation"]
+      "required": ["type", "annotation"]
     },
     "unsupported-expression": {
       "type": "object",
       "properties": {
+        "type": { "const": "expression" },
         "annotation": { "$ref": "#/$defs/unsupported-annotation" }
       },
-      "required": ["annotation"]
+      "required": ["type", "annotation"]
     },
     "expression": {
       "oneOf": [
@@ -103,13 +109,43 @@
       ]
     },
 
+    "markup-open": {
+      "type": "object",
+      "properties": {
+        "type": { "const": "markup" },
+        "kind": { "enum": ["open", "standalone"] },
+        "name": { "type": "string" },
+        "options": { "$ref": "#/$defs/options" }
+      },
+      "required": ["type", "kind", "name"]
+    },
+    "markup-close": {
+      "type": "object",
+      "properties": {
+        "type": { "const": "markup" },
+        "kind": { "const": "close" },
+        "name": { "type": "string" }
+      },
+      "required": ["type", "kind", "name"]
+    },
+    "markup": {
+      "oneOf": [
+        { "$ref": "#/$defs/markup-open" },
+        { "$ref": "#/$defs/markup-close" }
+      ]
+    },
+
     "pattern": {
       "type": "object",
       "properties": {
         "body": {
           "type": "array",
           "items": {
-            "oneOf": [{ "type": "string" }, { "$ref": "#/$defs/expression" }]
+            "oneOf": [
+              { "type": "string" },
+              { "$ref": "#/$defs/expression" },
+              { "$ref": "#/$defs/markup" }
+            ]
           }
         }
       },
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -12,7 +12,7 @@ their handling during formatting is specified here as well.
 
 Formatting of a _message_ is defined by the following operations:
 
-- **_Expression Resolution_** determines the value of an _expression_,
+- **_Expression and Markup Resolution_** determines the value of an _expression_ or _markup_,
   with reference to the current _formatting context_.
   This can include multiple steps,
   such as looking up the value of a variable and calling formatting functions.
@@ -83,9 +83,10 @@ At a minimum, it includes:
 
 Implementations MAY include additional fields in their _formatting context_.
 
-## Expression Resolution
+## Expression and Markup Resolution
 
 _Expressions_ are used in _declarations_, _selectors_, and _patterns_.
+_Markup_ is only used in _patterns_.
 
 In a _declaration_, the resolved value of the _expression_ is bound to a _variable_,
 which is available for use by later _expressions_.
@@ -98,7 +99,7 @@ but also the _variable_ to which the resolved value of the _variable-expression_
 
 In _selectors_, the resolved value of an _expression_ is used for _pattern selection_.
 
-In a _pattern_, the resolved value of an _expression_ is used in its _formatting_.
+In a _pattern_, the resolved value of _expressions_ and _markup_ is used in their _formatting_.
 
 The shapes of resolved values are implementation-dependent,
 and different implementations MAY choose to perform different levels of resolution.
@@ -210,17 +211,7 @@ the following steps are taken:
    Implementations are not required to implement _namespaces_ or installable
    _function registries_.
 
-3. Resolve the _options_ to a mapping of string identifiers to values.
-   If _options_ is missing, the mapping will be empty.
-   For each _option_:
-   - Resolve the _identifier_ of the _option_.
-   - If the _option_'s _identifier_ already exists in the resolved mapping of _options_,
-     emit a Duplicate Option Name error.
-   - If the _option_'s right-hand side successfully resolves to a value,
-     bind the _identifier_ of the _option_ to the resolved value in the mapping.
-   - Otherwise, bind the _identifier_ of the _option_ to an unresolved value in the mapping.
-     Implementations MAY later remove this value before calling the _function_.
-     (Note that an Unresolved Variable error will have been emitted.)
+3. Perform _option resolution_.
 
 4. Call the function implementation with the following arguments:
 
@@ -247,6 +238,37 @@ the following steps are taken:
    If the call fails or does not return a valid value,
    emit a Resolution error and use a _fallback value_ for the _expression_.
 
+#### Option Resolution
+
+The result of resolving _option_ values is a mapping of string identifiers to values.
+
+For each _option_:
+
+- Resolve the _identifier_ of the _option_.
+- If the _option_'s _identifier_ already exists in the resolved mapping of _options_,
+   emit a Duplicate Option Name error.
+- If the _option_'s right-hand side successfully resolves to a value,
+   bind the _identifier_ of the _option_ to the resolved value in the mapping.
+- Otherwise, bind the _identifier_ of the _option_ to an unresolved value in the mapping.
+   Implementations MAY later remove this value before calling the _function_.
+   (Note that an Unresolved Variable error will have been emitted.)
+
+Errors MAY be emitted during _option resolution_,
+but it always resolves to some mapping of string identifiers to values.
+This mapping can be empty.
+
+### Markup Resolution
+
+Unlike _functions_, the resolution of _markup_ is not customizable.
+
+The resolved value of _markup_ includes the following fields:
+
+- The type of the markup: open, standalone, or close
+- The _name_ of the _markup_
+- For _markup-open_ and _markup_standalone_,
+  the resolved _options_ values after _option resolution_.
+
+The resolution of _markup_ MUST always succeed.
 
 ### Fallback Resolution
 
@@ -304,7 +326,7 @@ The _fallback value_ depends on the contents of the _expression_:
   > the message formats to `{$arg}`.
 
 - _function_ _expression_ with no _operand_:
-  the _function_ starting sigil followed by its _identifier_
+  U+003A COLON `:` followed by the _function_ _identifier_
 
   > Examples:
   > In a context where `:func` fails to resolve, `{:func}` resolves to the _fallback value_ `:func`.
@@ -599,7 +621,7 @@ one {{Category match}}
 ## Formatting
 
 After _pattern selection_,
-each _text_ and _expression_ part of the selected _pattern_ is resolved and formatted.
+each _text_ and _placeholder_ part of the selected _pattern_ is resolved and formatted.
 
 _Formatting_ is a mostly implementation-defined process,
 as it depends on the implementation's shape for resolved values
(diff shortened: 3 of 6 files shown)


macchiati (commented):
> Markup-open and markup-close parts SHOULD be paired, but this is not a requirement. A markup-open MAY be present in a message without a corresponding markup-close, and vice versa. Markup-standalone parts are not paired.

There is no mention of whether markup is forced to nest or not. It seems like we should have a statement about it, rather than leaving the question open in reader's minds. 

Maybe it would be better to state that whether markup-open and markup-close are paired and/or nested depends on the markup standard that they are following, not this specification.

aphillips:
@macchiati 
> There is no mention of whether markup is forced to nest or not. It seems like we should have a statement about it, rather than leaving the question open in reader's minds.

I agree that we should spell this out. Note that the design document does cover it: markup is neither actually paired nor does it require any specific nesting.

spec/data-model/README.md
```diff
@@ -110,38 +110,45 @@ interface CatchallKey {
...
-Each element of the `body` array MUST either be a non-empty string or an `Expression` object.
+Each element of the `body` array MUST either be a non-empty string or an `Expression` or `Markup` object.
 String values represent literal _text_,
-while `Expression` wraps each of the potential _expression_ shapes.
+while `Expression` and `Markup` wrap each of the potential _expression_ and _markup_ shapes.
 The `body` strings are the "cooked" _text_ values, i.e. escape sequences are processed.
 
-Implementations MUST NOT rely on the set of `Expression` interfaces being exhaustive,
+Implementations MUST NOT rely on the set of `Expression` and `Markup` interfaces being exhaustive,
 as future versions of this specification MAY define additional expressions.
```
aphillips, line 119: I think the term "expression" is overloaded here. Also, the second normative word shouldn't be normative (it's permitting a future version of this specification to do something). Perhaps:

```suggestion
Implementations MUST NOT rely on the set of `Expression` and 
`Markup` interfaces defined in this document being exhaustive.
Future versions of this specification might define additional
types of expression or markup.
```

spec/data-model/README.md
```diff
@@ -172,17 +179,13 @@ interface VariableRef {
 ```
 
 A `FunctionAnnotation` represents a _function_ _annotation_.
-In a `FunctionAnnotation`,
-the `kind` corresponds to the starting sigil of a _function_:
-`'open'` for `+`, `'close'` for `-`, and `'value'` for `:`.
-The `name` does not include this starting sigil.
+The `name` does not include the `:` starting sigil.
```
aphillips, line 182: I think there's a problem here. We're using the word `name` (here and in the interface definitions above and below). This is fine for `VariableRef`, but not for `FunctionAnnotation`, which should use `identifier` (that is, including the namespace)

spec/data-model/README.md
```diff
@@ -214,11 +217,43 @@ that the implementation attaches to that _annotation_.
...
+  sigil: "!" | "@" | "%" | "^" | "&" | "*" | "+" | "<" | ">" | "?" | "~";
   source: string;
 }
 ```
 
+## Markup
+
+Each of _markup-open_, _markup-close_, and _markup-standalone_ are represented
+by the corresponding `MarkupOpen`, `MarkupStandalone`, and `MarkupClose`.
+The `name` in these does not include the starting sigils `#` and `/`.
```
aphillips, line 229: `name` => `identifier` in this section

spec/data-model/README.md
```diff
@@ -214,11 +217,43 @@ that the implementation attaches to that _annotation_.
...
-  sigil: "!" | "@" | "#" | "%" | "^" | "&" | "*" | "<" | ">" | "/" | "?" | "~";
+  sigil: "!" | "@" | "%" | "^" | "&" | "*" | "+" | "<" | ">" | "?" | "~";
   source: string;
 }
 ```
 
+## Markup
+
+Each of _markup-open_, _markup-close_, and _markup-standalone_ are represented
+by the corresponding `MarkupOpen`, `MarkupStandalone`, and `MarkupClose`.
```
aphillips, line 228: This reads oddly (both in terms of English and in terms of stylistically). The other sections start with the interface names, while this starts with what is represented. Perhaps:

```suggestion
Each of the markup interfaces `MarkupOpen`, `MarkupStandalone`,
and `MarkupClose` represent the _markup_ `type` of the corresponding
`kind`.
```

spec/data-model/message.json
```diff
@@ -103,13 +109,43 @@
...
+    "markup-open": {
+      "type": "object",
+      "properties": {
+        "type": { "const": "markup" },
+        "kind": { "enum": ["open", "standalone"] },
+        "name": { "type": "string" },
+        "options": { "$ref": "#/$defs/options" }
+      },
+      "required": ["type", "kind", "name"]
+    },
```
aphillips, line 121: why isn't this two objects, `markup-open` and `markup-standalone`, like the data model has?

See previous comment about `name` vs. `identifier`

spec/formatting.md
```diff
@@ -247,6 +238,37 @@ the following steps are taken:
...
+This mapping can be empty.
+
+### Markup Resolution
+
+Unlike _functions_, the resolution of _markup_ is not customizable.
+
+The resolved value of _markup_ includes the following fields:
+
+- The type of the markup: open, standalone, or close
+- The _name_ of the _markup_
```
aphillips, line 267: identifier?

spec/formatting.md
```diff
@@ -247,6 +238,37 @@ the following steps are taken:
...
+   (Note that an Unresolved Variable error will have been emitted.)
+
+Errors MAY be emitted during _option resolution_,
+but it always resolves to some mapping of string identifiers to values.
+This mapping can be empty.
+
+### Markup Resolution
+
+Unlike _functions_, the resolution of _markup_ is not customizable.
+
```
aphillips, line 263: This might be too opinionated? If I'm understanding this correctly, this doesn't permit implementations to support plug-in markup handlers that work similar to functions. While that might be the right solution for some implementations, it might be be what everyone expects. 

We do want, I think, to say that markup's fallback is to nothing?

spec/formatting.md
```diff
@@ -618,13 +640,18 @@ appropriate data type or structure. Some examples of these include:
...
 
 Implementations SHOULD provide _formatting_ result types that match user needs,
 including situations that require further processing of formatted messages.
 Implementations SHOULD encourage users to consider a formatted localised string
 as an opaque data structure, suitable only for presentation.
 
+When formatting to a string, the default representation of all _markup_
+MUST be an empty string.
+Implementations MAY offer functionality for customizing this,
+such as by emitting XML-ish tags for each _markup_.
```
aphillips, line 653: I would tighten this up a bit. The normative text is appropriate, but I would avoid the MUST. I would also make the example a bit more fleshed out.

```suggestion
The default representation of all types of _markup_ placeholder
is an empty string.
That is, unless otherwise customized, open, close and standalone
markup placeholders are omitted from the output.
Implementations MAY offer functionality to customize this behavior.

> For example, an implementation might replace markup with
> styling attributes in an attributed string API 
> or it might replace the placeholders with HTML markup.
```

spec/syntax.md
```diff
@@ -509,37 +509,19 @@ what _options_ and _option_ values are valid,
 and what outputs might result.
 See [function registry](./) for more information.
 
-_Functions_ can be _standalone_, or can be an _opening element_ or _closing element_.
-
-A **_<dfn>standalone</dfn>_** _function_ is not expected to be paired with another _function_.
-An **_<dfn>opening element</dfn>_** is a _function_ that SHOULD be paired with a _closing element_.
-A **_<dfn>closing element</dfn>_** is a _function_ that SHOULD be paired with an _opening element_.
+A _function_ starts with a prefix sigil `:` followed by an _identifier_.
+The _name_ MAY be followed by one or more _options_.
```
aphillips, line 513: name => identifier

spec/syntax.md
```diff
@@ -669,6 +651,45 @@ reserved-char = %x00-08        ; omit HTAB and LF
               / %xE000-10FFFF
 ```
 
+## Markup
+
+**_<dfn>Markup</dfn>_** _placeholders_ are _pattern_ parts
+that can be used to represent non-language parts of a _message_,
+such as inline elements or styling that should apply to a span of parts.
```
aphillips, line 658: I'm not sure about the language here. Perhaps:

```suggestion
**_<dfn>Markup</dfn>_** are _placeholders_ in a _pattern_ that
represent externally defined, non-formattable parts of a _message_.
Unless otherwise provisioned, _markup_ is omitted from the formatted _message_.
Examples of markup might include inline styling or translator
instructions to be applied to a span of _message_ parts.
```

spec/syntax.md
```diff
@@ -669,6 +651,45 @@ reserved-char = %x00-08        ; omit HTAB and LF
...
+Markup comes in three forms:
+
+**_<dfn>Markup-open</dfn>_** starts with `#` and
+is a _pattern_ part starting a span.
+It MAY include _options_, but these are not required.
+
+If a _markup-open_ includes a `/` immediately before its closing `}`,
+it is a self-closing **_<dfn>markup-standalone</dfn>_** part
+representing a standalone element within a message,
+such as an inline image or any other opaque content.
```
aphillips, line 668: This should be the same as the other two types.

```suggestion
**_<dfn>Markup-standalone</dfn>_** starts with U+0023 NUMBER SIGN `#`
and has a U+002F SOLIDUS `/` immediately before its closing `}`
representing a self-closing or standalone element within the _message_.
It MAY include _options_, but these are not required.
```

spec/syntax.md
```diff
@@ -669,6 +651,45 @@ reserved-char = %x00-08        ; omit HTAB and LF
...
 ```
 
+## Markup
+
+**_<dfn>Markup</dfn>_** _placeholders_ are _pattern_ parts
+that can be used to represent non-language parts of a _message_,
+such as inline elements or styling that should apply to a span of parts.
+Markup comes in three forms:
+
+**_<dfn>Markup-open</dfn>_** starts with `#` and
```
aphillips, line 661: Let's start to be fastidious about character references

```suggestion
**_<dfn>Markup-open</dfn>_** starts with U+0023 NUMBER SIGN `#` and
```

spec/syntax.md
```diff
@@ -669,6 +651,45 @@ reserved-char = %x00-08        ; omit HTAB and LF
...
+is a _pattern_ part starting a span.
```
aphillips, line 662: This is slightly misleading, as we don't guarantee closure.

```suggestion
represents an opening element within the _message_,
such as markup used to start a span.
```

spec/syntax.md
```diff
@@ -669,6 +651,45 @@ reserved-char = %x00-08        ; omit HTAB and LF
...
+It MAY include _options_, but these are not required.
+
+If a _markup-open_ includes a `/` immediately before its closing `}`,
+it is a self-closing **_<dfn>markup-standalone</dfn>_** part
+representing a standalone element within a message,
+such as an inline image or any other opaque content.
+
+**_<dfn>Markup-close</dfn>_** starts with `/` and
```
aphillips, line 670:
```suggestion
**_<dfn>Markup-close</dfn>_** starts with U+002F SOLIDUS `/` and
```

aphillips (changes requested):
A more thorough review. Lots of nitty-gritty details.

spec/formatting.md
```diff
@@ -12,7 +12,7 @@ their handling during formatting is specified here as well.
 
 Formatting of a _message_ is defined by the following operations:
 
-- **_Expression Resolution_** determines the value of an _expression_,
+- **_Expression and Markup Resolution_** determines the value of an _expression_ or _markup_,
```
stasm, line 15: I think we should make *markup resolution* a separate bullet in this list. Resolving markup is a very different operations from resolving expressions due to markup's not calling any functions, due to the default ignorable property in `formatToString`, and due to the fact that it can only happen inside patterns.

In fact, I'm not even sure if we should call it *resolution*. I guess we still resolve options, but crucially, in its current design, markup is encoded in the data model rather directly.

spec/message.abnf
```diff
@@ -23,13 +23,18 @@ expression = literal-expression / variable-expression / annotation-expression
...
 
+markup       = "{" [s] markup-open [s] ["/"] "}"
+             / "{" [s] markup-close [s] "}"
+markup-open  = "#" identifier *(s option)
+markup-close = "/" identifier
+
 literal = quoted / unquoted
 variable = "$" name
-function = (":" / "+" / "-") identifier
+function = ":" identifier *(s option)
```
stasm, line 37: Please move `*(s option)` back to the `annotation` production. It's helpful that `variable` and `function` define *names*. In fact, we should consider renaming them to `veriable-name` and `function-name`, respectively.

spec/message.abnf
```diff
@@ -3,7 +3,7 @@ message = simple-message / complex-message
 simple-message = [simple-start pattern]
 simple-start = simple-start-char / text-escape / placeholder
 pattern = *(text-char / text-escape / placeholder)
-placeholder = expression
+placeholder = expression / markup
```
stasm, line 6: I realize this is cosmetic (or academic?), but I think I was assuming we'd instead do:

```abnf
pattern = *(text-char / text-escape / placeholder / markup)
placeholder = expression
```

That is, that markup won't end up being placeholders, but rather a new type of pattern elements, similar to how you implemented it in the data model. I presume that ABNF will have the most impact on shaping MF2's vocabulary, which is why I think we should be careful here.

Thoughts?

spec/syntax.md
```diff
@@ -669,6 +651,45 @@ reserved-char = %x00-08        ; omit HTAB and LF
...
+## Markup
+
+**_<dfn>Markup</dfn>_** _placeholders_ are _pattern_ parts
+that can be used to represent non-language parts of a _message_,
+such as inline elements or styling that should apply to a span of parts.
+Markup comes in three forms:
+
+**_<dfn>Markup-open</dfn>_** starts with `#` and
+is a _pattern_ part starting a span.
+It MAY include _options_, but these are not required.
```
stasm, line 663: The `but...` is not needed here?

```suggestion
It MAY include _options_.
```

spec/syntax.md
```diff
@@ -669,6 +651,45 @@ reserved-char = %x00-08        ; omit HTAB and LF
...
+markup-close = "/" identifier
+```
+
+> A _message_ with one `button` markup span and a standalone `img` markup element.
+>
+> ```
+> {#button}Submit{/button} or {#img alt=|Cancel| /}.}
+> ```
+
+_Markup-open_ and _markup-close_ parts SHOULD be paired,
```
stasm, line 687: Isn't that more of a `MAY`? There exist completely valid systems where markup is stack-based (e.g. TTY) and where I could imagine only `markup-open` being used with success.

stasm (changes requested):

spec/formatting.md
eemeli: Your suggestion leaves out the "When formatting to a string" qualifier. Is that intentional?

spec/syntax.md
eemeli: This seems to presume that "formatting" is a process that always produces a string. I don't think that's an accurate portrayal for this spec, which is supporting non-string formatting targets even if it does not explicitly define them.

spec/syntax.md
eemeli:
```suggestion
_Markup-open_ and _markup-close_ parts MAY be paired,
```

spec/data-model/README.md
```diff
@@ -110,38 +110,45 @@ interface CatchallKey {
 ## Patterns
 
 Each `Pattern` represents a linear sequence, without selectors.
-Each element of the `body` array MUST either be a non-empty string or an `Expression` object.
+Each element of the `body` array MUST either be a non-empty string or an `Expression` or `Markup` object.
 String values represent literal _text_,
-while `Expression` wraps each of the potential _expression_ shapes.
+while `Expression` and `Markup` wrap each of the potential _expression_ and _markup_ shapes.
```
eemeli, line 115:
```suggestion
`Expression` wraps each of the potential _expression_ shapes.
`Markup` wraps each of the potential _markup_ shapes.
```

spec/data-model/README.md
```diff
@@ -110,38 +110,45 @@ interface CatchallKey {
 ## Patterns
 
 Each `Pattern` represents a linear sequence, without selectors.
-Each element of the `body` array MUST either be a non-empty string or an `Expression` object.
+Each element of the `body` array MUST either be a non-empty string or an `Expression` or `Markup` object.
 String values represent literal _text_,
```
eemeli, line 114:
```suggestion
String values represent literal _text_.
```

spec/data-model/README.md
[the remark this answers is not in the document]
eemeli: Applied as two separate suggestions, to work around GitHub limitations.

spec/data-model/message.json
eemeli: I merged them here because for the JSON Schema definition it's an internal detail that doesn't show up in user data, and the two shapes are otherwise exactly the same.

Would you be more comfortable if the whole object was duplicated?

spec/formatting.md
eemeli: Note that this statement is specifically about _resolution_, and does not limit what may be done afterwards. A "plug-in markup handler" could post-process resolved values into different shapes, and it could even have access to the whole span of parts between open & close elements.

spec/formatting.md
eemeli: > I think we should make _markup resolution_ a separate bullet in this list. Resolving markup is a very different operations from resolving expressions due to markup's not calling any functions, due to the default ignorable property in `formatToString`, and due to the fact that it can only happen inside patterns.

I would prefer to keep these together, because the similarities are much greater than the differences.
- Markup resolution does not call external functions, but its results may be defined in the same way as for expressions. We also have expressions such as literals and variables which do not necessarily call any functions.
- Markup resolution uses the same option resolution as functions.
- The behaviour of markup during string formatting is a concern during _formatting_, but not _resolution_.
- Markup resolution only being used for pattern placeholders does not change how it is done.

> In fact, I'm not even sure if we should call it _resolution_. I guess we still resolve options, but crucially, in its current design, markup is encoded in the data model rather directly.

Could you clarify what you mean by "data model" here? The only data model we define in detail is for the message before formatting, from which we need to select, resolve, and format a pattern during formatting. We left out formatted parts, so the output structure is not well defined.

spec/message.abnf
eemeli: This part of the ABNF was explicitly included in the design doc, and so I'd really rather not change that here. I also think that having a catch-all name for "stuff in a pattern that has curly braces" is useful, and that "placeholder" works for that.

I would strongly prefer discussing this change as a separate follow-up to this PR.

spec/message.abnf
eemeli: On the other hand, isn't it also useful for a _function_ to have _options_? Previously, they were only a part of the _annotation_, which could effectively have any shape as _private-use_ and _reserved annotations_.

spec/syntax.md
eemeli:
```suggestion
The _identifier_ MAY be followed by one or more _options_.
```

spec/formatting.md
eemeli:
```suggestion
- The _identifier_ of the _markup_
```

spec/syntax.md
```diff
@@ -669,6 +651,46 @@ reserved-char = %x00-08        ; omit HTAB and LF
...
+>
+> ```
+> {#button}Submit{/button} or {#img alt=|Cancel| /}.}
+> ```
+
+_Markup-open_ and _markup-close_ parts MAY be paired,
+but this is not a requirement.
+A _markup-open_ MAY be present in a message without
+a corresponding _markup-close_, and vice versa.
+_Markup-standalone_ parts are not paired.
```
eemeli, line 692: @macchiati @aphillips Would this work?

```suggestion
No validity requirements are imposed for the proper pairing or nesting of _markup_.
A _markup-open_ MAY be present in a message without
a corresponding _markup-close_, and vice versa.
```

spec/data-model/README.md
eemeli: I've spun this off into a separate issue #575, as it's a separate concern from the markup implementation.

spec/data-model/README.md
eemeli: See #575.

spec/data-model/message.json
aphillips: Yes: the open and standalone items are meant to be semantically different. And they wouldn't be exactly duplicated. One would have `kind: open` and the other `kind: standalone`

spec/syntax.md
aphillips: No, that's a jump. It doesn't say anything about string resolution--it just says that the markup can be omitted ("unless otherwise provisioned"). It doesn't say what form any of that takes. We could expand on the examples, I suppose.

spec/formatting.md
aphillips: My first reaction to the draft was like @stasm's: it's okay to be somewhat repetitious in a spec. Keeping a wall between the two similar-but-unlike concepts might be useful. 

I didn't make that comment in the end, because there is a lot of commonality and the resolution into different buckets does happen later.

> We also have expressions such as literals and variables which do not necessarily call any functions.

Are you sure? If there is a placeholder `{|zebra|}`, I'm pretty sure that invokes the `:string` formatting function. If there is a placeholder `{$foo}`, I'd allow the implementation to introspect `foo`'s type, but I would expect *some* function to be called in the end. We don't want it to be an error to write messages like `Hello {$user}`...

spec/syntax.md
aphillips: I noodled with this for a while. I think we want to allow markup regimes to impose requirements while saying that the message is otherwise "valid".

```suggestion
A _markup-open_ can appear without a corresponding _markup-close_.
A _markup-close_ can appear without a corresponding _markup-open_.
_Markup_ _placeholders_ can appear in any order without making the _message_ invalid.
However, specifications or implementations defining _markup_ might impose requirements
on the pairing, ordering, or contents of _markup_ during _formatting_.
```

spec/data-model/message.json
eemeli: Added.

spec/syntax.md
eemeli: It's still referring to markup as "non-formattable", and states that "_markup_ is omitted [unless] otherwise provisioned". Our design doc, on the other hand, includes this: https://github.com/unicode-org/message-format-wg/blob/2a69d2a17ee7476282bfa824fd4ca62e0a03f6e1/exploration/open-close-placeholders.md?plain=1#L253-L259

So they are formattable, and are not omitted by default _except when formatting to a string_.

Also, we did end up includes standalone, for which placeholders like `{#img/}` or `{#button/}` are rather likely to end up as concrete parts of the output

aphillips: You're right, but then... how are `img` or `button` different from `b`?

I think what we might be trying to say is:

```suggestion
**_<dfn>Markup</dfn>_** is an implementation-specific
(but otherwise unsupported) mechanism intended to provide for
the inclusion of styling, annotation, or various markup syntaxes
in a _message_ in ways that are compatible with translation.
A _markup_ _placeholder_ can only appear in a _pattern_
and might be omitted or remain unprocessed during the formatting process.
```

eemeli: > You're right, but then... how are `img` or `button` different from `b`?

They don't have a span of parts within the message that they're wrapping. They explicitly enable the representation of inline elements that should be called out in the explanation of _markup_.

> **_Markup_** is an implementation-specific (but otherwise unsupported) mechanism [...]

What does this mean, and why does it need to be called out as the first thing about markup? How is markup "unsupported"?

My understanding of the MF2 position on _markup_ is that we support syntax for it, and define how to represent it in non-string output. I don't really see how it could be more "supported" than that? Defining the output more precisely would've been much easier if we could have included formatted parts in the spec, but alas, that was not to be.

Maybe @mihnita already has some idea how the ICU4J implementation will handle markup, to give us another point of view?

macchiati:
I think it is "unsupported" in the sense that MF itself attaches no
specific meaning to any of the markup identifiers; it's just a 'pass
through' mechanism.

spec/syntax.md
aphillips: I wrote the above hurriedly yesterday, regretting "unsupported" even as I suggested it. I don't think it's necessarily "wrong" though. 

The problem with markup is that we don't provide real support for it. It doesn't use functions per-se and we don't provide anything in the registry for it. Open/close don't mean anything, other than as semantic constructs--we don't require them to be in order or paired.

What we're doing is reserving syntax for a partially specified feature. I think there are three ways that implementations will respond to this: (i) they won't support markup; (ii) they'll provide "pass through" support for it using the data model and/or some other means; or (iii) they'll implement either specific markup support (such as for HTML) or some form of pluggable (generic) support.

All implementations can check the syntax of a placeholder and can resolve any variables. Therefore all implementations MUST emit a syntax error for a malformed placeholder and MUST emit a resolution error if they encounter one.

Let's consider an implementation that supports markup. I think we should provide support for implementers who want to put additional requirements on (specific) markup, such as requiring open/close pairing, requiring open/close to be in order, require specific options (or validate option values). Users will appreciate it if a typo in their markup produces an appropriate error (@stasm has cited this in the past). This suggests that we provide an *optional* MarkupError that implementations MAY emit.

What about pass through? I think this is what we currently describe, albeit we require `formatToString` to remove the markup (pass through is to the data model or parts. Here we might want to relax the requirement to allow implementations to decide how to pass markup through (callers might need to reparse the output as a result).

What about non-supporting implementations? They might still emit data model parts or they might eat them. I think we might want to require that they not fail on markup but permit them to not emit the markup.

A key consideration is not to allow markup to become a security attack vector. We don't want `<scr{#bad}ipt>` to become a script tag (or other abusive sequence) in ways that are hard to detect. This can happen if the markup doesn't collapse to nothing until late.

The last two considered together might look like:

- An implementation SHOULD NOT emit the markup placeholder when formatting to string.
- An implementation MAY (SHOULD?) emit unrecognized markup placeholders in the data model (i.e. they do something other than formatToString)
- An implementation SHOULD NOT apply any additional validation (beyond syntax and resolution) to unrecognized or unsupported markup.
- An implementation MAY emit a MarkupError for unrecognized or unsupported markup. (this has to be allowed if implementations that support markup are to identify typos)

spec/data-model/README.md
```diff
@@ -214,11 +220,44 @@ that the implementation attaches to that _annotation_.
...
+  sigil: "!" | "@" | "%" | "^" | "&" | "*" | "+" | "<" | ">" | "?" | "~";
   source: string;
 }
 ```
 
+## Markup
+
+Each of the markup interfaces `MarkupOpen`, `MarkupStandalone`,
+and `MarkupClose` represent the _markup_ `type` of the corresponding
+`kind`.
```
gibson042, line 232: I don't know how to read this sentence... what is it trying to communicate? Perhaps something like this?
```suggestion
A `Markup` object is either `MarkupOpen`, `MarkupStandalone`, or `MarkupClose`,
which are differentiated by `kind`.
```

spec/data-model/message.json
```diff
@@ -22,27 +22,29 @@
...
-    "option": {
-      "type": "object",
-      "properties": {
-        "name": { "type": "string" },
-        "value": {
-          "oneOf": [{ "$ref": "#/$defs/literal" }, { "$ref": "#/$defs/variable" }]
-        }
-      },
-      "required": ["name", "value"]
+    "options": {
```
gibson042, line 25: This works, but aesthetically I'd prefer keeping the shallower approach and just duplicating ```"options": { "type": "array", "items": { "$ref": "#/$defs/option" } } ``` (it's nice to have an easily-referenced "option" type).

spec/formatting.md
```diff
@@ -247,6 +238,37 @@ the following steps are taken:
...
+Unlike _functions_, the resolution of _markup_ is not customizable.
+
+The resolved value of _markup_ includes the following fields:
+
+- The type of the markup: open, standalone, or close
+- The _identifier_ of the _markup_
+- For _markup-open_ and _markup_standalone_,
+  the resolved _options_ values after _option resolution_.
+
+The resolution of _markup_ MUST always succeed.
```
gibson042, line 271: This doesn't look quite right for a use of RFC 2119 "MUST".

spec/formatting.md
eemeli: > If there is a placeholder `{|zebra|}`, I'm pretty sure that invokes the `:string` formatting function. If there is a placeholder `{$foo}`, I'd allow the implementation to introspect `foo`'s type, but I would expect _some_ function to be called in the end.

I agree that this is a reasonable thing for an implementation to do, but I don't think we should expect it to _always_ happen. There are also reasonable further questions, such as: What formatter is called for `{|zebra|}` if an implementation allows for overriding default formatters, and `:string` is so overridden? Is it the core `:string` formatter, or the user-provided `:string` formatter that handles the zebra?

I think that's a question we ought to _not_ resolve in this spec, leaving it instead to implementations.

spec/formatting.md
eemeli: The usage here is intentional, at least on my part: An implementation must not allow for the resolution of any `{#foo}` part to ever fail.

spec/data-model/message.json
eemeli: Could you clarify what use case you have in mind for an externally-referrable JSON schema option type?

In my experience (and according to the [reference docs](https://json-schema.org/understanding-json-schema/structuring#defs)), these `$defs` are meant only for internal use within this schema, and that only the top-level entity here should be considered as an external API.

spec/formatting.md
aphillips: > I think that's a question we ought to not resolve in this spec, leaving it instead to implementations.

I agree. Hmm... what concrete change are we proposing with this thread?

spec/data-model/README.md
```diff
@@ -110,38 +110,48 @@ interface CatchallKey {
 ## Patterns
 
 Each `Pattern` represents a linear sequence, without selectors.
-Each element of the `body` array MUST either be a non-empty string or an `Expression` object.
-String values represent literal _text_,
-while `Expression` wraps each of the potential _expression_ shapes.
+Each element of the `body` array MUST be a non-empty string, an `Expression` or a `Markup` object.
```
aphillips, line 113: Nit: use the Oxford comma

```suggestion
Each element of the `body` array MUST be a non-empty string, an `Expression`, or a `Markup` object.
```

spec/data-model/README.md
```diff
@@ -110,38 +110,48 @@ interface CatchallKey {
...
+`Expression` wraps each of the potential _expression_ shapes.
+`Markup` wraps each of the potential _markup_ shapes.
 The `body` strings are the "cooked" _text_ values, i.e. escape sequences are processed.
 
-Implementations MUST NOT rely on the set of `Expression` interfaces being exhaustive,
-as future versions of this specification MAY define additional expressions.
+Implementations MUST NOT rely on the set of `Expression` and 
+`Markup` interfaces defined in this document being exhaustive.
+Future versions of this specification might define additional
+types of expression or markup.
```
aphillips, line 122: Minor wording. "Types of" is unncessary and also misleading, since we might use the same "type" of expression for a new function (for example)

```suggestion
expressions or markup.
```

spec/data-model/README.md
```diff
@@ -214,11 +220,43 @@ that the implementation attaches to that _annotation_.
 ```ts
-  sigil: "!" | "@" | "#" | "%" | "^" | "&" | "*" | "<" | ">" | "/" | "?" | "~";
+  sigil: "!" | "@" | "%" | "^" | "&" | "*" | "+" | "<" | ">" | "?" | "~";
```
aphillips, line 223: what order is this list in? Probably we should sort the reserved sigils into code point order?

spec/data-model/README.md
```diff
@@ -214,11 +220,43 @@ that the implementation attaches to that _annotation_.
...
   source: string;
 }
 ```
 
+## Markup
+
+A `Markup` object is either `MarkupOpen`, `MarkupStandalone`, or `MarkupClose`,
+which are differentiated by `kind`.
+The `name` in these does not include the starting sigils `#` and `/`.
```
aphillips, line 232: We should call out the closer too

```suggestion
The `name` in these does not include the starting sigils `#` and `/` 
or the ending sigil `/`.
```

spec/data-model/README.md
```diff
@@ -214,11 +220,43 @@ that the implementation attaches to that _annotation_.
...
-  sigil: "!" | "@" | "#" | "%" | "^" | "&" | "*" | "<" | ">" | "/" | "?" | "~";
+  sigil: "!" | "@" | "%" | "^" | "&" | "*" | "+" | "<" | ">" | "?" | "~";
   source: string;
 }
 ```
 
+## Markup
+
+A `Markup` object is either `MarkupOpen`, `MarkupStandalone`, or `MarkupClose`,
+which are differentiated by `kind`.
```
aphillips, line 231: Should we call out that open/standalone/close carries no semantic meaning, unless assigned by the implementation?

spec/formatting.md
```diff
@@ -98,7 +99,7 @@ but also the _variable_ to which the resolved value of the _variable-expression_
 
 In _selectors_, the resolved value of an _expression_ is used for _pattern selection_.
 
-In a _pattern_, the resolved value of an _expression_ is used in its _formatting_.
+In a _pattern_, the resolved value of _expressions_ and _markup_ is used in their _formatting_.
```
aphillips, line 102: "their" is incorrect, since it refers to "value". Perhaps a bit of avoidance can be added too:

```suggestion
In a _pattern_, the resolved value of an _expression_ or _markup_ is used in its _formatting_.
```

spec/formatting.md
aphillips: What I'm saying is: it's not "afterwards" at all, it might be **now** as part of the resolution, if the implementation chooses to make it work that way.

Instead of saying this (which doesn't really help implementations either way), perhaps state what our limits are:

```suggestion
This specification does not fully define how _markup_ is resolved.
Implementations can handle _markup_ according to their own needs, so long as the following
requirements are met.
```

spec/formatting.md
aphillips: Yes.

spec/message.abnf
```diff
@@ -65,8 +70,8 @@ reserved-keyword = "." name
 
 ; Reserve additional sigils for use by future versions of this specification.
 reserved-annotation = reserved-annotation-start reserved-body
-reserved-annotation-start = "!" / "@" / "#" / "%" / "*"
-                          / "<" / ">" / "/" / "?" / "~"
+reserved-annotation-start = "!" / "@" / "%" / "*" / "+"
+                          / "<" / ">" / "?" / "~"
```
aphillips, line 74: As noted above, we should probably sort these

aphillips (commented):
Mostly minor comments. Getting close.

spec/data-model/README.md
eemeli: I think it's currently something close to the layout order on a US keyboard? I've no opinion on the order here.

spec/data-model/README.md
eemeli: I don't think that necessarily belongs here in the data model, given that it's an optional part of the spec?

spec/data-model/message.json
gibson042: As I said, it's just aesthetic. Referencing a singular is both shallower and a bit easier to keep in my human memory when contemplating the schema.

spec/message.abnf
```diff
@@ -23,13 +23,18 @@ expression = literal-expression / variable-expression / annotation-expression
...
 annotation-expression = "{" [s] annotation [s] "}"
-annotation = (function *(s option))
+annotation = function
            / private-use-annotation
            / reserved-annotation
 
+markup       = "{" [s] markup-open [s] ["/"] "}"
+             / "{" [s] markup-close [s] "}"
+markup-open  = "#" identifier *(s option)
+markup-close = "/" identifier
```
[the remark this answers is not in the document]
stasm: It's there as `["/"]` in the first line. It's not a separate production because it would require an infinite lookahead.

spec/data-model/README.md
aphillips: I think we should call it out, but agree that it doesn't belong here.

spec/syntax.md
aphillips: I feel strongly about this thread, but think we need to merge the overall changes into the spec. I am filing a separate issue to track this conversation.

macchiati (commented):
I think the term "shape" is confusing for many readers and may well be misleading: I stumble across a term like "implementation's shape".

It appears that "structure" is what is meant. If so, it should be replaced by that. 

If this is merged, I can move this comment to the new issue for cleanup.

aphillips:
@macchiati I think your suggested (editorial) change is worth considering. I am filing an issue on your behalf to track it, so that we can merge this work.

aphillips:
Per the 2024-11-08 teleconference, merging this work. Please use new issues or PRs to address specific items within it.

PR #576, opened 2023-12-24
Forbid {�} as a valid expression

eemeli:
Implementation feedback from working on updates to the JS `messageformat` library:

At the moment, the `name` rule includes U+FFFD as a valid `name-start` character. This means that `{�}` is a valid unquoted literal.

This is a potential footgun for implementations that choose to extend our "always emit something, report errors via side channel" formatting behaviour also to their stringification API. If such a stringifier is asked to stringify a data model with unsupported contents, `�` is an obvious character to use in the output, especially given how we're already using it. Currently, however, this would mean that the broken output of the stringifier would parse as valid MF2 syntax.

There is no good reason why `�` should be supported as a `name`, and broken output should not appear valid.

### DIFF
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -91,9 +91,9 @@ name-start = ALPHA / "_"
            / %xC0-D6 / %xD8-F6 / %xF8-2FF
            / %x370-37D / %x37F-1FFF / %x200C-200D
            / %x2070-218F / %x2C00-2FEF / %x3001-D7FF
-           / %xF900-FDCF / %xFDF0-FFFD / %x10000-EFFFF
+           / %xF900-FDCF / %xFDF0-FFFC / %x10000-EFFFF
 name-char  = name-start / DIGIT / "-" / "."
-           / %xB7 / %x300-36F / %x203F-2040
+           / %xB7 / %x300-36F / %x203F-2040 / %xFFFD
 
 text-escape     = backslash ( backslash / "{" / "}" )
 quoted-escape   = backslash ( backslash / "|" )
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -93,7 +93,7 @@ name-start = ALPHA / "_"
            / %x2070-218F / %x2C00-2FEF / %x3001-D7FF
            / %xF900-FDCF / %xFDF0-FFFC / %x10000-EFFFF
 name-char  = name-start / DIGIT / "-" / "."
-           / %xB7 / %x300-36F / %x203F-2040 / %xFFFD
+           / %xB7 / %x300-36F / %x203F-2040
 
 text-escape     = backslash ( backslash / "{" / "}" )
 quoted-escape   = backslash ( backslash / "|" )
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -793,9 +793,9 @@ name-start = ALPHA / "_"
            / %xC0-D6 / %xD8-F6 / %xF8-2FF
            / %x370-37D / %x37F-1FFF / %x200C-200D
            / %x2070-218F / %x2C00-2FEF / %x3001-D7FF
-           / %xF900-FDCF / %xFDF0-FFFD / %x10000-EFFFF
-name-char  = name-start / DIGIT / "-" / "." / ":"
-           / %xB7 / %x300-36F / %x203F-2040
+           / %xF900-FDCF / %xFDF0-FFFC / %x10000-EFFFF
+name-char  = name-start / DIGIT / "-" / "."
+           / %xB7 / %x300-36F / %x203F-2040 / %xFFFD
 ```
 
 ### Escape Sequences
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -795,7 +795,7 @@ name-start = ALPHA / "_"
            / %x2070-218F / %x2C00-2FEF / %x3001-D7FF
            / %xF900-FDCF / %xFDF0-FFFC / %x10000-EFFFF
 name-char  = name-start / DIGIT / "-" / "."
-           / %xB7 / %x300-36F / %x203F-2040 / %xFFFD
+           / %xB7 / %x300-36F / %x203F-2040
 ```
 
 ### Escape Sequences


aphillips:
Actually, the error is on our part. We want `name` to be NCName and NCName uses [XML's Char production](https://www.w3.org/TR/REC-xml/#NT-Char).

Char is defined:

```abnf
  Char ::= | #x9 \| #xA \| #xD \| [#x20-#xD7FF] \| [#xE000-#xFFFD] \| [#x10000-#x10FFFF] | /* any Unicode character, excluding the surrogate blocks, FFFE, and FFFF. */
```

spec/message.abnf
```diff
@@ -91,9 +91,9 @@ name-start = ALPHA / "_"
            / %xC0-D6 / %xD8-F6 / %xF8-2FF
            / %x370-37D / %x37F-1FFF / %x200C-200D
            / %x2070-218F / %x2C00-2FEF / %x3001-D7FF
-           / %xF900-FDCF / %xFDF0-FFFD / %x10000-EFFFF
+           / %xF900-FDCF / %xFDF0-FFFC / %x10000-EFFFF
 name-char  = name-start / DIGIT / "-" / "."
-           / %xB7 / %x300-36F / %x203F-2040
+           / %xB7 / %x300-36F / %x203F-2040 / %xFFFD
```
aphillips, line 96: Lose this change. We don't want replacement characters anywwhere.

spec/syntax.md
```diff
@@ -793,9 +793,9 @@ name-start = ALPHA / "_"
            / %xC0-D6 / %xD8-F6 / %xF8-2FF
            / %x370-37D / %x37F-1FFF / %x200C-200D
            / %x2070-218F / %x2C00-2FEF / %x3001-D7FF
-           / %xF900-FDCF / %xFDF0-FFFD / %x10000-EFFFF
-name-char  = name-start / DIGIT / "-" / "." / ":"
-           / %xB7 / %x300-36F / %x203F-2040
+           / %xF900-FDCF / %xFDF0-FFFC / %x10000-EFFFF
+name-char  = name-start / DIGIT / "-" / "."
+           / %xB7 / %x300-36F / %x203F-2040 / %xFFFD
```
aphillips, line 798: ibid.

aphillips (changes requested):
With the minor tweak, this should be merged.

duerst:
This may have to go into a separate issue/PR, but it seems related, so I'm posting it here:
The grammar currently has the following:
```
simple-start-char = %x0-2D
quoted-char = %x0-5B         ; omit \
reserved-char  = %x00-08        ; omit HTAB and LF
```
This not only allows CR/LF/tab/FF, which may be okay, but also all kinds of other characters with essentially undefined or in one or another way weird (think e.g. about ESC) behavior. I think excluding those would be very prudent. This would avoid security issues and other nastities (e.g. searching for invisible differences if one of these characters is invisible).

duerst:
> This not only allows CR/LF/tab/FF, which may be okay, but also all kinds of other characters with essentially undefined or in one or another way weird (think e.g. about ESC) behavior.

On rereading, I realized I should have been a bit more specific. I'm talking about the C0 area, %x00-19.

gibson042:
> > This not only allows CR/LF/tab/FF, which may be okay, but also all kinds of other characters with essentially undefined or in one or another way weird (think e.g. about ESC) behavior.
> 
> On rereading, I realized I should have been a bit more specific. I'm talking about the C0 area, %x00-19.

Rejecting control characters was explicitly considered, cf. #268 and #282 and #290. I believe the outcome is explained in [syntax.md](https://github.com/gibson042/message-format-wg/blob/main/spec/syntax.md) (emphasis mine):

> The syntax should define as few special characters and sigils as possible. Note that this necessitates extra care when presenting messages for human consumption, because they may contain invisible characters such as U+200B ZERO WIDTH SPACE, **control characters such as U+0000 NULL and U+0009 TAB**, permanently reserved noncharacters (U+FDD0 through U+FDEF and U+nFFFE and U+nFFFF where n is 0x0 through 0x10), private-use code points (U+E000 through U+F8FF, U+F0000 through U+FFFFD, and U+100000 through U+10FFFD), unassigned code points, and other potentially confusing content.

> Any Unicode code point is allowed [in _text_], except for surrogate code points U+D800 through U+DFFF inclusive.

> A _literal_ MAY include any Unicode code point except for surrogate code points U+D800 through U+DFFF.

duerst:
@gibson042 Thanks for your explanation. I think this is all okay, except when it comes to security considerations. Somewhere in the spec, the fact that a message can contain arbitrary control characters should be clearly called out as a security issue.

aphillips:
@duerst Thanks for the comment. Of course you're right: there are security risks here. For example, bidi-based spoofing is an issue. I created #579 to track this and other security considerations.

> Somewhere in the spec, the fact that a message can contain arbitrary control characters should be clearly called out as a security issue.

Our syntax permits arbitrary control characters to appear inside the pattern portions of a message, but this does not imply anything about environment in which the message is serialized, stored, or processed. We don't provide character escapes, for example, because it is expected that the host format or system will do so. Most users will see the message as it is layered in their source code format and this will often make control characters visible as escaped. 

Most string types are defined as a sequence of code points or code units with minimal restriction on the characters used. Our goal is that anything you could write as a string resource can be used in a pattern.

spec/message.abnf
eemeli:
```suggestion
           / %xB7 / %x300-36F / %x203F-2040
```

spec/syntax.md
eemeli:
```suggestion
           / %xB7 / %x300-36F / %x203F-2040
```

aphillips:
Ship it...
```
          |\___..--"/
   __..--``        /
  \_______________/   
```

aphillips (approved):

stasm (approved):
LGTM.

Would it make sense to treat U+FFFD the same way we treat surrogates, and forbid it globally from `simple-start-char` (which implies `text-char`), from `quoted-char`, and from `reserved-char`?

aphillips:
@stasm 
> Would it make sense to treat U+FFFD the same way we treat surrogates, and forbid it globally from simple-start-char (which implies text-char), from quoted-char, and from reserved-char?

No, because it is a valid text char (can appear in literals).

PR #577, opened 2024-01-01
Clarify that Duplicate Option Name is a data model error

eemeli:
I noticed during implementation that even though we define Duplicate Option Name as a data model error, we handle it as a formatting error. This should be clarified.

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -214,8 +214,6 @@ the following steps are taken:
    If _options_ is missing, the mapping will be empty.
    For each _option_:
    - Resolve the _identifier_ of the _option_.
-   - If the _option_'s _identifier_ already exists in the resolved mapping of _options_,
-     emit a Duplicate Option Name error.
    - If the _option_'s right-hand side successfully resolves to a value,
      bind the _identifier_ of the _option_ to the resolved value in the mapping.
    - Otherwise, bind the _identifier_ of the _option_ to an unresolved value in the mapping.
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -89,7 +89,8 @@ Attempting to parse a _message_ that is not _well-formed_ will result in a _Synt
 
 A _message_ is **_<dfn>valid</dfn>_** if it is _well-formed_ and
 **also** meets the additional content restrictions
-and semantic requirements about its structure defined below.
+and semantic requirements about its structure defined below for
+_declarations_, _matcher_ and _options_.
 Attempting to parse a _message_ that is not _valid_ will result in a _Data Model Error_.
 
 ## The Message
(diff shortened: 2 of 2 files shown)


spec/syntax.md
```diff
@@ -89,7 +89,8 @@ Attempting to parse a _message_ that is not _well-formed_ will result in a _Synt
 
 A _message_ is **_<dfn>valid</dfn>_** if it is _well-formed_ and
 **also** meets the additional content restrictions
-and semantic requirements about its structure defined below.
+and semantic requirements about its structure defined below for
+_declarations_, _matcher_ and _options_.
```
aphillips, line 93: Why is this addition necessary?

spec/syntax.md
```diff
@@ -554,6 +555,9 @@ The value of an _option_ can be either a _literal_ or a _variable_.
 Multiple _options_ are permitted in an _annotation_.
 Each _option_ is separated by whitespace.
 
+The _option_ _identifiers_ MUST be unique within each _expression_.
+Otherwise, the _message_ is not considered _valid_.
```
aphillips, line 559: I think some reformulation is needed here.

```suggestion
Multiple _options_ are permitted in an _annotation_.
_Options_ are separated from other parts of the _annotation_
and from each other by whitespace.
Each _option_'s _identifier_ MUST be unique within the _annotation_: 
an _annotation_ with duplicate _option_ _identifiers_ is not valid.
```

spec/formatting.md
```diff
@@ -214,8 +214,6 @@ the following steps are taken:
    If _options_ is missing, the mapping will be empty.
    For each _option_:
    - Resolve the _identifier_ of the _option_.
-   - If the _option_'s _identifier_ already exists in the resolved mapping of _options_,
-     emit a Duplicate Option Name error.
```
aphillips, line 218: Why remove this? An implementation might detect this during parsing (which, it sounds like, you're doing when populating the data model). But it might not (if it naively parses the message and doesn't evaluate options until formatting time). Does this directive do any harm?

aphillips (commented):
Good call out, I think. Some comments.

spec/formatting.md
eemeli: Because we don't do that for any of the data model errors, and keeping this here would create a requirement for throwing data model errors during formatting, rather than before.

We should probably move the syntax & data model error definitions out of `formatting.md` into their own doc, and make it clearer that a message must be validated against all of them before formatting.

spec/syntax.md
eemeli: It makes it clearer where in the doc to hunt for those "semantic requirements". IIRC we started with a single list that was pretty much the next paragraph after this, but the refactorings and later changes have moved them all over the place.

aphillips (approved):

PR #488, opened 2023-10-08
Use `input` and `local` keywords instead of `let` in declarations

eemeli:
Implements the changes proposed in #469. _Submitting this as a draft until that PR is merged as accepted._

The _expression_ rule is here redefined as
```abnf
expression = literal-expression / variable-expression / function-expression
literal-expression = "{" [s] literal [s annotation] [s] "}"
variable-expression = "{" [s] variable [s annotation] [s] "}"
function-expression = "{" [s] annotation [s] "}"
```

in order to make it easier for human readers, and to make _variable-expression_ reusable in _input-declaration_. This isn't strictly speaking LL1, but it may be transformed to its earlier shape to make it LL1 again.

PR #580, opened 2024-01-05
Clarify Duplicate Declaration error

eemeli:
During implementation, I noticed that while we included this in the variable mutability design doc: https://github.com/unicode-org/message-format-wg/blob/fb3da4c670c71fcee6a5551ef4f04ad9489ac408/exploration/variable-mutability.md?plain=1#L179-L186

we missed out on accounting for that in #488. So this PR adds a Forward Reference data model error for that situation. We might also want to consider this Duplicate Declaration error as a Forward Reference error instead: https://github.com/unicode-org/message-format-wg/blob/2a69d2a17ee7476282bfa824fd4ca62e0a03f6e1/spec/syntax.md?plain=1#L181


spec/formatting.md
```diff
@@ -834,6 +834,21 @@ These are divided into the following categories:
...
+    > Examples of invalid messages resulting in a Forward Reference error:
+    >
+    > ```
+    > .local $foo = {$bar :number}
+    > .local $bar = {42 :number}
+    > {{The answer is {$foo}}}
+    >
+    > .local $foo = {42 :number minimumFractionDigits=$bar}
+    > .local $bar = {2}
+    > {{The answer is {$foo}}}
```
aphillips, line 849: We should show `.input` examples also, to avoid the impression that this only happens with `.local`

```suggestion
{{The answer is {$foo}}}

.input $foo = {$bar}
.local $bar = {2}
```

spec/formatting.md
```diff
@@ -834,6 +834,21 @@ These are divided into the following categories:
     > {{{$var} cannot be redefined. {$var2} cannot refer to itself}}
     > ```
 
+  - A **Forward Reference error** occurs when a _declaration_ _expression_ includes a _variable_
+    that is declared after it in the _message_.
```
aphillips, line 838: Is it truly a "forward reference error" or it is actually a species of "duplicate name"?

That is, with `.local $foo = {$bar} .local $bar = {something}`, the first `$bar` might be resolved from the inputs (and that could fail), making the second `local` an attempt at reassignment.

aphillips (changes requested):

spec/formatting.md
eemeli: Which confusion is why we want this to be an error. I'd be fine describing this either way, as long as we do include the error.

spec/formatting.md
eemeli:
```suggestion
    > .local $foo = {$bar :number}
    > .local $bar = {42 :number}
    > {{The answer is {$foo}}}
    >
    > .input {$foo :number minimumFractionDigits=$bar}
    > .local $bar = {2}
    > {{The answer is {$foo}}}
```

eemeli:
I've refactored the PR to not add a new error, and instead clarify the existing Duplicate Declaration error to cover this case as well.

aphillips (changes requested):
Thanks for this. Minor tweaks suggested.

aphillips (approved):

stasm (approved):

eemeli:
@aphillips I've resolved the conflicts, and this should now merge cleanly.

PR #581, opened 2024-01-08
Add :stringified, :enum as alternative in the exact match selector design doc

stasm:
A follow-up to the [December 18](https://github.com/unicode-org/message-format-wg/blob/main/meetings/2023/notes-2023-12-18.md) meeting: document two more naming alternatives for the default exact match selector.

### DIFF
--- a/exploration/exact-match-selector-options.md
+++ b/exploration/exact-match-selector-options.md
@@ -127,6 +127,7 @@ Pros:
 
 Cons:
 - Somewhat generic name.
+  All selectors "select", not just this one.
 
 ### Option B. `:exact`
 
@@ -147,6 +148,7 @@ Cons:
 - Implies object or value equality, which may not be correct in all implementations
   or programming languages.
   The implications might be confusing to users in those cases.
+- See `:exact`: doesn't specify that the equality is string-based.
 
 ### Option D. `:string`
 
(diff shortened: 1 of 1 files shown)


eemeli (approved):

exploration/exact-match-selector-options.md
```diff
@@ -127,6 +127,7 @@ Pros:
 
 Cons:
 - Somewhat generic name.
+  All selectors "select", not just this one.
```
aphillips, line 130: nit

```suggestion
- All selectors "select", not just this one.
```

exploration/exact-match-selector-options.md
```diff
@@ -158,9 +160,26 @@ Cons:
 - Might confuse users who are comparing e.g. two numbers or two dates.
   In typed languages, the values being compared might not actually be strings.
 - Depends on the serialization.
+- Doesn't sound like it's doing selection.
 
+### Option E. `:stringified`
```
aphillips, line 165: Ick?

aphillips:
Are you really suggesting that we reopen the naming discussion for this selector? I know that there is a lack of love for `:string`, but I think I feel that deficit for other options too. Let's discuss with the group how to attack it.

aphillips (approved):

PR #584, opened 2024-01-09
Add new terms to glossary

SimonClark:
Adding some new terms from discussions in the group. These are my understandings of how we are using the words, so feel free to suggest better definitions.

### DIFF
--- a/docs/glossary-and-resources.md
+++ b/docs/glossary-and-resources.md
@@ -36,7 +36,6 @@
 - **format to parts** - A formatting call that returns an array of objects representing the message,
   or some part of it, rather than a single concatenated string.
   A formatted part may also represent a non-string value, such as an inline image.
-- **formatter** - in MessageFormat, a function that takes a set of parameters (eg: ```code: 'USD', amount: 3.9```), and returns formatted text that can be incorporated into the output string (eg: ```"<span class="cur"><span class="cur-symbol">$</span><span class="cur-amt">3.90</span> <span class="cur-code">USD</span>"```)
 - **function** - in MessageFormat, functionality that is callable in the context of a MessageFormat message to perform a variety of functions, from determining the plural form needed to represent a number, to formatting an option for output.
 - **gender** - In many languages, nouns are considered to have a gender which requires changes to the text around the noun. Common genders in linguistics are feminine, masculine, neuter, common (gendered) or epicene.
 - **grammeme** - a value of a grammatical category, such as "singular" or "plural" for numerical grammer, or "male" or "female" for gendered nouns
--- a/docs/glossary-and-resources.md
+++ b/docs/glossary-and-resources.md
@@ -24,7 +24,7 @@
 ### Terms for message formatting
 
 
-- **[ABNF](https://en.wikipedia.org/wiki/Augmented_Backus%E2%80%93Naur_form)** - (Augmented Backus-Naur Form). ABNF is a method of specifying valid syntax for a specification. The MessageFormat ABNF spec defines what is and is not required for a string to be valid MessageFormat. ABNF is specified by RFC 5234 and RFC 7405; x
+- **[ABNF](https://en.wikipedia.org/wiki/Augmented_Backus%E2%80%93Naur_form)** - (Augmented Backus-Naur Form). ABNF is a method of specifying valid syntax for a specification. The MessageFormat ABNF spec defines what is and is not required for a string to be valid MessageFormat. ABNF is specified by RFC 5234 and RFC 7405.
 - **API** - any function call(s) invoked by the user to perform message formatting.
 - **API argument format** - syntax of values passed to input of message formatting API. May also refer to structure of values represented by the syntax. May be similar or same as _authoring format_. See _message syntax_.
 - **application locale** - the locale for formatting (or formatting resources) requested by an application.
--- a/docs/glossary-and-resources.md
+++ b/docs/glossary-and-resources.md
@@ -53,7 +53,7 @@
 - **operand** - A character in a message that denotes an action taken during runtime execution, such as value assignment or comparison.
 - **options** - values passed in, along with an MF2 message, that will be used to determine the output. Examples may be item count in a search results string.
 - **placeholder** - An expression or markup in a message pattern.
-- **registry** - MF2's function registry, which allows reusable logic to be executed at runtime while resolving an MF2 string to output.
+- **registry** - MF2's function registry, which allows reusable logic to be executed at runtime while formatting an MF2 message.
 - **resource** - files bundled with an application that are loaded in by the executable code. UI strings, etc. and their locale-specific translations are typically stored as resources.
 - **roundtrip** - the process of transforming a message into another format or representation, then transforming it back into the original format.
 - **translation merging** - in l10n TMSes, the document-level interpolation of translated content. In other words, the replacing of translatable units in the source document with their equivalent translated units. See _interpolate_.
--- a/docs/glossary-and-resources.md
+++ b/docs/glossary-and-resources.md
@@ -53,7 +53,7 @@
 - **open/close** - syntax within the source test that changes the execution context of the parser from output mode to code mode, and back again.
 - **operand** - A character in a message that denotes an action taken during runtime execution, such as value assignment or comparison.
 - **options** - values passed in, along with an MF2 message, that will be used to determine the output. Examples may be item count in a search results string.
-- **placeholder** - Syntax in the MF2 message that deontes that an option should be inserted into that place in the string.
+- **placeholder** - An expression or markup in a message pattern.
 - **registry** - MF2's function registry, which allows reusable logic to be executed at runtime while resolving an MF2 string to output.
 - **resource** - files bundled with an application that are loaded in by the executable code. UI strings, etc. and their locale-specific translations are typically stored as resources.
 - **roundtrip** - the process of transforming a message into another format or representation, then transforming it back into the original format.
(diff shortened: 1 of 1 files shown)


docs/glossary-and-resources.md
```diff
@@ -23,23 +23,44 @@
...
+- **attribute** - A value in a message or a function call that provides additional information that may be needed to properly resolve. Attributes are key/value pairs expressed in the form "@key=value".
 - **authoring** - writing a message by hand that adheres to some syntax.
 - **authoring format** - syntax of message formatting inputs when constructed for a program, either manually (by a developer) or programtically (ex: a WYSIWYG tool used by a translator). May also refer to structure of values represented by the syntax. May be similar or same as _API argument format_. See _message syntax_.
+- **[CLDR](https://cldr.unicode.org/)** - Unicode's Common Locale Data Repository - a set of standardized building blocks that enable software to support languages and locales around the world.
 - **data model** - a syntax-independent description of the structure of values passed to the message formatting API.
-- **implementation** - code written to make the API achieve the intended behavior of output for a given input.
+- **format to parts** - context-aware functionality that can take a formatted string (such as a monetary amount, or a date), and return the seperated constituent parts.
```
SimonClark, line 36: Maybe this doesn't belong here because it is rarely discussed, but i also went with a generic description that may not be 100% accurate in the MF2 context.

docs/glossary-and-resources.md
```diff
@@ -23,23 +23,44 @@
...
+- **formatter** - in MessageFormat, a function that takes a set of parameters (eg: ```code: 'USD', amount: 3.9```), and returns formatted text that can be incorporated into the output string (eg: ```"<span class="cur"><span class="cur-symbol">$</span><span class="cur-amt">3.90</span> <span class="cur-code">USD</span>"```)
```
eemeli, line 37: This one is probably best left our for now; the shape of the value returned from a formatter is something that's explicitly left out of the spec as an implementation detail.

Your suggestion, for instance, presumes that a message is being formatted to a string.

```suggestion
```

docs/glossary-and-resources.md
```diff
@@ -23,23 +23,44 @@
...
+- **implementation** - code written to achieve the intended behavior of output for a given input. It is expected that MessageFormat will have implementations in all popular programming languages.
+- **inflection** - a variation in a word or phrase common in many languages, that expresses further context such as tense, mood, formality, etc.
 - **interchange format** - syntax/file format used to convert the inputs of message formatting into inputs for other systems (ex: l10n systems).
 - **interpolate** - inserting the contents of one string in the middle of another, at places indicated by a pattern / placeholder. See _translation merging_.
+- **LDML** - Locale Data Markup Language - the spec schema that defines CLDR
 - **locale fallback** - offering a reasonable substitute locale when the requested locale's resources are not available. Results may vary depending on context (ex: audio vs. text vs. video).
 - **locale matching** - computing the locale fallback.
+- **markup** - characters present in the MessageFormat source strings that are intended to provide structure, functionality, or formatting to the output string, as opposed to display text.
```
eemeli, line 48:
```suggestion
- **markup** - characters present in the MessageFormat source strings that are intended to provide structure, functionality, or formatting to the output, as opposed to display text.
```

docs/glossary-and-resources.md
```diff
@@ -23,23 +23,44 @@
...
+- **match** - a mechanism in MessageFormat that allows the message author to specify a variety of outcomes based on the combined values of one or more inputs.
+- **metadata** - Addition information about an MF2 message that aids in translation, or other pre-runtime processes, but has no impact on how the message will be rendered. This may be context notes, screenshots of the string in situ, etc. Metadata should never be relied upon at runtime.
```
eemeli, line 50: As we've left expression attributes out of the LDML 45 release and we've previously ruled out message-level metadata from the MF2 spec, we've avoided needing to explicitly agree whether or not metadata may have a runtime impact.

docs/glossary-and-resources.md
```diff
@@ -23,23 +23,44 @@
...
+- **open/close** - syntax within the source test that changes the execution context of the parser from output mode to code mode, and back again.
```
eemeli, line 51: This one confuses me; is this about pattern delimiters? Probably not markup, which also uses the "open" and "close" terms.

docs/glossary-and-resources.md
```diff
@@ -23,23 +23,44 @@
...
+- **operand** - A character in a message that denotes an action taken during runtime execution, such as value assignment or comparison.
```
eemeli, line 52: I think we more commonly refer to the positional argument of an annotation as an operand?

docs/glossary-and-resources.md
```diff
@@ -23,23 +23,44 @@
...
+- **options** - values passed in, along with an MF2 message, that will be used to determine the output. Examples may be item count in a search results string.
```
eemeli, line 53: Using "options" for this is misleading; that's used often for the named arguments of expressions and markup.

For this meaning, I think we usually say "arguments" or "parameters".

docs/glossary-and-resources.md
```diff
@@ -23,23 +23,44 @@
...
+- **placeholder** - Syntax in the MF2 message that deontes that an option should be inserted into that place in the string.
```
eemeli, line 54:
```suggestion
- **placeholder** - An expression or markup in a message pattern.
```

docs/glossary-and-resources.md
```diff
@@ -23,23 +23,44 @@
...
+- **registry** - MF2's function registry, which allows reusable logic to be executed at runtime while resolving an MF2 string to output.
```
eemeli, line 55:
```suggestion
- **registry** - MF2's function registry, which allows reusable logic to be executed at runtime while formatting an MF2 message.
```

docs/glossary-and-resources.md
```diff
@@ -23,23 +23,44 @@
...
 - **resource** - files bundled with an application that are loaded in by the executable code. UI strings, etc. and their locale-specific translations are typically stored as resources.
 - **roundtrip** - the process of transforming a message into another format or representation, then transforming it back into the original format.
 - **translation merging** - in l10n TMSes, the document-level interpolation of translated content. In other words, the replacing of translatable units in the source document with their equivalent translated units. See _interpolate_.
 - **selector** - see _placeholder type_.
 - **specification** - the rules we decide that describe what is passed to the API for message formatting (structure of data, syntax, etc.).
 - **serialization** - how to convert in-memory representations of data to/from a file/stream.
+- **sigil** - A character in a string that is used to denote a change in context.
```
eemeli, line 62: This isn't quite right; they used to be the leading prefix character, but now with standalone markup we also use one in a trailing position.

docs/glossary-and-resources.md
```diff
@@ -23,23 +23,44 @@
...
+- **spannable** - A construct in messages where there is an open and a close denotation (start and stop). The italics markup, for instance ("<i>This is italicized</i>") is a spannable.
```
eemeli, line 63:
```suggestion
- **spannable** - See _markup_.
```

eemeli (changes requested):

docs/glossary-and-resources.md
aphillips: This description is actually wrong, as "format to parts" functions generally don't start with the formatted string. They start with an object (whatever the input is) and return the object as a sequence of formatted parts. See, for example, Intl.NumberFormat

docs/glossary-and-resources.md
aphillips: This is confusing for the reasons @eemeli states.

aphillips (changes requested):
Thanks @SimonClark for suggesting these additions. 

Since this is an informative document, I might make two suggestions:

1. We should leave this work to post-45
2. The spec has many definitions carefully formatted into it. You might look for the `<dfn>` tags in syntax.md, formatting.md, etc. and see if the definitions there are mirrored here. Where the definitions in the spec are insufficient (require more than just some formatting changes to appear in this document), that might be cause for work on the source document.

docs/glossary-and-resources.md
SimonClark: ok, thanks.  I was inferring from how I'd heard it used in conversation, but I don't think we are all using is consistently yet.

docs/glossary-and-resources.md
```diff
@@ -23,23 +23,44 @@
 
 ### Terms for message formatting
 
+
+- **[ABNF](https://en.wikipedia.org/wiki/Augmented_Backus%E2%80%93Naur_form)** - (Augmented Backus-Naur Form). ABNF is a method of specifying valid syntax for a specification. The MessageFormat ABNF spec defines what is and is not required for a string to be valid MessageFormat. ABNF is specified by RFC 5234 and RFC 7405; x
```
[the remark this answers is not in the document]
SimonClark: Perhaps. I approached this as an audience of me, trying to become a useful participant/contributor as well as consumer of MF2. I can see that it makes sense to narrow the audience to just the latter.

docs/glossary-and-resources.md
SimonClark: It is a term I have heard many times in conversation, but not seen written down, so consider this a blind stab in the dark. 

I could drop it, or split it into separate terms.

docs/glossary-and-resources.md
SimonClark: I'm having trouble parsing your meaning here.  Perhaps an example?

aphillips: The `operand` is the one argument that, if present, is passed to the annotating function. That is, it is the thing being "operated on". In the following placeholder examples, it is the thing with `operand` in it:

```
{$operand :function option=value @annotation=value}
{|operand| :function option=value @annotation=value}
```

We don't have a production `operand` in the ABNF, so this is an independent term.

PR #585, opened 2024-01-10
Flatten JSON & TS data model Pattern element

eemeli:
Feedback from writing the JS implementation:

In the data model, the wrapping object in the `Pattern` which has a single `body` property is rather useless. We should flatten this to simplify the model a bit.

One reason why the wrapping object was included earlier was to enable an extension of the data model to function as a CST, but in practice I at least found that its requirements are sufficiently different for the CST to work better as a separately defined structure.

This PR overlaps a bit with #574; whichever lands second will need a small adjustment to deconflict the merge.

### DIFF
--- a/spec/data-model/README.md
+++ b/spec/data-model/README.md
@@ -109,11 +109,14 @@ interface CatchallKey {
 
 ## Patterns
 
-Each `Pattern` represents a linear sequence, without selectors.
+Each `Pattern` is a standalone linear sequence of text and placeholders
+that represents the output of a message.
+
 Each element of the `Pattern` MUST either be a non-empty string or an `Expression` object.
-String values represent literal _text_,
-while `Expression` wraps each of the potential _expression_ shapes.
-The `Pattern` strings are the "cooked" _text_ values, i.e. escape sequences are processed.
+String values represent literal _text_.
+String values include all processing of the underlying _text_ values,
+including escape sequence processing.
+`Expression` values wrap each of the _expression_ shapes.
 
 Implementations MUST NOT rely on the set of `Expression` interfaces being exhaustive,
 as future versions of this specification MAY define additional expressions.
--- a/spec/data-model/README.md
+++ b/spec/data-model/README.md
@@ -110,18 +110,16 @@ interface CatchallKey {
 ## Patterns
 
 Each `Pattern` represents a linear sequence, without selectors.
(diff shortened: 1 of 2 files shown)


spec/data-model/README.md
```diff
@@ -110,18 +110,16 @@ interface CatchallKey {
 ## Patterns
 
 Each `Pattern` represents a linear sequence, without selectors.
-Each element of the `body` array MUST either be a non-empty string or an `Expression` object.
+Each element of the `Pattern` MUST either be a non-empty string or an `Expression` object.
```
aphillips, line 113: I think this is a good change, especially since `body` rhymes with `complex-body`, which means something entirely else.

spec/data-model/README.md
```diff
@@ -110,18 +110,16 @@ interface CatchallKey {
 ## Patterns
 
 Each `Pattern` represents a linear sequence, without selectors.
```
aphillips, line 112: You might want to adjust this sentence also. It might a relic of earlier times, before the addition of pattern delimiters. Perhaps:

```suggestion
Each `Pattern` is a standalone linear sequence of text and placeholders
that represents the output of a message.
```

spec/data-model/README.md
```diff
@@ -110,18 +110,16 @@ interface CatchallKey {
...
-Each element of the `body` array MUST either be a non-empty string or an `Expression` object.
+Each element of the `Pattern` MUST either be a non-empty string or an `Expression` object.
 String values represent literal _text_,
 while `Expression` wraps each of the potential _expression_ shapes.
-The `body` strings are the "cooked" _text_ values, i.e. escape sequences are processed.
+The `Pattern` strings are the "cooked" _text_ values, i.e. escape sequences are processed.
```
aphillips, line 116: The whole "cooking" thing seems just a bit informal to me. Here we refer to "`Pattern` strings" and I think that points to the problem. Perhaps:

```suggestion
String values represent literal _text_.
String values include all processing of the underlying _text_ values, that is, after escape sequences have been processed.

`Expression` values wrap each of the potential _expression_ shapes.

```

aphillips (commented):
Looks good. Some editorial comments.

spec/data-model/README.md
eemeli: Added this with a few tweaks.

aphillips (approved):

gibson042 (approved):

PR #588, opened 2024-01-10
Add security considerations section

aphillips:
Addresses #579 

Adds a document for appendices and populates with with Security Considerations. I considered adding this to the front matter, but observe that most documents place these types of materials at the end.

Please suggest other considerations as needed.

@duerst, please also have a look.

### DIFF
--- a/spec/appendices.md
+++ b/spec/appendices.md
@@ -14,14 +14,15 @@ host environments, their serializations and resource formats,
 that might be sufficient to prevent most problems.
 However, MessageFormat itself does not supply such a restriction.
 
-MessageFormat _messages_ permit nearly all Unicode characters, excepting surrogates,
-to appear in a _pattern_.
+MessageFormat _messages_ permit nearly all Unicode code points,
+with the exception of surrogates, to appear in a _pattern_.
 This means that it can be possible for a _message_ to contain invisible unquoted
 control characters that abnormally affect the display of the _message_
 but do not generate errors from MessageFormat parsers or processing APIs.
 
-
-Users are also reminded that, since MessageFormat 2.0's syntax makes use of 
+Bidirectional text containing right-to-left characters (such as used for Arabic or Hebrew) 
+also poses a potential source of confusion for users. 
+Since MessageFormat 2.0's syntax makes use of 
 keywords and symbols that are left-to-right or consist of neutral characters 
 (including characters subject to mirroring under the Unicode Bidirectional Algorithm), 
 it is possible to create messages that,
--- a/spec/appendices.md
+++ b/spec/appendices.md
@@ -1,0 +1,32 @@
+# DRAFT Appendices
+
+## Security Considerations
+
+MessageFormat 2.0 _patterns_ are meant to allow a _message_ to include any string value
+which users might normally wish to use in their environment.
+Programming languages and other environments vary in what characters are permitted
+to appear in a valid string.
+In many cases, certain types of characters, such as invisible control characters,
+require escaping by these host formats.
+In other cases, strings are not permitted to contain certain characters at all.
+Since _messages_ are subject to the restrictions and limitations of their 
+host environments, their serializations and resource formats,
+that might be sufficient to prevent most problems.
+However, MessageFormat itself does not supply such a restriction.
+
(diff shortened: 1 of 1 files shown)


duerst:
@aphillips Thanks for putting this together. Just a few comment, nothing serious:
- "Unicode characters, excepting surrogates": Can be read as if surrogates were characters, an impression which we want to avoid at all costs.
- "Users are also reminded that": That paragraph is perfectly okay for somebody who already understands bidi, but is a bit difficult to read for people for whom bidi isn't a major concern. I'd rewrite the text so that something like "right-to-left text" (preferably with the addition "e.g. Arabic or Hebrew") is mentioned very early in the paragraph.

eemeli (approved):
This is a good start. We should also include discussion of custom functions as an attack vector. I presented this two years ago, but can't find the relevant meeting notes on it just now: https://docs.google.com/presentation/d/19Q90N6Z9HNF0t_bc2_jtUdYu2PKTPy1oFxuJsxrh2O8/edit?usp=drivesdk

aphillips:
@eemeli I think that's a good callout, but one we could include in a new PR.

spec/appendices.md
```diff
@@ -0,0 +1,33 @@
...
+In many cases, certain types of characters, such as invisible control characters,
+require escaping by these host formats.
+In other cases, strings are not permitted to contain certain characters at all.
+Since _messages_ are subject to the restrictions and limitations of their 
+host environments, their serializations and resource formats,
+that might be sufficient to prevent most problems.
+However, MessageFormat itself does not supply such a restriction.
+
+MessageFormat _messages_ permit nearly all Unicode code points,
+with the exception of surrogates, to appear in a _pattern_.
```
macchiati, line 18: with the exception of surrogates, to appear in a _pattern_.
```suggestion
with the exception of surrogates, to appear in literal text in a _pattern_.
```

spec/appendices.md
```diff
@@ -0,0 +1,33 @@
...
+This means that it can be possible for a _message_ to contain invisible unquoted
+control characters that abnormally affect the display of the _message_
+but do not generate errors from MessageFormat parsers or processing APIs.
+
+Bidirectional text containing right-to-left characters (such as used for Arabic or Hebrew) 
+also poses a potential source of confusion for users. 
+Since MessageFormat 2.0's syntax makes use of
```
macchiati, line 25: I think we need to separate 2 cases: what the MF message appears like to the programmer, vs what a resulting message appears like to the user. 

1. For the former, UTS55 is relevant.
2. For the latter, we can point out that MF conformance requires that placeholder replacements be generated in a way that avoids bidi reordering issues between the literal message text and the placeholder contents.

spec/appendices.md
```diff
@@ -0,0 +1,33 @@
...
+In other cases, strings are not permitted to contain certain characters at all.
+Since _messages_ are subject to the restrictions and limitations of their 
+host environments, their serializations and resource formats,
+that might be sufficient to prevent most problems.
+However, MessageFormat itself does not supply such a restriction.
+
+MessageFormat _messages_ permit nearly all Unicode code points,
+with the exception of surrogates, to appear in a _pattern_.
+This means that it can be possible for a _message_ to contain invisible unquoted
+control characters that abnormally affect the display of the _message_
```
macchiati, line 20: control characters that abnormally affect the display of the _message_
```suggestion
control characters (such as bidirectional format characters) that abnormally affect the display of the _message_
```

spec/appendices.md
aphillips: actually...

```suggestion
with the exception of surrogates, 
to appear in _literals_, including the text portions of a _pattern_
```

spec/appendices.md
aphillips: This text refers specifically to viewing the _message_ in a source code context (and thus UTS55 is relevant). 

That reminds me that there is probably the need to call out another security consideration, which is that the output of formatting needs to be examined for e.g. injection or other abuse.

spec/appendices.md
aphillips: @duerst's original comment was about C0 controls. The bidi controls also certainly affect the display. I'll cite both.

PR #592, opened 2024-01-12
Add @attributes as reserved

### LINKED ISSUE #453: What is the namespace for `expression attributes` ?
In #450, there are two classes of _expression attributes_: built-in values such as `@locale` and implementation or private-use values (which might include values like `@can-reorder` or `@_websiteId`)

@eemeli proposed in the PR to use the presence of a hyphen in the name (`@can-reorder`) as the namespacing mechanism.
@aphillips proposed to use a starting character of underscore (`@_myAttribute`) as the namespacing mechanism

Both are consistent with `name` in that they don't require modification to the syntax to make them valid.

What we choose here should be consistent with other naming rules. We currently don't have similar distinctions elsewhere, but we are discussing using a "second character" or separate sigil for local vs. external variables in the discussion on mutability. I'm writing this issue to ensure that we making our naming schemes consistent so that we don't have different naming requirements in different parts. If would be weird to require (for example) a hyphen in expression attributes but an underscore for local vars:

```
let $_local = {:amendLocale nu=Latn @locale=th-TH @some-attr=foo}
```

### LINKED ISSUE #587: Sort the reserved sigils, check for completeness
In #574 I suggested that the sigils should be sorted in the ABNF. This is to track that comment.

currently we have:

```abnf
reserved-annotation-start = "!" / "@" / "%" / "*" / "+"
                          / "<" / ">" / "?" / "~"
```

I would suggest this, which is in code point order:

```abnf
reserved-annotation-start = "!" / "%" / "*" / "+"
                          / "<" / ">" / "?" / "@" / "~"
```

Also, are we missing these ASCII punctuation items: `:`, `;`, `(`, `)` ? (Note that `&` and `^` are private use sigils)

eemeli:
As discussed during this week's call, this adds @attributes to _expression_ and _markup_ with runtime impact.

For the syntax & data model, this is mostly a rebase of #450 that leaves out any discussion of what attributes might be good for, while noting that "_Attributes_ are reserved for use by future specification versions."

For formatting, the language is rather more forbidding:
> _Attributes_ MUST NOT have any impact on message formatting, and MUST NOT be available in the formatted output. _Variables_ in _attribute_ _values_ are not resolved, and no errors are emitted for them.

Closes #453, as using _identifier_ brings in the same namespacing we're using for _options.

This also closes #587, as removing `@` leaves the remaining symbols sorted in code point order.

### DIFF
--- a/spec/data-model/README.md
+++ b/spec/data-model/README.md
@@ -135,24 +135,33 @@ interface LiteralExpression {
   type: "expression";
   arg: Literal;
   annotation?: FunctionAnnotation | UnsupportedAnnotation;
+  attributes?: Attribute[];
 }
 
 interface VariableExpression {
   type: "expression";
   arg: VariableRef;
   annotation?: FunctionAnnotation | UnsupportedAnnotation;
+  attributes?: Attribute[];
 }
 
 interface FunctionExpression {
   type: "expression";
   arg?: never;
   annotation: FunctionAnnotation;
+  attributes?: Attribute[];
 }
 
 interface UnsupportedExpression {
   type: "expression";
   arg?: never;
   annotation: UnsupportedAnnotation;
+  attributes?: Attribute[];
+}
+
+interface Attribute {
+  name: string;
+  value?: Literal | VariableRef;
 }
 ```
 
@@ -219,7 +228,7 @@ that the implementation attaches to that _annotation_.
 ```ts
 interface UnsupportedAnnotation {
   type: "unsupported-annotation";
-  sigil: "!" | "@" | "%" | "^" | "&" | "*" | "+" | "<" | ">" | "?" | "~";
+  sigil: "!" | "%" | "^" | "&" | "*" | "+" | "<" | ">" | "?" | "~";
   source: string;
 }
 ```
@@ -241,6 +250,7 @@ interface MarkupOpen {
   kind: "open";
   name: string;
   options?: Option[];
+  attributes?: Attribute[];
 }
 
 interface MarkupStandalone {
@@ -248,12 +258,14 @@ interface MarkupStandalone {
   kind: "standalone";
   name: string;
   options?: Option[];
+  attributes?: Attribute[];
 }
 
 interface MarkupClose {
   type: "markup";
   kind: "close";
   name: string;
+  attributes?: Attribute[];
 }
 ```
 
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -652,10 +652,9 @@ MUST be an empty string.
 Implementations MAY offer functionality for customizing this,
 such as by emitting XML-ish tags for each _markup_.
 
-_Attributes_ MUST NOT have any impact on message formatting,
-and MUST NOT be available in the formatted output.
-_Variables_ in _attribute_ _values_ are not resolved,
-and no errors are emitted for them.
+_Attributes_ are reserved for future standardization.
+Other than checking for valid syntax, they SHOULD NOT 
+affect the processing or output of a _message_.
 
 ### Examples
 
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -652,6 +652,11 @@ MUST be an empty string.
 Implementations MAY offer functionality for customizing this,
 such as by emitting XML-ish tags for each _markup_.
 
+_Attributes_ MUST NOT have any impact on message formatting,
+and MUST NOT be available in the formatted output.
+_Variables_ in _attribute_ _values_ are not resolved,
+and no errors are emitted for them.
+
 ### Examples
 
 _This section is non-normative._
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -699,13 +699,16 @@ on the pairing, ordering, or contents of _markup_ during _formatting_.
 
 ## Attributes
 
-An **_<dfn>attribute</dfn>_** is a key with an optional value that
-is attached to the _expression_ or _markup_ as a whole.
-_Attributes_ are reserved for use by future specification versions.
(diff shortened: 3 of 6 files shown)


spec/data-model/README.md
```diff
@@ -219,7 +228,7 @@ that the implementation attaches to that _annotation_.
 ```ts
-  sigil: "!" | "@" | "%" | "^" | "&" | "*" | "+" | "<" | ">" | "?" | "~";
+  sigil: "!" | "%" | "^" | "&" | "*" | "+" | "<" | ">" | "?" | "~";
```
aphillips, line 231: 😉

spec/formatting.md
```diff
@@ -652,6 +652,11 @@ MUST be an empty string.
 Implementations MAY offer functionality for customizing this,
 such as by emitting XML-ish tags for each _markup_.
 
+_Attributes_ MUST NOT have any impact on message formatting,
+and MUST NOT be available in the formatted output.
+_Variables_ in _attribute_ _values_ are not resolved,
+and no errors are emitted for them.
```
aphillips, line 658: This might be too strict? We do need to account for syntax errors. I would probably allow the variable resolution error.

Perhaps:

```suggestion
_Attributes_ are reserved for future standardization.
Implementations MUST emit a syntax error for any 
_attribute_ that is not well-formed. 
Implementations SHOULD NOT evaluate the value of
an _attribute_, but MAY emit an _Unrevolved Variable error_ if
one is encountered.
_Attributes_ SHOULD be omitted from any _fallback values_.
No other errors are permitted.
```

spec/syntax.md
```diff
@@ -691,6 +697,39 @@ _Markup_ _placeholders_ can appear in any order without making the _message_ inv
 However, specifications or implementations defining _markup_ might impose requirements
 on the pairing, ordering, or contents of _markup_ during _formatting_.
 
+## Attributes
+
+An **_<dfn>attribute</dfn>_** is a key with an optional value that
+is attached to the _expression_ or _markup_ as a whole.
+_Attributes_ are reserved for use by future specification versions.
```
aphillips, line 704: I would put this sentence first and perhaps in bold in its own paragraph? I would make it more emphatic because we give examples later and we don't want to give the wrong impression.

```suggestion
**_Attributes_ are reserved for standardization by future versions of this specification.**
Examples in this section are meant to be illustrative and
might not match future requirements or usage.
```

spec/syntax.md
```diff
@@ -691,6 +697,39 @@ _Markup_ _placeholders_ can appear in any order without making the _message_ inv
 However, specifications or implementations defining _markup_ might impose requirements
 on the pairing, ordering, or contents of _markup_ during _formatting_.
 
+## Attributes
+
+An **_<dfn>attribute</dfn>_** is a key with an optional value that
+is attached to the _expression_ or _markup_ as a whole.
```
aphillips, line 703: This is what we intend to use it for, but I would probably not say it like this. Note that `key` should be `identifier` and that "the expression" is slightly quirky. Perhaps:

```suggestion
An **_<dfn>attribute</dfn>_** is an identifier with an optional value
that appears in an _expression_ or in _markup_.
```

spec/syntax.md
```diff
@@ -691,6 +697,39 @@ _Markup_ _placeholders_ can appear in any order without making the _message_ inv
...
+_Attributes_ are reserved for use by future specification versions.
+
+_Attributes_ are prefixed by a U+0040 COMMERCIAL AT `@` sign,
+followed by an _identifier_.
+An _attribute_ MAY have a _value_ separated from the _identifier_
```
aphillips, line 708: Add a couple of words for clarity.

```suggestion
An _attribute_ MAY have a _value_ which is separated from the _identifier_
```

aphillips (changes requested):
Good start. Some suggestions.

spec/formatting.md
eemeli: I don't think syntax errors need to be specifically called out here, as they would invalidate the parsing of the source as a _variable_ or an _attribute_, so we never even consider this.

Otherwise the strictness here is necessary to ensure that attributes do not at this time provide for any runtime visibility. Even checking for Unresolved Variables could be externally observable.

spec/syntax.md
```diff
@@ -691,6 +697,42 @@ _Markup_ _placeholders_ can appear in any order without making the _message_ inv
...
 on the pairing, ordering, or contents of _markup_ during _formatting_.
 
+## Attributes
+
+An **_<dfn>attribute</dfn>_** is an _identifier_ with an optional value
+that appears in an _expression_ or in _markup_.
+
+**_Attributes_ are reserved for standardization by future versions of this specification.**
+Examples in this section are meant to be illustrative and
+might not match future requirements or usage.
```
eemeli, line 707:
```suggestion
**_Attributes_ are reserved for standardization by future versions of this specification.**
Examples in this section are meant to be illustrative and
might not match future requirements or usage.

An **_<dfn>attribute</dfn>_** is an _identifier_ with an optional value
that appears in an _expression_ or in _markup_.
```

spec/formatting.md
aphillips: I'm not sure "runtime visibility" is a problem. We haven't defined with attributes are (at least formally). We've reserved a syntax and said it doesn't do anything yet. Maybe less is more?

```suggestion
_Attributes_ are reserved for future standardization.
Other than checking for valid syntax, they SHOULD NOT 
affect the processing or output of a _message_.
```

aphillips (approved):
Notwithstanding my comment, I'm going to add a 'ship it' here. I'd like to merge this with one more positive review.

PR #594, opened 2024-01-15
Add `:string` as formatter and selector

eemeli:
This implements the [proposed design](https://github.com/unicode-org/message-format-wg/blob/main/exploration/exact-match-selector-options.md#proposed-design), and marks the design as approved.

### DIFF
--- a/exploration/exact-match-selector-options.md
+++ b/exploration/exact-match-selector-options.md
@@ -1,6 +1,6 @@
 # Design Proposal Template
 
-Status: **Proposed**
+Status: **Acctepted**
 
 <details>
 	<summary>Metadata</summary>
--- a/exploration/exact-match-selector-options.md
+++ b/exploration/exact-match-selector-options.md
@@ -1,6 +1,6 @@
 # Design Proposal Template
 
-Status: **Acctepted**
+Status: **Accepted**
 
 <details>
 	<summary>Metadata</summary>
--- a/spec/registry.xml
+++ b/spec/registry.xml
@@ -253,7 +253,7 @@
 
   <function name="string">
     <description>
-      String formatting and selection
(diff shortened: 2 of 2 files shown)


exploration/exact-match-selector-options.md
```diff
@@ -1,6 +1,6 @@
 # Design Proposal Template
 
-Status: **Proposed**
+Status: **Acctepted**
```
aphillips, line 3:
```suggestion
Status: **Accepted**
```

spec/registry.xml
```diff
@@ -6,6 +6,7 @@
   <validationRule id="positiveInteger" regex="0|([1-9]\d*)"/>
   <validationRule id="currencyCode" regex="[A-Z]{3}"/>
   <validationRule id="timeZoneId" regex="[a-zA-Z/]+"/>
+  <validationRule id="anything" regex=".*"/>
```
aphillips, line 9: Technically this should be the same as a post-processed literal, which would omit surrogates. Probably not worth dinking with?

spec/registry.xml
```diff
@@ -250,4 +251,18 @@
     </alias>
   </function>
 
+  <function name="string">
+    <description>
+      String formatting and selection
```
aphillips, line 256: Perhaps a bit more detail?

```suggestion
Formatting of strings as a literal and selection based on string equality
```

aphillips (changes requested):
Need to fix the spelling error

spec/registry.xml
eemeli: That's taken care of by the syntax rule, so a literal value will never include surrogates.

aphillips (approved):

PR #597, opened 2024-01-15
Add missing hourCycle to :datetime timeStyle/dateStyle <formatSignature>

eemeli:
As discussed in https://github.com/unicode-org/message-format-wg/pull/570#discussion_r1452487765, the `hourCycle` option is also valid when using the `timeStyle` or `dateStyle` options of `:datetime`.

I suggest this is considered as an editorial change and fast-tracked.

### DIFF
--- a/spec/registry.xml
+++ b/spec/registry.xml
@@ -46,6 +46,11 @@
           such as "Asia/Shanghai", "Asia/Kolkata", "America/New_York".
         </description>
       </option>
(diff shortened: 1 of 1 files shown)


eemeli:
Because they're separate changes, and I wasn't sure if we had consensus on adding `:date` and `:time`.

aphillips:
We don't have consensus on `:date` and `:time`. Let's see if we can get there in the call?

PR #599, opened 2024-01-15
Update default registry design doc to reflect LDML45 scope

aphillips:
In the 2024-01-15 call we agree that the functions listed in this document are the comprehensive set for the `LDML45` default registry. This updates our documentation to reflect that.

The list of options, operands, etc. remain open for discussion, but no other functions will be considered.

### DIFF
--- a/exploration/default-registry-and-mf1-compatibility.md
+++ b/exploration/default-registry-and-mf1-compatibility.md
@@ -1,6 +1,6 @@
 # Default Registry and MF1 Compatibility
 
-Status: **Proposed**
+Status: **Accepted**
 
 <details>
 	<summary>Metadata</summary>
@@ -23,6 +23,12 @@ along with their _operands_ and _options_.
 It also contains a section comparing MF1 (as embodied by ICU4J) and MF2
 to ensure that we don't forget something.
 
+Implementations MAY provide other selectors or functions using the mechanisms defined by MF2
+but these will not be considered for the LDML45 default registry.
+The default registry is a foundational set of selectors and functions that an implementation
+claiming MF2 compatibility is required to supply.
+Therefore, addition to this list requires a higher level of rigor.
+
 ## Default Registry Entries
 
 ### Numbers
@@ -31,8 +37,6 @@ Function name: `:number`
 
 Aliases: 
 - `:integer` (implies: `maximumFractionDigits=0`)
-- `:currency` (implies: `style=currency`)
-- `:percent` (implies: `style=percent`)
 - `:plural` (no format, implies `select=plural` which is default)
 - `:ordinal` (implies: `select=ordinal`; we are missing `style=ordinal`)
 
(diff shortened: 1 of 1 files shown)


exploration/default-registry-and-mf1-compatibility.md
```diff
@@ -64,7 +68,7 @@ Options:
 Function name: `:datetime`
 
 Aliases:
-- (none currently; should there be `:date` and `:time` aliases to shorthand getting particularly timeStyle?)
+- `:date` and `:time` have been proposed and will be considered for 45
```
eemeli, line 71: I think we should be explicit here about whether these are in or out. Or is that intentionally left for later consideration? Also, as currently envisioned these aren't technically "aliases". Maybe "sub-functions"? I don't really care how they're categorised, mind.

```suggestion
- `:date` (with `style=...` option corresponding to `:datetime dateStyle=...`)
- `:time` (with `style=...` option corresponding to `:datetime timeStyle=...`)
```

exploration/default-registry-and-mf1-compatibility.md
```diff
@@ -23,6 +23,12 @@ along with their _operands_ and _options_.
 It also contains a section comparing MF1 (as embodied by ICU4J) and MF2
 to ensure that we don't forget something.
 
+Implementations MAY provide other selectors or functions using the mechanisms defined by MF2
+but these will not be considered for the LDML45 default registry.
+The default registry is a foundational set of selectors and functions that an implementation
+claiming MF2 compatibility is required to supply.
```
eemeli, line 29: Selectors are functions too, so probably better to use a more explicit term.

```suggestion
Implementations MAY provide other selectors or formatters using the mechanisms defined by MF2
but these will not be considered for the LDML45 default registry.
The default registry is a foundational set of selectors and formatters that an implementation
claiming MF2 compatibility is required to supply.
```

eemeli (commented):
I'm presuming that the "assignment" of this PR was intended as a request for reviews?

macchiati:
FYI Something that is often forgotten is that date literal does not map to a unique start/end unless it is accompanied by a timezone.

eemeli (approved):

aphillips:
@macchiati Oh yes, we know.

I'm engaged in an update of our old friend "Working With Timezones". The editor's copy is:

https://w3c.github.io/timezone/

This is one of the areas where we did significant work at Amazon on our message formatter. The ability to float and unfloat time values and manage time zones is important in being able to get consistent and accurate results. Sections 3 and 4 of the above document capture some of this work (although you have to read between the lines to see how it plays into MF2).

aphillips:
@macchiati 

(PS> Unless I'm missing something all of this is covered by the existing `options` on `:datetime`, `:date`, or `:time`. If something is missing, please raise an issue for it.)

aphillips:
Can I get an additional ship-it so we can merge this?

macchiati (approved):

PR #604, opened 2024-01-16
Add a JSON test suite

### LINKED ISSUE #109: Message Format - Test Suite
We should have a group of standard sentences that could be used as a test cases for all possible implementations of message format. They must cover all possible casuistics 

We have an example [here](https://github.com/unicode-org/message-format-wg/blob/2452264e693a4273c6418c598b71da873e1dab4c/exploration/variants.md)  for selectors.

eemeli:
Closes #109
CC @mradbourne

As discussed on yesterday's call, we should bootstrap our official test suite with the one I've been putting together for the JS `messageformat` implementation, as it's the most comprehensive one currently available.

For the avoidance of any doubt, I am the original author of these tests and I hereby relicense my work included in this PR under the Unicode License.

I've here moved the tests originally added by @grhoten that were in `test/` to `test/xml/`, as I wasn't sure whether we wanted to keep them as well.


test/json/README.md
```diff
@@ -0,0 +1,28 @@
+The files in this directory were originally copied from the [messageformat project](https://github.com/messageformat/messageformat/tree/11c95dab2b25db8454e49ff4daadb817e1d5b770/packages/mf2-messageformat/src/__fixtures)
+and are here relicensed by their original author (Eemeli Aro) under the Unicode License.
```
aphillips, line 2: I'm not sure if this is kosher or not. I will check with legal about the best way to handle this.

test/json/test-core.json
```diff
@@ -0,0 +1,140 @@
+[
+  { "src": "hello", "exp": "hello" },
+  { "src": "hello {world}", "exp": "hello world" },
```
aphillips, line 3: comments would be useful in this file. It took me a second to remember that `world` is an unquoted.

test/json/test-core.json
```diff
@@ -0,0 +1,140 @@
...
+    "src": "{$one} and {$two}",
+    "params": { "one": 1.3, "two": 4.2 },
+    "exp": "1.3 and 4.2"
+  },
+  {
+    "src": "{$one} et {$two}",
+    "locale": "fr",
+    "params": { "one": 1.3, "two": 4.2 },
+    "exp": "1,3 et 4,2"
+  },
```
aphillips, line 41: These test something other than the `:string` formatter. It might be a good idea to move them with others of the same stripe or segregate them.

test/json/test-core.json
```diff
@@ -0,0 +1,140 @@
...
+  { "src": ".local $foo = {bar} {{bar {$foo}}}", "exp": "bar bar" },
+  { "src": ".local $foo = {|bar|} {{bar {$foo}}}", "exp": "bar bar" },
+  {
+    "src": ".local $foo = {|bar|} {{bar {$foo}}}",
```
aphillips, line 45: Might be nice to have some whitespace tests with assignments and selectors. I don't see them elsewhere in this file?

test/json/test-core.json
```diff
@@ -0,0 +1,140 @@
...
+      {
+        "type": "markup",
+        "kind": "open",
+        "name": "tag",
+        "options": { "foo": "bar" }
+      }
+    ]
+  },
+  {
+    "src": "{#tag a:foo=|foo| b:bar=$bar}",
```
aphillips, line 118: Missing standalone tags?

test/json/test-core.json.d.ts
```diff
@@ -0,0 +1,21 @@
+export type TestMessage = {
+  /** The MF2 syntax source. */
```
aphillips, line 2:
```suggestion
 /** The MF2 message to be tested. */
```

test/json/test-core.json.d.ts
```diff
@@ -0,0 +1,21 @@
...
+  src: string;
+  /** The locale to use for formatting. Defaults to 'en-US'. */
```
aphillips, line 4: Did you consider defaulting to `und`? The root locale is US English-like but also somewhat invariant.

test/json/test-core.json.d.ts
```diff
@@ -0,0 +1,21 @@
...
+  locale?: string;
+  /** Parameters to pass in to the formatter for resolving external variables. */
+  params?: Record<string, string | number | null | undefined>;
```
aphillips, line 7: include date type here?

test/json/test-functions.json
```diff
@@ -0,0 +1,259 @@
+{
+  "integer": [
+    { "src": "hello {4.2 :integer}", "exp": "hello 4" },
+    { "src": "hello {-4.20 :integer}", "exp": "hello -4" },
+    { "src": "hello {0.42e+1 :integer}", "exp": "hello 4" },
+    {
+      "src": ".match {$foo :integer} one {{one}} * {{other}}",
```
aphillips, line 7: when writing plural tests, I often find it a good idea to build out the variant matrix. Yes, `en-US` (or `und` for that matter) won't touch the `two` or `few` or whatever keyword. But it can help catch errors if one of the other keywords triggers incorrectly. It also allows the test itself to be recycled for other locales.

test/json/test-functions.json
```diff
@@ -0,0 +1,259 @@
...
+      "exp": "=1.2"
+    },
+    {
+      "src": ".match {$foo :integer} 1.2 {{=1.2}} |1,2| {{=1,2}} * {{other}}",
+      "locale": "fr",
+      "params": { "foo": 1.2 },
+      "exp": "=1.2"
+    }
+  ],
+  "number": [
```
aphillips, line 23: I would move `:number` ahead of `:integer`. Note that you don't test any of `:integer`'s options, e.g. `minimumIntegerDigits`, `signDisplay`, et al.

aphillips (commented):
Thank you! Some general observations follow.

test/json/test-core.json.d.ts
eemeli: Alas, JS doesn't support an `und` locale.

test/json/test-core.json.d.ts
eemeli: Nope, because this is JSON data which doesn't inherently support dates.

eemeli:
@aphillips I would very much prefer to not touch the data files at all in this PR, even though as you've pointed out there are a few cases where the tests are e.g. incorrectly categorised, and definitely incomplete.

As we squash our commits, that will allow this one PR to be the place where we adopt the `messageformat` test suite and for which we need to make sure that any copyright assignment or licensing is done right. Then subsequent PRs on top of this will be done wholly under the Unicode license.

test/json/README.md
aphillips: Based on our discussion with Anne (legal), as sole author you can commit this file, relying on the CLA to cover it.

aphillips (approved):
Per your comment, we can commit this without addressing my comments.

It would make sense to do a more thorough job of it, though. Should we work on these files or something else?

test/json/test-core.json.d.ts
aphillips: It works in Chrome. It also works in FF, except that the locale's name data isn't present.

eemeli: It only looks like it works, though. If you try this, for instance, you'll get your default locale (probably `en-US`) rather than `und`:
```js
new Intl.NumberFormat('und').resolvedOptions().locale
```

eemeli:
> Should we work on these files or something else?

My recommendation would be to get these in, and then start iterating on them. Starting from a copy has the benefit of having at least one implementation for which all of these tests pass, which makes it easier to validate any changes.

grhoten:
> I've here moved the tests originally added by @grhoten that were in `test/` to `test/xml/`, as I wasn't sure whether we wanted to keep them as well.

I say delete them. They don't fit well into the current design of MF2, and the files can always be resurrected from the history if necessary.

test/json/test-functions.json
```diff
@@ -0,0 +1,259 @@
...
+    {
+      "src": ".match {$foo :ordinal} one {{st}} two {{nd}} few {{rd}} * {{th}}",
+      "params": { "foo": 3 },
+      "exp": "rd"
+    },
+    {
+      "src": ".match {$foo :ordinal} one {{st}} two {{nd}} few {{rd}} * {{th}}",
+      "params": { "foo": 4 },
+      "exp": "th"
+    },
```
grhoten, line 213: Do you have the ability to reference ICU RBNF functionality?  ICU MessageFormat could reference them as the spellout and ordinal, which conflicts with this plural rule ordinal.

See here for reference: https://unicode-org.github.io/icu-docs/apidoc/released/icu4c/classMessageFormat.html

Please ignore duration from MessageFormat. It's deprecated, and you need a decent message formatting framework to do it right.

aphillips: You're missing the point of this test, which is testing the `ordinal` selector (== `selectordinal` in MF1). @eemeli  has used the English ordinal suffixes in the test for readability in a test that is certainly not ready for any locale other than en-US.

Note well that we are considering spellout, ordinal, and duration for inclusion into the default registry as formatters--in another set of issues and design documents. I recommended against including spellout and duration in the default registry. `ordinal` as a formatter is on the bubble (only ICU4J has it, not Intl or ICU4X) and there are obvious functionality problems with ordinals in the current implementation.

eemeli:
@aphillips Applied the updates discussed on the call; this should now be good to merge.

PR #607, opened 2024-01-18
Update why_mf_next.md

macchiati:
Make some spot fixes to the doc to update for reviewers. Some of the text is outdated, and/or needed some clarifications.

### DIFF
--- a/docs/why_mf_next.md
+++ b/docs/why_mf_next.md
@@ -57,10 +57,12 @@ with existing tools.
 ICU is old, but also very popular (right now it is the core i18n library
 for all major operating systems, browsers, and many other products).
 
-This is how it has both numeric and named parameters, allows partial strings in
-plural / select (technically concatenation, which is bad i18n), uses date / time
-patterns (bad i18n, when skeletons are the better way), has nesting selectors,
-and unfriendly escaping (think doubling the apostrophe `''` ), `#` in plurals.
+As a result of its age and design, MessageFormat 1.0 has both numeric (positional)
+ and named parameters.
+It still provides date and time patterns (picture strings), when skeletons or option
+bags provide far superior results.
+It allows selectors (such as plural and select) around only part of the overall message,
+which is a form of non-internationalized string concatenation.
 
 Most of it can't be “blamed” directly on a bad decision, it is just time
 teaching us what works (for instance skeletons did not exist when the
--- a/docs/why_mf_next.md
+++ b/docs/why_mf_next.md
@@ -86,7 +86,7 @@ But the stability requirements prevent any major cleanup.
   there are no guardrails to keep people from using these poorly
   internationalized features.
 - Unable to support grammatical or personal gender selection well. 
-  Using existing formatters such as `select` doesn't account
+  Existing selectors such as `select` cannot account
   for the grammatical needs of different gender categories across languages.
   Tools have no way to know what modifications are needed
   and developers have to understand the needs of current and future languages to succeed.
--- a/docs/why_mf_next.md
+++ b/docs/why_mf_next.md
@@ -8,8 +8,9 @@ Intro
 
 - `MessageFormat` is the Unicode API for software localization
 - It is 20 years old, well designed, proven solution
-- Its design is optimized for the software development model of 20y ago,
-  and implementers struggle with its shortcomings.
+Its design was optimized for the software development model
+of twenty years ago.
+Implementers, developers, and translators struggle with its shortcomings.
 
 The current wave of software development uses dynamic languages, modern UI
 frameworks and new forms of user interactions (voice, VR etc.).
--- a/docs/why_mf_next.md
+++ b/docs/why_mf_next.md
@@ -79,7 +79,7 @@ But the stability requirements prevent any major cleanup.
   grammatical agreement requires words outside select / plural to change.
   See https://en.wikipedia.org/wiki/Agreement_(linguistics)
 - Patterns in the date / time / number placeholders are bad i18n, should use skeletons
-- Unable to support gender selection. 
+- Unable to support gender selection well. 
   Using existing formatters such as `select` doesn't account
   for the grammatical needs of different gender categories across languages.
   Tools have no way to know what modifications are needed
--- a/docs/why_mf_next.md
+++ b/docs/why_mf_next.md
@@ -79,9 +79,11 @@ But the stability requirements prevent any major cleanup.
   grammatical agreement requires words outside select / plural to change.
   See https://en.wikipedia.org/wiki/Agreement_(linguistics)
 - Patterns in the date / time / number placeholders are bad i18n, should use skeletons
-- No official support for gender. It can be done with `select`, but that has problems for localization tooling:
-  strings can't account for the difference in gender categories across languages, and there are no standard “enum” identifiers:
-  developers use masculine/feminine, masc/fem, male/female, etc.
+- Unable to support gender selection. 
+  Using existing formatters such as `select` doesn't account
+  for the grammatical needs of different gender categories across languages.
+  Tools have no way to know what modifications are needed
+  and developers have to understand the needs of current and future languages to succeed.
 - Formatting for “literals” known at compile time.
   For example, literal numbers in a message need to be adjusted for the user's preferences for native vs ASCII digits and format.
 - Escaping with apostrophe is error prone. There is no reliable way to tell if
--- a/docs/why_mf_next.md
+++ b/docs/why_mf_next.md
@@ -126,13 +126,17 @@ localization flow:
 ### 4. Designed to be API only, plain text, UI, “imperative style”
 
 The typical use case for `MessageFormat` is: load the string from resources,
-replace placeholders, and return the string result with placeholders replaced. \
+replace placeholders, and return the string result with placeholders replaced.
 An i18n-aware `printf`, basically.
 
 It does not account for formatting tags (such as HTML),
 or “document-like” content (for example templating systems like
 [freemarker](https://freemarker.apache.org/),
 [mustache](https://mustache.github.io/), even JSP, PHP, etc.)
+
+Because it is only an API, it has no standard way to store strings in a
+serialized format or to carry information or directives for translators
+or localization tools.
 
 There is also no metadata “packaging”: comments, length limits, example, links,
 protecting non-translatable sections of text, etc.
@@ -146,9 +150,9 @@ Applications don't need to migrate all the strings to a new format and resource
 resolution only to support some more advanced features in a few messages.
 
 And since the string loading is left to the underlying tech stack it means that
-the locale resolution and fallback is consistent with everything else. \
-For example in Android there is locale based selection (with fallback) for
-styles, images, sounds, any kind of assets. \
+the locale resolution and fallback is consistent with everything else. 
+For example in Android there is locale-based selection (with fallback) for
+styles, images, sounds, any kind of assets. 
 So there is no risk that the string fallback is different than the sound
 fallback, for example.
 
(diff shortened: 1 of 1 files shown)


docs/why_mf_next.md
```diff
@@ -44,20 +44,21 @@ It also means most tools used to process these messages are built rigidly,
...
 (think localization tools, linters, friendly UIs, etc.).
 
-The most basic functionality would be adding a new formatter. Meantime ICU
-added other formatters: time intervals, measurement, lists. But MessageFormat
-did not keep up. And adding support for any of these new formats risks to break
+The most basic functionality would be adding a new formatter.
+There are a relatively small number of supported formatters, while over the years ICU
+added other formatters: date and time intervals, measurement units, lists, person names, and many others. 
+But MessageFormat did not keep up. And adding support for any of these new formats risks to break
 existing tools.
```
aphillips, line 51: This paragraph is difficult to read and less clear than it needs to be.

```suggestion
The most basic functionality would be adding a new formatting function.
MessageFormat 1.0 only supported a small number of basic formatting functions,
while over the years ICU added many new capabilities: date and time intervals,
measurement units, lists, person names, and many more.
Developers also sometimes want to define their own formatting functions.
Supporting additional formats risks breaking interoperability or compatibility
with existing tools.
```

docs/why_mf_next.md
```diff
@@ -44,20 +44,21 @@ It also means most tools used to process these messages are built rigidly,
...
+for all major operating systems, browsers, and many other products).
 
-This is how he have both numeric and named parameters, partial strings in
-plural / select (technically concatenation, which is bad i18n), date / time
-patterns (bad i18n, when skeletons are the better way), nesting selectors,
-unfriendly escaping (think doubling the apostrophe `''` ), `#` in plurals.
+This is how it has both numeric and named parameters, allows partial strings in
+plural / select (technically concatenation, which is bad i18n), uses date / time
+patterns (bad i18n, when skeletons are the better way), has nesting selectors,
+and unfriendly escaping (think doubling the apostrophe `''` ), `#` in plurals.
```
aphillips, line 61: Translation into English:

```suggestion
As a result of its age and design, MessageFormat 1.0 has both numeric (positional)
 and named parameters.
It still provides date and time patterns (picture strings), when skeletons or option
bags provide far superior results.
It allows selectors (such as plural and select) around only part of the overall message,
which is a form of non-internationalized string concatenation.
```

docs/why_mf_next.md
```diff
@@ -74,9 +75,11 @@ But the stability requirements prevent any major cleanup.
   grammatical agreement requires words outside select / plural to change.
   See https://en.wikipedia.org/wiki/Agreement_(linguistics)
 - Patterns in the date / time / number placeholders are bad i18n, should use skeletons
-- No official support for gender. It can be done with `select`, but it
-  is not the same thing (same as the difference between an `enum` and integer/strings). Developers can use masculine/feminine, masc/fem, male/female, etc.
-- Formatting for “parameters” known at compile time
+- No official support for gender. It can be done with `select`, but that has problems for localization tooling:
+  strings can't account for the difference in gender categories across languages, and there are no standard “enum” identifiers:
+  developers use masculine/feminine, masc/fem, male/female, etc.
```
aphillips, line 80: This has: double colons: and needs more work. Also, we don't support gender in 2.0-45 either.

```suggestion
- Unable to support gender selection. 
  Using existing formatters such as `select` doesn't account
  for the grammatical needs of different gender categories across languages.
  Tools have no way to know what modifications are needed
  and developers have to understand the needs of current and future languages to succeed.
```

docs/why_mf_next.md
```diff
@@ -74,9 +75,11 @@ But the stability requirements prevent any major cleanup.
...
+- Formatting for “literals” known at compile time.
+  For example, literal numbers in a message need to be adjusted for the user's preferences for native vs ASCII digits and format.
```
aphillips, line 82: I don't understand what this is supposed to represent?

docs/why_mf_next.md
```diff
@@ -85,17 +88,17 @@ But the stability requirements prevent any major cleanup.
 
 ### 3. Hard to map to the existing localization core structures
 
-The format is not well supported by any major localization system. \
+The format is not easy to support in localization tooling.
```
aphillips, line 91: No. This is precisely the opposite of what the rest of the paragraph says (and is not correct anyway).

```suggestion
While MessageFormat 1.0 and its syntax are widely supported by runtime environments,
the same cannot be said for localization tooling.

File: docs/why_mf_next.md [hunk 6]
```diff
@@ -85,17 +88,17 @@ But the stability requirements prevent any major cleanup.
...
-and gives back a translated string, usually with the same placeholders
+and give back a translated string, usually with the same placeholders
 (with some degree of flexibility).
 
-It makes it very difficult to translate things like plurals, where the input
-has (for example) 2 “message variants” (English, 1 / many, singular / plural),
-and return 4 message variants for Russian, for example.
+It requires that the translation software expand plurals, where the input
+has (for example) 2 “message variants” (English, singular / plural),
+but the message presented to a Arabic translator needs 6 message variants.
```
aphillips, line 101: This change makes the text less clear.

```suggestion
To get the right results, translation software needs to understand the message syntax.
For example, it needs to adjust the number of translated "patterns" to match the
grammatical needs of the target language.
Where the English input might have only two patterns (singular and plural),
the Arabic translator needs to supply six message variants,
and the Japanese translator only one.
```

aphillips (dismissed):
This is probably a good idea to undertake. I wish you had added it to my PR's branch, but this is fine too.

docs/why_mf_next.md
macchiati: > Unable to support gender selection. 
That is misleading, because people do use select for gender support. It just supports is badly!
Suggest:

Gender selection is badly supported.

docs/why_mf_next.md
macchiati: This the case where you have Arabic *literal* text in "There are 98 departments" (note no input placeholders). There is no way in MF1.0 to change the 98 to Arabic numerals (from ASCII) according to user preferences.

macchiati: I didn't understand the original text, so tried to reformulate it into something that made sense to me.

macchiati: That was the line "Formatting for “parameters” known at compile time"

macchiati: If you have suggested for clarity ... Or we could just delete "Formatting for “parameters” known at compile time" and not replace by "Formatting for "literals"..."

macchiati:
> This is probably a good idea to undertake. I wish you had added it to my PR's branch, but this is fine too.

Sorry for that!

docs/why_mf_next.md
aphillips: I think we should just remove this item.

macchiati: done

macchiati:
Fixed the last outstanding lines so should be good to go, once approved.

docs/why_mf_next.md
```diff
@@ -8,8 +8,8 @@ Intro
 
 - `MessageFormat` is the Unicode API for software localization
 - It is 20 years old, well designed, proven solution
-- Its design is optimized for the software development model of 20y ago and its
-  shortcomings result in mixed reception and adoption by the industry.
+- Its design is optimized for the software development model of 20y ago,
+  and implementers struggle with its shortcomings.
```
aphillips, line 12:
```suggestion
Its design was optimized for the software development model
of twenty years ago.
Implementers, developers, and translators struggle with its shortcomings.
```

aphillips (approved):

aphillips:
I am tempted to add a link to the UTW video.

https://www.youtube.com/watch?v=-DlS6KNopoU 

Otherwise ready. @macchiati I did some edits. If you'd review quickly (or if @eemeli gets to it first), then we can merge.

macchiati:
Excellent idea to add a link!

eemeli (changes requested):
See inline. Mostly editorial nitpicks, the block here is for resolving the discussion on the dropped section 4 paragraph.

docs/why_mf_next.md
```diff
@@ -108,23 +125,16 @@ localization flow:
...
 or “document-like” content (for example templating systems like
 [freemarker](https://freemarker.apache.org/),
 [mustache](https://mustache.github.io/), even JSP, PHP, etc.)
 
-Because it is API only it has no standard way to store the stings in a
-serialized format and to carry info or directives for translators or
-localization tools. \
-So there is no way for a message to reference another message, or to fallback
-to a different locale. That is all left to the "host resource manager"
-(whatever that is for the given tech stack)
```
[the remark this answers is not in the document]
aphillips: I agree about the first part.

I'm not sure I agree about the second part, which is a description of resource management and/or resource files. There are lots of formatter adjacent things that are not in scope for MF2 itself (and which MF1 never pretended to solve). I can't use 'suggestion' to restore text, so will separately commit.

docs/why_mf_next.md
```diff
@@ -68,34 +79,45 @@ But the stability requirements prevent any major cleanup.
 ### 2. Some existing problems
 
 - ICU added new formatters, but MessageFormat does not support them
-- Combined selectors (select + plural) results in unreadable and error
-  prone nesting
-- Select and plurals inside the message are difficult to translate because of
-  grammatical agreement requires words outside select / plural to change.
+- Messages with selectors (`select` and/or `plural`) are difficult to create
+  and edit because of the complex nesting requirements of the syntax.
+- `select` and `plural` placeholders inside a message are difficult to translate as
```
macchiati, line 84:
```suggestion
- `select` and `plural` placeholders inside a message are difficult to translate as
  grammatical agreement may require words _outside_ the `select`/`plural` to change.
```
Remove this item since it is covered above. (I wouldn't weigh too heavily on this, as systems can enforce "no select on parts", as Google does.)

docs/why_mf_next.md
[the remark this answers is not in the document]
macchiati: I think the first can be restored, but should be changed. There is a standard string serialization defined by the API, used all the time. The point to make is that it doesn't the kinds of information needed, so people have developed non-interoperable wrappers for it.
OLD
Because it is API only it has no standard way to store the stings in a
serialized format and to carry info or directives for translators or
localization tools.
NEW
The format defined by the API provides no standard way to carry the necessary info or directives for translators or localization tools, so implementations have had to develop their own, with no interoperability.

docs/why_mf_next.md
```diff
@@ -108,23 +130,20 @@ localization flow:
...
-localization tools. \
-So there is no way for a message to reference another message, or to fallback
-to a different locale. That is all left to the "host resource manager"
-(whatever that is for the given tech stack)
+Because it is only an API, it has no standard way to store strings in a
+serialized format or to carry information or directives for translators
+or localization tools.
 
-There is also no metadata: comments, length limits, example, links,
+There is also no metadata “packaging”: comments, length limits, example, links,
```
macchiati, line 146: This is a duplicate of the above; should be combined. (There are other duplicates in the text, mostly harmless and would take additional work to fix; given that we want to get this out pronto I didn't suggest changes.)

docs/why_mf_next.md
aphillips: They aren't the same thing. Nesting happens when you don't do select on parts but do have multiple selection. Coupled with the need to "explode" plurals, things can get messy. (At Amazon we used our resource format to break up the nesting, making it easier to write).

This second bullet is about naive use of MF1 (avoiding nesting works in English, but doesn't translate). You or I can enforce this in our own tooling, but MF1 itself does not.

In any case, I think this is fine?

docs/why_mf_next.md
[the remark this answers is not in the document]
aphillips: I will add your text.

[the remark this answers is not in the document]
macchiati: Thanks

docs/why_mf_next.md
macchiati: Ok for now, thanks.

macchiati:
If not, another PR can be opened.

PR #609, opened 2024-01-18
Macchiati patch 2

macchiati:


aphillips:
@macchiati What you wanted to do here was commit to my branch `aphillips-readme-update` (**_not_** make a PR). That would add the additional changes to my existing PR (and make you co-author of my PR). Now we have two PRs of mostly the same stuff. I can kill my PR and we can review here.

docs/goals.md
```diff
@@ -12,43 +12,43 @@ of some data value or values.
 
 The design goals are listed below.
 
-1.  Express grammatical features, such as plurals, genders, and inflections.
+1.  Express grammatical features, such as plurals and genders.
```
aphillips, line 15: We aren't supporting genders in this release. Either we should remove that too or we can keep inflections.

docs/goals.md
```diff
@@ -12,43 +12,43 @@ of some data value or values.
...
 
 2.  Express other variance in translation, due to linguistic and regional
     features, the presentation media, context, circumstance, and other factors.
+    _Note: not part of the LDML45 release._
```
aphillips, line 19: I'm not sure that we aren't providing this. Admittedly we aren't providing explicit support for e.g. display context, but one can use `:string` (_SelectFormat_) to do most of these.

To be honest, I'm not sure why this isn't just part of the first item.

docs/goals.md
```diff
@@ -12,43 +12,43 @@ of some data value or values.
...
 
-3.  Express formattable data, such as numbers, dates, currencies, or units,
+4.  Express formattable data, such as numbers, dates, currencies, or units,
```
aphillips, line 21: Don't renumber

docs/goals.md
```diff
@@ -12,43 +12,43 @@ of some data value or values.
...
     and optimized for different uses and audiences.
 
 ## Deliverables
 
 1.  A formal definition of the canonical data model for representing
     localizable _dynamic message strings_.
+    _Note: not part of the LDML45 release._
 
-2.  A formal definition of the canonical syntax for representing the data
+3.  A formal definition of the canonical syntax for representing the data
```
aphillips, line 38: Repair numbering

aphillips (changes requested):

docs/goals.md
macchiati: Let's remove then. Only a "such as"

docs/goals.md
macchiati: It is quite broad: for example, "linguistic" could reasonably include gender, other noun class, inflections, general agreement. And those are problematic because the tooling needs to do expansion/contraction. Unclear whether other items also need to be "enums" that expand/contract based on locale (context? Does that include formality/linguistic register?). Anyway, this list is far to misleading.

If you really mean "stuff that can be selected with :string (SelectFormat)", then that should be explicit so that this line doesn't mislead people.

macchiati: Or just delete the line. We aren't listing all of the design features anyway.

macchiati:
> I can kill my PR and we can review here.

Sorry, I misunderstood the process you were using. What's the best way to recover?

aphillips:
> Sorry, I misunderstood the process you were using. What's the best way to recover?

Don't worry about it. I merged my PR, since it had enough approval and fixed this one (including giving it a better title). We can just iterate on this PR.

docs/goals.md
aphillips:
```suggestion
1. Express grammatical features. For example: plurals.
```

docs/goals.md
macchiati: Deleting to resolve conversation.

aphillips: I think, actually, that this item and the first item might be better expressed as:

```suggestion
1. Allow users to write messages that are both grammatically
   correct and can be translated in a grammatically correct manner
   to languages that have different requirements. For example, plurals.
```

docs/goals.md
```diff
@@ -12,16 +12,12 @@ of some data value or values.
 
 The design goals are listed below.
 
-1.  Express grammatical features, such as plurals, genders, and inflections.
-
-2.  Express other variance in translation, due to linguistic and regional
-    features, the presentation media, context, circumstance, and other factors.
+1. Express grammatical features. For example: plurals.
```
aphillips, line 15: Bringing my suggestion back.

```suggestion
1. Allow users to write messages that are both grammatically
   correct and can be translated in a grammatically correct manner
   to languages that have different requirements. For example, plurals.
```

aphillips (approved):

docs/goals.md
```diff
@@ -33,17 +31,17 @@ The design goals are listed below.
 
 1.  A formal definition of the canonical data model for representing
     localizable _dynamic message strings_.
+    _Note: not part of the LDML45 release._
```
eemeli, line 34: Why not? It's right here: https://github.com/unicode-org/message-format-wg/tree/main/spec/data-model

```suggestion
```

docs/goals.md
```diff
@@ -12,16 +12,14 @@ of some data value or values.
 
 The design goals are listed below.
 
-1.  Express grammatical features, such as plurals, genders, and inflections.
-
-2.  Express other variance in translation, due to linguistic and regional
-    features, the presentation media, context, circumstance, and other factors.
+1. Allow users to write messages that are both grammatically
+   correct and can be translated in a grammatically correct manner
+   to languages that have different requirements. For example, plurals.
```
eemeli, line 17: Accepting just "plurals" as a sufficient example seems to expect a lot of the reader. Should this sentence fragment be expanded?

eemeli (changes requested):
I'm mostly fine with these, but changing our goals and deliverables should be discussed explicitly at a weekly call.

docs/goals.md
macchiati: I'm ok with this as a goal. I have no doubt that we make huge progress towards this with MF2.0. (I can relay caveats elsewhere.)

macchiati:
> I'm mostly fine with these, but changing our goals and deliverables should be discussed explicitly at a weekly call.

We can wait on this for Monday's meeting.

docs/goals.md
```diff
@@ -12,16 +12,16 @@ of some data value or values.
...
-
-2.  Express other variance in translation, due to linguistic and regional
-    features, the presentation media, context, circumstance, and other factors.
+1. Allow users to write messages that are both grammatically
+   correct and can be translated in a grammatically correct manner
+   to languages that have different requirements.
+   For example, providing a mechanism that uses CLDR's plural rules
+   to select between various strings to use based on a numeric input value.
 
 3.  Express formattable data, such as numbers, dates, currencies, or units,
```
mihnita, line 21: Nitpick: renumber. Now we have 1, 3, 4, 5, 6

PR #611, opened 2024-01-19
Replace 'shape' with better terminology

aphillips:
Addresses #590

Replaces the amorphous term "shape" in Formatting with more appropriate text. Generally I used a reformulation using some variation of the word "form".

In one case, I deleted the paragraph. The paragraph in question was non-normative and basically said the same thing as a normative paragraph just below it--that how formatting works is implementation-defined.

---

In fixing this, I corrected one typo.

I also transformed the eager/lazy note into an "IMPORTANT" (which also restores it's intended normativity)

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -16,13 +16,15 @@ Formatting of a _message_ is defined by the following operations:
   with reference to the current _formatting context_.
   This can include multiple steps,
   such as looking up the value of a variable and calling formatting functions.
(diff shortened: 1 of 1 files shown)


macchiati (approved):

aphillips:
@macchiati I added a normative change, so canceling your previous approve.

macchiati (approved):

eemeli (approved):

PR #614, opened 2024-01-23
Correct bogus ABNF syntax matching literal numbers

alerque:
The `0` here was evidently meant to represent a literal character 0, but as I understand ABNF this would be interpreted as a counter modifier for a later item. It would be quoted as a string `"0"`, but the later range `[0-9]` is not quite so easy. To be a range one kind of needs to specify the actual character codes (as is done elsewhere in this very grammar). I encoded the zero as `%x30` and the range one-to-nine as `%x31-39` so all the literal digits are represented the same way.


gibson042 (approved):
Wow, I'm surprised we missed this for so long. Thanks!

aphillips:
@alerque Thank you so much!

Please sign the CLA or contact me if you have any questions/problems.

alerque:
Signed.

This bogus syntax was introduced in #553.

PR #613, opened 2024-01-23
Beautify the ABNF

aphillips:
Fixes #600 

As we lock down the ABNF, start to fix spacing and alignment. In this PR all of the changes are editorial.

Note that I have moved a few of the productions in an attempt to organize items more logically. I have also added comments for individual blocks.

PR #615, opened 2024-01-23
Port ABNF fix to relevant mention in syntax document

alerque:
Follow up to #614.

I also notice some of the beautification from #613 didn't land here so maybe you have other plans to update these code blocks to match, but I didn't bother trying to apply any of them as they *should be* only decorative and I didn't know if you wanted them in an unrelated PR. This change fixing an actual syntax problem needs updating even if the formatting changes aren't urgent to touchup.

### DIFF
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -789,7 +789,7 @@ quoted-char    = content-char / s / "." / "@" / "{" / "}"
 
 unquoted       = name
                / number-literal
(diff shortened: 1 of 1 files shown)


aphillips (approved):

aphillips:
This is fine. The beautification was landed in the ABNF first. The copies in syntax.md are informative.

gibson042 (approved):

aphillips:
fast-track requirements met

PR #618, opened 2024-01-26
Clean up MF1 compatibility matrix

eemeli:
The important part here is dropping the `{$num :percent}` and `{$num :currency}` examples, but while at it I added some `:date` and `:time` examples and tried to also otherwise prettify it.

Also added `.match {$num :number select=exact}` as a solution for some MF1 `select` usage.

### DIFF
--- a/exploration/default-registry-and-mf1-compatibility.md
+++ b/exploration/default-registry-and-mf1-compatibility.md
@@ -131,7 +131,7 @@ How to write an MF1 format or selector in MF2:
 | Number             | `{num,number}`                 | `{$num :number}`                                                    |                                 |
 | Integer            | `{num,number,integer}`         | `{$num :integer}`<br/>`{$num :number maximumFractionDigits=0}`      |                                 |
 | Percent            | `{num,number,percent}`         | `{$num :number style=percent}`                                      |                                 |
-| Currency           | `{num,number,currency}`        | `{$num :number currency=$code}`                                     |                                 |
+| Currency           | `{num,number,currency}`        | `{$num :number style=currency currency=$code}`                      |                                 |
 | Plural (selector)  | `{num,plural, ...}`            | `.match {$num :plural}`<br/>`.match {$num :number}`                 |                                 |
 | Ordinal (selector) | `{num,selectordinal, ...}`     | `.match {$num :ordinal}`<br/>`.match {$num :number select=ordinal}` |                                 |
 | Ordinal (format)   | `{num,ordinal}`                |                                                                     | missing                         |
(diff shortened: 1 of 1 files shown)


exploration/default-registry-and-mf1-compatibility.md
```diff
@@ -124,23 +124,23 @@ The following functionality was deliberately excluded:
...
-| Datetime | {date,time,::skeleton} | {$date :datetime weekday=short etc.}                       | supported through options bag |
-| Spellout | {num,spellout}       |                                                              | missing |
-| Duration | {num,duration}       |                                                              | missing |
-| Choice   | {num,choice, ...}    |                                                              | deprecated in MF1 |
+| MF1                | Syntax                         | MF2                                                                 | Comment                         |
+| ------------------ | ------------------------------ | ------------------------------------------------------------------- | ------------------------------- |
+| String             | `{var}`                        | `{$var}`<br/>`{$var :string}`                                       |                                 |
+| Select             | `{var,select, ...}`            | `.match {$var :string}`<br/>`.match {$num :number select=exact}`    |                                 |
```
aphillips, line 130: I don't agree with this change. `{$num :number select=exact}` is not quite the same semantic, I suspect. Some implementations will use numeric equality rather than string equality. Consider:

```
.input {$num :number minimumFractionDigits=2}
.match {$num :number select=exact}
1   {Does this match the value 1.00?}
2.3 {Does this match the value 2.30?}
```

exploration/default-registry-and-mf1-compatibility.md
```diff
@@ -124,23 +124,23 @@ The following functionality was deliberately excluded:
...
+| Number             | `{num,number}`                 | `{$num :number}`                                                    |                                 |
+| Integer            | `{num,number,integer}`         | `{$num :integer}`<br/>`{$num :number maximumFractionDigits=0}`      |                                 |
+| Percent            | `{num,number,percent}`         | `{$num :number style=percent}`                                      |                                 |
+| Currency           | `{num,number,currency}`        | `{$num :number currency=$code}`                                     |                                 |
```
aphillips, line 134: We need, probably, to discuss currency further. This is fine, tho.

eemeli: Actually, I just noticed that this is missing an option.

```suggestion
| Currency           | `{num,number,currency}`        | `{$num :number style=currency currency=$code}`                      |                                 |
```

exploration/default-registry-and-mf1-compatibility.md
eemeli: We should write a spec description of `:number` selection; that probably should not be an implementation detail.

mihnita:
I don't think we should add aliases.
We have the functionality without them.

aphillips:
accepted in 2024-01-29 call. addison to follow up with issue about select=exact

PR #621, opened 2024-02-03
Describe number selection fully

aphillips:
Update the design document as part of my action item in order to describe number selection completely per last week's call. The core text is intended to go somewhat verbatim into the spec.


exploration/number-selection.md
```diff
@@ -92,16 +95,78 @@ ICU MF1 messages using `plural` and `selectordinal` should be representable in M
...
+
+### Selection
+
+When performing selection, the resolved value of the `operand` is compared to 
+each of the variant keys.
+The result of such a comparison can be one of the following:
+
+- If the value of the key does not match the production `number-literal`
+  or is not one of the plural/ordinal keywords, the key is invalid;
+  return a _Selection Error_.
```
eemeli, line 152: This should specify the fallback behaviour, so probably we should emit the error and then ignore the invalid key in further processing?

exploration/number-selection.md
```diff
@@ -92,16 +95,78 @@ ICU MF1 messages using `plural` and `selectordinal` should be representable in M
...
+- If the value of the key matches the production `number-literal`:
+  - If the parsed value of the key is an [exact match](#determining-exact-literal-match)
+    of the value of the operand, then the key matches the selector with the highest
+    level match supported by the selector.
```
eemeli, line 156: It would be clearer to express the preferential order as "a list of string keys in preferential order", which is how the spec expects the [MatchSelectorKeys](https://github.com/unicode-org/message-format-wg/blob/main/spec/formatting.md#resolve-preferences) function to work.

exploration/number-selection.md
```diff
@@ -92,16 +95,78 @@ ICU MF1 messages using `plural` and `selectordinal` should be representable in M
...
+- Else, if the value of the key is a keyword and the value of `select` is not `exact`:
+  - If the keyword value of running plural(operand) or ordinal(operand)
+    equals the key, the key matches the selector with a level of match
+    lower than the highest level, but higher than the default.
```
eemeli, line 160: Are these `plural(operand)` and `ordinal(operand)` functions defined somewhere? And shouldn't they take the options as well? I would prefer describing their behaviour here in words instead of pseudocode.

exploration/number-selection.md
```diff
@@ -92,16 +95,78 @@ ICU MF1 messages using `plural` and `selectordinal` should be representable in M
...
+> [!NOTE]
+> The default keyword `other` can produce this level of match.
```
eemeli, line 162: I disagree. According to the description above, `other` is only matched if the operand's plural category really is `other`.
```suggestion
```

exploration/number-selection.md
```diff
@@ -92,16 +95,78 @@ ICU MF1 messages using `plural` and `selectordinal` should be representable in M
...
+- Else, if the value of the key is `*`
+  - The key matches the selector at the lowest (default) level of match.
```
eemeli, line 164: This never happens, because a `*` is never compared to the selector value. Its preferential position is automatic.
```suggestion
```

exploration/number-selection.md
```diff
@@ -92,16 +95,78 @@ ICU MF1 messages using `plural` and `selectordinal` should be representable in M
...
+- Else, return `no_match` (eliminating the variant from the selection operation)
+
+The `select` type `plural` is the default because this value is presumed to be the most common use case.
+It affords the least bad fallback when used incorrectly:
+using "plural" isntead of "exact" still selects exactly matching cases,
```
eemeli, line 169:
```suggestion
using "plural" instead of "exact" still selects exactly matching cases,
```

exploration/number-selection.md
```diff
@@ -116,20 +181,52 @@ but can cause problems in target locales that the original developer is not cons
...
+- the `select` type of this selector is always `plural`
+- the function is not valid as a formatter
+
+The function `:ordinal` operates identically to `:number`, except:
+- the `option` of `select` is not valid
+- the `select` type of this selector is always `ordinal`
+  (This function is valid as a formatter)
+
+The function `:integer` operates identically to `:number`, except:
+- the `minimumFractionDigits` and `maximumFranctionDigits` options
```
eemeli, line 197:
```suggestion
- the `minimumFractionDigits` and `maximumFractionDigits` options
```

exploration/number-selection.md
```diff
@@ -116,20 +181,52 @@ but can cause problems in target locales that the original developer is not cons
...
+  default to `0` even for floating point numeric types
+> [!NOTE]
+> The fraction digits can be overridden for the purposes of formatting
+> and selection, but the result must be that any fractional part of
+> the operand is reduced to `0`
```
eemeli, line 202: What does this mean?

exploration/number-selection.md
```diff
@@ -116,20 +181,52 @@ but can cause problems in target locales that the original developer is not cons
...
+### Determining Exact Literal Match
+A numeric literal key value exactly matches the resolved value of an operand as follows:
+
+> this is still in progress, but basically has to consider:
+
+- signs match
+- minimum fraction digits match
+- scientific notation?
+- other formatting gorp doesn't apply (e.g. currency or measure unit)
+- no grouping
```
eemeli, line 219: Given that we have `:string` separately, `:number` should so numerical comparison:
```suggestion
To determine if a numeric literal key value
exactly matches the resolved value of an operand,
the parsed numerical value represented by the key
is compared to the numerical value of the _operand_.
Selector options such as `minimumFractionDigits` have no effect on this comparison.
```

With that, the keys `1.0` and `1` would both match a number 1, with equal weight.

exploration/number-selection.md
aphillips: Actually, I think you are agreeing? What I'm trying to call out here is that `other` can be ranked here. Perhaps I should just omit the word "default"?

exploration/number-selection.md
aphillips: It means that this:

```
{|1.2345| :integer minimumFractionDigits=1}
```

Formats as:

```
1.0
```

exploration/number-selection.md
aphillips:
```
.match {$num :number}
oen  {{I ain't nothing but one misspelled}}
*    {{Should this fire instead?}}
```

Does "ignore the key in further processing" mean that the above selects `*`? Or does it trigger the fallback?

exploration/number-selection.md
aphillips: I know that you've written the algorithm to never do the comparison. I think that might be wrong. I think a selector should be exposed to all of the keys and decide how to sort `*` into the mix. Among other things, in `:number` this allows the plural rule `other` to be a true synonym to `*`

exploration/number-selection.md
eemeli: With the above message, the `*` variant is always selected.

exploration/number-selection.md
eemeli: To me, the intent of the algorithmic step above is pretty clear:
> If the keyword value of running plural(operand) or ordinal(operand) equals the key, the key matches the selector [...].

But then this note calling out `other` makes it unclear. Do you intend for it to be treated as special in any way?

exploration/number-selection.md
eemeli: This is why I called out during our last call reviewing the current selection algorithm, because if there is any desire to change it, we really should have a PR available for review ASAP. I don't think such a change would be necessary or beneficial, but I'd be open and interested to review specific changes.

My own sense is that the current algorithm is capable of supporting any actual selection behaviour, and that further changes would add unnecessary complexity.

exploration/number-selection.md
eemeli: That seems rather surprising. If we do that, it'll be the only place in the built-in functions where we're transforming the inputs. Do we really need this? Couldn't we just not support `minimumFractionDigits` on `:integer`?

exploration/number-selection.md
aphillips: Yes. I guess the question is whether `oen` should produce an error or just not match. My text says it's an error. I could imagine creating a sub-species of Selection Error for bad key values?

exploration/number-selection.md
aphillips: My instinct was to propose changes to that algorithm first, which is why this took so long to start writing. In the end I decided to work on this as an aid to proposing changes there. I see what you're saying. I'm re-writing the algorithm "your" way this morning and we'll see if that makes more sense.

exploration/number-selection.md
aphillips: We don't really transform the inputs, but I see what you mean. I'll forbid all of the fraction digits on `:integer` for now.

exploration/number-selection.md
```diff
@@ -92,16 +95,78 @@ ICU MF1 messages using `plural` and `selectordinal` should be representable in M
...
+
+> For example, in Java, any subclass of `java.lang.Number` plus the primitive
+> types (`boolean`, `byte`, `short`, `int`, `long`, `float`, `double`, etc.) 
+> might be considered as the "implementation-defined numeric types".
+> Implementations in other programming languages would define different types
+> or classes according to their local needs.
+
+Implementations SHOULD treat numeric literals that produce parse errors as
+non-numeric string literals.
+That is, numeric selection MUST emit a _Selection Error_ for these values.
```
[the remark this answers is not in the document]
aphillips: Fixed.

exploration/number-selection.md
aphillips: See new text

exploration/number-selection.md
aphillips: Done

exploration/number-selection.md
aphillips: Floating point vs. integer representation might introduce some corner cases. I'm not sure about "equal weight" either. Our specification requires a list at the moment, so we should provide a tie-break so that the order is deterministic.

We also have to be cautious. Suppose the selector is `{$num :number maximumFractionDigits=1}`, the value in the operand is `1.0001`, and the key is `1.0`. This should probably be a match, right? Even though the numerical values are not (quite) equal?

macchiati: A couple of points:
- As @aphillips notes, many languages have different behavior depending on the presence or absence of trailing zero fraction digits, so those must be maintained. That is, |1|, |1.0|, |1.00|, etc.
- The interaction between max/min digits and significant digits is surprisingly complicated. Describing the precise behavior should not be attempted in this spec.

eemeli: No, that should not match as the numerical values are not equal.

While I agree that the behaviour around non-integer values may be a little surprising, as far as I know they are only theoretical, and should not influence the overall design unduly. Or are there any real-world examples of current messages that need to work around the limitations of MF1, Fluent, or any other syntax that supports selection on exact numerical values? I have searched a bit, but have not found any.

exploration/number-selection.md
```diff
@@ -92,45 +95,241 @@ ICU MF1 messages using `plural` and `selectordinal` should be representable in M
...
+- `currencySign`
+  - `accounting`
+  - `standard` (default)
+- `notation`
+   - `standard` (default)
+   - `scientific`
+   - `engineering`
+   -  `compact`
+- `numberingSystem`
+   -  (one of the ISO15927 script codes or extensions defined by ICU and/or Intl.NumberFormat
```
macchiati, line 182: This can point to CLDR instead, which is what ICU draws from.

exploration/number-selection.md
```diff
@@ -92,45 +95,241 @@ ICU MF1 messages using `plural` and `selectordinal` should be representable in M
...
+
+- `select`
+   -  `plural` (default)
+   -  `ordinal`
+   -  `exact`
+- `compactDisplay`
+   - `short` (default)
+   - `long`
+- `currency`
+   -  (`alpha3` ISO 4217 currency code)
```
macchiati, line 167: I'd suggest https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeCurrencyIdentifier

exploration/number-selection.md
aphillips: @eemeli I didn't take you text verbatim, but I took the idea into account. Check it out.

@macchiati Thanks. The matching behavior for literal keys I don't think is affected by this, but certainly the plural/ordinal rule computation is. I think we've accounted for that. 

I agree that we don't want to get into signfiicant digits and such, although I am worried about developers not being able to figure out why setting the number of digits or modifying the rounding mode doesn't cause operand values to match certain key literals (as @eemeli suggests). That is, why doesn't `{|1.00000001| :number maximumFractionDigits=1}` match `1.0` (since it formats as "1.0" or the equivalent)?

aphillips: @macchiati I think we understand.

Part of the problem here is that your example doesn't annotate `$averageWater`. It would either print `1` as an integer or `1.0`, but not both. If `$averageWater` is annotated to produce a fraction part, plural rules would handle getting the right keyword (including for `1.0`). You wouldn't write the literal `|1|`.

Let me rewrite your example a bit:

```
.match {$averageWater :number minimumFractionDigits=1}
one    {{This never fires in English; {$averageWater} liter per person}}
*      {{This can fire in English;    {$averageWater} liters per person}}
1      {{This matches;                {$averageWater} liter per person}}
1.0    {{This also matches;           {$averageWater} liters per person}}
```

The problem with this message and maybe @eemeli's suggestion is that `1` matches because it has the value `1.0`. This makes writing non-grammatical messages perhaps slightly more likely. It's also unclear what user expectations would be for such a message.

My suggestion would be that the key has to match the significant digits in the selector function in order to match. This might involve trimming or rounding the operand. Describing this is much more complicated that just parsing the key and doing `.equals(operand)` on it. It would probably be more like "serialize the operand according to the JSON rules to the number of significant fraction digits specified in the numeric selector function; The key and operand are an exact match if the resulting string is code point equal to the key".

Is this significantly better than the parse solution? It might be if floating point or fractional values are commonly used as literal keys. My experience, though, is that most literal keys are integers in integer applications. Should we provide an affordance for these more common cases? Or is full control worth the extra cost and complexity?

exploration/number-selection.md
```diff
@@ -92,45 +95,241 @@ ICU MF1 messages using `plural` and `selectordinal` should be representable in M
...
+       of the value of the `operand`, then `key` matches the selector.
+       Add `key` to the front of the `return_value` list.
+   - Else, if the value of `key` is a keyword:
+      - Let `keyword` be a string which is the result of [rule selection](#rule-selection).
+      - If `keyword` equals `key`, then `key` matches the selector.
+        Append `key` to the end of the `return_value` list.
+   - Else, `key` is not a match.
+     Do not add `key` to `return_value`
+- Return `return_value`
+
```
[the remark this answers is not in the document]
aphillips: Will do when this moves to the spec

exploration/number-selection.md
```diff
@@ -92,45 +95,241 @@ ICU MF1 messages using `plural` and `selectordinal` should be representable in M
...
+- `maximumFractionDigits`
+  - (positive integer)
+- `minimumSignificantDigits`
+  - (positive integer)
+- `maximumSignificantDigits`
+  - (positive integer)
+
+
+> [!NOTE]
+> We have a requirement for `offset` but have not provided for it yet??
```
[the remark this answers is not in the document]
macchiati: Right, something like that would need to be added. Lacking it means that we can't emulate a major feature of MF1 behavior with MF2 messages, so that needs to be documented clearly in the spec for the tech preview, so that people don't try to do it.

exploration/number-selection.md
macchiati: I think the the most robust approach would be the following description (I'm not completely clear if this is just restating what your words were or not).

A literal matches a variable using {$variable :number ...}, iff after formatting the literal using the same settings as the $variable has, there is an exact string match. 

 It can clearly be optimized in production, but if we state it like this it establishes the principle. It also means that we can detect that 1 are equivalent: will either both match or neither match, which can be detected statically.

aphillips: Let me give it a shot at wording. I think this is the better approach.

aphillips: MF2 has _multiple selection_ and that might work better than trying to shove everything into a single selector:

```
.local $offset = {$numGuides :number offset=1}
.match {$numGuests :number} {$offset}
0   *   {{No guests were invited}}
1   *   {{Only {$guest1} was invited}}
*   one {{Only {$guest1} guest and {$offset} more guest}}
*   *   {{Only {$guest1} and {$offset} more were invited}}
```

(Note that other languages would write `one one`/`one few`/`one *`/etc. messages to fit grammatical needs. The keys `0 *` and `1 *` act as guardrails against the weird behavior of `$offset`)

macchiati: Good point; that would involve just one additional operand ('offset', or maybe 'minus'). Not terse, but works.

```
.local $offset = {$numGuests :number offset=1}
.match {$numGuests :number} {$offset}
0   *   {{No guests were invited}}
1   *   {{Only {$guest1} was invited}}
*   one {{{$guest1} and {$offset} more guest was invited}}
*   *   {{{$guest1} and {$offset} more guests were invited}}
```

exploration/number-selection.md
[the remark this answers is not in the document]
eemeli: This could be supported by an option like `icu:offset`, much like skeletons, another MF1 feature that the initial core set of built-in functions won't support.

exploration/number-selection.md
```diff
@@ -92,44 +95,274 @@ ICU MF1 messages using `plural` and `selectordinal` should be representable in M
...
+### Selection
+
+When implementing [`MatchSelectorKeys`](spec/formatting.md#resolve-preferences), 
+numeric selectors perform as described below.
+
+- Let `return_value` be a new empty list of strings.
+- Let `operand` be the resolved value of the _operand_.
+  If the `operand` is not a number type, emit a _Selection Error_
+  and return `return_value`.
+  > This might result in the fallback message (logo) being rendered.
```
eemeli, line 244: This is not the case according to our current spec where an all-`*` variant is required.
```suggestion
```

exploration/number-selection.md
```diff
@@ -92,44 +95,274 @@ ICU MF1 messages using `plural` and `selectordinal` should be representable in M
...
+   - If the value of `key` matches the production `number-literal`:
+     - If the parsed value of `key` is an [exact match](#determining-exact-literal-match)
+       of the value of the `operand`, then `key` matches the selector.
+       Add `key` to the front of the `return_value` list.
+   - Else, if the value of `key` is a keyword:
+      - Let `keyword` be a string which is the result of [rule selection](#rule-selection).
+      - If `keyword` equals `key`, then `key` matches the selector.
+        Append `key` to the end of the `return_value` list.
+   - Else, `key` is not a match.
+     Do not add `key` to `return_value`
```
eemeli, line 261: The error case is included twice here.

```suggestion
   - If the value of `key` matches the production `number-literal`:
     - If the parsed value of `key` is an [exact match](#determining-exact-literal-match)
       of the value of the `operand`, then `key` matches the selector.
       Add `key` to the front of the `return_value` list.
   - Else, if the value of `key` is a keyword:
      - Let `keyword` be a string which is the result of [rule selection](#rule-selection).
      - If `keyword` equals `key`, then `key` matches the selector.
        Append `key` to the end of the `return_value` list.
   - Else, `key` is invalid;
     emit a _Selection Error_.
     Do not add `key` to `return_value`
```

exploration/number-selection.md
```diff
@@ -92,44 +95,274 @@ ICU MF1 messages using `plural` and `selectordinal` should be representable in M
...
+- Return `return_value`
+
+### Plural/Ordinal Keywords
+The _plural/ordinal keywords_ are: `zero`, `one`, `two`, `few`, `many`, and
+`other`.
+The value `other` is both a valid output of [rule selection](#rule-selection)
+and serves a default value.
```
eemeli, line 268: Mentioning this here would be misleading; it's better to just describe the actual behaviour in Rule Selection.

```suggestion
```

exploration/number-selection.md
```diff
@@ -92,44 +95,274 @@ ICU MF1 messages using `plural` and `selectordinal` should be representable in M
...
+If the option `select` is set to `plural`, selection should be based on CLDR plural rule data
+of type `cardinal`. See [charts](https://www.unicode.org/cldr/charts/latest/supplemental/language_plural_rules.html)
+for examples.
+
+If the option `select` is set to `ordinal`, selection should be based on CLDR plural rule data
+of type `ordinal`. See [charts](https://www.unicode.org/cldr/charts/latest/supplemental/language_plural_rules.html)
+for examples.
+
+Apply the rules defined by CLDR to the resolved value of the operand and return the resulting keyword.
+If no rules match, return `other`.
```
eemeli, line 288:
```suggestion
Apply the rules defined by CLDR to the resolved value of the operand and the function options,
and return the resulting keyword.
If no rules are available for the requested locale or locales, return `other`.
```

exploration/number-selection.md
```diff
@@ -92,44 +95,274 @@ ICU MF1 messages using `plural` and `selectordinal` should be representable in M
...
-	<setOption name="select" value="plural"/>
-</alias>
-<alias name="ordinal" supports="match">
-	<setOption name="select" value="ordinal"/>
-</alias>
-```
+
+### Determining Exact Literal Match
+
+Number literals in the MessageFormat v2 syntax use the
```
eemeli, line 316:
```suggestion
Number literals in the MessageFormat 2 syntax use the 
```

exploration/number-selection.md
```diff
@@ -92,44 +95,274 @@ ICU MF1 messages using `plural` and `selectordinal` should be representable in M
...
-```
+
+### Determining Exact Literal Match
+
+Number literals in the MessageFormat v2 syntax use the 
+[format defined for a JSON number](https://www.rfc-editor.org/rfc/rfc8259#section-6).
+The resolved value of an `operand` exactly matches a numeric literal `key`
+if, when the `operand` is serialized using the format for a JSON number
+including the exact number of fraction digits specified by the selector
+function or its options, the two strings are equal.
```
eemeli, line 321: This places an unreasonable burden on all selections in order to support a theoretical corner case. Consider, for instance, that we also support the options `minimumIntegerDigits`, `minimumSignificantDigits`, and `maximumSignificantDigits`. They either need to be considered here as well, or ignored.

I'm starting to think we don't need to specify this here, and should leave it up to implementations. Comparisons against floating point values are fraught in general, and should be avoided.

```suggestion
The resolved value of an `operand` exactly matches a numeric literal `key`
if, when the `operand` is serialized using the format for a JSON number,
the two strings are equal.
All formatting options are ignored when applying this serialization.
Using keys representing fractional numbers is not recommended,
and may produce different results with different implementations.
```

If this suggestion is accepted, the examples below will also need to be updated.

eemeli (changes requested):

exploration/number-selection.md
```diff
@@ -92,44 +95,274 @@ ICU MF1 messages using `plural` and `selectordinal` should be representable in M
...
+or is some serialization of a number that the implementation recognizes.
+
+> For example, in Java, any subclass of `java.lang.Number` plus the primitive
+> types (`byte`, `short`, `int`, `long`, `float`, `double`, etc.) 
+> might be considered as the "implementation-defined numeric types".
+> Implementations in other programming languages would define different types
+> or classes according to their local needs.
+
+Implementations MUST treat literals that do not match `number-literal` as
+non-numeric string literals.
```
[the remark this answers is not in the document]
aphillips: Agreed. Perhaps:

```suggestion
Implementations MUST treat literals that do not match `number-literal` as
plain string literals.
```

exploration/number-selection.md
aphillips: Although if the operand isn't a number now, it probably won't format later. But that could result in a more usable fallback message I suppose.

exploration/number-selection.md
aphillips: I'm going to commit your change, although, in fact, "`key` is not a match" is not the same thing as the error case.

exploration/number-selection.md
aphillips: This would be incorrect. If the specific locale requested has no rules, the rules used will fall back up the locale hierarchy, ultimately to the root locale. There is always a root locale set of resources.

My text "If no rules match..." covers any errors in CLDR data.

exploration/number-selection.md
aphillips: Integer and significant digits (and, fwiw, grouping) formatting specifications don't enter into it because they don't affect plural selection the way that fraction digits do. They surely affect the appearance of the final product, but not "equality".

This means that overlong keys probably don't work:

```
.match {$num :number minimumIntegerDigits=3}
000 {{I can never be selected}}
010 {{I can't either}}
100 {{But I work}}
```

> Comparisons against floating point values are fraught in general, and should be avoided.

Yes. I'll note that people using this feature are on thin ice (this is actually the one place where `ChoiceFormat` did something useful in the modern era). Nearly all exact number selections that I'm familiar with use integers.

exploration/number-selection.md
```diff
@@ -92,44 +95,274 @@ ICU MF1 messages using `plural` and `selectordinal` should be representable in M
...
+
+The function `:number` is the default selector for numeric values.
+
+The function `:plural` operates identically to `:number`, except:
+- the `select` type of this selector is always `plural`
+- the function is **not** valid as a formatter
+
+The function `:ordinal` operates identically to `:number`, except:
+- the `select` type of this selector is always `ordinal`
+  (This function is valid as a formatter)
```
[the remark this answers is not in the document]
aphillips: I could agree to leave `:ordinal` out of formatting. ICU provides a formatter in MF1, but it is not a Good Thing as discussed previously. Allowing `:icu:ordinal` as a formatter seems a better compatibility thing so that we don't spread bad ordinal formatters in other implementations.

I think we have consensus to have `:plural` as a selector-only function.

exploration/number-selection.md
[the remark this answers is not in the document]
eemeli: Your fix doesn't seem to address my issue? The questionable word in this phrase is "literal", because in our common usage it's referring to a value that's defined directly in the syntax source. My concern is that the spec should allow for an implementation to consider the values of e.g. a literal operand `42` and a variable operand with the resolved value of the _string_ `42` to be indistinguishable from each other.

exploration/number-selection.md
eemeli: Ah. I had presumed we could assume the CLDR data to be infallible, but that is indeed a risky supposition.

exploration/number-selection.md
eemeli: Ok, I'll grant you that `minimumIntegerDigits` can be discounted, but:

```js
new Intl.NumberFormat('en', { maximumSignificantDigits: 2 }).format(1.03) // "1"
new Intl.NumberFormat('en', { maximumSignificantDigits: 3 }).format(1.03) // "1.03"

new Intl.NumberFormat('en', { minimumSignificantDigits: 2 }).format(1.3) // "1.3" 
new Intl.NumberFormat('en', { minimumSignificantDigits: 3 }).format(1.3) // "1.30"
```

exploration/number-selection.md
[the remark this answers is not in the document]
macchiati: Note: I think we would need to add icu:offset in the tech preview in ICU, because otherwise we wouldn't be able to emulate MF1.0 with MF2.0, which is an important test.

exploration/number-selection.md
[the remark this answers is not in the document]
macchiati: 👍

exploration/number-selection.md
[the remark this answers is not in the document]
macchiati: We shouldn't leave ordinals out from first release; they are distinct from plurals. However, we could delay support until after tech preview.

Ordinals are **formatted** as integers. 
Their **selection** is different than plurals. The purpose is to be able to support numeric ordinals where the pre/suffixes are in the literal text, such as:

one: The {$rank}st prize.
two: The {$rank}nd prize.
few: The {$rank}rd prize.
other: The {$rank}th prize.

(English is similar to Slavics in how it handles ordinals, interestingly.)

The reason for not formatting $rank alone as (say) "21st" is that for many languages, the pre/suffixes are inflected.

exploration/number-selection.md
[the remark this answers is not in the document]
macchiati: That gets into what the "resolved" value is. The *formatted" value of a variable might be `४२` (numerically equal to 42, but not equal in string value).

exploration/number-selection.md
macchiati: Well, nothing is infallible.

A more accurate picture is that CLDR does not have data for all 7,000ish languages, so some languages will fall back to root.
And of course, we add locales over time, so a language that falls back to root now, may change in the future; and every now and then we find that we need to tweak plural rules and/or number formats in existing locale data because we didn't fully capture usage (or even, sometimes, that usage changes over time).

exploration/number-selection.md
macchiati: > This places an unreasonable burden on all selections in order to support a theoretical corner case

I disagree. The suggested language for this will break for some cases, and not just theoretical ones.

The language "Apply the rules defined by CLDR to the resolved value of the operand and the function options,
and return the resulting keyword." works, and is not an excessive burden. Worst case is that as well as formatting the variable, you also format a few literals. However, it is very easy to optimize the implementations to avoid the extra formatting.

exploration/number-selection.md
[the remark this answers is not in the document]
eemeli: Support for selection on ordinals is definitely in, with `:number select=ordinal`. The possibly open question is whether to leave its `:ordinal` shorthand out, and how to define its behaviour when formatting.

As we don't have consensus for something like MF1's ordinal formatter, I don't think we should define formatting for `:ordinal`. In English, I would expect `{3 :ordinal}` to format as "3rd", and in Finnish, I'd expect "3.". Getting "3" would be surprising in nearly all cases.

exploration/number-selection.md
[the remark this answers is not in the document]
aphillips: @macchiati Note that we won't be able to emulate the formats that we agreed to exclude: `spellout`, `duration`, and `choice` unless ICU chooses to implement them using namespacing (`:icu:spellout`, etc.)

exploration/number-selection.md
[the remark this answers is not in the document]
aphillips: TL;DR: I support `:ordinal` as a selector but not as a formatter for the tech preview. @macchiati already calls out awareness of one problem with the formatter (inflection), although the fallback problem is equally noisome.

@eemeli At least in ICU it works they way you expect--in the locales you chose as examples. However, 
this is bothersome because it requires users (probably translators) in some languages to make changes (probably excruciatingly clever ones) to their messages to avoid the "well, it works in English" ordinal formatter with the added negative of "but it's in the I18N API". Also, JS and ICU4X don't currently have the formatter.

For those playing at home, some examples of the inflection problem:
```
es: 1.º 2.º 3.º 4.º 5.º 6.º   <- vs. e.g. 1ª
fr: 1er   <- vs. e.g. 1ère
```

And the fallback problem: locale data for ordinals is relatively sparse, so a lot of locales produce obviously goofy output stemming from the root locale's `other` resource and English's "Slavic" ordinals (I would suggest that maybe the root locale should produce a better degraded form or that locales that use `other` be provided with an insulating resource):

```
uz-UZ: 1th 2th 3th 4th 5th 6th   // there are lots more like this
xog-UG: 1th 2th 3th 4th 5th 6th  // these are near the bottom of the list
zu-ZA: 1th 2th 3th 4th 5th 6th
```

---

Try it yourself (requires ICU4J):

```java
    public static void ordinalsGalore() {
        com.ibm.icu.text.MessageFormat mf = new com.ibm.icu.text.MessageFormat(
                "{6}: {0,ordinal} {1,ordinal} {2,ordinal} {3,ordinal} {4,ordinal} {5,ordinal}");
        final Object[] args = {1, 2, 3, 4, 5, 6, ""};
        for (Locale l : sortAllLocales()) {
            mf.setLocale(l);
            args[6] = l.toLanguageTag();
            System.out.println(mf.format(args));
        }
    }

    public static List<Locale> sortAllLocales() {
        final List<Locale> list = Arrays.asList(Locale.getAvailableLocales());
        list.sort((l1, l2) -> (int) l1.toLanguageTag().compareTo(l2.toLanguageTag()));
        return list;
    }
```

exploration/number-selection.md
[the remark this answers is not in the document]
aphillips: I think my edits this morning (1c0bf19) address this. I removed the MUSTard in favor of text very similar (intentionally similar) to that in date/time. 

@macchiati "resolved value" means mostly that we unwind the chain of assignments to have the actual value in hand. That is:

```
Input: `{ 'a' : 'wildebeest'}`
.local $b = {$a}
.local $c = {$b}
.local $d = {$c}
{{The resolved value of {$d} is the literal 'wildebeest'}}
```

This is not the same thing as the _formatted_ value, although any annotations pick up along the way are available.

FWIW, ` |४२|` is not a valid numeric literal because it does not match the `numeric-literal` production. It's just some string.

exploration/number-selection.md
```diff
@@ -92,44 +95,274 @@ ICU MF1 messages using `plural` and `selectordinal` should be representable in M
...
+> might be considered as the "implementation-defined numeric types".
+> Implementations in other programming languages would define different types
+> or classes according to their local needs.
+
+Implementations MUST treat literals that do not match `number-literal` as
+non-numeric string literals.
+Non-numeric values, including non-numeric string literals, result in
+a _Selection Error_.
+
+> _Ed note:_ Alternatively we can require that non-numeric values result in `*`
```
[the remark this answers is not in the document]
aphillips: I'll buy that. 

I don't think this can be a _Resolution Error_, since the value resolves correctly. It is the number function that is not accepting the input. In the case of a selector, this means that it can produce a _Selection Error_ (to indicate the type mismatch). In the case of a formatter, this means that it can produce a _Formatting Error_. If you read the list of _Resolution Errors_, you'll see that none of them fit this case. And I do think that it is _internal_ to `:number` (and friends) whether to accept a given value.

The behavior of the selector is as you suggest, with the `*` case being the pattern selected. If the value is used in the pattern, that value will use its fallback resolution. Either way there will be an error available to be emitted (in an implementation-defined way).

exploration/number-selection.md
aphillips: As noted in a previous comment, many locales produce `other` and the `other` resource in `root` is not very helpful.

I also included the last statement to guard against implementations returning nothing.

The other fix seems valid.

```suggestion
Apply the rules defined by CLDR to the resolved value of the operand and the function options,
and return the resulting keyword.
If no rules match, return `other`.
```

exploration/number-selection.md
aphillips: @eemeli Fair enough. I added significant digits to the match, including examples.

I do notice that both FF and Chrome's implementation of `Intl.NumberFormat` prioritizes significant digits ahead of fraction ones:

```javascript
const number = 1.23456;

console.log(
  new Intl.NumberFormat('de-DE', { maximumSignificantDigits:3, maximumFractionDigits:0 }).format(
    number,
  ),
);
```

Produces: `1,23` 

(ICU4J `NumberFormatter` does not replicate this, so this might be a bug)

```java
        NumberFormatter.withLocale(Locale.US)
           .precision(Precision.minSignificantDigits(3))
           .precision(Precision.maxFraction(0))
           .format(1.2345)
           .toString()  // produces `1`
       );
```

In any case, I didn't deal with the interplay between these two in the spec and \* _sigh_ \* I probably need to.

exploration/number-selection.md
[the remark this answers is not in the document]
eemeli: While the operand value does resolve correctly in this case, it's the resolution of the expression that fails. Consider this nonsensical but formally valid message, for instance:

```
.local $x = {horse :number}
.match {$x}
1 {{I'm a cow}}
* {{Herding {$x} along}}
```

Here, we need to emit an error when resolving the `{horse :number}` expression, and then when using its value for selection. That same order of errors should be retained if the selector expression itself fails to resolve.

> If you read the list of _Resolution Errors_, you'll see that none of them fit this case.

Good point. The specific kind of resolution error isn't currently defined: https://github.com/unicode-org/message-format-wg/blob/e588afba76b0bc31251a96c2760c4905618ed733/spec/formatting.md?plain=1#L240-L241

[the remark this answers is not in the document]
aphillips: @eemeli Thanks for this.

I think your example is a good one, since we want the expression to fail in the `.local` rather than waiting for the `.match` or the formatting in the pattern.

Regarding the quoted lines, I'm not sure, having thought about it, what an "invalid value" is (we don't define a valid value). `null` might be an acceptable value for some kinds of calls? Ultimately, though I think we need a specific error type (so we can test for it, if nothing else). I think something like `Invalid Expression`, one subtype of which would be `Type Mismatch`.

I'll create an issue to track and a PR (if I have time) to propose it.

exploration/number-selection.md
macchiati: @sffc is the expert on this. The best interplay between them where there are seeming conflicts is tricky to get right, and as I recall, _not_ just a matter of having one have precedence over the other.

sffc: My previous research showed that min/max significant digits are much less popular than min/max fraction digits outside of compact notation. If you don't need compact notation, I would suggest considering dropping significant digits.

There is a way to map between ECMA-402 digit options and NumberFormatter digit options. I maintain that the NumberFormatter options are simpler and less error-prone, but the ECMA-402 options were set long ago so we did the best we could within that space.

aphillips: Thanks @sffc. Significant digits are definitely going to be rarer than fraction digits--I can't recall using them in any kind of "normal" application, aside from compact (and then only sporadically). 

I have no problem with ECMA-402's options. I just note that it produced some surprise for me to get a fractional part after I explicitly asked not to (e.g. significant > fraction). I expected that significant digits would step on the *value* of the formattable number, while fraction digits stepped on the number of decimal places shown.

We *only* have those options that overlap between 402 and ICU4J (so, for example, no rounding control) and since significant digits is in both, that's why it's there. I would shed zero tears to drop it in preview, although it's just going to come back again later (or in the form of `:icu:maximumSignificantDigits`/`:js:maximumSignficantDigits` 😉)

In any case, we have differences in implementations, so I should tread carefully in MFv2 and make the interplay implementation defined (assuming you do not agree that ECMA-402's behavior is a bug).

sffc: > We only have those options that overlap between 402 and ICU4J (so, for example, no rounding control)

ICU4J is a superset of ECMA-402. Note that ECMA-402 contains rounding increment and rounding mode as of NFv3 which was stabilized about a year ago.

> In any case, we have differences in implementations, so I should tread carefully in MFv2 and make the interplay implementation defined

I don't agree with the characterization that there is a "difference in implementations". Rather, there is a difference in APIs. It should not be assumed that just because there is an option with the same / similar name in ECMA-402 and ICU4J that the two options have the same / similar behavior. When the APIs were designed (I was involved with both), the respective committees preferred to learn from the mistakes of the previous generation than try to match it bug-for-bug.

exploration/number-selection.md
```diff
@@ -117,7 +117,16 @@ The function `:plural` operates identically to `:number`, except:
 
 The function `:ordinal` operates identically to `:number`, except:
 - the `select` type of this selector is always `ordinal`
-  (This function is valid as a formatter)
+- formatting behavior of this function is implementation defined
+
+> [!NOTE]
+> CLDR data has known shortcomings related to automatic formatting
```
macchiati, line 123:
```suggestion
> CLDR data was designed to only format numeric ordinals, using a similar mechanism as plural selection, because the suffixes tend to need inflection.
```

macchiati (changes requested):

exploration/number-selection.md
```diff
@@ -92,44 +95,303 @@ ICU MF1 messages using `plural` and `selectordinal` should be representable in M
...
+>
+> For example, if the value of the variable `num` were the string
+> `-1234.567`, it would behave identically to the local
+> variable in this example:
+> ```
+> .local $example = {|-1234.567| :number}
+> {{{$num :number} == {$example}}}
+> ```
+
+### Options
```
sffc, line 163: Issue: I really don't think this matrix of options belongs in the MF spec.

Number formatting options are very complex. I've spent a lot of time over the last several years trying to reduce bugs and align more with client needs.

There are two parts of the model for number formatting that I think should land in MF2:

1. Specific functions aligned to specific use cases. Rather than an all-in-one `:number` function, instead tailor them to specific use cases. Engines can implement the functions to serve their use case, not necessarily being bound to a specific options matrix.
    - If we _must_ include number formatting options in MF2, we need to open a more deep discussion about which options to include and how they should behave.
2. The numeric type should be a Decimal. I think this is consistent with the `number-literal` grammar except that the extracted number should be required to be taken at face value, including trailing zeros. In other words, the "implementation-defined numeric type" must have capabilities similar to `java.math.BigDecimal`.
    - This eliminates the need for the `number` function to include options such as `minimumFractionDigits`. It also fixes one long-standing bug in MF1 which is that the `plural` function _needs to know_ the final rounded number and all of its digits in order to return the correct plural form. By pushing this out of MessageFormat, MF2 can be simpler and more accurate.

sffc (changes requested):

exploration/number-selection.md
aphillips: I think what you're suggesting is a different approach than what MFWG has been working on. The syntax of a message necessarily includes the options needed for users (developers, translators) to control which patterns are selected for formatting and, relatedly, how values are inserted or formatted into those patterns.

If we didn't provide the options, each implementation on each platform would be different. Users would have to relearn the details of number and date formatting for each implementation. Tools (i.e. CAT tools) would have to know about each variation and which variation to apply to each message. This is antithetical to our goals.

> The numeric type should be a Decimal

MFv2 does not have types. We require certain formats of string literal to be accepted as inputs for some functions (number and datetime are examples) and we allow implementation-specific types to be formatted. But MFv2 does not itself have types. 

The `number-literal` in our syntax is misleading. It exists **_solely_** to allow string literals that represent numbers to be unquoted in the syntax. Otherwise numbers like `42` and `-57.3` would have to be quoted in expressions and selector keys using our "beloved" quoting syntax (note that quoting **_is_** _permitted_). See examples below. Note that number literals and "word token" literals are the most common literals in current messages.

```
.local $unquoted = {42 :number}
.local $quoted   = {|42| :number}

.match {$num}
|42|  {{Quoted}}
42    {{Unquoted}}
|-42| {{Quoted}}
-42   {{Unquoted}}
```

> This eliminates the need for the number function to include options such as minimumFractionDigits

The options in the spec are for *formatting*, not for the number itself. They provide control over the kinds of display features that UX designers want developers to implement. How else do users get a number to show a specific number of fraction digits? Or zero fill a number?

> It also fixes one long-standing bug in MF1 which is that the plural function needs to know the final rounded number and all of its digits in order to return the correct plural form. By pushing this out of MessageFormat, MF2 can be simpler and more accurate.

There are two ways to look at this.

On the one hand, it is not a "bug" to know that the plural selector for `2` should potentially select a different category from the plural selector for `2.0`.

On the other hand, read this design doc carefully: **MessageFormat** does not know anything about how plural selection work, how numbers are formatted, what the correct seperators are, usw. 

What MF2 requires is that implementers provide the minimum set of functions we specify and that those functions accept certain options. We value message portability, not because we think messages will be formatted across many implementations but because portability guarantees that what you learn about how to format a message is transferable (and note that "what you learn" means "what your tools can assume" too).

> If we must include number formatting options in MF2, we need to open a more deep discussion about which options to include and how they should behave.

I think we absolutely must format numbers and date/time values. No one needs MF2 if all we do is insert strings into strings. This design doc *is* that discussion. This is on the [F2F agenda](https://docs.google.com/document/d/1NSeimrSFzuKxne7BHmuo4Nseusa1Gfoif0fBdbF2obU/edit) for tomorrow morning, albeit not the first item in the agenda.

If you want to make an alternate proposal, we *require* a design doc--I would absolutely welcome a constructive proposal. You can make a PR against this design doc, perhaps adding an alternative considered or you can make a new one. The template is under "exploration".

sffc: > I think what you're suggesting is a different approach than what MFWG has been working on. The syntax of a message necessarily includes the options needed for users (developers, translators) to control which patterns are selected for formatting and, relatedly, how values are inserted or formatted into those patterns. ... If we didn't provide the options, each implementation on each platform would be different. Users would have to relearn the details of number and date formatting for each implementation. Tools (i.e. CAT tools) would have to know about each variation and which variation to apply to each message. This is antithetical to our goals.

I think I'm aligned on MF2 taking the role of a common set of operations to align implementations across the ecosystem. I'm just not aligned on a giant options bag as necessarily being the right mechanism to do that. I was suggesting that we take a more use-case-driven approach.

> MFv2 does not have types. We require certain formats of string literal to be accepted as inputs for some functions (number and datetime are examples) and we allow implementation-specific types to be formatted. But MFv2 does not itself have types. ... The options in the spec are for formatting, not for the number itself. They provide control over the kinds of display features that UX designers want developers to implement. How else do users get a number to show a specific number of fraction digits? Or zero fill a number?

I'm trying to make the point that specifying a number of fraction digits _should not_ be a formatting option. There is a difference between "1" and "1.0" (the latter has greater precision); the trailing zeros should be part of the data model. MF2 should request that implementations prefer numeric types that represent a formattable decimal number.

> The number-literal in our syntax is misleading. It exists solely to allow string literals that represent numbers to be unquoted in the syntax.

I was referring to this paragraph in the PR: "The _operand_ of a number function is either an implementation-defined numeric type or a literal that matches the `number-literal` production in the [ABNF](/main/spec/message.abnf)."

I think "implementation-defined numeric type" should be a type that is capable of representing trailing zeros. Failing that, a string matching the `number-literal` production is acceptable as input to the `:number` function.

> I think we absolutely must format numbers and date/time values.

I have similarly well formed opinions on date/time formatting options and would like to weigh in on that discussion as well.

aphillips: > I think "implementation-defined numeric type" should be a type that is capable of representing trailing zeros. Failing that, a string matching the number-literal production is acceptable as input to the :number function.

Note that "operand" in our spec means "input". Consider this pseudo-code fragment:

```java
MessageFormat2 mf2 = MessageFormat2.withPattern("The number is {$num}");
ImmutableMap args = ImmutableMap.of("num", aNumber);
System.out.println(mf2.format(args).toString());
```

In this example, `$num` is the _operand_. And its value is whatever `aNumber` is. The text you quote says the value `aNumber` can be whatever `MessageFormat2` accepts. This could be an `int` or a `double`. It could be a `BigDecimal`. *Any* numeric type might reasonably be input. As a user, my expectation is that I don't have to do anything to my numbers in order to format them.

I think what you're talking about is the *output*. In ICU, the result of the number formatter is likely to be a `FormattedNumber` whose internal representation is probably a `BigDecimal`. 

MF2 doesn't care.

exploration/number-selection.md
macchiati: See my comment on a different thread; I've been thinking all the time about the selectOrdinal, which **has** sufficient coverage to support inclusion. (I'm not as sure about the *formatting* ordinal; I think it is supported in about half the coverage=modern locales.)

I'm ok with keeping ordinal out of the tech review, and I don't even think we need to cite a reason for that; we could just say that it is less used.

aphillips: This PR is about number _selection_. I don't see a reason to remove `:ordinal` from that. Formatting is something else.

exploration/number-selection.md
sffc: If the input is an int or a float, that's fine. You just display the numeric value. And if you want trailing zeros, you should use BigDecimal as your input. We're just moving the concept of trailing zeros from a formatting concern to a data concern, which is where I believe it belongs.

exploration/number-selection.md
```diff
@@ -92,44 +95,303 @@ ICU MF1 messages using `plural` and `selectordinal` should be representable in M
...
 
-The default `plural` value is presumed to be the most common use case,
-and it affords the least bad fallback when used incorrectly:
-Using "plural" for "exact" still selects exactly matching cases,
-whereas using "exact" for "plural" will not select LDML category matches.
-This might not be noticeable in the source language,
+### Operands
+
+The _operand_ of a number function is either an implementation-defined numeric type or
+a literal that matches the `number-literal` production in the [ABNF](/main/spec/message.abnf).
```
eemeli, line 140:
```suggestion
The _operand_ of a number function is either an implementation-defined numeric type or
a literal that matches the `number-literal` production in the [ABNF](/main/spec/message.abnf).
Implementation-defined numeric types MAY be compound values that include
a currency code, a unit, or other options that are taken into account by the implementation.
```

exploration/number-selection.md
```diff
@@ -92,44 +95,303 @@ ICU MF1 messages using `plural` and `selectordinal` should be representable in M
...
+> .local $example = {|-1234.567| :number}
+> {{{$num :number} == {$example}}}
+> ```
+
+### Options
+
+All number functions have the following options internally.
+Unless otherwise specified in [functions](#functions) all of these options
+can be used in _annotations_.
+
```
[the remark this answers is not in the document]
aphillips: I would say SHOULD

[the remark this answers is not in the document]
aphillips: Obsoleted by moving these styles to post-45

exploration/number-selection.md
```diff
@@ -92,44 +95,347 @@ ICU MF1 messages using `plural` and `selectordinal` should be representable in M
...
+   -  `never`
+- `style`
+  - `decimal` (default)
+  - `percent` (see [Percent Style](#percent-style) below)
+- `useGrouping`
+  - `true` (default)
+  - `false`
+- `minimumIntegerDigits`
+  - (non-negative integer, default: `1`)
+- `minimumFractionDigits`
```
macchiati, line 184: minimum default is zero?

exploration/number-selection.md
```diff
@@ -92,44 +95,347 @@ ICU MF1 messages using `plural` and `selectordinal` should be representable in M
...
+  - (non-negative integer)
+- `maximumFractionDigits`
```
macchiati, line 186: is there a maximum? Or if unstated would I see 0.3333333333333333333333333333333333 for the value of 1/3?

macchiati: etc

exploration/number-selection.md
```diff
@@ -92,44 +95,347 @@ ICU MF1 messages using `plural` and `selectordinal` should be representable in M
...
+   -  `never`
+- `style`
+  - `decimal` (default)
+  - `percent` (see [Percent Style](#percent-style) below)
+- `useGrouping`
+  - `true` (default)
+  - `false`
+- `minimumIntegerDigits`
+  - (non-negative integer, default: `1`)
+- `minimumFractionDigits`
```
macchiati, line 218: It seems like the min/maxFractions default to 0 for integer, those being the only difference from :number

exploration/number-selection.md
```diff
@@ -92,44 +95,347 @@ ICU MF1 messages using `plural` and `selectordinal` should be representable in M
...
+### Options
+
+The following options and their values are required in the default registry to be available on the 
+function `:number`:
+- `select`
+   -  `plural` (default)
+   -  `ordinal`
+   -  `exact`
+- `compactDisplay`
+   - `short` (default)
```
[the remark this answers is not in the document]
aphillips: MDN says:

> Whether to use short or long form when using compact notation. This is the value provided in the [options.compactDisplay](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#compactdisplay) argument of the constructor, or the default value: "short". The value is only present if notation is set to "compact", and otherwise is undefined.

So probably this option should be moved under `notation`

exploration/number-selection.md
aphillips: These items (min*/max*) don't have defaults because they are overrides. The default is implementation defined.

exploration/number-selection.md
aphillips: see above

exploration/number-selection.md
aphillips: They default to zero and cannot be set for `:integer`

exploration/number-selection.md
```diff
@@ -92,44 +95,343 @@ ICU MF1 messages using `plural` and `selectordinal` should be representable in M
...
+- `style`
+  - `decimal` (default)
+  - `percent` (see [Percent Style](#percent-style) below)
+- `useGrouping`
+  - `true` (default)
+  - `false`
+- `minimumIntegerDigits`
+  - (non-negative integer, default: `1`)
+- `minimumSignificantDigits`
+  - (non-negative integer)
```
sffc, line 219: Please remove `minimumSignificantDigits` from `:integer`; it serves no purpose other than to create confusion when adding trailing zeros which it seems we want to avoid in `:integer` mode.

exploration/number-selection.md
```diff
@@ -92,44 +95,343 @@ ICU MF1 messages using `plural` and `selectordinal` should be representable in M
...
+   -  `auto` (default)
+   -  `always`
+   -  `exceptZero`
+   -  `never`
+- `style`
+  - `decimal` (default)
+  - `percent` (see [Percent Style](#percent-style) below)
+- `useGrouping`
+  - `true` (default)
+  - `false`
```
sffc, line 181: `useGrouping` should not be a boolean. It is an enumeration in ICU and in ECMA-402.

exploration/number-selection.md
```diff
@@ -92,44 +95,343 @@ ICU MF1 messages using `plural` and `selectordinal` should be representable in M
...
+   - `engineering`
+   - `compact`
+- `numberingSystem`
+   - valid [Unicode Number System Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeNumberSystemIdentifier)
+     (default is locale-specific)
+- `signDisplay`
+   -  `auto` (default)
+   -  `always`
+   -  `exceptZero`
+   -  `never`
```
sffc, line 175: There are 5 signDisplay values as of a few years ago. The fifth is `negative`

exploration/number-selection.md
```diff
@@ -92,44 +95,343 @@ ICU MF1 messages using `plural` and `selectordinal` should be representable in M
...
+
+
+Number literals in the MessageFormat 2 syntax use the 
+[format defined for a JSON number](https://www.rfc-editor.org/rfc/rfc8259#section-6).
+The resolved value of an `operand` exactly matches a numeric literal `key`
+if, when the `operand` is serialized using the format for a JSON number
+the two strings are equal.
+Implementations SHOULD take into account including the exact number of fraction digits 
+and the exact number of significant digits specified by the selector
+function or its options.
```
sffc, line 372: I'm not really comfortable with this type of hand-wavy language. We have established that there are multiple ways to process min/max fraction/significant digits, which differ between ICU and ECMA-402. Switching implementations should not cause my rule selection to change. I suggest that we use ECMA-402 rules for merging fraction/significant options, which are defined here:

https://tc39.es/ecma402/#sec-setnfdigitoptions

Alternatively, if we don't like that algorithm, remove `significantDigits` from the tech preview and this problem largely goes away.

Alternatively, why do we need exact literal match? This seems like ingredients for failure especially with floating point. It would be simpler and less error-prone if we either did not have this or if we supported only integer matching.

exploration/number-selection.md
aphillips: Thanks! Fixed.

exploration/number-selection.md
aphillips: Thanks. Fixed.

exploration/number-selection.md
aphillips: Significant digits alone doesn't resolve the problem.

We need to define exact literal match in order for cases like `.match {$num} 0 {{You have no zebras}}` to work, where `0` is the key. To your point, that is integer matching and I included a note about it:

> Important
>
> The exact behavior of exact literal match is only defined for non-zero-filled integer values. Annotations that use fraction digits or significant digits might work in specific implementation-defined ways. Users should avoid depending on these types of keys in message selection.

What the `SHOULD` text you quote says, basically, is that implementations ought to provide more match behavior for the (exceedingly rare, I would think) cases in which a fractional value is needed for a match. I could relax this to `MAY` with no loss of functionality.

exploration/number-selection.md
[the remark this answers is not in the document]
aphillips: Done. See above.

exploration/number-selection.md
sffc: I would be happy if you made such non-integer-match behavior optional and removed the examples that imply it is supposed to work a certain way.

exploration/number-selection.md
sffc: You went too far; I am actually okay with `maximumSignificantDigits` on `:integer`, just not `minimum*` because the only purpose that serves is to create trailing zeros which are not needed in `:integer` mode.

exploration/number-selection.md
```diff
@@ -92,44 +95,351 @@ ICU MF1 messages using `plural` and `selectordinal` should be representable in M
...
+   -  `never`
+- `style`
+  - `decimal` (default)
+  - `percent` (see [Percent Style](#percent-style) below)
+- `useGrouping`
+  - `auto` (default)
+  - `true`
+  - `false`
+  - `min2`
+  - `always`
```
sffc, line 185: The true/false values in ECMA are legacy. Without inventing anything too new, I would suggest

```suggestion
- `useGrouping`
  - `auto` (default)
  - `always`
  - `never`
  - `min2`
```

exploration/number-selection.md
```diff
@@ -92,44 +95,351 @@ ICU MF1 messages using `plural` and `selectordinal` should be representable in M
...
+When implementing `style=percent`, the numeric value of the operand
+MUST be divided by 100 for the purposes of formatting.
+
+### Selection
+
+When implementing [`MatchSelectorKeys`](spec/formatting.md#resolve-preferences), 
+numeric selectors perform as described below.
+
+- Let `return_value` be a new empty list of strings.
+- Let `operand` be the resolved value of the _operand_.
```
sffc, line 294:
```suggestion
- Let `operand` be the decimal value of _operand_ after formatting options from `:number` have been applied conforming to [the `sampleValue` grammar of UTS 35](https://unicode.org/reports/tr35/tr35-numbers.html#Plural_rules_syntax).
```

The intermediate formatted value may have notation information associated with it, such as "1.0c3" for "1.0K" or potentially "2.3e4" for "2.3*10^4". We should be clear that this is the amount of information we want from the intermediate value, and it also allows us to select on compact values, which may be important in French, for example.

exploration/number-selection.md
aphillips: Fixed.

exploration/number-selection.md
aphillips: The problem here is message portability. Probably we should remove the SHOULD and just tell users that only integer exact match works in Tech Preview "but please write to us if you find an application for fraction digits"

I'll remove the examples once we've decided on what we're doing here. They're useful illustrations for the discussion.

aphillips: (update) I implemented my comment in a8746ed

exploration/number-selection.md
aphillips: "resolved value" has special meaning in our specification (even though it is not highlighted here). It's important to say that `operand` is the "resolved value". We can add requirements *after* that, such as what you suggest.

Note that `operand` is untyped by us. It is either an object or it is a literal.

We say what to do with the literal (which ones are supported and how to parse them)

We mustn't define what the object is. That's up to the implementation. The implementer is responsible for resolving the decimal value, if that's what they need.

sffc: Hmm, are you referencing this language?

>  The form of the resolved value is implementation defined and the
>  value might not be evaluated or formatted yet.
>  However, it needs to be "formattable", i.e. it contains everything required
>  by the eventual formatting.

So the "resolved value" may or may not be the intermediate value?

exploration/number-selection.md
```diff
@@ -45,6 +47,7 @@ but ordinal rules use `one` (_1st_, _21st_, etc.), `few` (_2nd_, _22nd_, etc.),
 Additionally,
 MF1 provides `ChoiceFormat` selection based on a complex rule set
 (and which allows determining if a number falls into a specific range).
+This capability is deprecated.
```
eemeli, line 50: As I understand it, this is already effectively deprecated (but technically supported) in MF1, so being clearer that we're not doing it at all might be better. Though it's clearly still somewhat present in our design, as e.g. the MatchSelectKeys function interface could not be simplified because someone might want to implement something like ChoiceFormat in MF2.

```suggestion
This capability is not supported by the default functions of MF2.
```

exploration/number-selection.md
```diff
@@ -92,44 +95,360 @@ ICU MF1 messages using `plural` and `selectordinal` should be representable in M
...
+
+### Options
+
+The following options and their values are required in the default registry to be available on the 
+function `:number`:
+- `select`
+   -  `plural` (default)
+   -  `ordinal`
+   -  `exact`
+- `compactDisplay` // this option only had meaning with the option `notation=compact`
```
eemeli, line 160:
```suggestion
- `compactDisplay` // this option only has meaning when combined with the option `notation=compact`
```

exploration/number-selection.md
```diff
@@ -92,44 +95,360 @@ ICU MF1 messages using `plural` and `selectordinal` should be representable in M
...
+> 1    {{Does not match}}
+> 1.3  {{Does not match}}
+> 1.30 {{Matches}}
+>
+> .local $num = {1234}
+> .match {$num :number maximumSignificantDigits=2}
+> 1200   {{Matches}}
+> 1234   {{Does not match}}
+> 1200.0 {{Does not match}}
+> ```
```
eemeli, line 450: Including these examples seems distracting, as they're showing messages that we're saying _should not_ be used, and they're indicating that specific variants would match.

```suggestion
```

eemeli (approved):
Overall fine, some suggested improvements inline.

exploration/number-selection.md
aphillips: I agree but want to keep them for this morning's discussion **_only_**

PR #623, opened 2024-02-06
Describe date/time formatting functions and options

aphillips:
Based on the 2024-02-05 teleconference discussion, describe the date/time formatting functions in a manner similar to the number selection design document.

Open issues:

- I reference BCP175 and the IANA time zone database. Should we have additional zone IDs supported, such as the BCP47 compatible ones from ULIs?
- We need to choose the supported serialization(s) for date literals. RFC3339, XMLSchema, and JSON all provide some things here. SEDATE's extension to support Temporal should be added when its ready.

### DIFF
--- a/exploration/default-registry-and-mf1-compatibility.md
+++ b/exploration/default-registry-and-mf1-compatibility.md
@@ -96,13 +96,27 @@ Dates and times have the following functions:
 
 #### Operands
 
-> [!IMPORTANT]
-> The list of serializations needs more work.
+The operand of a date/time function is either 
+an implementation-defined date/time type (passed in as an argument)
+or a literal that can be parsed to an implementation-defined date/time type.
 
-The operand of a date/time function is either a literal that any of the serializations 
-defined by RFC3339 or SEDATE and which is parsed by the 
-implementation into an implementation defined date/time type; 
-or any implementation-defined date/time types.
+Date/time literals are required to be an XMLSchema 
+[dateTime](https://www.w3.org/TR/xmlschema11-2/#dateTime),
+[time](https://www.w3.org/TR/xmlschema11-2/#time),
+or [date](https://www.w3.org/TR/xmlschema11-2/#date).
+The `timezoneOffset` of each of these formats is optional. 
+When the offset is not present, conversion to incremental time types is required
+to treat the offset identically to UTC.
+Conversion of such values to floating time types
+(such as Java's `java.time.LocalDateTime`) should omit time zone.
+For more information, see [Working with Timezones](https://w3c.github.io/timezone).
+
+> [!NOTE]
+> True time zone support in serializations is expected to coincide with the adoption
+> of Temporal in JavaScript.
+> The form of these serializations is known and is a de facto standard.
+> Support for these extensions is expected to be required in the post-tech preview.
+> See: https://datatracker.ietf.org/doc/draft-ietf-sedate-datetime-extended/
 
 #### Options
 
--- a/exploration/default-registry-and-mf1-compatibility.md
+++ b/exploration/default-registry-and-mf1-compatibility.md
@@ -135,7 +135,7 @@ The following options provide high-level control over date/time formats:
   - valid [Unicode Calendar Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeCalendarIdentifier)
 - `numberingSystem` (default is locale-specific)
    - valid [Unicode Number System Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeNumberSystemIdentifier)
-- `timezone` (default is system default time zone or UTC)
+- `timeZone` (default is system default time zone or UTC)
   - valid identifier per [BCP175](https://www.rfc-editor.org/rfc/rfc6557)
 - `hourCycle` (default is locale-specific)
   - `h11`
--- a/exploration/default-registry-and-mf1-compatibility.md
+++ b/exploration/default-registry-and-mf1-compatibility.md
@@ -88,9 +88,11 @@ If no options are specified, each of the functions defaults to the following:
 
 #### Operands
 
-The operand of a date/time function is either 
+The _operand_ of a date/time function is either 
 an implementation-defined date/time type (passed in as an argument)
 or a _date/time literal value_, as defined below.
+All other _operand_ values produce a _Selection Error_ when evaluated for selection
+or a _Formatting Error_ when formatting the value.
 
 A **_<dfn>date/time literal value</dfn>_** is a non-empty string consisting of 
 one of the following:
@@ -104,7 +106,7 @@ When the offset is not present, implementations should use a floating time type
 For more information, see [Working with Timezones](https://w3c.github.io/timezone).
 
 > [!IMPORTANT]
-> The [ABNF](./spec/message.abnf) and [syntax](./spec/syntax.md) of MFv2
+> The [ABNF](/main/spec/message.abnf) and [syntax](/main/spec/syntax.md) of MFv2
 > do not formally define date/time literals. 
 > This means that a _message_ can be syntactically valid but produce
 > runtime errors due to what amounts to a "type mismatch".
--- a/exploration/default-registry-and-mf1-compatibility.md
+++ b/exploration/default-registry-and-mf1-compatibility.md
@@ -98,18 +98,37 @@ Dates and times have the following functions:
 
 The operand of a date/time function is either 
 an implementation-defined date/time type (passed in as an argument)
-or a literal that can be parsed to an implementation-defined date/time type.
+or a _date/time literal value_, as defined below.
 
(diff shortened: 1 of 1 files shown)


exploration/default-registry-and-mf1-compatibility.md
```diff
@@ -63,37 +63,137 @@ Options:
...
+    - `hourCycle`
+- `:time`
+  - defaults to `timeStyle=short`
+  - Does not permit any of these options:
+    - `dateStyle`
+    - `weekday`
+    - `era`
+    - `year`
+    - `month`
+    - `day`
```
eemeli, line 95: Why this divergence from the simpler `:date` and `:time` options bags proposed in #570? Those should each only support a single `style` option in addition to the locale options, and not any of the date-time component options.

exploration/default-registry-and-mf1-compatibility.md
```diff
@@ -63,37 +63,137 @@ Options:
...
+an implementation-defined date/time type (passed in as an argument)
+or a literal that can be parsed to an implementation-defined date/time type.
+
+Date/time literals are required to be an XMLSchema 
+[dateTime](https://www.w3.org/TR/xmlschema11-2/#dateTime),
+[time](https://www.w3.org/TR/xmlschema11-2/#time),
+or [date](https://www.w3.org/TR/xmlschema11-2/#date).
+The `timezoneOffset` of each of these formats is optional. 
+When the offset is not present, conversion to incremental time types is required
+to treat the offset identically to UTC.
```
eemeli, line 109: This is highly problematic, as the default timezone uses a different definition. Those two definitions must match so that e.g. `{14:00 :time}` will be formatted as some representation of 2pm, rather than varying depending on the system time's offset from UTC.

exploration/default-registry-and-mf1-compatibility.md
```diff
@@ -63,37 +63,137 @@ Options:
...
+    - `month`
+    - `day`
+
+#### Operands
+
+The operand of a date/time function is either 
+an implementation-defined date/time type (passed in as an argument)
+or a literal that can be parsed to an implementation-defined date/time type.
+
+Date/time literals are required to be an XMLSchema
```
eemeli, line 103: These two phrases appear to give different definitions of valid literals, the first lax and the second strict?

exploration/default-registry-and-mf1-compatibility.md
```diff
@@ -63,37 +63,137 @@ Options:
...
+- `timeStyle`
+  - `full`
+  - `long`
+  - `medium`
+  - `short` (default)
+- `calendar` (default is locale-specific)
+  - valid [Unicode Calendar Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeCalendarIdentifier)
+- `numberingSystem` (default is locale-specific)
+   - valid [Unicode Number System Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeNumberSystemIdentifier)
+- `timezone` (default is system default time zone or UTC)
```
eemeli, line 138:
```suggestion
- `timeZone` (default is system default time zone or UTC)
```

eemeli (changes requested):

exploration/default-registry-and-mf1-compatibility.md
aphillips: I could change to using `style` instead of `xxxStyle`.

The component options are important though. I handled it here by exclusion, but could do it by inclusion instead. We should give users of `:date` and `:time` control over the output, though, not just the canned stuff.

What's your thinking about excluding the date-time components?

exploration/default-registry-and-mf1-compatibility.md
aphillips: They appear to. My intention was to define "literal that can be parsed" with the strict item below it. Can clarify, assuming we all agree about the strict version.

exploration/default-registry-and-mf1-compatibility.md
eemeli: > What's your thinking about excluding the date-time components?

Keeping the `:date` and `:time` simple, but sufficient for most purposes. So if you need to e.g. only format the year from the input value, you are expected to use something like `{$date :datetime year=numeric}` for that.

Idea here is that if/once you need to go that specific, needing to use `:datetime` is a minimal burden, in particular compared to the added complexity of having them available _also_ on `:date`.

aphillips: Dates and times have tons of variations. I don't particularly think that a shorthand for `dateStyle=short`, for example, is necessarily a great idea: all-numeric dates can be ambiguous. 

`<rant>`
To be honest, what I would want _personally_ would be skeletons, as these make a nice balance of brevity and power--and it simplifies what I have to teach developers. Option bags are a pain in the ass because you have to laboriously spell out _everything_ that you want included (`{$date :datetime year=numeric}` prints `2024`). If I want to ensure an abbreviated month name (in locales that do such a thing), I need to do this, remembering that `short` means "abbreviation" for `month`:

```
{$date :datetime year=numeric month=short day=numeric}
```

Using a skeleton I have:

```
{$date :date skeleton=yMMMd}
```

I have to learn the arcane syntax, but it's a single page in the reference. No one, translators included, ever has to touch them again and they don't clutter up my patterns with all this option goo.

The option bags are fine _in JavaScript code_, but in a pattern they make a lot of visual noise. Compare:

```
{{On {$date :date year=numeric month=short day=numeric} at
    {$date :time hour=short minute=numeric} there was a disturbance in the Force.}}
{{On {$date :date skeleton=yMMMd} at {$date :date skeleton=jm} there was a disturbance in the Force.}}
```
`</rant>`

Anyhoooo... we need to agree on what to put into "the registry", which means the description in this document in the LDML45 release. I'm guessing here from our group's conversation on Monday (2024-02-05) as to what is wanted. You (@eemeli) and I could probably come to some agreement, but what we actually need is group consensus. The fact that we're having this thread suggests I need to make a matrix of options for discussion this coming week so we can all "disagree and commit" together 😜

aphillips: Never mind. I refactored based on some of my comments above and elsewhere.

exploration/default-registry-and-mf1-compatibility.md
```diff
@@ -71,46 +71,57 @@ This subsection describes the functions and options for date/time formatting.
...
-When the offset is not present, conversion to incremental time types is required
-to treat the offset identically to UTC.
-Conversion of such values to floating time types
-(such as Java's `java.time.LocalDateTime`) should omit time zone.
+When the offset is not present, implementations should use a floating time type
+(such as Java's `java.time.LocalDateTime`) to represent the time value.
 For more information, see [Working with Timezones](https://w3c.github.io/timezone).
 
+> [!IMPORTANT]
+> The [ABNF](./spec/message.abnf) and [syntax](./spec/syntax.md) of MFv2
```
eemeli, line 107:
```suggestion
> The [ABNF](../spec/message.abnf) and [syntax](../spec/syntax.md) of MF2
```

eemeli (approved):
Eh, this is fine. I appreciate the explicit note about variable string values.

exploration/default-registry-and-mf1-compatibility.md
```diff
@@ -63,37 +63,178 @@ Options:
...
+- an XMLSchema 1.1 [time](https://www.w3.org/TR/xmlschema11-2/#time)
+- an XMLSchema 1.1 [date](https://www.w3.org/TR/xmlschema11-2/#date)
+
+The `timezoneOffset` of each of these formats is optional. 
+When the offset is not present, implementations should use a floating time type
+(such as Java's `java.time.LocalDateTime`) to represent the time value.
+For more information, see [Working with Timezones](https://w3c.github.io/timezone).
+
+> [!IMPORTANT]
+> The [ABNF](/main/spec/message.abnf) and [syntax](/main/spec/syntax.md) of MFv2
```
eemeli, line 109:
```suggestion
> The [ABNF](/spec/message.abnf) and [syntax](/spec/syntax.md) of MF2
```

aphillips: Fixed by using `/main/spec/syntax.md` (etc.)

exploration/default-registry-and-mf1-compatibility.md
aphillips: I fixed this in a different way

aphillips:
I'm going to merge this now as a basis for discussion next week in the F2F.

exploration/default-registry-and-mf1-compatibility.md
eemeli: I saw that change, but those links are broken now: https://github.com/unicode-org/message-format-wg/blob/main/exploration/default-registry-and-mf1-compatibility.md#operands

aphillips: Thanks. I did a direct fix.

PR #631, opened 2024-02-07
Clarify when "additional processing" of a named value occurs

catamorphism:
This change clarifies that when an implementation "performs additional processing" on the expression of a declaration, that expression must be a variable with no annotation.

This came up when discussing #589 and is a purely editorial change.

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -150,7 +150,7 @@ that consists only of a _variable_.
 > the implementation could interpret the _placeholder_ `{$date}` as if
 > the pattern included the function `:datetime` with some set of default options.
 
-Else, the _expression_ must consist of a _literal_.
+Else, the _expression_ consists of a _literal_.
 Its resolved value is defined by _literal resolution_.
 
 > **Note**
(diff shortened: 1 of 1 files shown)


eemeli (approved):

spec/formatting.md
```diff
@@ -151,8 +150,8 @@ when resolving the value of the _expression_.
 > the implementation could interpret the _placeholder_ `{$date}` as if
 > the pattern included the function `:datetime` with some set of default options.
 
-Else, if the _expression_ consists of a _literal_,
-its resolved value is defined by _literal resolution_.
+Else, the _expression_ must consist of a _literal_.
```
aphillips, line 153: "must" is a 2119 keyword. It's not used normatively here (and thus you have it as lower case), but it would be easy to avoid using it at all.

```suggestion
Else, the _expression_ consists of a _literal_.
```

The lack of "if" communicates that this is a terminal state.

aphillips (changes requested):

aphillips (approved):

aphillips:
fast-tracking as this is not a material change

PR #632, opened 2024-02-09
Remove expression attributes from the data model

stasm:
A follow-up to #620.

Expression attributes were proposed in [a design doc](https://github.com/unicode-org/message-format-wg/blob/main/exploration/expression-attributes.md), which was deemed out of scope for LDML 45. Instead, we decided to reserve the most likely syntax for the; this  was done in #592.

That PR reserved the *syntax*, but also added attributes as a concept in the data model. I think that's wrong and I'd like to suggest to remove expression attributes from the data model.

Our formatting spec explicitly states that the reason for reserving attributes was to be able to perform *syntax* validity checks:

> _Attributes_ are reserved for future standardization. Other than checking for valid syntax, they SHOULD NOT affect the processing or output of a _message_.

Furthermore, our stability policy doesn't guarantee anything about the data model:

> Updates to this specification will not change the syntactical meaning, the runtime output, or other behaviour of valid messages written for earlier versions of this specification that only use functions and expression attributes defined in this specification. Updates to this specification will not remove any syntax provided in this version. Future versions MAY add additional structure or meaning to existing syntax.

Interestingly, the policy mentions attributes by name, which I'm also fixing in this PR. The mention was added in #472 together with the rest of the wording of the policy; a mere month after attributes were first proposed in a doc.

### DIFF
--- a/spec/README.md
+++ b/spec/README.md
@@ -84,7 +84,7 @@ A reference to a _term_ looks like this.
 Updates to this specification will not change
 the syntactical meaning, the runtime output, or other behaviour
 of valid messages written for earlier versions of this specification
-that only use functions and expression attributes defined in this specification.
+that only use functions defined in this specification.
 Updates to this specification will not remove any syntax provided in this version.
 Future versions MAY add additional structure or meaning to existing syntax.
 
--- a/spec/data-model/README.md
+++ b/spec/data-model/README.md
@@ -136,33 +136,24 @@ interface LiteralExpression {
   type: "expression";
   arg: Literal;
   annotation?: FunctionAnnotation | UnsupportedAnnotation;
-  attributes?: Attribute[];
 }
 
 interface VariableExpression {
   type: "expression";
   arg: VariableRef;
   annotation?: FunctionAnnotation | UnsupportedAnnotation;
-  attributes?: Attribute[];
 }
 
 interface FunctionExpression {
   type: "expression";
   arg?: never;
   annotation: FunctionAnnotation;
-  attributes?: Attribute[];
 }
 
 interface UnsupportedExpression {
   type: "expression";
   arg?: never;
   annotation: UnsupportedAnnotation;
-  attributes?: Attribute[];
-}
-
-interface Attribute {
-  name: string;
-  value?: Literal | VariableRef;
 }
 ```
 
@@ -251,7 +242,6 @@ interface MarkupOpen {
   kind: "open";
   name: string;
   options?: Option[];
-  attributes?: Attribute[];
 }
 
 interface MarkupStandalone {
@@ -259,14 +249,12 @@ interface MarkupStandalone {
   kind: "standalone";
   name: string;
   options?: Option[];
-  attributes?: Attribute[];
 }
 
 interface MarkupClose {
   type: "markup";
   kind: "close";
   name: string;
-  attributes?: Attribute[];
 }
 ```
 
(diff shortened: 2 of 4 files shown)


eemeli:
Expression attributes were discussed on the 8 Jan call, when we agreed to specifically reserve _only_ their formatting runtime meaning. The [meeting notes](https://github.com/unicode-org/message-format-wg/blob/main/meetings/2024/notes-2024-01-08.md#topic-expression-attributes) do not appear to have explicitly captured it, but the discussion did include the data model, where they were kept; this is noted in the subsequent PR #592.

stasm:
Hmm, that's different from what I remember, but I may be wrong. My understanding was that there was no agreement about whether they're buildtime-only, runtime-only, or both, which is what made us only reserve the syntax.

Also, the last comment in the meeting minutes reads (emphasis mine):

> APP - we will incorporate syntax for attrib as reserved, on expression and markup and **make them currently explicitly not do anything at runtime**. Reserved for future standardization. You do not error on them, you do not do anything with them.

spec/README.md
```diff
@@ -84,7 +84,7 @@ A reference to a _term_ looks like this.
 Updates to this specification will not change
 the syntactical meaning, the runtime output, or other behaviour
 of valid messages written for earlier versions of this specification
-that only use functions and expression attributes defined in this specification.
+that only use functions defined in this specification.
```
aphillips, line 87: I think we'll want to keep this change, but reject the rest of this, since `attribute` is part of the syntax and we guarantee round trip with the data model (we do not claim that attributes in the data model do anything)

aphillips (changes requested):
I think we need to fix the dangling reference in the stability policy, but otherwise forego this change? See comment.

stasm:
There have been a number of developments since I originally filed this PR:

* #643 make some guarantees about the stability of the *data model*; they are less strict than the guarantees about the syntax.
* #644 defines the goals of recommending a canonical data model.
* #633 now explains the differences between the *canonical* data model and the JSON and XML serializations of it.

In light of these, I think we should keep the `attribute` field, but change its type from the concrete `Attribute` interface to something similar to `UnsupportedAnnotation`.

The interface is currently defined as:

```
interface Attribute {
   name: string;
   value?: Literal | VariableRef;
}
```

If in the future we decide to un-reserve attributes, we will be constrained by this exact data model, by our stability policy. Alternatively, we will add a new interface, under a new name, since `Attribute` will already be taken.

Instead, we should now start with a generic representation of the reserved syntax for the purpose of the round-trip, and add the proper `Attribute` later on, if ever.

```
// Now:
interface LiteralExpression {
   type: "expression";
   arg: Literal;
   annotation?: FunctionAnnotation | UnsupportedAnnotation;
   attributes: UnsupportedAnnotation[];
}

// Maybe in the future:
interface LiteralExpression {
   type: "expression";
   arg: Literal;
   annotation?: FunctionAnnotation | UnsupportedAnnotation;
   attributes: Attribute[] | UnsupportedAnnotation[];
}
```

eemeli:
> If in the future we decide to un-reserve attributes, we will be constrained by this exact data model, by our stability policy. Alternatively, we will add a new interface, under a new name, since `Attribute` will already be taken.

Given that attributes are defined in the syntax as https://github.com/unicode-org/message-format-wg/blob/0c1d40aaf3503b6bde6cc7914dca19758995e13a/spec/message.abnf#L45
what would be the benefit of defining attributes differently in the data model? I don't really see how we might end up with a different representation, given that we're constrained here by the syntax.

spec/README.md
stasm: #654 was the cherry-pick PR.

PR #634, opened 2024-02-12
[DESIGN] Maintaining the Standard, Optional and Unicode Namespace Function Sets

aphillips:
Capture for discussion the general ideas for how to organize the registry in the post-45 time period.

Notably, @sffc [mentioned number skeletons](https://github.com/unicode-org/message-format-wg/pull/621#discussion_r1482260505) in a comment on #621:

> Note: My original idea from 5 years ago was that MF would use number skeletons and then become the specification that defines number skeletons since they are currently specified only in the ICU documentation.

This goaded me to write down some of the concepts floating around in my head about how to manage the registry going forwards.

### DIFF
--- a/exploration/maintaining-registry.md
+++ b/exploration/maintaining-registry.md
@@ -10,7 +10,7 @@ Status: **Proposed**
 		<dt>First proposed</dt>
 		<dd>2024-02-12</dd>
 		<dt>Pull Requests</dt>
-		<dd>#000</dd>
+                <dd>#634</dd>
 	</dl>
 </details>
 
--- a/exploration/maintaining-registry.md
+++ b/exploration/maintaining-registry.md
@@ -36,7 +36,7 @@ runtimes in a wholly consistent manner.
 Because we want broad adoption in many different programming environments
 and because the capabilities 
 and functionality available in these environments varies widely,
-the default function registry must be conservative in what it required.
+the default function registry must be conservative in what it requires.
 
 At the same time, we want to promote message interoperability.
 Even though a feature cannot be adopted by all platforms,
--- a/exploration/maintaining-registry.md
+++ b/exploration/maintaining-registry.md
@@ -18,18 +18,17 @@ Status: **Proposed**
 
 _What is this proposal trying to achieve?_
 
-Describe how MFv2 will manage the default function registry
-as well was related function registries intended to promot interoperability.
+Describe how to manage the default function registry
+as well was related function registries intended to promote interoperability.
 
 ## Background
 
 _What context is helpful to understand this proposal?_
 
-> [!NOTE]
-> This design proposal is for the Tech Preview period.
-
-MessageFormat v2 envisions providing a "default registry" that all conformant
-implementations are required to implement.
+MessageFormat v2 includes a "default function registry".
+Implementations are required to implement all of the _selectors_ 
+and _formatters_ in this registry,
+including _operands_, _options_, and option values.
 Our goal is to be as universal as possible, 
 making MFv2's message syntax available to developers in many different
 runtimes in a wholly consistent manner.
@@ -62,14 +61,17 @@ for the various types of function registry.
 _What use-cases do we see? Ideally, quote concrete examples._
 
 As an implementer, I want to know what functions, options, and option values are
-required to claim support for MF2.
-I want to know what the options and their values mean.
-I also need to be able to implement all of the required functions in my runtime environment
-without difficulty.
-I don't want to be required to exactly follow CLDR or a specific release of CLDR,
-in case my local I18N support differs from what CLDR provides.
+required to claim support for MF2:
+- I want to know what options I am required to implement.
+- I want to know what the values of each option are.
+- I want to know what the options and their values mean.
+- I want to be able to implement all of the required functions using my runtime environment
+  without difficulty.
+- I want to be able to use my local I18N APIs, which might use an older release of CLDR
+  or might not be based on CLDR data at all.
+  This means that my output might not match that of an CLDR-based implementation.
 
-As an implementer, user, translators, tools author I expect functions, options
+As an implementer, user, translator, or tools author I expect functions, options
 and option values to be stable.
 The meaning and use of these, once established, should never change.
 Messages that work today should work tomorrow.
@@ -83,24 +85,41 @@ without being required to implement other APIs that I'm not ready for.
 As an implementer, I want to be assured that functions or options added in the future
 will not conflict with functions or options that I have created for my local users.
 
-As a developer, I want to be able to implement local functions or local options
+As a developer, I want to be able to implement my own local functions or local options
 and be assured that these do not conflict with future additions by the core standard.
 
 As a tools developer, I want to track both required and optional function development
 so that I can produce consistent support for messages that use these features.
 
-As a translator, I want all messages to be consistent in their meaning.
+As a translator, I want messages to be consistent in their meaning.
 I want functions and options to work consistently.
-I don't want to relearn selection or formatting rules for different implementations.
+I want to selection and formatting rules to be consistent so that I only have
+to learn them once and so that there are no local quirks.
 
-As a user, I want to be able to use required functions and their options.
+As a user, I want to be able to use required functions and their options in my messages.
 I want to be able to quickly adopt new additions as my implementation supports them
 or be able to choose plug-in or shim implementations.
-I never want to have to find/rewrite a message because a function or its option has changed.
+I never want to have to rewrite a message because a function or its options have changed.
 
 ## Requirements
 
 _What properties does the solution have to manifest to enable the use-cases above?_
+
+The default registry needs to describe the minimum set of selectors and formatters
+needed to create messages effectively.
+This must be compatible with ICU MessageFormat 1 messages.
+
+There must be a clear process for the creation of new selectors that are required
+by the default registry, 
+which includes a maturation process that permits implementer feedback.
+
+There must be a clear process for the creation of new formatters that are required
+by the default registry, 
+which includes a maturation process that permits implementer feedback.
+
+There must be a clear process for the addition of options or option values that are required
+by the default registry, 
+which includes a maturation process that permits implementer feedback.
 
 ## Constraints
 
@@ -110,7 +129,7 @@ _What prior decisions and existing conditions limit the possible design?_
 
 _Describe the proposed solution. Consider syntax, formatting, errors, registry, tooling, interchange._
 
-To address the above problem, in addition to the default registry,
+To address the above requirements, in addition to the default registry,
 create a template for registry functions
 and a process for proposing and evaluating these functions for inclusion
 into the MessageFormat v2 function registry.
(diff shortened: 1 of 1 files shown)


macchiati (approved):
I like the direction. There are a few sections that remain to be filled out (eg _What properties does the solution have to manifest to enable the use-cases above?_
). Approving, assuming you want to go ahead with this, and flesh out more later.

exploration/maintaining-registry.md
```diff
@@ -0,0 +1,114 @@
...
+> This design proposal is for the Tech Preview period.
+
+MessageFormat v2 envisions providing a "default registry" that all conformant
+implementations are required to implement.
+Our goal is to be as universal as possible, 
+making MFv2's message syntax available to developers in many different
+runtimes in a wholly consistent manner.
+Because we want broad adoption in many different programming environments
+and because the capabilities 
+and functionality available in these environments varies widely,
```
catamorphism, line 38:
```suggestion
and functionality available in these environments vary widely,
```

exploration/maintaining-registry.md
```diff
@@ -0,0 +1,114 @@
...
+the default function registry must be conservative in what it required.
```
catamorphism, line 39:
```suggestion
the default function registry must be conservative in what it requires.
```

exploration/maintaining-registry.md
```diff
@@ -0,0 +1,114 @@
...
+into the MessageFormat v2 function registry.
+
+There would be three levels of expected maturity:
+
+- **Default Registry** includes functions that are normatively required by all implementations.
+  Such entries will be limited in scope to functions that can be
+  implemented in nearly any programming environment.
+  > Examples: `:string`, `:number`, `:date`
+- **Recommended for General Implementation**
+  ("RGI", deliberately similar to RGI in emoji, tho' we may want to change the name)
```
sffc, line 88: Thought: should these "RGI" functions be prefixed in some way, like a vendor prefix, draft prefix, etc? For example, the definition of `:person` might need to change over time as more environments adopt it. Perhaps it should start as `:person:draft` or `:person:2024a` or something like that, only being promoted to `:person` once the design is finalized.

exploration/maintaining-registry.md
```diff
@@ -0,0 +1,114 @@
...
+  includes functions that are not
+  normatively required but whose names, operands, and options are recommended.
+  Implementations are _strongly_ encouraged to use these function signatures
+  when implementing the described functionality.
+  This will promote message interoperability
+  and reduce the learning curve for developers, tools, and translators.
+  > Examples: We don't currently have any, but potential work here
+  > might includes personal name formatting, gender-based selectors, etc.
+- **Unicode Extensions** includes optional functionality that implementations
```
eemeli, line 97: I get the RGI bucket of definitions, but I'm having trouble seeing what use we'd make of the `u` space. If it's just for stuff that we'll likely standardise in the anon namespace later on, I think some other namespace name would be better for that, like `x` or `next`. Or is it meant for something different?

exploration/maintaining-registry.md
```diff
@@ -0,0 +1,114 @@
...
+
+There would be three levels of expected maturity:
+
+- **Default Registry** includes functions that are normatively required by all implementations.
+  Such entries will be limited in scope to functions that can be
+  implemented in nearly any programming environment.
+  > Examples: `:string`, `:number`, `:date`
+- **Recommended for General Implementation**
+  ("RGI", deliberately similar to RGI in emoji, tho' we may want to change the name)
+  includes functions that are not
```
eemeli, line 89: This should also include options, so that e.g. `:number skeleton` could be specified, but would not be required.

exploration/maintaining-registry.md
aphillips: RGI functions could be prefixed, but I think they are meant to pollute the common function namespace. We have `:u` namespace reserved below. The natural order of progression would be:

- `:u:person` <= early days proposal, leads to:
- `:person` "RGI" <= implement this if you wish to, leads eventually (maybe) to:
- `:person` <= implement this if you want to claim you support MF2

exploration/maintaining-registry.md
aphillips: One example is something @sffc mentioned. Suppose we wanted to define date skeletons or number skeletons in a referenceable way. We don't want to add them to `:datetime` or `:number`, because then everyone would have to implement them. But we could define them as `:u:skeleton` options which implementers SHOULD follow if they want to provide this feature.

Similarly, this provides a way to prototype functionality before making it RGI.

We can debate spelling of it. I chose `u` for familiarity.

exploration/maintaining-registry.md
aphillips: This was my intention. I'll add text.

exploration/maintaining-registry.md
eemeli: I'd prefer the prototyping to use the same name as what we ultimately intend for the thing, be it a function or an option. We could classify each defined thing according to some progressive schema like alpha/beta/stable, and only promise that the "stable" stuff will actually be stable.

exploration/maintaining-registry.md
```diff
@@ -0,0 +1,165 @@
...
+
+<details>
+	<summary>Metadata</summary>
+	<dl>
+		<dt>Contributors</dt>
+		<dd>@aphillips</dd>
+		<dt>First proposed</dt>
+		<dd>2024-02-12</dd>
+		<dt>Pull Requests</dt>
+		<dd>#000</dd>
```
aphillips, line 13:
```suggestion
      <dd>#634</dd>
```

exploration/maintaining-registry.md
```diff
@@ -155,73 +169,80 @@ _What prior decisions and existing conditions limit the possible design?_
...
-The three registries are:
-
-1. **Default Registry**
-This is a _function_ registry.
-It includes _functions_ (and only functions) that are normatively required to be
-implemented by all implementations.
-Each function describes it's operand or operands,
+The MessageFormat WG will a set of specifications
+that standardize the implementation of functions and options in the default namespace of 
+MessageFormat v2 beginning with the LDML46 release.
```
eemeli, line 174: This phrase seems to be missing some words.

exploration/maintaining-registry.md
```diff
@@ -155,73 +169,80 @@ _What prior decisions and existing conditions limit the possible design?_
...
+Future updates to these sets of functions and options will coincide with LDML releases.
+
+Each _function_ is described by a single specification document.
+Each such document will use a common template.
+A _function_ can be a _formatting function_,
+a _selector_,
+or both.
+
+The specification will indicate if the _formatting function_,
+the _selector function_, or, where applicable, both are `Standard` or `Optional`.
```
eemeli, line 186: We really should finally align on how to speak about resolved values, because that's clearly relevant here as well. Note also that _formatting function_ and _selector function_ are not terms that we've defined so far, and I don't think that they should be.

exploration/maintaining-registry.md
```diff
@@ -155,73 +169,80 @@ _What prior decisions and existing conditions limit the possible design?_
...
+Each _option_ must define which values it accepts.
+An _option_  is either `Standard` or `Optional`.
+
+_Functions_ or _options_ that have an `Optional` status
+must have a maturity level assigned.
+The maturity levels are:
+- **Proposed**
+- **Accepted**
+- **Released**
+- **Deprecated**
```
eemeli, line 199: Are these terms used by CLDR or ICU already, or are we inventing something new for MF2? My own preference would be to use instead something like alpha/beta/[no tag]/deprecated, which probably are better understood by a layperson.

I also can't figure out what the difference between "Accepted" and "Released" might be.

exploration/maintaining-registry.md
```diff
@@ -155,73 +169,80 @@ _What prior decisions and existing conditions limit the possible design?_
...
-promotion to the RGI or default registry in a future release.
+The CLDR-TC reserves the `u:` namespace for use by the Unicode Consortium.
+This namespace can contain _functions_ or _options_.
+Implementations are not required to implement these _functions_ or _options_
+and may adopt or ignore them at their discretion,
+but are encouraged to implement these items.
+
+Items in the Unicode Reserved Namespace are stable and subject to stability guarantees.
+This namespace might sometimes be used to incubate functionality before
+promotion to the default namespace in a future release.
```
eemeli, line 245: Given that the `u:` space is stable, doesn't this mean that anything "incubated" there would need to remain there even after its "promotion"? And given that we are also proposing a process for introducing new items directly in the root namespace, why/what would we want to incubate in such a manner? I'd prefer this to be left out.

exploration/maintaining-registry.md
```diff
@@ -230,28 +251,31 @@ In such cases, the `u:` namespace version is retained, but deprecated.
 > but it is not universally available and could represent a barrier to adoption
 > if normatively required.
 
-Any registry entry goes through a development process that includes these levels of maturity:
+All `Standard`, `Optional`, and Unicode namespace function or option specifications goes through 
+a development process that includes these levels of maturity:
```
eemeli, line 255: The list of levels below doesn't match the lists above, which don't include Rejected at all and only include Proposed and Accepted for Optional items.

eemeli (commented):
This seems like a definite improvement; see line comments below. On a meta level, I'd appreciate more explicit communication about the status of this long-running PR, especially as it's being deeply refactored like this. I can't tell if this is truly ready for review, or if it'll still be iterated on before further consideration.

exploration/maintaining-registry.md
aphillips: "Accepted" is a specification that has been reviewed and will be part of the next LDML release (but it not part of the current one).

These terms are similar to those used by the UTC, but there's no real attempt to be rigorous about them. The names are provisional and I'm open to suggestions. alpha/beta would be okay, once proposals are accepted, but you only have to look at emoji or language subtags to see that a lot of proposals get proposed that would never reach `alpha`.

exploration/maintaining-registry.md
aphillips: > Given that the u: space is stable, doesn't this mean that anything "incubated" there would need to remain there even after its "promotion"?

Yes. I think I say that somewhere?

> And given that we are also proposing a process for introducing new items directly in the root namespace, why/what would we want to incubate in such a manner? I'd prefer this to be left out.

This is permission to promote something from `u:` to standard, sort of like the way browsers use prefixes in CSS. This might be a bad idea.

exploration/maintaining-registry.md
echeran: Here's what I dug up about status labels for ICU/CLDR/Unicode

ICU:
* [ICU API statuses](https://unicode-org.github.io/icu/userguide/icu/design.html#icu-api-compatibility) - `draft`, `stable`, `deprecated`, `obsolete` (and `discouraged` in Java only?)
* `technology preview` is not defined, but has been used for a very long time ([example from ICU v4.4](https://unicode-org.github.io/icu/userguide/strings/utf-8.html#abstract-text-apis)). It is like draft, and draft is more or less like beta, but draft additionally signals to the user that there is no guarantee of stability, while draft means that the design iteration seems done so there probably won't be major changes.

CLDR: [the processes](https://cldr.unicode.org/index/process) have different labels depending on what it is. Data fields have statuses like unconfirmed. But for releases of CLDR, you have `alpha`, `beta`, and the unqualified final release itself.

Unicode data: at https://www.unicode.org/Public/, you have a `draft` subfolder that acts like the `main` trunk. Each versioned release has alpha, then beta, then the final release.

macchiati:
In CLDR, we introduced the term 'technology preview' in CLDR 24 (2013),
with a meaning corresponding to ICU's, except applied to XML structure &
spec, instead of APIs. That is a status of a portion of the spec and/or
data *within* a release.

For the data, we have 4 statuses, as Elango said, but those aren't relevant.

PR #643, opened 2024-02-13
Add data model to stability policy

### LINKED ISSUE #640: add data model to stability policy

aphillips:
Fixes #640

### DIFF
--- a/spec/README.md
+++ b/spec/README.md
@@ -124,16 +124,29 @@ All other names in these categories are reserved for the use of implementations 
 > Users defining custom names SHOULD include at least one character outside these ranges
 > to ensure that they will be compatible with future versions of this specification.
 
-The data model is not _forward_ compatible. That is, later versions of this specification MAY introduce changes in the data model that
-are incompatible with the data model in this version of the specification.
+Later versions of this specification will not introduce changes
+to the data model that would result in a data model representation
+based on this version being invalid.
 
-> For example, a future version of this spec might introduce a new keyword
-> which is represented by a new `interface` in the data model.
-> Such a data model could not be processed by the data model in this spec.
+> For example, existing interfaces or fields will not be removed.
 
-The data model is backwards compatible. That is, later versions of this specification will not introduce changes to the data model
-that make existing data model representations invalid.
+Later versions of this specification MAY introduce changes
+to the data model that would result in future data model representations
+not being valid for implementations of this version of the data model.
 
-> For example, a future version of this specification could add additional fields
-> to an existing `interface`, but could not remove any fields or change their
-> names.
+> For example, a future version could introduce a new keyword,
+> whose data model representation would be a new interface
+> that is not recognized by this version's data model.
+
+Later specification versions will not introduce syntax that cannot be
+processed by this version of the data model.
+However, the results of such processing might not be the same as that produced
+but future version of the data model.
+
+> For example, a future version could introduce a new keyword.
+> The future version's data model would provide an interface for that keyword
+> while this version of the data model would parse the value into
+> the interface `Unsupported_Statement`.
+> Both data models would be "valid" in their context,
+> but this version's would be missing any functionality for the new statement type.
+
--- a/spec/README.md
+++ b/spec/README.md
@@ -144,7 +144,7 @@ represented by this version of the data model.
 > For example, a future version could introduce a new keyword.
 > The future version's data model would provide an interface for that keyword
 > while this version of the data model would parse the value into
-> the interface `Unsupported_Statement`.
+> the interface `UnsupportedStatement`.
 > Both data models would be "valid" in their context,
 > but this version's would be missing any functionality for the new statement type.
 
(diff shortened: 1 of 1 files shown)


spec/README.md
```diff
@@ -123,3 +123,17 @@ All other names in these categories are reserved for the use of implementations
 > [!NOTE]
 > Users defining custom names SHOULD include at least one character outside these ranges
 > to ensure that they will be compatible with future versions of this specification.
+
+Later versions of this specification MAY introduce changes in the data model that
```
macchiati, line 127:
```suggestion
The data model is not _forward_ compatible. That is, later versions of this specification MAY introduce changes in the data model that
```

spec/README.md
```diff
@@ -123,3 +123,17 @@ All other names in these categories are reserved for the use of implementations
...
+are incompatible with the data model in this version of the specification.
+
+> For example, a future version of this spec might introduce a new keyword
+> which is represented by a new `interface` in the data model.
+> Such a data model could not be processed by the data model in this spec.
+
+Later versions of this specification will not introduce changes to the data model
```
macchiati, line 134:
```suggestion
The data model is backwards compatible. That is, later versions of this specification will not introduce changes to the data model
```

spec/README.md
```diff
@@ -123,3 +123,17 @@ All other names in these categories are reserved for the use of implementations
 > [!NOTE]
 > Users defining custom names SHOULD include at least one character outside these ranges
 > to ensure that they will be compatible with future versions of this specification.
+
```
macchiati, line 126: Reverse the order of these, so that the "yes it is backwards compatible" message comes first.

macchiati (changes requested):

spec/README.md
aphillips: I incorped your suggestions, but think they might be saying the wrong thing.

The data model is intended (I think) to be "forwards compatible" from the point of view that a valid data model representation of this version will still be valid later.

The data mode is intended (I think) to be _potentially_ non-backwards compatible from the point of view that a future valid data model representation might not be readable by current data model implementations (because it uses a new feature).

Future version of the syntax are guaranteed to be compatible with the current data model, but the results might be different from the future data model's representation.

aphillips: I made a set of changes based on the above.

macchiati: LGTM now

spec/README.md
```diff
@@ -123,3 +123,30 @@ All other names in these categories are reserved for the use of implementations
...
+Later versions of this specification MAY introduce changes
+to the data model that would result in future data model representations
+not being valid for implementations of this version of the data model.
+
+> For example, a future version could introduce a new keyword,
+> whose data model representation would be a new interface
+> that is not recognized by this version's data model.
+
+Later specification versions will not introduce syntax that cannot be
+processed by this version of the data model.
```
catamorphism, line 142:
```suggestion
represented by this version of the data model.
```

spec/README.md
```diff
@@ -123,3 +123,30 @@ All other names in these categories are reserved for the use of implementations
...
+However, the results of such processing might not be the same as that produced
+but future version of the data model.
```
catamorphism, line 144:
```suggestion
However, a future version of the specification could change the meaning
of a particular data model.
```

I'm not sure if this is exactly what you meant, but "the results of such processing" confused me. I think you're trying to say that future changes to the spec may change the semantics of a particular data model instance, but will never expand the set of valid data model instances?

spec/README.md
```diff
@@ -123,3 +123,30 @@ All other names in these categories are reserved for the use of implementations
...
+
+> For example, a future version could introduce a new keyword.
+> The future version's data model would provide an interface for that keyword
+> while this version of the data model would parse the value into
+> the interface `Unsupported_Statement`.
```
eemeli, line 149:
```suggestion
> the interface `UnsupportedStatement`.
```

spec/README.md
eemeli: I think it may actually be clearer to leave out this sentence, especially as the following note clarifies the meaning sufficiently.
```suggestion
```

catamorphism (approved):
LGTM

spec/README.md
```diff
@@ -123,3 +123,28 @@ All other names in these categories are reserved for the use of implementations
 > [!NOTE]
 > Users defining custom names SHOULD include at least one character outside these ranges
 > to ensure that they will be compatible with future versions of this specification.
+
+Later versions of this specification will not introduce changes
+to the data model that would result in a data model representation
+based on this version being invalid.
+
+> For example, existing interfaces or fields will not be removed.
```
stasm, line 131: What about changes to the existing fields and interfaces?

IIUC, this parts says that if we every un-reserve, say, the `%` sigil, we will also need to keep it on the list of sigils inside `UnsupportedAnnotation`:

```
  sigil: "!" | "%" | "^" | "&" | "*" | "+" | "<" | ">" | "?" | "~";
  source: string;
}
```

Otherwise, (i.e. if we remove `%` from the list), data model representations produced by older implementations will not be valid.

---

What was the reason for including the `sigil` field in `UnsupportedAnnotation`? Could we move the sigil into the `source`?

eemeli: Good point. Including it in the `source` string would be a better choice for future compatibility.

ryzokuken (approved):

PR #646, opened 2024-02-14
Update spec as if PR #645 was accepted

catamorphism:
**DO NOT MERGE**

This PR is for reference purposes when reviewing #645, to show how the spec would change if the design doc in #645 was accepted. It should not be merged as-is, and also reflects a slightly earlier version of the design doc. It should still serve to give a sense of what the spec _would_ look like if this design doc was accepted.


aphillips:
(chair hat off)

I think this is too specific. We give implementations a lot of leeway about how to implement things internally. This is one good way of organizing it. There are elements that, on an initial read, I would nitpick, but let's discuss the high-level requirements before diving into that. I appreciate the effort, though.

PR #648, opened 2024-02-14
Implement changes to date/time functions

aphillips:
This is per the 2024-02-14 teleconference.

### DIFF
--- a/exploration/default-registry-and-mf1-compatibility.md
+++ b/exploration/default-registry-and-mf1-compatibility.md
@@ -241,7 +241,7 @@ The function `:datetime` has the following options:
 > The following options do not have default values because they are only to be used
 > as overrides for locale-and-value dependent implementation-defined defaults.
 
-The function `:time` has following additional options:
+The function `:time` has following options:
 - `hourCycle` (default is locale-specific)
   - `h11`
   - `h12`
--- a/exploration/default-registry-and-mf1-compatibility.md
+++ b/exploration/default-registry-and-mf1-compatibility.md
@@ -155,12 +155,12 @@ The function `:datetime` has these function-specific _style_ options.
   - `full`
   - `long`
   - `medium`
-  - `short` (default)
+  - `short`
 - `timeStyle`
   - `full`
   - `long`
   - `medium`
-  - `short` (default)
+  - `short`
 
 The function `:date` has these function-specific _style_ options:
 - `style`
--- a/exploration/default-registry-and-mf1-compatibility.md
+++ b/exploration/default-registry-and-mf1-compatibility.md
@@ -154,26 +154,26 @@ The function `:datetime` has these function-specific _style_ options.
   - `full`
   - `long`
   - `medium`
-  - `short`
+  - `short` (default)
 - `timeStyle`
   - `full`
   - `long`
   - `medium`
-  - `short`
+  - `short` (default)
 
 The function `:date` has these function-specific _style_ options:
 - `style`
   - `full`
   - `long`
   - `medium`
-  - `short`
+  - `short` (default)
 
 The function `:time` has these function-specific _style_ options:
 - `style`
   - `full`
   - `long`
   - `medium`
-  - `short`
+  - `short` (default)
 
 ##### Field Options
 
@@ -181,16 +181,14 @@ Field options describe which fields to include in the formatted output
 and what format to use for that field.
 Only those fields specified in the _annotation_ appear in the formatted output.
 
+> [!NOTE]
+> Field options do not have default values because they are only to be used
+> to compose the formatter.
+> Each value must be specified.
+
 The _field_ options are defined as follows:
 
 All functions have the following option:
-- `timeZoneName`
-  - `long`
-  - `short`
-  - `shortOffset`
-  - `longOffset`
-  - `shortGeneric`
-  - `longGeneric`
 
 The function `:datetime` has the following options:
 - `weekday`
@@ -231,14 +229,32 @@ The function `:datetime` has the following options:
   - `h12`
   - `h23`
   - `h24`
+- `timeZoneName`
+  - `long`
+  - `short`
+  - `shortOffset`
+  - `longOffset`
+  - `shortGeneric`
+  - `longGeneric`
 
-The function `:time` has following option, whose purpose is to override the default hour cycle:
+> [!NOTE]
+> The following options do not have default values because they are only to be used
+> as overrides for locale-and-value dependent implementation-defined defaults.
+
+The function `:time` has following additional options:
 - `hourCycle` (default is locale-specific)
   - `h11`
   - `h12`
   - `h23`
   - `h24`
-
+- `timeZoneName`
+  - `long`
+  - `short`
+  - `shortOffset`
+  - `longOffset`
+  - `shortGeneric`
+  - `longGeneric`
+  
 The followind date/time options are *not* part of the default registry.
 Implementations SHOULD avoid creating options that conflict with these, but
 are encouraged to track development of these options during Tech Preview:
--- a/exploration/default-registry-and-mf1-compatibility.md
+++ b/exploration/default-registry-and-mf1-compatibility.md
@@ -188,8 +188,6 @@ Only those fields specified in the _annotation_ appear in the formatted output.
 
 The _field_ options are defined as follows:
 
-All functions have the following option:
-
 The function `:datetime` has the following options:
 - `weekday`
   - `long`
--- a/exploration/default-registry-and-mf1-compatibility.md
+++ b/exploration/default-registry-and-mf1-compatibility.md
@@ -145,7 +145,8 @@ A function can use either the appropriate _style_ options for that function
 or can use a collection of _field options_ (but not both) to control the formatted 
 output.
 
-If both are specified, the _style_ option takes precedence.
+If both are specified, an _Invalid Expression_ error MUST be emitted
+and a _fallback value_ used as the resolved value of the _expression_.
 
 ##### Style Options
 
--- a/exploration/default-registry-and-mf1-compatibility.md
+++ b/exploration/default-registry-and-mf1-compatibility.md
@@ -117,7 +117,7 @@ For more information, see [Working with Timezones](https://w3c.github.io/timezon
 > The [ABNF](/spec/message.abnf) and [syntax](/spec/syntax.md) of MFv2
 > do not formally define date/time literals. 
 > This means that a _message_ can be syntactically valid but produce
-> runtime errors due to what amounts to a "type mismatch".
+> an _Operand Mismatch Error_ at runtime.
 
 > [!NOTE]
 > String values passed as variables in the _formatting context_'s
--- a/exploration/default-registry-and-mf1-compatibility.md
+++ b/exploration/default-registry-and-mf1-compatibility.md
@@ -185,7 +185,6 @@ Only those fields specified in the _annotation_ appear in the formatted output.
 > [!NOTE]
 > Field options do not have default values because they are only to be used
 > to compose the formatter.
-> Each value must be specified.
 
 The _field_ options are defined as follows:
 
--- a/exploration/default-registry-and-mf1-compatibility.md
+++ b/exploration/default-registry-and-mf1-compatibility.md
@@ -33,6 +33,9 @@ Therefore, addition to this list requires a higher level of rigor.
 
 ### Numbers
 
+> [!IMPORTANT]
+> This section is replaced by the design for number selection
+
 Function name: `:number`
 
 Aliases: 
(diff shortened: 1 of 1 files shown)


exploration/default-registry-and-mf1-compatibility.md
```diff
@@ -226,6 +232,23 @@ The functions `:datetime` and `:time` have the following options:
...
+  - `h24`
+
+The followind date/time options are *not* part of the default registry.
+Implementations SHOULD avoid creating options that conflict with these, but
+are encouraged to track development of these options during Tech Preview:
+- `calendar` (default is locale-specific)
+  - valid [Unicode Calendar Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeCalendarIdentifier)
+- `numberingSystem` (default is locale-specific)
+   - valid [Unicode Number System Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeNumberSystemIdentifier)
+- `timeZone` (default is system default time zone or UTC)
```
macchiati, line 249: not having the timezone setting will make it difficult to test in server environments where the 'default' is not used.

aphillips: I agree, even in the case where the implementation provides a message-context API for setting the time zone.

The argument is that you "ought" to use a Temporal type with real time zone support--which I don't disagree that one _should_ do that. But there are lots of applications that can't or systems that do not provide anything but an incremental time value and need to set the time zone.

Not providing an option to set that seems problematic.

On the other hand, these specific options will likely enter Tech Preview as "RGI" and we'll get feedback from users.

exploration/default-registry-and-mf1-compatibility.md
```diff
@@ -173,6 +175,12 @@ The function `:time` has these function-specific _style_ options:
...
 
+##### Field Options
+
+Field options describe which fields to include in the formatted output
+and what format to use for that field.
+Only those fields specified in the _annotation_ appear in the formatted output.
+
 The _field_ options are defined as follows:
 
 All functions have the following option:
```
sffc, line 186: `timeZoneName` should not be in `:date`

exploration/default-registry-and-mf1-compatibility.md
macchiati: Shane has a good point. This is a bit like currencies with numbers; the input should include both, and if you don't have an input timezone, then the assumption is UTC.

aphillips: > I would suggest formatting with "GMT+?" if the time zone is not specified but the format requires a time zone.

Yes, if the format requires a time zone and one is not available. Although that generally means a floating time value such as `PlainDate` or `LocalTime`. The system time zone will be wrong in a server context, but not all code runs in a server context. The correct way to handle this is to provide message-level contextual time zone (an implementation detail) including API access to same--not to produce shitty date/time display for users of non-temporal data types.

> This is a bit like currencies with numbers; the input should include both, and if you don't have an input timezone, then the assumption is UTC.

The input **_should_** include both, absolutely. However, there are plenty of `java.util.Date` and JS `Date` objects in the world which are epoch times in UTC for which the display should be in some specific time zone. And Temporal isn't quite shipped in JS.

For these sorts of incremental time values, the ability to say:

```
Your package arrived: {$deliveryDate :datetime timeZone=$customerTZ}
```

... is sometimes useful.

I even have ample implementation experience of zoned values needing to be floated or re-zoned for display purposes. Doing this in Amazon's message formatter was easy to explain and worked effectively, although in our case we passed the time zone via the formatting context and it applied to the whole message, not to specific fields.

Note that there are no options for "floating" a time value for display (that is, turning an `Instant` or `Zoned*` into a `Plain*`/`Local*` etc.), which is a super-common thing to want to do _at format time_. You could argue that developers should have to convert the value themselves, but its convenient to let time-handling experts do this (i.e. inside the MF implementation) rather than developers getting it wrong (explaining how to work with time values is hard work relatively speaking)

> The system or locale time zone should never be implicitly used

I do not agree. It means that simple messages never display local time when using old-fashioned time value like `java.util.Date`. I want to write the following and use `System.currentTimeMillis()` or `new Date()` if I feel like it:

```
This time is now {$now :time}
```

exploration/default-registry-and-mf1-compatibility.md
aphillips: fixed.

exploration/default-registry-and-mf1-compatibility.md
sffc: Yeah, I think we're aligned but let me clarify. The system time zone can be used when localizing a timestamp (absolute time / instant). Ideally that happens outside of MF because MF shouldn't be in the business of date arithmetic.

aphillips: MF is definitely not in the _anything_ arithmetic business 😄 but our functions can sometimes be.

What I have in mind here is that users be able (but not required) to pass a time zone when they need the underlying formatter to do one of the following:
- attach a floating time to the timeline (e.g. `ZonedDateTime.of(myLocalDateTime, zoneId)`)
- change the time zone of a zoned time value
- set the time zone for displaying an unzoned instant (like `java.util.Date`)

As noted, this could be done in the formatting context, but an application might need to do any of these for a specific expression field:

```
The time {$now :time} is {$now :time timezone=$usersZone} in {$usersZone}.
```

Notice that in all three cases you want the value and the target time zone and the desired operation can be determined through reflection. Re-floating a time value isn't generally necessary in the formatter unless what you want to do is get rid of the time zone name in the display.

>Ideally that happens outside of MF because MF shouldn't be in the business of date arithmetic.

As above: _ideally_ yes. The option is for a non-ideal world.

One downside to the above: the option can take a value and the value can be an argument. But the value can't be null/empty.

sffc: One of the challenges with `{$now :time timezone=$usersZone}` is defining, how it is supposed to behave based on the input type?

- Temporal.Absolute: probably combine the offset with the time zone and format as a ZonedDateTime
- Temporal.ZonedDateTime: a few things one could do in principle
    - Keep the ZonedDateTime time zone
    - Use the MF time zone and project the absolute timestamp into that time zone
    - Use the MF time zone without changing the clock time of the ZonedDateTime
    - Throw an exception
- Temporal.PlainDate: no time of day is available, so throw an exception?
- Temporal.PlainDateTime: a few things are possible
    - Use the PlainDateTime at face value and attach the time zone, formatting as a ZonedDateTime
    - Assume the PlainDateTime is in the system time zone and perform the conversion (yuck, don't do this)
- Temporal.PlainTime: similar to Temporal.PlainDateTime

Due to interoperability concerns, I _do_ think MF2 should specify the data flow. I'm in full alignment that MF2 shouldn't specify output formats, but it _should_ specify the flow of information so that two different implementations produce _semantically equivalent_ output.

exploration/default-registry-and-mf1-compatibility.md
```diff
@@ -133,58 +141,56 @@ For more information, see [Working with Timezones](https://w3c.github.io/timezon
...
-- `timeZone` (default is system default time zone or UTC)
-  - valid identifier per [BCP175](https://www.rfc-editor.org/rfc/rfc6557)
-
-In addition to the above high-level options, a function can use either the appropriate
-_style_ options for that function
+A function can use either the appropriate _style_ options for that function
 or can use a collection of _field options_ (but not both) to control the formatted 
 output.
 
+If both are specified, the _style_ option takes precedence.
```
sffc, line 148: Did this change? I recall earlier that the _field options_ took precedence. In ECMA-402, we throw an exception. Can we do that here?

exploration/default-registry-and-mf1-compatibility.md
```diff
@@ -225,6 +229,41 @@ The functions `:datetime` and `:time` have the following options:
...
+  - `shortOffset`
+  - `longOffset`
+  - `shortGeneric`
+  - `longGeneric`
+
+> [!NOTE]
+> The following options do not have default values because they are only to be used
+> as overrides for locale-and-value dependent implementation-defined defaults.
+
+The function `:time` has following additional options:
```
sffc, line 244:
```suggestion
The function `:time` has following options:
```

exploration/default-registry-and-mf1-compatibility.md
aphillips: We can emit an error (specifically `Invalid Expression`). That would result in the fallback rendering. 

This changed from the point of view that we didn't specify previously. Error would perhaps be preferable.

exploration/default-registry-and-mf1-compatibility.md
aphillips: Good catch. Previously these were additional. Now they aren't.

exploration/default-registry-and-mf1-compatibility.md
eemeli:
```suggestion
If both are specified, an _Invalid Expression_ error MUST be emitted
and a _fallback value_ used as the resolved value of the _expression_.
```

exploration/default-registry-and-mf1-compatibility.md
```diff
@@ -133,58 +141,56 @@ For more information, see [Working with Timezones](https://w3c.github.io/timezon
...
+
+##### Style Options
+
 The function `:datetime` has these function-specific _style_ options.
 - `dateStyle`
   - `full`
   - `long`
   - `medium`
-  - `short`
+  - `short` (default)
```
eemeli, line 157: As discussed previously in https://github.com/unicode-org/message-format-wg/pull/623#discussion_r1480148093, these "default" notes should be dropped.

```suggestion
  - `short`
```

exploration/default-registry-and-mf1-compatibility.md
```diff
@@ -133,58 +141,56 @@ For more information, see [Working with Timezones](https://w3c.github.io/timezon
...
 - `timeStyle`
   - `full`
   - `long`
   - `medium`
-  - `short`
+  - `short` (default)
```
eemeli, line 162:
```suggestion
  - `short`
```

exploration/default-registry-and-mf1-compatibility.md
```diff
@@ -133,58 +141,56 @@ For more information, see [Working with Timezones](https://w3c.github.io/timezon
...
+Only those fields specified in the _annotation_ appear in the formatted output.
+
+> [!NOTE]
+> Field options do not have default values because they are only to be used
+> to compose the formatter.
+> Each value must be specified.
 
 The _field_ options are defined as follows:
 
 All functions have the following option:
```
eemeli, line 191:
```suggestion
```

exploration/default-registry-and-mf1-compatibility.md
```diff
@@ -225,6 +229,41 @@ The functions `:datetime` and `:time` have the following options:
...
+  - `h23`
+  - `h24`
+- `timeZoneName`
+  - `long`
+  - `short`
+  - `shortOffset`
+  - `longOffset`
+  - `shortGeneric`
+  - `longGeneric`
+
```
eemeli, line 257: If we're dropping all but style and leaving the locale-ish options out for now, I think these should not be included on `:time`. My feelings aren't too strong here, though.

```suggestion
```

eemeli (changes requested):
Small stuff, inline.

exploration/default-registry-and-mf1-compatibility.md
```diff
@@ -133,58 +141,56 @@ For more information, see [Working with Timezones](https://w3c.github.io/timezon
...
+##### Field Options
+
+Field options describe which fields to include in the formatted output
+and what format to use for that field.
+Only those fields specified in the _annotation_ appear in the formatted output.
+
+> [!NOTE]
+> Field options do not have default values because they are only to be used
+> to compose the formatter.
+> Each value must be specified.
```
stasm, line 188: I think this is potentially misleading. The previous sentence is probably enough?

```suggestion
```

stasm (approved):

exploration/default-registry-and-mf1-compatibility.md
aphillips: Time zone name I could live without here. Hour cycle feels important though: it's a common override on time values. @sffc's point about having a 12/24 hour toggle (`hour12`) also feels like a feature for `:time` specifically.

exploration/default-registry-and-mf1-compatibility.md
```diff
@@ -164,27 +167,28 @@ The function `:date` has these function-specific _style_ options:
...
-- `timeZoneName`
-  - `long`
-  - `short`
-  - `shortOffset`
-  - `longOffset`
-  - `shortGeneric`
-  - `longGeneric`
+Field options describe which fields to include in the formatted output
+and what format to use for that field.
+Only those fields specified in the _annotation_ appear in the formatted output.
```
sffc, line 183: One of the features of semantic skeleta is preventing nonsensical skeletons like "hours and seconds" without minutes. Mapping from standard skeleta to semantic skeleta may fill in some of those missing fields. Even now, fields such as hour cycle and era may appear even if not in the skeleton. Therefore, it would be better to remove the second sentence giving any guarantees about which fields appear in the formatted output.

eemeli (approved):

exploration/default-registry-and-mf1-compatibility.md
```diff
@@ -109,7 +117,7 @@ For more information, see [Working with Timezones](https://w3c.github.io/timezon
 > The [ABNF](/spec/message.abnf) and [syntax](/spec/syntax.md) of MFv2
```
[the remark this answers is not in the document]
aphillips: `timezoneOffset` is a field in XMLSchema's various specified literal representations, not our spec.

PR #633, opened 2024-02-09
Make arrays in the data model not optional

stasm:
A follow-up to #620, see the discussion there.

PR #651, opened 2024-02-15
Review data model

mihnita:
Addresses issue #639

I will also update the .json and .dtd, but after we get an agreement on this.

### DIFF
--- a/spec/data-model/README.md
+++ b/spec/data-model/README.md
@@ -22,39 +22,43 @@ Two equivalent definitions of the data model are also provided:
 
 ## Messages
 
-A `SelectMessage` corresponds to a syntax message that includes _selectors_.
-A message without _selectors_ and with a single _pattern_ is represented by a `PatternMessage`.
+A `SimpleMessage` corresponds to a syntax message that includes _selectors_.
+A message without _selectors_ and with a single _pattern_ is represented by a `ComplexMessage`.
 
 In the syntax,
-a `PatternMessage` may be represented either as a _simple message_ or as a _complex message_,
+a `ComplexMessage` may be represented either as a _simple message_ or as a _complex message_,
 depending on whether it has declarations and if its `pattern` is allowed in a _simple message_.
 
 ```ts
-type Message = PatternMessage | SelectMessage;
+type Message = SimpleMessage | ComplexMessage;
 
-interface PatternMessage {
-  type: "message";
-  declarations: Declaration[];
+interface SimpleMessage {
+  type: "simple-message";
   pattern: Pattern;
 }
 
-interface SelectMessage {
-  type: "select";
-  declarations: Declaration[];
+interface Matcher {
   selectors: Expression[];
   variants: Variant[];
+}
+
+interface ComplexMessage {
+  type: "complex-message";
+  declarations?: Declaration[];
+  body: Pattern | Matcher;
 }
 ```
 
 Each message _declaration_ is represented by a `Declaration`,
 which connects the `name` of a _variable_
 with its _expression_ `value`.
-The `name` does not include the initial `$` of the _variable_.
+The `name` does not include the `$` sigil of the _variable_,
+or the starting and ending `|` of the literals.
 
 The `name` of an `InputDeclaration` MUST be the same
 as the `name` in the `VariableRef` of its `VariableExpression` `value`.
 
-An `UnsupportedStatement` represents a statement not supported by the implementation.
+An `ReservedStatement` represents a statement not supported by the implementation.
 Its `keyword` is a non-empty string name (i.e. not including the initial `.`).
 If not empty, the `body` is the "raw" value (i.e. escape sequences are not processed)
 starting after the keyword and up to the first _expression_,
@@ -68,7 +72,7 @@ The non-empty `expressions` correspond to the trailing _expressions_ of the _res
 > this data model.
 
 ```ts
-type Declaration = InputDeclaration | LocalDeclaration | UnsupportedStatement;
+type Declaration = InputDeclaration | LocalDeclaration | ReservedStatement;
 
 interface InputDeclaration {
   type: "input";
@@ -82,18 +86,17 @@ interface LocalDeclaration {
   value: Expression;
 }
 
-interface UnsupportedStatement {
-  type: "unsupported-statement";
+interface ReservedStatement {
+  type: "reserved-statement";
   keyword: string;
   body?: string;
   expressions: Expression[];
 }
 ```
 
-In a `SelectMessage`,
-the `keys` and `value` of each _variant_ are represented as an array of `Variant`.
-For the `CatchallKey`, a string `value` may be provided to retain an identifier.
-This is always `'*'` in MessageFormat 2 syntax, but may vary in other formats.
+In a `Matcher` the `keys` and `value` of each _variant_ are represented as an array of `Variant`. \
+For the `CatchallKey`, a `Pattern` `value` MUST be provided.
+A `CatchallKey` is always `'*'` in MessageFormat 2 syntax, but may vary in other formats.
 
 ```ts
 interface Variant {
@@ -103,23 +106,25 @@ interface Variant {
 
 interface CatchallKey {
   type: "*";
-  value?: string;
+  value: Pattern;
 }
 ```
 
 ## Patterns
 
-Each `Pattern` contains a linear sequence of text and placeholders corresponding to potential output of a message.
+Each `Pattern` contains a linear sequence of text and placeholders corresponding
+to potential output of a message.
 
-Each element of the `Pattern` MUST either be a non-empty string, an `Expression`, or a `Markup` object.
-String values represent literal _text_.
+Each element of the `Pattern` MUST either be a non-empty string, an `Expression`,
+or a `Markup` object. \
+String values represent literal _text_. \
 String values include all processing of the underlying _text_ values,
-including escape sequence processing.
-`Expression` wraps each of the potential _expression_ shapes.
+including escape sequence processing. \
+`Expression` wraps each of the potential _expression_ shapes. \
 `Markup` wraps each of the potential _markup_ shapes.
 
-Implementations MUST NOT rely on the set of `Expression` and 
-`Markup` interfaces defined in this document being exhaustive.
+Implementations MUST NOT rely on the set of `Expression` and
+`Markup` interfaces defined in this document being exhaustive. \
 Future versions of this specification might define additional
 expressions or markup.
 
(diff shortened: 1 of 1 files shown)


mihnita:
I don't consider it 100% done, but it is the result of a first (slow and careful) pass.
My plan is to take a break, and take another look.
But I thought it might be good to give some extra time to our colleagues in Europe.

spec/data-model/README.md
```diff
@@ -22,39 +22,43 @@ Two equivalent definitions of the data model are also provided:
 
 ## Messages
 
-A `SelectMessage` corresponds to a syntax message that includes _selectors_.
-A message without _selectors_ and with a single _pattern_ is represented by a `PatternMessage`.
+A `SimpleMessage` corresponds to a syntax message that includes _selectors_.
+A message without _selectors_ and with a single _pattern_ is represented by a `ComplexMessage`.
```
aphillips, line 26: This one is backwards.

```suggestion
A `SimpleMessage` is a _message_ without _selectors_ and with a single _pattern_.
A `ComplexMessage` is any _message_ that includes one or more _declarations_, a _selector_, or both.

**Review by aphillips — COMMENTED** [review]
Thanks for this.

The top part looks like it needs more work. Other parts looks like they might be in better harmony with the syntax. I didn't comment fully.

File: spec/data-model/README.md [hunk 2]
```diff
@@ -22,39 +22,43 @@ Two equivalent definitions of the data model are also provided:
...
 
 In the syntax,
-a `PatternMessage` may be represented either as a _simple message_ or as a _complex message_,
+a `ComplexMessage` may be represented either as a _simple message_ or as a _complex message_,
 depending on whether it has declarations and if its `pattern` is allowed in a _simple message_.
 
 ```ts
-type Message = PatternMessage | SelectMessage;
+type Message = SimpleMessage | ComplexMessage;
```
eemeli, line 33: Why this change in representation? This seems wholly unnecessary, as it links the data model with syntactic details like the representation of a message with a single pattern depending on whether it starts with a `.` or not.

I would have expected a change like this to have at least some argumentation about why it's being made, and some consideration of how it's impacting data model users and operations.

spec/data-model/README.md
```diff
@@ -22,39 +22,43 @@ Two equivalent definitions of the data model are also provided:
...
 with its _expression_ `value`.
-The `name` does not include the initial `$` of the _variable_.
+The `name` does not include the `$` sigil of the _variable_,
+or the starting and ending `|` of the literals.
 
 The `name` of an `InputDeclaration` MUST be the same
 as the `name` in the `VariableRef` of its `VariableExpression` `value`.
 
-An `UnsupportedStatement` represents a statement not supported by the implementation.
+An `ReservedStatement` represents a statement not supported by the implementation.
```
eemeli, line 61: Why this rename? Again, I would expect some argumentation here.

spec/data-model/README.md
```diff
@@ -103,23 +106,25 @@ interface Variant {
 
 interface CatchallKey {
   type: "*";
-  value?: string;
+  value: Pattern;
```
eemeli, line 109: This seems like a misunderstanding of this data model entry.

eemeli (changes requested):
While the changes proposed here bring the data model much closer to the MF2 syntax, they also make it less useful for any other purpose. None of the proposed changes are presented with any discussion about why they are being made, and they are sufficiently extensive that I would expect a design document discussing them, should we wish to consider their incorporation in the spec.

In short, this does not appear to be a review of the data model, but a rewriting of it. I do not think that can be in scope for our consideration this week.

spec/data-model/README.md
```diff
@@ -22,39 +22,43 @@ Two equivalent definitions of the data model are also provided:
...
 ## Messages
 
-A `SelectMessage` corresponds to a syntax message that includes _selectors_.
-A message without _selectors_ and with a single _pattern_ is represented by a `PatternMessage`.
+A `SimpleMessage` corresponds to a syntax message that includes _selectors_.
+A message without _selectors_ and with a single _pattern_ is represented by a `ComplexMessage`.
 
 In the syntax,
-a `PatternMessage` may be represented either as a _simple message_ or as a _complex message_,
+a `ComplexMessage` may be represented either as a _simple message_ or as a _complex message_,
```
catamorphism, line 29: I think it's confusing to use "complex" to mean two different things (in the data model, having only a single pattern, regardless of declarations; in the concrete syntax, having no declarations and no selectors.)

I don't love the terms `PatternMessage` and `SelectMessage`, but I agree with Eemeli that it's probably best to keep it that way.

spec/data-model/README.md
```diff
@@ -22,39 +22,43 @@ Two equivalent definitions of the data model are also provided:
...
-  type: "select";
-  declarations: Declaration[];
+interface Matcher {
   selectors: Expression[];
   variants: Variant[];
 }
+
+interface ComplexMessage {
+  type: "complex-message";
+  declarations?: Declaration[];
```
catamorphism, line 47: This should be `declarations: Declaration[]` per #633

spec/data-model/README.md
```diff
@@ -22,39 +22,43 @@ Two equivalent definitions of the data model are also provided:
...
+  body: Pattern | Matcher;
+}
 ```
 
 Each message _declaration_ is represented by a `Declaration`,
 which connects the `name` of a _variable_
-The `name` does not include the initial `$` of the _variable_.
+The `name` does not include the `$` sigil of the _variable_,
+or the starting and ending `|` of the literals.
```
catamorphism, line 56: I'm confused about this -- since this part is describing what appears on the left-hand side of a declaration, where do literals come in?

spec/data-model/README.md
```diff
@@ -82,18 +86,17 @@ interface LocalDeclaration {
...
   body?: string;
   expressions: Expression[];
 }
 ```
 
-In a `SelectMessage`,
-the `keys` and `value` of each _variant_ are represented as an array of `Variant`.
-For the `CatchallKey`, a string `value` may be provided to retain an identifier.
-This is always `'*'` in MessageFormat 2 syntax, but may vary in other formats.
+In a `Matcher` the `keys` and `value` of each _variant_ are represented as an array of `Variant`. \
```
catamorphism, line 97:
```suggestion
In a `Matcher`,
the `keys` and `value` of each _variant_ are represented as an array of `Variant`. 
```

(Use semantic line breaks; remove stray backslash)

spec/data-model/README.md
catamorphism: Specifically: I think it should be `string` because the intention here is for it to be a different string representing the catchall key in some syntax other than MF2. And it should be optional.

spec/data-model/README.md
```diff
@@ -103,23 +106,25 @@ interface Variant {
...
-  value?: string;
+  value: Pattern;
 }
 ```
 
 ## Patterns
 
-Each `Pattern` contains a linear sequence of text and placeholders corresponding to potential output of a message.
+Each `Pattern` contains a linear sequence of text and placeholders corresponding
+to potential output of a message.
```
catamorphism, line 116:
```suggestion
to the potential output of a message.
```

spec/data-model/README.md
```diff
@@ -103,23 +106,25 @@ interface Variant {
...
 
-Each element of the `Pattern` MUST either be a non-empty string, an `Expression`, or a `Markup` object.
-String values represent literal _text_.
+Each element of the `Pattern` MUST either be a non-empty string, an `Expression`,
+or a `Markup` object. \
```
catamorphism, line 119:
```suggestion
or a `Markup`.
```
Delete "object" for consistency with the rest of the sentence (it doesn't say "an `Expression` object").

And, I don't think the backslashes are necessary? (Here and elsewhere.)

spec/data-model/README.md
```diff
@@ -172,46 +188,58 @@ The `Literal` and `VariableRef` correspond to the the _literal_ and _variable_ s
 When they are used as the `body` of an `Expression`,
 they represent _expression_ values with no _annotation_.
 
-`Literal` represents all literal values, both _quoted_ and _unquoted_.
-The presence or absence of quotes is not preserved by the data model.
-The `value` of `Literal` is the "cooked" value (i.e. escape sequences are processed).
+`Literal` represents all literal values, both numeric (_unquoted_) and string (_quoted_).
+The presence or absence of quotes (`'|'` in MessageFormat 2) is not preserved by the data model.
+The `value` of `StringLiteral` is the "cooked" value (i.e. escape sequences are processed).
+The `value` of `NumberLiteral` is the numeric value after parsing (for example the trailing zeros are not preserved).
```
catamorphism, line 194:
```suggestion
The `value` of `NumberLiteral` is the numeric value after parsing,
(for example the trailing zeros are not preserved).
```

Use semantic line breaks.

More substantively: making this the "parsed" numeric value makes it hard to test the roundtrip property (maybe I missed an issue where this was discussed?)

spec/data-model/README.md
```diff
@@ -172,46 +188,58 @@ The `Literal` and `VariableRef` correspond to the the _literal_ and _variable_ s
...
-_private-use annotation_ not supported by the implementation or a _reserved annotation_.
-The `sigil` corresponds to the starting sigil of the _annotation_.
-The `source` is the "raw" value (i.e. escape sequences are not processed)
+An `PrivateUseAnnotation` represents a _private-use annotation_,
+not supported by the implementation.
+These are annotations that might be supported by some implementations,
+and are guaranteed to not be in conflict with future standard development.
+
+An `ReservedAnnotation` represents a _reserved annotation_.
+not supported by the implementations.
```
catamorphism, line 235:
```suggestion
not supported by the implementation.
```

(Consistency with the previous para)

spec/data-model/README.md
```diff
@@ -172,46 +188,58 @@ The `Literal` and `VariableRef` correspond to the the _literal_ and _variable_ s
...
+These are annotations that might be defined in the future version of the standars.
```
catamorphism, line 236:
```suggestion
These are annotations that might be defined in the future version of the standard.
```

spec/data-model/README.md
```diff
@@ -221,53 +249,49 @@ and does not include the starting `sigil`.
...
-supported by the implementation,
+or _reserved annotation_ supported by the implementation,
 the implementation SHOULD represent it in the data model
 using an interface appropriate for the semantics and meaning
 that the implementation attaches to that _annotation_.
 
+## Identifiers
+
+An `Identifier` is used to identify functions, options, attributes, markup,
+and supporting namespaces to avoid conflict between various proprietary extensions. \
```
catamorphism, line 260:
```suggestion
An `Identifier` is used to identify functions, options, attributes, and markup.
It also includes a namespace, to avoid conflict between various proprietary extensions.
```

catamorphism (changes requested):

aphillips:
Propose we move this post-tech preview.

PR #656, opened 2024-02-15
Reduce strictness of datetime field options

sffc:
Unaddressed feedback from https://github.com/unicode-org/message-format-wg/pull/648#discussion_r1491319250:

> One of the features of semantic skeleta is preventing nonsensical skeletons like "hours and seconds" without minutes. Mapping from standard skeleta to semantic skeleta may fill in some of those missing fields. Even now, fields such as hour cycle and era may appear even if not in the skeleton. Therefore, it would be better to remove the second sentence giving any guarantees about which fields appear in the formatted output.


aphillips (approved):
This change seems fine to me. It's functionally editorial and correct.

eemeli (approved):

PR #657, opened 2024-02-16
Port ABNF changes and fomatting to syntax document

alerque:
The ABNF has recieved a major overhaul with difrenet formatting, and also along the way several small changes and fixes landed. These changes are not reflected in the relevant snippets in the syntax document.

This brings all the snippets up to date with the actual ABNF. One could easily make a case that the formatting doesn't need to match and each snippet could have its own whitespace choices to match the local context, but it was much easier to find changes and reason about them if the lines exactly matched the ABNF file, so I went ahead and updated all the formatting.

In the process the actual changes showed through. Some changes are typo fixes (e.g. `reserve-escape`→`reserved-escape`), others have more exact definitions (e.g. `reserved-keyword` and `reserved-statement`), and a couple items have just been moved around in the source. I moved a couple definitions when they were not relevant to the bit of the document to closer match where they are in the ABNF (e.g.g `quoted-char` and `reserved-char`).

### DIFF
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -127,7 +127,7 @@ A **_<dfn>message</dfn>_** is the complete template for a specific message forma
 A _message_ can be a _simple message_ or it can be a _complex message_.
 
 ```abnf
-message = simple-message / complex-message
+message           = simple-message / complex-message
 ```
 
 A **_<dfn>simple message</dfn>_** contains a single _pattern_,
@@ -135,8 +135,8 @@ with restrictions on its first character.
 An empty string is a valid _simple message_.
 
 ```abnf
-simple-message = [simple-start pattern]
-simple-start = simple-start-char / text-escape / placeholder
+simple-message    = [simple-start pattern]
+simple-start      = simple-start-char / text-escape / placeholder
 ```
 
 A **_<dfn>complex message</dfn>_** is any _message_ that contains _declarations_,
@@ -148,7 +148,7 @@ and consists of:
 2. a _complex body_
 
 ```abnf
-complex-message = *(declaration [s]) complex-body
+complex-message   = *(declaration [s]) complex-body
 ```
 
 ### Declarations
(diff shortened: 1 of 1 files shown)


eemeli (changes requested):
While the actual rules should absolutely be the same, I don't think the whitespace changes make sense.

aphillips (approved):
Thank you for doing this!

The changes look good to me. I'll admit that I didn't actually check them against the ABNF.

stasm (approved):
LGTM, thanks @alerque! I looked at `syntax.md` from your branch and `message.abnf` from `main` side by side, and I think they're equivalent, modulo the ordering of productions.

spec/syntax.md
```diff
@@ -289,6 +289,8 @@ be preserved during formatting.
 ```abnf
+quoted-char       = content-char / s / "." / "@" / "{" / "}"
+reserved-char     = content-char / "."
```
eemeli, line 293: These are already listed elsewhere in `syntax.md`. Is there a reason to bring them here, rather than keeping them in their current places?

aphillips: I'm not seeing this? Where are the productions doubled in @alerque's [text](https://github.com/alerque/message-format-wg/blob/update-abnf-snippets-in-spec/spec/syntax.md)?

eemeli: The `quoted-char` is currently under Literals right after `quoted`, and `reserved-char` under Reserved Annotations right after `reserved-body`. The proposed change moves their listings from those places (which are their only users) to here.

As is, their meaning is obvious from context. If they're moved here, then we should add some text here explaining what they are.

alerque: Those definitions moved in the abnf source too, and I found their new/current locations much easier to grok than the old ones the syntax explanations were using because they are now grouped with more related context with other similar definitions.

If consensuses is that isn't a good change they can go back to their original places, but the abnf snippets will no longer be all blocks where the whole snippet comes from a contiguous bit of source. Having all the snippets match some contiguous bit from the original file makes it quite a bit easier to verify that it is in sync.

eemeli: As we do not intend to make further syntax changes, mirroring such changes in `syntax.md` as well as the ANBF should not need to be done later.

I have a weak preference for keeping these as they currently are, but I'm fine to go with whatever most people think.

aphillips:
I'm going to merge this. We can tweak the duplicates in the spec tomorrow, if we feel strongly enough about it.

PR #658, opened 2024-02-16
Rename simple-start to simple-begin

stasm:
Our ABNF uses the following convention (or at last seems to be using it):

* names of productions which match single characters end with `-char`,
* names of a subset of such productions which define the first characters of longer tokens end with `-start`.

Hence, for example: `name-char` and `name-start`.

There is however a small inconsistency in how we use this convention: `simple-start` is not a terminal single-char production but instead is defined as:

```abnf
simple-start      = simple-start-char / text-escape / placeholder
```

`simple-start-char` is the only production in which we use `-start-char` in the name.

I'd like to fix this. I see two ways:

1. Rename all single-char `*-start` productions to `*-start-char`. This is more explicit and also follows [XML's precedent](https://www.w3.org/TR/REC-xml/#NT-Name), but produces longer production names, and requires changes to the ABNF in a few more places than option 2.
2. Rename `simple-start` to `simple-begin`, and `simple-start-char` to `simple-start`.

This PR proposes the second solution. If approved, I'll rebase on top of #657 and make corresponding changes to the rest of the spec.

### DIFF
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -1,7 +1,7 @@
 message           = simple-message / complex-message
 
-simple-message    = [simple-start pattern]
-simple-start      = simple-start-char / text-escape / placeholder
+simple-message    = [simple-begin pattern]
+simple-begin      = simple-start / text-escape / placeholder
 pattern           = *(text-char / text-escape / placeholder)
 placeholder       = expression / markup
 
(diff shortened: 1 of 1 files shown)


stasm:
cc @alerque.

aphillips:
These are good thoughts, but...

I think we froze the ABNF (pace actual errors) in yesterday's call. I don't see anyone having difficulty understanding the syntax as it sits now (I think it is pretty obvious what is going on), so would suggest we skip this? 

What do others think?

eemeli:
To me this sounds like a semantics argument about a thing that only the people reading the ABNF might ever notice. I think the rules are currently clear in context, and all of the `-start` rules start something. Introducing `-begin` adds a new concept, which feel heavy.

I feel rather "meh" about this, and don't have a really strong opinion here.

stasm:
It's confusing to me that `simple-start` is not a single-char terminal. Do other people, and in particular early implementers, also feel that way?

I realize that this is coming a few hours too late, but then again, we haven't yet balloted, and also I don't plan any other cleanups like this one.

aphillips:
I'm going to put this out of scope. @stasm Open an issue if you want to ask the group to consider it.

PR #659, opened 2024-02-16
Implement the default registry in the spec

aphillips:
This PR copies the number and date/time default functionality from design to the specification.

The copy was done verbatim, except:

- I moved some parts of number to make the structure logically consistent
- I added an example to the `style=percent` note
- I turned the information about the default selector into an "important" note
- I added the surrounding document structure

### DIFF
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -310,36 +310,17 @@ All other values produce a _Resolution Error_.
 
 ### Options
 
+Some options do not have default values defined in this specification.
+The defaults for these options are implementation-dependent.
+In general, the default values for such options depend on the locale, 
+the value of other options, or both.
+
 The following options and their values are required in the default registry to be available on the 
 function `:number`:
 - `select`
-   -  `plural` (default)
+   -  `plural` (default; see [Default Value of `select` Option](#default-value-of-select-option) below)
    -  `ordinal`
    -  `exact`
-
- 
-> [!IMPORTANT]
-> **Default Value of `select` Option**
->
-> The value `plural` is default for the option `select` 
-> because it is the most common use case for numeric selection.
-> It can be used for exact value matches but also allows for the grammatical needs of 
-> languages using CLDR's plural rules.
-> This might not be noticeable in the source language (particularly English), 
-> but can cause problems in target locales that the original developer is not considering.
->
-> For example, a naive developer might use a special message for the value `1` without
-> considering a locale's need for a `one` plural:
->
->```
-> .match {$var}
-> 1   {{You have one last chance}}
-> one {{You have {$var} chance remaining}} // needed by languages such as Polish or Russian
->                                          // such locales typically require other keywords
->                                          // such as two, few, many, and so forth
-> *   {{You have {$var} chances remaining}}
->```
-
 - `compactDisplay` (this option only has meaning when combined with the option `notation=compact`)
    - `short` (default)
    - `long`
@@ -367,11 +348,6 @@ function `:number`:
   - `min2`
 - `minimumIntegerDigits`
   - (non-negative integer, default: `1`)
-> [!NOTE]
-> The following options do not have default values because they are only to be used
-> as overrides for an existing locale-and-value dependent implementation-defined
-> default
-
 - `minimumFractionDigits`
   - (non-negative integer)
 - `maximumFractionDigits`
@@ -401,18 +377,10 @@ function `:integer`:
   - `percent` (see [Percent Style](#percent-style) below)
 - `useGrouping`
   - `auto` (default)
-  - `true`
-  - `false`
-  - `min2`
   - `always`
+  - `min2`
 - `minimumIntegerDigits`
   - (non-negative integer, default: `1`)
-
-> [!NOTE]
-> The following option does not have a default value because it is only to be used
-> as an override for an existing locale-and-value dependent implementation-defined
-> default
-
 - `maximumSignificantDigits`
   - (non-negative integer)
 
@@ -446,6 +414,26 @@ are encouraged to track development of these options during Tech Preview:
    - `long`
    - `short` (default)
    - `narrow`
+
+#### Default Value of `select` Option
+
+The value `plural` is default for the option `select` 
+because it is the most common use case for numeric selection.
+It can be used for exact value matches but also allows for the grammatical needs of 
+languages using CLDR's plural rules.
+This might not be noticeable in the source language (particularly English), 
+but can cause problems in target locales that the original developer is not considering.
+
+For example, a naive developer might use a special message for the value `1` without
+considering a locale's need for a `one` plural:
+```
+.match {$var}
+1   {{You have one last chance}}
+one {{You have {$var} chance remaining}} // needed by languages such as Polish or Russian
+                                         // such locales typically require other keywords
+                                         // such as two, few, many, and so forth
+*   {{You have {$var} chances remaining}}
+```
 
 #### Percent Style
 When implementing `style=percent`, the numeric value of the _operand_ 
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -344,7 +344,7 @@ and formatting numeric values as integers.
 
 The _operand_ of a number function is either an implementation-defined type or
 a literal whose contents match the `number-literal` production in the [ABNF](/spec/message.abnf).
-All other values produce an _Invalid Expression Error_.
+All other values produce an _Invalid Expression_ error.
 
 > For example, in Java, any subclass of `java.lang.Number` plus the primitive
 > types (`byte`, `short`, `int`, `long`, `float`, `double`, etc.) 
@@ -645,7 +645,7 @@ If no options are specified, each of the functions defaults to the following:
 The _operand_ of a date/time function is either 
 an implementation-defined date/time type (passed in as an argument)
 or a _date/time literal value_, as defined below.
-All other _operand_ values produce a _Selection Error_.
+All other _operand_ values produce an _Invalid Expression_ error.
 
 A **_<dfn>date/time literal value</dfn>_** is a non-empty string consisting of 
 one of the following:
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -609,8 +609,8 @@ the two strings are equal.
 
 > [!NOTE]
 > Implementations are not expected to implement this exactly as written,
-as there are clearly optimizations that can be applied.
-However, the observed behavior must be consistent with what is described here.
+> as there are clearly optimizations that can be applied.
+> However, the observed behavior must be consistent with what is described here.
 
 > [!NOTE]
 > Only integer matching is required in the Technical Preview.
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -255,3 +255,502 @@ which only expects the `accord` option:
 > .input {$object :noun case=nominative}
 > {{You see {$color :adjective article=indefinite accord=$object} {$object}!}}
 >```
+
+# Default Registry for LDML45
+
+This section describes the functions which each implementation MUST provide to be conformant with
+the LDML45 Technical Preview of MessageFormat v2.
+
+## String Value Selection and Formatting
+
+(Pending)
+
+## Numeric Value Selection and Formatting
+
+### Functions
+
+The following functions use numeric selection:
+
+The function `:number` is the default selector for numeric values.
+
+The function `:integer` provides a reduced set of options for selecting
+and formatting numeric values as integers.
+
+### Operands
+
+The _operand_ of a number function is either an implementation-defined type or
+a literal that matches the `number-literal` production in the [ABNF](/main/spec/message.abnf).
+All other values produce a _Selection Error_ when evaluated for selection
+or a _Formatting Error_ when attempting to format the value.
+
+> For example, in Java, any subclass of `java.lang.Number` plus the primitive
+> types (`byte`, `short`, `int`, `long`, `float`, `double`, etc.) 
+> might be considered as the "implementation-defined numeric types".
+> Implementations in other programming languages would define different types
+> or classes according to their local needs.
+
+> [!NOTE]
+> String values passed as variables in the _formatting context_'s
+> _input mapping_ can be formatted as numeric values as long as their
+> contents match the `number-literal` production in the [ABNF](/main/spec/message.abnf).
+>
+> For example, if the value of the variable `num` were the string
+> `-1234.567`, it would behave identically to the local
+> variable in this example:
+> ```
+> .local $example = {|-1234.567| :number}
+> {{{$num :number} == {$example}}}
+> ```
+
+> [!NOTE]
+> Implementations are encouraged to provide support for compound types or data structures
+> that provide additional semantic meaning to the formatting of number-like values.
+> For example, in ICU4J, the type `com.ibm.icu.util.Measure` can be used to communicate
+> a value that include a unit
+> or the type `com.ibm.icu.util.CurrencyAmount` can be used to set the currency and related
+> options (such as the number of fraction digits).
+
+
+### Options
+
+The following options and their values are required in the default registry to be available on the 
+function `:number`:
+- `select`
+   -  `plural` (default)
+   -  `ordinal`
+   -  `exact`
+
+ 
+> [!IMPORTANT]
+> **Default Value of `select` Option**
+>
+> The value `plural` is default for the option `select` 
+> because it is the most common use case for numeric selection.
+> It can be used for exact value matches but also allows for the grammatical needs of other 
+> languages using CLDR's plural rules.
+> This might not be noticeable in the source language (particularly English), 
+> but can cause problems in target locales that the original developer is not considering.
+>
+> > For example, a naive developer might use a special message for the value `1` without
+> > considering other locale's need for a `one` plural:
+> >
+> >```
+> > .match {$var}
+> > 1   {{You have one last chance}}
+> > one {{You have {$var} chance remaining}} // needed by languages such as Polish or Russian
+> >                                          // such locales typically require other keywords
+> >                                          // such as two, few, many, and so forth
+> > *   {{You have {$var} chances remaining}}
+> >```
+
+- `compactDisplay` // this option only has meaning when combined with the option `notation=compact`
+   - `short` (default)
+   - `long`
+- `notation`
+   - `standard` (default)
+   - `scientific`
+   - `engineering`
+   - `compact`
+- `numberingSystem`
+   - valid [Unicode Number System Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeNumberSystemIdentifier)
+     (default is locale-specific)
+- `signDisplay`
+   -  `auto` (default)
+   -  `always`
+   -  `exceptZero`
+   -  `negative`
+   -  `never`
+- `style`
+  - `decimal` (default)
+  - `percent` (see [Percent Style](#percent-style) below)
+- `useGrouping`
+  - `auto` (default)
+  - `always`
+  - `never`
+  - `min2`
+- `minimumIntegerDigits`
+  - (non-negative integer, default: `1`)
+  - 
+> [!NOTE]
+> The following options do not have default values because they are only to be used
+> as overrides for an existing locale-and-value dependent implementation-defined
+> default
+
+- `minimumFractionDigits`
+  - (non-negative integer)
+- `maximumFractionDigits`
+  - (non-negative integer)
+- `minimumSignificantDigits`
+  - (non-negative integer)
+- `maximumSignificantDigits`
+  - (non-negative integer)
+
+The following options and their values are required in the default registry to be available on the 
+function `:integer`:
+- `select`
+   -  `plural` (default)
+   -  `ordinal`
+   -  `exact`
+- `numberingSystem`
+   - valid [Unicode Number System Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeNumberSystemIdentifier)
+     (default is locale-specific)
+- `signDisplay`
+   -  `auto` (default)
+   -  `always`
+   -  `exceptZero`
+   -  `negative`
+   -  `never`
+- `style`
+  - `decimal` (default)
+  - `percent` (see [Percent Style](#percent-style) below)
+- `useGrouping`
+  - `auto` (default)
+  - `true`
+  - `false`
+  - `min2`
+  - `always`
+- `minimumIntegerDigits`
+  - (non-negative integer, default: `1`)
+
+> [!NOTE]
+> The following option does not have a default value because it is only to be used
+> as an override for an existing locale-and-value dependent implementation-defined
+> default
+
+- `maximumSignificantDigits`
+  - (non-negative integer)
+
+> [!NOTE]
+> The following options or option values are being developed during the Technical Preview
+> period.
+
+The following values for the option `style` are _not_ part of the default registry.
+Implementations SHOULD avoid creating options that conflict with these, but
+are encouraged to track development of these options during Tech Preview:
+- `currency`
+- `unit`
+
+The following options are _not_ part of the default registry.
+Implementations SHOULD avoid creating options that conflict with these, but
+are encouraged to track development of these options during Tech Preview:
+- `currency`
+   - valid [Unicode Currency Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeCurrencyIdentifier)
+     (no default)
+- `currencyDisplay`
+   - `symbol` (default)
+   - `narrowSymbol`
+   - `code`
+   - `name`
+- `currencySign`
+  - `accounting`
+  - `standard` (default)
+- `unit`
+   - (anything not empty)
+- `unitDisplay`
+   - `long`
+   - `short` (default)
+   - `narrow`
+
+### Percent Style
+When implementing `style=percent`, the numeric value of the _operand_ 
+MUST be divided by 100 for the purposes of formatting.
+
+> For example,
+> `The total was {|100| :number style=percent}.` should format
+> in a manner similar to:
+> > The total was 1%.
+
+### Selection
+
+Number selection has three modes:
+- `exact` selection matches the operand to explicit numeric keys exactly
+- `plural` selection matches the operand to explicit numeric keys exactly
+  or to plural rule categories if there is no explicit match
+- `ordinal` selection matches the operand to explicit numeric keys exactly
+  or to ordinal rule categories if there is no explicit match
+
+When implementing [`MatchSelectorKeys`](spec/formatting.md#resolve-preferences), 
+numeric selectors perform as described below.
+
+- Let `return_value` be a new empty list of strings.
+- Let `operand` be the resolved value of the _operand_.
+  If the `operand` is not a number type, emit a _Selection Error_
+  and return `return_value`.
+- Let `keys` be a list of strings containing keys to match.
+  (Hint: this list is an argument to `MatchSelectorKeys`)
+- For each string `key` in `keys`:
+   - If the value of `key` matches the production `number-literal`:
+     - If the parsed value of `key` is an [exact match](#determining-exact-literal-match)
+       of the value of the `operand`, then `key` matches the selector.
+       Add `key` to the front of the `return_value` list.
+   - Else, if the value of `key` is a keyword:
+      - Let `keyword` be a string which is the result of [rule selection](#rule-selection).
+      - If `keyword` equals `key`, then `key` matches the selector.
+        Append `key` to the end of the `return_value` list.
+   - Else, `key` is invalid;
+     emit a _Selection Error_.
+     Do not add `key` to `return_value`
+- Return `return_value`
+
+### Plural/Ordinal Keywords
+The _plural/ordinal keywords_ are: `zero`, `one`, `two`, `few`, `many`, and
+`other`.
+
+### Rule Selection
+
+If the option `select` is set to `exact`, rule-based selection is not used.
+Return the empty string.
+
+> [!NOTE]
+> Since keys cannot be the empty string in a numeric selector, returning the
+> empty string disables keyword selection
+
+If the option `select` is set to `plural`, selection should be based on CLDR plural rule data
+of type `cardinal`. See [charts](https://www.unicode.org/cldr/charts/latest/supplemental/language_plural_rules.html)
+for examples.
+
+If the option `select` is set to `ordinal`, selection should be based on CLDR plural rule data
+of type `ordinal`. See [charts](https://www.unicode.org/cldr/charts/latest/supplemental/language_plural_rules.html)
+for examples.
+
+Apply the rules defined by CLDR to the resolved value of the operand and the function options,
+and return the resulting keyword.
+If no rules match, return `other`.
+
+> **Example.**
+> In CLDR 44, the Czech (`cs`) plural rule set can be found
+> [here](https://www.unicode.org/cldr/charts/44/supplemental/language_plural_rules.html#cs).
+>
+> A message in Czech might be:
+> ```
+> .match {$numDays :number}
+> one  {{{$numDays} den}}
+> few  {{{$numDays} dny}}
+> many {{{$numDays} dne}}
+> *    {{{$numDays} dní}}
+> ```
+> Using the rules found above, the results of various `operand` values might look like:
+> | Operand value | Keyword | Formatted Message |
+> |---|---|---|
+> | 1 | `one` | 1 den |
+> | 2 | `few` | 2 dny |
+> | 5 | `other` | 5 dní |
+> | 22 | `few` | 22 dny |
+> | 27 | `other` | 27 dní |
+> | 2.4 | `many` | 2,4 dne |
+
+### Determining Exact Literal Match
+
+> [!IMPORTANT]
+> The exact behavior of exact literal match is only defined for non-zero-filled
+> integer values.
+> Annotations that use fraction digits or significant digits might work in specific
+> implementation-defined ways.
+> Users should avoid depending on these types of keys in message selection.
+
+
+Number literals in the MessageFormat 2 syntax use the 
+[format defined for a JSON number](https://www.rfc-editor.org/rfc/rfc8259#section-6).
+The resolved value of an `operand` exactly matches a numeric literal `key`
+if, when the `operand` is serialized using the format for a JSON number
+the two strings are equal.
+
+> [!NOTE]
+> Implementations are not expected to implement this exactly as written,
+> as there are clearly optimizations that can be applied.
+
+> [!NOTE]
+> Only integer matching is required in the Technical Preview.
+> Feedback describing use cases for fractional and significant digits-based
+> selection would be helpful.
+> Otherwise, users should avoid using matching with fractional numbers or significant digits.
+
+## Date and Time Value Formatting
+
+This subsection describes the functions and options for date/time formatting.
+Selection based on date and time values is not required in this release.
+
+#### Functions
+
+Functions for formatting [date/time values](#operands) in the default registry are:
+
+- `:datetime`
+- `:date`
+- `:time`
+
+If no options are specified, each of the functions defaults to the following:
+- `{$d :datetime}` is the same as `{$d :datetime dateStyle=short timeStyle=short}`
+- `{$d :date}` is the same as `{$d :date style=short}`
+- `{$t :time}` is the same as `{$t :time style=short}`
+
+> [!NOTE]
+> Pattern selection based on date/time values is a complex topic and no support for this
+> is required in this release.
+
+> [!NOTE]
+> The default formatting behavior of `:datetime` is inconsistent with `Intl.DateTimeFormat`
+> in JavaScript and with `{d,date}` in MessageFormat v1.
+> This is because, unlike those implementations, `:datetime` is distinct from `:date` and `:time`.
+
+#### Operands
+
+The _operand_ of a date/time function is either 
+an implementation-defined date/time type (passed in as an argument)
+or a _date/time literal value_, as defined below.
+All other _operand_ values produce a _Selection Error_ when evaluated for selection
+or a _Formatting Error_ when formatting the value.
+
+A **_<dfn>date/time literal value</dfn>_** is a non-empty string consisting of 
+one of the following:
+- an XMLSchema 1.1 [dateTime](https://www.w3.org/TR/xmlschema11-2/#dateTime)
+- an XMLSchema 1.1 [time](https://www.w3.org/TR/xmlschema11-2/#time)
+- an XMLSchema 1.1 [date](https://www.w3.org/TR/xmlschema11-2/#date)
+
+The `timezoneOffset` of each of these formats is optional. 
+When the offset is not present, implementations should use a floating time type
+(such as Java's `java.time.LocalDateTime`) to represent the time value.
+For more information, see [Working with Timezones](https://w3c.github.io/timezone).
+
+> [!IMPORTANT]
+> The [ABNF](/spec/message.abnf) and [syntax](/spec/syntax.md) of MFv2
+> do not formally define date/time literals. 
+> This means that a _message_ can be syntactically valid but produce
+> an _Operand Mismatch Error_ at runtime.
+
+> [!NOTE]
+> String values passed as variables in the _formatting context_'s
+> _input mapping_ can be formatted as date/time values as long as their
+> contents are date/time literals.
+>
+> For example, if the value of the variable `now` were the string
+> `2024-02-06T16:40:00Z`, it would behave identically to the local
+> variable in this example:
+> ```
+> .local $example = {|2024-02-06T16:40:00Z| :datetime}
+> {{{$now :datetime} == {$example}}}
+> ```
+
+> [!NOTE]
+> True time zone support in serializations is expected to coincide with the adoption
+> of Temporal in JavaScript.
+> The form of these serializations is known and is a de facto standard.
+> Support for these extensions is expected to be required in the post-tech preview.
+> See: https://datatracker.ietf.org/doc/draft-ietf-sedate-datetime-extended/
+
+#### Options
+
+A function can use either the appropriate _style_ options for that function
+or can use a collection of _field options_ (but not both) to control the formatted 
+output.
+
+If both are specified, an _Invalid Expression_ error MUST be emitted
+and a _fallback value_ used as the resolved value of the _expression_.
+
+##### Style Options
+
+The function `:datetime` has these function-specific _style_ options.
+- `dateStyle`
+  - `full`
+  - `long`
+  - `medium`
+  - `short`
+- `timeStyle`
+  - `full`
+  - `long`
+  - `medium`
+  - `short`
+
+The function `:date` has these function-specific _style_ options:
+- `style`
+  - `full`
+  - `long`
+  - `medium`
+  - `short` (default)
+
+The function `:time` has these function-specific _style_ options:
+- `style`
+  - `full`
+  - `long`
+  - `medium`
+  - `short` (default)
+
+##### Field Options
+
+Field options describe which fields to include in the formatted output
+and what format to use for that field.
+The implementation may use this _annotation_ to configure which fields
+appear in the formatted output.
+
+> [!NOTE]
+> Field options do not have default values because they are only to be used
+> to compose the formatter.
+
+The _field_ options are defined as follows:
+
+The function `:datetime` has the following options:
+- `weekday`
+  - `long`
+  - `short`
+  - `narrow`
+- `era`
+  - `long`
+  - `short`
+  - `narrow`
+- `year`
+  - `numeric`
+  - `2-digit`
+- `month`
+  - `numeric`
+  - `2-digit`
+  - `long`
+  - `short`
+  - `narrow`
+- `day`
+  - `numeric`
+  - `2-digit`
+- `hour`
+  - `numeric`
+  - `2-digit`
+- `minute`
+  - `numeric`
+  - `2-digit`
+- `second`
+  - `numeric`
+  - `2-digit`
+- `fractionalSecondDigits`
+  - `1`
+  - `2`
+  - `3`
+- `hourCycle` (default is locale-specific)
+  - `h11`
+  - `h12`
+  - `h23`
+  - `h24`
+- `timeZoneName`
+  - `long`
+  - `short`
+  - `shortOffset`
+  - `longOffset`
+  - `shortGeneric`
+  - `longGeneric`
+
+> [!NOTE]
+> The following options do not have default values because they are only to be used
+> as overrides for locale-and-value dependent implementation-defined defaults.
+
+The followind date/time options are *not* part of the default registry.
+Implementations SHOULD avoid creating options that conflict with these, but
+are encouraged to track development of these options during Tech Preview:
+- `calendar` (default is locale-specific)
+  - valid [Unicode Calendar Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeCalendarIdentifier)
+- `numberingSystem` (default is locale-specific)
+   - valid [Unicode Number System Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeNumberSystemIdentifier)
+- `timeZone` (default is system default time zone or UTC)
+  - valid identifier per [BCP175](https://www.rfc-editor.org/rfc/rfc6557)
+
+#### Selection
+
+Selection based on date/time types is not required by MFv2.
+Implementations should use care when defining selectors based on date/time types.
+The types of queries found in implementations such as `java.time.TemporalAccessor`
+are complex and user expectations may be inconsistent with good I18N practices.
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -588,7 +588,7 @@ If no options are specified, each of the functions defaults to the following:
 > in JavaScript and with `{d,date}` in ICU MessageFormat 1.0.
 > This is because, unlike those implementations, `:datetime` is distinct from `:date` and `:time`.
 
-#### Operands
+### Operands
 
 The _operand_ of a date/time function is either 
 an implementation-defined date/time type (passed in as an argument)
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -366,7 +366,7 @@ All other values produce an _Invalid Expression_ error.
 > [!NOTE]
 > String values passed as variables in the _formatting context_'s
 > _input mapping_ can be formatted as numeric values as long as their
-> contents match the `number-literal` production in the [ABNF](/main/spec/message.abnf).
+> contents match the `number-literal` production in the [ABNF](/spec/message.abnf).
 >
 > For example, if the value of the variable `num` were the string
 > `-1234.567`, it would behave identically to the local
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -326,6 +326,10 @@ the result of converting `operand` to a string literal.
 > in each string.
 > As a result, variations in how text can be encoded can affect the performance of matching.
 > The function `:string` does not perform case folding or Unicode Normalization of string values.
+> Users SHOULD encode _messages_ and their parts (such as _keys_ and _operands_),
+> in Unicode Normalization Form C (NFC) unless there is a very good reason
+> not to.
+> See also: [String Matching](https://www.w3.org/TR/charmod-norm)
 
 > [!NOTE]
 > Unquoted string literals in a _variant_ do not include spaces.
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -303,26 +303,20 @@ The function `:string` has no options.
 
 ### Selection
 
-When implementing [`MatchSelectorKeys(rv, keys)`](spec/formatting.md#resolve-preferences), 
+When implementing [`MatchSelectorKeys(resolvedSelector, keys)`](/spec/formatting.md#resolve-preferences)
+where `resolvedSelector` is the resolved value of a _selector_ _expression_
+and `keys` is a list of strings,
 the `:string` selector performs as described below.
 
-- Let `return_value` be a new empty list of strings.
-- Let `operand` be _rv_.
-If `operand` is not a string literal, let `operand` be
-the result of converting `operand` to a string literal.
-  or, optionally: emit a _Selection Error_ and return `return_value`.
-- Let `keys` be a list of strings containing keys to match.
-  (Hint: this list is an argument to `MatchSelectorKeys`)
-- For each string `key` in `keys`:
-   - If the value of `key` is equal to the string value of `operand`
-     then `key` matches the selector.
-     A `key` and an `operand` are equal if they consist of the same
-     sequence of Unicode code points.
-     Add `key` to the front of the `return_value` list.
-- Return `return_value`.
+1. Let `compare` be the string value of `resolvedSelector`.
+1. Let `result` be a new empty list of strings.
+1. For each string `key` in `keys`:
+   1. If `key` and `compare` consist of the same sequence of Unicode code points, then
+      1. Append `key` as the last element of the list `result`.
+1. Return `result`.
 
 > [!NOTE]
-> Matching of `key` and `operand` values is sensitive to the sequence of code points
+> Matching of `key` and `compare` values is sensitive to the sequence of code points
 > in each string.
 > As a result, variations in how text can be encoded can affect the performance of matching.
 > The function `:string` does not perform case folding or Unicode Normalization of string values.
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -555,7 +555,7 @@ Return the empty string.
 
 > [!NOTE]
 > Since keys cannot be the empty string in a numeric selector, returning the
-> empty string disables keyword selection
+> empty string disables keyword selection.
 
 If the option `select` is set to `plural`, selection should be based on CLDR plural rule data
 of type `cardinal`. See [charts](https://www.unicode.org/cldr/charts/latest/supplemental/language_plural_rules.html)
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -278,8 +278,7 @@ and formatting numeric values as integers.
 
 The _operand_ of a number function is either an implementation-defined type or
 a literal that matches the `number-literal` production in the [ABNF](/spec/message.abnf).
-All other values produce a _Selection Error_ when evaluated for selection
-or a _Formatting Error_ when attempting to format the value.
+All other values produce a _Resolution Error_.
 
 > For example, in Java, any subclass of `java.lang.Number` plus the primitive
 > types (`byte`, `short`, `int`, `long`, `float`, `double`, etc.) 
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -258,8 +258,8 @@ which only expects the `accord` option:
 
 # Default Registry
 
-This section describes the functions which each implementation MUST provide to be conformant with
-this specification.
+This section describes the functions which each implementation MUST provide
+to be conformant with this specification.
 
 ## String Value Selection and Formatting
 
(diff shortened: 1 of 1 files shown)


spec/registry.md
```diff
@@ -255,3 +255,502 @@ which only expects the `accord` option:
 > .input {$object :noun case=nominative}
 > {{You see {$color :adjective article=indefinite accord=$object} {$object}!}}
 >```
+
+# Default Registry for LDML45
```
eemeli, line 259:
```suggestion
# Default Registry
```

spec/registry.md
```diff
@@ -255,3 +255,502 @@ which only expects the `accord` option:
...
+
+This section describes the functions which each implementation MUST provide to be conformant with
+the LDML45 Technical Preview of MessageFormat v2.
```
eemeli, line 262:
```suggestion
this specification.
```

spec/registry.md
```diff
@@ -255,3 +255,502 @@ which only expects the `accord` option:
...
+
+(Pending)
+
+## Numeric Value Selection and Formatting
+
+### Functions
+
+The following functions use numeric selection:
+
+The function `:number` is the default selector for numeric values.
```
eemeli, line 274:
```suggestion
The function `:number` is the default selector and formatter for numeric values.
```

spec/registry.md
```diff
@@ -255,3 +255,502 @@ which only expects the `accord` option:
...
+
+The function `:integer` provides a reduced set of options for selecting
+and formatting numeric values as integers.
+
+### Operands
+
+The _operand_ of a number function is either an implementation-defined type or
+a literal that matches the `number-literal` production in the [ABNF](/main/spec/message.abnf).
```
eemeli, line 282: As before, including `/main` in these links is invalid. So the same applies to all of them:
```suggestion
a literal that matches the `number-literal` production in the [ABNF](/spec/message.abnf).
```

spec/registry.md
```diff
@@ -255,3 +255,502 @@ which only expects the `accord` option:
...
+All other values produce a _Selection Error_ when evaluated for selection
+or a _Formatting Error_ when attempting to format the value.
```
eemeli, line 284: It's probably fine to not mention that these values will first emit a _Resolution Error_, and then because the whole expression resolves to a fallback value, we emit the above-mentioned further errors. Or should some of that be mentioned here?

spec/registry.md
```diff
@@ -255,3 +255,502 @@ which only expects the `accord` option:
...
+> > considering other locale's need for a `one` plural:
+> >
+> >```
+> > .match {$var}
+> > 1   {{You have one last chance}}
+> > one {{You have {$var} chance remaining}} // needed by languages such as Polish or Russian
+> >                                          // such locales typically require other keywords
+> >                                          // such as two, few, many, and so forth
+> > *   {{You have {$var} chances remaining}}
+> >```
```
eemeli, line 344: This shouldn't need a second indentation level.

```suggestion
> For example, a naive developer might use a special message for the value `1` without
> considering other locale's need for a `one` plural:
>
>```
> .match {$var}
> 1   {{You have one last chance}}
> one {{You have {$var} chance remaining}} // needed by languages such as Polish or Russian
>                                          // such locales typically require other keywords
>                                          // such as two, few, many, and so forth
> *   {{You have {$var} chances remaining}}
>```
```

spec/registry.md
```diff
@@ -255,3 +255,502 @@ which only expects the `accord` option:
...
+
+- `compactDisplay` // this option only has meaning when combined with the option `notation=compact`
```
eemeli, line 346:
```suggestion
- `compactDisplay` (this option only has meaning when combined with the option `notation=compact`)
```

spec/registry.md
```diff
@@ -255,3 +255,502 @@ which only expects the `accord` option:
...
+  - `decimal` (default)
+  - `percent` (see [Percent Style](#percent-style) below)
+- `useGrouping`
+  - `auto` (default)
+  - `always`
+  - `never`
+  - `min2`
+- `minimumIntegerDigits`
+  - (non-negative integer, default: `1`)
+  -
```
eemeli, line 373:
```suggestion

```

spec/registry.md
```diff
@@ -255,3 +255,502 @@ which only expects the `accord` option:
...
+> [!NOTE]
+> The following option does not have a default value because it is only to be used
+> as an override for an existing locale-and-value dependent implementation-defined
+> default
+
+- `maximumSignificantDigits`
+  - (non-negative integer)
+
+> [!NOTE]
+> The following options or option values are being developed during the Technical Preview
```
eemeli, line 424:
```suggestion
> The following options and option values are being developed during the Technical Preview
```

spec/registry.md
```diff
@@ -255,3 +255,502 @@ which only expects the `accord` option:
...
+   - `narrow`
+
+### Percent Style
+When implementing `style=percent`, the numeric value of the _operand_ 
+MUST be divided by 100 for the purposes of formatting.
+
+> For example,
+> `The total was {|100| :number style=percent}.` should format
+> in a manner similar to:
+> > The total was 1%.
```
eemeli, line 461: We got this wrong when discussing it in the meeting, the numbers should be _multiplied_ for formatting so e.g. `0.5` is formatted as "50%".

Also, all the other examples use code blocks, rather than being inline.

```suggestion
When implementing `style=percent`, the numeric value of the _operand_ 
MUST be multiplied by 100 for the purposes of formatting.

> For example,
> ```
> The total was {0.5 :number style=percent}.
> ```
> should format in a manner similar to:
> > The total was 50%.
```

spec/registry.md
```diff
@@ -255,3 +255,502 @@ which only expects the `accord` option:
...
+### Selection
+
+Number selection has three modes:
+- `exact` selection matches the operand to explicit numeric keys exactly
+- `plural` selection matches the operand to explicit numeric keys exactly
+  or to plural rule categories if there is no explicit match
+- `ordinal` selection matches the operand to explicit numeric keys exactly
+  or to ordinal rule categories if there is no explicit match
+
+When implementing [`MatchSelectorKeys`](spec/formatting.md#resolve-preferences),
```
eemeli, line 472:
```suggestion
When implementing [`MatchSelectorKeys`](formatting.md#resolve-preferences), 
```

spec/registry.md
```diff
@@ -255,3 +255,502 @@ which only expects the `accord` option:
...
+   - Else, if the value of `key` is a keyword:
+      - Let `keyword` be a string which is the result of [rule selection](#rule-selection).
+      - If `keyword` equals `key`, then `key` matches the selector.
+        Append `key` to the end of the `return_value` list.
+   - Else, `key` is invalid;
+     emit a _Selection Error_.
+     Do not add `key` to `return_value`
+- Return `return_value`
+
+### Plural/Ordinal Keywords
```
eemeli, line 495:
```suggestion
#### Plural/Ordinal Keywords
```

spec/registry.md
```diff
@@ -255,3 +255,502 @@ which only expects the `accord` option:
...
+The _plural/ordinal keywords_ are: `zero`, `one`, `two`, `few`, `many`, and
+`other`.
+
+### Rule Selection
```
eemeli, line 499:
```suggestion
#### Rule Selection
```

spec/registry.md
```diff
@@ -255,3 +255,502 @@ which only expects the `accord` option:
...
+> Feedback describing use cases for fractional and significant digits-based
+> selection would be helpful.
+> Otherwise, users should avoid using matching with fractional numbers or significant digits.
+
+## Date and Time Value Formatting
+
+This subsection describes the functions and options for date/time formatting.
+Selection based on date and time values is not required in this release.
+
+#### Functions
```
eemeli, line 573:
```suggestion
### Functions
```

spec/registry.md
```diff
@@ -255,3 +255,502 @@ which only expects the `accord` option:
...
+- `{$d :date}` is the same as `{$d :date style=short}`
+- `{$t :time}` is the same as `{$t :time style=short}`
+
+> [!NOTE]
+> Pattern selection based on date/time values is a complex topic and no support for this
+> is required in this release.
+
+> [!NOTE]
+> The default formatting behavior of `:datetime` is inconsistent with `Intl.DateTimeFormat`
+> in JavaScript and with `{d,date}` in MessageFormat v1.
```
eemeli, line 592:
```suggestion
> in JavaScript and with `{d,date}` in ICU MessageFormat 1.0.
```
as per https://github.com/unicode-org/message-format-wg/blob/3e30503810046b123267e1cc8b86e0b632f446d9/spec/README.md?plain=1#L54-L55

spec/registry.md
```diff
@@ -255,3 +255,502 @@ which only expects the `accord` option:
...
+> This is because, unlike those implementations, `:datetime` is distinct from `:date` and `:time`.
+
+#### Operands
+
+The _operand_ of a date/time function is either 
+an implementation-defined date/time type (passed in as an argument)
+or a _date/time literal value_, as defined below.
+All other _operand_ values produce a _Selection Error_ when evaluated for selection
+or a _Formatting Error_ when formatting the value.
```
eemeli, line 601: Same as with `:number`, these errors are emitted because we first emit a _Resolution Error_ and end up with a fallback value.

spec/registry.md
```diff
@@ -255,3 +255,502 @@ which only expects the `accord` option:
...
+- an XMLSchema 1.1 [time](https://www.w3.org/TR/xmlschema11-2/#time)
+- an XMLSchema 1.1 [date](https://www.w3.org/TR/xmlschema11-2/#date)
+
+The `timezoneOffset` of each of these formats is optional. 
+When the offset is not present, implementations should use a floating time type
+(such as Java's `java.time.LocalDateTime`) to represent the time value.
+For more information, see [Working with Timezones](https://w3c.github.io/timezone).
+
+> [!IMPORTANT]
+> The [ABNF](/spec/message.abnf) and [syntax](/spec/syntax.md) of MFv2
```
eemeli, line 615:
```suggestion
> The [ABNF](/spec/message.abnf) and [syntax](/spec/syntax.md) of MF2
```

spec/registry.md
```diff
@@ -255,3 +255,502 @@ which only expects the `accord` option:
...
+> ```
+
+> [!NOTE]
+> True time zone support in serializations is expected to coincide with the adoption
+> of Temporal in JavaScript.
+> The form of these serializations is known and is a de facto standard.
+> Support for these extensions is expected to be required in the post-tech preview.
+> See: https://datatracker.ietf.org/doc/draft-ietf-sedate-datetime-extended/
+
+#### Options
```
eemeli, line 640:
```suggestion
### Options
```

spec/registry.md
```diff
@@ -255,3 +255,502 @@ which only expects the `accord` option:
...
+
+A function can use either the appropriate _style_ options for that function
+or can use a collection of _field options_ (but not both) to control the formatted 
+output.
+
+If both are specified, an _Invalid Expression_ error MUST be emitted
+and a _fallback value_ used as the resolved value of the _expression_.
+
+##### Style Options
```
eemeli, line 649:
```suggestion
#### Style Options
```

spec/registry.md
```diff
@@ -255,3 +255,502 @@ which only expects the `accord` option:
...
+  - `short` (default)
+
+The function `:time` has these function-specific _style_ options:
+- `style`
+  - `full`
+  - `long`
+  - `medium`
+  - `short` (default)
+
+##### Field Options
```
eemeli, line 677:
```suggestion
#### Field Options
```

spec/registry.md
```diff
@@ -255,3 +255,502 @@ which only expects the `accord` option:
...
+  - `shortOffset`
+  - `longOffset`
+  - `shortGeneric`
+  - `longGeneric`
+
+> [!NOTE]
+> The following options do not have default values because they are only to be used
+> as overrides for locale-and-value dependent implementation-defined defaults.
+
+The followind date/time options are *not* part of the default registry.
```
eemeli, line 741:
```suggestion
The following date/time options are **not** part of the default registry.
```

spec/registry.md
```diff
@@ -255,3 +255,502 @@ which only expects the `accord` option:
...
+Implementations SHOULD avoid creating options that conflict with these, but
+are encouraged to track development of these options during Tech Preview:
+- `calendar` (default is locale-specific)
+  - valid [Unicode Calendar Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeCalendarIdentifier)
+- `numberingSystem` (default is locale-specific)
+   - valid [Unicode Number System Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeNumberSystemIdentifier)
+- `timeZone` (default is system default time zone or UTC)
+  - valid identifier per [BCP175](https://www.rfc-editor.org/rfc/rfc6557)
+
+#### Selection
```
eemeli, line 751:
```suggestion
### Selection
```

spec/registry.md
```diff
@@ -255,3 +255,502 @@ which only expects the `accord` option:
...
+
+Selection based on date/time types is not required by MFv2.
```
eemeli, line 753:
```suggestion
Selection based on date/time types is not required by MF2.
```

eemeli (changes requested):

spec/registry.md
aphillips: This is intended to be a subsection of `selection`, not on the same level?

spec/registry.md
aphillips: No, we should just say _Resolution Error_ here. Implementers need to know what to throw/emit and that's what they'll emit when parsing the literal.

```suggestion
All other values produce a _Resolution Error_.
```

spec/registry.md
eemeli: Exactly, and Selection has `###` so this needs `####`.

spec/registry.md
```diff
@@ -255,3 +255,498 @@ which only expects the `accord` option:
...
+> | Operand value | Keyword | Formatted Message |
+> |---|---|---|
+> | 1 | `one` | 1 den |
+> | 2 | `few` | 2 dny |
+> | 5 | `other` | 5 dní |
+> | 22 | `few` | 22 dny |
+> | 27 | `other` | 27 dní |
+> | 2.4 | `many` | 2,4 dne |
+
+### Determining Exact Literal Match
```
eemeli, line 538:
```suggestion
#### Determining Exact Literal Match
```

spec/registry.md
```diff
@@ -255,3 +255,498 @@ which only expects the `accord` option:
...
+> [!NOTE]
+> Pattern selection based on date/time values is a complex topic and no support for this
+> is required in this release.
+
+> [!NOTE]
+> The default formatting behavior of `:datetime` is inconsistent with `Intl.DateTimeFormat`
+> in JavaScript and with `{d,date}` in ICU MessageFormat 1.0.
+> This is because, unlike those implementations, `:datetime` is distinct from `:date` and `:time`.
+
+#### Operands
```
eemeli, line 591:
```suggestion
### Operands
```

eemeli (changes requested):
Noticed a couple more incorrect title levels, and a few of my previous line comments are still open. Most significantly, the `style=percent` multiplication instead of division.

eemeli (changes requested):
This still needs too be addressed: https://github.com/unicode-org/message-format-wg/pull/659#discussion_r1492735437

spec/registry.md
```diff
@@ -255,3 +255,500 @@ which only expects the `accord` option:
...
+  - (non-negative integer)
+- `maximumFractionDigits`
+  - (non-negative integer)
+- `minimumSignificantDigits`
+  - (non-negative integer)
+- `maximumSignificantDigits`
+  - (non-negative integer)
+
+The following options and their values are required in the default registry to be available on the 
+function `:integer`:
```
gibson042, line 385: Suggested restructuring (and likewise for Date and Time Value Formatting and other sections):

> ## Numeric Value Selection and Formatting
> …
> ### `:number`
> The function `:number` is the default selector and formatter for numeric values.
> 
> #### Operand
> See [Number Function Operands](#number-function-operands).
> 
> #### Options
> * all [Number Function Common Options](#number-function-options)
> * `compactDisplay` (this option only has meaning when combined with the option `notation=compact`)
>   * `short`
>   * `long`
> * `notation`
>   * `standard` (default)
>   * `scientific`
>   * `engineering`
>   * `compact`
> * `useGrouping`
>   * `auto` (default)
>   * `always`
>   * `never`
>   * `min2`
> * `minimumFractionDigits`
>   * (non-negative integer) (default is implementation-dependent and may depend upon locale and/or value and/or other options)
> * `maximumFractionDigits`
>   * (non-negative integer) (default is implementation-dependent and may depend upon locale and/or value and/or other options)
> * `minimumSignificantDigits`
>   * (non-negative integer) (default is implementation-dependent and may depend upon locale and/or value and/or other options)
> 
> ### `:integer`
> The function `:integer` provides a reduced set of options for selecting and formatting numeric values as integers.
> 
> #### Operand
> See [Number Function Operands](#number-function-operands).
> 
> #### Options
> * all [Number Function Common Options](#number-function-options)
> * `useGrouping`
>   * `auto` (default)
>   * `true`
>   * `false`
>   * `min2`
>   * `always`

spec/registry.md
```diff
@@ -255,3 +255,500 @@ which only expects the `accord` option:
...
+   -  `never`
+- `style`
+  - `decimal` (default)
+  - `percent` (see [Percent Style](#percent-style) below)
+- `useGrouping`
+  - `auto` (default)
+  - `true`
+  - `false`
+  - `min2`
+  - `always`
```
gibson042, line 407: Out of curiosity, what determines the order in which to display possible values?

spec/registry.md
```diff
@@ -255,3 +255,500 @@ which only expects the `accord` option:
...
+> default
+
+- `minimumFractionDigits`
+  - (non-negative integer)
+- `maximumFractionDigits`
+  - (non-negative integer)
+- `minimumSignificantDigits`
+  - (non-negative integer)
+- `maximumSignificantDigits`
+  - (non-negative integer)
```
gibson042, line 382: I'm not convinced that this note is helpful; the same dynamic default behavior is also true of `numberingSystem`.
```suggestion
- `minimumFractionDigits`
  - (non-negative integer) (default is implementation-dependent and may depend upon locale and/or value and/or other options)
- `maximumFractionDigits`
  - (non-negative integer) (default is implementation-dependent and may depend upon locale and/or value and/or other options)
- `minimumSignificantDigits`
  - (non-negative integer) (default is implementation-dependent and may depend upon locale and/or value and/or other options)
- `maximumSignificantDigits`
  - (non-negative integer) (default is implementation-dependent and may depend upon locale and/or value and/or other options)
```

spec/registry.md
```diff
@@ -255,3 +255,500 @@ which only expects the `accord` option:
...
+> Annotations that use fraction digits or significant digits might work in specific
+> implementation-defined ways.
+> Users should avoid depending on these types of keys in message selection.
+
+
+Number literals in the MessageFormat 2 syntax use the 
+[format defined for a JSON number](https://www.rfc-editor.org/rfc/rfc8259#section-6).
+The resolved value of an `operand` exactly matches a numeric literal `key`
+if, when the `operand` is serialized using the format for a JSON number
+the two strings are equal.
```
gibson042, line 554:
```suggestion
The resolved value of an `operand` exactly matches a numeric literal `key`
if, when the `operand` is serialized using the format for a JSON number,
the two strings are equal.
```

spec/registry.md
```diff
@@ -255,3 +255,500 @@ which only expects the `accord` option:
...
+
+If no options are specified, each of the functions defaults to the following:
+- `{$d :datetime}` is the same as `{$d :datetime dateStyle=short timeStyle=short}`
+- `{$d :date}` is the same as `{$d :date style=short}`
+- `{$t :time}` is the same as `{$t :time style=short}`
+
+> [!NOTE]
+> Pattern selection based on date/time values is a complex topic and no support for this
+> is required in this release.
+
```
gibson042, line 587: Redundant with text above.
```suggestion
```

spec/registry.md
```diff
@@ -255,3 +255,500 @@ which only expects the `accord` option:
...
+The function `:number` is the default selector and formatter for numeric values.
+
+The function `:integer` provides a reduced set of options for selecting
+and formatting numeric values as integers.
+
+### Operands
+
+The _operand_ of a number function is either an implementation-defined type or
+a literal that matches the `number-literal` production in the [ABNF](/spec/message.abnf).
+All other values produce a _Resolution Error_.
```
gibson042, line 281: This is inconsistent with Date and Time Value Formatting:
> All other _operand_ values produce a _Selection Error_ when evaluated for selection or a _Formatting Error_ when formatting the value.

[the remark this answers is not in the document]
aphillips: Yes, we do! See the folder 'test' in our repo. I have filed an issue to capture this comment.

aphillips: If we didn't correct date/time previously, I will do so. This was a discussion with @eemeli. The _Resolution Error_ is what comes first.

spec/registry.md
```diff
@@ -255,3 +255,500 @@ which only expects the `accord` option:
...
+The following options and their values are required in the default registry to be available on the 
+function `:number`:
+- `select`
+   -  `plural` (default)
+   -  `ordinal`
+   -  `exact`
+
+ 
+> [!IMPORTANT]
+> **Default Value of `select` Option**
```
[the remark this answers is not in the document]
aphillips: Will fix.

spec/registry.md
```diff
@@ -255,3 +255,500 @@ which only expects the `accord` option:
...
+>```
+> .match {$var}
+> 1   {{You have one last chance}}
+> one {{You have {$var} chance remaining}} // needed by languages such as Polish or Russian
+>                                          // such locales typically require other keywords
+>                                          // such as two, few, many, and so forth
+> *   {{You have {$var} chances remaining}}
+>```
+
+- `compactDisplay` (this option only has meaning when combined with the option `notation=compact`)
```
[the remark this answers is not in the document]
aphillips: We don't do that for all of the other values--and sometimes this is very deliberate. We might refer to MDN or 402 for the compact display gunk, though, since otherwise this gives implementations the "freedom" to implement `short` and `long` but not other forms of `notation=compact`.

What this really argues for is: we should excommunicate `compactDisplay` and `notation` from Tech Preview and consider it for RGI instead. JS and ICU can implement as an extension. (Scientific and engineering notation are also specialized)

spec/registry.md
aphillips: Could we instead put something above all of the options in a non-NOTE style that says something like:

> Some options do not have default values defined in this specification.
> The defaults for these options are implementation-dependent.
> In general, the defaults depend on the locale, the value of other options,
> or both.

spec/registry.md
aphillips:
```suggestion
All other _operand_ values produce a _Selection Error_.
```

spec/registry.md
aphillips: Ah... I see. I _did_ fix it---in `:number` but not in `:integer`. Glad you asked!

spec/registry.md
aphillips: I implemented this. See what you think.

spec/registry.md
```diff
@@ -255,3 +255,548 @@ which only expects the `accord` option:
...
+The _operand_ of a date/time function is either 
+an implementation-defined date/time type (passed in as an argument)
+or a _date/time literal value_, as defined below.
+All other _operand_ values produce a _Selection Error_.
+
+A **_<dfn>date/time literal value</dfn>_** is a non-empty string consisting of 
+one of the following:
+- an XMLSchema 1.1 [dateTime](https://www.w3.org/TR/xmlschema11-2/#dateTime)
+- an XMLSchema 1.1 [time](https://www.w3.org/TR/xmlschema11-2/#time)
+- an XMLSchema 1.1 [date](https://www.w3.org/TR/xmlschema11-2/#date)
```
eemeli, line 653: We should clarify which functions take which of these values. All can't be valid for all, as e.g. an XML time is insufficient input for `:date`.

Also, are any timezones in literal inputs used in the formatted outputs? For example, when formatting
```
{12:34+06:00 :time}
```
is the resulting formatted time in the system timezone, or in +06?

How about UTC, what timezone does this render in?
```
{12:34Z :time}
```

As we've left out `timeZone` from the available options, also leaving out timezone parsing would be one way to resolve this.

eemeli (changes requested):

spec/registry.md
aphillips: Oh boy, where do I start? 😈 

The values can parse into various kinds of date/time objects. Many (most?) of these can be formatted by any of the functions given here, although the results may be not be very useful. Specifically, a `time` formatted as a date or a `date` formatted as a time is probably not very interesting.

XMLSchema date/time values _do not have timezones_. They only support _zone offsets_. The only zone offset that can be 100% certainly made into a time zone is `Z` because that one means "UTC".

The time zone is very important in rendering date and time values. I argued at length in the previous PR (and tried to make the point in the call) that we should not excommunicate `timezone` as an option to "future". Without that option, there are only the following options for managing time zone in formatting:

- the implementation uses the system default
- the implementation provides an API for setting the time zone of an entire message
  (this is not a behavior we specify because it is beyond our remit, although probably
  we should point it out)
- the implementation assumes UTC

In addition, date/time _literals_ with no offset should be handled as floating time values (even if the implementation uses some specific time zone, the value should be formatted so that the field values don't change--often this is done using UTC as the time zone, ignoring local settings)

Note well that "time zone parsing" doesn't exist here. The _zone offset_ (usually include `Z`) parses into the underlying incremental ("instant") value. That is, all of the following are the *same* time value:

- `2007-01-01T01:00:00.000-01:00`
- `2007-01-01T01:00:00.000Z` 
- `2007-01-01T01:00:00.000+01:00` 
- `2006-12-31T23:00:00.000-01:00`

Literals missing one part should be permitted to treat that part as `0`. Some examples:

| Literal | Pattern | Output | Comment |
|---|---|---|---|
| `2024-02-17` | `{$d :datetime dateStyle=medium}` | Feb 17, 2024 00:00 a.m. | No time; treat as floating |
| `12:34:56.000` | {$d :datetime dateStyle=medium}` | Jan 1, 1970 12:34 p.m. | No date; treat as floating |
| `2024-02-17T12:34:56.000Z` | {$d :dateTime dateStyle=medium} | Feb 17, 2024 4:34:56 a.m. Pacific | Surprised? |
| `2024-02-17T12:34:56.000-08:00` | {$d :dateTime dateStyle=medium} | Feb 17, 2024 12:34 p.m. Pacific | |

If we didn't do this, users used to working with partial values would have messages fail instead of doing what their code usually does. The group will need to decide if we bring `timeZone` back in Tech Preview. We should also discuss whether to specify any of this behavior.

Some of what I express above might be controversial, in that we could make different decisions about forcing floating time behavior or permitting partial values.

macchiati:
Ah, I hadn't realized that timezones were not handled; only offsets is
pretty brain-dead for end users.

aphillips:
@macchiati 

We have to be careful not to conflate things here. We could support XMLSchema date/time/dateTime types for parsing literals. These do not have time zones, only offsets, but that can be okay. Values with an offset parse into incremental times (such as JS `Date`) without heartburn.

Setting/managing time zone for the formatting function is should be handled separately. The three cases I enumerate above aren't terrible, but they are limiting. Temporal support should come with Temporal in JS (and corresponding changes in serializations). Java can already do the Temporal thing, but we're trying to be the "strict subset".

So...

In LDML45, I propose:

- `:datetime` accepts XMLSchema `date, `time`, and `dateTime` types as literal operands.

I also propose that we bring the option `timeZone` back above the fold.

```
.local $time = {|2024-02-17| :time}
{{Today is {$time :date}}}

.local $date = {|01:23:45Z| :date}
{{It is now {$date :time}}}

.local $time = {|2024-02-17| :time}
{{Today is {$time :datetime}}}
```

Having a minimal guardrail seems like a good idea, though.

macchiati:
Your whole picture sounds good to me. (I have a concern for literals being
expressive enough, mostly for those languages where the user has a choice
of formats. However, that can wait.)

aphillips:
@macchiati wrote:
> I have a concern for literals being
> expressive enough, mostly for those languages where the user has a choice
> of formats.

I'm not sure I follow? The literals allow users to type a date/time value into a message. This is going to be rare, since it is uncommon to statically hardcode a specific date or time vs. passing it as a variable. In most cases, we'll actually be working with `java.util.Date`, `java.time.Temporal`, `time_t`, JS `Date`, etc.

I didn't offer all of the XMLSchema date/time related serializations, the complete list of which is:
> [duration](https://www.w3.org/TR/xmlschema11-2/#duration) · [dateTime](https://www.w3.org/TR/xmlschema11-2/#dateTime) · [time](https://www.w3.org/TR/xmlschema11-2/#time) · [date](https://www.w3.org/TR/xmlschema11-2/#date) · [gYearMonth](https://www.w3.org/TR/xmlschema11-2/#gYearMonth) · [gYear](https://www.w3.org/TR/xmlschema11-2/#gYear) · [gMonthDay](https://www.w3.org/TR/xmlschema11-2/#gMonthDay) · [gDay](https://www.w3.org/TR/xmlschema11-2/#gDay) · [gMonth](https://www.w3.org/TR/xmlschema11-2/#gMonth)

And we'll want the SEDATE extensions that provide real time zones for `date`, `time`, and `dateTime`.

I was somewhat conservative since I expect date/time handling to evolve with work on skeletons, Temporal, and such.

But what I don't see is how we would need to change literals to support different formats. Are you thinking about alternate calendars?

spec/registry.md
aphillips: Note to @eemeli... see disconnected conversation below ⬇️

spec/registry.md
```diff
@@ -255,3 +255,548 @@ which only expects the `accord` option:
 > .input {$object :noun case=nominative}
 > {{You see {$color :adjective article=indefinite accord=$object} {$object}!}}
 >```
+
+# Default Registry
+
+This section describes the functions which each implementation MUST provide to be conformant with
+this specification.
```
catamorphism, line 262:
```suggestion
This section describes the functions which each implementation MUST provide
to be conformant with this specification.
```

spec/registry.md
```diff
@@ -255,3 +255,548 @@ which only expects the `accord` option:
...
+
+### Functions
+
+The following functions are provided:
+
+The function `:string` provides string selection and formatting.
+
+### Operands
+
+The _operand_ is any literal or an implementation-defined set of string or
```
catamorphism, line 274: I wonder if we should use the term "string" or "string literal" here (and include a definition of "string literal" as an arbitrary string which the formatting function may parse in any way) to avoid confusion with `Literal` in the grammar.

spec/registry.md
```diff
@@ -255,3 +255,548 @@ which only expects the `accord` option:
...
+character-sequence types.
+
+> [!NOTE]
+> This should probably include individual character types, such as `char`.
+
+In addition, implementations MAY perform formatting and selection on 
+`operand` values that do not otherwise have a formatting function registered.
```
catamorphism, line 281: Is this saying that the `:string` formatter/selector may be used as a "default" by the implementation? If so, it's a little confusing because it's unclear whether to read the sentence that way, or read it to mean that the implementation of `:string` may do arbitrary formatting on its operand.

spec/registry.md
```diff
@@ -255,3 +255,548 @@ which only expects the `accord` option:
...
+
+The function `:string` has no options.
+
+> [!NOTE]
+> Proposals for string transformation options or implementation
+> experience with user requirements is desired during the Tech Preview.
+
+### Selection
+
+When implementing [`MatchSelectorKeys`](spec/formatting.md#resolve-preferences),
```
catamorphism, line 293:
```suggestion
When implementing [`MatchSelectorKeys(rv, keys)`](spec/formatting.md#resolve-preferences), 
```

spec/registry.md
```diff
@@ -255,3 +255,548 @@ which only expects the `accord` option:
...
+
+When implementing [`MatchSelectorKeys`](spec/formatting.md#resolve-preferences), 
+the `:string` selector performs as described below.
+
+- Let `return_value` be a new empty list of strings.
+- Let `operand` be the resolved value of the _operand_.
+  If the `operand` is not a string literal, convert the value to a string literal,
+  or, optionally: emit a _Selection Error_ and return `return_value`.
+- Let `keys` be a list of strings containing keys to match.
+  (Hint: this list is an argument to `MatchSelectorKeys`)
```
catamorphism, line 301: These two lines can be deleted if my previous suggestion is used (including the argument names to `MatchSelectorKeys` above. That would make it clearer what's an argument to the procedure and what's a local variable.

spec/registry.md
```diff
@@ -255,3 +255,548 @@ which only expects the `accord` option:
...
+> experience with user requirements is desired during the Tech Preview.
+
+### Selection
+
+When implementing [`MatchSelectorKeys`](spec/formatting.md#resolve-preferences), 
+the `:string` selector performs as described below.
+
+- Let `return_value` be a new empty list of strings.
+- Let `operand` be the resolved value of the _operand_.
+  If the `operand` is not a string literal, convert the value to a string literal,
```
catamorphism, line 298:
```suggestion
- Let `operand` be _rv_.
If `operand` is not a string literal, let `operand` be
the result of converting `operand` to a string literal.
```

(See above suggestoin)

spec/registry.md
```diff
@@ -255,3 +255,548 @@ which only expects the `accord` option:
...
+  or, optionally: emit a _Selection Error_ and return `return_value`.
+- Let `keys` be a list of strings containing keys to match.
+  (Hint: this list is an argument to `MatchSelectorKeys`)
+- For each string `key` in `keys`:
+   - If the value of `key` is equal to the string value of `operand`
```
catamorphism, line 303:
```suggestion
   - If the value of `key` is equal to `operand`
```

`operand` is already a string value

spec/registry.md
```diff
@@ -255,3 +255,548 @@ which only expects the `accord` option:
...
+     then `key` matches the selector.
+     A `key` and an `operand` are equal if they consist of the same
```
catamorphism, line 305:
```suggestion
    Two strings are equal if they consist of the same
```

spec/registry.md
```diff
@@ -255,3 +255,548 @@ which only expects the `accord` option:
...
+     sequence of Unicode code points.
+     Add `key` to the front of the `return_value` list.
```
catamorphism, line 307: I think this is wrong. The `return_value` list should have length 0 or 1, per this part of the formatting spec:

"The returned list MUST contain only unique elements of the input list `keys`."

The only way that `return_value` could have length > 1 is if it includes duplicates, since all of the elements in the list must be equal to each other (given the previous step).

spec/registry.md
```diff
@@ -255,3 +255,548 @@ which only expects the `accord` option:
...
+- Return `return_value`
```
catamorphism, line 308:
```suggestion
- Return `return_value`.
```

spec/registry.md
```diff
@@ -255,3 +255,548 @@ which only expects the `accord` option:
...
+The function `:number` is the default selector and formatter for numeric values.
+
+The function `:integer` provides a reduced set of options for selecting
+and formatting numeric values as integers.
+
+### Operands
+
+The _operand_ of a number function is either an implementation-defined type or
+a literal whose contents match the `number-literal` production in the [ABNF](/spec/message.abnf).
+All other values produce a _Resolution Error_.
```
catamorphism, line 346:
```suggestion
All other values produce an _Invalid Expression Error_.
```

Per #630

spec/registry.md
```diff
@@ -255,3 +255,548 @@ which only expects the `accord` option:
...
+- `unit`
+   - (anything not empty)
+- `unitDisplay`
+   - `long`
+   - `short` (default)
+   - `narrow`
+
+#### Default Value of `select` Option
+
+The value `plural` is default for the option `select`
```
catamorphism, line 485:
```suggestion
The value `plural` is the default for the option `select` 
```

spec/registry.md
```diff
@@ -255,3 +255,548 @@ which only expects the `accord` option:
...
+### Selection
+
+Number selection has three modes:
+- `exact` selection matches the operand to explicit numeric keys exactly
+- `plural` selection matches the operand to explicit numeric keys exactly
+  or to plural rule categories if there is no explicit match
+- `ordinal` selection matches the operand to explicit numeric keys exactly
+  or to ordinal rule categories if there is no explicit match
+
+When implementing [`MatchSelectorKeys`](formatting.md#resolve-preferences),
```
catamorphism, line 523: Same general suggestions as the ones I made for `:string` (name the arguments here, then refer to them below)

spec/registry.md
```diff
@@ -255,3 +255,548 @@ which only expects the `accord` option:
...
+- Let `operand` be the resolved value of the _operand_.
+  If the `operand` is not a number type, emit a _Selection Error_
+  and return `return_value`.
+- Let `keys` be a list of strings containing keys to match.
+  (Hint: this list is an argument to `MatchSelectorKeys`)
+- For each string `key` in `keys`:
+   - If the value of `key` matches the production `number-literal`:
+     - If the parsed value of `key` is an [exact match](#determining-exact-literal-match)
+       of the value of the `operand`, then `key` matches the selector.
+       Add `key` to the front of the `return_value` list.
```
catamorphism, line 536: Duplicates?

spec/registry.md
```diff
@@ -255,3 +255,548 @@ which only expects the `accord` option:
...
+   - Else, if the value of `key` is a keyword:
+      - Let `keyword` be a string which is the result of [rule selection](#rule-selection).
+      - If `keyword` equals `key`, then `key` matches the selector.
+        Append `key` to the end of the `return_value` list.
```
catamorphism, line 540:
```suggestion
        Append `key` onto the `return_value` list.
```

spec/registry.md
```diff
@@ -255,3 +255,548 @@ which only expects the `accord` option:
...
+   - Else, `key` is invalid;
+     emit a _Selection Error_.
+     Do not add `key` to `return_value`
```
catamorphism, line 543:
```suggestion
     Do not add `key` to `return_value`.
```

spec/registry.md
```diff
@@ -255,3 +255,548 @@ which only expects the `accord` option:
...
+- Return `return_value`
```
catamorphism, line 544:
```suggestion
- Return `return_value`.
```

spec/registry.md
```diff
@@ -255,3 +255,548 @@ which only expects the `accord` option:
...
+`other`.
+
+#### Rule Selection
+
+If the option `select` is set to `exact`, rule-based selection is not used.
+Return the empty string.
+
+> [!NOTE]
+> Since keys cannot be the empty string in a numeric selector, returning the
+> empty string disables keyword selection
```
catamorphism, line 557:
```suggestion
> empty string disables keyword selection.
```

spec/registry.md
```diff
@@ -255,3 +255,548 @@ which only expects the `accord` option:
...
+
+Number literals in the MessageFormat 2 syntax use the 
+[format defined for a JSON number](https://www.rfc-editor.org/rfc/rfc8259#section-6).
+The resolved value of an `operand` exactly matches a numeric literal `key`
+if, when the `operand` is serialized using the format for a JSON number,
+the two strings are equal.
+
+> [!NOTE]
+> Implementations are not expected to implement this exactly as written,
+> as there are clearly optimizations that can be applied.
```
catamorphism, line 611: Add something about how the observable behavior must be the same as what is written?

catamorphism (changes requested):
I left a few comments, particularly wrt the `MatchSelectorKeys` algorithms for `:string` and `:number`, but overall this is great!

spec/registry.md
aphillips: Actually, this text really **_means_** the production `Literal` in the grammar. The operand is any such literal _or_ implementation defined types.

spec/registry.md
aphillips: The list can have duplicates. Consider:

```
.match {$foo :string}{$num :integer}
foo 0   {{You have no foo}}
foo one {{You have {$num} foo}}
foo *   {{You have {$num} foo}}
*   0   {{You have no {$foo}}}
*   one {{You have {$num} {$foo}}}
*   *   {{You have {$num} {$foo}}}
```

spec/registry.md
aphillips: Actually, we're both wrong. This is a failure inside the function, so it's Invalid Expression.

spec/registry.md
aphillips:
```suggestion
as there are clearly optimizations that can be applied.
However, the observed behavior must be consistent with what is described here.
```

aphillips:
@eemeli @gibson042 @stasm @catamorphism @macchiati 

I would like to commit this today (preferably before the open house!)

- I have an issue open against compact notation #664 (cf. our conversation @gibson042)
- I have an issue open to restore time zone #665 (cf. our conversation @eemeli)
- I have fixed the error emitted for a bad operand

Anything else that must go in before we commit?

spec/registry.md
catamorphism: What's confusing me is that there are several different meanings of "literal":

* any string
* a data model node representing a `Literal` node in the tree
* a string that is generated by one of the following three productions in the grammar: `*(quoted-char / quoted-escape)`; `name` or `number-literal`

Put differently, the string "|1|" is a valid syntactic representation of a number literal, but not a valid argument to the number formatter.

catamorphism:
> @eemeli @gibson042 @stasm @catamorphism @macchiati
> 
> I would like to commit this today (preferably before the open house!)
> 
>     * I have an issue open against compact notation [Remove `:number` (etc) support for `notation` and related option `compactDisplay` from preview #664](https://github.com/unicode-org/message-format-wg/issues/664) (cf. our conversation @gibson042)
> 
>     * I have an issue open to restore time zone [Bring back option `timeZone` and restrict literals for `:date` and `:time` #665](https://github.com/unicode-org/message-format-wg/issues/665) (cf. our conversation @eemeli)
> 
>     * I have fixed the error emitted for a bad operand
> 
> 
> Anything else that must go in before we commit?

Sorry, I didn't see this before the open house. I would still like the "literal" part to be clarified per my previous comment, but I won't block on that.

catamorphism (approved):
Ideally would like to see the two nontrivial open comments I left before addressed, but I won't block on that.

aphillips:
@eemeli @gibson042 @stasm @macchiati @ everyone...
> 
>  * I have an issue open against compact notation [Remove `:number` (etc) support for `notation` and related 
>   option `compactDisplay` from preview #664](https://github.com/unicode-org/message-format-wg/issues/664)
>   (cf. our conversation @gibson042)
> 
>  * I have an issue open to restore time zone [Bring back option `timeZone` and restrict literals for `:date` and `:time` #665](https://github.com/unicode-org/message-format-wg/issues/665) (cf. our conversation @eemeli)
> 
>  * I have fixed the error emitted for a bad operand
> 
> 
> Anything else that must go in before we commit?

Thanks to @catamorphism!

I need one more approval to merge this. We don't have a specification without it. Please help.

spec/registry.md
catamorphism: `{|1| :number}` is valid syntax. But the implementation parses `|1|` into a `Literal` data model node (or does something observably the same), and is then responsible for extracting the string "1" from that node and passing it into the runtime function that implements `:number`. If it passed the string `|1|` to `:number` instead, that wouldn't be spec-compliant.

spec/registry.md
```diff
@@ -255,3 +255,550 @@ which only expects the `accord` option:
...
+
+### Functions
+
+The following functions are provided:
+
+The function `:string` provides string selection and formatting.
+
+### Operands
+
+The _operand_ is any literal or an implementation-defined set of string or
```
catamorphism, line 274:
```suggestion
The _operand_ is any string generated by the
`*(quoted-char / quoted-escape)`, `name`, or `number-literal` productions in the grammar;
or an implementation-defined set of string or
```

spec/registry.md
aphillips: I think we're saying the same thing? The literal value of `|1|` is the string `1` and doesn't include the quoting pipe symbols. Does my suggestion help express that? What should I change?

catamorphism (approved):
Still approving, just added more elaboration on the "literal" terminology

spec/registry.md
catamorphism: See the suggestion I made (at the same time as you wrote this comment :)

spec/registry.md
aphillips: I see what you're doing!

The `name` or `number-literal` productions are narrower than the first description (which is the guts of a quoted literal). That narrower production defines what's allowed in an unquoted literal, which has no real bearing here (it's a detail of the syntax). 

The verb "generated" also bothers me, as one doesn't generate the operand. I kind of want to say the "resolved value of a _literal_", since one of the things that should happen is that the `quoted-escape` becomes unescaped in the data-model or in-memory representation. It's only escaped in the syntax to avoid conflicting with the grammar.

Perhaps:

```suggestion
The _operand_ is any _literal_'s value less any quotes and with any `quoted-escape` sequences
converted to their escaped characters
or an implementation-defined set of string or
```

We should add an example too.

aphillips: I rewrote and added an example. Thanks!

spec/registry.md
```diff
@@ -255,3 +255,563 @@ which only expects the `accord` option:
...
+> The exact behavior of exact literal match is only defined for non-zero-filled
+> integer values.
+> Annotations that use fraction digits or significant digits might work in specific
+> implementation-defined ways.
+> Users should avoid depending on these types of keys in message selection.
+
+
+Number literals in the MessageFormat 2 syntax use the 
+[format defined for a JSON number](https://www.rfc-editor.org/rfc/rfc8259#section-6).
+The resolved value of an `operand` exactly matches a numeric literal `key`
```
macchiati, line 619: I can't find anywhere where we define what "resolved value" means.

spec/registry.md
```diff
@@ -255,3 +255,563 @@ which only expects the `accord` option:
...
+> |---|---|---|
+> | \|1\| | `1` | Does not include quotes `\|` |
+> | \|ab\\\|cd\| | `ab\|cd` | Unescape escaped `\|` |
+> | hello | `hello` | Unquoted literal |
+> | -13.2 | `-13.2` | Unquoted number-literal |
+> | \| spaces \| | ` spaces ` | Includes one space before and after |
+
+
+> [!NOTE]
+> This should probably include individual character types, such as `char`.
```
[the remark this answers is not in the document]
aphillips: That's exactly what I mean: don't overlook `char` and `java.lang.Character`

spec/registry.md
```diff
@@ -255,3 +255,563 @@ which only expects the `accord` option:
...
+     sequence of Unicode code points.
+     Add `key` to the front of the `return_value` list.
+- Return `return_value`.
+
+> [!NOTE]
+> Matching of `key` and `operand` values is sensitive to the sequence of code points
+> in each string.
+> As a result, variations in how text can be encoded can affect the performance of matching.
+> The function `:string` does not perform case folding or Unicode Normalization of string values.
+
```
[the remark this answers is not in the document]
aphillips: https://www.w3.org/TR/charmod-norm/

Requiring implementations to do normalization can be a big ask (I spent a long time at W3C asking...) and there are counterarguments. We should certainly **recommend** that users stick to normalized sequences. That is what this note is trying to do. We can mention NFC here if you think that would help.

spec/registry.md
aphillips: The term "resolved value" lacks a formal definition, but the various sorts of resolved values are defined [here](https://github.com/unicode-org/message-format-wg/blob/main/spec/formatting.md#expression-and-markup-resolution) (in the formatting spec). This may be a gap.

I will do a PR with a formal definition because you're not the first to ask.

spec/registry.md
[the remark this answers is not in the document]
macchiati: Yes, I think that would be useful. (I wasn't proposing requiring it.)

spec/registry.md
macchiati: Sounds good.

spec/registry.md
macchiati: I was confused as to what you meant by a 'resolved value', and could see two possible interpretations. My comment was specific to :string (because as you say the results are always string comparisons anyway).

The spec always has to work in terms of strings, because it knows nothing about datatypes, and can't require particular datatypes because they vary by environment and implementation capabilities.

However, I would expect many MF2 implementations to actually operate with inputs that are not strings, but rather datatypes: both primitive and non-primitive. We can't have a spec that precludes implementations of MF2 from doing that. And those implementations will want to keep the values in datatypes where feasible, instead of always stringifying.

So, for example, it should be possible for an implementation to compile a MF2 into a data structure whereby when it knows that $input is a long, it can convert a literal for matching also into a long in that data structure, and just do long comparison.

Similarly, for plural matching, the compiled data structure could contain an enum value `few', instead of converting the an internal value for the plural category into a string and then doing a string comparison.

macchiati (approved):
My comments should not stand in the way of approval.

spec/registry.md
```diff
@@ -255,3 +255,567 @@ which only expects the `accord` option:
...
+> Users SHOULD encode _messages_ and their parts (such as _keys_ and _operands_),
+> in Unicode Normalization Form C (NFC) unless there is a very good reason
+> not to.
+> See also: [String Matching](https://www.w3.org/TR/charmod-norm)
+
+> [!NOTE]
+> Unquoted string literals in a _variant_ do not include spaces.
+> If users wish to match strings that include whitespace
+> (including U+3000 `IDEOGRAPHIC SPACE`)
+> to a key, the `key` needs to be quoted.
```
mihnita, line 338: Nitpick:
   s / needs to be quoted / MUST be quoted /
?

mihnita (approved):

spec/registry.md
```diff
@@ -255,3 +255,567 @@ which only expects the `accord` option:
...
+      - If `keyword` equals `key`, then `key` matches the selector.
+        Append `key` to the end of the `return_value` list.
+   - Else, `key` is invalid;
+     emit a _Selection Error_.
+     Do not add `key` to `return_value`.
+- Return `return_value`.
+
+#### Plural/Ordinal Keywords
+The _plural/ordinal keywords_ are: `zero`, `one`, `two`, `few`, `many`, and
+`other`.
```
eemeli, line 566: **Blocking**:
Rephrasing this the same way as for `:string` above. One nominal change in behaviour here is the ordering of exact matches, such that earlier exact matches are favoured over later exact matches. This does not have any effect with the current algorithm as that's requiring a very strong equality between `key` and `exact`, but it may have an effect as we consider #675.

Also inlining the section defining keywords, and moving up the note about implementation exactness.

```suggestion
When implementing [`MatchSelectorKeys(resolvedSelector, keys)`](/spec/formatting.md#resolve-preferences)
where `resolvedSelector` is the resolved value of a _selector_ _expression_
and `keys` is a list of strings,
numeric selectors perform as described below.

1. Let `exact` be the JSON string representation of the numeric value of `resolvedSelector`.
   (See [Determining Exact Literal Match](#determining-exact-literal-match) for details)
1. Let `keyword` be a string which is the result of [rule selection](#rule-selection) on `resolvedSelector`.
1. Let `resultExact` be a new empty list of strings.
1. Let `resultKeyword` be a new empty list of strings.
1. For each string `key` in `keys`:
   1. If the value of `key` matches the production `number-literal`, then
      1. If `key` and `exact` consist of the same sequence of Unicode code points, then
         1. Append `key` as the last element of the list `resultExact`.
   1. Else if `key` is one of the keywords `zero`, `one`, `two`, `few`, `many`, or `other`, then
      1. If `key` and `keyword` consist of the same sequence of Unicode code points, then
         1. Append `key` as the last element of the list `resultKeyword`.
   1. Else, emit a _Selection Error_.
1. Return a new list whose elements are the concatenation of the elements (in order) of `resultExact` followed by the elements (in order) of `resultKeyword`.

> [!NOTE]
> Implementations are not required to implement this exactly as written.
> However, the observed behavior must be consistent with what is described here.
```

spec/registry.md
```diff
@@ -255,3 +255,567 @@ which only expects the `accord` option:
...
+
+#### Rule Selection
+
+If the option `select` is set to `exact`, rule-based selection is not used.
+Return the empty string.
+
+> [!NOTE]
+> Since keys cannot be the empty string in a numeric selector, returning the
```
eemeli, line 574:
```suggestion
> Since valid keys cannot be the empty string in a numeric expression, returning the
```

spec/registry.md
```diff
@@ -255,3 +255,567 @@ which only expects the `accord` option:
...
+>
+> A message in Czech might be:
+> ```
+> .match {$numDays :number}
+> one  {{{$numDays} den}}
+> few  {{{$numDays} dny}}
+> many {{{$numDays} dne}}
+> *    {{{$numDays} dní}}
+> ```
+> Using the rules found above, the results of various `operand` values might look like:
```
eemeli, line 601:
```suggestion
> Using the rules found above, the results of various _operand_ values might look like:
```

spec/registry.md
```diff
@@ -255,3 +255,567 @@ which only expects the `accord` option:
...
+> Annotations that use fraction digits or significant digits might work in specific
+> implementation-defined ways.
+> Users should avoid depending on these types of keys in message selection.
+
+
+Number literals in the MessageFormat 2 syntax use the 
+[format defined for a JSON number](https://www.rfc-editor.org/rfc/rfc8259#section-6).
+The resolved value of an `operand` exactly matches a numeric literal `key`
+if, when the `operand` is serialized using the format for a JSON number,
+the two strings are equal.
```
eemeli, line 625:
```suggestion
A `resolvedSelector` exactly matches a numeric literal `key`
if, when the numeric value of `resolvedSelector` is serialized using the format for a JSON number,
the two strings are equal.
```

spec/registry.md
```diff
@@ -255,3 +255,567 @@ which only expects the `accord` option:
...
+
+> [!NOTE]
+> Implementations are not expected to implement this exactly as written,
+> as there are clearly optimizations that can be applied.
+> However, the observed behavior must be consistent with what is described here.
+
```
eemeli, line 631: **Blocking**:
This note should be on the whole selection algorithm, see above.
```suggestion
```

spec/registry.md
```diff
@@ -255,3 +255,567 @@ which only expects the `accord` option:
...
+> [!NOTE]
+> The default formatting behavior of `:datetime` is inconsistent with `Intl.DateTimeFormat`
+> in JavaScript and with `{d,date}` in ICU MessageFormat 1.0.
+> This is because, unlike those implementations, `:datetime` is distinct from `:date` and `:time`.
+
+### Operands
+
+The _operand_ of a date/time function is either 
+an implementation-defined date/time type (passed in as an argument)
```
eemeli, line 664:
```suggestion
an implementation-defined date/time type
```

spec/registry.md
```diff
@@ -255,3 +255,567 @@ which only expects the `accord` option:
...
+> [!NOTE]
+> True time zone support in serializations is expected to coincide with the adoption
+> of Temporal in JavaScript.
+> The form of these serializations is known and is a de facto standard.
+> Support for these extensions is expected to be required in the post-tech preview.
+> See: https://datatracker.ietf.org/doc/draft-ietf-sedate-datetime-extended/
+
+### Options
+
+A function can use either the appropriate _style_ options for that function
```
eemeli, line 707:
```suggestion
A function can use either the appropriate _style options_ for that function
```

spec/registry.md
```diff
@@ -255,3 +255,567 @@ which only expects the `accord` option:
...
+or can use a collection of _field options_ (but not both) to control the formatted 
+output.
+
+If both are specified, an _Invalid Expression_ error MUST be emitted
+and a _fallback value_ used as the resolved value of the _expression_.
+
+#### Style Options
+
+The function `:datetime` has these function-specific _style_ options.
```
eemeli, line 716:
```suggestion
The function `:datetime` has these function-specific _style options_.
```

spec/registry.md
```diff
@@ -255,3 +255,567 @@ which only expects the `accord` option:
...
+  - `long`
+  - `medium`
+  - `short`
+- `timeStyle`
+  - `full`
+  - `long`
+  - `medium`
+  - `short`
+
+The function `:date` has these function-specific _style_ options:
```
eemeli, line 728:
```suggestion
The function `:date` has these function-specific _style options_:
```

spec/registry.md
```diff
@@ -255,3 +255,567 @@ which only expects the `accord` option:
...
+- `style`
+  - `full`
+  - `long`
+  - `medium`
+  - `short` (default)
+
+The function `:time` has these function-specific _style_ options:
```
eemeli, line 735:
```suggestion
The function `:time` has these function-specific _style options_:
```

spec/registry.md
```diff
@@ -255,3 +255,567 @@ which only expects the `accord` option:
...
+Field options describe which fields to include in the formatted output
+and what format to use for that field.
+The implementation may use this _annotation_ to configure which fields
+appear in the formatted output.
+
+> [!NOTE]
+> Field options do not have default values because they are only to be used
+> to compose the formatter.
+
+The _field_ options are defined as follows:
```
eemeli, line 753:
```suggestion
_Field options_ describe which fields to include in the formatted output
and what format to use for that field.
The implementation may use this _annotation_ to configure which fields
appear in the formatted output.

> [!NOTE]
> _Field options_ do not have default values because they are only to be used
> to compose the formatter.

The _field options_ are defined as follows:
```

eemeli (changes requested):
The inline comments marked as **Blocking** will need to be resolved; the others can be considered nitpicks.

spec/registry.md
eemeli: This is a non-normative note, so using MUST would be out of place.

spec/registry.md
aphillips: "style options" is not a term.

```suggestion
The function `:time` has these function-specific _options_ for _style_:
```

spec/registry.md
```diff
@@ -255,3 +255,567 @@ which only expects the `accord` option:
...
+> | hello | `hello` | Unquoted literal |
+> | -13.2 | `-13.2` | Unquoted number-literal |
+> | \| spaces \| | ` spaces ` | Includes one space before and after |
+
+
+> [!NOTE]
+> This should probably include individual character types, such as `char`.
+
+In addition, implementations MAY perform formatting and selection on 
+`operand` values that do not otherwise have a formatting function registered.
```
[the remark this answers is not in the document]
aphillips: The Java note is not right (for example, probably subclasses of `java.lang.CharSequence`). 

In general this is better... and also approximately where we started.

spec/registry.md
aphillips:
```suggestion
The function `:time` has these function-specific _style options_:
```

No, I was wrong.

spec/registry.md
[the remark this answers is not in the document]
eemeli: Very happy to have the Java note corrected. Note when doing so that `java.lang.CharSequence` is an interface rather than a class.

eemeli (approved):

aphillips:
> Just to be clear, I'm still fine with this landing without further discussion in a live call as long as the issues above are resolved; I just don't think it's appropriate to consider this "fast-tracked".

Removing the label technically means holding this for telecon. I'm not going to do that. Note that, as promised in email, this will be followed "immediately" (in some hours, after I have prepared it) with an editorial-only PR (which will also have a `fast-track` label).

aphillips:
Merging this [per my email](https://groups.google.com/a/chromium.org/g/message-format-wg/c/kJAmtx3WbU0) on this topic.

spec/registry.md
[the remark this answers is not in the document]
aphillips: > I would also appreciate a recommendation for NFC normalization specifically for this comparison.

If you mean requiring or suggesting that implementations normalize the key and value when performing the comparison, it is important that implementations **NOT** do that. That would make implementations work differently depending on whether they normalize or not.

aphillips:
@sffc wrote (in part):
> ... and decided to cut number formatting options that carry semantic meaning, such as currency and unit, until we can discuss their design further.

Please notice #664

PR #660, opened 2024-02-16
Define `:string`

aphillips:
Flesh out `:string` using the format we used for dates and numbers.

### DIFF
--- a/exploration/string-selection-formatting.md
+++ b/exploration/string-selection-formatting.md
@@ -16,7 +16,7 @@ Status: **Proposed**
 
 ## Objective
 
-Define how selection and formatting of string literals takes place.
+Define how selection and formatting of string values takes place.
 
 ## Background
 
--- a/exploration/string-selection-formatting.md
+++ b/exploration/string-selection-formatting.md
@@ -61,8 +61,7 @@ character-sequence types.
 > This should probably include individual character types, such as `char`.
 
 In addition, implementations MAY perform formatting and selection on 
-`operand` values that do not otherwise have a formatting function registered, 
-although emitting an `Unsupported Expression` error is RECOMMENDED.
+`operand` values that do not otherwise have a formatting function registered.
 
 ### Options
 
--- a/exploration/string-selection-formatting.md
+++ b/exploration/string-selection-formatting.md
@@ -70,7 +70,7 @@ The function `:string` has no options.
 
 > [!NOTE]
 > Proposals for string transformation options or implementation
-> experience with user requirements is desired in the Technical Preview.
+> experience with user requirements is desired during the Tech Preview.
 
 ### Selection
 
--- a/exploration/string-selection-formatting.md
+++ b/exploration/string-selection-formatting.md
@@ -98,15 +98,16 @@ the `:string` selector performs as described below.
 
 > [!NOTE]
 > Unquoted string literals in a _variant_ do not include spaces.
-> If users wish to match strings that begin or end with spaces
+> If users wish to match strings that include whitespace
 > (including U+3000 `IDEOGRAPHIC SPACE`)
-> to a key, the `key` needs to have the spaces quoted.
-> > For example:
-> >```
-> > .match {$string :string}
-> > | space key | {{Matches the string " space key "}}
-> > *             {{Matches the string "space key"}}
-> >```
+> to a key, the `key` needs to be quoted.
+>
+> For example:
+>```
+> .match {$string :string}
+> | space key | {{Matches the string " space key "}}
+> *             {{Matches the string "space key"}}
+>```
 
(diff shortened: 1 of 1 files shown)


macchiati (approved):

exploration/string-selection-formatting.md
```diff
@@ -0,0 +1,114 @@
...
+		<dt>First proposed</dt>
+		<dd>2024-02-16</dd>
+		<dt>Pull Request</dt>
+		<dd><a href="https://github.com/unicode-org/message-format-wg/pull/660">#660</a></dd>
+	</dl>
+</details>
+
+## Objective
+
+Define how selection and formatting of string literals takes place.
```
eemeli, line 19: Not just literals, also variables.
```suggestion
Define how selection and formatting of string values takes place.
```

exploration/string-selection-formatting.md
```diff
@@ -0,0 +1,114 @@
...
+
+The _operand_ is any literal or an implementation-defined set of string or
+character-sequence types.
+
+> [!NOTE]
+> This should probably include individual character types, such as `char`.
+
+In addition, implementations MAY perform formatting and selection on 
+`operand` values that do not otherwise have a formatting function registered, 
+although emitting an `Unsupported Expression` error is RECOMMENDED.
```
eemeli, line 65: I don't think we should be recommending errors. At least in the JS Intl.MessageFormat proposal, `:string` will likely work for nearly all values, and only fail for objects that actively prevent their own stringification by e.g. throwing an error in their `toString()` method.

If a user explicitly asks for a `:string`, they should get that if possible.

```suggestion
In addition, implementations MAY perform formatting and selection on 
`operand` values that do not otherwise have a formatting function registered.
```

exploration/string-selection-formatting.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Options
+
+The function `:string` has no options.
+
+> [!NOTE]
+> Proposals for string transformation options or implementation
+> experience with user requirements is desired in the Technical Preview.
```
eemeli, line 73: We should be consistent in our naming, and also match how CLDR refers to this phase as a "tech preview".

```suggestion
> experience with user requirements is desired during the Tech Preview.
```

exploration/string-selection-formatting.md
```diff
@@ -0,0 +1,114 @@
...
+     A `key` and an `operand` are equal if they consist of the same
+     sequence of Unicode code points.
+     Add `key` to the front of the `return_value` list.
+- Return `return_value`
+
+> [!NOTE]
+> Matching of `key` and `operand` values is sensitive to the sequence of code points
+> in each string.
+> As a result, variations in how text can be encoded can affect the performance of matching.
+> The function `:string` does not perform case folding or Unicode Normalization of string values.
```
eemeli, line 98: Should we, though? The operand and the key are likely to be coming from rather distinct sources, and a mismatch here could be really hard to debug given that the strings may _look_ exactly the same.

exploration/string-selection-formatting.md
```diff
@@ -0,0 +1,114 @@
...
+> Unquoted string literals in a _variant_ do not include spaces.
+> If users wish to match strings that begin or end with spaces
+> (including U+3000 `IDEOGRAPHIC SPACE`)
+> to a key, the `key` needs to have the spaces quoted.
+> > For example:
+> >```
+> > .match {$string :string}
+> > | space key | {{Matches the string " space key "}}
+> > *             {{Matches the string "space key"}}
+> >```
```
eemeli, line 110: Keys with _any_ whitespace need to be quoted.

```suggestion
> If users wish to match strings that include whitespace
> (including U+3000 `IDEOGRAPHIC SPACE`)
> to a key, the `key` needs to be quoted.
>
> For example:
>```
> .match {$string :string}
> | space key | {{Matches the string " space key "}}
> *             {{Matches the string "space key"}}
>```
```

eemeli (changes requested):

exploration/string-selection-formatting.md
aphillips: I have a long document about this: [Charmod: String Matching](https://www.w3.org/TR/charmod-norm/).

If we want to offer case folding, normalization, or other text transformation (such as trimming), we should add options to control it. The default really should be code point sequences. Sometimes those variations in encoding are intended to be significant. I actually think that `:string` options will be a useful and popular add-on.... sometime in the fall...

If one wants to write a `SelectFormat`, it's a Good Idea to stick to ASCII tokens, preferably non-language tokens. `SHOUTED_SNAKE_CASE` recommends itself 😉. But that's for the user guide, not here.

eemeli (approved):

PR #666, opened 2024-02-18
Address #619: add a note about `*` variant

### LINKED ISSUE #619: Add warning about `*`
In the 2024-01-29 call we agreed that we would continue to require the `*` variant for all selectors.

There remains concern that message authors will find this awkward in cases where a selector can be exhausted without the default fallback value. We agreed to seek feedback during tech preview and to add a note/warning to the spec about it. This issue tracks the need for a PR.

aphillips:
Per my action-item, add an appropriate note about the requirement for the fallback variant, including text seeking feedback in tech preview.

Fixes #619

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -375,9 +375,21 @@ according to their _key_ values and selecting the first one.
 > [!NOTE]
 > At least one _variant_ is required to have all of its _keys_ consist of
 > the fallback value `*`.
-> Some _selectors_ might be implemented in a way that this _variant_
+> Some _selectors_ might be implemented in a way that the key value `*`
 > cannot be selected in a _valid_ _message_.
-> In other cases, this _variant_ might be unreachable only in certain locales.
+> In other cases, this key value might be unreachable only in certain locales.
+> This could result in the need in some locales to create
+> one or more _variants_ that do not make sense grammatically for that language.
+> > For example, in the `pl` (Polish) locale, this _message_ cannot reach
+> > the `*` _variant_:
+> > ```
+> > .match {$num :integer}
+> > 0    {{ }}
+> > one  {{ }}
+> > few  {{ }}
+> > many {{ }}
+> > *    {{Only used by fractions in Polish.}}
+> > ```
 >
 > In the Tech Preview, feedback from users and implementers is desired about
 > whether to relax the requirement that such a "fallback _variant_" appear in
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -372,6 +372,18 @@ to provide the _pattern_ for the formatting operation.
 This is done by ordering and filtering the available _variant_ statements
 according to their _key_ values and selecting the first one.
 
+> [!NOTE]
+> At least one _variant_ is required to have all of its _keys_ consist of
+> the fallback value `*`.
+> Some _selectors_ might be implemented in a way that this _variant_
+> cannot be selected in a _valid_ _message_.
+> In other cases, this _variant_ might be unreachable only in certain locales.
+>
+> In the Tech Preview, feedback from users and implementers is desired about
+> whether to relax the requirement that such a "fallback _variant_" appear in
+> every message, versus the potential for a _message_ to fail at runtime
+> because no matching _variant_ is available.
+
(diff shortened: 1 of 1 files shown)


mihnita (approved):

spec/formatting.md
```diff
@@ -372,6 +372,18 @@ to provide the _pattern_ for the formatting operation.
 This is done by ordering and filtering the available _variant_ statements
 according to their _key_ values and selecting the first one.
 
+> [!NOTE]
+> At least one _variant_ is required to have all of its _keys_ consist of
+> the fallback value `*`.
+> Some _selectors_ might be implemented in a way that this _variant_
+> cannot be selected in a _valid_ _message_.
+> In other cases, this _variant_ might be unreachable only in certain locales.
```
stasm, line 380: I'm not sure the note about locales is relevant here? How about:

```suggestion
> In other cases, this _variant_ may be interpreted as a valid catch-all variant, rather than a fallback in case of failed selection.
```

stasm (approved):

eemeli (approved):

spec/formatting.md
```diff
@@ -372,6 +372,30 @@ to provide the _pattern_ for the formatting operation.
...
+> > For example, in the `pl` (Polish) locale, this _message_ cannot reach
+> > the `*` _variant_:
+> > ```
+> > .match {$num :integer}
+> > 0    {{ }}
+> > one  {{ }}
+> > few  {{ }}
+> > many {{ }}
+> > *    {{Only used by fractions in Polish.}}
+> > ```
```
eemeli, line 392: Would this be a simpler example for what's being demostrated here?

```suggestion
> > For example, in the `en` (English) locale, this _message_ cannot reach
> > the `*` _variant_ unless `$num` is not a valid number:
> > ```
> > .match {$num :number}
> > one   {{ }}
> > other {{ }}
> > *     {{ }}
> > ```
```

aphillips: Do we really want to promote the use of `other` with the number functions? In English? I thought the Polish example was clever because it naturally exposes the issue at hand (and begs the question "what does that message say?"). But happy to accept this too...

eemeli: As I've understood, in Polish `many` _is_ the fallback category, so explicitly including it (as in your example) is not really any different from my proposal.

spec/formatting.md
```diff
@@ -372,6 +372,30 @@ to provide the _pattern_ for the formatting operation.
 This is done by ordering and filtering the available _variant_ statements
 according to their _key_ values and selecting the first one.
 
+> [!NOTE]
+> At least one _variant_ is required to have all of its _keys_ consist of
+> the fallback value `*`.
+> Some _selectors_ might be implemented in a way that the key value `*`
+> cannot be selected in a _valid_ _message_.
```
stasm, line 379: How about we give an example of such a selector? 

```suggestion
> Some _selectors_ might be implemented in a way that the key value `*`
> cannot be selected in a _valid_ _message_.
> 
> ```
> .match {$result :boolean}
> true {{Result is true}}
> false {{Result is false}}
> * {{Unreachable}}
> ```
```

I think this would make it easier to understand why the next sentence starts with "in other cases... only in certain locales".

spec/formatting.md
```diff
@@ -372,6 +372,30 @@ to provide the _pattern_ for the formatting operation.
 This is done by ordering and filtering the available _variant_ statements
 according to their _key_ values and selecting the first one.
 
+> [!NOTE]
+> At least one _variant_ is required to have all of its _keys_ consist of
+> the fallback value `*`.
```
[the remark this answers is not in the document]
aphillips: Thanks. The phrasing is a bit chewy. I didn't want to create a new term like "exhaustive variant" just for the note.

spec/formatting.md
aphillips: I'd like to avoid using functions that don't exist 😉 

```
.match {$boolean :string}
true  {{Result is true}}
false {{Result is false}}
*     {{Unreachable?}}
```

aphillips:
I'm going to merge this and we can mess with it more later. It's a non-normative note and we need to focus on the normative stuff.

PR #668, opened 2024-02-18
Address #575: say that data model uses `name` for `identifier`

### LINKED ISSUE #575: The data model uses "name" for both namespaced and non-namespaced names
We currently have https://github.com/unicode-org/message-format-wg/blob/2a69d2a17ee7476282bfa824fd4ca62e0a03f6e1/spec/data-model/README.md?plain=1#L168-L171
but also https://github.com/unicode-org/message-format-wg/blob/2a69d2a17ee7476282bfa824fd4ca62e0a03f6e1/spec/data-model/README.md?plain=1#L183-L193

In the former, the "name" is a non-namespaced `name`, while in the latter it's an `identifier`, which may also include a namespace.

Which of the following should we do?
1. Change the function & option "names" to be called something else.
2. Add the namespace to the data model as a separate field.
3. Note in the text that these have different contents.
4. Something else?

This issue was originally raised  by @aphillips in https://github.com/unicode-org/message-format-wg/pull/574#discussion_r1434369952, as markup will also be affected.

aphillips:
Per our discussion of #575, we want to say that the field `name` in the data model encompasses the production `identifier` in cases where the name is an identifier.

Fixes #575.

### DIFF
--- a/spec/data-model/README.md
+++ b/spec/data-model/README.md
@@ -48,6 +48,16 @@ declarations, options, and attributes to be optional rather than required proper
 
 ## Messages
 
+> [!IMPORTANT]
+> The data model uses the field name `name` to denote various interface identifiers.
+> In the MessageFormat 2 syntax, the source for these `name` fields sometimes uses
+> the production `identifier`.
+> This happens when the named item, such as a _function_, supports namespacing.
+>
+> In the Tech Preview, feedback on whether to separate the `namespace` from the `name`
+> and represent both separately, or just, as here, use an opaque single field `name`
+> is desired.
+
 A `SelectMessage` corresponds to a syntax message that includes _selectors_.
 A message without _selectors_ and with a single _pattern_ is represented by a `PatternMessage`.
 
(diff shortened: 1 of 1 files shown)


spec/data-model/README.md
```diff
@@ -48,6 +48,16 @@ declarations, options, and attributes to be optional rather than required proper
 
 ## Messages
```
eemeli, line 50: The note should probably be at the end of the intro, rather than within the "Messages" section.
```suggestion
```

spec/data-model/README.md
```diff
@@ -48,6 +48,16 @@ declarations, options, and attributes to be optional rather than required proper
...
+> [!IMPORTANT]
+> The data model uses the field name `name` to denote various interface identifiers.
+> In the MessageFormat 2 syntax, the source for these `name` fields sometimes uses
+> the production `identifier`.
+> This happens when the named item, such as a _function_, supports namespacing.
+>
+> In the Tech Preview, feedback on whether to separate the `namespace` from the `name`
+> and represent both separately, or just, as here, use an opaque single field `name`
+> is desired.
+
```
eemeli, line 60:
```suggestion

## Messages

```

eemeli (approved):
See slight reordering inline.

aphillips:
@stasm Can I get a ship-it?

PR #672, opened 2024-02-19
Drop "quoted" remnants from JSON Schema & DTD data models

### LINKED ISSUE #667: [FEEDBACK] The Literal 'quoted' property is obsolete
In the data model descriptions, there is a contradiction whether the `Literal` object contains a `quoted` property or not:

* In the JSON schema https://github.com/unicode-org/message-format-wg/blob/main/spec/data-model/message.json and the XML DTD https://github.com/unicode-org/message-format-wg/blob/main/spec/data-model/message.dtd a `quoted` property is mentioned.
* In the data model description https://github.com/unicode-org/message-format-wg/blob/main/spec/data-model/README.md it is not mentioned.

Because of the statement "Implementations MUST NOT distinguish between quoted and unquoted literals that have the same sequence of code points." in https://github.com/unicode-org/message-format-wg/blob/main/spec/syntax.md, and because of the purpose of the data model is "to allow interchange of the logical representation", I believe this property should be removed. That is, the `value` property is the unquoted literal.

eemeli:
Fixes #667, CC @bhaible.

This was missed out in #443.

### DIFF
--- a/spec/data-model/message.dtd
+++ b/spec/data-model/message.dtd
@@ -29,7 +29,6 @@
 )>
 
 <!ELEMENT literal (#PCDATA)>
-<!ATTLIST literal quoted (true | false) #REQUIRED>
 
 <!ELEMENT variable (EMPTY)>
 <!ATTLIST variable name NMTOKEN #REQUIRED>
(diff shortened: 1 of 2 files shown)


bhaible:
Thanks. Looks correct.

aphillips (approved):

PR #673, opened 2024-02-19
Fix whitespace conformance to match UAX31 (including permitting LRM/RLM)

aphillips:
This partially addresses #661 by allowing the LRM character in message whitespace. This is whitespace **_outside_** pattern `text`.
Tools can use this to help ensure that messages are formatted visually in a way consistent with LTR presentation of a message.


aphillips:
@eggrobin For review

spec/syntax.md
```diff
@@ -906,16 +906,28 @@ backslash       = %x5C ; U+005C REVERSE SOLIDUS "\"
...
+The character `U+200E LEFT-TO-RIGHT MARK` (LRM) MAY be prepended to _whitespace_ outside
+_patterns_ and _quoted literals_ to assist with presentation to users.
+Tools SHOULD generate these LRM characters following _identifiers_, _unquoted literals_, or 
+_option_ values that use right-to-left characters so that the _message_ displays
+intelligibly in a left-to-right context.
+
+This definition of _whitespace_ implements 
+[UTR#31 Rule 3a-2](https://www.unicode.org/reports/tr31/#R3a-2).
+It is a profile of R3a-1 in that specification because only the
+whitespace characters listed are permitted as whitespace.
```
eggrobin, line 929: UAX31-R2a-2 says you need to
> define that profile with a precise specification of the characters that are added to or removed from the set of code points defined by the Pattern_White_Space property, and of any changes to the criteria under which a character or sequence of characters is interpreted as an end of line, as ignorable format controls, or as horizontal space.

The point is that the reader of such a conformance statement sees the difference from the default, which are the things that may need special attention when interoperating with an implementation based on the defaults.

So, if I am reading this right:
- Form feed, next line, line separator, paragraph separator, and right-to-left mark are removed from the set of white space characters;
- ideographic space is added to the set of white space characters;
- the sole ignorable format control (LRM) is allowed only in contexts UAX31-I1 and UAX31-I3 (not UAX31-I2), where UAX31-I1 is further restricted to the beginning of a sequence of horizontal spaces, and UAX31-I3 is further restricted to only the end and not the beginning of a line.

eggrobin: > further restricted to only the end and not the beginning of a line

Also it is disallowed on a blank line, I think?

eggrobin: > Note that quoted literals or pattern text can contain bidi controls that might cause Trojan source effects unless/until we address placeholder isolation.

Mostly that is something to be dealt with at a higher level (in editors and tooling); the main thing languages need to do is to treat the right things in the right way so that standardized tooling can deal with the issue, see UTS55.

> Note that UAX31-I2 is not permitted because that would break our sigil-identifier syntax.

I don’t really understand what you mean here; UAX31-I2 does not say you should allow `$⟨LRM⟩identifier`, it says that you should allow it _if_ you allow `$    identifier`.

The reason why you do not have UAX31-I2 is that you do not allow an LRM wherever you have `[s]`, so for instance in 

https://github.com/unicode-org/message-format-wg/blob/ae712d70d60e9b5d06bc72d5c7788811aa728f8a/spec/message.abnf#L35-L36

you do not allow `{⟨LRM⟩#whatever⟨LRM⟩}`, only `{⟨LRM⟩ #whatever⟨LRM⟩ }`, right?

spec/message.abnf
```diff
@@ -104,5 +104,5 @@ quoted-escape   = backslash ( backslash / "|" )
 reserved-escape = backslash ( backslash / "{" / "|" / "}" )
 backslash       = %x5C ; U+005C REVERSE SOLIDUS "\"
 
-; Whitespace
-s = 1*( SP / HTAB / CR / LF / %x3000 )
+; Whitespace, optionally prepended with LRM
+s = [%x200E] 1*( SP / HTAB / CR / LF / %x3000 )
```
gibson042, line 108: This means that LRM is _not_ valid after a line terminator, e.g. before the quoted patterns in a message like
```
.match {$count :number}
one
	{{You have {$count} notification.}}
*
	{{You have {$count} notifications.}}
```

Is that intentional?

spec/syntax.md
eggrobin: > you do not allow {⟨LRM⟩#whatever⟨LRM⟩}, only {⟨LRM⟩ #whatever⟨LRM⟩ }, right?

Note that this probably throws a wrench into the UTS55 conversion to plain text algorithm.

spec/message.abnf
aphillips: No, not really.

Probably what needs to happen here is a distinguishing of optional and non-optional whitespace. Everywhere we have `[s]` should use a production that can be just LRM (or RLM, fwiw) or nothing, e.g.:

```abnf
; optional whitespace
owsp = *( SP/ HTAB / CR / LF / %x3000 / %x200E / %x200F )
```

And everywhere that requires positive whitespace (i.e. just `s`) permit controls to either side:

```abnf
; required whitespace
wsp = [ (%x200E / %x200F) ] 1*( SP / HTAB / CR / LF / %x3000) [ (%x200E / %x200F) ]
```

eggrobin: I think that would be an improvement (and then you can drop the convoluted explanation from the conformance statement, and the profile limits itself to changing the sets of characters).

spec/syntax.md
```diff
@@ -906,23 +906,44 @@ backslash       = %x5C ; U+005C REVERSE SOLIDUS "\"
...
+[UTR#31 Rule 3a-2](https://www.unicode.org/reports/tr31/#R3a-2).
+It is a profile of R3a-1 in that specification because:
+the following pattern whitespace characters are not allowed:
+`U+000B LINE FEED`, 
+`U+000C VERTICAL TABULATION`, 
+`U+0085 NEXT LINE`, 
+`U+2028 LINE SEPARATOR` and 
+`U+2029 PARAGRAPH SEPARATOR`;
+the character `U+3000 IDEOGRAPHIC SPACE`
+_is_ included in pattern whitespace;
```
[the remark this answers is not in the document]
aphillips: I know, which is why I didn't say Pattern_White_Space. "interpreted as" is somehow not quite right in our space. But we both have other comments below that probably make this line moot.

spec/syntax.md
aphillips: Yes. See my response to @gibson042's comment.

spec/syntax.md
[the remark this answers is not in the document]
aphillips: ... not to mention that we overload the term "pattern"...

spec/syntax.md
```diff
@@ -905,24 +905,55 @@ backslash       = %x5C ; U+005C REVERSE SOLIDUS "\"
...
+It is a profile of R3a-1 in that specification because:
+the following pattern whitespace characters are not allowed:
+`U+000B FORM FEED`, 
+`U+000C VERTICAL TABULATION`, 
+`U+0085 NEXT LINE`, 
+`U+2028 LINE SEPARATOR` and 
+`U+2029 PARAGRAPH SEPARATOR`;
+the character `U+3000 IDEOGRAPHIC SPACE`
+_is_ interpreted as whitespace;
+ and the following character is not included in ignorable format controls:
```
eggrobin, line 940:
```suggestion
 and the following character is not interpreted as whitespace (in particular, it is not treated as an ignorable format control):
```
This is admittedly pedantry (but what did you expect from a standardizer?), but given the way R3a is written I could technically read « the following character is not included in ignorable format controls » as « that character is whitespace, but not treated as an ignorable format control, and thus is horizontal space ».

spec/syntax.md
```diff
@@ -905,24 +905,55 @@ backslash       = %x5C ; U+005C REVERSE SOLIDUS "\"
...
+but which users might want to include to increase the readability of a _message_.
+**_<dfn>Required whitespace</dfn>_** is whitespace that is required by the syntax.
+
+Tools SHOULD generate `U+200E LEFT-TO-RIGHT MARK` or `U+200F RIGHT-TO-LEFT MARK` 
+characters where permitted by the syntax before or following _identifiers_,
+_unquoted literals_, or _option_ values that use right-to-left characters 
+so that the _message_ displays intelligibly.
+
+These definitions of _whitespace_ implement
+[UTR#31 Rule 3a-2](https://www.unicode.org/reports/tr31/#R3a-2).
```
eggrobin, line 930:
```suggestion
[UTR#31 Requirement R3a-2](https://www.unicode.org/reports/tr31/#R3a-2).
```

eggrobin (approved):
The conformance statement seems good now.
There may be a discussion to be had as to whether the profile should be made smaller (in particular in the line break area discrepancies could lead to spoofing concerns), but this is probably not urgent.

In any case this clear conformance statement should facilitate CLDR-TC’s job when it comes to review—at least I should hope it will help the chair, who is my co-editor on UAX&nbsp;#&zwnj;31 :-)

spec/syntax.md
eggrobin: On the classification of UAX, UTS (synchronized and non-synchronized), and UTR, see https://www.unicode.org/reports/about-reports.html.

(And yes, that page does say that a Unicode Technical Report (UTR) is a specific type of Unicode Technical Report. 🤪)

spec/syntax.md
aphillips: Matching suggestion to our formatting.

```suggestion
and the following character is not interpreted as whitespace
(in particular, it is not treated as an ignorable format control):
```

spec/syntax.md
```diff
@@ -905,24 +905,56 @@ backslash       = %x5C ; U+005C REVERSE SOLIDUS "\"
...
+`U+000B FORM FEED`, 
+`U+000C VERTICAL TABULATION`, 
+`U+0085 NEXT LINE`, 
+`U+2028 LINE SEPARATOR` and 
+`U+2029 PARAGRAPH SEPARATOR`;
+the character `U+3000 IDEOGRAPHIC SPACE`
+_is_ interpreted as whitespace;
+and the following character is not interpreted as whitespace
+(in particular, it is not treated as an ignorable format control):
+`U+200F RIGHT-TO-LEFT MARK`.
```
aphillips, line 942: Since I added U+200F, this text no longer applies.

```suggestion
_is_ interpreted as whitespace.
```

eggrobin: Oh, right.

spec/message.abnf
aphillips: I did the hard change. We now need to get WG approval.

spec/syntax.md
```diff
@@ -905,24 +905,53 @@ backslash       = %x5C ; U+005C REVERSE SOLIDUS "\"
...
 Whitespace is not significant outside translatable text, except where required by the syntax.
 
+There are two whitespace productions in the syntax.
+**_<dfn>Optional whitespace</dfn>_** is whitespace that is not required by the syntax, 
+but which users might want to include to increase the readability of a _message_.
+**_<dfn>Required whitespace</dfn>_** is whitespace that is required by the syntax.
+
+Tools SHOULD generate `U+200E LEFT-TO-RIGHT MARK` or `U+200F RIGHT-TO-LEFT MARK` 
+characters where permitted by the syntax before or following _identifiers_,
+_unquoted literals_, or _option_ values that use right-to-left characters
```
macchiati, line 926: I don't see why there is the restriction on **unquoted** literals. Shouldn't it be any literal? That is, anywhere a unquoted literal can appear, and unquoted one can. So it seems like either both (aka just 'literal') or neither can appear.

I could match on:
```
X y ⎨⎨$count⎬⎬
```
Where X is a RTL character, or on
```
⎸X⎸ y ⎨⎨$count⎬⎬
```

In both cases the result is jumbled (disregard the direction of the fake braces, the tool just reorders.
```
⎬	⎬	y		⎨	⎨	$	c	o	u	n	t		X

⎬	⎬	y		⎨	⎨	$	c	o	u	n	t		⎸	X	⎸
```
Now, in this case I could put LRMs before the first ⎸ and after the second, because both positions allow whitespace. Are there circumstances around literals where it makes a difference in the insertability of LRM/RLM because of the quoting?

macchiati: Other than that, the changes look good to me. Note that if there are any issues in the WG about this, we refrain from these changes until after the v45 release, just leaving a note that we're looking at the bidi ordering issues...

aphillips: You're right. The key thing I think is to remind tool writers not to quote the mark onto the value.

aphillips: > Note that if there are any issues in the WG about this, we refrain from these changes until after the v45 release, just leaving a note that we're looking at the bidi ordering issues...

The changes are to the syntax and I think important enough to merit doing the change now--the better to stabilize the syntax. It does represent a relaxation of what is allowed in free whitespace. I would like to avoid having a lot of Tech Preview implementations reject bidi-friendlier messages in the fall. 

OTOH, it does represent a departure from how we set up the `s` production.

macchiati: Right, for a LTR reading, you want to put LRM around any 'element' that could contain RTL characters. Each literal being matched, literals in option values, etc. So in 

```
{{STUFF {$value option=|JUNK| ...} TO READ}}
```
You want to insert like:
```
{{<LRM>STUFF {<LRM>$value option=<LRM>|JUNK|<LRM> ...<LRM>} TO READ<LRM>}}
```

Of course:
- With LRM/RLM you can't the the RTL message parts to reorder around the {$value}, but the order is predictable and far better than the raw message.
- in practice you want **tooling** to do this, not humans.

macchiati:
While a departure, I think it is cleaner than before....

spec/syntax.md
aphillips: > `{{<LRM>STUFF {<LRM>$value option=<LRM>|JUNK|<LRM> ...<LRM>} TO READ<LRM>}}`

You definitely do **not** want the LRMs in the _text_ part of the message, which is where the ones after the `{{` and before the `}}` are. Instead there should be an LRM following the pattern so that any keys in the next variant aren't reordered:

```
<LRM>KEY1<LRM> KEY2<LRM> {{STUFF {<LRM>$value option=<LRM>|JUNK|<LRM>...<LRM>} TO READ}}
<LRM>key1<LRM> key2<LRM> {{ ... next variant...}}
```

Agreed that this is a job for tools.

macchiati: Whoops, yes, imediately outside {{ and }}, not inside.

eemeli:
I'm a little puzzled about the explicit choice made here of only allowing RLM and LRM, as opposed to other directional formatting characters. Why is that preferable here to LRI/RLI/FSI/PDI, which we're exclusively using in our [default bidi isolation strategy?](https://github.com/unicode-org/message-format-wg/blob/main/spec/formatting.md#handling-bidirectional-text)?

I've not spent very long (yet...) with bidi concerns, but one aspect that I'm concerned with is the implementation and understanding of the recommendations added by this change. To me, isolates seems more in line with the shape of MF2 syntax and its nestings of code and message text. As a programmer, they're also somewhat easier to reason about as their effects are more direct and, well, isolated.

I'm also a bit concerned about the effects that our allowance for RTL names and identifiers may have, esp. when they are mixed in with LTR names and identifiers, and our general allowance for newlines in whitespace.

Rather than allowing directional formatting characters in any/all whitespace, could we be much more restrictive about which characters we allow, and where? And also perhaps include some explicit guidance about the preferred rendering order for syntax, such as the LTR order within expressions of operand > function name > options > attributes?

For instance, would it be appropriate to only allow the following?
1. LRM after a syntax whitespace newline
2. LRI/RLI/FSI before the `quoted-pattern` start `{{`
3. PDI after the `quoted-pattern` end`}}`
4. LRI after the `expression` or `markup` start `{`
5. PDI before the `expression` or `markup` end `}`
6. LRI/RLI/FSI before `variable` or `literal`
7. LRI/RLI/FSI before the sigil (if any) prefixing an `identifier`
8. PDI after `variable`, `literal`, or `identifier`

The intent with the above would be to ensure that it's possible to have a valid message for which the "code" portions always have an LTR paragraph direction, while allowing for all user-customizable strings to define their own direction.

aphillips:
@eemeli That's a good point.

Your list would cover the isolation requirements, albeit being somewhat hard to specify. I would permit the LRM solution also (it's simpler if one is prudent about using RTL tokens in placeholders or keys).

It's a bit hard to specify in the grammar. I think we could add isolates to the various placeholder quotes (`{`/`}`) and to pattern quotes (`{{`/`}}`) and probably should do.

One part of our syntax that makes me nervous is the key array. Key values can be literals and these are only separated by whitespace. RLI/PDI around keys (or LRM after them) keeps this from happening (numbers added to show the _logical_ order of the keys):

image

It also guards against this kind of "exchange" (the message actually has keys `|1| |2 3|`):

image

eggrobin:
> It's a bit hard to specify in the grammar.

Generally that is the thrust of the recommendations of UTS55, reflecting the consensus arising from a year and a half of discussion with security experts and implementers at various levels from compilers to editors in the in the Source Code Working Group[^1]: syntaxes should allow the characters that are needed to fix things, and higher-level tooling should deal with inserting that in the right places, ensuring the stateful controls are correctly paired, etc.

Again I can see how it would make sense to add the isolates to ignorable format controls in this context; but trying to restrict it to the right places is messy to deal with in the grammar, because you have to ensure that they are properly terminated (and note of course that stateful controls that are in literal text would interact with those that you allow in whitespace).

[^1]: The SCWG was a limited-duration working group of the Properties and Algorithms Group of the Unicode Technical Committee, see [UTC-170-C2](https://www.unicode.org/L2/L2022/22016.htm#170-C2).

macchiati:
It is quite tricky, and we should not derail the tech preview release for
this. That's why I (more strongly) urge that we capture this issue in a
note in the spec for tech preview, and make the fix afterwards.

aphillips:
A short term fix would be to permit the isolates *around* whitespace (without attempting to forcibly pair them) and to permit (correct side) decoration of the placeholder and pattern delimiters with LRI/PDI. I'm noodling around with it in another browser tab before proposing it. Otherwise, yes, this should be a Tech Preview item with a note in the specification. I will add the note to this PR also.

spec/syntax.md
```diff
@@ -905,24 +926,64 @@ backslash       = %x5C ; U+005C REVERSE SOLIDUS "\"
...
 ```abnf
-s = 1*( SP / HTAB / CR / LF / %x3000 )
+; optional whitespace
+owsp = *( s / %x200E / %x200F )
+
+; required whitespace
+wsp = [ (%x200E / %x200F) ] 1*s [ (%x200E / %x200F) ]
+
+; whitespace characters
+s = ( SP / HTAB / CR / LF / %x3000 )
```
aphillips, line 986: Update to match:

```suggestion
; optional whitespace
owsp = *( s / %x200E / %x200F / %x2066-2069 )

; required whitespace
wsp = [ (%x200E / %x200F / %x2066-2069 ) ] 1*s [ (%x200E / %x200F / %x2066-2069 ) ]

; whitespace characters
s = ( SP / HTAB / CR / LF / %x3000 )
```

spec/syntax.md
```diff
@@ -905,24 +926,64 @@ backslash       = %x5C ; U+005C REVERSE SOLIDUS "\"
...
+[UAX#31 Requirement R3a-2](https://www.unicode.org/reports/tr31/#R3a-2).
+It is a profile of R3a-1 in that specification because:
+the following pattern whitespace characters are not allowed:
+`U+000B FORM FEED`, 
+`U+000C VERTICAL TABULATION`, 
+`U+0085 NEXT LINE`, 
+`U+2028 LINE SEPARATOR` and 
+`U+2029 PARAGRAPH SEPARATOR`;
+the character `U+3000 IDEOGRAPHIC SPACE`
+_is_ interpreted as whitespace.
```
eggrobin, line 971:
```suggestion
the character `U+3000 IDEOGRAPHIC SPACE`
_is_ interpreted as whitespace,
as well as the directional isolates U+2066..U+2069
(these having the Default_Ignorable_Code_Point property, they are ignorable format controls).
```

eggrobin (approved):

spec/syntax.md
aphillips: Is that true, though? I added them directly to `owsp`, meaning they can be used by themselves in optional whitespace locations (without having to insert any actual whitespace), while `wsp` requires that these characters (optionally) appear outside (before/after) a required run of whitespace. We don't impute any "whitespace-ness" to the controls. They're just doing their default-ignorable jobs 😉

eggrobin: Yes, but being an ignorable format control (which does not count as a required space, and is only allowed as an optional space) counts as one of the three ways in which something can be _interpreted as whitespace_ in the sense of [UAX31-R3a](https://www.unicode.org/reports/tr31/#R3a) (the other two being horizontal space, and line terminator or part of a line terminator).

spec/syntax.md
```diff
@@ -905,24 +926,64 @@ backslash       = %x5C ; U+005C REVERSE SOLIDUS "\"
...
 Inside _patterns_ and _quoted literals_,
 whitespace is part of the content and is recorded and stored verbatim.
 Whitespace is not significant outside translatable text, except where required by the syntax.
 
+There are two whitespace productions in the syntax.
+**_<dfn>Optional whitespace</dfn>_** is whitespace that is not required by the syntax, 
+but which users might want to include to increase the readability of a _message_.
+**_<dfn>Required whitespace</dfn>_** is whitespace that is required by the syntax.
+
+Tools or users SHOULD insert paired isolating controls `U+2066 LEFT-TO-RIGHT ISOLATE`
```
macchiati, line 945: - The SHOULD should be limited to those pieces that contain RTL text.
- The SHOULD should be limited to "tooling".
I think "or users" sets unrealistic expectations; manual insertion can be tricky, and runs the risk with isolates of having unpaired bidi controls.
Maybe instead add at the end a note that users may also manually insert these characters, 
but tooling can make sure that the bidi controls are in the right positions,
and that all and only the bidi controls necessary are included.
- The two sections about inserting isolates and inserting marked should be clearly separated.

macchiati (changes requested):

spec/syntax.md
eggrobin: (Yes, this is terribly confusing, and part of the general pattern that in Unicode contexts _whitespace_ means a host of different things, from [characters that are whitespace](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3Awhitespace%3A%5D&g=&i=) to [characters that should be interpreted as whitespace in parsing](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3Apattern_white_space%3A%5D&g=&i=) to [characters that are categorized as spaces](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3Aspace+separator%3A%5D&g=&i=) to the zero width space and no-break space that are called spaces, but are not spaces at all and instead are controls serving as overrides for some segmentation algorithms.

And then of course you have [characters that behave like space in line breaking](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3Aline_break%3Dspace%3A%5D&g=&i=), [characters that behave like space in bidirectional ordering](https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3Abc%3Dwhite_space%3A%5D&g=&i=), etc.

Unicode Properties & Algorithms are an interesting place.)

spec/syntax.md
aphillips: @macchiati 

The SHOULD is already limited to RTL. I added an RTL part of the note about literals.

I split the isolate and marks paragraphs apart.

There are two ways to go on the "tools or users" thing. An alternative to getting into who should do it is to say "Messages SHOULD include..." and "Messages MAY include..." and then having a note about tools, which is how I've handled this.

Note that the syntax requires expression and quoted-pattern controls to be paired (although the whitespace rules break this)

spec/syntax.md
aphillips: Yes, I'm aware. Note well that these controls are not in the `s` production, which is what we call whitespace. I realize that other interpretations of whitespace are possible in Unicode. Your note addition would, I think, be confusing, since the controls are optional and are definitely *not* counted as whitespace in required whitespace parts of our syntax.

Is there harm in leaving this out? If so, I'll commit it.

eggrobin: I think it the conformance statement really ought to mention that you are adding to the set of ignorable format controls, but I agree that my proposed wording is needlessly pedantic and confusing. How about this:
```suggestion
the character `U+3000 IDEOGRAPHIC SPACE`
_is_ interpreted as whitespace,
 and the directional isolates U+2066..U+2069
 are treated as ignorable format controls.
```

aphillips: Works for me! Done.

spec/syntax.md
```diff
@@ -905,24 +926,77 @@ backslash       = %x5C ; U+005C REVERSE SOLIDUS "\"
...
+or _option_ values that use right-to-left characters,
+so that the _message_ displays intelligibly.
+_Literals_, especially each individual _key_ in a variant, that use right-to-left
+characters SHOULD be isolated using these paired controls.
+
+_Messages_ MAY also include `U+200E LEFT-TO-RIGHT MARK` or `U+200F RIGHT-TO-LEFT MARK`
+characters where permitted by the syntax before or following _identifiers_,
+_literals_ (taking care not to include the mark inside any quotes), 
+or _option_ values that use right-to-left characters 
+so that the _message_ displays intelligibly.
```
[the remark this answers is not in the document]
aphillips: The guidance on "never inside" is not necessarily correct, unless one adds something like "never inside unless intended to be part of the message's output"

I will add merge your suggestion, and then will edit it for spec styling requirements (and grammar)

macchiati:
Looks great. Will approve once I'm back at my computer

macchiati (approved):

spec/message.abnf
```diff
@@ -3,43 +3,45 @@ message           = simple-message / complex-message
...
+annotation-expression = owsp annotation *(wsp attribute) owsp
 
 annotation            = function
                       / private-use-annotation
                       / reserved-annotation
 
-markup = "{" [s] "#" identifier *(s option) *(s attribute) [s] ["/"] "}"  ; open and standalone
-       / "{" [s] "/" identifier *(s option) *(s attribute) [s] "}"  ; close
+markup = owsp "#" identifier *(wsp option) *(wsp attribute) owsp ["/"]  ; open and standalone
+       / owsp "/" identifier *(wsp option) *(wsp attribute) owsp        ; close
```
mihnita, line 38: I think that the `"{"`  and `"}"` got lost.

```suggestion
markup = "{" owsp "#" identifier *(wsp option) *(wsp attribute) owsp ["/"] "}" ; open and standalone
       / "{" owsp "/" identifier *(wsp option) *(wsp attribute) owsp "}"       ; close
```

mihnita (commented):
Thank you!

eemeli (changes requested):
Quoting @macchiati from https://github.com/unicode-org/message-format-wg/issues/673#issuecomment-1957206166:
> It is quite tricky, and we should not derail the tech preview release for
> this. That's why I (more strongly) urge that we capture this issue in a
> note in the spec for tech preview, and make the fix afterwards.

I am not comfortable landing this in the timeframe required for LDML 45. I think we need to take the time to consider and address this properly, rather than rushing through a solution right now.

spec/message.abnf
```diff
@@ -104,5 +106,11 @@ quoted-escape   = backslash ( backslash / "|" )
 reserved-escape = backslash ( backslash / "{" / "|" / "}" )
 backslash       = %x5C ; U+005C REVERSE SOLIDUS "\"
 
-; Whitespace
-s = 1*( SP / HTAB / CR / LF / %x3000 )
+; optional whitespace
+owsp = *( s / %x200E / %x200F / %x2066-2069 )
+
+; required whitespace
+wsp = [ (%x200E / %x200F / %x2066-2069 ) ] 1*s [ (%x200E / %x200F / %x2066-2069 ) ]
```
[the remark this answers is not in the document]
aphillips: I like it.

spec/message.abnf
aphillips: No, they didn't. They are in `placeholder`. 😸

mihnita (commented):
I am split about this.
Looks OK, but incomplete.
We should also look at isolates, and it is a bit too close to deadline.

aphillips:
@mihnita noted:

> Looks OK, but incomplete.
> We should also look at isolates, and it is a bit too close to deadline.

Can you clarify? What is incomplete? Also, this uses isolates, so your last comment is mysterious to me.

I agree that the deadlines are an issue. My concern here is that (a) we are a Unicode WG and this is a Unicode set of requirements. Even if we don't include it in 45, we need to deal with it and (b) syntax stability is important to me. Permitting bidi controls now in a somewhat (but not entirely!!) loose manner will prevent unnecessary churn later.

Anyway, to discuss in a few minutes in our call 😉

aphillips:
WG will consider in post-45. @aphillips to create a 45-timed PR with a note.

spec/syntax.md
```diff
@@ -955,24 +963,91 @@ and inside _patterns_ only escape `{` and `}`.
...
+   `U+200F RIGHT-TO-LEFT MARK` as permitted by the ABNF around
+   parts of any _message_ containing RTL characters:
+   -  _identifiers_
+   - _literals_ (taking care not to include the mark inside any quotes), 
+   - _option_ values
+
+> [!IMPORTANT]
+> Always take care **not** to add a bidi control where it is semantically significant:
+> - put them outside of _literal_ quotes, such as `<LRM>|...|<LRM>`
+> - put them outside of quoted _patterns_, such as `<LRI>{{...}}<PDI>`
```
echeran, line 1003: I suggest adding words Just to make it clear what the list is telling people to _not_ do (vs. whether they are supposed to do it). Also, I think you want a newline to ensure that the following sentences of the note don't get collapsed into the second bullet point.

```suggestion
> - do not put them outside of _literal_ quotes, such as `<LRM>|...|<LRM>`
> - do not put them outside of quoted _patterns_, such as `<LRI>{{...}}<PDI>`
>
```

echeran (approved):
Looks good. Added a minor suggestion.

spec/syntax.md
aphillips: Good point. I am making changes to this text as I incorporate it into a new PR that replaces this one and implements the bidi design.

aphillips:
Closing as obsolete

PR #679, opened 2024-02-22
Rename <key default> as <key catchall> in message.dtd

### LINKED ISSUE #676: [FEEDBACK] Definition of variant key in DTD does not match data model and JSON schema
In https://github.com/unicode-org/message-format-wg/blob/main/spec/data-model/README.md a variant key is described like this:

"For the CatchallKey, a string value may be provided to retain an identifier. This is always '*' in MessageFormat 2 syntax, but may vary in other formats.
```
interface Variant {
  keys: Array<Literal | CatchallKey>;
  value: Pattern;
}

interface CatchallKey {
  type: "*";
  value?: string;
}
```
So, a variant key is `Literal | CatchallKey`.

This is consistent with the definition of `variant` and `variant-key` in the JSON schema `message.json`:
```
    "variant-key": {
      "oneOf": [
        { "$ref": "#/$defs/literal" },
        {
          "type": "object",
          "properties": {
            "type": { "const": "*" },
            "value": { "type": "string" }
          },
          "required": ["type"]
        }
      ]
    },
```

However, the definition in the XML DTD is different:
```
<!ELEMENT variant (key+,pattern)>
<!ELEMENT key (#PCDATA)>
<!ATTLIST key default (true | false) "false">
```
Or, as an equivalent XML schema element:
```
  <element name="variant">
    <complexType>
      <sequence>
        <element ref="t:key" maxOccurs="unbounded"/>
        <element ref="t:pattern"/>
      </sequence>
    </complexType>
  </element>
  <element name="key">
    <complexType mixed="true">
      <attribute name="default" use="default" value="false">
        <simpleType>
          <restriction base="string">
            <enumeration value="true"/>
            <enumeration value="false"/>
          </restriction>
        </simpleType>
      </attribute>
    </complexType>
  </element>
```
Here, the variant key is a string with a boolean property named `default`. This is not immediately relatable to the `Literal | CatchallKey` definition above.

I can guess how a mapping between the two could be done: `default="false"` means `Literal`, `default="true"` means `CatchallKey`.

I would suggest to revisit this XML DTD element, for consistency.

Maybe by changing the attribute name from `default` to `catchall`.

Maybe also by changing the `value` in the CatchallKey to `identifier`, and adding an attribute `identifier` to the XML DTD element (instead of overloading the meaning of the XML element's contents).

eemeli:
Fixes #676

This resolves the naming inconsistency in the data model DTD.

The further recommendation by @bhaible
> Maybe also by changing the `value` in the CatchallKey to `identifier`, and adding an attribute `identifier` to the XML DTD element (instead of overloading the meaning of the XML element's contents).

is not included here, as it would potentially add more confusion. In the MF2 syntax, the non-default variant keys are _literal_ values, rather than _identifiers_, and the data model's catchall `value` is not used when representing MF2 syntax. It's there to support the lossless representation of other syntaxes, which may assign a literal value to their catchall variants.

### DIFF
--- a/spec/data-model/message.dtd
+++ b/spec/data-model/message.dtd
@@ -19,7 +19,7 @@
 <!ELEMENT selectors (expression)+>
 <!ELEMENT variant (key+,pattern)>
 <!ELEMENT key (#PCDATA)>
-<!ATTLIST key default (true | false) "false">
+<!ATTLIST key catchall (true | false) "false">
 
 <!ELEMENT pattern (#PCDATA | expression | markup)*>
 


aphillips (approved):

aphillips:
Needs one more approval to merge.

gibson042 (approved):

aphillips:
Two approves. LGTM

PR #681, opened 2024-02-22
Add note about XML C0 characters adding restrictions

### LINKED ISSUE #669: [FEEDBACK] C0 control characters are problematic for XML serialization
As far as I understand from the existence of https://github.com/unicode-org/message-format-wg/blob/main/spec/data-model/message.dtd and from the statement "The syntax should make a single message easily embeddable inside many container formats: ... XML, ..." in https://github.com/unicode-org/message-format-wg/blob/main/spec/syntax.md, it is important to be able to embed the data model and the string representation of a message in XML.

The `content-char` definition in the ABNF https://github.com/unicode-org/message-format-wg/blob/main/spec/message.abnf has three problems.

1) The comment "omit SP (%x20)" between the ranges %x0E-19 and %x21-2D indicates that it was the editor's intention to exclude %x20 only here, not the range %x1A-20. If so, the first range should be changed to %x0E-1F. If not, the comment should be changed.

2) The use of the NUL character (U+0000), which is disallowed in XML (both XML 1.0 and XML 1.1, see https://en.wikipedia.org/wiki/Valid_characters_in_XML), will lead to serialization problems with XML. Also, Eemeli Aro mentioned in Brussels that a C implementation was planned or ongoing, as part of the ICU project. But the NUL character makes the entire C API weird, because the usual string representation in C is with NUL-terminated strings.

3) The use of the characters U+0001..U+0008, U+000B..U+000C, U+000E..U+001F, which are disallowed in XML 1.0 and disallowed in document representations of XML 1.1 (see https://en.wikipedia.org/wiki/Valid_characters_in_XML, https://www.w3.org/TR/xml/#charsets definition [2], and https://www.w3.org/TR/2006/REC-xml11-20060816/#charsets definitions [2a] and [78]), will lead to serialization problems with XML

I would suggest to remove these characters from the `content-char` definition.

eemeli:
See #669, CC @bhaible

I agree with @aphillips that we should not restrict `content-char` characters because they're not all representable in XML. But we should explicitly note to readers that there is a concern here, and that they may need to add restrictions for their own message validation if they rely on XML.

### DIFF
--- a/spec/data-model/README.md
+++ b/spec/data-model/README.md
@@ -48,11 +48,12 @@ declarations, options, and attributes to be optional rather than required proper
 
 > [!NOTE]
 > Users relying on XML representations of messages should note that
-> XML 1.0 does not allow for the representation of all C0 control characters (U+0000-U+0019),
+> XML 1.0 does not allow for the representation of all C0 control characters (U+0000-U+001F)
 > and XML 1.1 does not allow for the representation of a U+0000 NULL character.
 > These characters are allowed in MessageFormat 2 messages,
 > so systems and users relying on XML as a resource format might need to
-> apply additional restrictions on message characters.
+> apply additional restrictions on message characters
+> or provide an alternative escape mechanism for these characters.
 
 > [!IMPORTANT]
 > The data model uses the field name `name` to denote various interface identifiers.
--- a/spec/data-model/README.md
+++ b/spec/data-model/README.md
@@ -46,6 +46,14 @@ the JSON and DTD definitions are intended for interchange between systems and pr
 To that end, they relax some aspects of the data model, such as allowing
 declarations, options, and attributes to be optional rather than required properties.
 
+> [!NOTE]
+> Users relying on XML representations of messages should note that
+> XML 1.0 does not allow for the representation of all C0 control characters (U+0000-U+0019),
+> and XML 1.1 does not allow for the representation of a U+0000 NULL character.
+> These characters are allowed in MessageFormat 2 messages,
+> so systems and users relying on XML as a resource format might need to
+> apply additional restrictions on message characters.
+
 > [!IMPORTANT]
 > The data model uses the field name `name` to denote various interface identifiers.
 > In the MessageFormat 2 [syntax](/spec/syntax.md), the source for these `name` fields
--- a/spec/data-model/README.md
+++ b/spec/data-model/README.md
@@ -50,9 +50,9 @@ declarations, options, and attributes to be optional rather than required proper
 > Users relying on XML representations of messages should note that
 > XML 1.0 does not allow for the representation of all C0 control characters (U+0000-U+001F).
 > Except for U+0000 NULL , these characters are allowed in MessageFormat 2 messages,
-> so systems and users relying on XML as a resource format might need to
-> apply additional restrictions on message characters
-> or provide an alternative escape mechanism for these characters.
+> so systems and users relying on this XML representation for interchange
+> might need to supply an alternate escape mechanism to support messages
+> that contain these characters.
 
 > [!IMPORTANT]
 > The data model uses the field name `name` to denote various interface identifiers.
--- a/spec/data-model/README.md
+++ b/spec/data-model/README.md
@@ -48,9 +48,8 @@ declarations, options, and attributes to be optional rather than required proper
 
 > [!NOTE]
 > Users relying on XML representations of messages should note that
-> XML 1.0 does not allow for the representation of all C0 control characters (U+0000-U+001F)
-> and XML 1.1 does not allow for the representation of a U+0000 NULL character.
-> These characters are allowed in MessageFormat 2 messages,
+> XML 1.0 does not allow for the representation of all C0 control characters (U+0000-U+001F).
+> Except for U+0000 NULL , these characters are allowed in MessageFormat 2 messages,
 > so systems and users relying on XML as a resource format might need to
 > apply additional restrictions on message characters
 > or provide an alternative escape mechanism for these characters.


bhaible:
The C0 control characters are U+0000-**U+001F**, not U+0000-U+0019.

spec/data-model/README.md
```diff
@@ -46,6 +46,14 @@ the JSON and DTD definitions are intended for interchange between systems and pr
 To that end, they relax some aspects of the data model, such as allowing
 declarations, options, and attributes to be optional rather than required properties.
 
+> [!NOTE]
+> Users relying on XML representations of messages should note that
+> XML 1.0 does not allow for the representation of all C0 control characters (U+0000-U+0019),
```
aphillips, line 51:
```suggestion
> XML 1.0 does not allow for the representation of all C0 control characters (U+0000-U+001F)
```

spec/data-model/README.md
```diff
@@ -46,6 +46,14 @@ the JSON and DTD definitions are intended for interchange between systems and pr
...
+> and XML 1.1 does not allow for the representation of a U+0000 NULL character.
+> These characters are allowed in MessageFormat 2 messages,
+> so systems and users relying on XML as a resource format might need to
+> apply additional restrictions on message characters.
```
aphillips, line 55: ... or provide an alternative escape mechanism for these characters.

RDF* for example uses `\u` and `\U` notations to allow non-XML chars. Note that by design our syntax does not provide these domain-specific escapes... exactly so that domains *can* provide them.

eemeli:
```suggestion
> so systems and users relying on XML as a resource format might need to
> apply additional restrictions on message characters
> or provide an alternative escape mechanism for these characters.
```

spec/data-model/README.md
eemeli: D'oh!

eemeli:
This will need to be rebased if #682 is accepted, or the note updated in that PR if this lands first.

aphillips (approved):

spec/data-model/README.md
```diff
@@ -46,6 +46,15 @@ the JSON and DTD definitions are intended for interchange between systems and pr
...
 declarations, options, and attributes to be optional rather than required properties.
 
+> [!NOTE]
+> Users relying on XML representations of messages should note that
+> XML 1.0 does not allow for the representation of all C0 control characters (U+0000-U+001F)
+> and XML 1.1 does not allow for the representation of a U+0000 NULL character.
+> These characters are allowed in MessageFormat 2 messages,
+> so systems and users relying on XML as a resource format might need to
+> apply additional restrictions on message characters
+> or provide an alternative escape mechanism for these characters.
```
stasm, line 56: I may be missing something, but these work-arounds don't address @bhaible's concern about interoperability between implementations. They recommend a path for producing valid XML within a single implementation.

Can you give an example of restrictions or escape mechanism that you had in mind here?

eemeli: A possible restriction would be to require all messages to contain no C0 control codes.

A possible escape mechanism would be to use backslash escapes like `\u0000` in the serialised XML for control codes, which would necessitate a custom parser and serialiser.

Both of these could be applied on top of what we are presenting here. Alternatively, if we accept #682, then sticking to XML 1.1 would provide built-in support for all valid MF2 messages.

stasm: Thanks. If I understand @bhaible's feedback, both of these mechanism will render the implementation not operable with other, different implementations. In other words, if I restrict all messages to avoid C0 codes, then I cannot accept as input a message produced by an implementation which doesn't have that restriction.

I think #682 + recommending XML 1.1 would be a more robust work-around, as you say.

stasm, line 56: How about something like the following instead?

```suggestion
> Users relying on XML representations of messages should note that
> XML 1.0 does not allow for the representation of all C0 control characters (U+0000-U+001F).
> These characters are allowed in valid MessageFormat 2 messages.
> Thus, XML 1.1 is recommended for representing serialized MessageFormat 2 messages.
>  
> However, in MessageFormat 2 Tech Preview, valid messages may contain the U+0000 NULL character,
> which cannot be represented in XML 1.1.
> Systems and users relying on XML as a resource format might need to
> apply additional restrictions on message characters
> or provide an alternative escape mechanism for the U+0000 NULL character.
```

bhaible: > I think https://github.com/unicode-org/message-format-wg/pull/682 + recommending XML 1.1 would be a more robust work-around

When it comes to interoperability between implementations, most likely an XML document will be exchanged between the two implementations. Since XML 1.1, for XML documents, also forbids the C0 control characters (see https://www.w3.org/TR/2006/REC-xml11-20060816/#sec-well-formed definitions [1] and [2a]), there is no advantage in recommending XML 1.1. In this respect, XML 1.1 is as bad as XML 1.0.

eemeli: But XML 1.1 does allow for character reference escapes for the control characters #x1 through #x1F, yes?

bhaible: > But XML 1.1 does allow for character reference escapes for the control characters #x1 through #x1F, yes?

Yes, it does. Quoting https://www.w3.org/TR/2006/REC-xml11-20060816/#sec-xml11 :
"XML 1.1 allows the use of character references to the control characters #x1 through #x1F, most of which are forbidden in XML 1.0. For reasons of robustness, however, these characters still cannot be used directly in documents."

So, I withdraw my statement that "XML 1.1 is as bad as XML 1.0".

But trying to validate the XML 1.1 document
```
<?xml version="1.1" encoding="UTF-8" ?>
<foo>
  <bar>C0 control &#0001;</bar>
  <bar>C0 control &#0002;</bar>
  <bar>C0 control &#0003;</bar>
  <bar>C0 control &#0004;</bar>
  <bar>C0 control &#0005;</bar>
  <bar>C0 control &#0006;</bar>
  <bar>C0 control &#0007;</bar>
  <bar>C0 control &#0008;</bar>
  <bar>C0 control &#0011;</bar>
  <bar>C0 control &#0012;</bar>
  <bar>C0 control &#0014;</bar>
  <bar>C0 control &#0015;</bar>
  <bar>C0 control &#0016;</bar>
  <bar>C0 control &#0017;</bar>
  <bar>C0 control &#0018;</bar>
  <bar>C0 control &#0019;</bar>
  <bar>C0 control &#0020;</bar>
  <bar>C0 control &#0021;</bar>
  <bar>C0 control &#0022;</bar>
  <bar>C0 control &#0023;</bar>
  <bar>C0 control &#0024;</bar>
  <bar>C0 control &#0025;</bar>
  <bar>C0 control &#0026;</bar>
  <bar>C0 control &#0027;</bar>
  <bar>C0 control &#0028;</bar>
  <bar>C0 control &#0029;</bar>
  <bar>C0 control &#0030;</bar>
  <bar>C0 control &#0031;</bar>
</foo>
```
I see that all three validators that I had used in https://github.com/unicode-org/message-format-wg/issues/669#issuecomment-1952811014 don't support XML 1.1, only XML 1.0.

So, while relying on XML 1.1 would fix the problem in theory, in practice it loses, since XML 1.1 is hardly supported by software.

Even the XML support in Microsoft .NET does not support version 1.1: https://stackoverflow.com/questions/76124030/

aphillips:
XML 1.1 is effectively a dead letter. We shouldn't depend on or refer to it.

spec/data-model/README.md
aphillips: (as individual contributor) 

It isn't just NULL that's a problem in XML 1.0 and, as noted, XML 1.1 isn't really a thing we can use.

Instead of mucking around with this, I think we're better off saying that some messages can't be serialized in the TP's XML data model. We can fix this _during_ the preview.

stasm: Agreed. I was trying to suggest an improvement to the note's phrasing based on the assumption that we *could* recommend XML 1.1, and to make room for #682. At this point, however, it's best to not make any late changes hastily.

stasm (approved):

spec/data-model/README.md
mihnita: +1

If localizable strings are stored in XML and some platform needs the controls for some reason,
they might already have a way to represent them for non MF2 messages.
If we add our own mechanism we only add to the confusion.

For example XLIFF has [`<cp>`](https://docs.oasis-open.org/xliff/xliff-core/v2.1/os/xliff-core-v2.1-os.html#cp)

Let's say my exiting XML use the same convention.
If a tool want to save "Hello <the real 0003 character> world" into such a file it has no way to know if it should escape as `<cp hex="0003"/>`, or as `\u0003` (MF2).
Because just by looking at a string there is no way to tell what API is used on it.
So the safe thing is to escape with the storage convention (`<cp hex="0003"/>`),
And we only introduce the MF2 complexity for no good reason.

mihnita:
TLDR: I don't think we should do this.

spec/data-model/README.md
```diff
@@ -46,6 +46,15 @@ the JSON and DTD definitions are intended for interchange between systems and pr
 To that end, they relax some aspects of the data model, such as allowing
 declarations, options, and attributes to be optional rather than required properties.
 
+> [!NOTE]
+> Users relying on XML representations of messages should note that
+> XML 1.0 does not allow for the representation of all C0 control characters (U+0000-U+001F)
+> and XML 1.1 does not allow for the representation of a U+0000 NULL character.
+> These characters are allowed in MessageFormat 2 messages,
```
eemeli, line 53:
```suggestion
> XML 1.0 does not allow for the representation of all C0 control characters (U+0000-U+001F).
> Except for U+0000 NULL , these characters are allowed in MessageFormat 2 messages,
```

spec/data-model/README.md
```diff
@@ -46,6 +46,14 @@ the JSON and DTD definitions are intended for interchange between systems and pr
 To that end, they relax some aspects of the data model, such as allowing
 declarations, options, and attributes to be optional rather than required properties.
 
+> [!NOTE]
+> Users relying on XML representations of messages should note that
+> XML 1.0 does not allow for the representation of all C0 control characters (U+0000-U+001F).
+> Except for U+0000 NULL , these characters are allowed in MessageFormat 2 messages,
+> so systems and users relying on XML as a resource format might need to
+> apply additional restrictions on message characters
+> or provide an alternative escape mechanism for these characters.
```
aphillips, line 55: The mention of "resource format" is not applicable here? This is the data model! Perhaps:

```suggestion
> so systems and users relaying on this XML representation for interchange
> might need to apply additional restrictions on

aphillips (changes requested):
Maybe a bit of tweaking? The note should be about the data model, not just any XML format.

spec/data-model/README.md
eemeli: Updated suggestion to not change "relying" to "relaying".

eemeli:
> The note should be about the data model, not just any XML format.

I think I either disagree, or don't understand? The data model and its JSON representation are fully capable of representing content that includes any and all characters, so the note here is specifically about its XML representation, which can't.

stasm (approved):

aphillips:
> > The note should be about the data model, not just any XML format.
> I think I either disagree, or don't understand? The data model and its JSON representation are fully capable of 
> representing content that includes any and all characters, so the note here is specifically about its XML representation, 
> which can't.

Probably I didn't express myself well enough. My comment was about the XML representation of the data model, not just any XML format. It should be explicit that it's about that.

aphillips (approved):
Good enough!

PR #682, opened 2024-02-22
Do not allow null characters in message source

### LINKED ISSUE #669: [FEEDBACK] C0 control characters are problematic for XML serialization

eemeli:
Closes #669

As @bhaible points out, U+0000 NUL characters are challenging in null-terminated strings as well as XML. These are real concerns, but U+0000 should still be representable in MF2. Rather than requiring custom solutions that are likely to not be mutually compatible, we should make a specific affordance for U+0000, removing it from `content-char` and adding `\0` as an MF2 escape.

### DIFF
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -88,7 +88,7 @@ simple-start-char = content-char / s / "@" / "|"
 text-char         = content-char / s / "." / "@" / "|"
 quoted-char       = content-char / s / "." / "@" / "{" / "}"
 reserved-char     = content-char / "."
-content-char      = %x01-08        ; omit NUL (%x00), HTAB (%x09) and LF (%x0A)
+content-char      = %x01-08        ; omit NULL (%x00), HTAB (%x09) and LF (%x0A)
                   / %x0B-0C        ; omit CR (%x0D)
                   / %x0E-1F        ; omit SP (%x20)
                   / %x21-2D        ; omit . (%x2E)
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -88,7 +88,7 @@ simple-start-char = content-char / s / "@" / "|"
 text-char         = content-char / s / "." / "@" / "|"
 quoted-char       = content-char / s / "." / "@" / "{" / "}"
 reserved-char     = content-char / "."
-content-char      = %x00-08        ; omit HTAB (%x09) and LF (%x0A)
+content-char      = %x01-08        ; omit NUL (%x00), HTAB (%x09) and LF (%x0A)
                   / %x0B-0C        ; omit CR (%x0D)
                   / %x0E-1F        ; omit SP (%x20)
                   / %x21-2D        ; omit . (%x2E)
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -271,8 +271,8 @@ A _quoted pattern_ MAY be empty.
 ### Text
 
 **_<dfn>text</dfn>_** is the translateable content of a _pattern_.
-Any Unicode code point is allowed, except for surrogate code points U+D800
-through U+DFFF inclusive.
+Any Unicode code point is allowed, except for U+0000 NUL
+and the surrogate code points U+D800 through U+DFFF inclusive.
 The characters U+005C REVERSE SOLIDUS `\`,
 U+007B LEFT CURLY BRACKET `{`, and U+007D RIGHT CURLY BRACKET `}`
 MUST be escaped as `\\`, `\{`, and `\}` respectively.
@@ -291,7 +291,7 @@ simple-start-char = content-char / s / "@" / "|"
 text-char         = content-char / s / "." / "@" / "|"
 quoted-char       = content-char / s / "." / "@" / "{" / "}"
 reserved-char     = content-char / "."
-content-char      = %x00-08        ; omit HTAB (%x09) and LF (%x0A)
+content-char      = %x01-08        ; omit NUL (%x00), HTAB (%x09) and LF (%x0A)
                   / %x0B-0C        ; omit CR (%x0D)
                   / %x0E-1F        ; omit SP (%x20)
                   / %x21-2D        ; omit . (%x2E)
@@ -794,7 +794,7 @@ as a _key_ value,
 as the _operand_ of a _literal-expression_,
 or in the value of an _option_.
 A _literal_ MAY include any Unicode code point
-except for surrogate code points U+D800 through U+DFFF.
+except for U+0000 NUL or the surrogate code points U+D800 through U+DFFF.
 
 All code points are preserved.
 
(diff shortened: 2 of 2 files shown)


aphillips:
I think this is a reasonable solution and the change looks fine. C

I think we should hold this for post 45.

macchiati:
We purposely wanted to minimize the escaping structure because it can
collide with an enclosing envelope's escaping format. Supporting null in
message strings is a complete edge case: I'd rather see us just limit the
valid characters to exclude null.

bhaible:
> Supporting null in
message strings is a complete edge case: I'd rather see us just limit the
valid characters to exclude null.

I agree that excluding U+0000 is a better solution than escaping as `\0`:
- It is simpler.
- If someone is trying to use U+0000 as a separator character, for example in order to stuff several messages into a single message, they should better use one of the ordinary separator characters instead (such as U+2029 PARAGRAPH SEPARATOR). This is likely to work better with surrounding software.

aphillips:
@macchiati isn't not just NULL, it's an array of C0 controls.

The JSON representation of the data model can escape the characters, so would work.

An XML representation of the data model, by definition, needs to provide either escapes or just not support interchange of these (troublesome) characters (which is probably inappropriate given the goals of the data model). 

If it were just NULL, I could grit my teeth and provide an escape. C and C strings exist, as @stasm put it. But I don't support adding any other escapes to our syntax: we don't want that. These characters will often *already* be escaped in their source format. I'd much rather limit the escaping to the data model representation. It does mean some layer of additional processing beyond just grabbing the value, but there will likely already be some additional processing 😸 

(Note: I recently encountered this exact problem in the RDF* specifications and their solution is to use `\u` and `\U` escapes. We can certainly do that too--when we flesh out the DM.)

mihnita:
I can't think of any decent use case for NULLs in localizable strings.
So I would rather exclude 0 from the syntax completely.

And the C0, even if they have a use case, should probably be left to the storage format, whatever that is.
After all, not all messages stored in the "resource bundle" go through `MessageFormat2`.
So if one need to use C0 in messages used "as is", they have a way to store them.
They can use the same storage solution for MF2 messages.

---

My TLDR: exclude NULL, and leave everything else as is.

duerst:
I agree with @mihnita. If we introduce some kind of escaping (e.g. \0, or \uABCD, or whatever else), it will interact with the storage format. That's exactly what we wanted to avoid.

And please note that even if only \0 is escaped, that means that we then need to escape a backslash as \\ or whatever.

aphillips:
> After all, not all messages stored in the "resource bundle" go through MessageFormat2.

Another way to look at this is that all such strings are simple messages.

> And please note that even if only \0 is escaped, that means that we then need to escape a backslash as \ or whatever.

Note that we already do escape backslash, since we provide an escape for `{` and `}` in patterns and for `|` in literals. Each of these represent syntax-meaningful characters. But I agree to not escaping any more than is necessary.

Note that my comment was to provide escapes _in the XML data model_, which is a different thing than providing them in the MF2 syntax.

aphillips:
To be retitled and repurposed. NUL to be excluded.

eemeli:
Rebased on `main`, now only banning NUL. If you need it in your output, you'll need to pass it in as a variable or define a custom `:nul` (or similar) for it.

spec/message.abnf
```diff
@@ -88,7 +88,7 @@ simple-start-char = content-char / s / "@" / "|"
 text-char         = content-char / s / "." / "@" / "|"
 quoted-char       = content-char / s / "." / "@" / "{" / "}"
 reserved-char     = content-char / "."
-content-char      = %x00-08        ; omit HTAB (%x09) and LF (%x0A)
+content-char      = %x01-08        ; omit NUL (%x00), HTAB (%x09) and LF (%x0A)
```
aphillips, line 91: The name of the character is actually `NULL`

```suggestion
content-char      = %x01-08        ; omit NULL (%x00), HTAB (%x09) and LF (%x0A)
```

spec/syntax.md
```diff
@@ -271,8 +271,8 @@ A _quoted pattern_ MAY be empty.
 ### Text
 
 **_<dfn>text</dfn>_** is the translateable content of a _pattern_.
-Any Unicode code point is allowed, except for surrogate code points U+D800
-through U+DFFF inclusive.
+Any Unicode code point is allowed, except for U+0000 NUL
```
aphillips, line 274:
```suggestion
Any Unicode code point is allowed, except for U+0000 NULL
```

spec/syntax.md
```diff
@@ -291,7 +291,7 @@ simple-start-char = content-char / s / "@" / "|"
 text-char         = content-char / s / "." / "@" / "|"
 quoted-char       = content-char / s / "." / "@" / "{" / "}"
 reserved-char     = content-char / "."
-content-char      = %x00-08        ; omit HTAB (%x09) and LF (%x0A)
+content-char      = %x01-08        ; omit NUL (%x00), HTAB (%x09) and LF (%x0A)
```
aphillips, line 294:
```suggestion
content-char      = %x01-08        ; omit NULL (%x00), HTAB (%x09) and LF (%x0A)
```

aphillips (approved):

stasm (approved):

PR #684, opened 2024-02-22
Address #683: make registry data model non-TP

### LINKED ISSUE #683: Update status of registry spec for Tech Preview
@macchiati suggested in #671 

---

In the Conformance and Use:

The term 'registry' is being used ambiguously. That is, the distinction is not clearly enough drawn between the default registry, and the machine-readable format that is used by the default registry, but can also be used as implementations to provide machine-readable documentation/specification of their own registry, with additions to the default registry.

For example, "The MessageFormat 2.0 Registry was created to describe the core set of formatting and selection functions," seems really to be describing the default registry, not the registry format. So it should say "The MessageFormat 2.0 Default Registry"

In the next paragraph "The registry provides a machine-readable description" seems to be talking about the use as a format. It should have its own term, like 'registry format' (but doesn't have to be exactly that term).

I suggest that we don't have any 'naked' mentions of the word 'registry' in the spec, but rather qualify it to resolve the ambiguity.

---

I made other comments there. This issue tracks making registry terms consistent and adding a note for tech preview

aphillips:
This clarifies the status of the registry data model as experimental.

Fixes #683

### DIFF
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -2,7 +2,7 @@
 
 Implementations and tooling can greatly benefit from a
 structured definition of formatting and matching functions available to messages at runtime.
-The _registry_ is a mechanism for storing such declarations in a portable manner.
+This specification is intended to provide a mechanism for storing such declarations in a portable manner.
 
 ## Goals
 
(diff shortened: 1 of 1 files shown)


spec/registry.md
```diff
@@ -30,12 +30,21 @@ in order to support the following goals and use-cases:
 _This section is normative._
 
 To be conformant with MessageFormat 2.0, an implementation MUST implement
-all of the formatting and selection _functions_ described in the default registry,
-including all of the _options_ and _option_ values, _operands_ and outputs
-described by the default registry.
+the _functions_/_selectors_, _options_ and _option_ values, _operands_ and outputs
```
eemeli, line 33:
```suggestion
the _functions_, _options_ and _option_ values, _operands_ and outputs
```

spec/registry.md
```diff
@@ -258,6 +270,10 @@ which only expects the `accord` option:
 
 # Default Registry
 
+> [!IMPORTANT]
+> That part of the specification is part of the Tech Preview
```
eemeli, line 274:
```suggestion
> This part of the specification is part of the Tech Preview
```

eemeli (approved):

stasm (approved):

macchiati:
I would would change "experimental" to "experimental and incomplete".

IMO It is in such bad shape, given more recent discussions and changes, that it should be pulled entirely from the spec OR at least moved to the very end.

PR #686, opened 2024-02-23
Add note to "Function Resolution" section about function argument and result types

catamorphism:
Currently, the introduction to the spec states:

>   The form of the resolved value is implementation defined and the
>  value might not be evaluated or formatted yet.
> However, it needs to be "formattable", i.e. it contains everything required
> by the eventual formatting.

And the "Expression and Markup Resolution" section says:

> Since a _variable_ can be referenced in different ways later,
> implementations SHOULD NOT immediately fully format the value for output.

However, the "Function Resolution" section is not as clear as it could be about the implications of these requirements for the interface with formatting functions.

I added some text that effectively implies that functions have the same operand type and result type. If this wasn't the case, it wouldn't make sense to bind the result of a function to a variable and use that result as an operand for another function call.

I think it's useful guidance for implementors to state this explicitly rather than letting it be inferred from the two existing passages that I quoted.

This relates to #515; some version of #645 would make this much more precise, but it's a start.

The reason this came up was that I was discussing the API for custom functions with members of the ICU TC, who were puzzled at first about why formatting functions take and return the same type (in my implementation). 

If an implementor instead requires custom functions to take a "formattable" thing as an argument, and return a "formatted" thing, examples like the first use case in #515 (with the two calls to `:number`) wouldn't work. In my opinion, the current spec doesn't say that you can't do this -- it could be read as saying that a "resolved _operand_ value" as mentioned in step 4, and the value referred to by the text "resolve the value of the _expression_ as the result of that function call", might refer to different kinds of "resolved values".

(Taking and returning the same type makes formatting functions seem more like "transformers" than "formatters" -- and that's in line with the text in [syntax.md](https://github.com/unicode-org/message-format-wg/blob/main/spec/syntax.md) saying, "Functions are used to evaluate, format, select, or otherwise process data values during formatting." -- but changing the name might be more controversial.)

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -267,7 +267,7 @@ the following steps are taken:
 > {{$n1}}
 > ```
 >
-> is implementation-dependent.
+> is currently implementation-dependent.
 > Depending on whether the options are preserved
 > between the resolution of the first `:number` _annotation_
 > and the resolution of the second `:number` _annotation_,
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -247,12 +247,8 @@ the following steps are taken:
    MAY define an interface for custom function implementations.
    Such an interface SHOULD define an implementation-specific
    argument type `T` and return type `U`
-   for implementations of formatting functions
+   for implementations of functions
    such that `U` can be coerced to `T` without loss of information.
-   The type `U`
-   (or a type that `U` can be coerced to without loss of information)
-   SHOULD also be the input type of implementations of
-   custom selector functions.
    Implementations of specific functions MAY
    signal errors if supplied an _operand_ that
    does not make sense for the particular function
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -274,9 +274,13 @@ the following steps are taken:
 > a conformant implementation
 > could produce either "001.000" or "1.000"
 >
-> Each function implementation MAY have
+> Each function **specification** MAY have
 > its own rules to preserve some options in the returned structure
-> and discard others.
+> and discard others. 
+> In instances where a function specification does not determine whether an option is preserved or discarded,
+> each function **implementation** of that specification MAY have
+> its own rules to preserve some options in the returned structure
+> and discard others. 
 >
 >
 > [!NOTE]
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -234,8 +234,13 @@ the following steps are taken:
    ```
    the second call to `:number` composes with the first call.
 
+   In addition, selector functions compose with formatting functions
+   in the sense that a selector function's _operand_
+   may be the output of any formatting function.
+
    Implementations SHOULD provide a means for formatting functions
-   to compose with each other.
+   to compose with each other
+   and for formatting functions to compose with selector functions.
    Implementations that provide a means for defining custom functions
    SHOULD provide a means for those functions to return values
    that contain enough information
@@ -245,8 +250,11 @@ the following steps are taken:
    For example, an implementation in a typed programming language
    MAY define an interface that custom functions implement.
    Such an interface SHOULD define an implementation-specific
-   argument type `T` and return type `U` for custom functions
+   argument type `T` and return type `U` for custom formatting functions
    such that `U` can be coerced to `T` without loss of information.
+   The type `U`
+   (or a type that `U` can be coerced to without loss of information)
+   SHOULD also be the input type of custom selector functions.
 
 > [!NOTE]
 > In the Tech Preview, the spec leaves the behavior of the previous
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -222,6 +222,11 @@ the following steps are taken:
 
    The form that resolved _operand_ and _option_ values take is implementation-defined.
 
+   Since the result of a function call can be bound to a _variable_,
+   the output of one _function_ may be the input of another _function_.
+   Thus, formatting functions SHOULD use a structure for the resolved _operand_ value
+   that is interconvertible with the structure for the result of the _function_.
+
    An implementation MAY pass additional arguments to the function,
    as long as reasonable precautions are taken to keep the function interface
    simple and minimal, and avoid introducing potential security vulnerabilities.
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -230,7 +230,7 @@ the following steps are taken:
    For example, in
    ```
    .input {$n :number minIntegerDigits=3}
-   .local {$n1 :number maxFractionDigits=3}
+   .local $n1 = {$n :number maxFractionDigits=3}
    ```
    the second call to `:number` composes with the first call.
 
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -256,7 +256,7 @@ the following steps are taken:
 
 > [!NOTE]
 > The behavior of the previous example is
-> implementation-dependent. Supposing that
+> currently implementation-dependent. Supposing that
 > the external input variable `n` is bound to the string `"1"`,
 > and that the implementation formats to a string,
 > the formatted result of the following message:
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -225,40 +225,41 @@ the following steps are taken:
    A _local-declaration_ binds the resolved value of an _expression_
    to a _variable_.
    Thus, the output of one _function_ is potentially the _operand_
-   of another _function_. In other words, formatting functions
-   compose with each other.
+   of another _function_,
+   or the value of one of the _options_ for another function.
    For example, in
    ```
    .input {$n :number minIntegerDigits=3}
    .local $n1 = {$n :number maxFractionDigits=3}
    ```
-   the second call to `:number` composes with the first call.
+   the output of the first call to `:number`
+   is the input of the second call to `:number`.
 
-   In addition, selector functions compose with formatting functions
-   in the sense that a selector function's _operand_
-   may be the output of any formatting function.
-
-   Implementations SHOULD provide a means for formatting functions
-   to compose with each other
-   and for formatting functions to compose with selector functions.
    Implementations that provide a means for defining custom functions
-   SHOULD provide a means for those functions to return values
-   that contain enough information
+   SHOULD provide a means for function implementations
+   to return values that contain enough information
    (e.g. the resolved _operand_ and _option_ values
    that the function was called with)
-   to be used as inputs to subsequent function calls.
-   For example, an implementation in a typed programming language
-   MAY define an interface that custom functions implement.
+   to be used as arguments to subsequent calls
+   to the function implementations.
+   For example, a MessageFormat implementation in a typed programming language
+   MAY define an interface for custom function implementations.
    Such an interface SHOULD define an implementation-specific
-   argument type `T` and return type `U` for custom formatting functions
+   argument type `T` and return type `U`
+   for implementations of formatting functions
    such that `U` can be coerced to `T` without loss of information.
    The type `U`
    (or a type that `U` can be coerced to without loss of information)
-   SHOULD also be the input type of custom selector functions.
+   SHOULD also be the input type of implementations of
+   custom selector functions.
+   Implementations of specific functions MAY
+   signal errors if supplied an _operand_ that
+   does not make sense for the particular function
+   being implemented.
 
 > [!NOTE]
-> In the Tech Preview, the spec leaves the behavior of the previous
-> example implementation-dependent. Supposing that
+> The behavior of the previous example is
+> implementation-dependent. Supposing that
 > the external input variable `n` is bound to the string `"1"`,
 > and that the implementation formats to a string,
 > the formatted result of the following message:
@@ -270,12 +271,24 @@ the following steps are taken:
 > ```
 >
 > is implementation-dependent.
-> Depending on whether the options are preserved across
-> the two calls to `:number`, a conformant implementation
+> Depending on whether the options are preserved
+> between the resolution of the first `:number` _annotation_
+> and the resolution of the second `:number` _annotation_,
+> a conformant implementation
 > could produce either "001.000" or "1.000"
-> Feedback from users and implementers is desired
-> about whether to require one interpretation or the other
-> in the spec.
+>
+> Each function implementation MAY have
+> its own rules to preserve some options in the returned structure
+> and discard others.
+>
+>
+> [!NOTE]
+> During the Technical Preview,
+> feedback on how the registry describes
+> the flow of _resolved values_ and _options_
+> from one _function_ to another,
+> and on what requirements this specification should impose,
+> is highly desired.
 
    An implementation MAY pass additional arguments to the function,
    as long as reasonable precautions are taken to keep the function interface
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -245,8 +245,7 @@ the following steps are taken:
    that the function was called with)
    to be used as arguments to subsequent calls
    to the function implementations.
-   For example, a MessageFormat implementation in a typed programming language
-   MAY define an interface for custom function implementations.
+   For example, an implementation might define an interface that allows custom function implementation.
    Such an interface SHOULD define an implementation-specific
    argument type `T` and return type `U`
    for implementations of functions
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -238,7 +238,8 @@ the following steps are taken:
    Implementations that provide a means for defining custom functions
    SHOULD provide a means for function implementations
    to return values that contain enough information
-   (e.g. the resolved _operand_ and _option_ values
+   (e.g. a representation of
+   the resolved _operand_ and _option_ values
    that the function was called with)
    to be used as arguments to subsequent calls
    to the function implementations.
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -225,9 +225,49 @@ the following steps are taken:
    A _local-declaration_ binds the resolved value of an _expression_
    to a _variable_.
    Thus, the output of one _function_ is potentially the _operand_
-   of another _function_.
-   Thus, formatting functions SHOULD use a structure for the resolved _operand_ value
-   that is interconvertible with the structure for the result of the _function_.
+   of another _function_. In other words, formatting functions
+   compose with each other.
+   For example, in
+   ```
+   .input {$n :number minIntegerDigits=3}
+   .local {$n1 :number maxFractionDigits=3}
+   ```
+   the second call to `:number` composes with the first call.
+
+   Implementations SHOULD provide a means for formatting functions
+   to compose with each other.
+   Implementations that provide a means for defining custom functions
+   SHOULD provide a means for those functions to return values
+   that contain enough information
+   (e.g. the resolved _operand_ and _option_ values
+   that the function was called with)
+   to be used as inputs to subsequent function calls.
+   For example, an implementation in a typed programming language
+   MAY define an interface that custom functions implement.
+   Such an interface SHOULD define an implementation-specific
+   argument type `T` and return type `U` for custom functions
+   such that `U` can be coerced to `T` without loss of information.
+
+> [!NOTE]
+> In the Tech Preview, the spec leaves the behavior of the previous
+> example implementation-dependent. Supposing that
+> the external input variable `n` is bound to the string `"1"`,
+> and that the implementation formats to a string,
+> the formatted result of the following message:
+>
+> ```
+> .input {$n :number minIntegerDigits=3}
+> .local {$n1 :number maxFractionDigits=3}
+> {{$n1}}
+> ```
+>
+> is implementation-dependent.
+> Depending on whether the options are preserved across
+> the two calls to `:number`, a conformant implementation
+> could produce either "001.000" or "1.000"
+> Feedback from users and implementers is desired
+> about whether to require one interpretation or the other
+> in the spec.
 
    An implementation MAY pass additional arguments to the function,
    as long as reasonable precautions are taken to keep the function interface
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -250,10 +250,9 @@ the following steps are taken:
    argument type `T` and return type `U`
    for implementations of functions
    such that `U` can be coerced to `T` without loss of information.
-   Implementations of specific functions MAY
-   signal errors if supplied an _operand_ that
-   does not make sense for the particular function
-   being implemented.
+   Implementations of a _function_ SHOULD emit an 
+   _Invalid Expression_ error for _operands_ whose resolved value
+   or type is not supported.
 
 > [!NOTE]
 > The behavior of the previous example is
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -249,7 +249,7 @@ the following steps are taken:
    Such an interface SHOULD define an implementation-specific
    argument type `T` and return type `U`
    for implementations of functions
-   such that `U` can be coerced to `T` without loss of information.
+   such that `U` can be coerced to `T`.
    Implementations of a _function_ SHOULD emit an 
    _Invalid Expression_ error for _operands_ whose resolved value
    or type is not supported.
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -224,7 +224,7 @@ the following steps are taken:
 
    A _declaration_ binds the resolved value of an _expression_
    to a _variable_.
-   Thus, the output of one _function_ is potentially the _operand_
+   Thus, the result of one _function_ is potentially the _operand_
    of another _function_,
    or the value of one of the _options_ for another function.
    For example, in
@@ -232,8 +232,10 @@ the following steps are taken:
    .input {$n :number minIntegerDigits=3}
    .local $n1 = {$n :number maxFractionDigits=3}
    ```
-   the output of the first call to `:number`
-   is the input of the second call to `:number`.
+   the value bound to `$n` is the
+   resolved value used as the _operand_
+   of the `:number` _function_
+   when resolving the value of the _variable_ `$n1`.
 
    Implementations that provide a means for defining custom functions
    SHOULD provide a means for function implementations
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -266,7 +266,7 @@ the following steps are taken:
 >
 > ```
 > .input {$n :number minIntegerDigits=3}
-> .local {$n1 :number maxFractionDigits=3}
+> .local $n1 = {$n :number maxFractionDigits=3}
 > {{$n1}}
 > ```
 >
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -222,7 +222,7 @@ the following steps are taken:
 
    The form that resolved _operand_ and _option_ values take is implementation-defined.
 
-   A _local-declaration_ binds the resolved value of an _expression_
+   A _declaration_ binds the resolved value of an _expression_
    to a _variable_.
    Thus, the output of one _function_ is potentially the _operand_
    of another _function_,
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -222,8 +222,10 @@ the following steps are taken:
 
    The form that resolved _operand_ and _option_ values take is implementation-defined.
 
-   Since the result of a function call can be bound to a _variable_,
-   the output of one _function_ may be the input of another _function_.
+   A _local-declaration_ binds the resolved value of an _expression_
+   to a _variable_.
+   Thus, the output of one _function_ is potentially the _operand_
+   of another _function_.
    Thus, formatting functions SHOULD use a structure for the resolved _operand_ value
    that is interconvertible with the structure for the result of the _function_.
 


spec/formatting.md
```diff
@@ -222,6 +222,11 @@ the following steps are taken:
 
    The form that resolved _operand_ and _option_ values take is implementation-defined.
 
+   Since the result of a function call can be bound to a _variable_,
+   the output of one _function_ may be the input of another _function_.
+   Thus, formatting functions SHOULD use a structure for the resolved _operand_ value
+   that is interconvertible with the structure for the result of the _function_.
```
eemeli, line 228: A few observations:
1. It is misleading to refer to "formatting" functions here, given that their output may also be used for selection. Note how the text around this avoids that term.

2. I at least have never encountered the term "interconvertible", and using it here should be avoided.

3. There are cases where it makes sense for the operand type to be wider than the output type. For example, consider the resolution of `{$n :number}`, where the implicit input variable `$n` has a string value `'42'`. In a programming language like JS, it's easier for the custom function implementation to accept that its input may be a number, bigint, string, or object, rather than requiring each of those to come pre-wrapped as suggested by the SHOULD. See steps 6 and 7 [here](https://tc39.es/proposal-intl-messageformat/#sec-messageformat-numberfunctions) for the JS implementation details of this particular case.

   I do fully agree that the output of a function is expected to be in a shape that's acceptable as input, but I am not convinced that the input must always have that same shape.

catamorphism: > A few observations:
> 
>     1. It is misleading to refer to "formatting" functions here, given that their output may also be used for selection. Note how the text around this avoids that term.
> 

I wrote it that way since selector functions don't have an output (or at least, not in the way that the "output" is being described here.) Though I'll look and see if Addison's suggested changes address that.

>     2. I at least have never encountered the term "interconvertible", and using it here should be avoided.

OK.

> 
>     3. There are cases where it makes sense for the operand type to be wider than the output type. For example, consider the resolution of `{$n :number}`, where the implicit input variable `$n` has a string value `'42'`. In a programming language like JS, it's easier for the custom function implementation to accept that its input may be a number, bigint, string, or object, rather than requiring each of those to come pre-wrapped as suggested by the SHOULD. See steps 6 and 7 [here](https://tc39.es/proposal-intl-messageformat/#sec-messageformat-numberfunctions) for the JS implementation details of this particular case.

That seems to not be ruled out by my original text, since even if the number formatter will never return a string "42", there might be other formatting functions that _do_ just return their inputs, in some cases. 

The goal here is to make a statement about _all_ custom functions, and I think the only logical thing we can say about all of them is that (ignoring options and context), they have the type signature `T -> T`, for some `T`. In JS, `T` would mean something like `number` ∪ `BigInt` ∪ `string` ∪ `object`. That wouldn't need to be explicitly written down in JS, but in C++ (for example), you do need a type to describe the interface that functions need to implement. 

I'm not sure if this is clear, but one reason why it's hard to be precise about is that we're slipping between the object language sense of operand types and output types (as expressed in the specification of the function registry) and the meta-language sense (as expressed in an implementation's description of the calling conventions for functions).

>        I do fully agree that the output of a function is expected to be in a shape that's acceptable as input, but I am not convinced that the input must always have that same shape.

spec/formatting.md
```diff
@@ -222,6 +222,11 @@ the following steps are taken:
 
    The form that resolved _operand_ and _option_ values take is implementation-defined.
 
+   Since the result of a function call can be bound to a _variable_,
+   the output of one _function_ may be the input of another _function_.
```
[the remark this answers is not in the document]
catamorphism: I made a slightly different change instead (in order to avoid implying that an expression has output, which isn't really a concept in the spec); let me know what you think.

spec/formatting.md
catamorphism: > Perhaps: [...]

I took some inspiration from this suggestion, but didn't use it exactly.

spec/formatting.md
```diff
@@ -222,6 +222,61 @@ the following steps are taken:
...
+   to a _variable_.
+   Thus, the output of one _function_ is potentially the _operand_
+   of another _function_. In other words, formatting functions
+   compose with each other.
+   For example, in
+   ```
+   .input {$n :number minIntegerDigits=3}
+   .local {$n1 :number maxFractionDigits=3}
+   ```
+   the second call to `:number` composes with the first call.
```
aphillips, line 235: I think adding the idea of "compose" is going too far. We should be very conservative here. Also, this still has the "output" of a function in play. Perhaps:

```suggestion
Thus, the _operand_ for one _function_ might be the resolved value
of another _function_.
Further, the _options_ for one _expression_ might affect the operation
of another.

>For example, if the value of the variable `n` were `1`:
> ```
> .input {$n :number minimumFractionDigits=1}
> .local $num = {$n :number minimumIntegerDigits=3}
> .match {$num}
> *   {{Prints 001.0 for {$num}}}
> ```
> ... because the _options_ for the `.input` and `.local` are both applied to the value
> for the purposes of both formatting and selection.
> (Note that in English, fractional values match the plural rule `other`)
```

Also note that the `.local` in the original is incorrect.

spec/formatting.md
```diff
@@ -222,6 +222,61 @@ the following steps are taken:
...
+> ```
+> .input {$n :number minIntegerDigits=3}
+> .local {$n1 :number maxFractionDigits=3}
+> {{$n1}}
+> ```
+>
+> is implementation-dependent.
+> Depending on whether the options are preserved across
+> the two calls to `:number`, a conformant implementation
+> could produce either "001.000" or "1.000"
```
aphillips, line 275: I would leave all of this out.

spec/formatting.md
```diff
@@ -222,6 +222,61 @@ the following steps are taken:
...
+> Feedback from users and implementers is desired
+> about whether to require one interpretation or the other
+> in the spec.
```
aphillips, line 278: And I would highlight this...

```suggestion
> [!NOTE]
> During the Technical Preview, feedback on how the registry
> describes how _functions_ inherit resolved values and _options_
> and what requirements this specification should impose
> are highly desired.
```
```

**Review by aphillips — CHANGES_REQUESTED** [review]
I think our comments should be more cautious, even though in general I think we are on the right track here. We also need to describe resolved value handling carefully.

File: spec/formatting.md (hunk 3 above)
Comment (eemeli) [reply spec/formatting.md:235]: Fixed the indents in the above suggestion. Its example should be updated, because we should not be showing any matching (even if only with a single `*` variant) on a non-integral number. Then could also drop the parenthetical and irrelevant bit about `other`.

File: spec/formatting.md [hunk 6]
```diff
@@ -222,6 +222,61 @@ the following steps are taken:
...
+   For example, in
+   ```
+   .input {$n :number minIntegerDigits=3}
+   .local {$n1 :number maxFractionDigits=3}
+   ```
+   the second call to `:number` composes with the first call.
+
+   In addition, selector functions compose with formatting functions
+   in the sense that a selector function's _operand_
+   may be the output of any formatting function.
```
eemeli, line 239: As I mention in https://github.com/unicode-org/message-format-wg/pull/686#discussion_r1500555831, we really should avoid "formatting function" and "selector function" as terms here, given that the resolved value of a function can theoretically be used for both.

This also applies to the next paragraph.

eemeli (commented):
I agree with @aphillips's points above.

spec/formatting.md
eemeli: So the example could read:

```
.input {$n :number minimumFractionDigits=1}
.local $num = {$n :number minimumIntegerDigits=3}
{{Prints 001.0 for {$num}}}
```

stasm: I don't mind using the word "compose" here, but I'm also OK with dropping it and replacing with a description of the observed behavior, like @aphillips proposed.

I would, however, edit @aphillips's suggestion slightly:

```diff
-Thus, the _operand_ for one _function_ might be the resolved value
-of another _function_.
+Thus, the resolved value of one _function_ might be the _operand_
+or an _option_ value for one another _function_.
 Further, the _options_ for one _expression_ might affect the operation of another.
```

This does two things:

* The subject of the sentence is the resolved value, similar to the second sentence.
* It's not only operands that can be resolved values of other expressions; option values can as well, as per our syntax and data model.

spec/formatting.md
```diff
@@ -222,6 +222,61 @@ the following steps are taken:
...
+
+   Implementations SHOULD provide a means for formatting functions
+   to compose with each other
+   and for formatting functions to compose with selector functions.
+   Implementations that provide a means for defining custom functions
+   SHOULD provide a means for those functions to return values
+   that contain enough information
+   (e.g. the resolved _operand_ and _option_ values
+   that the function was called with)
+   to be used as inputs to subsequent function calls.
```
stasm, line 249: This is the key part and I wouldn't want to drop it, as other reviewers suggested. Furthermore, I think we need to say *something somewhere* about resolved values. Otherwise, @aphillips's example from his suggestion above end up being the only place in the spec where we *implicitly* require that resolved values carry some extra information. I'd prefer to be explicit.

spec/formatting.md
```diff
@@ -222,6 +222,61 @@ the following steps are taken:
...
 
+   A _local-declaration_ binds the resolved value of an _expression_
+   to a _variable_.
+   Thus, the output of one _function_ is potentially the _operand_
+   of another _function_. In other words, formatting functions
+   compose with each other.
+   For example, in
+   ```
+   .input {$n :number minIntegerDigits=3}
+   .local {$n1 :number maxFractionDigits=3}
```
stasm, line 233: As @aphillips noticed, the `.local` needs a name. Making this suggestion to make sure we don't miss it, in case the other review comment isn't committed.

```suggestion
   .input {$n :number minIntegerDigits=3}
   .local $x = {$n1 :number maxFractionDigits=3}
```

spec/formatting.md
```diff
@@ -222,6 +222,61 @@ the following steps are taken:
...
+   Such an interface SHOULD define an implementation-specific
+   argument type `T` and return type `U` for custom formatting functions
+   such that `U` can be coerced to `T` without loss of information.
+   The type `U`
+   (or a type that `U` can be coerced to without loss of information)
+   SHOULD also be the input type of custom selector functions.
+
+> [!NOTE]
+> In the Tech Preview, the spec leaves the behavior of the previous
+> example implementation-dependent. Supposing that
```
stasm, line 261: Isn't this PR still in scope of the Tech Preview? Or do you mean that it's implementation-dependent because of the `SHOULD`?

stasm (approved):
In my mind, this PR sufficiently clarifies the expectations of the resolution mechanism to be useful to implementors and to allow early adopters to experiment with function composition.

Note that there's also a mention of resolved values in lines 106-118, similar to the wording in line 223:

https://github.com/unicode-org/message-format-wg/blob/9caacb6fe3f5ac7fb18f0a9530a88a4495906326/spec/formatting.md#L106-L118

spec/formatting.md
aphillips: >  because we should not be showing any matching (even if only with a single * variant) on a non-integral number.

`:number` really does need to do plural matching on fractions and this isn't a problem. Your concern is, I think, about exact matching, which I do not show. I actually think the `.match` is important to call out precisely because the fraction bits happen to `$n`

eemeli: Ah, fair point, I was confused. Including a `one {{This is never selected}}` would clarify things a bit.

spec/formatting.md
aphillips: But we need to avoid being too specific about how it works. "return" has a specific meaning and I don't necessarily think it's a good idea to think of this as a function call's return value.

Our language is a declarative language and we call the stuff in an expression an "annotation" for a reason. Perhaps:

```suggestion
When resolving the value of an _operand_ or other variable
(such as the value of an _option)
implementations SHOULD provide interfaces so that _annotation_
applied in statements can accompany the value where appropriate.
Implementations of _functions_ SHOULD define whether they change the 
value of the _operand_ in any way.
Implementations of _functions_ SHOULD define whether the value of
each _option_ is transitive or local.
```

I mention "statements" here because `.match` can be where the annotation is applied (not just `.local` or `.input`)

Some examples might help here. Here might be an example of non-transitive options (we might say that field options are non-transitive in the spec):

```
.input {$d :datetime weekday=short month=medium day=numeric}
.local $d1 = {$d :datetime hour=|2-digit| minute=numeric}
{{The transaction was on {$d} at {$d1}.}}
```

Here's a similar example:

```
.input {$d :datetime timeZone=|Europe/Paris|}
.local $date = {$d :datetime dateStyle=short}
.local $time = {$date :datetime timeStyle=short}
{{What does {$date} and {$time} print?}}
```

I think it is _less_ surprising if `$time` forgets the earlier style annotation but _not_ the time zone.

We don't currently define any functions that change the value of an operand, but we certainly might do:

```
.local $regular = {|Addison| :string}
.local $shouted = {$regular :transform to=uppercase}
.match {$shouted}
ADDISON {{... is selected... }}
* {{ ... }}
```

spec/formatting.md
aphillips: This PR's text is part of the Tech Preview.

spec/formatting.md
aphillips: LOL. Actually, I had it originally, but removed it before submitting--I thought for clarity!

spec/formatting.md
stasm: I see your point about "calling" and "returning".

> (...) so that _annotation_ applied in statements can accompany the value where appropriate.

This sounds OK, although I'm now not sure about the exact meaning of "the value" here. I realize that this was the whole point of @catamorphism's opening the other PR...  

Btw. I think _expressions_ would be more appropriate than _statements_ — it's possible to annotate inside placeholders, too.

> Implementations of _functions_ SHOULD define whether they change the 
> value of the _operand_ in any way.
> Implementations of _functions_ SHOULD define whether the value of
> each _option_ is transitive or local.

Maybe move this part to the note about seeking feedback? We don't really know yet what defining these constraints and requirements should look like.

spec/formatting.md
catamorphism: "the resolved value   of another _function_" doesn't really make sense here. The resolved value of a _function_ is a thing that represents the function; the resolved value of a function applied to arguments (using more conventional declarative-language terminology) is something left unspecified, but is _not_ a representation of a function itself. The spec doesn't really offer us the language for disambiguating the two.

catamorphism: I'm not sure if replacing the original example with the one with `.match` clarifies the point, although more examples are generally a good thing. I wanted to show an example with two formatting functions, because that's where it might not be obvious that functions compose.

spec/formatting.md
catamorphism: I'm working on a revised version of this, but note that the word "inherit" should be completely off-limits since it connotes object-oriented inheritance, which would just confuse the issue.

spec/formatting.md
```diff
@@ -222,6 +222,61 @@ the following steps are taken:
...
+   the second call to `:number` composes with the first call.
+
+   In addition, selector functions compose with formatting functions
+   in the sense that a selector function's _operand_
+   may be the output of any formatting function.
+
+   Implementations SHOULD provide a means for formatting functions
+   to compose with each other
+   and for formatting functions to compose with selector functions.
+   Implementations that provide a means for defining custom functions
```
macchiati, line 244: I agree with the sense of this. We should also note somewhere that any two functions are not required to meaningfully compose; there is no requirement or expectation that the following make a meaningful composition:
```
.input {$date :datetime}
.local $person = {$date :x:personname}
```

spec/formatting.md
macchiati: We need to make clear that the 'return type' is not the 'thing bound to a variable'. The 'thing bound to a variable' also includes options.

spec/formatting.md
catamorphism: I would also like to avoid using the word "transitive", because transitivity is a property of a mathematical relation and we haven't defined any such relations in the spec.

catamorphism: Re: this example:

```
.input {$d :datetime weekday=short month=medium day=numeric}
.local $d1 = {$d :datetime hour=|2-digit| minute=numeric}
{{The transaction was on {$d} at {$d1}.}}
```

It's not obvious to me that the options from the first `:datetime` call shouldn't be preserved -- should `$d1` be a formatted date with the union of all the options shown in both `:datetime` annotations? Or just the `hour` and `minute` options and defaults for the others? To me there's no "obvious" answer, though maybe it's obvious to people who have more experience with message formatting.

It's certainly worth thinking about what the example means, but I'm not sure if it's the best example if the goal is to show something where certain options obviously shouldn't be preserved.

spec/formatting.md
catamorphism: That's one of the questions, though. If the contract with function implementations is that they are responsible for returning a thing containing all the options they want to preserve, then whatever a function returns _is_ the thing bound to a variable (modulo possible lazy evaluation). If function implementations have no such responsibility, then yes, the formatter has to do additional processing to transform the "thing returned by a function" into the "thing bound to a variable".

spec/formatting.md
```diff
@@ -225,40 +225,41 @@ the following steps are taken:
...
-   MAY define an interface that custom functions implement.
+   to be used as arguments to subsequent calls
+   to the function implementations.
+   For example, a MessageFormat implementation in a typed programming language
+   MAY define an interface for custom function implementations.
    Such an interface SHOULD define an implementation-specific
-   argument type `T` and return type `U` for custom formatting functions
+   argument type `T` and return type `U`
+   for implementations of formatting functions
    such that `U` can be coerced to `T` without loss of information.
```
macchiati, line 250: Look at the following mutating function. Do we expect to be able to extract

.local $foo {$date :extractDay calendar=georgian}
.local $fii {$foo :extractMonth calendar=georgian}

Would the above text mean that  :extractDay SHOULD have a "return type" such that it preserves the month value from $date?

spec/formatting.md
catamorphism: > For #3, we have _not_ delved into what the specification for outbound communication from an RV to an expression (as operand or option value) are. 

I think this is a good way of putting it; as it is, the spec doesn't say anything about how to go from the return value of a function (as implemented in an underlying programming language) to an expression (in MessageFormat). If there's going to be a custom function interface at all, I don't know how to _not_ specify that.

catamorphism:
In the hopes of making the discussion easier to follow, I'll summarize the feedback and how I did or didn't address it in [5803dd8](https://github.com/unicode-org/message-format-wg/pull/686/commits/5803dd8f0ad074cff3fc5f2aa847f3989f1f4a05):

* Fixed the syntax of the example
* Eliminated "compose" and "composability" per @aphillips 
* In language referring to return values, changed "function" to "function implementation" to make it clear that we're speaking in object language rather than metalanguage. I don't know how to express these properties in metalanguage. I don't think this imposes any requirements on the internal structure of resolved values in the implementation. It does make recommendations about the interface between custom functions and the formatter, but I don't see how we can avoid specifying that interface.
* For now, I kept in the passage that @aphillips suggested dropping, but rewrote it to refer to function implementations rather than functions.
* Generalized references to formatting/selection functions to just "functions", per @eemeli 
* Added text saying that the result of one call can also be the option value of another call, per @stasm 
* Added text reiterating that function implementations are free to signal errors for any unhandled input, per @macchiati 's comment about "functions are not required to meaningfully compose" (unfortunately, not using "compose" means I can't say it quite so elegantly).

There's one more comment from @aphillips that I didn't address yet; will do that in another commit.

spec/formatting.md
catamorphism: No; to be concrete, I'll use a simplified version of the C++ implementation.

`:extractDay` and `:extractMonths` would be implemented as classes that provide a `format()` method. The type of that method (ignoring options and context, for simplicity) is:

```
FormattedPlaceholder format(FormattedPlaceholder&& argument);
```

A particular instance of the type `FormattedPlaceholder` could contain any set of options, or none.

Sorry for talking in implementation terms, but I'm not sure how else to say it, given that my goal here is to suggest that implementors not do the "obvious" thing, which would be to define something like a `FormatterInput` and `FormatterOutput` type, which are not coercible to each other, and define the interface like:

```
FormatterOutput format(FormatterInput&& argument);
```

(I would expect every implementation to have some sort of interface between the formatter and calls to custom functions (I say "custom" because built-in functions could be handled inside the formatter), so I think it's meaningful to refer to it in this note. In a unityped implementation language like JavaScript, there's less of a hazard since there's only one type, trivially guaranteeing the property stated in the note.)

spec/formatting.md
```diff
@@ -222,6 +222,74 @@ the following steps are taken:
...
+> The behavior of the previous example is
+> implementation-dependent. Supposing that
+> the external input variable `n` is bound to the string `"1"`,
+> and that the implementation formats to a string,
+> the formatted result of the following message:
+>
+> ```
+> .input {$n :number minIntegerDigits=3}
+> .local {$n1 :number maxFractionDigits=3}
+> {{$n1}}
```
gibson042, line 270: This is not valid syntax. Did you mean the below?
```suggestion
> .input {$n :number minIntegerDigits=3}
> .local $n1 = {$n :number maxFractionDigits=3}
> {{$n1}}
```

spec/formatting.md
catamorphism: On the whole, I understand the suggestion and think it's getting at something useful, but I don't see how it makes sense in the current framework of the spec, in which there is no data model for runtime values. 

I think the first sentence ("...implementations SHOULD provide interfaces so that _annotation_ applied in statements can accompany the value where appropriate.") is too vague, but I'm also not sure how to make it less vague. I would argue that the text in my revised commit is better because it's focused on functions, and the boundary between "inside the formatter" and "in a function implementation" is the one place where information is likely to get "lost". 

I don't get the concept of "changing the value of the operand" even if I mentally replace that with "mapping the operand onto a new operand" or something like that. In your `:transform` example, I would understand `:transform` as returning something like this (if I borrow some of the machinery from #645 but simplify it for presentation):

```
AnnotatedFormattableValue {
  source: AnnotatedFormattableValue { source: Formattable("Addison"), value: FormattedValue("Addison") },
  formatter: "transform",
  options: { "to": "uppercase" },
  value: FormattedValue("ADDISON")
}
```

As with functions in other examples, it passes the same operand through (a wrapped thing ultimately representing the string "Addison"), and the transformed operand (the string "ADDISON") is the "formatted value" thing inside the structure representing the return value.

Obviously this isn't necessarily going to be the data model for runtime values, but still, it's not obvious to me that the result is just "ADDISON" with a bunch of options, rather than something that encapsulates the input "Addison", the output "ADDISON", and the options.

Finally, the sentence about defining options that are or aren't preserved (I would use "preserved" rather than "transitive" as I already said) is important, but we don't have a way in the registry to declare that info, currently.

spec/formatting.md
catamorphism: Thanks, I forgot to correct this error in both places

spec/formatting.md
stasm: +1 to this. I'm not sure if I was able to explain this in yesterday's call, but I definitely agree that there should be no requirement for *any two* functions to meaningfully compose, as you've nicely put it. This is why I've been using the term *cooperative composition* (although I'm happy to call it something else), i.e. the functions must be aware of each other's interface in order to meaningfully compose.

catamorphism:
As a meta-comment, my objective with this PR is to help implementers not paint themselves into a corner, so while it would fit more naturally with the spec to use metalanguage (talking about annotations, expressions, resolved values, etc.) rather than object language, in the absence of the machinery that would be needed for that, I used the escape hatch of object language (e.g. "function implementations") instead.

spec/formatting.md
```diff
@@ -222,6 +222,74 @@ the following steps are taken:
...
+   .local $n1 = {$n :number maxFractionDigits=3}
+   ```
+   the output of the first call to `:number`
+   is the input of the second call to `:number`.
+
+   Implementations that provide a means for defining custom functions
+   SHOULD provide a means for function implementations
+   to return values that contain enough information
+   (e.g. the resolved _operand_ and _option_ values
+   that the function was called with)
```
eemeli, line 242: As this is an example and therefore non-normative this is _ok_, but I'd consider it surprising for a function to provide its input operand & option values rather than the resolved values.

Consider, for example, `{42 :number}`. The resolved value of the operand there is something like the _string_ `'42'`. Does it really make sense for that to be considered the "value" of this expression? Or is it more appropriate to allow and expect for the `:number` implementation to parse that string into a number `42`, and to present that as the value in its output?

spec/formatting.md
```diff
@@ -222,6 +222,74 @@ the following steps are taken:
...
+   Thus, the output of one _function_ is potentially the _operand_
+   of another _function_,
+   or the value of one of the _options_ for another function.
+   For example, in
+   ```
+   .input {$n :number minIntegerDigits=3}
+   .local $n1 = {$n :number maxFractionDigits=3}
+   ```
+   the output of the first call to `:number`
+   is the input of the second call to `:number`.
```
aphillips, line 236: This is saying outputs and inputs, which, as I noted before, isn't correct.

spec/formatting.md
```diff
@@ -222,6 +222,74 @@ the following steps are taken:
 
    The form that resolved _operand_ and _option_ values take is implementation-defined.
 
+   A _local-declaration_ binds the resolved value of an _expression_
```
aphillips: Technically, any declaration (not just the local ones)

spec/formatting.md
```diff
@@ -222,6 +222,74 @@ the following steps are taken:
...
+   For example, a MessageFormat implementation in a typed programming language
+   MAY define an interface for custom function implementations.
+   Such an interface SHOULD define an implementation-specific
+   argument type `T` and return type `U`
+   for implementations of formatting functions
+   such that `U` can be coerced to `T` without loss of information.
+   The type `U`
+   (or a type that `U` can be coerced to without loss of information)
+   SHOULD also be the input type of implementations of
+   custom selector functions.
```
eemeli, line 254: **Blocking:**
The explicit references to "formatting functions" and "selector functions" really do need to get dropped. That also allows the second sentence here to be dropped.

Consider this message:
```
.input {$n :number}
.match {$n}
one {{{$n} is one!!1}}
* {{{$n} is not one.}}
```

Here, the resolved value of `{$n :number}` is used both for selection and formatting. So is the `:number` implementation a "formatting function", or a "selector function"? It is neither, it's just a "function", or if we really must add a qualifier, it's a "message function".

To be clear, it is possible for an implementation to implement this using two separate "formatter" and "selector" functions, but the spec does not and should not mandate this. Instead, such an implementation carries an extra burden e.g. ensuring that any Resolution Error in the resolution of `{$n :number}` is only emitted once, rather than separately for the selector and formatter.

```suggestion
   Such an interface SHOULD define an implementation-specific
   argument type `T` and return type `U`
   for implementations of functions
   such that `U` can be coerced to `T` without loss of information.
```

spec/formatting.md
```diff
@@ -222,6 +222,74 @@ the following steps are taken:
...
+
+   Implementations that provide a means for defining custom functions
+   SHOULD provide a means for function implementations
+   to return values that contain enough information
+   (e.g. the resolved _operand_ and _option_ values
+   that the function was called with)
+   to be used as arguments to subsequent calls
+   to the function implementations.
+   For example, a MessageFormat implementation in a typed programming language
+   MAY define an interface for custom function implementations.
```
aphillips, line 246: This is an example, so it cannot be normative. Also, it probably applies to weakly or untyped programming languages too. How would we test this?

```suggestion
For example, an implementation might define an interface that allow custom function implementation.
```

spec/formatting.md
aphillips: This is still in an example and must not be normative. Even if it weren't an example, it's way too specific to put in the specification.

If we want to provide guidance to users/implementers, instead of pouring stuff into the spec, we should write some user guide material.

eemeli (changes requested):
Overall this is good, but I've left one "Blocking" comment below that needs to be addressed before merging this. If that is done, please don't hesitate to dismiss this review when merging, as I might not be awake at end-of-day Pacific Time.

spec/formatting.md
```diff
@@ -222,6 +222,74 @@ the following steps are taken:
...
+   for implementations of formatting functions
+   such that `U` can be coerced to `T` without loss of information.
+   The type `U`
+   (or a type that `U` can be coerced to without loss of information)
+   SHOULD also be the input type of implementations of
+   custom selector functions.
+   Implementations of specific functions MAY
+   signal errors if supplied an _operand_ that
+   does not make sense for the particular function
+   being implemented.
```
aphillips, line 258: This is wrong guidance. There is a specific error (Invalid Expression) and it should be at least SHOULD emit, as in:

```suggestion
Implementations of a _function_ SHOULD emit an 
_Invalid Expression_ error for _operands_ whose revolve value
or type is not supported.
```

See for example the descriptions in `registry.md`, such as this one for `:string`:

> The operand of :string is either any implementation-defined type that is a string or for which 
> conversion to a string is supported, or any literal value. All other values produce an Invalid Expression 
> error.

aphillips (commented):
(as contributor)

I really believe that less is more here. We're getting into the weeds of how "functions" are "called" and what their "inputs" and "outputs" are. This is imperative programming thinking.

In my opinion, we should focus on calling out "here be dragons" and following up with carefully considered text across the spec (or with user guide like material)

spec/formatting.md
catamorphism: That's a good point. I changed it to: "a representation of the resolved...", which hopefully leaves enough room to make it clear that your example can work that way.

spec/formatting.md
catamorphism: I'm not super comfortable with this change because it implies that a selector has to have a return type that's the same as whatever a formatter returns. At the implementation level, a selector returns a list of keys. I'm not sure how the same implementation (as opposed to two different implementations that happen to be registered to the same name in the function registry) can return both. I haven't looked at other implementations closely besides my own ICU4C one and ICU4J, but in those two implementations, it's clear that these are really different interfaces.

However, I guess one reading of your suggested text is that functions that don't return any value can be defined, so I guess that's compatible at least with my own implementation, where selectors are passed a list of keys to fill in as an out-parameter, and return `void`.

catamorphism:
> (as contributor)
> 
> I really believe that less is more here. We're getting into the weeds of how "functions" are "called" and what their "inputs" and "outputs" are. This is imperative programming thinking.
> 
> In my opinion, we should focus on calling out "here be dragons" and following up with carefully considered text across the spec (or with user guide like material)

I don't quite know how to specify a foreign function interface without talking about inputs and outputs (or arguments and return values). This is not specific to imperative languages; all functional languages have a concept of function application, arguments, and return values.

eemeli (approved):
I believe that my concerns about this PR have been sufficiently addressed.

spec/formatting.md
aphillips: +100 although there are some different things going on here. In @macchiati's example, the types don't match. The `.local` declaration should emit an `Invalid Expression` error because `$date` isn't of a supported type for `:x:personname` (presumably).

In other cases, the types can match but the annotations might not be supported:

```
.input {$date :date style=long}
.local $foo = {$date :time}  // not style=long because style means "dateStyle"
```

For now, we should basically say something like "functions can decide what operands to support" and "functions can decide what functions or function options to support". We permit "composition" without requiring it **_or_** prohibiting it.

In the default registry we can add (in the TP period) some guidance for date/time/datetime and number functions as a guide.

aphillips:
(chair hat)

I have two approvals, which is what I need to merge this. I'm going to make a last call in this comment, ignoring my own review. Any objection to merging this?

spec/formatting.md
catamorphism: I tried to address this in [8a5f589](https://github.com/unicode-org/message-format-wg/pull/686/commits/8a5f5893a866a2bab49c0f64b92948c2ccab01f7)

spec/formatting.md
macchiati: Agreed.

spec/formatting.md
```diff
@@ -222,6 +222,71 @@ the following steps are taken:
...
+   argument type `T` and return type `U`
+   for implementations of functions
+   such that `U` can be coerced to `T` without loss of information.
+   Implementations of a _function_ SHOULD emit an 
+   _Invalid Expression_ error for _operands_ whose resolved value
+   or type is not supported.
+
+> [!NOTE]
+> The behavior of the previous example is
+> implementation-dependent. Supposing that
```
macchiati, line 259:
```suggestion
> currently implementation-dependent. Supposing that
```

I disagree with this note as stated. I agree that the Tech Preview specification of the functions it has is _not sufficient_ to determine what the results of the cited example should be.

But that doesn't mean that our hands are tied from doing that by the time of the stable version in v46, which this seems to do.

spec/formatting.md
```diff
@@ -222,6 +222,71 @@ the following steps are taken:
...
+> and that the implementation formats to a string,
+> the formatted result of the following message:
+>
+> ```
+> .input {$n :number minIntegerDigits=3}
+> .local $n1 = {$n :number maxFractionDigits=3}
+> {{$n1}}
+> ```
+>
+> is implementation-dependent.
```
macchiati, line 270:
```suggestion
> is currently implementation-dependent.
```

spec/formatting.md
catamorphism: Totally agreed.

spec/formatting.md
```diff
@@ -222,6 +222,71 @@ the following steps are taken:
 
    The form that resolved _operand_ and _option_ values take is implementation-defined.
 
+   A _declaration_ binds the resolved value of an _expression_
```
macchiati, line 225: Not asking for any change in the tech preview, but this will definitely need to be revised afterwards.
The whole notion of a 'resolved value' is very muddy, unless it literally means "what is bound to a variable by an expression declaration", which means it *also* includes selection/formatting options and their values whenever those are carried over.

spec/formatting.md
```diff
@@ -222,6 +222,75 @@ the following steps are taken:
...
+   (e.g. a representation of
+   the resolved _operand_ and _option_ values
+   that the function was called with)
+   to be used as arguments to subsequent calls
+   to the function implementations.
+   For example, an implementation might define an interface that allows custom function implementation.
+   Such an interface SHOULD define an implementation-specific
+   argument type `T` and return type `U`
+   for implementations of functions
+   such that `U` can be coerced to `T` without loss of information.
```
macchiati, line 252:
```suggestion
   such that `U` can be coerced to `T`.
```

macchiati (changes requested):

stasm (approved):
LGTM after the most recent updates.

I acknowledge and agree that we need to spend more time and deliberation specifying this fully for 46. That said, I do think that the current working does a good job of giving guidance to implementors and of stating that it's a work-in-progress.

spec/formatting.md
macchiati: I think this needs to be included in the changes before merging.

spec/formatting.md
catamorphism: #645 is an attempt to address this. (But it could go even further.)

macchiati (approved):

PR #685, opened 2024-02-22
Reformat default registry

aphillips:
This PR reformats the default registry consistent with the suggestion in @gibson042's comment [here](https://github.com/unicode-org/message-format-wg/pull/659#discussion_r1493061933)

There are no normative changes.

PR #687, opened 2024-02-23
Require ISO 8601 date or datetime literals

### LINKED ISSUE #680: Restrict literals for `:date` and `:time`
_This is a spin-off from #665, which conflates this issue's nominal topic with a discussion of the `timeZone` options. The text below is an extract of the original by @aphillips._

---

@eemeli noted that some literal values don't make sense with some of the date/time functions. Notably `date` literal with `:time` or `time` literal with `:date`.

In LDML45, I propose:

- `:datetime` accepts XMLSchema `date`, `time`, and `dateTime` types as literal operands.

```
.local $date = {|2024-02-17| :datetime} // don't blow up!
{{Today is {$date}.}}
```

eemeli:
Fixes #680, see discussion there.

The regexp included here is the subset that should be supported by all reasonable implementations.

This may need a rebase once #685 lands.

### DIFF
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -646,13 +646,18 @@ an implementation-defined date/time type
 or a _date/time literal value_, as defined below.
 All other _operand_ values produce an _Invalid Expression_ error.
 
-A **_<dfn>date/time literal value</dfn>_** is a non-empty string consisting of 
-one of the following:
-- an XMLSchema 1.1 [dateTime](https://www.w3.org/TR/xmlschema11-2/#dateTime)
-- an XMLSchema 1.1 [time](https://www.w3.org/TR/xmlschema11-2/#time)
-- an XMLSchema 1.1 [date](https://www.w3.org/TR/xmlschema11-2/#date)
+A **_<dfn>date/time literal value</dfn>_** is a non-empty string consisting of an ISO 8601 date,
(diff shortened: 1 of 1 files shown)


aphillips:
I think I oppose this change as written. We need to discuss as a group. 😿

spec/registry.md
```diff
@@ -885,14 +885,19 @@ an implementation-defined date/time type
...
-A **_<dfn>date/time literal value</dfn>_** is a non-empty string consisting of 
-one of the following:
-- an XMLSchema 1.1 [dateTime](https://www.w3.org/TR/xmlschema11-2/#dateTime)
-- an XMLSchema 1.1 [time](https://www.w3.org/TR/xmlschema11-2/#time)
-- an XMLSchema 1.1 [date](https://www.w3.org/TR/xmlschema11-2/#date)
-
-The `timezoneOffset` of each of these formats is optional. 
-When the offset is not present, implementations should use a floating time type
+A **_<dfn>date/time literal value</dfn>_** is a non-empty string consisting of an ISO 8601 date,
+or an ISO 8601 datetime optionally followed by a timezone offset.
```
sffc, line 889: Issue: why only a timezone offset? In general IANA time zone annotations should be used instead.

Did you consider citing IXDTF?

sffc (changes requested):

PR #691, opened 2024-02-26
Note the need to quote `2-digit`

aphillips:
Per 2024-02-26 teleconference, add a note about the value `2-digit`

### DIFF
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -786,6 +786,14 @@ appear in the formatted output.
 
 The _field options_ are defined as follows:
 
+> [!IMPORTANT]
+> The value `2-digit` for some _field options_ **_must_** be quoted
+> in the MessageFormat syntax.
+> ```
+> .local $correct = {$someDate :datetime year=|2-digit|}
+> .local $syntaxError = {$someDate :datetime year=2-digit}
+> ```
+
 The function `:datetime` has the following options:
 - `weekday`
   - `long`
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -787,7 +787,7 @@ appear in the formatted output.
 The _field options_ are defined as follows:
 
 > [!IMPORTANT]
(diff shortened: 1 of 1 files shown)


catamorphism (approved):
Looks good. I had the thought that maybe it's worth noting that the reason you have to quote it isn't specific to the string "2-digit", but is because the character '-' can't appear in an unquoted literal. But it might be better to keep it brief. Leaving that up to you.

echeran (approved):
Ditto @catamorphism's comment

aphillips:
> maybe it's worth noting that the reason you have to quote it isn't specific to the string "2-digit", but is because the 
> character '-' can't appear in an unquoted literal. But it might be better to keep it brief. Leaving that up to you.

Only that's not the reason. The character `-` is definitely permitted in values (but not at the start except for numbers.) The reason is that you cannot start with a digit except for `number-literals`. Participants will note that I argued long and hard for starts-with-digit (and lost) > 1 year ago.

spec/registry.md
```diff
@@ -786,6 +786,14 @@ appear in the formatted output.
 
 The _field options_ are defined as follows:
 
+> [!IMPORTANT]
+> The value `2-digit` for some _field options_ **_must_** be quoted
```
eemeli, line 790: We use bold+italic for term definitions.

```suggestion
> The value `2-digit` for some _field options_ **must** be quoted
```

eemeli (approved):

spec/registry.md
```diff
@@ -786,6 +786,14 @@ appear in the formatted output.
 
 The _field options_ are defined as follows:
 
+> [!IMPORTANT]
+> The value `2-digit` for some _field options_ **must** be quoted
+> in the MessageFormat syntax.
```
aphillips, line 791:
```suggestion
> in the MessageFormat syntax because it starts with a number
> but does not match the `number-literal` production in the ABNF.
```

PR #696, opened 2024-02-28
Add tests from messageformat@4.0.0-7

eemeli:
Update the test suite with new and changed tests to match the latest spec, as a copy from https://github.com/messageformat/messageformat/tree/messageformat@4.0.0-7/packages/mf2-messageformat/src/__fixtures.

These add tests for markup-close with options, `:date`, `:time`, and `:datetime`, as well as various corner cases encountered during development.


test/test-functions.json
```diff
@@ -1,4 +1,85 @@
 {
+  "date": [
+    { "src": "{:date}", "exp": "{:date}", "errors": [{ "type": "bad-input" }] },
+    {
+      "src": "{horse :date}",
+      "exp": "{|horse|}",
+      "errors": [{ "name": "RangeError" }]
```
aphillips, line 7: I believe the spec says this should be an invalid expression error. Your code might call this a `RangeError`. Fix?

eemeli: Ah, true. I currently don't add a wrapper for the built in JS RangeError that's the cause here, as the `:number` spec requires.

test/syntax-errors.json
```diff
@@ -17,6 +16,10 @@
   "missing space {|foo|:func}",
   "missing space {|foo|@bar}",
   "missing space {:func@bar}",
+  "missing space {:func @bar@baz}",
+  "missing space {:func @bar=42@baz}",
+  "missing space {+reserved@bar}",
+  "missing space {&private@bar}",
```
gibson042, line 22: These don't seem like "missing space".
```suggestion
  "missing equals {:func @bar@baz}",
  "missing missing quotes {:func @bar=42@baz}",
  "reserved {+reserved@bar}",
  "private {&private@bar}",
```

mihnita:
Suggestion (does not have to be in this PR): unify schemas?

A good target would be the one used for functions.

For example `syntax-errors`:
```
{
    "errors" : [
        {
           "src": "missing space {|foo|:func}",
        }
    ]
}
```

And `core`:
```
{
    "errors" : [
       ... the current cases
   ]
}
```

test/syntax-errors.json
eemeli: All of those can be turned into valid syntax with the addition of a single space:
```suggestion
  "missing space {:func @bar @baz}",
  "missing space {:func @bar=42 @baz}",
  "missing space {+reserved @bar}",
  "missing space {&private @bar}",
```
(the above is not intended as an actual suggestion, just to show a nice diff)

eemeli:
> Suggestion (does not have to be in this PR): unify schemas?
> 
> A good target would be the one used for functions.

Not really sure what you mean here. At least for now, all of the `syntax-errors.json` entries are spec'd just to produce a syntax or data model error, and formatting those messages should never even get to calling any functions. The errors in the other files are runtime errors.

mihnita:
> Not really sure what you mean here. 
> At least for now, all of the syntax-errors.json entries are spec'd just to produce a syntax or data model error

What I mean is that `syntax-errors.json` contains a list of strings that should fail parsing.
But there is no way to tell what is the error they fail with.

For example:
```json
[
  ...
  "bad {:placeholder option=x option=x}",
  ...
  ".input {$foo} .input {$foo} {{_}}"
  ...
]
```
vs.
```json
{
  "syntax-errors": [
    ...
  ],
  "data-model-errors": [      
    {
      "src": "bad {:placeholder option=x option=x}",
      "errors": [{ "type": "duplicate-option-name" }]
    }
    {
      "src": ".input {$foo} .input {$foo} {{_}}",
      "errors": [{ "type": "duplicate-declaration" }]
    }     
  ]
}
```
---

The second benefit is that there is one single schema for all test files.
Making it easier to parse (at least in some languages :-).

eemeli:
@aphillips Could this be merged, so that further work on the tests can build on top of it, such as #707 and #708? I understood that to be the conclusion reached during the last call.

mihnita (approved):

test/syntax-errors.json
aphillips: Technically, though, they are invalid name tests. You created them by removing non-optional spaces, of course.

aphillips:
Per our conversation, we were waiting on people not named @aphillips and @eemeli to review before merging. But we're holding up additional work by retaining this as a PR.

PR #699, opened 2024-03-02
Drop obsolete mention of "kind" in function-annotation JSON Schema definition

### LINKED ISSUE #698: [FEEDBACK] Mistake in JSON schema
This definition in https://github.com/unicode-org/message-format-wg/blob/main/spec/data-model/message.json makes no sense:
```
    "function-annotation": {
      "type": "object",
      "properties": {
        "type": { "const": "function" },
        "name": { "type": "string" },
        "options": { "$ref": "#/$defs/options" }
      },
      "required": ["type", "kind", "name"]
    },
```

The `required` line does not list the property `options` (why should an array be optional?) and references another property `kind`.

Suggestion: Replace
```
      "required": ["type", "kind", "name"]
```
with
```
      "required": ["type", "name", "options"]
```

eemeli:
Fixes #698, CC @bhaible

This was missed in #574.

### DIFF
--- a/spec/data-model/message.json
+++ b/spec/data-model/message.json
@@ -61,7 +61,7 @@
         "name": { "type": "string" },
         "options": { "$ref": "#/$defs/options" }
       },
-      "required": ["type", "kind", "name"]
+      "required": ["type", "name"]
     },
     "unsupported-annotation": {
       "type": "object",


aphillips (approved):
Yeah, this is an obvious bug.

PR #710, opened 2024-03-07
Clarify (again) that Duplicate Option Name is a data model error

eemeli:
As noted in https://github.com/unicode-org/message-format-wg/pull/704#pullrequestreview-1921417562, this was already fixed once in #577, but mistakenly reverted in the merge of #593.

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -324,8 +324,6 @@ The result of resolving _option_ values is a mapping of string identifiers to va
 For each _option_:
 
 - Resolve the _identifier_ of the _option_.
-- If the _option_'s _identifier_ already exists in the resolved mapping of _options_,
-   emit a _Duplicate Option Name_ error.
 - If the _option_'s right-hand side successfully resolves to a value,
    bind the _identifier_ of the _option_ to the resolved value in the mapping.
 - Otherwise, bind the _identifier_ of the _option_ to an unresolved value in the mapping.


spec/formatting.md
```diff
@@ -324,8 +324,6 @@ The result of resolving _option_ values is a mapping of string identifiers to va
 For each _option_:
 
 - Resolve the _identifier_ of the _option_.
-- If the _option_'s _identifier_ already exists in the resolved mapping of _options_,
-   emit a _Duplicate Option Name_ error.
```
aphillips, line 328: _Duplicate Option Name_ **_is_** a data model error?

eemeli: The current formatting spec language assumes that it's dealing with a _valid_ message, i.e. one with no data model errors. [Duplicate Option Name](https://github.com/unicode-org/message-format-wg/blob/main/spec/errors.md#duplicate-option-name) errors are primarily defined in `errors.md`, but for messages coming from syntax, they're also defined here: https://github.com/unicode-org/message-format-wg/blob/e76196481b23e6e9245923a1239282e19484efd0/spec/syntax.md?plain=1#L567-L568
and as with all other data model errors they're required to be detected and emitted before formatting: https://github.com/unicode-org/message-format-wg/blob/e76196481b23e6e9245923a1239282e19484efd0/spec/errors.md?plain=1#L11-L14

So in order to get to this point in formatting, the current spec text must presume that this error has already been checked, and that therefore the condition for triggering it can't happen. So this statement should be removed to remove the current uncertainty it presents.

We also explicitly only allow for Resolution and Formatting errors to be ignored in expressions that aren't being formatted, and _not_ data model errors: https://github.com/unicode-org/message-format-wg/blob/e76196481b23e6e9245923a1239282e19484efd0/spec/errors.md?plain=1#L19-L21

aphillips: Thanks @eemeli. You're right, we do have that text in `errors.md`, etc.

I guess what I'm grappling with is: did we make the right choice here. @catamorphism and @mihnita might weigh in about how they've implemented the checks required by data model (that is, when do they check for duplicate options or missing variant keys or duplicate declarations). I'm fine with blowing up the message for all of these with the possible exception of duplicate option names (and even there I still want to blow up)

aphillips (approved):

spec/formatting.md
catamorphism: Like I mentioned in #703 (just repeating it here to make the threads easier to follow) my implementation does a "static checking" pass for data model errors that are not checked during parsing, which is consistent with what @eemeli is saying.

aphillips:
Need one more approval to commit

catamorphism (approved):
Instead of deleting the text, I'd like to replace it with something like "assert that the option doesn't already exist". However, that only makes sense if #703 lands, since otherwise, the implementation would be free to recover from the previously-emitted data model error. So this is fine, but I added a note to that effect in #703.

macchiati (approved):
As I've said, I think the current spec is much too proscriptive about the exact "phase" in which an error must be emitted, because requires a particular implementation model.

So removing these kinds of lines is definitely the right direction.

PR #714, opened 2024-03-09
Remove tests that require non-integer matching

catamorphism:
These tests may be able to be re-added after resolving https://github.com/unicode-org/message-format-wg/issues/675


aphillips (approved):
Let's get this in. Need one more approve.

macchiati (approved):

PR #719, opened 2024-03-11
Remove references to "RangeError" in spec tests

catamorphism:
I'm presuming that `RangeError` is supposed to be `bad-input`.

In general #706 is still needed: e.g. it's not clear that `bad-input` refers to an ["Operand Mismatch Error"](https://github.com/unicode-org/message-format-wg/blob/main/spec/errors.md#invalid-expression), but I'm assuming so. For now, this PR just tries to make the test suite internally consistent.

### DIFF
--- a/test/test-functions.json
+++ b/test/test-functions.json
@@ -4,7 +4,7 @@
     {
       "src": "{horse :date}",
       "exp": "{|horse|}",
-      "errors": [{ "name": "RangeError" }]
+      "errors": [{ "name": "bad-input" }]
     },
     { "src": "{|2006-01-02| :date}", "exp": "1/2/06" },
     { "src": "{|2006-01-02T15:04:06| :date}", "exp": "1/2/06" },
@@ -23,7 +23,7 @@
     {
       "src": "{horse :time}",
       "exp": "{|horse|}",
-      "errors": [{ "name": "RangeError" }]
+      "errors": [{ "name": "bad-input" }]
     },
     { "src": "{|2006-01-02T15:04:06| :time}", "exp": "3:04 PM" },
     {
--- a/test/test-functions.json
+++ b/test/test-functions.json
@@ -23,7 +23,7 @@
     {
       "src": "{horse :time}",
       "exp": "{|horse|}",
-      "errors": [{ "name": "bad-input" }]
+      "errors": [{ "type": "bad-input" }]
     },
     { "src": "{|2006-01-02T15:04:06| :time}", "exp": "3:04 PM" },
     {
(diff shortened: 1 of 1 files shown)


test/test-functions.json
```diff
@@ -4,7 +4,7 @@
     {
       "src": "{horse :date}",
       "exp": "{|horse|}",
-      "errors": [{ "name": "RangeError" }]
+      "errors": [{ "name": "bad-input" }]
```
eemeli, line 7:
```suggestion
      "errors": [{ "type": "bad-input" }]
```

test/test-functions.json
```diff
@@ -23,7 +23,7 @@
     {
       "src": "{horse :time}",
       "exp": "{|horse|}",
-      "errors": [{ "name": "RangeError" }]
+      "errors": [{ "name": "bad-input" }]
```
eemeli, line 26:
```suggestion
      "errors": [{ "type": "bad-input" }]
```

eemeli (approved):
Should use `"type"` rather than `"name"` to be internally consistent.

test/test-functions.json
```diff
@@ -4,7 +4,7 @@
     {
       "src": "{horse :date}",
       "exp": "{|horse|}",
-      "errors": [{ "name": "RangeError" }]
+      "errors": [{ "type": "bad-input" }]
```
aphillips, line 7: Our specification has a specific name for this. It is _Operand Mismatch Error_. We should use some formulation of that, perhaps:

```suggestion
      "errors": [{ "name": "OperandMismatchError" }]
```

test/test-functions.json
```diff
@@ -23,7 +23,7 @@
     {
       "src": "{horse :time}",
       "exp": "{|horse|}",
-      "errors": [{ "name": "RangeError" }]
+      "errors": [{ "type": "bad-input" }]
```
aphillips, line 26: ibid

aphillips (changes requested):

test/test-functions.json
eemeli: That sounds like something to apply separately from this PR, for all the `type: 'bad-input'` errors.

aphillips: Do you disagree with my observation?

Also, some "bad-inputs" might be syntax errors (I'd have to look to know).

eemeli: I think the right context for that discussion in #706, because in the absence of a defined list of error codes, it's hard to say whether `bad-input`, `OperandMismatchError`, or maybe `operand-mismatch` is the right code to use.

If any of the current `bad-input` errors are indeed syntax errors, that should be fixed separately.

eemeli: I think we should merge this small, limited fix-up PR in its current state, and work separately towards resolving #706 either iteratively or holistically.

PR #727, opened 2024-03-13
Disallow U+3000 as the last character of a reserved-body.

### LINKED ISSUE #725: [FEEDBACK] syntax: an ambiguity regarding resolved-body in expression
There is an ambiguity in
https://github.com/unicode-org/message-format-wg/blob/main/spec/syntax.md
and
https://github.com/unicode-org/message-format-wg/blob/main/spec/message.abnf

Inside an `expression`, the `annotation` may be followed by whitespace:
```
expression            = literal-expression
                      / variable-expression
                      / annotation-expression
literal-expression    = "{" [s] literal [s annotation] *(s attribute) [s] "}"
variable-expression   = "{" [s] variable [s annotation] *(s attribute) [s] "}"
annotation-expression = "{" [s] annotation *(s attribute) [s] "}"
```

An `annotation` of type `private-use-annotation` or `reserved-annotation` ends with a `resolved-body`:
```
reserved-annotation       = reserved-annotation-start reserved-body
private-use-annotation    = private-start reserved-body
```

A `resolved-body` may end with U+3000:
```
reserved-body             = *([s] 1*(reserved-char / reserved-escape / quoted))
reserved-char     = content-char / "."
```
Examples (using \u escapes for legibility): The input strings
```
{ % foo bar \u3000\u3000 }
```
and
```
{ & foo bar \u3000\u3000 @x }
```
contain a `resolved-body` as part of the `annotation` in the `expression`.

How does it get parsed? There are 3 possibilities:
```
' foo bar \u3000\u3000' parsed as reserved-body
' ' parsed as [s]

' foo bar \u3000' parsed as reserved-body
'\u3000 ' parsed as [s]

' foo bar' parsed as reserved-body
' \u3000\u3000 ' parsed as [s]
```

It appears that the contents of the `reserved-body` is meant to appear as the `body` field of an `UnsupportedStatement` element in the data model (cf. https://github.com/unicode-org/message-format-wg/blob/main/spec/data-model/README.md ). Therefore it matters which of these 3 possibilities the parser chooses.

Please, specify how this ambiguity should be resolved.

bhaible:
In all places where a 'reserved-body' can occur:
  - in a reserved-statement,
  - after a reserved-annotation or private-use-annotation, in a literal-expression, variable-expression, or annotation-expression, it can be followed by whitespace ('s' nonterminal).

A syntax ambiguity exists, because - as reported in #721 and #725 - a U+3000 character can occur as the last character of a 'reserved-body' (via a 'reserved-char') and also as first character of whitespace ('s' nonterminal).

According to the principles explained in #725, it is not desired that a 'reserved-body' ends with a U+3000 character; rather, the U+3000 character is meant to be interpreted as part of the following whitespace.

Test cases (written with \u escapes, for legibility): For reserved-statement:
.regex   /foo/\u3000\u3000{xyz}{{hello}}
For reserved-annotation:
{ % foo bar \u3000\u3000 }
For private-use-annotation:
{ & foo bar \u3000\u3000 @x }

This patch removes this ambiguity, by disallowing U+3000 as the last character of a 'reserved-body'.

It thus fixes #725 and the second part of #721.

Details:
  - U+3000 gets removed from 'content-char' and 'reserved-char'.
  - Whereas simple-start-char, text-char, quoted-char stay the same (since U+3000 is already part of 's').

### DIFF
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -95,7 +95,8 @@ content-char      = %x01-08        ; omit NULL (%x00), HTAB (%x09) and LF (%x0A)
                   / %x2F-3F        ; omit @ (%x40)
                   / %x41-5B        ; omit \ (%x5C)
                   / %x5D-7A        ; omit { | } (%x7B-7D)
-                  / %x7E-D7FF      ; omit surrogates
+                  / %x7E-2FFF      ; omit IDEOGRAPHIC SPACE (%x3000)
+                  / %x3001-D7FF    ; omit surrogates
                   / %xE000-10FFFF
 
 ; Character escapes
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -311,7 +311,8 @@ content-char      = %x01-08        ; omit NULL (%x00), HTAB (%x09) and LF (%x0A)
                   / %x2F-3F        ; omit @ (%x40)
                   / %x41-5B        ; omit \ (%x5C)
                   / %x5D-7A        ; omit { | } (%x7B-7D)
-                  / %x7E-D7FF      ; omit surrogates
+                  / %x7E-2FFF      ; omit IDEOGRAPHIC SPACE (%x3000)
+                  / %x3001-D7FF    ; omit surrogates
                   / %xE000-10FFFF
 ```
 


aphillips:
The appears to be an obvious oversight in the addition of U+3000 to whitespace.

aphillips (approved):

PR #728, opened 2024-03-14
Add Resolved Values and Function Handler sections to formatting

### LINKED ISSUE #678: Define `resolved value` formally
There have been multiple comments on the need to define the term "resolved value", which we use formally but have defined somewhat informally.

eemeli:
This PR is effectively an extended comment on the ongoing discussion in #515 / #645 / #646. It also closes #678.

The intent with the change proposed here is to keep our formatting spec close to its current shape, but more clearly define what "resolved values" are, and via a non-normative example guide the reader towards a mental model of what a "formattable" or "selectable" value might look like.

I _think_ this could be included already in LDML45 as its doesn't actually change anything, but that's of course presuming a lot of consensus.


spec/formatting.md
```diff
@@ -85,10 +85,16 @@ At a minimum, it includes:
 
 Implementations MAY include additional fields in their _formatting context_.
 
-## Expression and Markup Resolution
+## Resolved Values
 
-_Expressions_ are used in _declarations_, _selectors_, and _patterns_.
-_Markup_ is only used in _patterns_.
+This specification allows for the same value to be used for:
```
aphillips, line 90:
```suggestion
This specification allows for the same variable reference to be used in multiple parts of a _message_:
```

spec/formatting.md
```diff
@@ -85,10 +85,16 @@ At a minimum, it includes:
...
+- formatting in a _placeholder_,
+- selection in a _selector_ _expression_,
+- as the _operand_ of another _expression_, or
```
aphillips, line 93: I think we want to mention declarations explicitly. I think "another" can just be "an" (another presumes there is already an expression...)

```suggestion
- as the _operand_ of an _expression_ (including of a _declaration_), or
```

spec/formatting.md
```diff
@@ -106,16 +112,44 @@ In a _pattern_, the resolved value of an _expression_ or _markup_ is used in its
...
-> Alternatively, it could be an instance of an ICU4J `FormattedNumber`,
-> or some other locally appropriate value.
+> With this approach:
+> - An _expression_ could be used as a _placeholder_ if
+>   calling the `formatToString()` or `formatToX()` method of its _resolved value_
+>   did not throw an error.
+> - An _expression_ could be used as a _selector_ _expression_ if
+>   calling the `selectKeys(keys)` method of its _resolved value_
+>   did not throw an error.
+> - An _expression_ could be used as an _operand_ or _option_ value if
```
aphillips, line 136: Expression here is wrong? You can't put an _expression_ as an operand or option value, but you can reference the annotated value.

Would this actually be "the resolved value of an expression could be used..."?

spec/formatting.md
```diff
@@ -106,16 +112,44 @@ In a _pattern_, the resolved value of an _expression_ or _markup_ is used in its
 The form that resolved values take is implementation-dependent,
 and different implementations MAY choose to perform different levels of resolution.
 
-> For example, the resolved value of the _expression_ `{|0.40| :number style=percent}`
-> could be an object such as
+> While this specification does not require it,
+> a _resolved value_ could be implemented by requiring each function implementation to
+> return a value matching the following interface:
```
aphillips, line 117: I would make this more of a normal example. Perhaps:

```suggestion
> [!NOTE]
> This is an illustration of how an implementation might handle these requirements.
>
> An implementation might provide an interface similar to the following
> that each _function_ is required to return:
```

aphillips (commented):
Thanks for taking at stab at this.

spec/formatting.md
eemeli: Would that be clearer, though? Variable references are the mechanism that's used, by the key point here is that it's the same _value_ that's being used for multiple different purposes.

spec/formatting.md
eemeli: Why call out declarations as anything special here? Exactly the same sort of processing is done for example in both of the following equivalent messages:
```
.local $foo = {42}
.local $bar = {$foo :number}
{{The answer is {$bar}}}
```

```
.local $foo = {42}
{{The answer is {$foo :number}}}
```

spec/formatting.md
eemeli: I'm fine with rephrasing this, but two thoughts:
1. The "This is an illustration..." phrase seems a bit out of place here.
2. Saying that this is what a _resolved value_ could look like is kinda the key point here, and should not be lost.

spec/formatting.md
eemeli:
```suggestion
> - Using a _variable reference_, the resolved value of an _expression_
>   could be used as an _operand_ or _option_ value if
```

spec/formatting.md
aphillips: I think these are saying kind of the same thing. The thing that's getting us into trouble is an (apparently?) slippery meaning of the word "value".

spec/formatting.md
aphillips: The reason is so that declarations are mentioned explicitly. When I made the mental checklist of "places expressions can appear", declarations were missing. This also serves as a reminder of how an expression gets to be the value of an option or operand (as we don't permit nesting).

catamorphism (commented):
I know that this is non-normative text, but my conclusion after reading it is that there should be a design doc.

Some of it _should be_ normative: specifically, the question of what functions take and return, which has to be answered (IMO) for the registry spec to make sense. How to describe that takes some thought, as I said in one of the comments. The task is to design a foreign function interface that's abstract enough to be adapted to a range of implementation languages, while concrete enough to make the spec something that can be implemented in a way that affords confidence that an implementation conforms to the spec. That's not easy.

Given the amount of discussion the question of "what is the domain and range of a function?" has generated, and how closely coupled that question is with "resolved values", I think it's worth a design doc. This could be #645 or someone could start afresh.

macchiati:
I like the direction this is going, but I think we really never want to use
the naked 'value' term. There are different kinds of resolved values, and
it is important to keep them separate to prevent confusion. The value of
{|3| :number minSignificantDigits=4} is **not** 3, nor is it 3.000.

But high bit, this is a difficult issue, and if we rush it it is not
unlikely that the spec for tech preview will be worsened. At this point, I
think we would almost be better off having a NOTE explaining that we will
be revamping this section to clarify the language in the following ways: ...

aphillips:
In to 2024-03-25 call we agreed to work on this in the Tech Preview period.

spec/formatting.md
bearfriend: Right, on first read it may seem like "value" means "resolved value". It's complicated by placeholders not being expressions. Otherwise you could say "allows for identical expressions to be used for:"

spec/formatting.md
macchiati: +1

aphillips:
In the 2024-05-06 call, we asked @mihnita to review this before discussing further.

spec/formatting.md
```diff
@@ -106,16 +112,45 @@ In a _pattern_, the resolved value of an _expression_ or _markup_ is used in its
...
 >
+> ```ts
+> interface MessageValue {
```
mihnita, line 120: Overall the direction seems OK.
But I am uneasy with specifying this interface.
Because:
- in "intertwines" even more the formatters and selectors (which I always maintained are different concepts, with different interfaces)
- it is just specified, without explaining how to use it, how it works.

So it feels over-specified.

mihnita:
I think this update strongly conflicts with #645 (_"[DESIGN] dataflow for composability"_)

That one proposes to throw away "resolved values" as too ambiguous (_"In the current spec, a "resolved value" can be a nameable value, formattable value, or preformatted value, depending on context."_)

This PR seems to make resolved values even more ambiguous, while in the same time restricting the ability of implementations to implement things in a different (cleaner) way.

catamorphism:
I experimented with something like the suggested `MessageValue` interface in the ICU4C implementation, in a branch. Despite my previous comments, I think it's pretty good except for: `resolvedOptions(): { [key: string]: unknown }`. I think it should be made clearer that an option can be a value with options. I realize that in TypeScript, that's already implicit, but someone coming from a different implementation language reading this might assume that `unknown` is any implementation-specific type and not notice that it has to be something with options (in order to make examples like the one from #515 work).

So I suggest that instead of `unknown`, it should be: `resolvedOptions(): { [key: string]: MessageValue }`.

spec/formatting.md
```diff
@@ -106,16 +112,45 @@ In a _pattern_, the resolved value of an _expression_ or _markup_ is used in its
...
+>   formatToString(): string
+>   formatToX(): X // where X is an implementation-defined type
+>   getValue(): unknown
+>   resolvedOptions(): { [key: string]: unknown }
```
eemeli, line 124: As suggested by @catamorphism in https://github.com/unicode-org/message-format-wg/pull/728#issuecomment-2174083395

```suggestion
>   resolvedOptions(): { [key: string]: MessageValue }
```

spec/formatting.md
```diff
@@ -106,16 +112,45 @@ In a _pattern_, the resolved value of an _expression_ or _markup_ is used in its
...
-> { value: Number('0.40'),
->   formatter: NumberFormat(locale, { style: 'percent' }) }
-> ```
 >
-> Alternatively, it could be an instance of an ICU4J `FormattedNumber`,
-> or some other locally appropriate value.
+> With this approach:
+> - An _expression_ could be used as a _placeholder_ if
+>   calling the `formatToString()` or `formatToX()` method of its _resolved value_
+>   did not throw an error.
```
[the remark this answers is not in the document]
eemeli: You're right; the spec currently tends to use "emit".

```suggestion
>   did not emit an error.
```

spec/formatting.md
```diff
@@ -106,16 +112,45 @@ In a _pattern_, the resolved value of an _expression_ or _markup_ is used in its
...
+> - Using a _variable reference_, the resolved value of an _expression_
+>   could be used as an _operand_ or _option_ value if
+>   calling the `getValue()` method of its _resolved value_ did not throw an error.
+>   In this use case, the `resolvedOptions()` method could also
+>   provide a set of option values that could be taken into account by the called function.
+>
+> Extensions of the base `MessageValue` interface could be provided for different data types,
+> such as numbers or strings,
+> for which the `unknown` return types of `getValue()` and `resolvedOptions()`
```
eemeli, line 144:
```suggestion
> for which the `unknown` return type of `getValue()` and
> the generic `MessageValue` type used in `resolvedOptions()`
```

catamorphism (approved):
I tried changing ICU4C to use a value representation that's similar to this, and it works. The change in the PR shows what machinery needs to be there in order to implement examples like the one in #515. In an impl that uses the proposed `MessageValue` interface, it should be possible to write custom functions that use options that have options, and so on. Maybe we can get to a more normative version of this in the future, but I think this is a good step.

spec/formatting.md
eemeli:
```suggestion
- as the _operand_ of another _expression_ (including _local declarations_), or
```

spec/formatting.md
```diff
@@ -85,10 +85,16 @@ At a minimum, it includes:
 
 Implementations MAY include additional fields in their _formatting context_.
 
-## Expression and Markup Resolution
+## Resolved Values
 
-_Expressions_ are used in _declarations_, _selectors_, and _patterns_.
-_Markup_ is only used in _patterns_.
+This specification allows for the same value to be used for:
+- formatting in a _placeholder_,
```
stasm, line 91: That's not right, I think. The *result* of formatting is a string or some kind of a "formatted part" type. When talking about resolved values, I think we want to identify the intermediate type that's the result of evaluating the placeholder, before it's formatted.

spec/formatting.md
```diff
@@ -85,10 +85,16 @@ At a minimum, it includes:
...
+- selection in a _selector_ _expression_,
+- as the _operand_ of another _expression_ (including _local declarations_), or
+- as an _option_ value in another _expression_,
```
stasm, line 94:
```suggestion
This specification allows for the same interface to be used as:
- a result of evaluating a _placeholder_, before it's formatted,
- a result of evaluating a _selector_ _expression_, before it's selected on,
- as the _operand_ of another _expression_ (including _local declarations_), or
- as an _option_ value in another _expression_,
```

spec/formatting.md
```diff
@@ -106,16 +112,46 @@ In a _pattern_, the resolved value of an _expression_ or _markup_ is used in its
...
-> For example, the resolved value of the _expression_ `{|0.40| :number style=percent}`
-> could be an object such as
+> While this specification does not require it,
+> a _resolved value_ could be implemented by requiring each function implementation to
+> return a value matching the following interface:
 >
+> ```ts
+> interface MessageValue {
+>   formatToString(): string
+>   formatToX(): X // where X is an implementation-defined type
```
stasm, line 122: I may have missed some discussions about this, but why wouldn't we want to be more specific here about the parts?

```suggestion
>   formatToParts(): IterableIterator<MessagePart>; // MessagePart is implementation-defined.
```

stasm (changes requested):

stasm:
(My review comes in late. I'm OK with its not being considered blocking this PR. I can file a new PR if you prefer to discuss my suggestions elsewhere.)

spec/formatting.md
aphillips: I think the key thing here is the difference between the _output_ of an expression and the _evaluation_ of an expression. The "resolved value" is only important (or even visible) when it will be consumed by another expression later.

I think @stasm is correct that the output of formatting (or selection) is not the "resolved value". In the case of formatting, it is the string representation (or "formatted foo" parts representation), which is different from the "resolved value". Similarly, the output of selection is a filtered and stack ranked list of patterns (which is definitely not the "resolved value").

The "resolved value", to me, means "what is the value of the operand _after_ the expression has been evaluated".

If I have a message `You have {$x :integer} wildebeest`, the _value_ of `$x` is constrained to be an implementation-defined numeric type (it might not be an integer type, even though the formatting function formats it like one), but the output of the expression is a string.

I think that each function needs to explicitly say whether it "covers up" (changes) the operand's resolved value and what it changes it to be. Using my example:

> Function: `:integer`
> * Operand: an implementation-defined type or a literal whose contents match the number-literal production in the ABNF
> * Format output: a string
> * Selection output: number selector (see design number-selection)
> * Resolved value: an implementation-defined numeric type

Notice that this eliminates the string representation between the operand (input) and resolved value. It also, probably, allows the implementation defined type to be changed by the function.

macchiati: > Operand: an implementation-defined type or a literal whose contents match the number-literal production in the ABNF

There is a third kind of entity that can be an operand.

An operand is a defined type, or literal, _or the "result of a previous functional application"_ (call that a **function-result** for now). A function-result is (logically) not just an implementation-defined datatype. If so, it wouldn't have access to the information that any composing function would need.

Instead, I think the best conceptual model of a function-result is that of an object with methods to get information. Such as (of course, the names can be chosen to protect the innocent):

- getSourceFunction() —the function used to create it
- getSourceOperand() — the operand used to create it
- getsourceOptions() — the options used to create it
- getFormattedString()
- getFormatToPartsStructure()
- getSelector()
- getResolvedValue()

Internally, it can lazy-evaluate. For example, it only needs to generate the formatted string if the method getFormattedString() is called. If the function-result is only used for selection, that never needs to be generated. If the function-result is passed to an :uppercase function, then the getResolvedValue() would never need to be called, etc.

In particular a function can look at its operand, and if it is a function-result, see what options it was passed and use them if the type of the function is known.


The description of a function needs to specify what its inputs are, and how its function-result behaves. Part of that involves specifying which functions it can compose with, and how it treats that function's operand and options.

spec/formatting.md
eemeli: The intent with this paragraph is to communicate about end user capabilities, not about internal implementation details. So this is about being able to use the same value (say, a number passed into the implementation's formatting method as an argument) for multiple purposes.

In the syntax, this is effected by _variable references_, _declarations_, and other syntax features, and in the spec's runtime definition, by _resolved values_.

This proposed change flips the meaning of the text by discussing internal rather than external use cases. It also introduced the term "interface", which further increases the distance from concrete use cases, as well as rooting what follows in an object-oriented approach. This isn't invalid by any means, but it's presuming the reader to have a lot more context than the current text.

spec/formatting.md
eemeli: As this is a non-normative example and we've not reached consensus on formatted parts, I'd rather not introduce them here implicitly and somewhat out of context, when the key aspect of the `formatToX()` inclusion is to remind the reader that a string is not the only possible formatting target.

eemeli:
I went through the spec and linkified all "resolved value" mentions, and added examples for how literal and fallback values could be implemented using the previously given example. If this approach is acceptable, similar examples could be given for default functions, but that's outside the scope of this PR.

spec/formatting.md
```diff
@@ -168,28 +204,47 @@ Its resolved value is defined by _literal resolution_.
...
 
 To resolve the value of a _variable_,
 its _name_ is used to identify either a local variable or an input variable.
-If a _declaration_ exists for the _variable_, its resolved value is used.
+If a _declaration_ exists for the _variable_, its _resolved value_ is used.
 Otherwise, the _variable_ is an implicit reference to an input value,
 and its value is looked up from the _formatting context_ _input mapping_.
 
+An implemementation MAY choose to wrap or otherwise represent all input values
+using the same representation it uses for all _resolved values_.
```
eemeli, line 246: Highlighting this addition, which might otherwise be lost among the _resolved value_ link additions.

eemeli:
@lucacasonato, would be interested to hear if including the Resolved Values section and the examples would make it more readable from your PoV? https://github.com/unicode-org/message-format-wg/blob/define-resolved-values/spec/formatting.md

spec/formatting.md
```diff
@@ -85,54 +85,90 @@ At a minimum, it includes:
...
 >
-> Alternatively, it could be an instance of an ICU4J `FormattedNumber`,
-> or some other locally appropriate value.
+> With this approach:
+> - An _expression_ could be used as a _placeholder_ if
+>   calling the `formatToString()` or `formatToX()` method of its _resolved value_
+>   did not emit an error.
+> - An _expression_ could be used as a _selector_ _expression_ if
+>   calling the `selectKeys(keys)` method of its _resolved value_
+>   did not throw an error.
```
catamorphism, line 135:
```suggestion
>   did not emit an error.
```

(consistency)

spec/formatting.md
```diff
@@ -85,54 +85,90 @@ At a minimum, it includes:
...
+> - Using a _variable reference_, the _resolved value_ of an _expression_
+>   could be used as an _operand_ or _option_ value if
+>   calling the `getValue()` method of its _resolved value_ did not throw an error.
```
catamorphism, line 138:
```suggestion
>   calling the `getValue()` method of its _resolved value_ did not emit an error.
```

catamorphism (approved):
LGTM -- this will move things forward!

lucacasonato (commented):
@eemeli This does seem clearer than previously.

eemeli:
@aphillips As discussed yesterday, I've updated this PR to incorporate the consensus we reached a few weeks ago:

> A function MUST define its resolved value. The resolved value MAY be different from the value of the operand of the function. It MAY be an implementation specific type. It is not required to be the same type as the operand.
>
> A function MUST define its resolved options. The resolved options MAY be different from the options of the function.

After some thought, I found the best place for this to be a new section defining _function handlers_ as the real-world functions or methods that are called to get a _resolved value_. This now incorporates quite a bit of the language that was in step 4 of the Function Resolution process.

I opted for "handler" as we tend to use "implementation" to mean the whole of an MF2 formatter.

The exact MUSTard that's here is a little bit different from the above, as the requirements that a function "MUST define its resolved value" and "MUST define its resolved options" are not easily enforceable unless we were to define what "define" here means, and that's really rather tricky.

spec/formatting.md
```diff
@@ -77,50 +77,96 @@ At a minimum, it includes:
...
-_Expressions_ are used in _declarations_ and _patterns_.
-_Markup_ is only used in _patterns_.
+This specification allows for the same value to be used for:
+- formatting in a _placeholder_,
+- selection with a _selector_,
+- as the _operand_ of another _expression_ (including _local declarations_), or
+- as an _option_ value in another _expression_,
 
-In a _declaration_, the resolved value of the _expression_ is bound to a _variable_,
+To support this, the **_<dfn>resolved value</dfn>_** of each _expression_
```
aphillips, line 94: If we're going to `dfn` _resolved value_ here, we should write a self contained, concise definition, since every occurrence of the term will come here. On a mobile so not suggesting text just now.

spec/formatting.md
```diff
@@ -77,50 +77,96 @@ At a minimum, it includes:
...
+is an implementation-dependent value that supports some or all of the above use cases.
+
+In a _declaration_, the _resolved value_ of the _expression_ is bound to a _variable_,
 which is available for use by later _expressions_.
 Since a _variable_ can be referenced in different ways later,
 implementations SHOULD NOT immediately fully format the value for output.
```
aphillips, line 100: Maybe the key distinction here is that the resolved value and the formatted value are different things?

spec/formatting.md
```diff
@@ -156,25 +202,41 @@ Its resolved value is defined by _literal resolution_.
...
 or on the right-hand side of an _option_,
-the formatting function MUST treat its resolved value the same
+the formatting function MUST treat its _resolved value_ the same
 whether its value was originally a _quoted literal_ or an _unquoted literal_.
 
 > For example,
 > the _option_ `foo=42` and the _option_ `foo=|42|` are treated as identical.
 
-The resolution of a _text_ or _literal_ MUST resolve to a string.
+The resolution of a _text_ or _literal_ MUST resolve with a string value.
```
aphillips, line 217: Not sure about this change? The text or literal is a string, but the resolved value might depend on the function. E.g. `{1234 :number}` had a literal which MUST be a string, but the resolved value is a number? Too pedantic?

spec/formatting.md
```diff
@@ -202,108 +264,63 @@ the following steps are taken:
...
-
-> [!NOTE]
-> During the Technical Preview,
-> feedback on how the registry describes
-> the flow of _resolved values_ and _options_
-> from one _function_ to another,
-> and on what requirements this specification should impose,
-> is highly desired.
-
    An implementation MAY pass additional arguments to the function,
```
aphillips, line 275: To the function handler?

spec/formatting.md
```diff
@@ -202,108 +264,63 @@ the following steps are taken:
...
 
    If the call fails or does not return a valid value,
    emit the appropriate _Message Function Error_ for the failure.
 
-   Implementations MAY provide a mechanism for the _function_ to provide
+   Implementations MAY provide a mechanism for the _function handler_ to provide
    additional detail about internal failures.
    Specifically, if the cause of the failure was that the datatype, value, or format of the
    _operand_ did not match that expected by the _function_,
    the _function_ might cause a _Bad Operand_ error to be emitted.
```
aphillips, line 289: might => SHOULD ?

spec/formatting.md
```diff
@@ -202,108 +264,63 @@ the following steps are taken:
...
   
-   In all failure cases, use the _fallback value_ for the _expression_ as the resolved value.
+   In all failure cases, use the _fallback value_ for the _expression_ as its _resolved value_.
+
+#### Function Handler
+
+To resolve a _function_,
+a **_<dfn>function handler</dfn>_** is required.
+This is an implementation-defined process such as a function or method
+which accepts a set of arguments and returns a _resolved value_.
```
aphillips, line 298: Same comment as before about `dfn`. The definition is in the second sentence. Lead with that and then require it.

spec/formatting.md
```diff
@@ -202,108 +264,63 @@ the following steps are taken:
...
+
+The _resolved value_ returned by a _function handler_
+MAY be different from the value of the _operand_ of the _function_.
+It MAY be an implementation specified type.
+It is not required to be the same type as the _operand_.
+
+A _function handler_ MAY include resolved options in its _resolved value_.
+The resolved options MAY be different from the _options_ of the function.
+
+A _function handler_ SHOULD emit a
```
aphillips, line 316: Here's that SHOULD!

spec/formatting.md
```diff
@@ -400,6 +417,23 @@ _Option_ _identifiers_ and values are not included in the _fallback value_.
 
 _Pattern selection_ is not supported for _fallback values_.
```
aphillips, line 418: Should this be optional?  `*` might match?

spec/registry.md
```diff
@@ -374,12 +374,12 @@ All other values produce a _Bad Operand_ error.
 ### Digit Size Options
 
 Some _options_ of number _functions_ are defined to take a "digit size option".
-Implementations of number _functions_ use these _options_ to control aspects of numeric display
+_Function handlers_ of number _functions_ use these _options_ to control aspects of numeric display
```
aphillips, line 377: Perhaps "The _function handlers_ for..."?

spec/syntax.md
```diff
@@ -97,7 +97,7 @@ Attempting to parse a _message_ that is not _valid_ will result in a _Data Model
 
 A **_<dfn>message</dfn>_** is the complete template for a specific message formatting request.
 
-A **_<dfn>variable</dfn>_** is a _name_ associated to a resolved value.
+A **_<dfn>variable</dfn>_** is a _name_ associated with a _resolved value_.
```
aphillips, line 100: I think the original was better? Associated to is specific.

aphillips (commented):
Looks good. Doing this on a cell phone, so no suggestions... sorry.

spec/formatting.md
macchiati: We actually don't care whether an implementation produces a formatted value that isn't used, so the "SHOULD NOT" is too strong.

spec/errors.md
```diff
@@ -269,7 +269,7 @@ a reference to a function which cannot be resolved.
 ### Bad Selector
 
 A **_<dfn>Bad Selector</dfn>_** error occurs when a message includes a _selector_
-with a resolved value which does not support selection.
+with a _resolved value_ which does not support selection.
```
macchiati, line 272: I'm confused. Can you give an example of a selector that supports selection where its resolved value doesn't?

That is, if I have
.input {$num :number}
.match $num
...
How could :number ever have a resolved value that didn't support selection? I think the notion of 'resolved value' doesn't help.

```suggestion
that does not support selection.
```

spec/formatting.md
aphillips: What I think we're trying to say is: implementations should not replace the variable with the formatted value. Rather, it should replace it with the resolved value. Whether or not it does some formatting eagerly or not is, as @macchiati notes, an internal detail.

There is possibly a relationship between the formatted and resolved value. Consider:

```
.local $a = {zebra :titlecase}
{{{$a}}}
```

... probably prints `Zebra` and probably the resolved value of `$a` is `Zebra`. However:

```
.local $a = {0xbeef :integer}
{{{$a}}}
```

... might print `48,879` and have a resolved value might be an `int` with a value of `0xBEEF` (although we don't support hex literals like that, so you'd actually get a Bad Operand). If the locale were `ar-EG`, though, the formatted value would be `٤٨٬٨٧٩` but the resolved value would still be `(int)0xBEEF`.

So +1 to not SHOULD NOT, but my point is different.

spec/errors.md
```diff
@@ -283,10 +283,10 @@ with a resolved value which does not support selection.
 ## Message Function Errors
 
 A **_<dfn>Message Function Error</dfn>_** is any error that occurs
-when calling a message function implementation
+when calling a _function handler_
 or which depends on validation associated with a specific function.
 
-Implementations SHOULD provide a way for _functions_ to emit 
+Implementations SHOULD provide a way for _function handlers_ to emit
```
macchiati, line 289: I couldn't find a definition of what a "function handler" is. Which file is that in?

spec/formatting.md
```diff
@@ -16,7 +16,7 @@ The formatting of a _message_ is defined by the following operations:
   For a message with no _selectors_, this is simple as there is only one _pattern_.
   With _selectors_, this will depend on their resolution.
 
-- **_<dfn>Formatting</dfn>_** takes the resolved values of the selected _pattern_,
+- **_<dfn>Formatting</dfn>_** takes the _resolved values_ of the selected _pattern_,
```
macchiati, line 19: What is the "resolved values" of a pattern? Do you mean a the resolved values **of the placeholders** in a selected pattern?

That is false, however. It takes not the "resolved values", but rather the *formatted values* of the placeholders.

spec/syntax.md
macchiati: I find this all somewhat muddled. Except in function composition, "resolved value" is not a useful abstraction. 

That is, outside of function composition, the relevant features are 
- a formatted value, and
- a selection value

macchiati (changes requested):

spec/errors.md
eemeli: > How could :number ever have a resolved value that didn't support selection?

Going with the example you posted above, if that message is formatted without a `num` argument, then the first step of [Function Resolution](https://github.com/unicode-org/message-format-wg/blob/main/spec/formatting.md#function-resolution) applies:

> If the _expression_ includes an _operand_, resolve its value. If this fails, use a _fallback value_ for the _expression_.

And then this part of [Fallback Resolution](https://github.com/unicode-org/message-format-wg/blob/main/spec/formatting.md#fallback-resolution) applies:

> _Pattern selection_ is not supported for _fallback values_.

And so we end up emitting two errors when formatting:
1. Unresolved Variable
2. Bad Selector

spec/errors.md
eemeli: It's added by this PR to `spec/formatting.md`.

spec/formatting.md
eemeli: Note that in `{1234 :number}` we have two separate resolved values that end up being considered:
1. The resolved value of the _operand_ `1234`.
2. The resolved value of the _expression_ `{1234 :number}`.

This phrase is strictly about the former, as we're requiring that the parsing of the string `'1234'` into a numerical value can only be done by the `:number` function handler.

spec/formatting.md
eemeli:
```suggestion
   An implementation MAY pass additional arguments to the _function handler_,
```

spec/formatting.md
eemeli:
```suggestion
   the _function_ SHOULD cause a _Bad Operand_ error to be emitted.
```

spec/formatting.md
eemeli: If a selector does not support pattern selection, then the `*` variant is selected. A selector is never asked about `*`, that's handled by the selection algorithm directly.

The fact that you need to consider the combined effects of Resolve Preferences 2.ii.b, Filter Variants 2.i.b, and Sort Variants 5.iii.c to see that is what I tried to mean by arguing that "best choice" is too complicated.

spec/registry.md
eemeli:
```suggestion
The _function handlers_ for number _functions_ use these _options_ to control aspects of numeric display
```

spec/syntax.md
eemeli:
```suggestion
A **_<dfn>variable</dfn>_** is a _name_ associated to a _resolved value_.
```

eemeli:
Replying to @macchiati in https://github.com/unicode-org/message-format-wg/pull/728#discussion_r1783673632:
> I find this all somewhat muddled. Except in function composition, "resolved value" is not a useful abstraction.
> 
> That is, outside of function composition, the relevant features are
> * a formatted value, and
> * a selection value

A _resolved value_ is also useful when considering the processing of a message like this:
```
.input {$num :number}
.match $num
1 {{You have a thing}}
* {{You have {$num} things}}
```
Because we're using `$num` for both selection and formatting, it's useful to be able to say that it has a _resolved value_ that supports both selection and formatting.

spec/formatting.md
eemeli: This text is no longer present in the PR.

spec/formatting.md
eemeli: @aphillips I'm closing this thread as it's outdated, but please open a new one if you think that something here is still actionable.

spec/formatting.md
```diff
@@ -85,10 +85,16 @@ At a minimum, it includes:
...
-## Expression and Markup Resolution
+## Resolved Values
 
-_Expressions_ are used in _declarations_, _selectors_, and _patterns_.
-_Markup_ is only used in _patterns_.
+This specification allows for the same value to be used for:
+- formatting in a _placeholder_,
+- selection in a _selector_ _expression_,
+- as the _operand_ of another _expression_, or
+- as the _option_ value of another _expression_,
```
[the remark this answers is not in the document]
eemeli: I've dropped this list from the PR.

spec/formatting.md
eemeli: Closing as out-of-date.

spec/formatting.md
eemeli: Dropped this phrase, as it's somewhat redundant.

eemeli:
I've rephrased the _resolved value_ definition a bit, and clarified that it applies to all message parts.

spec/formatting.md
aphillips: Maybe clearer as:

```suggestion
The _resolved value_ of a _text_ or _literal_ is a string.
```

spec/formatting.md
aphillips: I went to the context and read the section carefully. I guess I have a few questions.

This text suggests that the fallback state is transitive--that implementations must track if a fallback has occurred. That probably needs to be stated more clearly, if so. Otherwise, the fallback value looks just like a string i.e. a literal, which is what lead to my comment.

Let's use my bad example:

```
.local $a = {0xbeef :num}       // this is Bad Operand with a fallback of `|0xbeef|`
.local $b = {$a :uppercase}     // this might work on `0xbeef`? It's probably Bad Operand if the first line isn't
.match $b                       // let's say :uppercase is a type of string selector too...
48879 {{Definitely not this}}
0xBEEF {{Matches?}}
* {{Could match if above doesn't}}
```

My point is, how does `.match $b` know that its resolved value is a fallback value?

spec/errors.md
aphillips: I think @macchiati's finger is on the right thing. It's not a technical problem. This is just hard to understand. Perhaps:

```suggestion
A **_<dfn>Bad Selector</dfn>_** error is an error that occurs when a _message_
includes a _selector_ whose _resolved value_ 
contains a _function handler_ which does not support selection,
is missing a _function handler_,
or consists of a _fallback value_.
```

aphillips:
(chair hat 🎩 on)

I just re-requested reviews from everyone who'd left reviews on the old PR so that the status is clear.

spec/formatting.md
eemeli: Actually, we should drop this as it's duplicating the first paragraph of this section. Merge this suggestion if you agree?
```suggestion
```

spec/formatting.md
eemeli: That's a good point. Some of the current text is using "fails to resolve" in a way that's not clear. I'll apply some updates to improve this.

catamorphism (approved):

spec/errors.md
eemeli: Could we spin this suggestion off into a separate PR? As is, the only change that is being applied here is linkifying _resolved value_.

aphillips: Works for me as a solution.

macchiati:
>A resolved value is also useful when considering the processing of a message like this:

> .input {$num :number}
> .match $num
> 1 {{You have a thing}}
> * {{You have {$num} things}}

> Because we're using $num for both selection and formatting, it's useful to be able to say that it has a resolved value that supports both selection and formatting.

I really find this extremely confusing. 

It appears to me that the core concept is the following:

- Each variable (such as $num) references an entity that that performs selection and formatting.
- And that you are calling that entity a "resolved value". 

Is that what you mean, and I'm just confused by the terminology that you are using?

spec/errors.md
eemeli: Here's a link to the section in this PR's branch: https://github.com/unicode-org/message-format-wg/blob/define-resolved-values/spec/formatting.md#function-handler

eemeli:
> It appears to me that the core concept is the following:
> * Each variable (such as $num) references an entity that that performs selection and formatting.
> * And that you are calling that entity a "resolved value".
> 
> Is that what you mean, and I'm just confused by the terminology that you are using?

Yes, that's pretty much it. With the very slight caveat that we do not require that the resolved value entity itself performs selection and formatting, as it should be possible for it to be a pure-data thing with no methods but enough information such that a separate, external processor can do the appropriate selection and formatting.

aphillips:
> Yes, that's pretty much it. With the very slight caveat that we do not require that the resolved value entity itself performs selection and formatting, as it should be possible for it to be a pure-data thing with no methods but enough information such that a separate, external processor can do the appropriate selection and formatting.

To use a concrete example from ICU4J, think of the _resolved value_ of a `:number` as being a `FormattedNumber` object. It contains the number itself, as well as all of the resolved options for formatting the number. You can call `toString()` on it to get the formatted representation. If that class implemented MF2 selection, it might add a `matchSelectorKeys` method that you could call for selection.

macchiati:
BTW, since at least part of the motivation for 'resolved value' is in Pattern Selection, I took a look at that section. I think the spec could be much cleaner (with the same results) with a revision: I proposed an outline of that in https://github.com/unicode-org/message-format-wg/issues/898

PR #370, opened 2023-03-27
Use _underline_ for italics in syntax.md

eemeli:
To match common use in markdown.

Also correct one remaining use of `(quotes)` instead of `|quotes|`.

PR #731, opened 2024-03-15
Disallow whitespace as the first character of a reserved-body in a reserved-statement.

bhaible:
In the 'reserved-statement' nonterminal, there is an ambiguity if there is more than one whitespace character between the 'reserved-keyword' and the first non-whitespace character of the 'reserved-body', because these whitespace characters can be seen as part of the 's' nonterminal or as part of the 'reserved-body' nonterminal.

According to the principles explained in #725 and the proposed resolution of #721, it is not desired that a 'reserved-body' starts with a whitespace character; rather, such a whitespace character is meant to be interpreted as part of the preceding 's' nonterminal.

Test case:
```
.regex   /foo/{xyz}{{hello}}
```

This patch removes this ambiguity, by disallowing whitespace as the first character of a 'reserved-body' in a reserved-statement.

It thus fixes the first part of #721.

Details:
  - In the other occurrences of 'resolved-body' as well (in a 'reserved-annotation' or 'private-use-annotation') the leading whitespace is separated as well. This has no influence on the set of inputs that the 'reserved-annotation' and 'private-use-annotation' nonterminals can match, but highlights that the parser should better trim off this leading whitespace in these places before entering the 'resolved-body' into the data model.
  - Two nonterminals 'reserved-body-start' and 'resolved-body-part' are introduced, each referenced once. The purpose is clarity and to follow the common *-start / *-part idiom.

### DIFF
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -61,13 +61,14 @@ reserved-statement = reserved-keyword [s reserved-body] 1*([s] expression)
 reserved-keyword   = "." name
 
 ; Reserve additional sigils for use by future versions of this specification.
-reserved-annotation       = reserved-annotation-start reserved-body
+reserved-annotation       = reserved-annotation-start [[s] reserved-body]
 reserved-annotation-start = "!" / "%" / "*" / "+" / "<" / ">" / "?" / "~"
 
 ; Reserve sigils for private-use by implementations.
-private-use-annotation    = private-start reserved-body
+private-use-annotation    = private-start [[s] reserved-body]
 private-start             = "^" / "&"
-reserved-body             = *([s] 1*(reserved-char / reserved-escape / quoted))
+reserved-body             = reserved-body-part *([s] reserved-body-part)
+reserved-body-part        = reserved-char / reserved-escape / quoted
 
 ; Names and identifiers
 ; identifier matches https://www.w3.org/TR/REC-xml-names/#NT-QName
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -610,7 +610,7 @@ wish to use a syntax exactly like other functions. Specifically:
 A _private-use annotation_ MAY be empty after its introducing sigil.
 
 ```abnf
-private-use-annotation = private-start reserved-body
+private-use-annotation = private-start [[s] reserved-body]
 private-start          = "^" / "&"
 ```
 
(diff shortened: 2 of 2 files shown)


spec/message.abnf
```diff
@@ -61,13 +61,15 @@ reserved-statement = reserved-keyword [s reserved-body] 1*([s] expression)
...
 reserved-annotation-start = "!" / "%" / "*" / "+" / "<" / ">" / "?" / "~"
 
 ; Reserve sigils for private-use by implementations.
-private-use-annotation    = private-start reserved-body
+private-use-annotation    = private-start [[s] reserved-body]
 private-start             = "^" / "&"
-reserved-body             = *([s] 1*(reserved-char / reserved-escape / quoted))
+reserved-body             = reserved-body-start *reserved-body-part
+reserved-body-start       = reserved-char / reserved-escape / quoted
+reserved-body-part        = [s] 1*(reserved-char / reserved-escape / quoted)
```
eemeli, line 72: This can be simplified:

```suggestion
reserved-body             = reserved-body-part *([s] reserved-body-part)
reserved-body-part        = reserved-char / reserved-escape / quoted
```

eemeli (approved):
If we're going to encode in the syntax that the value of `reserved-body` does not include its leading whitespace, then this seems like a better way of doing so than #370.

Approving, but note my inline suggested simplification, which ought to be applied `syntax.md` as well.

This does introduce one concern that should be noted:

If someone introduces private syntax `{&foo}` for which they do not allow for space between the `&` and `foo`, a standard MF2 processor, when encountering a privately invalid expression `{& foo}`, could reserialize it as `{&foo}`, a privately valid expression.

Similarly, this change would effectively disallow private and future syntax from _requiring_ a space immediately after the initial sigil, as an MF2.0 processor would likely dismiss the space as insignificant.

I'm okay with accepting these limitations.

spec/message.abnf
bhaible: Thanks for the suggestion. I like it, and I'll apply it if there is consensus. (I had introduced the `reserved-body-start` nonterminal following @aphillips comment https://github.com/unicode-org/message-format-wg/pull/730#discussion_r1526378593 .)

spec/syntax.md
```diff
@@ -653,10 +653,12 @@ While a reserved sequence is technically "well-formed",
 unrecognized _reserved-annotations_ or _private-use-annotations_ have no meaning.
 
 ```abnf
-reserved-annotation       = reserved-annotation-start reserved-body
+reserved-annotation       = reserved-annotation-start [[s] reserved-body]
```
catamorphism, line 656: I think it's worth adding some prose here explaining why the grammar is structured this way. It wasn't obvious immediately that it had to be, so it won't be obvious to the reader either. Including the example would be good. 

I also think a note should be added to the "Reserved Statements" section of the same doc that explains what the limitations are of `reserved-statement` that arise because of this change. You could use your example from #721.

catamorphism (changes requested):
The grammar change looks good. It took me ages to understand what this is doing; maybe I'm just slow, but I think it might help future readers to add a note in syntax.md as I suggested.

spec/syntax.md
aphillips: Let's use a separate PR to do any text changes.

aphillips (approved):

spec/message.abnf
aphillips: Let's apply this here and below.

bhaible:
The simplification suggested by @eemeli is now included.

PR #737, opened 2024-03-19
Create ack section

aphillips:
Create the acknowledgements section. Addresses #711. See that issue for how this was created (although it is self-describing).

Additional names are welcome! Specific acknowledgements for specific deeds are also welcome (as mentioned in the 2024-03-18 call). No omission is intentional. Don't feel shy about putting your name on our work.

### DIFF
--- a/spec/appendices.md
+++ b/spec/appendices.md
@@ -72,7 +72,7 @@ Markus Scherer,
 Zibi Braniecki, 
 Matt Radbourne, 
 Bruno Haible,
-Rafael Xavier de Souza
+and Rafael Xavier de Souza.
 
 Addison Phillips was chair of the working group from January 2023.
 Prior to 2023, the group was governed by a chair group, consisting of
--- a/spec/appendices.md
+++ b/spec/appendices.md
@@ -44,3 +44,45 @@ including buffer overflow, code injection, user tracking,
 fingerprinting, and other types of bad behavior.
 Any installed code needs to be appropriately sandboxed.
 In addition, end-users need to be aware of the risks involved.
+
+## Acknowledgements
+
+Special thanks to the following people for their contributions to making MessageFormat v2.
+The following people contributed to our github repo and are listed in order by contribution size:
+
+Addison Phillips, 
+Eemeli Aro, 
+Romulo Cintra, 
+Stanisław Małolepszy, 
+Elango Cheran, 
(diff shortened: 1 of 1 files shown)


spec/appendices.md
```diff
@@ -44,3 +44,45 @@ including buffer overflow, code injection, user tracking,
...
+Daniel Minor, 
+Christopher Dieringer, 
+George Rhoten, 
+Ujjwal Sharma, 
+Daniel Ehrenberg, 
+Markus Scherer, 
+Zibi Braniecki, 
+Matt Radbourne, 
+Bruno Haible,
+Rafael Xavier de Souza
```
catamorphism, line 75:
```suggestion
and Rafael Xavier de Souza.
```
(For consistency with the other list of people)

catamorphism (approved):

eemeli (approved):

spec/appendices.md
```diff
@@ -44,3 +44,45 @@ including buffer overflow, code injection, user tracking,
 fingerprinting, and other types of bad behavior.
 Any installed code needs to be appropriately sandboxed.
 In addition, end-users need to be aware of the risks involved.
+
+## Acknowledgements
+
+Special thanks to the following people for their contributions to making MessageFormat v2.
+The following people contributed to our github repo and are listed in order by contribution size:
```
stasm, line 51: I think this may have a bit of a skew caused by the fact that git commits are not the only way people have participated. Is there maybe a way to also count the number of issue/pr comments?

stasm (approved):

spec/appendices.md
aphillips: That's fair. Note that this also counts issues. 

I could alphabetize by family name. The idea of organizing by what github gives us it that is provides some weighting for "participation" without trying to reconstruct specific contributions. I could also look back at meeting minutes to see if we're missing someone.

stasm: You make a good point: once we decide to use something other than what GitHub gives us by default, we're in the business of making it as close to the true representation of participation as possible, and that's very hard to do right.

I think it's OK to merge this as-is.

macchiati:
Just a note: we typically let people specify a native script version of their names in parentheses. If anyone wants that, I suggest a follow-on PR.

macchiati (approved):

PR #740, opened 2024-03-19
Remove tests that rely on nonexistent bad-option error

catamorphism:
I suggest we remove these two tests for now, because the property that they are testing is actually underspecified. There is nothing in errors.md corresponding to the `bad-option` error, and what's more, the spec doesn't require `:number` to signal an error for a non-integer `minimumFractionDigits` value.

I filed #738 and #739 to clarify those issues in the spec, but I think the tests should be removed in the meantime, and re-added once the behavior is fully specified.

### DIFF
--- a/test/test-functions.json
+++ b/test/test-functions.json
@@ -128,12 +128,6 @@
       "exp": "bar 4.20"
     },
     {
-      "src": ".local $foo = {$bar :number minimumFractionDigits=foo} {{bar {$foo}}}",
-      "params": { "bar": 4.2 },
-      "exp": "bar {$bar}",
-      "errors": [{ "type": "bad-option" }]
-    },
-    {
       "src": ".local $foo = {$bar :number} {{bar {$foo}}}",
       "params": { "bar": "foo" },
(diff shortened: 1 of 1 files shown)


aphillips:
I think this would be a mistake. I agree that there is no specific `bad-option` error. I think we should fix registry.md for some of the cases you point out. I think that number formatting is so common that these options should be well-clarified now--and that there will be general agreement about how to do that. I think these tests should be retained.

aphillips:
Note: I made #741 to (try to) address part of the gap. It doesn't mess with the `bad-option` problem.

eemeli (changes requested):
I don't think these should be dropped. The right fix here would be to better define the behaviour for `:number minimumFractionDigits=foo`, which should indeed be considered an error.

aphillips:
I'm marking this for Monday's agenda. I used `resolve-candidate` because at least two of us feel that this shouldn't go in, but that doesn't mean the WG won't feel differently.

aphillips:
In the 2024-03-25 call we agreed not to merge this.

PR #750, opened 2024-03-25
Clarify that .match selectors are not declarations

### LINKED ISSUE #748: Make `.match` explictly a non-declaration
In the 2024-03-25 call we agreed to document what our implementations currently do, which is that `.match` selectors are not declarations. This will be a `fast-track` change to 45.

eemeli:
Closes #748

Adds a clarifying note about `.match` not acting as a declaration and updates misleading examples.

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -490,8 +490,11 @@ When all of the _selectors_ have been processed,
 the earliest-sorted _variant_ in the remaining list of _variants_ is selected.
 
 > [!NOTE]
-> A _selector_ _expression_ is not a _declaration_,
-> and its _annotation_ has no effect on the value of its _operand_ during formatting.
+> A _selector_ is not a _declaration_.
+> Even when the same _function_ can be used for both formatting and selection
+> of a given _operand_
+> the _annotation_ that appears in a _selector_ has no effect on subsequent
+> _selectors_ nor on the formatting used in _placeholders_.
 > To use the same value for selection and formatting,
 > set its value with a `.input` or `.local` _declaration_.
 
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -489,6 +489,12 @@ Earlier _selectors_ in the _matcher_'s list of _selectors_ have a higher priorit
 When all of the _selectors_ have been processed,
 the earliest-sorted _variant_ in the remaining list of _variants_ is selected.
 
+> [!NOTE]
+> A _selector_ _expression_ is not a _declaration_,
+> and its _annotation_ has no effect on the value of its _operand_ during formatting.
+> To use the same value for selection and formatting,
+> set its value with a `.input` or `.local` _declaration_.
+
 This selection method is defined in more detail below.
 An implementation MAY use any pattern selection method,
 as long as its observable behavior matches the results of the method defined here.
@@ -692,10 +698,11 @@ and an `en` (English) locale,
 the pattern selection proceeds as follows for this message:
 
 ```
-.match {$count :number}
-one {{Category match}}
-1   {{Exact match}}
-*   {{Other match}}
+.input {$count :number}
+.match {$count}
+one {{Category match for {$count}}}
+1   {{Exact match for {$count}}}
+*   {{Other match for {$count}}}
 ```
 
 1. For the selector:<br>
@@ -716,7 +723,7 @@ one {{Category match}}
    This is then sorted as:<br>
    « ( 0, `1` ), ( 1, `one` ), ( 2, `*` ) »<br>
 
-4. The pattern `Exact match` of the most preferred `1` variant is selected.
+4. The pattern `Exact match for {$count}` of the most preferred `1` variant is selected.
 
 ## Formatting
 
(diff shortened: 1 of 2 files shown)


spec/formatting.md
```diff
@@ -489,6 +489,12 @@ Earlier _selectors_ in the _matcher_'s list of _selectors_ have a higher priorit
 When all of the _selectors_ have been processed,
 the earliest-sorted _variant_ in the remaining list of _variants_ is selected.
 
+> [!NOTE]
+> A _selector_ _expression_ is not a _declaration_,
+> and its _annotation_ has no effect on the value of its _operand_ during formatting.
```
aphillips, line 494: I think the term "expression" isn't needed here. We use just "selector" all through this section otherwise.

I think the term "value" has also been a source of woe for us of late and I'd rather avoid it if we can. Perhaps:

```suggestion
> A _selector_ is not a _declaration_.
> Even when the same _function_ can be used for both formatting and selection
> of a given _operand_
> the _annotation_ that appears in a _selector_ has no effect on subsequent
> _selectors_ nor on the formatting used in _placeholders_.
```

spec/formatting.md
```diff
@@ -692,10 +698,11 @@ and an `en` (English) locale,
...
 ```
-.match {$count :number}
-one {{Category match}}
-1   {{Exact match}}
-*   {{Other match}}
+.input {$count :number}
+.match {$count}
+one {{Category match for {$count}}}
+1   {{Exact match for {$count}}}
+*   {{Other match for {$count}}}
```
aphillips, line 705: Is the goal of this change to illustrate the use of a declaration with a selector? Otherwise the change here isn't necessary.

spec/syntax.md
```diff
@@ -408,7 +409,9 @@ There MAY be any number of additional _selectors_.
 > A message with two _selectors_:
 >
 > ```
-> .match {$numLikes :number} {$numShares :number}
+> .input {$numLikes :number}
+> .input {$numShares :number}
```
aphillips, line 413: may as well use `:integer` here?

aphillips (changes requested):

spec/formatting.md
eemeli: I thought that having at least one of the selection examples use a declaration would be good.

spec/syntax.md
eemeli:
```suggestion
> .input {$numLikes :integer}
> .input {$numShares :integer}
```

aphillips (approved):

spec/formatting.md
```diff
@@ -489,6 +489,15 @@ Earlier _selectors_ in the _matcher_'s list of _selectors_ have a higher priorit
 When all of the _selectors_ have been processed,
 the earliest-sorted _variant_ in the remaining list of _variants_ is selected.
 
+> [!NOTE]
+> A _selector_ is not a _declaration_.
+> Even when the same _function_ can be used for both formatting and selection
+> of a given _operand_
+> the _annotation_ that appears in a _selector_ has no effect on subsequent
+> _selectors_ nor on the formatting used in _placeholders_.
```
catamorphism, line 497:
```suggestion
> [!NOTE]
> _Expressions_, including _expressions_ that appear as _selectors_,
> have no side effects and are not _declarations_.
> That is, they do not introduce or override _variable_ names.
> This means that the _annotation_ of a _variable_ in a _selector_
> does not change what is bound to the _variable_,
> even if the same _function_ can be used for both formatting and selection
> of a given _operand_.
> The meaning of a _variable_ reference that occurs within a _variant_
> depends only on the explicit or implicit _declaration_ of that _variable_.
```

Just a suggestion. I think this text is difficult to understand outside the context of the discussions we've been having, and without an example. The same might be true of my suggested text. I guess if it's confusing, people will ask about it during the tech preview feedback period.

catamorphism (approved):
Feel free to use all or any part of my suggested text, or ignore it -- up to your discretion.

spec/formatting.md
aphillips: The right hand size of a `.local` is an expression, so your suggestion might be confusing from that respect?

catamorphism: Doesn't seem like it to me. An expression is not a declaration; it can appear in a declaration, but not every term that an expression can appear in is a declaration. ("Term" = "subtree of the data model".)

Alternately, the text could begin with something like "Only a `.local` or an `.input` _declaration_ can change the meaning of a _variable_." Though this isn't quite true, as free variables introduce what we call "implicit declarations" someplace. A `.match` can't contain declarations (not even implicit declarations, since the only place where an "implicit declaration" makes sense is in a context where further names can be introduced) so it's clear from that that nothing within a `.match` changes the meaning of any variable name. 

Every phrasing I can think of is made more complicated by implicit declarations, though.

aphillips: @catamorphism Would you object to my merging the text as is? We're definitely coming back to the whole "resolved value" space later 😄

catamorphism: @aphillips I don't object!

PR #751, opened 2024-03-25
Clarify that options and attributes are unordered

### LINKED ISSUE #749: Explicitly say that the order of options has no meaning
In the 2024-03-25 call we agreed to say that the order of `option`s has no meaning somewhere near here:

https://github.com/unicode-org/message-format-wg/blob/main/spec/syntax.md#options

eemeli:
Closes #749.
See #716 for context.

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -319,7 +319,7 @@ the following steps are taken:
 
 #### Option Resolution
 
-The result of resolving _option_ values is a mapping of string identifiers to values.
+The result of resolving _option_ values is an unordered mapping of string identifiers to values.
 
 For each _option_:
 
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -766,8 +766,6 @@ The _value_ of an _attribute_ can be either a _literal_ or a _variable_.
 
 Multiple _attributes_ are permitted in an _expression_ or _markup_.
 Each _attribute_ is separated by whitespace.
-Each _attribute_'s _identifier_ MUST be unique within the _expression_ or _markup_:
-an _expression_ or _markup_ with duplicate _attribute_ _identifiers_ is not valid.
 
 The order of _attributes_ is not significant.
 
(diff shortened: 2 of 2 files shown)


catamorphism (approved):

spec/syntax.md
```diff
@@ -764,6 +766,11 @@ The _value_ of an _attribute_ can be either a _literal_ or a _variable_.
 
 Multiple _attributes_ are permitted in an _expression_ or _markup_.
 Each _attribute_ is separated by whitespace.
+Each _attribute_'s _identifier_ MUST be unique within the _expression_ or _markup_:
+an _expression_ or _markup_ with duplicate _attribute_ _identifiers_ is not valid.
+
+The order of _attributes_ is not significant.
```
aphillips, line 772: Let's not introduce this now. In this version, attributes are reserved. We didn't agree to make the order insignificant or require uniqueness, even if both become restrictions later.

spec/formatting.md
```diff
@@ -319,7 +319,7 @@ the following steps are taken:
 
 #### Option Resolution
 
-The result of resolving _option_ values is a mapping of string identifiers to values.
+The result of resolving _option_ values is an unordered mapping of string identifiers to values.
```
aphillips, line 322: This might be too strong. An implementation would not be incorrect to order the options internally--that would be an implementation detail. What's important is that functions cannot count on the ordering or make the ordering significant.

aphillips (commented):
A couple of comments.

spec/formatting.md
eemeli: This has that effect. It's still possible for an implementation to use a container that retains order to represent the options, but with this qualifier functions can't rely on that.

spec/syntax.md
eemeli: They're still reserved, but this makes attributes easier to deal with. The more different attributes and options are, the more custom code is required for dealing with this reserved syntax feature.

aphillips: The uniqueness requirement requires checking if the attribute is, in fact, unique. If one is using a map, perhaps that's easy. But if one is not, it's extra work that we just requirement.

I could maybe buy the ordering requirement, as this allows implementations to reorder attributes.

eemeli: I checked our three current implementations, and indeed none currently check attribute uniqueness, so this should be dropped for now.

```suggestion

The order of _attributes_ is not significant.
```

PR #753, opened 2024-03-26
Add design doc on function composition

catamorphism:
This design doc partially addresses #515. It covers some of the same ground as the design doc in #645, but should be seen as a prequel to #645.

Before getting into solutions, I thought it would be useful to collect together some examples and questions, omitting the  last two sections from the design doc template.

This document is also an attempt to summarize the ICU-TC mailing list discussion that included @macchiati, @markusicu, @echeran and @richgillam . (Some of the comments, which fall more under "proposed solutions", will be included in a future sequel to this design doc.) 

My goal is to get this document into a state that everyone can agree on before finishing part 2 (with the proposed solutions).

I invited some reviewers on an "in case you're interested" basis.

### DIFF
--- a/exploration/function-composition-part-1.md
+++ b/exploration/function-composition-part-1.md
@@ -879,10 +879,74 @@ resembles the idea of specifying a type system for functions.
 Specifying rules for function composition, while also remaining typeless,
 seems difficult and potentially unpredictable.
 
-## Past work
+## Prior work
 
-[PR #198](https://github.com/unicode-org/message-format-wg/pull/198) -- attempt to define resolved values in the spec, from 2021
-["A Modular and Extensible MessageFormat 2.0"](https://github.com/unicode-org/message-format-wg/discussions/190) -- discussion that includes this topic, also from 2021 ("runtime model")
+In a previous iteration of the spec,
+[PR #198](https://github.com/unicode-org/message-format-wg/pull/198) (September 2021)
+proposed to add a general `Formattable` interface to represent runtime values.
+This proposal was not merged, and eventually, the term "resolved value"
+was incorporated into the spec instead to abstract over this interface.
+
+Currently, custom functions are passed their argument
+(_operand_ in MessageFormat syntax)
+and a mapping of option names to option values separately.
+In PR 198, a `Formattable` would have encapsulated
+both the argument and options.
+A variation on this idea might make it simpler
+to return values with options:
+if a function implementation has type `Formattable -> Formattable`,
+then it can preserve some options in the returned value,
+omit them, or add new ones.
+
+A wiki page from August 2021,
+["Data and Execution Model Differences"](https://github.com/unicode-org/message-format-wg/wiki/Data-&-Execution-Model-Differences#formatting-function-dependencies),
+includes a section on "Formatting Function Dependencies",
+which includes the question:
+"When implementing a formatting function,
+what values/arguments does it need to have access to?"
+
+A GitHub discussion also from August 2021,
+["A Modular and Extensible MessageFormat 2.0"](https://github.com/unicode-org/message-format-wg/discussions/190),
+includes discussion of the "runtime model" for values;
+a proposed design based on this document
+might also specify a runtime model for values.
+At the time, "pattern element formatters"
+and "formatting functions" were conceived
+as separate categories of functions.
+This document also suggests
+introducing more categories of functions,
+but split in a different way.
+The August 2021 discussion suggests giving
+"pattern element formatters" more access to message context.
+Such a distinction might be useful again:
+it might be desirable to allow function authors
+to declare either simple functions that
+don't compose with each other,
+and more complex functions with a richer interface
+that can compose with each other.
+
+There are some common elements between those past discussions
+and this document.
+But in contrast with previous proposals that give functions
+access to message context,
+this document suggests a purely functional approach
+in which values the function needs to have access to
+are encapsulated in a single argument type
+and values the function needs to return
+are encapsulated in a single return type.
+
+Another notable difference with some of the prior work
+is that in the current spec, the data model
+is completely separate from the model for values
+that functions operate on.
+That is, functions operate on values
+without needing to know or care
+whether those values were obtained from
+evaluating a MessageFormat _literal_ or _variable_.
+We do not propose changing that.
+Hence, revisiting the extensibility of the runtime model
+now that the data model is settled
+may result in a more workable solution.
 
 ## Proposed design and alternatives considered
 
--- a/exploration/function-composition-part-1.md
+++ b/exploration/function-composition-part-1.md
@@ -879,6 +879,11 @@ resembles the idea of specifying a type system for functions.
 Specifying rules for function composition, while also remaining typeless,
 seems difficult and potentially unpredictable.
 
+## Past work
+
+[PR #198](https://github.com/unicode-org/message-format-wg/pull/198) -- attempt to define resolved values in the spec, from 2021
+["A Modular and Extensible MessageFormat 2.0"](https://github.com/unicode-org/message-format-wg/discussions/190) -- discussion that includes this topic, also from 2021 ("runtime model")
+
 ## Proposed design and alternatives considered
 
 These sections are omitted from this document and will be added in
--- a/exploration/function-composition-part-1.md
+++ b/exploration/function-composition-part-1.md
@@ -959,10 +959,14 @@ would fall into one of the following categories:
 
 1. Provide a general mechanism for custom function authors
 to specify how functions compose with each other.
-2. Specify composition rules for built-in functions,
+1. Specify composition rules for built-in functions,
 but not in general, allowing custom functions
 to cooperate in an _ad hoc_ way.
-3. Restrict function composition for built-in functions
+1. Recommend a rich representation of resolved values
+without specifying any constraints on how these values
+are used.
+(This is the approach in [PR 645](https://github.com/unicode-org/message-format-wg/pull/645).)
+1. Restrict function composition for built-in functions
 (in order to prevent unintuitive behavior).
 
 ## Acknowledgments
--- a/exploration/function-composition-part-1.md
+++ b/exploration/function-composition-part-1.md
@@ -245,7 +245,8 @@ that were passed in to the function.
 
 While some use cases don't work well
 (or at least work surprisingly)
-if options are **not** encoded in named values,
+if options are **not** preserved
+in function outputs,
 defining what it means to compose options
 is not straightforward.
 
--- a/exploration/function-composition-part-1.md
+++ b/exploration/function-composition-part-1.md
@@ -322,6 +322,8 @@ This suggests a representation for named values
 that allows functions to choose whether to
 inspect the "formatted result", the "argument" and options,
 or both.
+Or, it may also suggest that we consider 
+whether we allow composition at all when the functions differ.
 
 ## Background
 
--- a/exploration/function-composition-part-1.md
+++ b/exploration/function-composition-part-1.md
@@ -846,6 +846,18 @@ The proposed solution might involve normative changes to the spec,
 or it might be sufficient to define a set of examples
 and how they should work.
 
+### Stay aligned with user needs
+
+It may be that in practice, MessageFormat users will not
+find it useful to compose functions in complex ways.
+MessageFormat was not initially meant to be
+a general-purpose programming language.
+If supporting function composition in its full generality
+leads to an excessively complex or restrictive spec,
+and if user feedback suggests such use cases are rare in practice,
+it might be more desirable to rule out ambiguous examples
+by means of syntactic restrictions and/or runtime errors.
+
 ## Constraints
 
 _What prior decisions and existing conditions limit the possible design?_
--- a/exploration/function-composition-part-1.md
+++ b/exploration/function-composition-part-1.md
@@ -784,8 +784,6 @@ The question is how to craft the spec in a way that is consistent with expectati
 
 ## Requirements
 
-_What properties does the solution have to manifest to enable the use-cases above?_
-
 In the rest of this document, we assume some version of Model 2.
 However, if Model 1 is more desired, the questions arise of how to
 forbid compositions of functions that would do surprising things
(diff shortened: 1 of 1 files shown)


exploration/function-composition-part-1.md
```diff
@@ -0,0 +1,824 @@
...
+affects observable behavior.
+
+  4. Provide useful names for different kinds of functions
+
+Currently, a given MessageFormat function can be thought of as a
+formatter, a selector, or both.
+Because the implementations for formatters and selectors
+naturally have different type signatures
+(a formatter consumes and produces a resolved value,
+while a selector produces a list of keys),
```
macchiati, line 119: I think this is a flaw in the current specification. I think it would be much more general for a selector to match keys, rather than produce them.

macchiati (approved):
I like this design document; it sets out the options and their consequences pretty clearly.

exploration/function-composition-part-1.md
```diff
@@ -0,0 +1,824 @@
...
+with incorrect uses of functions being resolution errors.
+
+(See "Different kinds of composition" under "Alternatives to consider".)
+
+### Summarizing use cases
+
+There seem to be two areas of ambiguity:
+
+* Are named values essentially strings with metadata,
+or are they structured? (Model 1 vs. Model 2)
```
eemeli, line 731: I'm not sure that I understand what the difference between these is, or what the ambiguity here is. Doesn't a "string with metadata" need to have an object wrapper of some sort providing slots for where to put the metadata? How is that different from being "structured"?

exploration/function-composition-part-1.md
```diff
@@ -0,0 +1,824 @@
...
+as long as it preserves the observable behavior
+of a message.
+
+The proposed solution should not impose
+unnecessary constraints on the implementation.
+However, it _should_ make the requirements
+for "resolved values"
+clear enough so that implementors can
+make a well-informed decision
+on what these types and operations should be.
```
eemeli, line 793: Is this the expected primary outcome of this work? As in, normative text or examples of (custom) functions operating on values with some expected results?

exploration/function-composition-part-1.md
```diff
@@ -0,0 +1,824 @@
...
+One prior decision is that the same definition of
+"resolved value" appears in multiple places in the spec.
+If "resolved value" is defined broadly enough
+(an annotated value with rich metadata),
+then this prior decision need not be changed.
+
+A second constraint is
+the difficulty of developing a precise definition of "resolved value"
+that can be made specific in the interface for custom functions,
+which is implementation-language-neutral.
```
eemeli, line 808: In general, I would be very happy for us to include a more concrete definition or description of "resolved values". Historically, that was one of the first part of the formatting spec that I tried to include, in #198. Unfortunately, that PR ended up being stuck for a year, before needing to be dismissed. The preceding lengthy discussion in #190 is also relevant background, though note of course that these discussions are from over two years ago, so will not fully reflect our current thinking.

Given this past history, I would find it useful to hear @mihnita's current thoughts on defining "resolved value" more explicitly, as he in particular was previously against the idea.

eemeli (commented):
Added some inline comments.

Not having participated in the preceding ICU-TC mailing list discussion, I find the text in the document a little hard to follow. To me, it seems to be muddling together somewhat two spec layers:
1. What's possible in general with functions and resolved values.
2. What's the right thing for `:number` to do.

Untangling those would make it easier to follow what's being considered, as we need answers to both, and the specific answer for `:number` is unlikely to be right for all other functions.

exploration/function-composition-part-1.md
eemeli: As [currently specified](https://github.com/unicode-org/message-format-wg/blob/main/spec/formatting.md#resolve-preferences), keys are indeed "matched", but the result of that match is a preferential list of keys.

It is also possible to simplify this API in an implementation so that the resolved value of the selector is expected to have a `selectKey(keys)` method that returns a single key from its input list, rather than a preferential ordering, though the selection algorithm then gets a bit more complicated as it needs to track additional state (as done [here](https://github.com/messageformat/messageformat/blob/main/packages/mf2-messageformat/src/select-pattern.ts)). This works because the method can be called multiple times, each time leaving out the previously-best key from the input, to determine the full preference order.

catamorphism:
> Added some inline comments.
> 
> Not having participated in the preceding ICU-TC mailing list discussion, I find the text in the document a little hard to follow. To me, it seems to be muddling together somewhat two spec layers:
> 
>     1. What's possible in general with functions and resolved values.
> 
>     2. What's the right thing for `:number` to do.
> 
> 
> Untangling those would make it easier to follow what's being considered, as we need answers to both, and the specific answer for `:number` is unlikely to be right for all other functions.

`:number` is just meant to be an example because it's a known function that has several different options. The same issue could potentially come up with any custom function. I tried to clarify this in [8e31845](https://github.com/unicode-org/message-format-wg/pull/753/commits/8e31845fe6cd777c5fc3292cc462e52b2c0d67ed). 

I suspect that just this one paragraph might not address your concern about untangling the two concerns, though. I'll try to keep that in mind while answering the other comments.

exploration/function-composition-part-1.md
catamorphism: You're right, this is vague. I tried to address this in [09aeff1](https://github.com/unicode-org/message-format-wg/pull/753/commits/09aeff1afa1916f4d74603ebb3cfae6a94bf204c), which runs the hazard of getting too implementation-specific, but I'm not sure how else to express it.

exploration/function-composition-part-1.md
catamorphism: The answer is that I'm not sure, and part of the goal of writing a doc that doesn't present a proposed solution is to narrow down what the outcome should be: normative changes and/or a set of use cases that must be supported (which would also be normative, but would be different from the style of the existing spec.) I was hoping that would happen as part of the discussion of this design doc.

I added some text along those lines in [cfa882a](https://github.com/unicode-org/message-format-wg/pull/753/commits/cfa882a3b9eb6fed277b8c847c5ccefb7918b46c). I also added some text at the end about different directions that a proposed design could go in. If there are any broad categories of solution that I left out, I'd like to include them. 

I think that to answer your question, we have to pick one of those general directions.

exploration/function-composition-part-1.md
catamorphism: I would also be interested to hear from @mihnita -- in terms of the implementations, I took inspiration from his code in defining a `FormattedPlaceholder` type that represents the return value of a function implementation, and trying to figure out how that type should actually work (in combination with #515) is how I got here.

catamorphism: (and I'll take a look at those threads.)

catamorphism: I tried to extract some of the top-level points from the lengthy discussion and added a "prior work" section: [fa625c3](https://github.com/unicode-org/message-format-wg/pull/753/commits/fa625c3185d34a3ad0a3505e37d01f0f945a5680).

If you think I've misrepresented anything or left out an important point, please let me know! I'm hoping that some of the questions that were challenging in those prior discussions might be easier to resolve now due to spec changes that have happened in the meantime. But not having been there for those discussions, I can't be certain of it.

eemeli: Re-reviewing this PR and the changes since my last review, I believe further progress on this is still blocked by waiting on input from @mihnita, as requested above. I would find it important to get buy-in on the general approach before spending more time on this.

exploration/function-composition-part-1.md
```diff
@@ -0,0 +1,974 @@
...
+converts a number to uppercase.
+
+This example only makes sense if `:uppercase`
+operates on the "formatted result"
+of evaluating the _expression_ bound to `$y`.
+
+This suggests a representation for named values
+that allows functions to choose whether to
+inspect the "formatted result", the "argument" and options,
+or both.
```
echeran, line 324: optional, although I think this would help clarity for when drafting a future design doc that takes a position, since we would be less likely to forget options that are on the table
```suggestion
or both.
Or, it may also suggest that we consider 
whether we allow composition at all when the functions differ.
```

exploration/function-composition-part-1.md
```diff
@@ -0,0 +1,974 @@
...
+### Ambiguous examples
+
+Returning to Example Y1, consider two possible models
+of the runtime behavior of function composition.
+
+**Model 1:**
+
+1. Evaluate `$num` to a value and pass it to the `:number` function,
+   along with named options `{"maxFrac": "2"}`
+2. Bind `$x` to the result, which is an object `X`
```
echeran, line 462: semantics question: I always thought you evaluate, get a value, and then bind to symbols/names. which "binds to" which?
```suggestion
2. Bind the result to `$x`, which is an object `X`
```

exploration/function-composition-part-1.md
```diff
@@ -0,0 +1,974 @@
...
+   encapsulating the following fields:
+     * The source value, `"0.33333"`
+     * The fully-evaluated options, `{"maxFrac": "2"}`
+     * The formatted result, a `FormattedNumber` object
+       representing the string `"0.33"`
+3. Evaluate `$y` to a value, which is `X`,
+   and pass it to the `:number` function,
+   along with named options `{"maxFrac": "5"}`
+4. Bind `$y` to the result, which is an object `Y`
```
echeran, line 471: ditto
```suggestion
4. Bind the result to `$y`, which is an object `Y`
```

exploration/function-composition-part-1.md
```diff
@@ -0,0 +1,974 @@
...
+
+In terms of implementation, the result depends on
+what the nature is of the value that is bound to
+a local variable in the environment used in evaluation
+within the message formatter.
+The structure of this value might not be exactly the same
+as the structure that is passed to functions:
+for example, in a lazy implementation, unevaluated thunks
+might be stored in the environment, and then evaluated
+before a function call.
```
echeran, line 511: interesting. that actually sounds like an important point, even if just an implementation detail. because if the thunk is something like `getSystemTimeInMillisNow()`, you want to evaluate and reuse, rather than calling it again & again and getting different user-visible results for the same thing in the same message :-)

echeran (approved):
This design doc is good -- it covers the space of design decisions well around functions whose return values effectively get passed to other functions.

I left a few comments and small optional suggested edits, but shouldn't be considered a blocker.

These issues are important to decide upon so that we can avoid ambiguity and complexity.

exploration/function-composition-part-1.md
```diff
@@ -0,0 +1,974 @@
...
+```
+
+These examples raise the question
+of whether functions return values
+that encode the option names and values
+that were passed in to the function.
+
+While some use cases don't work well
+(or at least work surprisingly)
+if options are **not** encoded in named values,
```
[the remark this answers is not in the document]
catamorphism: I see what you're getting at, but in this case I would rather not introduce the word "assign" with all its connotations; I'm changing it to "if options are **not** preserved in function outputs."

exploration/function-composition-part-1.md
catamorphism: I've usually seen "bind a name to a value", but I rewrote these steps (and the next set of steps) to hopefully make it clearer, in [981df0f](https://github.com/unicode-org/message-format-wg/pull/753/commits/981df0fc454642f55bf58f227bf9b7fda0926a15)

exploration/function-composition-part-1.md
catamorphism: This is indeed an important point, and not just an implementation detail IMO.

Call-by-need and call-by-name implementations of laziness are observationally equivalent (except when it comes to runtime errors) in a language that doesn't have mutability or side effects.

MF2 is immutable and has no side effects, _unless_ functions have side effects that are observable from messages. `getSystemTimeInMillisNow()` is a good example of a function that does.

It's orthogonal to this design doc, I think, but maybe it should be noted somewhere in the spec that if custom functions with observable side effects exist, then careful thought should be put into lazy implementations.

exploration/function-composition-part-1.md
```diff
@@ -0,0 +1,970 @@
...
+
+The problem defined in this design document is that
+the meaning of function composition is ambiguous.
+
+An obstacle to disambiguating its meaning
+is the absence of a definition of "resolved value".
+
+So we must address both problems together:
+
+* Problem 1: Define what it means for functions to compose with each other.
```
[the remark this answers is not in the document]
catamorphism: > I don't know that this is actually a problem. I think it is a side-effect of not resolving Problem 2.

I disagree with this, because I think that even if we resolved Problem 2, we would still have ambiguities.

In your first example: suppose we solve problem 2 by adopting a representation of function results that includes all the options that were passed in. Without further spec work, though, it's _still_ unclear whether `$t` is printed with the time zone or not. Problem 2 is "what do functions return?", while problem 1 is "what do functions do with their inputs?" Maybe the latter question can only be answered in a way that's specific to each function.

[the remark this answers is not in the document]
aphillips: I don't think example 1 is an exemplar of "what do functions do with their inputs" as I don't think that options are necessarily the same thing as "inputs". Also, it's probably not a great example, since Temporal types provide ways to float/unfloat time values (making the time zone a part of the datetime value--that it, modifying the operand, as it were) rather than merely being a random option.

My concern is that, whenever I see "we have to solve two problems at the same time", that suggests extra effort to ensure that there's been complete decomposition of the requirements.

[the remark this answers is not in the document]
catamorphism: Re: decomposition of the requirements, fair enough... what do you think about the change to this section I made in [86c26ff](https://github.com/unicode-org/message-format-wg/pull/753/commits/86c26ff8343913a2d46a1ab5b2cf31c98709a2b1) (emphasizing that this document is mainly about solving the resolved value problem, which is only a part of the bigger function composition problem)?

macchiati:
The following seems contradictory:
```
Model 1:
...
2. Let X be the result of the function. X is an object encapsulating the following fields:

- The source value, "0.33333"
- The fully-evaluated options, {"maxFrac": "2"}
- The formatted result, a FormattedNumber object representing the string "0.33"

Model 2:
2. Let F be the result of the function. F is a FormattedNumber object representing the string "0.33"
```

- Alternative 1: A function returns a "formatted value". **This matches model 1**, where formatted values are bound to names.

- Alternative 2: A function returns a composite value that conceptually pairs a base value (possibly the operand of the function, but possibly not; see Example B1) with options. **This matches model 2**. If we preserve the single usage of "resolved value" in the spec, this implies that the (base value, options) representation applies to all resolved values, not just those returned by functions.


That is, Alternative 1 (formatted value only) seem to match Model 2, not Model 1.

This happens in a few other places as well. I think the readability would be improved dramatically if the Models had names instead of numbers, something like:

Model 1 --> Bundle Model
Model 2 --> Formattable Model

PR #754, opened 2024-03-27
[DESIGN] Bidi usability

aphillips:
Addresses #746.

Provide a design for bidi isolation of MFv2 _syntax_ elements, allowing users and tools to provide messages in right-to-left languages that look normal or have minimal disruption from the Unicode Bidirectional Algorithm.

This design does not require the use of isolates or markers, so it is still possible to create messages that look funky.

This design ignores the UAX31 whitespace definition requirements addressed by #673. Note that the design provided here is more closely tailored to ensure isolate sequences, once opened, are closed and that they tightly wrap only the things desired. There is some funkiness around `name` that we might improve (see the doc).

### DIFF
--- a/exploration/bidi-usability.md
+++ b/exploration/bidi-usability.md
@@ -195,6 +195,17 @@ it should be possible to include "local effect" strongly directional marks in an
 > {⁦$م1‎ :م2‎:ن3‎⁩} with isolates and LRMs
 >```
 
+Newlines inside of messages should not harm later syntax.
+
+```
+* * {{\u0645<br>\u0646}} 123 456 {{ No LRM==bad }}
+* * {{م
+ن}} 123 456 {{  No LRM==bad }}
+
+* * {{\u0645<br>\u0646}}\u200e 123 456 {{ LRM }}
+* * {{م
+ن}}‎ 123 456 {{ LRM }}
+```
 
 ## Constraints
 
@@ -220,6 +231,16 @@ This works at the cost of allowing spurious markers.
 ## Proposed Design
 
 _Describe the proposed solution. Consider syntax, formatting, errors, registry, tooling, interchange._
+
+To start with, we should establish that _message_ editing should always use a left-to-right
+base direction.
+Further, each _line_ of a message should be displayed for editing with a base paragraph direction of LTR.
+This is because the syntax of a _message_ depends on LTR word tokens,
+as well as token ordering (as in a placeholder or with variant keys).
+This is not the disadvantage to RTL languages that it might first appear:
+- Bidi inside of patterns works normally;
+  only placeholders/markup have special usage of bidi controls and this usage is isolated
+  so that placeholders and markup are treated as neutrals.
 
 Permit isolating bidi controls to be used on the **outside** of the following:
 - unquoted literals
--- a/exploration/bidi-usability.md
+++ b/exploration/bidi-usability.md
@@ -41,6 +41,52 @@ _What context is helpful to understand this proposal?_
 If you are unfamiliar with bidirectional or right-to-left text, there is a basic introduction 
 [here](https://www.w3.org/International/articles/inline-bidi-markup/uba-basics).
 
+MessageFormat _message_ strings are created and edited primarily by humans.
+The original _message_ is often written by a software developer or user experience designer.
+Translators need to work with the target-language versions of each _message_.
+Like many templating or domain-specific languages, MFv2 uses neutrally-directional symbols
+to form portions of the syntax.
+When the _message_ contains right-to-left (RTL) translations or uses values that are RTL,
+the plain-text of the message and the Unicode Bidirectional Algorithm (UBA, UAX#9)
+interact in ways that make the _message_ unintelligible or difficult to parse visually.
+
+Machines do not have a problem parsing _messages_ that contain RTL characters,
+but users need to be able to discern what a _message_ does,
+what _variant_ will be selected,
+or what a _placeholder_ will evaluate into.
+
+In addition, it is possible to construct messages that use bidi characters to spoof
+users into believing that a _message_ does something different than what it actually does.
+
+The current syntax does not permit bidi controls in _name_ tokens,
+_unquoted_ literal values,
+or in the whitespace portions of a _message_.
+
+Permitting the **isolate** controls and the standalone strongly-directional markers
+would enable tools, including translation tools, and users who speak RTL languages
+to format a _message_ so that it's plain-text representation and its function
+are unambiguous.
+
+The isolate controls are paired invisible control characters inserted around a portion of a string.
+The start of an isolate sequence is one of:
+- U+2066 LEFT-TO-RIGHT ISOLATE (LRI)
+- U+2067 RIGHT-TO-LEFT ISOALTE (RLI)
+- U+2068 FIRST-STRONG ISOLATE (FSI)
+
+The end of an isolate sequence is U+2069 POP DIRECTIONAL ISOLATE (PDI).
+
+The characters inside an isolate sequence have the initial string (paragraph) direction
+corresponding to the starting control (LTR for LRI, RTL for RLI, auto for FSI).
+The isolate sequence is **isolated** from surrounding text.
+This means that the surrounding text treats it as-if the sequence were a single neutral character.
+
+> [!NOTE]
+> One of the side-effects of using `{`/`}` and `{{`/`}}` to delimit _expressions_
+> and _patterns_ is that these paired enclosing punctuations provide a measure of
+> isolation in UBA.
+> This is an additional reason not to change over to quote marks (which are not enclosing)
+> around patterns.
+
 ## Use-Cases
 
 _What use-cases do we see? Ideally, quote concrete examples._
@@ -128,6 +174,22 @@ Permit isolating bidi controls to be used on the **outside** of the following:
 - quoted literals
 - quoted patterns
 
+This would change the ABNF as follows:
+```abnf
+literal        = ( open-isolate (quoted / unquoted) close-isolate)
+               / (quoted / unquoted)
+quoted-pattern = ( open-isolate "{{" pattern "}}" close-isolate)
+               / ("{{" pattern "}}")
+
+open-isolate   = %x2066-2068
+close-isolate  = %x2069
+```
+
+> [!IMPORTANT]
+> The isolating controls go on the **_outside_** of the various _literal_ and _pattern_
+> productions because characters on the **_inside_** of these are part of the normal text.
+> We need to allow users to include bidi controls in the output of MFv2.
+
 Permit the use of LRM or RLM controls immediately following:
 - name (note that this includes _identifiers_ as well as names of
   _functions_, _variables_, and _unquoted_ literals
@@ -135,8 +197,40 @@ Permit the use of LRM or RLM controls immediately following:
 > The one tricky part with `name` is whether we permit it between the `namespace` and `name`
 > part of an `identifier`.
 
+This would change the ABNF as follows:
+```abnf
+namespace = name-start *name-char ; same as name but lacks bidi close
+name      = name-start *name-char [%x200E-200F]
+```
+
+> [!NOTE]
+> Ideally we do not want RLM/LRM to be part of the `name` or part of any
+> production that consumes `name` (such as `variable`, `reserved-keyword`, or `unquoted`).
+> This is complicated to do in ABNF because each of these tokens is followed either by
+> whitespace or by some closing marker such as `}`.
+> The workaround in #763 is to permit these characters _before_ or _after_ whitespace
+> using the various whitespace productions.
+> This works at the cost of allowing spurious markers.
+
 ## Alternatives Considered
 
 _What other solutions are available?_
 _How do they compare against the requirements?_
 _What other properties they have?_
+
+### Nothing
+We could do nothing.
+
+A likely outcome of doing nothing is that RTL users would insert bidi controls into
+_messages_ in an attempt to make the _pattern_ and/or _placeholders_ to display correctly.
+These controls would become part of the output of the _message_,
+showing up inappropriately at runtime.
+Because these characters are invisible, users might be very frustrated trying to manage
+the results or debug what is wrong with their messages.
+
+By contrast, if users insert too many or the wrong controls using the recommended design,
+the _message_ would still be functional and would emit no undesired characters.
+
+### Deeper Syntax Changes
+We could alter the syntax to make it more "bidi robust", 
+such as by using strongly directional instead of neutrals.
--- a/exploration/bidi-usability.md
+++ b/exploration/bidi-usability.md
@@ -88,6 +88,17 @@ This means that the surrounding text treats it as-if the sequence were a single 
 > This is an additional reason not to change over to quote marks (which are not enclosing)
 > around patterns.
 
+This design also allows for the use of strongly directional marker characters.
+These include:
+- U+200E LEFT-TO-RIGHT MARK (LRM)
+- U+200F RIGHT-TO-LEFT MARK (RLM)
+- U+061C ARABIC LETTER MARK (ALM)
+
+These characters are invisible strongly-directional characters used in bidirectional
+text to coerce certain directional behavior (usually to mark the end of 
+a sequence of characters that would otherwise be ambiguous or interact with
+neutrals or opposite direction runs in an unhelpful way).
+
 ## Use-Cases
 
 _What use-cases do we see? Ideally, quote concrete examples._
@@ -98,6 +109,13 @@ _What use-cases do we see? Ideally, quote concrete examples._
 م2صر 0 {{The {$م2صر} is actually the first key}}
 م2صر * {{This one appears okay}}
 ```
+
+> ![NOTE]
+> The first _variant_ in the use case above is actually:
+>```
+> \u06452\u0635\u0631 0 {{The {$\u06452\u0635\u0631} is actually the first key}}
+>```
+
 
 2. Presentation in an expression can change if portions of the expression
    are not isolated or do not restore LTR order:
@@ -119,8 +137,8 @@ You have {$م1صر‎ :م2صر‎ م3صر‎=م4صر‎} <- LRM after each RTL t
    I don't want to have to manage a lot of paired controls, when I can get the right effect using
    strongly directional mark characters (LRM, RLM, ALM)
 
-4. As a translation tool or MF2 implementation, I want to automatically generate normalized
-   _messages_ that display correctly in RTL languages or containing RTL substrings with minimal user intervention.
+4. As a translation tool or MF2 implementation, I want to automatically generate 
+   _messages_ which display correctly when they contain RTL text or substring with minimal user intervention.
 
 ## Requirements
 
@@ -158,7 +176,7 @@ _option value_,
 or _literal_.
 These controls must not be included into the _identifier_, _name_, _option value_, or _literal_,
 that is, it must be possible to distinguish these characters from the identifier,
-name, value, or literal in question.
+name, option value, or literal in question.
 
 >```
 > You can use {$م1صر‎ :م2صر‎ م3صر‎=م4صر‎}
@@ -201,18 +219,29 @@ close-isolate  = %x2069
 > productions because characters on the **_inside_** of these are part of the normal text.
 > We need to allow users to include bidi controls in the output of MF2.
 
-Permit the use of LRM, RLM, or ALM controls immediately following:
-- name (note that this includes _identifiers_ as well as names of
-  _functions_, _variables_, and _unquoted_ literals
-
-> The one tricky part with `name` is whether we permit it between the `namespace` and `name`
-> part of an `identifier`.
+Permit the use of LRM, RLM, or ALM controls immediately following any of the items that
+**end** with the `name` production the ABNF. 
+This includes _identifiers_ found in the names of
+_functions_ 
+and _options_,
+plus the names of _variables_,
+as well as the contents of _unquoted_ literals.
 
 This would change the ABNF as follows:
 ```abnf
 namespace = name-start *name-char ; same as name but lacks bidi close
 name      = name-start *name-char [%x200E-200F / %x061C]
 ```
+
+> The one tricky part with `name` is whether we permit it between the `namespace` and `name`
+> part of an `identifier`.
+> Consider:
+>```
+> {$a1 :b2:c3}
+> {⁦$م1‎ :م2:ن⁩3‎}
+>```
+> Notice that the namespace is `:م2` and the name is `:ن⁩3`, but the sequence is displayed
+> with a spillover effect.
 
 > [!NOTE]
 > Ideally we do not want RLM/LRM/ALM to be part of the parsed
--- a/exploration/bidi-usability.md
+++ b/exploration/bidi-usability.md
@@ -87,8 +87,9 @@ it should be possible to bidi isolate a _quoted-pattern_.
 > isolate {{البحرين مصر الكويت!}}
 >```
 
-To prevent _placeholders_ or _expressions_ from having spillover effects with other parts of a _message_
-it should be possible to bidi isolate the contents of an _expression_.
+To prevent _markup_, _placeholders_, or _expressions_ from having spillover effects 
+with other parts of a _message_
+it should be possible to bidi isolate the contents of a _markup_ or an _expression_.
 
 >```
 > You can find it in {$مصر}.
@@ -121,6 +122,18 @@ syntax as described in this design so that _messages_ display appropriately as p
 ## Proposed Design
 
 _Describe the proposed solution. Consider syntax, formatting, errors, registry, tooling, interchange._
+
+Permit isolating bidi controls to be used on the **outside** of the following:
+- unquoted literals
+- quoted literals
+- quoted patterns
+
+Permit the use of LRM or RLM controls immediately following:
+- name (note that this includes _identifiers_ as well as names of
+  _functions_, _variables_, and _unquoted_ literals
+
+> The one tricky part with `name` is whether we permit it between the `namespace` and `name`
+> part of an `identifier`.
 
 ## Alternatives Considered
 
--- a/exploration/bidi-usability.md
+++ b/exploration/bidi-usability.md
@@ -70,7 +70,7 @@ are unambiguous.
 The isolate controls are paired invisible control characters inserted around a portion of a string.
 The start of an isolate sequence is one of:
 - U+2066 LEFT-TO-RIGHT ISOLATE (LRI)
-- U+2067 RIGHT-TO-LEFT ISOALTE (RLI)
+- U+2067 RIGHT-TO-LEFT ISOLATE (RLI)
 - U+2068 FIRST-STRONG ISOLATE (FSI)
 
 The end of an isolate sequence is U+2069 POP DIRECTIONAL ISOLATE (PDI).
--- a/exploration/bidi-usability.md
+++ b/exploration/bidi-usability.md
@@ -110,7 +110,8 @@ _What use-cases do we see? Ideally, quote concrete examples._
 م2صر * {{This one appears okay}}
 ```
 
-> ![NOTE]
+> [!NOTE]
+> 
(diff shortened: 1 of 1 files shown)


exploration/bidi-usability.md
```diff
@@ -0,0 +1,236 @@
...
+In addition, it is possible to construct messages that use bidi characters to spoof
+users into believing that a _message_ does something different than what it actually does.
+
+The current syntax does not permit bidi controls in _name_ tokens,
+_unquoted_ literal values,
+or in the whitespace portions of a _message_.
+
+Permitting the **isolate** controls and the standalone strongly-directional markers
+would enable tools, including translation tools, and users who speak RTL languages
+to format a _message_ so that it's plain-text representation and its function
```
macchiati, line 67: Minor
```suggestion
to format a _message_ so that its plain-text representation and its function
```

exploration/bidi-usability.md
```diff
@@ -0,0 +1,236 @@
...
+it should be possible to bidi isolate a _quoted_ or _unquoted_ _literal_.
+
+>```
+> .local $title = {|البحرين مصر الكويت!|}
+> .local $egypt = {مصر :string}
+>```
+
+To prevent _patterns_ from having spillover effects with other parts of a _message_,
+particularly with _keys_ in a _variant_,
+it should be possible to bidi isolate a _quoted-pattern_.
```
macchiati, line 129:
```suggestion
it should be possible to bidi-isolate a _quoted-pattern_.
```

macchiati (approved):

exploration/bidi-usability.md
```diff
@@ -0,0 +1,236 @@
...
+```
+
+> [!IMPORTANT]
+> The isolating controls go on the **_outside_** of the various _literal_ and _pattern_
+> productions because characters on the **_inside_** of these are part of the normal text.
+> We need to allow users to include bidi controls in the output of MFv2.
+
+Permit the use of LRM or RLM controls immediately following:
+- name (note that this includes _identifiers_ as well as names of
+  _functions_, _variables_, and _unquoted_ literals
```
eemeli, line 195: Why LRM/RLM rather than isolates?

Why allow for RLM?

exploration/bidi-usability.md
```diff
@@ -0,0 +1,236 @@
...
+
+This would change the ABNF as follows:
+```abnf
+namespace = name-start *name-char ; same as name but lacks bidi close
+name      = name-start *name-char [%x200E-200F]
+```
+
+> [!NOTE]
+> Ideally we do not want RLM/LRM to be part of the `name` or part of any
+> production that consumes `name` (such as `variable`, `reserved-keyword`, or `unquoted`).
```
eemeli, line 208: I don't think this is an important concern; it's easy to add language to `syntax.md` stating that these characters are not considered as a part of the value of these rules.

exploration/bidi-usability.md
```diff
@@ -0,0 +1,236 @@
...
+syntax as described in this design so that _messages_ display appropriately as plain text.
+
+## Proposed Design
+
+_Describe the proposed solution. Consider syntax, formatting, errors, registry, tooling, interchange._
+
+Permit isolating bidi controls to be used on the **outside** of the following:
+- unquoted literals
+- quoted literals
+- quoted patterns
```
eemeli, line 175: Should we also allow for an LRI/FSI pair immediately inside expressions and markup, or is there a reason not to do so?

exploration/bidi-usability.md
```diff
@@ -0,0 +1,236 @@
...
+showing up inappropriately at runtime.
+Because these characters are invisible, users might be very frustrated trying to manage
+the results or debug what is wrong with their messages.
+
+By contrast, if users insert too many or the wrong controls using the recommended design,
+the _message_ would still be functional and would emit no undesired characters.
+
+### Deeper Syntax Changes
+We could alter the syntax to make it more "bidi robust", 
+such as by using strongly directional instead of neutrals.
```
eemeli, line 236: One way of doing this would be to only allow RTL characters in text and literals, and not in `name`.

exploration/bidi-usability.md
```diff
@@ -0,0 +1,236 @@
...
+The isolate controls are paired invisible control characters inserted around a portion of a string.
+The start of an isolate sequence is one of:
+- U+2066 LEFT-TO-RIGHT ISOLATE (LRI)
+- U+2067 RIGHT-TO-LEFT ISOALTE (RLI)
+- U+2068 FIRST-STRONG ISOLATE (FSI)
+
+The end of an isolate sequence is U+2069 POP DIRECTIONAL ISOLATE (PDI).
+
+The characters inside an isolate sequence have the initial string (paragraph) direction
+corresponding to the starting control (LTR for LRI, RTL for RLI, auto for FSI).
```
eemeli, line 79: Do all editors reset the paragraph direction after a newline? For example, if there's a newline between an LRI and an FSI, how is the paragraph direction of the second line determined?

eemeli (commented):
I didn't add line comments for them, but I continue to note that we should be consistent in calling this spec "MessageFormat 2" or "MF2".

exploration/bidi-usability.md
macchiati: The normal application of the bidi algorithm requires a reset on each paragraph, wherein a newline breaks paragraphs. 

"The algorithm reorders text only within a paragraph; characters in one paragraph have no effect on characters in a different paragraph. Paragraphs are divided by the Paragraph Separator or appropriate Newline Function (for guidelines on the handling of CR, LF, and CRLF, see Section 4.4, Directionality, and Section 5.8, Newline Guidelines of [[Unicode](https://www.unicode.org/reports/tr41/tr41-32.html#Unicode)]). Paragraphs may also be determined by higher-level protocols: for example, the text in two different cells of a table will be in different paragraphs."

exploration/bidi-usability.md
aphillips: Agreed, but I want to call out that these characters are not intended to parse into the name.

exploration/bidi-usability.md
aphillips: We could do that also. It doesn't solve the problem of expression/markup internal bidi, though.

exploration/bidi-usability.md
aphillips: I would only ever support this if `name` were limited to ASCII letters/digits. We use `name` _everywhere_ (including for unquoted). Also, what's the point of allowing a wide range of Unicode, only to disable RTL language users?

exploration/bidi-usability.md
aphillips: @macchiati is correct. That's why it's called "paragraph direction". Note that newlines don't help us that much: they are optional in our syntax (outside literals) and technically normalize to space (or nothing). That is, the newline doesn't help us if we end up writing the message as a single-line.

exploration/bidi-usability.md
eemeli: I'm mostly here thinking of content like:
```js
a = 'אחד'
b = 'שתיים'
s = a + '{$' + b + '}'
```

where we have an RTL variable name inside a placeholder in an RTL pattern.

How, except with an LRI/FSI pair inside the braces, can we get that to render so that the `:` is to the left of the name?

exploration/bidi-usability.md
eemeli: I agree, and would find the limitation to ASCII weird and rather parochially American.

exploration/bidi-usability.md
eemeli: Given that we allow a number of neutral direction characters also in `name-start`, doesn't the same apply to the beginning of the name as well?

From an automation PoV, using FSI/PDI to wrap names seems like it would "just work", whereas LRM/RLM/ALM would require inspecting the contents of the string to figure out what might be needed.

exploration/bidi-usability.md
macchiati: I do too!

exploration/bidi-usability.md
macchiati: See #discussion_r1542105763
For those implementations, RLM/LRM are the best one can do.

exploration/bidi-usability.md
eemeli: Ok, so given that we allow for newlines within "code" and, specifically, expressions, I think we need to account for that so that we can keep the direction of the code as left-to-right, even when the first strongly directional character on the line is RTL.

As I understand it, not even an LRI/FSI pair inside the braces is always enough to keep the `$` on the left side of its name if it's preceded by a newline:

```js
a = 'אחד'
b = 'שתיים'
s = a + '{\u2066\n$' + b + '\u2069}'
```

exploration/bidi-usability.md
```diff
@@ -0,0 +1,236 @@
...
+
+Permitting the **isolate** controls and the standalone strongly-directional markers
+would enable tools, including translation tools, and users who speak RTL languages
+to format a _message_ so that its plain-text representation and its function
+are unambiguous.
+
+The isolate controls are paired invisible control characters inserted around a portion of a string.
+The start of an isolate sequence is one of:
+- U+2066 LEFT-TO-RIGHT ISOLATE (LRI)
+- U+2067 RIGHT-TO-LEFT ISOALTE (RLI)
```
catamorphism, line 73:
```suggestion
- U+2067 RIGHT-TO-LEFT ISOLATE (RLI)
```

exploration/bidi-usability.md
aphillips: If we are strictly talking about machines doing the "bidi annotation", tightly wrapping with isolates will generally work. 

FSI is not always going to get the right results, as not all tokens have the correct direction strongly directional character nearest the front. There is an element of judgement (machines don't have enough information, generally, to decide this, although sometimes they do).

But going back to my initial statement: humans write these strings and create translations of these strings. Sometimes the easiest way for them to make the message look correct is to add a strongly directional mark vs. wrapping. (Note that ICU produces marks on some number and date formats to coerce proper display).

You are correct that the design document should call out these use cases separately so that the reader (and ultimately the WG) can weight supporting these mechanisms appropriately.

macchiati (approved):

exploration/bidi-usability.md
eemeli: @macchiati:
> Some implementations don't handle bidi isolates well yet.

Is there a list available anywhere of software that does not yet support bidi isolates?

macchiati:
The W3C internationalization group might have some information on browsers,
but one would also have to consider other applications and tools.

aphillips:
@macchiati noted:

> The W3C internationalization group might have some information on browsers,
> but one would also have to consider other applications and tools.

@eemeli already spotted this: https://www.w3.org/International/i18n-tests/results/bidi-algorithm

The browsers already support isolates. Other environments, of course, may not (although much progress has been made there too)

exploration/bidi-usability.md
```diff
@@ -0,0 +1,244 @@
...
+> .local $egypt = {مصر :string}
+>```
+
+To prevent _patterns_ from having spillover effects with other parts of a _message_,
+particularly with _keys_ in a _variant_,
+it should be possible to bidi-isolate a _quoted-pattern_.
+
+>```
+> .match {$foo :string}
+> isolate {{البحرين مصر الكويت!}}
```
catamorphism, line 141: Is `isolate` proposed syntax, or a key?

exploration/bidi-usability.md
```diff
@@ -0,0 +1,244 @@
...
+>```
+
+To prevent _markup_, _placeholders_, or _expressions_ from having spillover effects
```
catamorphism, line 144:
```suggestion
To prevent _placeholders_ (_markup_ or _expressions_) from having spillover effects 
```

exploration/bidi-usability.md
```diff
@@ -0,0 +1,244 @@
...
+To prevent _markup_, _placeholders_, or _expressions_ from having spillover effects 
+with other parts of a _message_
+it should be possible to bidi isolate the contents of a _markup_ or an _expression_.
+
+>```
+> You can find it in {$مصر}.
+>```
+
+To prevent RTL identifiers from having spillover effects with other parts of an _expression_,
+it should be possible to include "local effect" bidi controls following an _identifier_,
```
catamorphism, line 153: Can't we omit "identifier" since an identifier ends with a name?

exploration/bidi-usability.md
```diff
@@ -0,0 +1,244 @@
...
+_name_,
+_option value_,
+or _literal_.
+These controls must not be included into the _identifier_, _name_, _option value_, or _literal_,
+that is, it must be possible to distinguish these characters from the value in question.
```
catamorphism, line 158:
```suggestion
that is, it must be possible to distinguish these characters from the term in question.
```

"Value" is very confusing since there are also "option values". I suggest "term" since _names, _option values_, etc. are all terms in the grammar.

Or, it might be more precise to say something like "the characters should not appear in parsed output" (i.e. the relevant nodes in the data model).

exploration/bidi-usability.md
```diff
@@ -0,0 +1,244 @@
...
+quoted-pattern = ( open-isolate "{{" pattern "}}" close-isolate)
+               / ("{{" pattern "}}")
+
+open-isolate   = %x2066-2068
+close-isolate  = %x2069
+```
+
+> [!IMPORTANT]
+> The isolating controls go on the **_outside_** of the various _literal_ and _pattern_
+> productions because characters on the **_inside_** of these are part of the normal text.
```
catamorphism, line 198:
```suggestion
> productions because characters on the **_inside_** of these are uninterpreted.
```

exploration/bidi-usability.md
```diff
@@ -0,0 +1,244 @@
...
+> We need to allow users to include bidi controls in the output of MF2.
```
catamorphism, line 199: Is this worth adding to the "constraints" section? Constraint: we must allow bidi controls as either literal (interpreted by the MF2 parser) or escaped (treated as regular text). (Their position introduces an implicit escape.)

exploration/bidi-usability.md
```diff
@@ -0,0 +1,244 @@
...
+
+Permit the use of LRM, RLM, or ALM controls immediately following:
+- name (note that this includes _identifiers_ as well as names of
+  _functions_, _variables_, and _unquoted_ literals
```
catamorphism, line 203:
```suggestion
- name (note that this includes _unquoted_ literals_, _identifiers_, and _variables_;
and that _identifiers_ include the names of _functions_.)
```

(It's a bit confusing to say that an unquoted literal has a name.)

exploration/bidi-usability.md
```diff
@@ -0,0 +1,244 @@
...
+
+This would change the ABNF as follows:
+```abnf
+namespace = name-start *name-char ; same as name but lacks bidi close
+name      = name-start *name-char [%x200E-200F / %x061C]
+```
+
+> [!NOTE]
+> Ideally we do not want RLM/LRM/ALM to be part of the `name` or part of any
+> production that consumes `name` (such as `variable`, `reserved-keyword`, or `unquoted`).
```
catamorphism, line 216:
```suggestion
> Ideally we do not want RLM/LRM/ALM to be part of the parsed
> `name`, `variable`, `reserved-keyword`, `unquoted`, or any other term
> defined in terms of `name`.
```

It reads a bit oddly to me to say that a production "consumes" something (the grammar is declarative; parsers consume things.)

exploration/bidi-usability.md
```diff
@@ -0,0 +1,244 @@
...
+
+_What other solutions are available?_
+_How do they compare against the requirements?_
+_What other properties they have?_
+
+### Nothing
+We could do nothing.
+
+A likely outcome of doing nothing is that RTL users would insert bidi controls into
+_messages_ in an attempt to make the _pattern_ and/or _placeholders_ to display correctly.
```
catamorphism, line 233:
```suggestion
_messages_ in an attempt to make the _pattern_ and/or _placeholders_ display correctly.
```

exploration/bidi-usability.md
catamorphism: Even if it's not an alternative anyone would support, maybe it's worth documenting Eemeli's suggestion as another alternative considered, if only to say it's an immediate non-starter.

catamorphism (changes requested):
Feel free to ignore some of my comments if you think it's already clear to a less naïve reader. However, there are a few things that are really ambiguous to me, like the syntax of examples.

exploration/bidi-usability.md
```diff
@@ -0,0 +1,244 @@
...
+
+If you are unfamiliar with bidirectional or right-to-left text, there is a basic introduction 
+[here](https://www.w3.org/International/articles/inline-bidi-markup/uba-basics).
+
+MessageFormat _message_ strings are created and edited primarily by humans.
+The original _message_ is often written by a software developer or user experience designer.
+Translators need to work with the target-language versions of each _message_.
+Like many templating or domain-specific languages, MF2 uses neutrally-directional symbols
+to form portions of the syntax.
+When the _message_ contains right-to-left (RTL) translations or uses values that are RTL,
```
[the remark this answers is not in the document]
aphillips: Values here means identifiers or tokens, that is, syntactic values such as `$مصر` or ` ⁦:مصر⁩ ` or an option name or value. I don't mean the "resolved value" behind a variable name.

exploration/bidi-usability.md
```diff
@@ -0,0 +1,244 @@
...
+the plain-text of the message and the Unicode Bidirectional Algorithm (UBA, UAX#9)
+interact in ways that make the _message_ unintelligible or difficult to parse visually.
+
+Machines do not have a problem parsing _messages_ that contain RTL characters,
+but users need to be able to discern what a _message_ does,
+what _variant_ will be selected,
+or what a _placeholder_ will evaluate into.
```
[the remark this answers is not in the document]
aphillips: Developers write messages too. It's from both perspectives.

exploration/bidi-usability.md
```diff
@@ -0,0 +1,244 @@
...
+
+In addition, it is possible to construct messages that use bidi characters to spoof
+users into believing that a _message_ does something different than what it actually does.
+
+The current syntax does not permit bidi controls in _name_ tokens,
+_unquoted_ literal values,
+or in the whitespace portions of a _message_.
+
+Permitting the **isolate** controls and the standalone strongly-directional markers
+would enable tools, including translation tools, and users who speak RTL languages
```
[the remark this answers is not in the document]
aphillips: that's a reasonable change, although really the verb we want is "write" 😉

exploration/bidi-usability.md
```diff
@@ -0,0 +1,244 @@
...
+
+## Use-Cases
+
+_What use-cases do we see? Ideally, quote concrete examples._
+
+1. Presentation of keys can change if values are not isolated:
+```
+.match {$م2صر :string}{$num :integer}
+م2صر 0 {{The {$م2صر} is actually the first key}}
+م2صر * {{This one appears okay}}
```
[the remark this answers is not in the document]
aphillips: The output isn't the issue here.

The first variant is actually:

```
\u06452\u0635\u0631 0 {{The {$\u06452\u0635\u0631} is actually the first key}}
```

That is, the first key is `م2صر `, not `0` (as it appears to be visually). The `\u` encoding shows the _actual logical order_ of the keys. Try selecting the text on that line...

exploration/bidi-usability.md
```diff
@@ -0,0 +1,244 @@
...
+```
+
+2. Presentation in an expression can change if values are not isolated or restore LTR order:
```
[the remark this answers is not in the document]
aphillips: It means the tokens in the syntax.

exploration/bidi-usability.md
aphillips: Not sure I follow? The point of the note is to show that bidi controls are just normal text _inside_ literal contexts (the body of a pattern or inside of quoted literals)

exploration/bidi-usability.md
aphillips: Perhaps, but that's how `unquoted` is defined:

```abnf
unquoted       = name / number-literal
```

exploration/bidi-usability.md
eemeli: A key.

exploration/bidi-usability.md
eemeli: The spillover can also occur in declarations and the `.match` statement. It won't have an effect on the parsing, but the appearance to a user.

exploration/bidi-usability.md
aphillips: Identifiers end with names, but also contain names in the namespace position. I wanted to be clear that we meant the end of an identifier in this case.

exploration/bidi-usability.md
aphillips: I disagree with this change. Perhaps omit "normal" or perhaps:

```suggestion
> productions because characters on the **_inside_** of these are not part of the MF2 syntax.
```

exploration/bidi-usability.md
aphillips: I changed the approach to make this text moot. I then moved this text to the "constraints" section, in hopes that the design addresses it.

exploration/bidi-usability.md
aphillips: That's correct. Getting the sigils to stay on the left side needs a base direction of LTR. An LRM doesn't help in your example either (except to prevent spillover with the following annotation if there were any). My proposal is not 100% bulletproof (and requires some action on the part of tools or users).

A bulletproof design would require more isolates and would probably be limited to using LRI/PDI pairs. It would be difficult to work with, given that there would be a lot of invisible control characters inside subcomponents of an expression, e.g.:

```
<LRI><LRI>option<PDI>[whitespace]=[whitespace]<LRI>value<PDI><PDI>
```

exploration/bidi-usability.md
```diff
@@ -0,0 +1,244 @@
...
+> the string to make the bidi effects visible.
+> The numbers correspond to:
+> 1. operand
+> 2. function
+> 3. option name
+> 4. option value
+
+```
+You have {$م1صر :م2صر م3صر=م4صر} <- no controls
+You have {$م1صر‎ :م2صر‎ م3صر‎=م4صر‎} <- LRM after each RTL token
```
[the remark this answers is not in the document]
aphillips: fixed

exploration/bidi-usability.md
```diff
@@ -0,0 +1,244 @@
...
+```
+
+3. As a developer or translator, I want to make RTL literal or names appear correctly
+   in my plain-text editing environment.
+   I don't want to have to manage a lot of paired controls, when I can get the right effect using
+   strongly directional mark characters (LRM, RLM, ALM)
```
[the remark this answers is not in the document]
aphillips: fixed

exploration/bidi-usability.md
```diff
@@ -0,0 +1,244 @@
...
+
+## Requirements
+
+_What properties does the solution have to manifest to enable the use-cases above?_
+
+To prevent RTL _literals_ from having spillover effects with surrounding syntax,
+it should be possible to bidi isolate a _quoted_ or _unquoted_ _literal_.
+
+>```
+> .local $title = {|البحرين مصر الكويت!|}
```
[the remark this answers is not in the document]
catamorphism: Ah, my eyes were deceiving me. I thought the literal ended here: |الكويت -- but I see that that's a different character that just looks like a `|`.

catamorphism (approved):
I don't have time to re-read in detail, but given that it's a design doc, I think that's fine; I followed it for the most part, and the parts that weren't clear to me are probably due to my lack of familiarity with the area.

exploration/bidi-usability.md
```diff
@@ -0,0 +1,329 @@
...
+- expressions
+- markup
+
+This would change the ABNF as follows (assuming the above changes are also incorporated):
+```abnf
+expression            = "{" open-isolate (literal-expression / variable-expression / annotation-expression) close-isolate "}"
+                      / "{" (literal-expression / variable-expression / annotation-expression) "}"
+literal-expression    = [s] literal [s annotation] *(s attribute) [s]
+variable-expression   = [s] variable [s annotation] *(s attribute) [s]
+annotation-expression = [s] annotation *(s attribute) [s]
```
eemeli, line 258: These aren't changing at all, right?
```suggestion
```

exploration/bidi-usability.md
```diff
@@ -0,0 +1,329 @@
...
+> or _pattern_'s textual content.
+> We need to allow users to include bidi controls in the output of MF2.
+
+Permit isolating bidi controls to be used **immediately inside** the following:
+- expressions
+- markup
+
+This would change the ABNF as follows (assuming the above changes are also incorporated):
+```abnf
+expression            = "{" open-isolate (literal-expression / variable-expression / annotation-expression) close-isolate "}"
```
eemeli, line 254: At the start we should only allow `%x2066`, i.e. LTR isolation.

eemeli (changes requested):
> This adds pressure on the parser, since there are four "end of markup" sequences: `}`, `\u2069}`, `/}` and `/\u2069}`.

Only two of those are valid at a time, though, as it's not ok to skip the end if you start it.

exploration/bidi-usability.md
aphillips: They are changed. Originally they contained the `"{"` and `"}"` markers. I moved those up a level (which is debateable) to `expression` to keep the noise level down.

exploration/bidi-usability.md
aphillips: We permit RLI elsewhere. We *could* insist on left-to-right order. Allowing RLI might help with expressions that contain only an RTL literal:

```
{⁧لت-123-م...⁩} // RLI isolated
{لت-123-م...}
```

eemeli: While that could work for rendering, it's conceptually really rather messy. We should in that case allow the isolation of the expression contents, _and_ the literal. So we'd have:
```js
s = 'لت-123-م'
e = '{\u2066\u2067' + s + '\u2069\u2069}'
```
where the outer `\u2066..\u2069` is for the expression and the inner `\u2067..\u2069` for the literal.

My point here is that immediately within braces we're always in "code" mode, and that should _always_ be LTR. Right inside that context we could well have a RTL literal, but for that the isolation shouldn't be coming from the `expression` rule, but the `literal` one.

aphillips: I understand. I also need to add the explicit "these message things are always LTR" to the design doc.

eemeli: This is still not fixed. As expressions are code, and code is always LTR, the `expression` rule should not allow for anything other that an LRI.

Yes, this does introduce an ambiguity if the expression has a literal argument that can also be LRI-isolated, but it's pretty much the same ambiguity as we already have about whitespace.

exploration/bidi-usability.md
```diff
@@ -0,0 +1,350 @@
...
+variable-expression   = [s] variable [s annotation] *(s attribute) [s]
+annotation-expression = [s] annotation *(s attribute) [s]
+markup = "{" [s] "#" identifier *(s option) *(s attribute) [s] ["/"] "}"                             ; open and standalone
+       / "{" [s] "/" identifier *(s option) *(s attribute) [s] "}"                                   ; close
+       / "{" open-isolate [s] "#" identifier *(s option) *(s attribute) [s] ["/"] close-isolate "}"  ; open and standalone
+       / "{" open-isolate [s] "/" identifier *(s option) *(s attribute) [s] close-isolate "}"        ; close
+```
+
+Permit the use of LRM, RLM, or ALM stronly directional marks immediately following any of the items that
+**end** with the `name` production in the ABNF.
```
eemeli, line 287: I still think this is too messy, and doesn't solve the problem as well as isolation, but I'm ok with considering that separately.

My preferred overall solution would be to:
1. Optionally LR/RL/FS -isolate `quoted-pattern`, `quoted`, and `name`;
2. Optionally LR-isolate `expression`; and
3. Allow for a single LRM after a newline in code, or may at the end of whitespace containing a newline.

Put together, that should allow for rendering all code as LTR, and all possibly-RTL content as RTL.

eemeli (changes requested):

eemeli (approved):

PR #616, opened 2024-01-26
Add docs/design/, with accepted & rejected design documents

eemeli:
Context: #523

This moves the accepted & rejected design docs to a more likely location under `docs/design/`, and fixes their titles & metadata. A README is added to that folder.

[Delimiting of Patterns in Complex Messages](https://github.com/unicode-org/message-format-wg/blob/main/exploration/delimiting-variant-patterns.md) is left in `exploration/`. Even though it's marked as "Accepted", it does not contain a description of the selected design. Its previous status of "Balloting" was updated in #559.

PR #755, opened 2024-03-27
[DESIGN] Effect of selectors on placeholders

### LINKED ISSUE #736: Match expressions should be re-usable in placeholders
While preparing yet another presentation on MF2, I needed to write a simple example that got me thinking:
```
.input {$count :number}
.match {$count}
one {{You have {$count} apple}}
* {{You have {$count} apples}}
```

It's not necessarily obvious that the above is what we expect, when this _looks_ like it'll work just as well:

```
.match {$count :number}
one {{You have {$count} apple}}
* {{You have {$count} apples}}
```

However, that doesn't format the `$count` explicitly as a number; we just presume it does, because "count" sounds numeric. 

So the thought I had here is that this is pretty much exactly why & how MF1 ended up with `#` being special in plural selectors, and that the solution we're providing is much less obvious and requires writing a whole new `.input` statement.

Could we consider making the `.match` expressions also act as implicit declarations, and make them usable in placeholders? The somewhat obvious way to address them is by index position:

```
.match {$count :number}
one {{You have {$0} apple}}
* {{You have {$0} apples}}
```

Assigning values to `$0`, `$1`, ... would not conflict with any input values, as numbers are invalid `name-start` characters. That's by design so that we encourage at least _some_ name for each variable; here that's effectively provided by the `.match` expressions.

I suspect that adding this shorthand would provide a more ergonomic solution for most `.input` use cases, and would enable the representation of many messages without any declarations, which currently would require one to avoid significant repetition.

The syntax change required by this would probably look something like this:
```diff
-variable = "$" name
+variable = "$" (name / %x30-39)
```
with accompanying spec language making numeric variables resolve to the `.match` selectors in placeholders, and a data model error otherwise.

aphillips:
Fixes #736, Works on #747 

This design document describes the effect of selectors on subsequent pattern placeholders of the same operand. It presents several alternatives.

This is still in "DRAFT" form and not yet ready for the WG as a whole.

### DIFF
--- a/exploration/selection-declaration.md
+++ b/exploration/selection-declaration.md
@@ -156,7 +156,7 @@ _What other properties they have?_
 
 In this alternative, we modify the syntax to allow selectors to 
 annotate an input variable (as with `.input`)
-or assign a local variable (as with `.local`).
+or bind a local variable (as with `.local`).
 Either annotation is immutable and results in a Duplicate Declaration error
 if it attempts to annotate a variable previously annotated.
 
--- a/exploration/selection-declaration.md
+++ b/exploration/selection-declaration.md
@@ -66,6 +66,44 @@ one {{You have {$togo} mile to go.}}
 
 _What use-cases do we see? Ideally, quote concrete examples._
 
+1. As a user, I want my formatting to match my selector.
+   This is one of the reasons why MF2 requires that selectors be annotated.
+   When I write a selector, the point is to choose the pattern to use as a template
+   for formatting the value being selected on.
+   Mismatches between these are undesirable.
+
+   ```
+   .match {$num :number minimumFractionDigits=1}
+   one {{This case can never happen in an English locale}}
+   *   {{I expect this formats num with one decimal place: {$num}}}
+   ```
+
+2. As a user, I want to use the least amount of MF special syntax possible.
+3. As a user, I don't want to repeat formatting, particularly in large selection matrices.
+   ```
+   .match {$num1 :integer} {$num2 :number minimumFractionDigits=1}
+   0    0    {{You have {$num1 :integer} ({$num2 :number minimumFractionDigits=1}) wildebeest.}}
+   0    one  {{You have {$num1 :integer} ({$num2 :number minimumFractionDigits=1}) wildebeest.}}
+   0    *    {{You have {$num1 :integer} ({$num2 :number minimumFractionDigits=1}) wildebeest.}}
+   one  0    {{ }}
+   one  one  {{ }}
+   one  *    {{ }}
+   // more cases for other locales that use two/few/many
+   *    0    {{ }}
+   *    one  {{ }}
+   *    *    {{ }}
+   ```
+
+4. As a user (especially as a translator), I don't want to have to modify
+   declarations and selectors to keep them in sync.
+   ```
+   .input {$num :number minimumFractionDigits=1}
+   .match {$num}
+   * {{Shouldn't have to modify the selector}}
+   ```
+   > Note that this is currently provided for by the spec.
+
+
 **Multiple Selection Conflicts**
 If a user writes multiple selectors for the same operand, which one formats the placeholder?
 
@@ -77,9 +115,15 @@ If a user writes multiple selectors for the same operand, which one formats the 
 * * {{Which selector formats {$num}?}}
 ```
 
-If both formats are needed in the message, how does one reference one or the other?
+If both formats are needed in the message (presumably they are or why the selector), 
+how does one reference one or the other?
 
 **Selection Different From Format**
+If a user wants to separate selection and formatting, 
+but selectors have a formatting side-effect,
+how do I write a selector that doesn't interfere with the formatting of the _same_ operand?
+(Note that users could do this using a `.local` declaration to split the operand)
+
 We don't support selection on dates in LDML45, but it's easy to conceptualize.
 How can a user write a _selector_ that doesn't mess up a _formatter_?
 
@@ -107,3 +151,91 @@ _Describe the proposed solution. Consider syntax, formatting, errors, registry, 
 _What other solutions are available?_
 _How do they compare against the requirements?_
 _What other properties they have?_
+
+#### Do What Comes Last
+
+In this alternative, selectors override declarations.
+
+```
+.input {$num :integer}
+.match {$num :number minimumFractionDigits=1}
+* {{Formats {$num} like 1.0}}
+```
+
+**Pros**
+- ... thinking, thinking...
+
+**Cons**
+- Violates immutability that we've established everywhere else
+
+#### Declaring Selectors with Immutability
+
+In this alternative, selectors are treated as declaration-selectors.
+That is, an annotation in a selector works like a `.input`.
+However, it is an error for the selector to try to modify a previous declaration
+(just as it is an error for a declaration to try to modify a previous declaration).
+This permits `.match` selectors to be a shorthand when no declarations exist.
+
+It is also an error for a selector to modify a previous selector
+
+```
+.match {$num :integer}
+* {{Formats {$num} as integer}}
+
+.input {$num :integer}
+.match {$num :number maximumFractionDigits=0}
+* {{This message produces a Duplicate Declaration error}}
+
+.input {$num :integer} {$num :number}
+* * {{This message produces a Duplicate Declaration error}}
+```
+
+**Pros**
+- Shorthand version works intuitively with minimal typing
+
+**Cons**
+- Selectors can't provide additional selection-specific options
+  if the value has already been annotated
+- Doesn't allow multiple selection on the same operand, e.g.
+  ```
+  .input {$date :datetime skeleton=yMdjm}
+  .match {$date :datetime field=month} {{$date ::datetime field=dayOfWeek}
+  * * {{This message produces a Duplicate Declaration error
+        even though selection is separate from formatting.}}
+  ```
+
+#### Provide a `#`-like Feature
+
+(Copy-pasta adapted from @eemeli's proposal in #736)
+
+Make the `.match` expression also act as implicit declarations accessed by index position:
+
+```
+.match {$count :number}
+one {{You have {$0} apple}}
+* {{You have {$0} apples}}
+```
+
+Assigning values to `$0`, `$1`, ... would not conflict with any input values, 
+as numbers are invalid `name-start` characters.
+That's by design so that we encourage at least _some_ name for each variable; 
+here that's effectively provided by the `.match` expressions.
+
+ABNF would be modified:
+```diff
+-variable = "$" name
++variable = "$" (name / %x30-39)
+```
+
+...with accompanying spec language making numeric variables resolve to the `.match` selectors in placeholders, 
+and a data model error otherwise.
+
+**Pros**
+- More ergonomic for most `.input` cases
+- Enables representation of many messages without any declarations
+
+**Cons**
+- Confusing that the operand name can't be used in the pattern?
+  Removes some self-documentation from the pattern.
+- Requires the pattern to change if the selectors are modified.
+- Limits number of referenceable selectors to 10 (in the current form)
--- a/exploration/selection-declaration.md
+++ b/exploration/selection-declaration.md
@@ -176,6 +176,7 @@ Examples:
 - No changes required.
 - `.local` can be used to solve problems with variations in selection and formatting
 - Supports multiple selectors on the same operand
+
 **Cons**
 - May require the user to annotate the operand for both formatting and selection.
 - Can produce a mismatch between formatting and selection, since the operand's formatting
--- a/exploration/selection-declaration.md
+++ b/exploration/selection-declaration.md
@@ -19,7 +19,7 @@ Status: **Proposed**
 _What is this proposal trying to achieve?_
 
 Define what effect (if any) the _annotation_ of a _selector_ has on subsequent _placeholders_
-that access the same _operand_ value.
+that access the same _variable_.
 
 ## Background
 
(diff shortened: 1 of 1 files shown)


aphillips:
Adding some review requests... looking for early feedback and suggestions.

exploration/selection-declaration.md
```diff
@@ -0,0 +1,296 @@
...
+_How do they compare against the requirements?_
+_What other properties they have?_
+
+### Allow both local and input declarative selectors with immutability
+
+In this alternative, we modify the syntax to allow selectors to 
+annotate an input variable (as with `.input`)
+or assign a local variable (as with `.local`).
+Either annotation is immutable and results in a Duplicate Declaration error
+if it attempts to annotate a variable previously annotated.
```
catamorphism, line 161:
```suggestion
if it attempts to re-declare a previously declared variable.
```

exploration/selection-declaration.md
```diff
@@ -0,0 +1,296 @@
...
+.match {$input :function} $local = {$input :function}
+* * {{This annotates {$input} and assigns {$local} a value.}}
+
+.match $local1 = {$input :function} $local2 = {$input :function2}
+* * {{This assigns two locals}}
+
+.input {$input :function}
+.local $local = {$input :function}
+.match {$input :function} {$local :function}
+* * {{This produces two duplicate declaration errors.}}
```
catamorphism, line 174: Why? I only see one declaration for each of the two variables.

exploration/selection-declaration.md
```diff
@@ -0,0 +1,296 @@
...
+
+**Pros**
+- Shorthand is consistent with the rest of the syntax
+- Shorthand version works intuitively with minimal input
+- Preserves immutability
+- Produces an error when users inappropriately annotate some items
+
+**Cons**
+- Selectors can't provide additional selection-specific options
+  if the value has already been annotated
```
catamorphism, line 191:
```suggestion
  if the variable name is already in scope
```

exploration/selection-declaration.md
```diff
@@ -0,0 +1,296 @@
...
+- Doesn't allow multiple selection on the same operand, e.g.
+  ```
+  .input {$date :datetime skeleton=yMdjm}
+  .match {$date :datetime field=month} {$date :datetime field=dayOfWeek}
+  * * {{This message produces a Duplicate Declaration error
+        even though selection is separate from formatting.}}
```
catamorphism, line 197: As with the previous example, I'm confused. I thought that in this alternative, you need the '=' sign for a declaration, and the `.match` here wouldn't introduce any declarations.

exploration/selection-declaration.md
```diff
@@ -0,0 +1,296 @@
...
+  ```
+  However, this design does allow for a local variable to be easily created
+  for the purpose of selection.
+
+#### Allow "Declarative" Selectors with _Mutability_
+
+In this alternative, selectors are treated as declaration-selectors.
+That is, an annotation in a selector works like a `.input`.
```
catamorphism, line 205: That doesn't seem right to me. `.input` can only take an argument (free variable), but I think what you're proposing is for selectors to also shadow local variables (bound variables).

exploration/selection-declaration.md
```diff
@@ -0,0 +1,296 @@
...
+This permits `.match` selectors to be a shorthand when no declarations exist.
```
catamorphism, line 206: I don't understand what this means.

exploration/selection-declaration.md
```diff
@@ -0,0 +1,296 @@
...
+
+It is not an error to overwrite a previous annotation.
```
catamorphism, line 208:
```suggestion
It is not an error to re-declare a variable that is in scope.
```

exploration/selection-declaration.md
```diff
@@ -0,0 +1,296 @@
...
+Instead the selector's annotation replaces what came before.
```
catamorphism, line 209:
```suggestion
Instead the selector implicitly shadows the name of its variable operand.
```

(This is another place where the distinction between literal and variable operands is important. You can shadow variables, but changing the value of a literal doesn't make sense.)

exploration/selection-declaration.md
```diff
@@ -0,0 +1,296 @@
...
+- Shorthand version works intuitively with minimal typing.
+
+**Cons**
+- Violates immutability that we've established everywhere else
+
+#### Allow "Declarative" Selectors with _Immutability_
+
+In this alternative, selectors are treated as declaration-selectors.
+That is, an annotation in a selector works like a `.input`.
+However, it is an error for the selector to try to modify a previous declaration
```
catamorphism, line 227:
```suggestion
However, it is an error for the selector to shadow a previous declaration
```

exploration/selection-declaration.md
```diff
@@ -0,0 +1,296 @@
...
+(just as it is an error for a declaration to try to modify a previous declaration).
+This permits `.match` selectors to be a shorthand when no declarations exist.
+
+It is also an error for a selector to modify a previous selector.
```
catamorphism, line 231:
```suggestion
It is also an error for a selector to shadow a previous selector.
```

(Again, this needs some qualifiers... writing `.match {-1 :integer} {-1 :number}` shouldn't be an error.)

exploration/selection-declaration.md
```diff
@@ -0,0 +1,296 @@
...
+This implies that multiple selecton on the same operand is pointless.
```
catamorphism, line 232: Not so much "pointless" as "always emitting an error"?

exploration/selection-declaration.md
```diff
@@ -0,0 +1,296 @@
...
+
+...with accompanying spec language making numeric variables resolve to the `.match` selectors in placeholders, 
+and a data model error otherwise.
+
+**Pros**
+- More ergonomic for most `.input` cases
+- Enables representation of many messages without any declarations
+
+**Cons**
+- Confusing that the operand name can't be used in the pattern?
```
catamorphism, line 293: Why does the operand name have to be disallowed from appearing in the pattern?

exploration/selection-declaration.md
```diff
@@ -0,0 +1,296 @@
...
+  Removes some self-documentation from the pattern.
+- Requires the pattern to change if the selectors are modified.
+- Limits number of referenceable selectors to 10 (in the current form)
```
catamorphism, line 296: I think one more alternative is missing, which is to do nothing: require a declaration when you want a declaration instead of introducing a third (maybe third and fourth?) declaration form, and maintain the invariant that the meaning of a variable name is the same regardless of where it appears.

catamorphism (commented):
This does a good job of laying out the alternatives. The use of "side effect" or "mutable" should be avoided, since the only thing being talked about is name binding, and likewise, since we're talking about name binding, it should be clarified when "operands" really means "variables". On the whole, I was able to follow it and I can't think of any missing alternatives (except for doing nothing.)

exploration/selection-declaration.md
```diff
@@ -0,0 +1,296 @@
...
+- Selectors can't provide additional selection-specific options
+  if the value has already been annotated
+- Doesn't allow multiple selection on the same operand, e.g.
+  ```
+  .input {$date :datetime skeleton=yMdjm}
+  .match {$date :datetime field=month} {$date :datetime field=dayOfWeek}
+  * * {{This message produces a Duplicate Declaration error
+        even though selection is separate from formatting.}}
+  ```
+
```
eemeli, line 261: Let's add my proposal from https://github.com/unicode-org/message-format-wg/issues/736#issuecomment-2016408105:

````suggestion

#### Match on variables instead of expressions

In this alternative, the `.match` syntax is simplified
to work on variable references rather than expressions:

```
.input {$count :number}
.match $count
one {{You have {$count} apple}}
* {{You have {$count} apples}}

.local $empty = {$theList :isEmpty}
.match $empty
true {{You bought nothing}}
* {{You bought {$theList}!}}
```

The ABNF change would look like:
```diff
 match-statement   = match 1*([s] selector)
-selector          = expression
+selector          = variable
```


**Pros**
- Overall the syntax is simplified.
- Preserves immutability.

**Cons**
- A separate declaration is required for each selector.

````

File: exploration/selection-declaration.md (hunk 10 above)
Comment (aphillips) [reply exploration/selection-declaration.md:231]: The example you gave doesn't modify a previous selector: the literal `-1` in the first is a separate literal from that in the second. "Shadow" doesn't really say what is going on?

File: exploration/selection-declaration.md (hunk 11 above)
Comment (aphillips) [reply exploration/selection-declaration.md:232]: No, one can do this just fine:
```
.input {$foo :number}
.match {$foo} {$foo}
one one {{why select twice, when the second selector will rank the same as the first?}}
```

File: exploration/selection-declaration.md (hunk 12 above)
Comment (aphillips) [reply exploration/selection-declaration.md:324]: Perhaps not worded correctly. You can use the operand name in the pattern, but it's annotation won't/might not be applied.

```
.match {$count :integer}
* {{You have {$count} apples might print 2 or 2.0 with input count=|2.0|}}
```

File: exploration/selection-declaration.md (hunk 13 above)
Comment (aphillips) [reply exploration/selection-declaration.md:327]: Will add

File: exploration/selection-declaration.md (hunk 2 above)
Comment (aphillips) [reply exploration/selection-declaration.md:174]: `.input {$input}` and `.match {$input}` both "declare" `$input` in this design. Ditto the `.local` and later selector.

File: exploration/selection-declaration.md (hunk 12 above)
Comment (catamorphism) [reply exploration/selection-declaration.md:324]: Right, so maybe something like: "The meaning of the operand name in the context of the pattern might not be what the user expects"?

File: exploration/selection-declaration.md (hunk 10 above)
Comment (catamorphism) [reply exploration/selection-declaration.md:231]: If a selector implicitly introduces a name, then what you're referring to as "modifying" is actually "shadowing". If you write, in this scenario:

```
.match {$x :integer} {$x :number} 
```

then the second `$x` doesn't modify the first one. Rather, it introduces a new name `$x`, so that in the patterns in the variants, `$x` now has a different meaning. That's what I mean by name shadowing.

File: exploration/selection-declaration.md [hunk 15]
```diff
@@ -0,0 +1,296 @@
...
+.match {$num :number minimumFractionDigits=2} {$num :integer}
+* * {{Which selector formats {$num}?}}
+```
+
+If both formats are needed in the message (presumably they are or why the selector), 
+how does one reference one or the other?
+
+### Selection Is Different From Format
+If a user wants to separate selection and formatting, 
+but selectors have a formatting side-effect,
```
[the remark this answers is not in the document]
aphillips: I converted the bottom two sub-section-y things to user stories and reworded.

exploration/selection-declaration.md
```diff
@@ -0,0 +1,296 @@
...
+how do I write a selector that doesn't interfere with the formatting of the _same_ operand?
+(Note that users could do this using a `.local` declaration to split the operand)
```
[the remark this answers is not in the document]
aphillips: Refactored and included an example.

exploration/selection-declaration.md
```diff
@@ -0,0 +1,296 @@
...
+   ```
+   .input {$num :number minimumFractionDigits=1}
+   .match {$num}
+   * {{Shouldn't have to modify the selector}}
+   ```
+   > Note that this is currently provided for by the spec.
+
+
+### Multiple Selection Conflicts
+If a user writes multiple selectors for the same operand, which one formats the placeholder?
```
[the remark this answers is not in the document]
aphillips: Done.

exploration/selection-declaration.md
```diff
@@ -0,0 +1,296 @@
...
+
+```
+.match {$num :integer} {$num :number minimumFractionDigits=2}
+* * {{Which selector formats {$num}?}}
+
+.match {$num :number minimumFractionDigits=2} {$num :integer}
+* * {{Which selector formats {$num}?}}
+```
```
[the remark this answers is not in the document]
aphillips: I added a "Do nothing" option that hopefully illustrates

exploration/selection-declaration.md
```diff
@@ -0,0 +1,296 @@
...
+		<dd>#000</dd>
+	</dl>
+</details>
+
+## Objective
+
+_What is this proposal trying to achieve?_
+
+Define what effect (if any) the _annotation_ of a _selector_ has on subsequent _placeholders_
+that access the same _operand_ value.
```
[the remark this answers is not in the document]
aphillips: Variable is good.

exploration/selection-declaration.md
aphillips: I re-titled this option. In this option, we do not have the `.local`-like feature. These selectors behave exactly like `.input` declarations. A `.input` is a declaration too and doesn't use an `=` sign.

exploration/selection-declaration.md
```diff
@@ -0,0 +1,296 @@
...
+  ```
+  .input {$date :datetime skeleton=yMdjm}
+  .match {$date :datetime field=month} {$date :datetime field=dayOfWeek}
+  * * {{This message produces a Duplicate Declaration error
+        even though selection is separate from formatting.}}
+  ```
+  However, this design does allow for a local variable to be easily created
+  for the purpose of selection.
+
+#### Allow "Declarative" Selectors with _Mutability_
```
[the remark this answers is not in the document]
aphillips: I re-titled, but I kept the words "mutability" and "immutability" because I think it's a lot clearer than talking about name shadowing. YMMV.

exploration/selection-declaration.md
aphillips: Nope. Only the first (well, now second after "do nothing") option provides a local variable binding (which alters the syntax!)

exploration/selection-declaration.md
```diff
@@ -0,0 +1,359 @@
...
+4. As a user (especially as a translator), I don't want to have to modify
+   declarations and selectors to keep them in sync.
+   ```
+   .input {$num :number minimumFractionDigits=1}
+   .match {$num}
+   * {{Shouldn't have to modify the selector}}
+   ```
+   > Note that this is currently provided for by the spec.
+
+5. As a user, I want to write multiple selectors for the same operand.
```
gibson042, line 106:
```suggestion
5. As a user, I don't want to write multiple selectors for the same operand.
```

aphillips: @gibson042 I don't understand the change proposed? This use case is illustrating a user who _does_ want to select on the same variable using different criteria.

aphillips:
Are we waiting on something to merge this **design** document? The meeting notes for 2024-04-29 (which I missed and was chaired by @eemeli) seem to indicate that @mihnita has a comment, but no other objections.

eemeli (approved):
I think this is fine to land as "proposed". We should also make our in-flight design docs more prominent and separate from the accepted ones, as I tried to do in #616.

PR #758, opened 2024-04-03
Change format of test/data-model-errors.json file

catamorphism:
While modifying ICU4C tests to read JSON directly from the test files in this repo, I noticed that the data-model-errors.json file I'd previously contributed is not actually valid JSON. 

Added braces to make it valid, and updated README file accordingly.

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -5,9 +5,8 @@ These test files are intended to be useful for testing multiple different messag
 
 - `syntax-errors.json` — An array of strings that should produce a Syntax Error when parsed.
 
-- `data-model-errors.json` - An array of objects.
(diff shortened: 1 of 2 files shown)


eemeli (changes requested):
Why not do the smaller fix of changing the JSON outermost `[]` into `{}` so that it matches the README?

catamorphism:
> Why not do the smaller fix of changing the JSON outermost `[]` into `{}` so that it matches the README?

You're right, that's simpler. Done in [85b77f0](https://github.com/unicode-org/message-format-wg/pull/758/commits/85b77f01e6f0805dd0c5238156d7147970203749)

eemeli (approved):

aphillips (approved):

PR #759, opened 2024-04-04
Match date/time formatter output to ICU formatting

catamorphism:
In trying to integrate the spec tests directly into the ICU4C tests (without modifying the JSON files), I noticed an inconsistency between the `time` and `datetime` functions' output as specified in the test, and what ICU outputs. (I'd noticed this before, but as I was importing the tests manually, I changed the expected output manually.)

ICU puts a narrow non-breaking space (0x202F) before the "AM" or "PM" in formatted date/time strings, which is a requirement from CLDR. I did a quick test that JS (at least running in Node) does the same thing, and it does:

```
$ node
Welcome to Node.js v18.13.0.
Type ".help" for more information.
> s = (Intl.DateTimeFormat("en-us", { timeStyle: "long" }).format(Date.now()))
'5:28:32 PM PDT'
> s.charCodeAt(7).toString(16)
'202f'
```

This change makes the expected output for time and datetime match what ICU outputs in these cases.

### DIFF
--- a/test/test-functions.json
+++ b/test/test-functions.json
@@ -25,18 +25,18 @@
       "exp": "{|horse|}",
       "errors": [{ "type": "bad-input" }]
     },
-    { "src": "{|2006-01-02T15:04:06| :time}", "exp": "3:04 PM" },
+    { "src": "{|2006-01-02T15:04:06| :time}", "exp": "3:04\u202FPM" },
     {
       "src": "{|2006-01-02T15:04:06| :time style=medium}",
-      "exp": "3:04:06 PM"
+      "exp": "3:04:06\u202FPM"
     },
     {
       "src": ".local $t = {|2006-01-02T15:04:06| :time style=medium} {{{$t :time}}}",
-      "exp": "3:04:06 PM"
+      "exp": "3:04:06\u202FPM"
     },
     {
       "src": ".local $d = {|2006-01-02T15:04:06| :date} {{{$d :time}}}",
-      "exp": "3:04 PM"
+      "exp": "3:04\u202FPM"
     }
   ],
   "datetime": [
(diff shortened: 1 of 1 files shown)


catamorphism:
A discussion could be had about what these tests should test for, since the spec doesn't specify exactly how dates/times are formatted. But if the tests are going to be used for checking implementations' conformance, and the expected output of test cases is a single hard-coded string, I think this is the right change. Maybe in the longer term the expected output should be specified differently. The `exp` strings could be replaced with regexes (at least to treat all whitespace characters the same way). 

What I want to use the tests for is to make sure my implementation is handling all the options to `:date`/`:time`/`:datetime` correctly, so it's useful to me if the output matches what CLDR requires, so that I don't have to use a modified version of the spec .json files or something nasty like that.

aphillips:
I think all of these tests need to be refactored--and not to contain any date/time formats at all. The tests for built-in functions should test that there is a required function, that it accepts the required literals (implementation defined types are up to the implementation), that it accepts the required options with the values specified by the spec, and that it emits the errors required.

The exception to this might be the string formatter.

If we want to test actual formatting, that belongs in ICU or maybe CLDR.

eemeli (changes requested):
This specific change is in fact a rather good example of why we should _not_ try to match the specific output of ICU. The change in the space character separating the AM/PM marker was made in CLDR 42, and prompted the following browser patches:
- https://phabricator.services.mozilla.com/D160188
- https://bugs.chromium.org/p/v8/issues/detail?id=13494
- https://github.com/WebKit/WebKit/pull/6556

If you test your code with Node.js 20 or later, you should find that it'll use a plain space rather than `\u202f`.

catamorphism:
So should we just remove the expected output for the tests in `test-functions.json` (except for `:string`), so as to specify that these strings should be formatted without errors, but the spec tests don't require a particular output?

aphillips:
(as contributor)

Yes, that's my opinion. Eventually a bit more than that. But not create tests that depend on the implementation or on specific CLDR data in this particular test suite.

catamorphism:
Closing in favor of #760

PR #760, opened 2024-04-05
Remove expected output from date/time formatter tests

catamorphism:
Per discussion on #759 , this PR removes expected output for `:date`, `:time`, and `:datetime`, except for error cases (since fallback handling is part of the spec).

Also added a note in the README to explain that.

I didn't remove the expected output for `:number`/`:integer` test cases, since that seems less likely to vary? But perhaps that should be done as well.

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -19,11 +19,15 @@ These test files are intended to be useful for testing multiple different messag
   - `parts: object[]` — The expected result of formatting the message to parts.
   - `cleanSrc: string` — A normalixed form of `src`, for testing stringifiers.
   - `errors: { type: string }[]` — The runtime errors expected to be emitted when formatting the message.
+     If `errors` is either absent or empty, the message must be formatted without errors.
   - `only: boolean` — Normally not set. A flag to use during development to only run one or more specific tests.
 
 - `test-function.json` — An object with string keys and arrays of test cases as values,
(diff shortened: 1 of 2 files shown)


eemeli (approved):
Could include a Prettier formatting run of the changed files, to prevent them being reformatted later.

aphillips (approved):
Looks good. Might be nice to convert the `exp` that were removed to comments, though.

macchiati:
One thought. Could retain the data, and document in the header that
conformance doesn't require exactly the same value. Non-cldr
implementations can test simply that a value is produced.

aphillips:
> One thought. Could retain the data, and document in the header that
> conformance doesn't require exactly the same value. Non-cldr
> implementations can test simply that a value is produced.

CLDR ones too. Date/time formats evolve over time; a good set of tests probably encompasses many locales and the full option set.

There should probably be some text about testing that says, effectively, "insert your platform integration tests here"

macchiati:
I agree that conformance needs to have tests that are very flexible. But we
can also generate version-specific test data for those implementations
using CLDR data. Even if the implementation patches CLDR data, those can be
pretty valuable to have, because they can often still use the tests with
some adjustments, because they know the patches that they make to CLDR
data. (I can elaborate more later.)

Anyway, not at all a blocker for tech review, of course.

PR #762, opened 2024-04-07
Remove registry.dtd and registry.xml from Tech Preview

aphillips:
Add comments to make clear that the DTD and XML forms of the default registry are not part of the tech preview, but that comments are welcome.

Note that we still need to do #596

### DIFF
--- a/spec/registry.dtd
+++ b/spec/registry.dtd
@@ -1,3 +1,7 @@
+<!-- 
+     This DTD is not part of the LDML45 Tech Preview of MessageFormat 2
+     Comments on this DTD are welcome. 
+-->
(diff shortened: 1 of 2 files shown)


spec/registry.dtd
```diff
@@ -1,3 +1,7 @@
+<!-- 
+     This DTD is not part of the LDML45 Tech Preview of MessageFormat 2
```
eemeli, line 2:
```suggestion
     This DTD is not part of the LDML45 Tech Preview of MessageFormat 2.
```

eemeli (approved):

mihnita (approved):
Thanks!

PR #765, opened 2024-04-10
Linkify the specification

aphillips:
This wasn't done in time for the LDML45 release, but capturing it publicly for later implementation. Can be added to the `tr35-messageformat.html` file.

### DIFF
--- a/docs/tools/linkify.js
+++ b/docs/tools/linkify.js
@@ -3,63 +3,63 @@
 // this has been tested on the tr35-messageformat.html file
 // but not implemented in LDML45
 
-    const terms = new Array();
-    var missing = new Set();
-    
-    function findTerms() {
-       document.querySelectorAll('dfn').forEach(generateAnchor);
-       // console.log(terms.length);
-   }
-    
-    function generateAnchor(item, index) {
-        // console.log(index + ": " + item.textContent);
-        let t = generateId(item.textContent);
-        terms.push(t);
-        item.setAttribute('id', t);
-    }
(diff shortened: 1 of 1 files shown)


srl295 (approved):
LGTM.  I think I can do this with a setting on the linkifier already being used

srl295:
> LGTM. I think I can do this with a setting on the linkifier already being used

https://github.com/unicode-org/cldr/blob/534a9278894440befa214917c6a413475d237632/tools/scripts/tr-archive/archive.js#L145

I think adding 'dfn' to that should do what it seems like you're trying to do. 
 since you are setup with the tools locally  you could PR that into the repo if it works locally

srl295:
see https://github.com/unicode-org/cldr/pull/3629

eemeli (changes requested):
This needs formatting by Prettier. With markdown I could be persuaded that sticking to it was too much of a hassle, but this code... is another matter.

Also, while you do get extra bonus brownie points for using `let` _and_ `const` _and_ `var`, you really shouldn't.

I'll look at this in more details once it's formatted.

aphillips:
@srl295 Adding 'dfn' to the archive.js does work to create anchors. I will PR that into the repo anon.

Turning `em` tags into links if-and-only-if there is an anchor would still need to be added.

eemeli (approved):
This should work? It's a bit messy and stylistically questionable, but it's probably fine. Presuming that there's calling code somewhere.

PR #767, opened 2024-04-12
Add new test schema

mradbourne:
This PR adds a schema for the JSON test files and refactors the existing test files to match the schema.

The schema is designed for easy interoperability with the `unicode/conformance` test suite. This is not a mandatory requirement but a nice-to-have.

Most of the content removed from the README is now included within `schema.json`.

The format of `params` has changed. They now use a `{ type, name, value }` structure instead of `{ name: value }`. Although instances of `{ "type": "string", "name": "foo", "value": "string-value" }` may seem unnecessary, this pattern allows us to specify params that are not part of the JSON spec (e.g. `{ "type": "datetime", "name": "foo", "value": "2006-01-02T15:04:06" }`.  

The TypeScript `.d.ts` files have been removed for now because they no longer match the JSON content. I'd be happy to re-add them if/when needed.

There is a Visual Studio Code settings file included. Although contributors use many editors and IDEs, I thought it would be useful to provide at least one editor/schema integration out-of-the-box.


test/schema.json
```diff
@@ -0,0 +1,475 @@
...
+      "type": "object",
+      "additionalProperties": {
+        "type": [
+          "string",
+          "number"
+        ]
+      }
+    },
+    "expErrors": {
+      "description": "The runtime errors expected to be emitted when formatting the message. If errors is either absent or empty, the message must be formatted without errors.",
```
eemeli, line 423:
```suggestion
      "description": "The runtime errors expected to be emitted when formatting the message. If expErrors is either absent or empty, the message must be formatted without errors.",
```

test/schema.json
```diff
@@ -0,0 +1,475 @@
...
+        "only": {
+          "type": "boolean",
+          "description": "Normally not set. A flag to use during development to only run one or more specific tests."
+        }
+      }
+    },
+    "description": {
+      "type": "string",
+      "description": "Information about the test."
+    },
```
eemeli, line 202: Should inline this under the `"test"` definition.

test/schema.json
```diff
@@ -0,0 +1,475 @@
...
+        "$ref": "#/$defs/var"
+      }
+    },
+    "var": {
+      "type": "object",
+      "additionalProperties": false,
+      "required": [
+        "type",
+        "value"
+      ],
```
eemeli, line 231: The type should be optional, as JSON already provides an implicit type that covers all the cases except for `"datetime"`.

This change will also need to be reflected in the `oneOf` array below.

```suggestion
      "required": [
        "value"
      ],
```

test/schema.json
```diff
@@ -0,0 +1,475 @@
...
+          "type"
+        ],
+        "properties": {
+          "type": {
+            "enum": [
+              "markup",
+              "literal",
+              "fallback"
+            ]
+          },
```
eemeli, line 323: The parts can have much more varied shapes. See [here](https://github.com/messageformat/messageformat/blob/main/packages/mf2-messageformat/src/formatted-parts.ts) for a TypeScript definition.

test/schema.json
```diff
@@ -0,0 +1,475 @@
...
+    },
+    "tests": {
+      "type": "array",
+      "items": {
+        "$ref": "#/$defs/test"
+      },
+      "minItems": 1
+    }
+  },
+  "$comment": "This allOf specifies required test properties that allow a default. A value will be required in 'defaultTestParams' if one is not provided for every individual test.",
```
eemeli, line 31: I like this. Had not come across this pattern before.

eemeli (changes requested):
Looks pretty good! Some minor-ish concerns mentioned inline.

We should also retain a human-friendly description of the test files in the README.

mihnita:
Some comments about the overall structure (we can fine tune later):

**1. I found the ability to group tests together in the same file pretty handy before (the way `data-model-errors.json` and `test-functions.json` used to be**

So instead of one hard-coded "tests" array, can we make that a map, where the key can be anything, not just "test"?

And with "defaultTestParams" at group level.

---

**2. In my implementation of ICU4J MF2, when trying to use the existing .json files, I found that some source strings are extremely long, unreadable.**

It would be handy (and something I did in ICU) to allow an array of source(s) as an alternative to `"src"`

Compare:

```json
"src": ".input {$exp :datetime timeStyle=short}\n.input {$user :string}\n.local $longExp = {$exp :datetime dateStyle=long}\n.local $zooExp = {$exp :datetime dateStyle=short timeStyle=$tsOver}\n{{Hello John, you want '{$exp}', '{$longExp}', or '{$zooExp}' or even '{$exp :datetime dateStyle=full}'?}}",
```
to:
```json
"srcs": [
    ".input {$exp :datetime timeStyle=short}\n",
    ".input {$user :string}\n",
    ".local $longExp = {$exp :datetime dateStyle=long}\n",
    ".local $zooExp = {$exp :datetime dateStyle=short timeStyle=$tsOver}\n",
    "{{Hello John, you want '{$exp}', '{$longExp}', or '{$zooExp}' or even '{$exp :datetime dateStyle=full}'?}}"
],
```

---

**3. When testing the results would be good to be able to specify expected results in a more flexible way**

That is to accommodate for different results between implementations, while still testing that the "plumbing" works and that we really format to a date / time / whatever.

Things like "one of":  `"expOneOf": [ "9:30 PM", "9:30\u202FPM" ]
Or reg-exp: :  `"expRegExp": [ "9:30.*[Pp].*[Mm]"]

---

**4. Allowing for an optional `"comment"` or `"description"` in each test**

test/tests/functions/ordinal.json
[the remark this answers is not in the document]
mradbourne: OK. I'll drop them for now because they're in version control history and we can discuss further if/how to include these.

test/tests/core.json
```diff
@@ -0,0 +1,415 @@
+{
+  "scenario": "Core syntax",
```
aphillips, line 2: consider naming this file `syntax.json` and avoiding the term "core". Are there "non-core" syntax elements?

test/schema.json
```diff
@@ -0,0 +1,475 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "title": "MessageFormat 2 data-driven tests source",
+  "description": "This documents the format of MessageFormat 2 source test data.",
+  "type": "object",
+  "additionalProperties": false,
```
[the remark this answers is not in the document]
mradbourne: Just summarising group discussion from 5 Apr meeting:
We've chosen to keep the schema more strict in an effort to enforce greater consistency in the test files.
I'll keep this property `false` but will explicitly version the schema. The version number will indicate any breaking changes in future.

test/tests/core.json
mradbourne: Sounds good 👍  I think we're unlikely to have "non-core".

test/schema.json
mradbourne: Thanks - I'll update the schema accordingly

mradbourne:
@mihnita Thanks for the comments! I'm replying here to capture our conversation from the WG meeting:

> **1. I found the ability to group tests together in the same file pretty handy before (the way `data-model-errors.json` and `test-functions.json` used to be**

As this is a non-breaking change, we'll keep it in mind as we write tests and propose it in a separate PR if needed.

My only concern was that having multiple `defaultTestParams` properties might make the test inputs more difficult to establish at a glance. The trade-off is less flexibility in the test file though. Lets revisit with some example tests that need that flexibility. I've added a `description` property to each test that might help with grouping (albeit not nesting) related tests.
 
> **2. In my implementation of ICU4J MF2, when trying to use the existing .json files, I found that some source strings are extremely long, unreadable.**
> 
> It would be handy (and something I did in ICU) to allow an array of source(s) as an alternative to `"src"`

Done 👍

> **3. When testing the results would be good to be able to specify expected results in a more flexible way**

As with point 1, we can discuss this non-breaking change up later. The first set of spec tests should be nice and deterministic. The `oneOf` would be for tests where results vary by implementation.

> **4. Allowing for an optional `"comment"` or `"description"` in each test**

Done 👍 . Thanks!

test/schemas/tests-schema-0.0.1.json
```diff
@@ -0,0 +1,445 @@
...
+        }
+      }
+    },
+    "testType": {
+      "description": "The category of test.",
+      "enum": [
+        "syntax",
+        "function"
+      ]
+    },
```
eemeli, line 206: What's this used for?

test/schemas/tests-schema-0.0.1.json
```diff
@@ -0,0 +1,445 @@
...
+      }
+    },
+    "var": {
+      "type": "object",
+      "oneOf": [
+        {
+          "additionalProperties": false,
+          "required": [
+            "value"
+          ],
```
eemeli, line 236: Shouldn't `"name"` also be required? Same goes for the other `"var"` variant.

eemeli (commented):
Did another review pass, a couple more questions inline.

Additionally, I think we need a way to mark inputs that are:
1. Not well formed, i.e. syntax errors.
2. Not valid, i.e. data model errors.

Maybe a new `expInvalid: 'syntax' | 'datamodel'` option, or some special values for `expErrors`?

test/schemas/tests-schema-0.0.1.json
aphillips: This comment and the ones @eemeli makes above suggest that we'd be well-served by a design document.

test/schemas/v0-0-1/tests.schema.json
```diff
@@ -0,0 +1,334 @@
...
+            }
+          }
+        },
+        {
+          "properties": {
+            "tests": {
+              "type": "array",
+              "items": {
+                "required": [
+                  "src"
```
mihnita, line 100: Having `"src"` be either a string or an array breaks some json parsers used by strict type languages.

For example in Java one can use parsers with annotations, or reflection, or other methods to automatically parse the JSON into a class.

Using GSon with this simple class "just works", without any extra code:
```java
class Unit {
    final String src;
    final List<String> srcs;
    final String locale;
    final Map<String, Object> params;
    final String exp;
    final String ignore;
    final List<Error> errors;
```

GSon will automatically map a JSON like this to that class:
```json
{
   "src": "Hello {$user}!",
   "locale": "en-US",
   "params": { "user", "John" },
   "exp": "Hello John!"
}
```

If the schema allows for a field to be an either / or it breaks these kinds of parsers.
All of a sudden you get an `Object` or `JsonObject` back and you must do type checking and explicit handling of the various types.
From all I know these parsers are available (and common) for Java, C++ and Rust, probably others.

test/schemas/v0-0-1/formatted-parts.schema.json
```diff
@@ -0,0 +1,206 @@
...
+      "properties": {
+        "type": {
+          "enum": [
+            "datetime",
+            "fallback",
+            "number",
+            "string",
+            "unknown"
+          ]
+        }
```
eemeli, line 78: This is still too strict, and unnecessary. We do not need to make this parts schema require extension every time a new function is added. All of the subtypes of this should be dropped, leaving `messageExpressionPart` to be as freeform as it really is.

```suggestion
        "type": "string"
```

test/schemas/v0-0-1/tests.schema.json
aphillips: This is one sort of problem users will experience in working with MF2 messages. Messages have to be stored in file and resource formats, typically as a string. We should ensure this dog food is yummy. 🐶🥫😋

test/schemas/tests-schema-0.0.1.json
mradbourne: @eemeli This was a way to indicate that different options/args are passed to the test executor for different types of test, solving a similar problem to https://github.com/unicode-org/message-format-wg/pull/769, where we need special behaviour for function tests.

For example, a value of `"testType": "function"` could automatically mock out instances of `:number` in the pattern to achieve similar behaviour to using `:test:number`. However, if #769 lands, I don't think there is a need for `testType`. I'll remove it now on that basis.

mradbourne: @aphillips No problem - I'm planning to update https://github.com/unicode-org/message-format-wg/blob/main/exploration/data-driven-tests.md to document the outcomes from this PR when we reach consensus, if that works?

test/schemas/v0-0-1/tests.schema.json
```diff
@@ -0,0 +1,334 @@
...
+      "items": {
+        "type": "object",
+        "additionalProperties": false,
+        "required": [
+          "type"
+        ],
+        "properties": {
+          "type": {
+            "enum": [
+              "bad-input",
```
janishorsts, line 296: My comment is outside the PR scope.

How do these errors correlate with errors described in [MessageFormat 2.0 Errors](https://github.com/unicode-org/message-format-wg/blob/main/spec/errors.md)?

RangeError (range-error) is bound to EcmaScript implementation.

test/schemas/v0-0-1/formatted-parts.schema.json
mradbourne: Sounds good. Updated.

test/schemas/tests-schema-0.0.1.json
mradbourne: This was originally done to allow `params` to take positional args. e.g.
```
"params": [
  { "value": "first" },
  { "value": "second" },
]
```
I don't think we need to do this though - `params` can just be an `object`.

mradbourne: Yes, you're right.
`name` was originally optional to support positional args. I don't think we need this though.

test/schemas/v0-0-1/tests.schema.json
```diff
@@ -276,7 +279,99 @@
...
+              },
+              "options": {
+                "type": "object"
+              }
+            }
+          },
+          {
+            "description": "Message expression part.",
+            "type": "object",
+            "additionalProperties": "false",
```
eemeli, line 337: Expression parts can have additional fields.
```suggestion
```

test/schemas/v0-0-1/tests.schema.json
```diff
@@ -276,7 +279,99 @@
...
+                    },
+                    "value": {}
+                  },
+                  "required": [
+                    "type"
+                  ]
+                }
+              },
+              "value": {}
+            }
```
eemeli, line 371: At most one of `"parts"` and `"value"` should be set, so I think something like this is needed:
```suggestion
            },
            "oneOf": [
              {
                "not": { "required": ["parts", "value"] }
              },
              {
                "required": ["parts"],
                "not": { "required": ["value"] }
              },
              {
                "required": ["value"],
                "not": { "required": ["parts"] }
              },
            ]
```

test/schemas/v0-0-1/tests.schema.json
mradbourne: It seems like there's more deliberation to be had about the 'src' vs 'srcs' list. Since we're not using the list version yet, I'll leave it out of the schema for now. We can introduce it as an enhancement when needed and have more in-depth discussion then. Thanks for flagging, Mihai.

test/schemas/v0-0-1/tests.schema.json
mradbourne: Done 👍

test/tests/functions/number.json
```diff
@@ -0,0 +1,409 @@
...
+        }
+      ]
+    },
+    {
+      "src": "{42 :number @foo @bar=13}",
+      "exp": "42",
+      "expParts": [
+        {
+          "type": "literal",
+          "value": "42"
```
mradbourne, line 404: The original expParts was:
`[{ "type": "number", "parts": [{ "type": "integer", "value": "42" }] }]`
Since the part doesn't have the `source` property that a message _expression_ part requires, I've assumed we want this as a string literal part.

echeran:
The PR looks good. I want to add automated guarantees that make use of this PR's schema to ensure that our test files are well-formed JSON and adhere to the schema. So I created https://github.com/mradbourne/message-format-wg/pull/1 

That PR is in @mradbourne 's personal fork with a destination of this PR. PTAL before merging.

test/schemas/v0-0-1/tests.schema.json
```diff
@@ -0,0 +1,390 @@
...
+            "required": [
+              "type",
+              "source"
+            ],
+            "not": {
+              "required": [
+                "parts",
+                "value"
+              ]
+            },
```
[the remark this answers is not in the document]
eemeli: This is asserting that `parts` and `value` are not both defined at the same time.

eemeli (approved):

eemeli:
I think I'd prefer a schema directory like `test/schemas/v0/` (with integer versions) or `test/schemas/v0.0.1/` instead of `test/schemas/v0-0-1/`, but that's a pretty minor quibble and can be addressed later.

mradbourne:
@echeran Thanks for raising https://github.com/mradbourne/message-format-wg/pull/1 
I'd like to do this as a separate PR to `unicode-org/message-format-wg` as it raises some new questions/decisions.

PR #769, opened 2024-04-15
Add :test:function, :test:select, and :test:format function definitions

eemeli:
CC @mradbourne

Given that we shouldn't rely on `:number` and `:datetime` producing exact results but we still want to ensure that our interfaces with them work, let's add `:test:number` and `:test:plural` as well-defined functions intended for tests only, with exactly defined behaviour.

This PR could be extended with test cases making use of these, but I figured it'd be more useful to consider the general case first, esp. in order to avoid conflicting with #767.


test/README.md
```diff
@@ -35,3 +35,78 @@ Some examples of test harnesses using these tests, from the source repository:
...
+
+As the behaviour of `:number`, `:datetime`, and other default message functions
+is dependent on locale-specific data and may vary between implementations,
+the following functions are defined for test use only:
+
+### `test:number`
+
+This function is valid both as a selector and as a formatter.
+The only option it recognizes is `fractionDigits`,
+all other option values passed to it are ignored.
```
aphillips, line 49: How would I use this?

A test like this still doesn't work:

```
You have {$num :number} elderberries.
```

I have to rewrite it to use `:test:number` and `:test:number` has different options from `:number`. Implementations cannot test full conformance with MF2 using the tests because default registry is included in full conformance.

Maybe a better approach would be to have an invariant locale or a test locale and to define what the outputs for that locale are. e.g. we could use the root locale (`und`) or possibly a patch on that (`und-x-test`) [note that pseudo-translation locales recommend themselves here as well--especially the potential for RTL pseudo-locales so both LTR and RTL messages can be tested]

Possible we should consider doing design first?

test/README.md
```diff
@@ -35,3 +35,90 @@ Some examples of test harnesses using these tests, from the source repository:
...
 - [Data model stringify tests](https://github.com/messageformat/messageformat/blob/11c95dab2b25db8454e49ff4daadb817e1d5b770/packages/mf2-messageformat/src/data-model/stringify.test.ts)
 - [Formatting tests](https://github.com/messageformat/messageformat/blob/11c95dab2b25db8454e49ff4daadb817e1d5b770/packages/mf2-messageformat/src/messageformat.test.ts)
+
+## Test Functions
+
+As the behaviour of `:number`, `:datetime`, and other default message functions
+is dependent on locale-specific data and may vary between implementations,
+the following functions are defined for **test use only**:
+
+### `test:func`
```
aphillips, line 45: `select` and `format` are fine, but `func` should probably be `function`

```suggestion
### `:test:function`
```

test/README.md
```diff
@@ -35,3 +35,90 @@ Some examples of test harnesses using these tests, from the source repository:
...
+
+This function is valid both as a selector and as a formatter.
+The only option it recognizes is `fd`,
+all other option values passed to it are ignored.
+
+When resolving a `test:func` expression,
+its _Input_ and _FD_ values are determined as follows:
+
+1. Let _FD_ be `0`.
+1. Let _arg_ be the resolved value of the expression argument.
```
aphillips, line 55: "expression argument"? Is that an "operand"?

test/README.md
```diff
@@ -35,3 +35,90 @@ Some examples of test harnesses using these tests, from the source repository:
 - [CST parse/stringify tests](https://github.com/messageformat/messageformat/blob/11c95dab2b25db8454e49ff4daadb817e1d5b770/packages/mf2-messageformat/src/cst/cst.test.ts)
 - [Data model stringify tests](https://github.com/messageformat/messageformat/blob/11c95dab2b25db8454e49ff4daadb817e1d5b770/packages/mf2-messageformat/src/data-model/stringify.test.ts)
 - [Formatting tests](https://github.com/messageformat/messageformat/blob/11c95dab2b25db8454e49ff4daadb817e1d5b770/packages/mf2-messageformat/src/messageformat.test.ts)
+
+## Test Functions
+
+As the behaviour of `:number`, `:datetime`, and other default message functions
```
aphillips, line 41: There's a specific list. I think `:string` shouldn't be on that list?

```suggestion
As the behavior of the functions 
`:number`,
`:integer`,
`:datetime`,
`:date`,
and `:time`
in the default registry depend 
on specific versions of locale data and on their local implementation
the following functions are defined for **test use only**.
```

test/README.md
```diff
@@ -35,3 +35,90 @@ Some examples of test harnesses using these tests, from the source repository:
...
+is dependent on locale-specific data and may vary between implementations,
+the following functions are defined for **test use only**:
+
+### `test:func`
+
+This function is valid both as a selector and as a formatter.
+The only option it recognizes is `fd`,
```
aphillips, line 48: What does `fd` mean?

We should probably follow the model in the registry.md document in defining these functions. There we have subsections

- operands
- options
- selection
- formatting
- (other stuff)

I think a little bit of design is probably useful here as well. It would be nice to be able to have `:function` (and `:select`) model `:number`'s behavior (with both exact and keyword matching) by defining what numbers match the plural keywords `zero`, `one`, `two`, `few`, `many`, and `other`. That way we can provide tests that implementers can use to see if they match the sorting mechanism in `formatting.md` without writing number/plural selection over.

Another thing to consider is that we don't need options if we can specify how `:function` and `:select` work internally. One thing we could specify is that they'll receive an ordered list or other set of arguments from the test (not in the message) that can be used to program their behavior in a simplistic way, e.g. a regexp that should match the operand.

test/README.md
eemeli: Rewrote it a bit, still leaving the list open so that adding a new default registry function doesn't require updating this as well.

test/README.md
eemeli: > What does `fd` mean?

It started life as `fractionDigits`, but you asked for this to be distanced from the real `:number` implementation, and I simplified its values to just `0` and `1`, so it makes sense to be a bit more abstract. So now it's just an option that has _some_ effect on the `:test:function` behaviour, and a small set of valid values.

> We should probably follow the model in the registry.md document in defining these functions. There we have subsections

That model doesn't exactly work here, because here we need to be explicit about exactly what this function does, and how it behaves in different settings.

> It would be nice to be able to have `:function` (and `:select`) model `:number`'s behavior (with both exact and keyword matching) by defining what numbers match the plural keywords `zero`, `one`, `two`, `few`, `many`, and `other`. That way we can provide tests that implementers can use to see if they match the sorting mechanism in `formatting.md` without writing number/plural selection over.

That's already supported. Selection on `{1 :test:function}` will only match the `1` key, while selection on `{1 :test:function fd=1}` will match both `1` and `1.0` keys, preferring the latter. This is explicitly different from `:number` selection, again, because you asked for this to be different.

> Another thing to consider is that we don't need options if we can specify how `:function` and `:select` work internally. One thing we could specify is that they'll receive an ordered list or other set of arguments from the test (not in the message) that can be used to program their behavior in a simplistic way, e.g. a regexp that should match the operand.

I would prefer working iteratively, and first seeing where this small family of functions gets us. If/as we discover parts of the spec that do need testing but can't be tested with these functions, then we can reconsider.

I would very much prefer not needing to specify even simple programmable behaviour at the level needed to guarantee exact behaviour across all implementations. For this, regexp is pretty much right out unless there's a ready reference somewhere for a minimum subset that's supported everywhere _and_ for which a validator is available to keep the test data honest.

aphillips: > but you asked for this to be distanced from the real :number implementation

Right. My thinking was that we want something that can be used for conformance testing now and later. We want it to be easy to implement, have deterministic behavior, and cover selection/formatting features required in our spec.

We have [an example](https://github.com/unicode-org/message-format-wg/blob/main/spec/formatting.md#example-3) in our spec now that should be testable. The example uses `:number`, but replacing it with `:function` could be straightforward.

> That model doesn't exactly work here, because here we need to be explicit about exactly what this function does, and how it behaves in different settings.

That model can work here, because we can describe the operands and options and the proceed to describe specifically how selection works and how formatting works. You're correct that, unlike `:number` or `:datetime`, we need to provide very specific implementation instructions.

Bear in mind that we need to flesh out how to write proposals for the default and "RGI" registry. The test functions should probably go through the motions and use the forms of that process.

> I would very much prefer not needing to specify even simple programmable behaviour at the level needed to guarantee exact behaviour across all implementations. For this, regexp is pretty much right out unless there's a ready reference somewhere for a minimum subset that's supported everywhere and for which a validator is available to keep the test data honest.

Fair enough.

eemeli: > We have [an example](https://github.com/unicode-org/message-format-wg/blob/main/spec/formatting.md#example-3) in our spec now that should be testable. The example uses `:number`, but replacing it with `:function` could be straightforward.

Sure, that should be pretty directly portable to using `:test:function`. I'll be happy to do so after #767 lands, either within this PR or in a follow-up.

> Bear in mind that we need to flesh out how to write proposals for the default and "RGI" registry. The test functions should probably go through the motions and use the forms of that process.

So could we land this first with its current freeform description, and then update it once the forms of that process are indeed fleshed out? I would really prefer not blocking improvements in tests on figuring out a new proposal process.

aphillips: > So could we land this first with its current freeform description, and then update it once the forms of that process are indeed fleshed out? I would really prefer not blocking improvements in tests on figuring out a new proposal process.

Sure, we can land this. 

However, nothing we do here will effect v45. We have _lots_ to do in the next few months. Each thing we do should not require later additional work to the degree reasonable.

Note that I'm not asking you to wait on a proposal process, but to try and use the flabby outlines of what we have to see what we can learn towards creating said process. If it's too hard, we probably need to look in another direction. FWIW, it would be amusing to see the registry XML or JSON description for `:function`, `:select`, and `:format`

eemeli: > Note that I'm not asking you to wait on a proposal process, but to try and use the flabby outlines of what we have to see what we can learn towards creating said process.

Done.

test/README.md
```diff
@@ -23,3 +23,104 @@ Some examples of test harnesses using these tests, from the source repository:
...
+- If the _Input_ is `1` and _FD_ is `1`,
+  the method will return some slice of the list « `'1.0'`, `'1'` »,
+  depending on whether those values are included in `keys`.
+- If the _Input_ is `1` and _FD_ is `0`,
+  the method will return the list « `'1'` » if `keys` includes `'1'`, or an empty list otherwise.
+- If the _Input_ is any other value, the method will return an empty list.
+
+When an expression with a `:test:function` annotation is assigned to a variable by a declaration
+and that variable is used as an option value,
+its resolved value is the _Input_ value.
```
aphillips, line 90: ...line breaks and italics for terms...

test/README.md
```diff
@@ -23,3 +23,104 @@ Some examples of test harnesses using these tests, from the source repository:
...
+   formatted as a sequence of decimal digit caharacters.
+1. If _FD_ is `1`, then
+   1. The character `.` U+002E Full Stop.
+   1. The single decimal digit character representing the value floor((abs(_Input_) - floor(abs(_Input_))) \* 10)
+
+If the formatting target is a sequence of parts,
+each of the above parts will be emitted separately
+rather than being concatenated into a single string.
+
+Note that for purposes of clarity, the formatting of `:test:function` does not perform any rounding.
```
aphillips, line 107: This shouldn't be a note as we are describing the implementation.

```suggestion
`:test:function` does not round the formatted value.
Implementations MAY impose an upper limit on the number of digits rendered,
based on local requirements (such as the size of the underlying decimal type)
```

test/README.md
```diff
@@ -23,3 +23,104 @@ Some examples of test harnesses using these tests, from the source repository:
...
+and that variable is used as an option value,
+its resolved value is the _Input_ value.
+
+When `:test:function` is used as a formatter,
+a placeholder resolving to a value with a `:test:function` expression
+is formatted as a concatenation of the following parts:
+
+1. If _Input_ is less than `0`, the character `-` U+002D Hyphen-Minus.
+1. The truncated absolute integer value of _Input_, i.e. floor(abs(_Input_)),
+   formatted as a sequence of decimal digit caharacters.
```
aphillips, line 98: Typo

```suggestion
   formatted as a sequence of decimal digit characters (U+0030...U+0039).
```

test/README.md
```diff
@@ -20,6 +20,10 @@ Some examples of test harnesses using these tests, from the source repository:
 - [Data model stringify tests](https://github.com/messageformat/messageformat/blob/11c95dab2b25db8454e49ff4daadb817e1d5b770/packages/mf2-messageformat/src/data-model/stringify.test.ts)
 - [Formatting tests](https://github.com/messageformat/messageformat/blob/11c95dab2b25db8454e49ff4daadb817e1d5b770/packages/mf2-messageformat/src/messageformat.test.ts)
 
+A [JSON schema](./schemas/) is included for the test files in this repository.
+
+For users of Visual Studio Code, a [settings file](./.vscode/settings.json) is included that enables schema validation while editing the test files.
```
aphillips, line 25: Line breaks

eemeli: This was added by #767, after which I merged `main` to this branch.

test/README.md
eemeli: The proposed MAY is incorrect, as the formatted digits are exactly defined above. If this is unclear, let's just remove it?
```suggestion
```

test/README.md
eemeli: Introducing italicized terms here would make it more difficult to discern the local _Input_ reference, which I'd prefer to avoid. Can it be avoided, or is there some other styling I should Use for _Input_ and _FD_?

aphillips: Use `code` like `Input`, perhaps??

sffc (commented):
The proposed `:test:function` appears super minimal, which seems good to me. I agree this is useful for testing, giving a way to avoid locale-dependent output. I'll defer on the exact spec text and scope.

test/README.md
```diff
@@ -23,3 +23,103 @@ Some examples of test harnesses using these tests, from the source repository:
...
+If the formatting target is a sequence of parts,
+each of the above parts will be emitted separately
+rather than being concatenated into a single string.
+
+
+### `:test:select`
+
+This _function_ accepts the same _operands_ and _options_,
+and behaves exactly the same as `:test:function`,
+except that it cannot be used for formatting.
```
echeran, line 112: I think having a selection function that isn't capable of formatting is strange because formatting can be a precursor to selection, and that is indeed the case for plurals/numbers, and this test function takes numbers.

What is the motivation for having such a function for selection but not formatting?

One concern is that such a function implies that selection and formatting aren't linked. Another concern is that if the constraint that `:test:select` is defined by is not super useful, and if its functionality is a subset of the full `:test:function`, then we can remove `:test:select` altogether without loss of functionality.

test/README.md
```diff
@@ -23,3 +23,103 @@ Some examples of test harnesses using these tests, from the source repository:
...
+
+When `:test:select` is used as a _formatter_,
+a "not-formattable" error is emitted and the _placeholder_ is formatted with
+a _fallback value_.
+
+### `:test:format`
```
echeran, line 118: Similar comment as for `:test:select`: can we remove this as redundant given the superset functionality in `:test:function`?

test/README.md
```diff
@@ -23,3 +23,103 @@ Some examples of test harnesses using these tests, from the source repository:
...
+
+This function is valid both as a _selector_ and as a _formatter_.
+
+#### Operands
+
+The function `:test:function` requires a [Number Operand](/spec/registry.md#number-operands) as its _operand_.
+
+#### Options
+
+The only _option_ `:test:function` recognizes is `fd`,
```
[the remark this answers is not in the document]
eemeli: It started life as `fractionDigits`, but got shortened to `fd` corresponding with @aphillips requests to make this less like `:number`.

test/README.md
```diff
@@ -23,3 +23,103 @@ Some examples of test harnesses using these tests, from the source repository:
...
+a _digit size option_ for which only `0` and `1` are valid values.
```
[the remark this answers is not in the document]
eemeli: https://github.com/unicode-org/message-format-wg/blob/main/spec/registry.md#digit-size-options

test/README.md
eemeli: The intent is to allow testing the capability in MF2 of defining a function that's only a formatter, such as `:datetime` in the default registry. With `:test:format`, it becomes possible to validate the implementation's behaviour when such a function is used as a selector. That's not possible to test with `:test:function`.

echeran: Ok, that makes sense. Sounds good.

test/README.md
echeran: Ok, that makes sense. IIUC, the `:test:select` function isn't so much about number formatting as much as standing in for the category of functions that fit the bill of select-only, for which there exist regular examples like gender. Sounds good.

test/README.md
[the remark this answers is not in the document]
echeran: suggestion: rename `fd` to `ds`, and `FD` to `DS`,  so that it at least corresponds to "digit size"? 

@aphillips What say you?

test/README.md
echeran, line 45: Linkifying digit size just as "Number Operand" is being linkified 5 lines above. Helps short circuit questions like mine for the reader.

```suggestion
a [_digit size option_](/spec/registry.md#digit-size-options) for which only `0` and `1` are valid values.
```

test/README.md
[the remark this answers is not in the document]
aphillips: My concern with `fractionDigits` was to stay away from confusing users with slightly different names (the ones in `:number` are `maximumFractionDigits` and `minimumFractionDigits`). Obscure names don't help anyone. I'd suggest `decimalPlaces`, or perhaps `decimalSize`. Alternatively, we could test drive number skeletons here.

test/README.md
aphillips: Underscore italics *are* linkified when the spec is turned into HTML. That's what [this](https://github.com/unicode-org/message-format-wg/blob/main/docs/tools/linkify.js) is for.

test/README.md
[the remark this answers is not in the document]
echeran: `decimalPlaces` sgtm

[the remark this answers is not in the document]
eemeli: Done.

aphillips (approved):

mihnita (approved):

PR #774, opened 2024-04-23
Refactor errors, adding section for Message Function Errors

eemeli:
CC @mradbourne, who should be added to the repo.

As [discussed yesterday](https://github.com/unicode-org/message-format-wg/blob/main/meetings/2024/notes-2024-04-22.md#738), our error categories could be improved. This PR effects the categorisation we discussed, with "Message Function Error" introduced as the supertype of the errors that are coming from a custom function and "Invalid Expression" renamed to avoid its re-definition of what "valid" means in an MF2 context.

This and #767 should not be considered as blockers for each other. Instead, we'll need a fixup PR after both of these land adding the kebab-case error codes to `errors.md` and updating the ones used in the tests.


spec/errors.md
```diff
@@ -34,7 +34,7 @@ SHOULD prioritise _Syntax Errors_ and _Data Model Errors_ over others.
 
 When an error occurs within a _selector_,
 the _selector_ MUST NOT match any _variant_ _key_ other than the catch-all `*`
-and a _Resolution Error_ or a _Selection Error_ MUST be emitted.
+and a _Resolution Error_ or a _Message Function Error_ MUST be emitted.
```
aphillips, line 37: Selection Error is a sub-species of Message Function Error, right? Any reason to expand?

spec/errors.md
```diff
@@ -241,49 +241,72 @@ or for private implementation use that is not supported by the current implement
...
 whose implementation-defined internal requirements produce an error during _function resolution_
 or when a _function_ returns a value (such as `null`) that the implementation does not support.
 
-An **_<dfn>Operand Mismatch Error</dfn>_** is an _Invalid Expression_ error that occurs when
+> For example, the following _message_ might produce a _Bad Expression_ error if the
+> the function `:function` threw an exception or otherwise emitted an error
+> rather than returning a valid value:
+>
+> ```
+> This has a bad expression {$var :function} because it has a bug in it.
```
aphillips, line 269:
```suggestion
This message emits a Bad Expression when {$var :function} throws an error.
```

spec/errors.md
```diff
@@ -241,49 +241,72 @@ or for private implementation use that is not supported by the current implement
...
-### Unsupported Statement
+### Bad Option
 
-An **_<dfn>Unsupported Statement</dfn>_** error occurs when a message includes a _reserved statement_.
+A **_<dfn>Bad Option</dfn>_** error is an error that occurs when
+an _option_ value provided to a _function_ during _function resolution_ does not match one of the
+expected implementation-defined types for that function;
+or in which a literal _option_ value does not have the required format
+and thus cannot be processed into one of the expected implementation-defined types
+for that specific _function_.
```
aphillips, line 297: I think we should allow implementations to emit other sorts of bad option errors, not just the bad value.

```suggestion
A **_<dfn>Bad Option</dfn>_** error is an error that occurs when there is an implementation-defined error with an _option_ or _option_ _value_.
These might include:
- a required _option_ is missing
- mutually exclusive _options_ are supplied
- an _option_ _value_ provided to a _function_ during _function resolution_
   does not match one of the implementation-defined types 
   for that _function_;
   or in which the literal _option_ _value_ does not have the required
   format
   and thus cannot be processed into one of the expected
   implementation-defined types for that specific _function_
- any other implementation-defined errors related to _options_
```

spec/errors.md
```diff
@@ -241,49 +241,72 @@ or for private implementation use that is not supported by the current implement
...
 
-> For example, attempting to format this message
-> would always result in an _Unsupported Statement_ error:
+> For example, the following _message_ might produce a _Bad Option_ error
+> because the literal `foo` does not match the production `number-literal`,
+> which is a requirement of the function `:number` for its `minimumFractionDigits` _option_:
```
aphillips, line 301: Slightly wrong. `|foo|` doesn't match [digit size option](https://github.com/unicode-org/message-format-wg/blob/main/spec/registry.md#digit-size-options)

aphillips (changes requested):
Good start.

spec/errors.md
```diff
@@ -16,10 +16,10 @@ and MUST be emitted as soon as possible.
 The other error categories are only emitted during formatting,
 but it might be possible to detect them with validation tools.
 
-During selection, an _expression_ handler MUST only emit _Resolution Errors_ and _Selection Errors_.
-During formatting, an _expression_ handler MUST only emit _Resolution Errors_ and _Formatting Errors_.
+During selection and formatting,
+_expression_ handlers MUST only emit _Message Function Errors_.
```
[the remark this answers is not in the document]
eemeli: Yes; those occur earlier.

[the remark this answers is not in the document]
aphillips: PS> added @mradbourne to the repo.

spec/errors.md
eemeli: Leaving it as Message Function allows for e.g. a lazy implementation to only discover at this point that one of its options that are only relevant for selection is badly formed, and so emit a more useful Bad Option error instead.

spec/errors.md
```diff
@@ -241,49 +241,72 @@ or for private implementation use that is not supported by the current implement
...
+
+> For example, attempting to format this message
+> would always result in an _Unsupported Statement_ error:
+>
+> ```
+> .some {|horse|}
+> {{The message body}}
+> ```
+
+## Message Function Errors
```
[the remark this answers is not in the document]
eemeli: No user-controllable code gets called for markup, but otherwise that looks good.

spec/errors.md
```diff
@@ -241,49 +241,72 @@ or for private implementation use that is not supported by the current implement
...
+### Bad Expression
+
+A **_<dfn>Bad Expression</dfn>_** error occurs when a _message_ includes an _expression_
 whose implementation-defined internal requirements produce an error during _function resolution_
 or when a _function_ returns a value (such as `null`) that the implementation does not support.
 
-An **_<dfn>Operand Mismatch Error</dfn>_** is an _Invalid Expression_ error that occurs when
+> For example, the following _message_ might produce a _Bad Expression_ error if the
+> the function `:function` threw an exception or otherwise emitted an error
+> rather than returning a valid value:
```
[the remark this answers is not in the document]
eemeli: This is a remnant from the preceding; I'd be happy to remove the `>` part completely?

spec/errors.md
eemeli: Edited with stylistic fixes, including removing _value_ as a thing we define.

spec/errors.md
eemeli:
```suggestion
> because the literal `foo` does not match the production `digit-size-option`,
> which is a requirement of the function `:number` for its `minimumFractionDigits` _option_:
```

spec/errors.md
```diff
@@ -302,9 +325,9 @@ An **_<dfn>Unsupported Statement</dfn>_** error occurs when a message includes a
 > * {{The value is not one.}}
 > ```
 
-## Formatting Errors
+### Formatting Error
 
-**_<dfn>Formatting Errors</dfn>_** occur during the formatting of a resolved value,
+A **_<dfn>Formatting Error</dfn>_** occurs during the formatting of a resolved value,
 for example when encountering a value with an unsupported type
 or an internally inconsistent set of options.
```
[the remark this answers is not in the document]
eemeli: The errors aren't necessarily exclusive, so both are still possible.

spec/errors.md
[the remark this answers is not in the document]
aphillips: > No user-controllable code gets called for markup

We do say that:

> The resolution of markup MUST always succeed.

... so I guess throwing an error can't happen.

spec/formatting.md
```diff
@@ -307,13 +307,13 @@ the following steps are taken:
    resolve the value of the _expression_ as the result of that function call.
 
    If the call fails or does not return a valid value,
-   emit a _Invalid Expression_ error.
+   emit a _Message Function Error_.
```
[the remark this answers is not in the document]
eemeli: That seems a bit too opinionated? As in, an implementation could have a hierarchy of errors like Error > MessageFormatError > MessageFunctionError > BadOption, and have a need to use the MessageFunctionError class directly if user code throws a random error that's not a MessageFunctionError.

spec/formatting.md
```diff
@@ -519,7 +519,7 @@ First, resolve the values of each _selector_:
    1. Else:
       1. Let `nomatch` be a resolved value for which selection always fails.
       1. Append `nomatch` as the last element of the list `res`.
-      1. Emit a _Selection Error_.
+      1. Emit a _Message Function Error_.
```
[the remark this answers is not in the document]
eemeli:
```suggestion
      1. Emit a _Selection Error_.
```

spec/formatting.md
```diff
@@ -735,7 +735,7 @@ each _text_ and _placeholder_ part of the selected _pattern_ is resolved and for
 
 Resolved values cannot always be formatted by a given implementation.
 When such an error occurs during _formatting_,
-an implementation SHOULD emit a _Formatting Error_ and produce a
+an implementation SHOULD emit a _Message Function Error_ and produce a
```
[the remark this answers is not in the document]
eemeli: We'll be able to test for the more general error class as well. @mradbourne or I will file a follow-up PR to his current one to allow for that.

spec/errors.md
```diff
@@ -241,49 +241,81 @@ or for private implementation use that is not supported by the current implement
...
-An **_<dfn>Operand Mismatch Error</dfn>_** is an _Invalid Expression_ error that occurs when
+> For example, the following _message_ might produce a _Bad Expression_ error if the
+> the function `:function` threw an exception or otherwise emitted an error
+> rather than returning a valid value:
+>
+> ```
+> This has a bad expression {$var :function} because it has a bug in it.
+> ```
+
+### Bad Operand
```
eemeli, line 276:
```suggestion
### Bad Operand
```

spec/formatting.md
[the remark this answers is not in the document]
aphillips: ... in which case the "appropriate" error is the parent class of error. My suggestion permits the implementation to emit one of the subclasses of error and encourages the implementer to figure out which one it is. Generally speaking, if there's a random exception thrown, it'll be emitted as a FormattingError or SelectionError (indicating what the implementation was trying to do when it failed). This would even be helpful in debugging for functions that do both selection and formatting, since it would tell me which part of the function the error took place in.

spec/errors.md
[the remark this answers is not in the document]
aphillips: Although I agree, I find it weird that the two examples appear to be for other types of errors that we provide separately. They are supposed to illustrate what a formatting error is. We have an extensive example underneath already. We could just omit this text.

```suggestion
A **_<dfn>Formatting Error</dfn>_** occurs during the formatting of a resolved value.
```

spec/formatting.md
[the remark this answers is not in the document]
aphillips: Agreed that we should create a test for it. My comment, though, is that we should nail this one down somewhat. Or do you disagree?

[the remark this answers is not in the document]
eemeli: I disagree. I think leaving it up to the function implementer to pick between the options will provide the best user experience, and it allows for lazy implementations to e.g. only check operands and options until they're absolutely required, and to emit Bad Operand or Bad Option errors at any time.

spec/errors.md
```diff
@@ -241,49 +241,73 @@ or for private implementation use that is not supported by the current implement
...
-An **_<dfn>Invalid Expression</dfn>_** error occurs when a _message_ includes an _expression_
+## Message Function Errors
+
+A **_<dfn>Message Function Error</dfn>_** is any error that occurs
+when calling a message function implementation
+or which depend on validation associated with a specific function.
+
+### Bad Expression
+
+A **_<dfn>Bad Expression</dfn>_** error occurs when a _message_ includes an _expression_
```
catamorphism, line 264: Can you add an example here? I'm still not sure what an example would be of a "bad expression" error that isn't a "bad operand" error.

catamorphism (changes requested):

spec/errors.md
eemeli: The examples for this aren't quite so clearcut as for Bad Operand and Bad Option, but I can think of at least the following:

> Consider a selection-only function `:platform` that takes no operand or options, and matches on identifiers like `macos` or `linux` for the current OS. A Bad Expression error could be emitted if its value is used as a markup option value, as in:
> ```
> .local $foo = {:platform}
> {{{#bar opt=$foo}}}
> ```

OTOH, a bare Message Function Error could be valid for that too, and for a number of other cases that for whatever reason don't exactly fit within the categories we define here. Perhaps it'd be better to drop Bad Expression and make it explicitly clear that this list is not exhaustive?

aphillips: > Perhaps it'd be better to drop Bad Expression and make it explicitly clear that this list is not exhaustive?

We need to make an exhaustive list, from the point of view that every error is categorized. I'm okay with there being an error that is the moral equivalent of a plain `Exception` or that (as we currently have) there are some overarching errors (wasn't that what Message Function Error was going to be??). Implementers shouldn't have to guess what type of error to emit ("should I emit a Formatting Error or a Bad Expression or a Resolution Error?") even if on some level it doesn't matter.

I think I would leave out Bad Expression, especially since the things that trigger it are not "badness" in the expression but rather "badness" in the function's operation.

spec/formatting.md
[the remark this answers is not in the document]
aphillips: This is a very specific bit of guidance: if there is a fully resolved value _and_ the function determines it cannot be formatted by that function during the formatting process, isn't that the definition of a Bad Operand? Plus, this is `SHOULD` not `MUST`, so an implementation is still permitted to emit something else if it has good reason. 

If I'm an implementer and I'm reading the spec, I'm going to be looking for `MUSTard` that I am required to implement (and the tests that go with it). If we intend to be wiggly, we need to make it clear. In that case, it would look more like:

```suggestion
Resolved values cannot always be formatted by a given implementation.
When a _function_ determines that it does not support the resolve value of its _operand_
its implementation SHOULD emit an informative error, such as _Bad Operand_,
and produce a
```

eemeli, line 738:
```suggestion
an implementation SHOULD emit an apropriate _Message Function Error_ and produce a
```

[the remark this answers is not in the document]
eemeli: I've updated the text here to now say:
> When such an error occurs during _formatting_,
> an implementation SHOULD emit an appropriate _Message Function Error_ and produce ...

which is in line with @aphillips's suggestion in https://github.com/unicode-org/message-format-wg/pull/774#discussion_r1576395507.

spec/errors.md
```diff
@@ -241,49 +241,71 @@ or for private implementation use that is not supported by the current implement
...
+## Message Function Errors
 
-An **_<dfn>Operand Mismatch Error</dfn>_** is an _Invalid Expression_ error that occurs when
+A **_<dfn>Message Function Error</dfn>_** is any error that occurs
+when calling a message function implementation
+or which depend on validation associated with a specific function.
+
+Implementations SHOULD provide a way for custom functions to specify
+which of the following errors ought to be emitted,
+as well as allowing for other implementation-defined _Message Function Errors_.
```
aphillips, line 264: Nit-picking the language here a bit. I got rid of "custom" because it applies to non-custom functions also.

```suggestion
Implementations SHOULD provide a way for _functions_ to emit 
(or cause to be emitted) any of the types of error defined in this section.
Implementations MAY also provide implementation-defined _Message Function Error_ types.
```

aphillips (commented):
One change proposed, otherwise looks good.

spec/errors.md
```diff
@@ -302,11 +324,9 @@ An **_<dfn>Unsupported Statement</dfn>_** error occurs when a message includes a
...
-## Formatting Errors
+### Formatting Error
 
-**_<dfn>Formatting Errors</dfn>_** occur during the formatting of a resolved value,
-for example when encountering a value with an unsupported type
-or an internally inconsistent set of options.
+A **_<dfn>Formatting Error</dfn>_** occurs during the formatting of a resolved value.
 
 > For example, attempting to format any of the following messages
 > might result in a _Formatting Error_ if done within a context that
```
gibson042, line 332: > ```
> Hello, {horse :get field=name}!
> ```

Wouldn't this be a Bad Operand error, because "_`:get`… requires its argument to be an object_"?

> ```
> Hello, {$user :get}!
> ```

Wouldn't this be a Bad Option error, because `:get` requires "_an option field to be provided with a string value_"?

> ```
> .local $id = {$user :get field=id}
> {{Hello, {$id :get field=name}!}}
> ```

Wouldn't this be a Bad Operand error, because "_`:get`… requires its argument to be an object_"?

> ```
> Your {$field} is {$id :get field=$field}
> ```

Wouldn't this be an Unresolved Variable error, because `$id` cannot be resolved?

----

As described in [Formatting](https://github.com/unicode-org/message-format-wg/blob/b76dc21172364da7cf7f30ea5ed91e4e7137b9d4/spec/formatting.md#formatting) (**bold emphasis** mine),
> After <em>pattern selection</em>, each <em>text</em> and <em>placeholder</em> part of the selected <em>pattern</em> is **resolved** and formatted.

Formatting is separated from formatting context construction pattern selection, but all those phases include resolution and therefore the potential for resolution errors—in other words, "Resolution Errors" as a category does not seem to imply specificity to any particular phase. If the goal is to establish a _tree_ of error types rather than merely a DAG, then I think it is necessary to either include phase in the name of errors that could otherwise occur in more than one or to define that e.g. other error types are subsumed into Formatting Error during the formatting phase.

To be frank, this all seems rather messy. We should probably just drop Selection Error and Formatting Error and instead recommend that implementations communicate in another way the point at which error-free formatting was interrupted, allowing the errors theirselves to maintain the desired level of specificity.

gibson042 (changes requested):

spec/errors.md
```diff
@@ -241,49 +241,71 @@ or for private implementation use that is not supported by the current implement
...
 > ```
 
-## Selection Errors
+### Selection Error
 
-**_<dfn>Selection Errors</dfn>_** occur when message selection fails.
+A **_<dfn>Selection Error</dfn>_** occurs when message selection fails.
 
 > For example, attempting to format either of the following messages
 > might result in a _Selection Error_ if done within a context that
```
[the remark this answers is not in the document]
eemeli: You should get both. First Bad Operand, then Selection Error.

spec/errors.md
aphillips, line 20: Maybe murder this normative text?

The MUST isn't testable. It would be okay to have a normative statement about what errors can be emitted in a specific phase of processing. But the boundaries of 'selection' and 'formatting' are vague _in this context_.

Perhaps try turning this upside down:

```suggestion
_Message Function Errors_ MUST only be emitted during selection and formatting.
```

spec/errors.md
```diff
@@ -16,10 +16,10 @@ and MUST be emitted as soon as possible.
...
 
-During selection, an _expression_ handler MUST only emit _Resolution Errors_ and _Selection Errors_.
-During formatting, an _expression_ handler MUST only emit _Resolution Errors_ and _Formatting Errors_.
+During selection and formatting,
+_expression_ handlers MUST only emit _Message Function Errors_.
 
-_Resolution Errors_ and _Formatting Errors_ in _expressions_ that are not used
+_Resolution Errors_ and _Message Function Errors_ in _expressions_ that are not used
 in _pattern selection_ or _formatting_ MAY be ignored,
 as they do not affect the output of the formatter.
```
aphillips, line 24: Thought: can a _Message Function Error_ ("MFE") occur in an unused _expression_? You can't have an error from a function that is never called. Admittedly, some of the MFEs might be discerned from the registry, e.g. checking option names or option values, during parsing. Again, maybe turn this upside down:

```suggestion
Implementations do not have to check for or emit _Resolution Errors_
or _Message Function Errors_ in _expressions_ that are not otherwise used by the _message_,
such as _placeholders_ in unselected _patterns_
or _declarations_ that are never referenced during _formatting_.
```

spec/errors.md
eemeli: The intent is to require all emitted errors to be Message Function Errors, rather than any random errors. The phrasing is perhaps a bit awkward as this used to have two separate pharses talking about selection and formatting. Would this work?

```suggestion
An _expression_ handler MUST only emit _Message Function Errors_.
```

eemeli:
As discussed yesterday and following @gibson042's suggestion in https://github.com/unicode-org/message-format-wg/pull/774#discussion_r1578237772, I've now dropped the Selection Error and Formatting Error types.

However, while doing so, I noticed that Selection Error was used in two places by our current spec:
1. During pattern selection, for selectors that don't support selection. So I added Bad Selector for this, as a resolution error.
2. In the `:number` `MatchSelectorKeys()` call, for variant keys that aren't numbers or plural categories. So I added Bad Variant Key for this, as a message value error.

In total, this means that two somewhat generic error types are replaced by two very specific ones.

I also think we should consider adding something like Bad Formatter for placeholders that use a selection-only function annotation. As with Bad Selector, that error isn't really emitted by the function implementation, but by the MF2 implementation. But we can handle this as a separate next step.

spec/errors.md
aphillips: I think my comment still applies? If our set of errors is more-or-less comprehensive and our instructions are clear about when to emit specific errors, we don't need to have normative text forbidding implementers from choosing one or another error type or about when they emit those errors. It's impossible to test anyway.

We'd be better off with non-normative text that signals our intention.

We talk about syntax/data model errors specifically and about message function errors specifically in this section, but only mention resolution errors in passing.

I think I'd rewrite this section (including the paragraph above that I can't `suggestion` into submission...) along the lines of:

```suggestion
_Syntax Errors_ and _Data Model Errors_ can always be detected during message parsing or processing.
(Other categories of error can sometimes be detected using validation tools.)

_Resolution Errors_ can be detected during _variable_, _function_, or _markup_ resolution.

_Message Function Errors_ apply to _selection_ or _formatting_
and include errors detected during the evaluation of an
_expression_ or during the execution of a _function_.
```

spec/errors.md
```diff
@@ -241,82 +242,53 @@ or for private implementation use that is not supported by the current implement
...
-> .match {$sel}
-> 1 {{The value is one.}}
-> * {{The value is not one.}}
+> .local $day = {|2024-05-01| :date}
+> .match {$day}
+> * {{The due date is {$day}}}
 > ```
 
-## Formatting Errors
+## Message Function Errors
```
aphillips, line 271: The header should be singular too, even though it encapsulates other errors

```suggestion
## Message Function Error
```

spec/errors.md
```diff
@@ -335,3 +307,65 @@ or an internally inconsistent set of options.
 > Your {$field} is {$id :get field=$field}
 > ```
 
+### Bad Operand
+
+A **_<dfn>Bad Operand</dfn>_** error is an error that occurs when
+an _operand_ provided to a _function_ during _function resolution_ does not match one of the
```
aphillips, line 313: I'd break this up a bit to provide some wiggle room for any cases we forgot and to relax when it happens. "... during function resolution..." is too specific, since it would potentially preclude a tool from detecting it during parsing, etc.

```suggestion
A **_<dfn>Bad Operand</dfn>_** error is any error that occurs due to the content or format of the _operand_,
such as when the _operand_ provided to a _function_ during _function resolution_ does not match one of the

File: spec/errors.md [hunk 20]
```diff
@@ -335,3 +307,65 @@ or an internally inconsistent set of options.
 > Your {$field} is {$id :get field=$field}
 > ```
 
+### Bad Operand
```
aphillips, line 310: Should there also be a `Missing Operand` (or `Missing Option`) error for cases where a required operand (or option) is missing (`{:date}`) or null?

spec/errors.md
```diff
@@ -335,3 +307,65 @@ or an internally inconsistent set of options.
...
+
+### Bad Option
+
+A **_<dfn>Bad Option</dfn>_** error is an error that occurs when there is
+an implementation-defined error with an _option_ or its value.
+These might include:
+- A required _option_ is missing.
+- Mutually exclusive _options_ are supplied.
+- An _option_ value provided to a _function_ during _function resolution_
+   does not match one of the implementation-defined types for that _function_;
```
aphillips, line 342:
```suggestion
   does not match one of the implementation-defined types or values for that _function_;
```

spec/errors.md
```diff
@@ -335,3 +307,65 @@ or an internally inconsistent set of options.
...
+   or in which the literal _option_ value does not have the required format
+   and thus cannot be processed into one of the expected
+   implementation-defined types for that specific _function_.
+- Any other implementation-defined errors related to _options_.
```
aphillips, line 346: The list stars with "These might include:` so we can leave this out?

```suggestion
```

spec/errors.md
```diff
@@ -335,3 +307,65 @@ or an internally inconsistent set of options.
...
+> ```
+> The answer is {42 :number minimumFractionDigits=foo}.
+> ```
+
+### Bad Variant Key
+
+A **_<dfn>Bad Variant Key</dfn>_** error is an error that occurs when a _variant_ _key_
+does not match the expected implementation-defined format.
+
+> For example, the following _message_ produces a _Bad Variant Key_ error
```
aphillips, line 361: This should probably be presented more suppositionally:

```suggestion
> For example, the following _message_ might produce a _Bad Variant Key_ error
```

aphillips (changes requested):
Looks good! Mostly editorial suggestions.

spec/errors.md
eemeli: All of the other titles at this level use "Errors".

spec/errors.md
eemeli: I'm not so sure; it seems like a missing operand is a Bad Operand.

spec/errors.md
eemeli: No; it's using `:number`, and the spec for that requires a Bad Variant Key for this.

spec/errors.md
eemeli: Note the preceding paragraph, which already talks of catching errors with validation tools. So repeating that here seems rather odd.

aphillips: The "suggestion" block is intended to replace the preceding paragraph also. So it's not a repeat.

spec/errors.md
aphillips: "Unsupported Statement", "Bad Selector", "Bad Operand", "Bad Option", and "Bad Variant Key" are all singular. I guess Syntax Error and Resolution Error are presented as plural. Either way, we probably should be consistent.

In this case, "Message Function Error" is both a _class_ of errors and an error in its own right.

catamorphism (approved):
LGTM

spec/errors.md
eemeli: It's a class of errors, next to Syntax Errors, Data Model Errors, and Resolution Errors. Its naming is consistent with those, both in how they're titled here, and the way the name includes the word "Error". The spec language is in fact quite deliberately careful about not requiring a bare Message Function Error to be emittable; instead it either refers directly to a specific Message Function Error, or an "appropriate" Message Function Error.

spec/errors.md
eemeli: So you're proposing that we also drop the requirement to emit syntax & data model errors as early as possible?

Thus far, this PR is only about changing the names & categorisation of errors, which is already a lot. Separately we also have #704 attempting to change a related aspect of error handling.

Could we leave error handling changes for separate consideration? This seems like a topic we'll need to discuss more, and I would much rather not get the rest of this PR blocked on that.

If that's acceptable, please feel free to resolve this thread by accepting my last suggestion above, or by closing it and keeping the current language.

spec/errors.md
```diff
@@ -241,82 +242,53 @@ or for private implementation use that is not supported by the current implement
...
-> would always result in an _Unsupported Statement_ error:
+> would result in an _Unsupported Statement_ error:
 >
 > ```
 > .some {|horse|}
 > {{The message body}}
 > ```
 
-## Selection Errors
+### Bad Selector
```
aphillips, line 257: This is under Data Model errors, but I'm not sure it should be? A selector might be syntactically well-formed and have an annotation (which are the DM requirements) but the implementation of the function not support selection (`:datetime` currently is like this). Detecting this requires reading the implementation's installed function registry/registries. Should that be a Message Function Error?

spec/errors.md
aphillips: If it's a class of errors, shouldn't there be a concrete subtype for all of the known errors?

I think the problem here is that its vaguely clear that there is a generic Message Function Error error type.

Currently:
- Syntax Error is a class and a type (and that's fine).
- Data Model Error has an exhaustive list of error types.
- Resolution Error has an exhaustive list of error types.
- Message Function Error has a partial list, where the missing type appears to be the "💩 Happened Error" aka "implementation defined error"

spec/errors.md
aphillips: Yeah, I'm proposing that we spend more time on error handling, noting that there's a problem here. I'll create an issue to track that proposal so that we can merge this.

spec/errors.md
eemeli: > If it's a class of errors, shouldn't there be a concrete subtype for all of the known errors?

No, I don't think we need to make this list exhaustive. The others are exhaustive as they're emitted from our own code, but we should not presume that we can imagine all the things that can happen, and all the things that might go wrong, with custom functions.

aphillips: Actually, I think I'm trying to say that. The "catch all" error is a message function error, since everything else that can go user-visibly wrong is accounted for.

eemeli: So this thread can be resolved and the title left as it currently is, matching the others which include "Syntax Errors"?

aphillips (approved):

PR #775, opened 2024-04-23
Define missing definitions for variables

aphillips:
During publication of v45, I found that we were missing a number of terms that we actively use in the specification. Among these are:

- variable
- external variable
- local variable

This PR defines those terms.

### DIFF
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -97,6 +97,17 @@ Attempting to parse a _message_ that is not _valid_ will result in a _Data Model
 
 A **_<dfn>message</dfn>_** is the complete template for a specific message formatting request.
 
+The purpose of a _message_ is to insert _variable_ values into text
+or to select the most appropriate text according to the value of a _variable_
+(or both).
+
+A **_<dfn>variable</dfn>_** is a data value bound to a _name_.
+
+An **_<dfn>external variable</dfn>_** is a _variable_ supplied by the caller
+to MessageFormat or available in the _formatting context_.
+
+A **_<dfn>local variable</dfn>_** is a _variable_ created by a _declaration_.
+
 > [!NOTE]
 > This syntax is designed to be embeddable into many different programming languages and formats.
 > As such, it avoids constructs, such as character escapes, that are specific to any given file
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -97,9 +97,6 @@ Attempting to parse a _message_ that is not _valid_ will result in a _Data Model
 
 A **_<dfn>message</dfn>_** is the complete template for a specific message formatting request.
 
-The purpose of a _message_ is to insert _variable_ values into text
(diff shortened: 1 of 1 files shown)


spec/syntax.md
```diff
@@ -97,6 +97,17 @@ Attempting to parse a _message_ that is not _valid_ will result in a _Data Model
 
 A **_<dfn>message</dfn>_** is the complete template for a specific message formatting request.
 
+The purpose of a _message_ is to insert _variable_ values into text
+or to select the most appropriate text according to the value of a _variable_
+(or both).
```
eemeli, line 102: Talking about the "purpose" of a message feels really deep, and I'm not sure that I completely agree with this characterisation. Do we need this here, or could we leave it out?

aphillips: We could leave it out.

```suggestion
```

spec/syntax.md
```diff
@@ -97,6 +97,14 @@ Attempting to parse a _message_ that is not _valid_ will result in a _Data Model
 
 A **_<dfn>message</dfn>_** is the complete template for a specific message formatting request.
 
+
+A **_<dfn>variable</dfn>_** is a data value bound to a _name_.
```
catamorphism, line 101: I don't think "name" has been defined -- or is it meant to refer to _name_ as defined in the grammar? But if so, then _variable_ doesn't need to be defined here either, since it's in the grammar.

Also, this conflicts with the later statement that: "A local-declaration binds a variable to the resolved value of an expression." Maybe that sentence should be changed to say "name" instead of "variable"?

spec/syntax.md
```diff
@@ -97,6 +97,14 @@ Attempting to parse a _message_ that is not _valid_ will result in a _Data Model
...
+
+An **_<dfn>external variable</dfn>_** is a _variable_ supplied by the caller
+to MessageFormat or available in the _formatting context_.
```
catamorphism, line 104: "Formatting context" hasn't been defined yet.

catamorphism (changes requested):

eemeli (approved):

spec/syntax.md
aphillips: > I don't think "name" has been defined

`name` has a definition [here](https://github.com/unicode-org/message-format-wg/blob/main/spec/syntax.md#names-and-identifiers)

_variable_ needs to be defined because we (try to) link it many times in the spec.

> Maybe that sentence should be changed to say "name" instead of "variable"?

Good point. Perhaps by saying something like:

> A `local-declaration` creates a _variable_ by binding a _name_ to the resolve value of an _expression_.

gibson042 (approved):

catamorphism (approved):

eemeli (approved):

mihnita (approved):

PR #778, opened 2024-04-30
Validate test cases against a schema

echeran:
This PR is to add automation so that we guarantee that our test data remains in a consistent state. This prevents problems we had previous of ill-formed JSON and/or invalid contents according to the schema.

This PR moves the work from https://github.com/mradbourne/message-format-wg/pull/1 to the upstream, given that #767 was merged already without this addition in order to keep things simple for #767.

From the discussion there, we have a couple of options for how we go about automating the test validation:

* Option 1: write a custom script to validate test cases' JSON files
  - Pros: Can add logic to pick up the latest schema version
  - Cons: We might have to maintain code in the future
* Option 2: use an existing tool (ex: [`ajv-cli`](https://ajv.js.org/guide/environments.html#command-line-interface)) to validate test cases' JSON files
  - Pros: Use an off the shelf tool in JS for JSON
  - Cons: We have to remember to maintain the CI workflow to when there is a new schema version

Aside: A decision point from #767 was the format of the version number for the schemas - should it be semantic versioning or an auto-increment integer or something else. #767 changed to auto-increment integer, which I have also have a slight preference for. We're all in agreement there, AFAICT.


eemeli (changes requested):
We do not need a custom script for this. My suggestion in https://github.com/mradbourne/message-format-wg/pull/1#discussion_r1582641326 works to replace the current behaviour, but without a custom Python script.

If we want to automatically use the latest schema version, here's how to do that:

```sh
ajv validate --spec=draft2020 \
-s $(ls -1v schemas/*/*schema.json | tail -1) \
-d 'tests/**/*.json'
```

.github/workflows/validate_tests.yml
```diff
@@ -0,0 +1,24 @@
+name: Validate test data
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches: '**'
```
eemeli, line 8: Leaving out the branch specifier does the same thing, and is clearer.
```suggestion
  pull_request:
```

.github/workflows/validate_tests.yml
```diff
@@ -0,0 +1,24 @@
...
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+      - name: Install CLI tool for JSON Schema validation
+        run: npm install --global ajv-cli
+      - name: Validate tests using the latest schema version
+        run: |
+            ajv validate --spec=draft2020 \
+            -s $(ls -1v schemas/*/*schema.json | tail -1) \
+            -d 'tests/**/*.json'
```
eemeli, line 23: With `|` block literals, newlines are literal and therefore the trailing `\` are needed so that the shell can combine these into a single command. With a `>` block literal, the newlines are collapsed into single spaces, and the `\` aren't required.

```suggestion
        run: >
            ajv validate --spec=draft2020
            -s $(ls -1v schemas/*/*schema.json | tail -1)
            -d 'tests/**/*.json'
```

eemeli (approved):

.github/workflows/validate_tests.yml
```diff
@@ -0,0 +1,24 @@
+name: Validate test data
+
+on:
+  push:
+    branches:
+      - main
```
eemeli, line 6:
```suggestion
  push:
    branches:
      - main
    paths:
      - test/**
```

.github/workflows/validate_tests.yml
echeran: Adding the `paths` constraint sounds good. I personally find it clearer in this case to be explicit rather than implicit about the `branches` constraint for the `pull_request` trigger. I'm not sure if it's because the `push` trigger does have & need `branches`, or if it is due to habit of seeing it that way for so long.

PR #780, opened 2024-05-06
Add contextual options to expression-attributes design doc

eemeli:
The core idea is to provide explicit and well-defined support for `u:id`, `u:locale`, and `u:dir` when used as function or markup options, and to thereby clarify that expression attributes have no runtime impact.


exploration/u-options.md
```diff
@@ -0,0 +1,149 @@
...
+Provide common definitions for contextual options.
+
+## Background
+
+Function options may influence the resolution, selection, and formatting of annotated expressions.
+These provide a great solution for options like `minFractionDigits`, `dateStyle`,
+or other similar factors that influence the formatted result.
+
+However, this single bag of options is not appropriate in all cases,
+in particular for options that change the formatting context for the function.
```
aphillips, line 28: I'm not sure this is what you mean. Perhaps:

```suggestion
Such options naturally correspond to function arguments or builder-style function constructors.
Each option is specific to the associated API.
Users (message authors, translators, developers) should not expect
that a given option name has the same meaning between functions
or that its behavior stays the same.

To reduce the learning curve for users and improve consistency,
it would be useful to have common options 
(generally those related to the formatting context)
shared between all functions.
```

exploration/u-options.md
```diff
@@ -0,0 +1,149 @@
...
+  and allows the parts of an expression to be explicitly addressed.
+
+  > Example identifying two literal numbers:
+  >
+  > ```
+  > The first number was {1234 :number u:id=first} and the second {56789 :number u:id=second}.
+  > ```
+
+- An override for the locale used to format the expression.
+  Should be expressed as a non-empty sequence of BCP 47 language codes.
```
aphillips, line 45: Sequence?

```suggestion
The value of this option is a Unicode Locale Identifier
(which is a BCP47 language tag).
The root locale is expressed using the tag `und`.
```

exploration/u-options.md
```diff
@@ -0,0 +1,149 @@
...
+
+  > Example embedding a French literal in an English message:
+  >
+  > ```
+  > In French, “{|bonjour| :string u:locale=fr}” is a greeting
+  > ```
```
aphillips, line 51: This isn't perhaps a great example.

```suggestion
> Example. Overriding the formatting locale to use French:
>```
> In French, this date would be displayed as {|2024-05-06| :date u:locale=fr}
>```
```

aphillips (commented):
Partial review

exploration/u-options.md
```diff
@@ -0,0 +1,149 @@
+# `u:` Options
```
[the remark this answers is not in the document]
eemeli: Maybe a bit shorter even?
```suggestion
# Contextual Options
```

exploration/u-options.md
```diff
@@ -0,0 +1,149 @@
...
+
+Status: **Accepted**
```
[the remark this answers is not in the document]
eemeli: I would prefer starting with Accepted, and then only if necessary first merging it as Proposed. We should not start from the presumption that multiple phases of discussion are required for everything we consider via design documents.

exploration/u-options.md
eemeli: I copied this from the expression-attributes text. As this is describing a use case, it's perhaps best to not get stuck in the technical details here.
```suggestion
```

exploration/u-options.md
eemeli:
```suggestion
  > Example embedding a French date in an English message:
  >
  > ```
  > In French, this date would be displayed as {|2024-05-06| :date u:locale=fr}
  > ```
```

exploration/u-options.md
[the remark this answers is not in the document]
aphillips: We're going to discuss it in an hour anyway 😄

exploration/contextual-options.md
```diff
@@ -0,0 +1,156 @@
...
+
+- `u:id` — A string value that is included as an `id` or other suitable value
+  in the formatted part(s) for the placeholder,
+  or any other structured formatted results.
+  Ignored when formatting to a string, but could show up in error messages.
+- `u:locale` — A comma-delimited list of BCP 47 language tags,
+  or an implementation-defined list of such tags.
+  The tags are parsed, and they replace the _locale_
+  defined in the _formatting context_ for this expression or markup.
+- `u:dir` — One of the string values `ltr`, `rtl`, or `auto`.
```
mihnita, line 97: Might not be enough?
I think that some kind if "isolate" would be handy.

The equivalent of `<bdi>` in HTML. Not sure whey that is a separate concept in HTML, and I suspect history / legacy.

The `auto` in HTML:
> `auto`, which lets the user agent decide. It uses a basic algorithm as it parses the characters inside the element until it finds a character with a strong directionality, then applies that directionality to the whole element.
> https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/dir

That sounds an awful lot like `bdi`, but less clear:
> Embedding the characters in `<span dir="auto">` has the same effect as using `<bdi>`, but its semantics are less clear.
> https://developer.mozilla.org/en-US/docs/Web/HTML/Element/bdi

exploration/contextual-options.md
```diff
@@ -0,0 +1,156 @@
...
+
+Define at least `locale` and `dir` as options for default registry functions,
+with handling internal to each function implementation.
+
+Other functions could use their own definitions and handling for similar options,
+such as `locale` or `x:lang`.
+
+Formatted parts for markup would not be able to directly include an identifier.
+
+### Use expression attributes
```
mihnita, line 146: Expression attributes can be used without a function (or an function implied from the type of the argument, which is the same thing).

So we can say: `{$expDate @locale=fr}`, but we can't say `{$expDate u:locale=fr}`, we are forced to say `{$expDate :date u:locale=fr}`
I don't see it as a problem, a bit more typing in some rare cases, but just saying.

exploration/expression-attributes.md
```diff
@@ -88,10 +60,10 @@ At least the following expression attributes should be considered:
     whether the expression should or should not be localised.
     The values here correspond to those used for this property in HTML and elsewhere.
 
-    > Example embedding a non-translatable French literal in an English message:
+    > Example embedding a non-translatable CLI command in a message:
```
mihnita, line 63: Non-translatable is a not a use case.
`@translate=no` is unnecessary.

A function is either specified, or "implied" from the argument type.
I the example the function is not specified, and the type is string. So the implied function is `:string`, which does nothing.
And arguments are always non-localizable.

So `{|bonjour|}` is already unlocalizable.

---

The locale is also unnecessary, since there is no function to do any transformation, locale-aware or not.

For markup the locale might be mapped to a `lang` attribute in HTML, or `<lang>` in SSML (https://cloud.google.com/text-to-speech/docs/ssml#lang).
So I can see a use case for markup.

But I see no reason for this to be a "global" attribute, it would be specific to the markup dialect.
Same as "href", which is it is HTML specific.

---

I think we really have to separate runtime attributes from the l10n directives.

exploration/expression-attributes.md
```diff
@@ -88,10 +60,10 @@ At least the following expression attributes should be considered:
...
     >
     > ```
-    > In French, "{|bonjour| @locale=fr @translate=no}" is a greeting
+    > Use {+code @translate=no}git ls-files{-code} to list all files in a repository.
```
mihnita, line 66: That is probably incorrect.

`@translate` is not a runtime attribute, so it is intended for translators (humans or machine, it does not matter).

Until now the placeholders and markup don't have side-effects on anything outside them.
A translator can move around placeholders / markup, maybe delete it. But it is free to changer any text outside them.
But this proposal changes that.

We already have a way to achieve this:
```
Use {+code}{|git ls-files|}{-code} to list all files in a repository.
```

mihnita:
TLDR: I am happy with namespaced attributes to represent global/universal options (non-function dependent)
But that makes the attributes even less useful.
So I would suggest we drop them.

aphillips:
In the 2024-05-20 call, we agreed **_not_** to merge this PR but to have @eemeli fold this into exploration/expression-attributes.md design.

eemeli:
Refactored the PR to be a change/expansion of the expression-attributes design doc rather than a separate one.

exploration/expression-attributes.md
```diff
@@ -27,17 +28,31 @@ Function options may influence the resolution, selection, and formatting of anno
...
 or other similar factors that influence the formatted result.
 
-However, this single bag of options is not appropriate in all cases,
-in particular for attributes that pertain to the expression as a selector or a placeholder.
-For example, many of the [XLIFF 2 inline element] attributes don't really make sense as function options.
+Such options naturally correspond to function arguments or builder-style function constructors.
+Each option is specific to the associated API.
+Users (message authors, translators, developers) should not expect
+that a given option name has the same meaning between functions
+or that its behavior stays the same.
```
aphillips, line 35: I think this is backwards: if the learning curve is to be manageable, we want functions to be consistent. Overriding the locale or suppressing localization (to use two examples) should work the same everywhere rather than being idiosyncratic for each MF2 implementation **_and_** each function withing those implementations.

```suggestion
Message authors such as translators and developers want consistent ways to do common tasks, 
such as providing hints to translation tools or overriding the locale,
but, unless MessageFormat provides otherwise, cannot rely on implementations
to consistently implement these.
```

aphillips (changes requested):

exploration/expression-attributes.md
```diff
@@ -27,17 +28,31 @@ Function options may influence the resolution, selection, and formatting of anno
...
+To reduce the learning curve for users and improve consistency,
+it would be useful to have common options
+(generally those related to the formatting context)
+shared between all functions.
+
+Separately from formatting concerns,
+it is often useful to attach other information to message expressions and markup.
+For example, presenting how an example value could be formatted can be very useful for the message's translation,
+and providing the original source representation of a placeholder may be essential for being able to format a non-MF2 message,
+if it has been transformed to MF2 to provide translators with a unified experience.
```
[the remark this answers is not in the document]
eemeli: Maybe breaking up the run-one sentence would improve things? The original source isn't primarily there for translators, but other tooling.

```suggestion
For example, presenting how an example value could be formatted can be very useful for the message's translation.
Providing the original source representation of a placeholder may be essential for being able to format a non-MF2 message,
if it has been transformed to MF2 to provide translators with a unified experience.
```

exploration/expression-attributes.md
```diff
@@ -98,31 +113,31 @@ At least the following expression attributes should be considered:
...
+  - `id` — An identifier for the expression or markup.
     This is included in the formatted part,
-    and allows the parts of an expression to be explicitly addressed.
+    and allows each part of a message to be explicitly addressed.
 
     > Example identifying two literal numbers:
     >
     > ```
-    > The first number was {1234 :number @id=first} and the second {56789 :number @id=second}.
+    > The first number was {1234 :number u:id=first} and the second {56789 :number u:id=second}.
```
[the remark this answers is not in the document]
eemeli: That design option is here presented as the proposed one.

exploration/expression-attributes.md
```diff
@@ -98,31 +113,31 @@ At least the following expression attributes should be considered:
...
 
   - `locale` — An override for the locale used to format the expression.
-    Should be expressed as a non-empty sequence of BCP 47 language codes.
 
-    > Example embedding a French literal in an English message:
+    > Example embedding a French date in an English message:
     >
     > ```
-    > In French, “{|bonjour| @locale=fr}” is a greeting
+    > In French, this date would be displayed as {|2024-05-06| :date u:locale=fr}
```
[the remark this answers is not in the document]
eemeli: Because you asked for it: https://github.com/unicode-org/message-format-wg/pull/780#discussion_r1591067064

exploration/expression-attributes.md
```diff
@@ -176,20 +191,16 @@ including expressions without an annotation.
 
 Attributes are distinct from function options.
 
-Common attributes are defined by the MF2 specification
-and must be supported by all implementations.
+Common options or attributes should work the same way in different functions.
+
+Special options or attributes should not conflict with other option names.
```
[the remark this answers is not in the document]
eemeli: Special, because that can account for implementations-specific things that may be "common" within their scope, but not universally. So e.g. `@amzn:marketplace=US` or `amzn:marketplace=US`.

exploration/expression-attributes.md
```diff
@@ -204,38 +215,124 @@ the reserved/private-use rules will need to be adjusted to support attributes.
...
-### Do not support expression attributes
+### Do nothing
+
+Continue to [caution](https://github.com/unicode-org/message-format-wg/blob/d38ff326d2381b3ef361e996c3431d1b251518d6/spec/syntax.md#attributes)
+function authors and other implementers away from creating function-specific or implementation-specific option values
+for the use cases presented above.
+
+As should be obvious, the current situation is not tenable in the long term, and should be resolved.
+
+### Do not provide any guidance
```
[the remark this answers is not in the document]
eemeli: "Do nothing" leaves the caution in the `syntax.md` file; this would remove that.

exploration/expression-attributes.md
```diff
@@ -204,38 +217,124 @@ the reserved/private-use rules will need to be adjusted to support attributes.
 
 ## Proposed Design
 
-Add support for option-like `@key=value` attribute pairs at the end of any expression.
+Solve the two disparate use cases separately,
```
catamorphism, line 220: It might be helpful here to reiterate what the two disparate use cases are (attributes with a formatting runtime impact, and those without runtime impact).

exploration/expression-attributes.md
```diff
@@ -204,38 +217,124 @@ the reserved/private-use rules will need to be adjusted to support attributes.
...
+
+Formatted parts for markup would not be able to directly include an identifier.
+
+Do not define a common way to communicate information
+about an expression or markup to translators or tools.
+
+### Use attributes also for contextual options
+
+Add support for standalone `@key` as well as
+option-like `@key=value` attribute pairs with a literal or variable values
```
catamorphism, line 312:
```suggestion
option-like `@key=value` attribute pairs with a literal or variable value
```

exploration/expression-attributes.md
eemeli:
```suggestion
Provide separate solutions for attributes that impact formatting,
and those which do not,
```

exploration/expression-attributes.md
```diff
@@ -27,17 +28,33 @@ Function options may influence the resolution, selection, and formatting of anno
 These provide a great solution for options like `minFractionDigits`, `dateStyle`,
 or other similar factors that influence the formatted result.
 
-However, this single bag of options is not appropriate in all cases,
-in particular for attributes that pertain to the expression as a selector or a placeholder.
-For example, many of the [XLIFF 2 inline element] attributes don't really make sense as function options.
+Such options naturally correspond to function arguments or builder-style function constructors.
+Each option is specific to the associated API.
+Message authors such as translators and developers want consistent ways to do common tasks,
```
aphillips, line 33: Englische-type nit

```suggestion
Message authors, such as translators or developers, want consistent ways to do common tasks,
```

exploration/expression-attributes.md
```diff
@@ -204,38 +217,125 @@ the reserved/private-use rules will need to be adjusted to support attributes.
...
+or overriding the message locale or directionality.
+
+Do not define a common way to communicate information
+about an expression or markup to translators or tools.
+
+This would mean not defining anything for default registry functions either,
+effectively requiring implementation-specific options like `icu:locale`.
+
+Other functions could use their own definitions and handling for similar options,
+such as `locale` or `x:lang`.
```
aphillips, line 288: I think spelling out the pros/cons here might be useful?

> **Pros**
> - implementations would be free to define these options/attributes to suit local implementation needs
> 
> **Cons**
> - users could not count on these options to be available in each function
> - users would have to learn different requirements/restrictions 
>   on option use for each function and for each separate implementation

exploration/expression-attributes.md
```diff
@@ -204,38 +217,125 @@ the reserved/private-use rules will need to be adjusted to support attributes.
...
+### Define options for default registry only
 
-A bit less bad than the previous, but still mixes attributes and options into the same namespace.
+Define at least `locale` and `dir` as options for default registry functions,
+with handling internal to each function implementation.
+
+Other functions could use their own definitions and handling for similar options,
+such as `locale` or `x:lang`.
+
+Formatted parts for markup would not be able to directly include an identifier.
```
aphillips, line 305: I don't follow this statement. Why wouldn't I be able to have a `{#b id=foo}`? Or a `{$x :number id=foo}`?

exploration/expression-attributes.md
```diff
@@ -204,38 +217,125 @@ the reserved/private-use rules will need to be adjusted to support attributes.
...
 
 Function options may be used as a workaround,
 but each implementation and user will end up with different practices.
 
-### Use function options, but with some suggested prefix like `_`
+### Define options for default registry only
 
-A bit less bad than the previous, but still mixes attributes and options into the same namespace.
+Define at least `locale` and `dir` as options for default registry functions,
+with handling internal to each function implementation.
```
aphillips, line 300: I think you should address how `id` _could_ work here. Presumably the `id` property of the resolved value could be queried?

exploration/expression-attributes.md
```diff
@@ -204,38 +217,125 @@ the reserved/private-use rules will need to be adjusted to support attributes.
...
+
+Other functions could use their own definitions and handling for similar options,
+such as `locale` or `x:lang`.
+
+Formatted parts for markup would not be able to directly include an identifier.
+
+Do not define a common way to communicate information
+about an expression or markup to translators or tools.
```
aphillips, line 308: This is either a dangling "Con" or wants to be the introduction to this sub-section, e.g.:

> ### Define options for default registry only
> Do not define a common way to communication information
> about an expression or markup to translators or tools.
> Instead, define `locale`, `dir`, and `id`, as options on each default registry function,
> with handling internal to each function's implementation,
> and serving as a model for non-default functions or for markup.

exploration/expression-attributes.md
```diff
@@ -204,38 +217,125 @@ the reserved/private-use rules will need to be adjusted to support attributes.
...
+
+Define the meaning and supported values of some attributes in the specification,
+including at least `@dir` and `@locale`.
+
+To support later extension of the specified set of attributes while allowing user extensibility,
+require custom attribute names to be namespaced.
+Examples: `@xliff:can-copy=no`, `@note:link=|https://...|`.
+
+Allow expression attributes to influence the formatting context,
+but do not directly pass them to user-defined functions.
```
aphillips, line 328: This is too suppositional.

```suggestion
Expression attributes are stored in or modify the formatting context,
rather than passed directly to functions (including user-defined functions).
Function implementations can access the values of attributes via the formatting
context, but are not required to do so.
```

exploration/expression-attributes.md
```diff
@@ -204,38 +217,125 @@ the reserved/private-use rules will need to be adjusted to support attributes.
...
+
+### Use function options, but with some suggested "discard" namespace like `_`
+
+Examples: `_:translate=yes` and `_:example=|World|`.
+
+Requires reserving an additional namespace.
+
+Requires cooperation from implementers to ignore all options using the namespace.
+
+Makes defining namespaced attributes difficult.
```
aphillips, line 338: Use pro/con for arguments for/against the option.

I would call out that the discard namespace `_` is functionally identical to `u`

aphillips (commented):
Lots of these comments are related to the formatting and structuring of the design doc rather than to the technical arguments.

exploration/expression-attributes.md
eemeli: This statement is specifically about formatted parts for markup, not expressions. Functions can define how expressions get formatted, but markup is not configurable. Sure, you could pass markup an `id=foo` option, but that option would be only present in the general bag of options in the formatted part, and would potentially conflict with other use.

For example, it's rather common for HTML elements to include an `id` value, which is already relevant for the CSS or JS executing on that document, but an identifier may be separately relevant for the post-processing of a formatted message, such as adding in parameters that were not included in the translation.

In other words, if we're translating a message like
```
Would you like to {#a u:id=continue}accept{/a} or {#a u:id=cancel}reject{/a} the offer?
```
we need a way to disambiguate the two formatted `{#a}` spans from each other, hence the `u:id`. That way the message can end up first formatted as:
```js
[
  { type: 'literal', value: 'Would you like to ' },
  { type: 'markup', kind: 'open', source: '#a', name: 'a', id: 'continue' },
  { type: 'literal', value: 'accept' },
  { type: 'markup', kind: 'close', source: '/a', name: 'a' },
  { type: 'literal', value: ' or ' },
  { type: 'markup', kind: 'open', source: '#a', name: 'a', 'id': 'cancel' },
  { type: 'literal', value: 'reject' },
  { type: 'markup', kind: 'close', source: '/a', name: 'a' },
  { type: 'literal', value: ' the offer?' }
]
```

And from there, we can construct an HTML snippet with additional properties:
```html
Would you like to <a href="continue?token=123abc">accept</a> or <a href="cancel?token=123abc">reject</a> the offer?
```

But if there's no common `u:id` definition, then the identifier ends up mixed together with the other markup options, and it gets much trickier to build tooling that's universal across different implementations and function authors, and doesn't risk bleeding through into the output HTML, where a property like `id="continue"` could very easily conflict with some other operations and expectations of the page.

exploration/expression-attributes.md
```diff
@@ -204,38 +217,125 @@ the reserved/private-use rules will need to be adjusted to support attributes.
...
 If not explicitly defined, less information will be provided to translators.
 
 Function options may be used as a workaround,
 but each implementation and user will end up with different practices.
 
-### Use function options, but with some suggested prefix like `_`
+### Define options for default registry only
 
-A bit less bad than the previous, but still mixes attributes and options into the same namespace.
+Define at least `locale` and `dir` as options for default registry functions,
```
eemeli, line 299:
```suggestion
Do not define a common way to communicate information
about an expression or markup to translators or tools.
Instead, define at least `locale` and `dir` as options for default registry functions,
```

exploration/expression-attributes.md
eemeli: I'm not sure what you mean by "suppositional". I think what you're describing is one possible way of implementing this alternative, but it's not the only one. Should we split this option into two, one of which only defines an explicit short list of attributes with well-defined runtime effects, and the other allows for values to be stored in the expression's formatting context?

exploration/expression-attributes.md
eemeli:
```suggestion
Requires cooperation from implementers to ignore all options using the namespace.

Makes defining namespaced attributes difficult.

Could be combined with the definition of `u:dir`, `u:locale`, and `u:id` contextual options.
```

exploration/expression-attributes.md
eemeli: At the moment, these alternative descriptions are not formatted as pro/con lists, and I would rather keep them internally consistent. Therefore, I'd much prefer not adding such a list to just one option, and I don't want to refactor the entire PR again to include such a reformatting. We could do such a reformatting as a follow-on change.

exploration/expression-attributes.md
eemeli: I'm honestly not sure how to make `id` sensibly work, if we're not defining it generally in the spec but only specifically for each standard registry function. That would mean not having any specific field for it in the formatting-internal resolved values, and intermingling it with the other resolved options.

I hope we do not end up with this option, and so can avoid needing to answer all the questions that it would introduce.

exploration/expression-attributes.md
aphillips: Okay, I see what you're saying now. My example omitted the `u:` by mistake. I still think that the text you have the design document is super-mysterious, though. Note that attributed strings can have overlapping markup (that would be illegal in HTML). For example:

> Would you like to {#span type=foo u:id=1}click {#span type=bar u:id=2}here{/span u:id=1} to get {/span u:id=2} illegal stuff.


Perhaps:

```suggestion
Lack of a built-in identifier would inhibit tools support for markup,
particularly when formatted to parts, 
when the relationship of open and close markup is ambiguous.
> For example, the two `{/a}` markup elements in this mess:
>```
> Would you like to {#a u:id=continue}accept{/a} or {#a u:id=cancel}reject{/a} the offer?
>```
```

exploration/expression-attributes.md
aphillips: I agree that we should not end up with this option. I'd probably address it by saying:

```suggestion
Define the `locale`, `dir`, and `id` options for all default registry functions
and establish a policy of including them in all future default registry functions,
with the hope that other functions and implementations will adopt this as a best practice.
Actually implementation would be internal to each function.
```

exploration/expression-attributes.md
aphillips: I said "suppositional" because it's hard to say what "influencing" the formatting context means?

I agree that my suggestion is one possible way of implementing this and that there could be others. However, by expressing it concretely, I hope to enable us to have a concrete conversation about which option to pursue. 

I probably wouldn't go to the trouble of splitting the option up. I'd just note that my suggestion is one route and limiting the attributes to the built-in set is another?

exploration/expression-attributes.md
eemeli: I think that a reference to "tools support" here would be misleading, as we're talking about the actual formatted parts. Also, as markup formats as an empty string, anything relating to it is only relevant with formatted parts or similar non-concatenated output formats.

In my experience, needing to rely on something like `u:id` to detangle improper nesting is relatively rare compared to needing to allow for localizations to reorder actions such as "accept" and "cancel"; in my previous comment's example I included a `u:id` to show how that can be supported.

```suggestion
Formatted parts for markup would not be able to directly include an identifier.
Implementations and users will need to invent their own practices with
markup option names like `l10n-id`, `mf:id`, or `markupId`
to refer to specific markup parts.
```

exploration/expression-attributes.md
eemeli: If we were to define `id` for default registry functions, we would also need to define what it does for them. Could you give some example for how you think that could be done without also defining `id` as somehow special in the spec?

aphillips: Oh no, I agree with you there. These would be somehow special. If we don't reserve them outright for all functions everywhere (and instead do it by policy and hope for imitation by local functions), then people will be confused later when local functions work funny. Note that this is one argument for why expression annotations make sense to me: since they are not options, there is no danger of conflict or redefinition.

eemeli: Given the clarifying text about hacky identifiers added a few lines below this, is there anything actionable left in this thread?

PR #781, opened 2024-05-06
Simplify source bidi isolation rules

eemeli:
Drop the `bidi` rule, and allow `name` to be LR/RL/FS -isolated.

Allow an LRI immediately after a non-content newline.

Relax expression & markup isolation to not require pairing on a syntactic level, as the LRI can also be terminated by a newline.

### DIFF
--- a/exploration/bidi-usability.md
+++ b/exploration/bidi-usability.md
@@ -249,7 +249,7 @@ This is not the disadvantage to right-to-left languages that it might first appe
   correct order and without spillover effects
 
 Permit isolating bidi controls to be used on the **outside** of the following:
-- names
+- unquoted literals
 - quoted literals
 - quoted patterns
 
@@ -258,15 +258,13 @@ the user to set the base direction of a _literal_ or _pattern_ according to its 
 actual contents.
 
 This would change the ABNF as follows:
+(Notice that this change includes a production `bidi` described further down
+in this document)
 ```abnf
-name           = (open-isolate name-body close-isolate) / name-body
-name-body      = name-start *name-char
-
-quoted         = (open-isolate quoted-body close-isolate) / quoted-body
-quoted-body    = "|" *(quoted-char / quoted-escape) "|"
-
-quoted-pattern = (open-isolate "{{" pattern "}}" close-isolate)
-               / "{{" pattern "}}"
+literal        = ( open-isolate (quoted / (unquoted [bidi])) close-isolate)
+               / (quoted / (unquoted [bidi]))
+quoted-pattern = ( open-isolate "{{" pattern "}}" close-isolate)
+               / ("{{" pattern "}}")
 
 open-isolate   = %x2066-2068
 close-isolate  = %x2069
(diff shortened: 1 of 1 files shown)


aphillips:
I wish you'd added this as a separate alternative.

I don't like that the isolates are part of the `name` rule---I worked hard to keep the isolates _outside_ the rules for important constructs (like `name`)

You removed unquoted literals from being amenable to bidi isolation, but they should still be isolatable, no?

eemeli:
> I don't like that the isolates are part of the `name` rule---I worked hard to keep the isolates _outside_ the rules for important constructs (like `name`)

Including the isolates in `name` doesn't change its parsed meaning, much like the `|` aren't a part of the parsed meaning of a quoted literal. It's the same situation as with isolated expressions, markup and patterns.

> You removed unquoted literals from being amenable to bidi isolation, but they should still be isolatable, no?

They are, covered by the change to `name`:

```abnf
unquoted       = name / number-literal
```
`number-literal` doesn't need isolation, because we've limited its valid values, so isolating `name` is enough.

aphillips:
The problem with allowing isolates into `name` is that it makes name comparison harder. Shouldn't the following two names be equal?

```
\u2066name\u2069
name
```

> number-literal doesn't need isolation, because we've limited its valid values, so isolating name is enough.

Actually, numbers are complicated in bidi because digits are weakly directional. The minus sign can swing around onto the "wrong" side visually.

The other reason I had unquoted and quoted together is that it simplifies what tools have to do. A tool can blindly isolate any literal _separate_ from the decision to quote it and can blindly remove isolates from literals without looking at the contents.

eemeli:
> The problem with allowing isolates into `name` is that it makes name comparison harder. Shouldn't the following two names be equal?
> 
> ```
> \u2066name\u2069
> name
> ```

As proposed, both of those strings would match the `name` rule, but as `\u2066` and `\u2069` are not valid `name-char` characters, they would be parsed according to the `open-isolate` and `close-isolate` rules, with `name-body` matching the four-character "name" string in both cases.

So the parsed value of the name would be "name" for both of the above, and they would be considered equal.

> > number-literal doesn't need isolation, because we've limited its valid values, so isolating name is enough.
> 
> Actually, numbers are complicated in bidi because digits are weakly directional. The minus sign can swing around onto the "wrong" side visually.

But `number-literal` only shows up in "code", which is always LTR, yes?

> The other reason I had unquoted and quoted together is that it simplifies what tools have to do. A tool can blindly isolate any literal _separate_ from the decision to quote it and can blindly remove isolates from literals without looking at the contents.

The proposed change doesn't change the number of constructs for which this can be done; it replaces "unquoted literals" with "names". Doing so lets us remove needing to separately and additionally pick out the LRM/RLM/ALM from the productions that include `name`.

exploration/bidi-usability.md
```diff
@@ -256,13 +258,15 @@ the user to set the base direction of a _literal_ or _pattern_ according to its
...
 
 This would change the ABNF as follows:
-(Notice that this change includes a production `bidi` described further down
-in this document)
 ```abnf
-literal        = ( open-isolate (quoted / (unquoted [bidi])) close-isolate)
-               / (quoted / (unquoted [bidi]))
-quoted-pattern = ( open-isolate "{{" pattern "}}" close-isolate)
-               / ("{{" pattern "}}")
+name           = (open-isolate name-body close-isolate) / name-body
```
aphillips, line 262: This is a problem because `name` is used to build a variety of other constructs (`variable`, `reserved-keyword`, `identifier`, etc.). This change puts the isolates _inside_ these constructs, e.g. `$\u2066name\u2069` rather than on the outside.

This will make it harder for implementations, since they can't take the parsed token and compare it immediately. They have to stop to remove isolates. My original design avoided this problem by making the isolates not parse into names/identifiers/tokens.

aphillips (changes requested):
Please change this PR to make your proposal an additional option, not overwriting the original design.

eemeli:
As requested, refactored as an alternative to the proposed solution. Also addressed the concerns identified in #787 and #788, and added an example showing how `name` isolation avoids a spillover the current proposal cannot.

I have also validated this solution by [implementing](https://github.com/messageformat/messageformat/compare/mf2-source-bidi) it in my parser.

exploration/bidi-usability.md
```diff
@@ -346,6 +348,66 @@ the results or debug what is wrong with their messages.
...
+
+### Loose isolation
+
+Apply a slightly different set of bidi isolates.
+The main differences to the proposed solution are:
+1. The open/close isolate characters are not syntactically required to be paired.
+   This avoids introducing parse errors for missing or required invisible characters,
+   which would lead to bad user experiences.
+2. Rather than patching the `name` rule with an optional trailing LRM/RLM/ALM,
+   allow for its proper isolation.
```
aphillips, line 360: I wouldn't call what we did above "patching". What we allow above with the strongly directional marks is allow bidi users to include them (to make the string look okay in a normal text editor) the way they might normally do when editing text. The productions we used don't make these marks part of the token, so they don't affect processing.

Allowing isolation is a separate consideration.

exploration/bidi-usability.md
```diff
@@ -346,6 +348,66 @@ the results or debug what is wrong with their messages.
...
+
+Quoted patterns, quoted literals, and names may be isolated by LRI/RLI/FSI...PDI.
+For names and quoted literals, the isolate characters are outside the body of the token,
+but for quoted patterns, the isolates are in the middle of the `{{` and `}}` characters.
```
aphillips, line 364: "middle" could mean anywhere inside the pattern quotes.

```suggestion
but for quoted patterns, the isolates are in between the `{` and `}` in the `{{` and `}}` sequences.
```

exploration/bidi-usability.md
```diff
@@ -346,6 +348,66 @@ the results or debug what is wrong with their messages.
...
+This avoids adding a lookahead requirement for detecting a `complex-message` start,
+and differentiating a `quoted-pattern` from a `quoted` `key` in a `variant`.
```
aphillips, line 366: Englishche-type nit...

```suggestion
and differentiates a `quoted-pattern` from a `quoted` `key` in a `variant`.
```

aphillips (changes requested):
The requested changes are editorial. Otherwise I would approve this addition.

I don't think I agree with this option. The strongly directional marks are included in the proposed solution for a different reason than might be assumed (I mention this below) and I don't think putting isolates into `name` has been fully accounted for.

exploration/bidi-usability.md
```diff
@@ -346,6 +348,66 @@ the results or debug what is wrong with their messages.
...
+
+An LRI is allowed immediately after a newline outside patterns and within expressions.
+This is intended to allow left-to-right representation for "code"
+even if it contains a newline followed by content
+that could otherwise prompt the paragraph direction to be detected as right-to-left.
+
+```abnf
+name           = [open-isolate] name-start *name-char [close-isolate]
+quoted         = [open-isolate] "|" *(quoted-char / quoted-escape) "|" [close-isolate]
+quoted-pattern = "{" [open-isolate] "{" pattern "}" [close-isolate] "}"
```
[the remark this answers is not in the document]
eemeli: Yes, that's the intent: `{\u2066{`

PR #791, opened 2024-05-13
Update design docs and their status

aphillips:
This PR updates the status of various design docs and otherwise fixes minor issues (mainly that the title wasn't set). It also adds notes to non-standard docs kept for historical purposes.

### DIFF
--- a/exploration/variable-mutability.md
+++ b/exploration/variable-mutability.md
@@ -1,4 +1,4 @@
-# Design Proposal Template
+# Variable Mutability
 
 Status: **Accepted**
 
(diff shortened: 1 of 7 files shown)


exploration/variable-mutability.md
```diff
@@ -1,4 +1,4 @@
-# Design Proposal Template
+# Variable Mutability
```
eemeli, line 1: The `<summary>` value below should also be corrected to "Metadata".

```suggestion
# Variable Namespacing and Mutability
```

eemeli (approved):
This improves things, but further changes will still be needed, such as the ones included in #616.

PR #792, opened 2024-05-14
[DESIGN] Add user stories / build-out of the expression attributes

aphillips:
Add user-stories and examples to the expression attributes design.


eemeli (commented):
Looks pretty good. The `quoting` of e.g. "locale" and "direction" in text looks a bit odd, and the code examples probably don't need `>` indentation.

A third user story could be added to account for the use case introduced in #772.

Remember to add this PR to the "Pull Requests" list.

PR #793, opened 2024-05-14
Recommend not escaping all the things

eemeli:
As [discussed](https://github.com/unicode-org/message-format-wg/blob/main/meetings/2024/notes-2024-05-13.md#743) on yesterday's call, we should instruct processors to only escape as necessary.

This adds a new normative SHOULD for "when programmatically emitting message syntax", which is novel for the spec.

### DIFF
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -948,9 +948,10 @@ backslash    = %x5C ; U+005C REVERSE SOLIDUS "\"
 > they do not need to be escaped, such as braces in a _quoted_ _literal_.
 > For example, `|foo {bar}|` and `|foo \{bar\}|` are synonymous.
 
-When programmatically emitting message syntax,
-only characters that are lexically meaningful in the current context
-SHOULD be escaped.
+When writing or generating a _message_, escape sequences SHOULD NOT be used
+unless required by the syntax.
+That is, inside _literals_ only escape `|` 
+and inside _patterns_ only escape `{` and `}`.
 
 ### Whitespace
 
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -943,6 +943,15 @@ escaped-char = backslash ( backslash / "{" / "|" / "}" )
 backslash    = %x5C ; U+005C REVERSE SOLIDUS "\"
 ```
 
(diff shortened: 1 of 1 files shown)


spec/syntax.md
```diff
@@ -943,6 +943,15 @@ escaped-char = backslash ( backslash / "{" / "|" / "}" )
...
 ```
 
+> [!NOTE]
+> The `escaped-char` rule allows escaping some characters in places where
+> they do not need to be escaped, such as braces in a _quoted_ _literal_.
+> For example, `|foo {bar}|` and `|foo \{bar\}|` are synonymous.
+
+When programmatically emitting message syntax,
+only characters that are lexically meaningful in the current context
+SHOULD be escaped.
```
alerque, line 953: I definitely support this, the normalized version should only have the minimum number of escapes. How about rewording where the SHOULD is though?

```suggestion
When programmatically emitting message syntax, escape sequences
SHOULD only be used for characters that are lexically meaningful in the current context.
```

The idea is to get the SHOULD to be closer to "only" rather than hanging out with "be escaped".  It just seems too easy to skim past the 'only' and end up with exactly the opposite idea.

alerque (changes requested):

spec/syntax.md
aphillips: I'd go further and not just cover programmatic generators, but people too.

```suggestion
When writing or generating a _message_, escape sequences SHOULD NOT be used
unless required by the syntax.
That is, only escape `|` inside of _literals_ 
and `{` and `}` inside of _patterns_ or _simple messages_.
```

alerque: Even better!

eemeli: Modified @aphillips's suggestion a bit, to ensure that we're not disallowing by omission escapes in _reserved_.

```suggestion
When writing or generating a _message_, escape sequences SHOULD NOT be used
unless required by the syntax.
That is, inside _literals_ only escape `|` 
and inside _patterns_ only escape `{` and `}`.
```

aphillips (approved):

PR #800, opened 2024-05-22
Use mappings for data model options and attributes

### LINKED ISSUE #716: [FEEDBACK] The data model could be simplified for options and attributes
_Parts of this issue have been broken out into their own issues #717 and #718 after this was initially posted._

While working on some Python code, I ended up needing to put together a [pythonic representation](https://github.com/eemeli/l10n-tools/blob/main/moz_l10n/message.py) of the message data model. While doing so, I encountered a few places where I could apply some simplifications to the data model with no loss of fidelity.

I think we should consider applying this change to how options and attributes are represented in the data model:

```diff
 interface FunctionAnnotation {
   type: "function";
   name: string;
-  options: Option[];
+  options: Map<string, Literal | VariableRef>;
 }

-interface Option { name: string, value: Literal | VariableRef }

 interface LiteralExpression { ...
 interface VariableExpression { ...
 interface FunctionExpression { ...
 interface UnsupportedExpression { ...
 interface Markup { ...
-  attributes: Attribute[];
+  attributes: Map<string, Literal | VariableRef | null>;
 }

-interface Attribute { name: string, value?: Literal | VariableRef }
```

As discussed most recently in #710, option names cannot be duplicated. This means that instead of an array of name+value wrapper objects, options could be represented by a 1:1 mapping. I'm using a JS Map here rather than an Object, as the former more explicitly does not include any prototype fields. In the JSON Schema, this could equivalently be represented by a JSON Object.

One explicit benefit of this change would be that messages coming from JSON interchange could not include any Duplicate Option Name errors.

The same applies for attributes as for options, but with the added note that values may be `null` because the attribute value is optional.

Separately, we probably should add a requirement for functions not to assign any meaning to option order, such that `:func foo=42 bar=13` and `:func bar=13 foo=42` would always be synonymous.

eemeli:
Closes #716

As discussed in the parent issue and its other follow-up #751, the order of options and attributes should not matter for their formatting or other processing. Reflecting this in the data model by making these use a mapping makes it clearer that e.g. `{:foo one=13 two=42}` and `{:foo two=42 one=13}` should be considered equivalent.

Note, though, that the order isn't completely lost here. The TS data model uses an ordered `Map` for these, and numerous JSON processors provide guarantees about retaining the order, even though the JSON spec does not require that. So making this change does not in practice make it harder for message syntax serialisers to retain the order that options or attributes may have had originally.

### DIFF
--- a/spec/data-model/README.md
+++ b/spec/data-model/README.md
@@ -180,34 +180,31 @@ interface LiteralExpression {
   type: "expression";
   arg: Literal;
   annotation?: FunctionAnnotation | UnsupportedAnnotation;
-  attributes: Attribute[];
+  attributes: Attributes;
 }
 
 interface VariableExpression {
   type: "expression";
   arg: VariableRef;
   annotation?: FunctionAnnotation | UnsupportedAnnotation;
-  attributes: Attribute[];
+  attributes: Attributes;
 }
 
 interface FunctionExpression {
   type: "expression";
   arg?: never;
   annotation: FunctionAnnotation;
-  attributes: Attribute[];
+  attributes: Attributes;
 }
 
 interface UnsupportedExpression {
   type: "expression";
   arg?: never;
   annotation: UnsupportedAnnotation;
-  attributes: Attribute[];
+  attributes: Attributes;
 }
 
-interface Attribute {
-  name: string;
-  value?: Literal | VariableRef;
-}
+type Attributes = Map<string, Literal | VariableRef | true>;
 ```
 
 ## Expressions
@@ -237,19 +234,16 @@ interface VariableRef {
 A `FunctionAnnotation` represents a _function_ _annotation_.
 The `name` does not include the `:` starting sigil.
 
-Each _option_ is represented by an `Option`.
+Each _option_ is represented by an `Options` key-value mapping.
 
 ```ts
 interface FunctionAnnotation {
   type: "function";
   name: string;
-  options: Option[];
+  options: Options;
 }
 
-interface Option {
-  name: string;
-  value: Literal | VariableRef;
-}
+type Options = Map<string, Literal | VariableRef>;
 ```
 
 An `UnsupportedAnnotation` represents a
@@ -276,15 +270,15 @@ A `Markup` object has a `kind` of either `"open"`, `"standalone"`, or `"close"`,
 each corresponding to _open_, _standalone_, and _close_ _markup_.
 The `name` in these does not include the starting sigils `#` and `/` 
 or the ending sigil `/`.
-The optional `options` for markup use the same `Option` as `FunctionAnnotation`.
+The optional `options` for markup use the same key-value mapping as `FunctionAnnotation`.
 
 ```ts
 interface Markup {
   type: "markup";
   kind: "open" | "standalone" | "close";
   name: string;
-  options: Option[];
-  attributes: Attribute[];
+  options: Options;
+  attributes: Attributes;
 }
 ```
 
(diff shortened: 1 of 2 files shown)


spec/data-model/README.md
```diff
@@ -180,34 +180,31 @@ interface LiteralExpression {
...
   annotation: UnsupportedAnnotation;
-  attributes: Attribute[];
+  attributes: Attributes;
 }
 
-interface Attribute {
-  name: string;
-  value?: Literal | VariableRef;
-}
+type Attributes = Map<string, Literal | VariableRef | true>;
```
aphillips, line 207: This changes the definition of attribute by adding `| true`. The syntax doesn't currently allow this?

spec/data-model/README.md
```diff
@@ -237,19 +234,16 @@ interface VariableRef {
 A `FunctionAnnotation` represents a _function_ _annotation_.
 The `name` does not include the `:` starting sigil.
 
-Each _option_ is represented by an `Option`.
+Each _option_ is represented by an `Options` key-value mapping.
```
aphillips, line 237: Document the contents of `Options` in its own place and document `Options` here.

```suggestion
`Options` is a map containing the _expression_'s _options_.
```

spec/data-model/README.md
```diff
@@ -276,15 +270,15 @@ A `Markup` object has a `kind` of either `"open"`, `"standalone"`, or `"close"`,
 each corresponding to _open_, _standalone_, and _close_ _markup_.
 The `name` in these does not include the starting sigils `#` and `/` 
 or the ending sigil `/`.
-The optional `options` for markup use the same `Option` as `FunctionAnnotation`.
+The optional `options` for markup use the same key-value mapping as `FunctionAnnotation`.
```
aphillips, line 273: "optional options" seems redundant? Consider making the text here exactly the same as in `FunctionAnnotation`

spec/data-model/README.md
```diff
@@ -237,19 +234,16 @@ interface VariableRef {
...
   name: string;
-  options: Option[];
+  options: Options;
 }
 
-interface Option {
-  name: string;
-  value: Literal | VariableRef;
-}
+type Options = Map<string, Literal | VariableRef>;
```
aphillips, line 246: Put documentation here for both `FunctionAnnotation` and Markup. (not using suggestion markdown because structural change would be needed)

> Options is a key-value map containing options.
> Implementations can use an ordered map if they want 
> to maintain the source ordering of options.
> Implementations might also canonicalize messages by ordering
> options into code point order.

aphillips (changes requested):
Good start. I think this is the right direction.

spec/data-model/README.md
eemeli: The `true` is meant to be used for representing value-less attributes like `@foo`, which the syntax does currently support.

spec/data-model/README.md
eemeli:
```suggestion
`Options` is a key-value mapping containing options,
and is used to represent the _annotation_ and _markup_ _options_.
```

spec/data-model/README.md
eemeli:
```suggestion
The `options` for markup use the same key-value mapping as `FunctionAnnotation`.
```

aphillips (approved):

catamorphism:
As part of this change, I think it would be useful to change https://github.com/unicode-org/message-format-wg/blob/main/spec/errors.md to add a new category of errors. I'm not sure what it would be called, but it seems weird to call "duplicate option name" a data model error if messages that have this error are unrepresentable in the data model. When I read "data model error" I imagine a property that can be violated by a data model. I think it would be useful to define a category of errors that aren't syntax errors, but that have to be checked before constructing a data model.

PR #802, opened 2024-05-27
Simplify terms in function-composition-part-1 design doc

catamorphism:
@macchiati suggested using names for models instead of "model 1" and "model 2". This introduces "the simple model" and "the preservation model".

### DIFF
--- a/exploration/function-composition-part-1.md
+++ b/exploration/function-composition-part-1.md
@@ -469,7 +469,10 @@ abstract "input" and "output" types.
 Returning to Example Y1, consider two possible models
 of the runtime behavior of function composition.
 
-**Model 1:**
+**Preservation Model (Model 1)**
+
+We call this the "preservation" model because it preserves
+the options in the result of the function.
 
 1. Evaluate `$num` to a value and pass it to the `:number` function,
    along with named options `{"maxFrac": "2"}`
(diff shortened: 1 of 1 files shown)


exploration/function-composition-part-1.md
```diff
@@ -469,7 +469,10 @@ abstract "input" and "output" types.
 Returning to Example Y1, consider two possible models
 of the runtime behavior of function composition.
 
-**Model 1:**
+**Preservation Model (Model 1)**
```
eemeli, line 472: Why are the numbers still retained here (parenthetically)?

exploration/function-composition-part-1.md
```diff
@@ -469,7 +469,10 @@ abstract "input" and "output" types.
...
+
+We call this the "preservation" model because it preserves
+the options in the result of the function.
```
aphillips, line 475: Use more direct language.

```suggestion
This model preserves the options in the result of the function.
```

exploration/function-composition-part-1.md
```diff
@@ -494,7 +497,10 @@ of the runtime behavior of function composition.
 
 then the formatted result is "0.33 0.33333".
 
-**Model 2:**
+**Simple Model (Model 2):**
+
+We call this the "simple" model because the result of the function
+is simpler than in the preservation model.
```
aphillips, line 503: This is unclear? What is simple or simpler?

```suggestion
### Formatted Value Model

This model preserves the formatted value of the function,
but not the options that were passed to the function.
```

exploration/function-composition-part-1.md
catamorphism: Fixed

exploration/function-composition-part-1.md
catamorphism: Sure, that's a good name. I applied this change and updated the rest of the document to reflect that.

eemeli (approved):

PR #803, opened 2024-05-29
Align test schema with error spec

mradbourne:
The error spec has been [recently updated](https://github.com/unicode-org/message-format-wg/pull/774/files#diff-be63d22facf59ed840140350b7431e109c4777901ec230052009aff7bb99eb75). This PR brings the test schema and existing tests in-line with the latest error definitions.

### DIFF
--- a/test/tests/functions/date.json
+++ b/test/tests/functions/date.json
@@ -11,7 +11,7 @@
       "exp": "{:date}",
       "expErrors": [
         {
-          "type": "invalid-expression"
+          "type": "bad-operand"
         }
       ]
     },
@@ -20,7 +20,7 @@
       "exp": "{|horse|}",
       "expErrors": [
         {
-          "type": "operand-mismatch"
+          "type": "bad-operand"
         }
       ]
     },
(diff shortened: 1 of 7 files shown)


test/tests/functions/date.json
```diff
@@ -11,7 +11,7 @@
       "exp": "{:date}",
       "expErrors": [
         {
-          "type": "invalid-expression"
+          "type": "bad-operand"
```
mradbourne, line 14: Just highlighting an assumption I've made:
This test results in `bad-operand` because it is effectively being passed an operand of 'undefined', which is not compatible with the `:date` function.

aphillips (approved):
Looks good, in general.

It's suspicious that we only have `bad-operand` tests in the PR. That suggests we lack tests.

catamorphism (approved):
LGTM!

eemeli (approved):
@aphillips This does include a couple of `bad-option` errors as well. But we definitely need more tests; many of the errors will need to use `:test:function` to be triggered.

srl295 (approved):
For the record, the specific spec PR seems to be https://github.com/unicode-org/message-format-wg/pull/774 (this PR just linked to latest)

PR #804, opened 2024-06-03
Add design doc for error handling

echeran:
Add design doc to contextualize the PR #795 based on the issue & discussion begun in #782 .

### DIFF
--- a/exploration/error-handling.md
+++ b/exploration/error-handling.md
@@ -42,7 +42,7 @@ In both languages, built-in error handling assumes a singular error.
 ICU4X operates in low resource environments for which returning at most 1 error is desirable
 because returning more than 1 error would require heap allocation.
 * Programming conventions and idioms: in [feedback from ICU-TC](https://docs.google.com/document/d/11yJUWedBIpmq-YNSqqDfgUxcREmlvV0NskYganXkQHA/edit#bookmark=id.lx4ls9eelh99),
-they found over the 25 years of maintaining the library that there was more cost than benefit in providing a default best effort return value at the same time as providing error information.
+they found over the 25 years of maintaining the library that there was more cost than benefit in additionally providing error information with a default best effort return value compared to just returning the default best effort value.
 The additional constraint in ICU4C's C++ style to return an error code rather than throwing errors using the STL further complicates the usefulness and likelihood to be used correctly during nested calls.
 
 ## Proposed Design
--- a/exploration/error-handling.md
+++ b/exploration/error-handling.md
@@ -11,8 +11,16 @@ Status: **Proposed**
 		<dd>2024-06-02</dd>
         <dt>Issues</dt>
 		<dd><a href="https://github.com/unicode-org/message-format-wg/issues/782">#782</a></dd>
+		<dd><a href="https://github.com/unicode-org/message-format-wg/issues/830">#830</a></dd>
+		<dd><a href="https://github.com/unicode-org/message-format-wg/issues/831">#831</a></dd>
 		<dt>Pull Requests</dt>
 		<dd><a href="https://github.com/unicode-org/message-format-wg/pull/795">#795</a></dd>
+		<dt>Meeting Notes</dt>
+		<dd><a href="https://github.com/unicode-org/message-format-wg/blob/main/meetings/2024/notes-2024-05-06.md">2024-05-06</a></dd>
+		<dd><a href="https://github.com/unicode-org/message-format-wg/blob/main/meetings/2024/notes-2024-05-13.md">2024-05-13</a></dd>
+		<dd><a href="https://github.com/unicode-org/message-format-wg/blob/main/meetings/2024/notes-2024-05-20.md">2024-05-20</a></dd>
+		<dd><a href="https://github.com/unicode-org/message-format-wg/blob/main/meetings/2024/notes-2024-07-15.md">2024-07-15</a></dd>
+		<dd><a href="https://github.com/unicode-org/message-format-wg/blob/main/meetings/2024/notes-2024-07-22.md">2024-07-22</a></dd>
 	</dl>
 </details>
 
--- a/exploration/error-handling.md
+++ b/exploration/error-handling.md
@@ -45,6 +45,18 @@ because returning more than 1 error would require heap allocation.
 they found over the 25 years of maintaining the library that there was more cost than benefit in additionally providing error information with a default best effort return value compared to just returning the default best effort value.
 The additional constraint in ICU4C's C++ style to return an error code rather than throwing errors using the STL further complicates the usefulness and likelihood to be used correctly by developers, especially during nested calls.
 
+> [!NOTE]
+> The wording in this document uses the word "signal" in regards to providing
+> information about an error rather than "return" or "emit" when referring to
+> a requirement that an implementation must at least indicate that an error has
+> occurred.
+> The word "signal" better accomodates more alternatives in the solution space
+> like those that only choose to indicate that an error occurred,
+> while still including those that additionally prefer to return the error
+> itself as an error object.
+> (By contrast, "return an error" implies that an error object will be thrown or
+> returned, and "emit an error" is ambiguous as to what is or isn't performed.)
+
 ## Proposed Design
 
 The following spec text is proposed:
--- a/exploration/error-handling.md
+++ b/exploration/error-handling.md
@@ -58,7 +58,7 @@ The additional constraint in ICU4C's C++ style to return an error code rather th
 > returned, and "emit an error" is ambiguous as to what is or isn't performed.)
 ## Use Cases
 
-As a software developer, I want calls to message format to signal runtime errors
+As a software developer, I want message formatting calls to signal runtime errors
 in a manner consistent with my programming language/environment.
 I would like error signals to include diagnostic information that allows me to debug errors.
 
@@ -67,8 +67,8 @@ even if a runtime error has occurred.
 
 As a software developer, I sometimes want to avoid "fatal" error signals,
 such as might occur due to unconstrained inputs,
-due to errors in translation of the message,
-or other reasons outside the developer's control.
+errors in translation of the message,
+or other reasons outside my control.
 For example, in Java, throwing an Exception is a common means of signaling an error.
 However, `java.text.NumberFormat` provide both throwing and non-throwing
 `parse` methods to allow developers to avoid a "fatal" throw of `ParseException`
@@ -96,7 +96,7 @@ which can be accomplished in different ways. Ex:
 * two APIs, a permissive one that always returns a best-effort formatted result
 and a stricter one that throws or returns an Error object when encountering an error
 * an API that always returns a best-effort formatted result
-and a global `boolean` values
+and sets a global `boolean` value
 
 This does not give implementations full freedom to return _nothing_ or some other behavior.
 
--- a/exploration/error-handling.md
+++ b/exploration/error-handling.md
@@ -99,6 +99,26 @@ For example, in ICU,
 [the suggested practice](https://docs.google.com/document/d/11yJUWedBIpmq-YNSqqDfgUxcREmlvV0NskYganXkQHA/edit#bookmark=id.lx4ls9eelh99)
 is to avoid additionally returning optional error codes when providing best-effort formatted results.
 
+### Require a best-effort message value and signaling of an error
+
+> In all cases, when encountering an error,
+> a message formatter MUST be able to signal an error or errors.
+> It MUST also provide the appropriate fallback representation of the _message_ defined
+> in this specification.
+
+This alternative maintains that an implementation must both return a best-effort message
+and must signal that an error has occurred.
+Compared to the current text,
+this alternative slightly loosens the constraint of 
+returning an error into only needing to signal an error.
+
+Similar to the current spec text,
+this alternative requires implementations to provide useful information:
+both a signal that an error occurred and a best effort message.
+A downside to this alternative is that these requirements together assume that
+all implementations will want to pay the cost of constructing a representative mesage
+after the occurrence of an error.
+
 ### Allow implementations to determine all details
 
 > When encountering an error during formatting,
--- a/exploration/error-handling.md
+++ b/exploration/error-handling.md
@@ -1,6 +1,6 @@
 # Error Handling
 
-Status: **Proposed**
+Status: **Accepted**
 
 <details>
 	<summary>Metadata</summary>
@@ -15,6 +15,7 @@ Status: **Proposed**
 		<dd><a href="https://github.com/unicode-org/message-format-wg/issues/831">#831</a></dd>
 		<dt>Pull Requests</dt>
 		<dd><a href="https://github.com/unicode-org/message-format-wg/pull/795">#795</a></dd>
+		<dd><a href="https://github.com/unicode-org/message-format-wg/pull/804">#804</a></dd>
 		<dt>Meeting Notes</dt>
 		<dd><a href="https://github.com/unicode-org/message-format-wg/blob/main/meetings/2024/notes-2024-05-06.md">2024-05-06</a></dd>
 		<dd><a href="https://github.com/unicode-org/message-format-wg/blob/main/meetings/2024/notes-2024-05-13.md">2024-05-13</a></dd>
--- a/exploration/error-handling.md
+++ b/exploration/error-handling.md
@@ -28,7 +28,7 @@ Decide whether and what implementations "MUST" / "SHOULD" / "MAY" perform after 
 
 In practice,
 runtime errors happen when formatting messages.
-It is useful to provide information about any formatting error(s) back to the callsite.
+It is useful to provide information about any errors back to the callsite.
 It is useful to the end user to provide best effort fallback representation of the message.
 Specifying the behavior in such cases promotes consistent results across conformant implementations.
 
@@ -43,7 +43,7 @@ ICU4X operates in low resource environments for which returning at most 1 error 
 because returning more than 1 error would require heap allocation.
 * Programming conventions and idioms: in [feedback from ICU-TC](https://docs.google.com/document/d/11yJUWedBIpmq-YNSqqDfgUxcREmlvV0NskYganXkQHA/edit#bookmark=id.lx4ls9eelh99),
 they found over the 25 years of maintaining the library that there was more cost than benefit in providing a default best effort return value at the same time as providing error information.
-The additional constraint in ICU4C's C++ style to use error code rather than throwing errors using the STL further complicates the usefulness and likelihood to be used correctly during nested calls.
+The additional constraint in ICU4C's C++ style to return an error code rather than throwing errors using the STL further complicates the usefulness and likelihood to be used correctly during nested calls.
 
 ## Proposed Design
 
--- a/exploration/error-handling.md
+++ b/exploration/error-handling.md
@@ -106,11 +106,14 @@ is to avoid additionally returning optional error codes when providing best-effo
 > It MUST also provide the appropriate fallback representation of the _message_ defined
 > in this specification.
 
-This alternative maintains that an implementation must both return a best-effort message
-and must signal that an error has occurred.
-Compared to the current text,
-this alternative slightly loosens the constraint of 
-returning an error into only needing to signal an error.
+This alternative requires that an implementation provide both an error signal
+and a means of accessing a "best-effort" fallback message.
+This slightly relaxes the requirement of "returning" an error
+(to allow a locally-appropriate signal of the error).
+
+Under this alternative, implementations can be conformant by providing
+two separate formatting methods or functions,
+one of which returns the fallback string and one of which signals the error.
 
 Similar to the current spec text,
 this alternative requires implementations to provide useful information:
--- a/exploration/error-handling.md
+++ b/exploration/error-handling.md
@@ -77,78 +77,15 @@ However, `java.text.NumberFormat` provide both throwing and non-throwing
 As a MessageFormat implementer, I want to be able to signal errors in an idiomatic way
 for my language and still be conformant with MF2 requirements.
 
-## Proposed Design
+## Accepted Design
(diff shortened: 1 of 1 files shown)


exploration/error-handling.md
```diff
@@ -0,0 +1,103 @@
...
+
+1. information about error(s)
+    - including, if relevant, the minimum number of errors for which such information is expected
+1. a fallback representation of the message
+
+## Background
+
+In practice,
+runtime errors happen when formatting messages.
+It is useful to provide information about any formatting error(s) back to the callsite.
```
aphillips, line 31: Not just formatting errors.

```suggestion
It is useful to provide information about any errors back to the callsite.
```

exploration/error-handling.md
```diff
@@ -0,0 +1,103 @@
...
+In Java, errors are thrown and subsequently caught in `try...catch` block.
+In Rust, fallible callsites (those which can return errors) should return a `Result<T, Err>` monad. 
+In both languages, built-in error handling assumes a singular error.
+* Environment constriants: as mentioned in [feedback from ICU4X](https://github.com/unicode-org/message-format-wg/issues/782#issuecomment-2103177417),
+ICU4X operates in low resource environments for which returning at most 1 error is desirable
+because returning more than 1 error would require heap allocation.
+* Programming conventions and idioms: in [feedback from ICU-TC](https://docs.google.com/document/d/11yJUWedBIpmq-YNSqqDfgUxcREmlvV0NskYganXkQHA/edit#bookmark=id.lx4ls9eelh99),
+they found over the 25 years of maintaining the library that there was more cost than benefit in providing a default best effort return value at the same time as providing error information.
+The additional constraint in ICU4C's C++ style to use error code rather than throwing errors using the STL further complicates the usefulness and likelihood to be used correctly during nested calls.
```
aphillips, line 46: typo

```suggestion
The additional constraint in ICU4C's C++ style to return an error code rather than throwing errors using the STL further complicates the usefulness and likelihood to be used correctly during nested calls.
```

exploration/error-handling.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Proposed Design
+
+The following spec text is proposed:
+
+> In all cases, when encountering an error during formatting,
+> a message formatter MUST provide some representation of the message,
+> or MUST provide an informative error or errors.
+> An implementation MAY provide both.
```
aphillips, line 55: To me, the fact of an error is more important than the fallback representation. Showing the fallback representation to end users is less good than having good error recovery (or no error at all). This wording makes the fallback representation first and thus, seemingly, the more important thing.

I also find the "MUST or MUST" formulation to be weird. I think this is trying to say "MUST provide  some representation of the message or an informative error or errors or both"

I think it would be weird if an implementation _never failed_, that is, if it returned the fallback representation silently with no signal of failure. When would this be a good thing? How would the caller find out about the error.

Note that we do not specify how errors are signaled. Just because (for example) Java often throws `Throwable` does not mean that the MF2 implementation has to use that as the error mechanism. See, for example, the use of `ParsePosition` in `NumberFormat`.

Thus I'd suggest an additional alternative:

```suggestion
> In all cases, when encountering an error,
> a message formatter MUST provide an informative error or errors.
> It MAY also provide the appropriate fallback representation of the _message_ defined
> in this specification.
```

exploration/error-handling.md
```diff
@@ -0,0 +1,103 @@
...
+
+This solution requires implementations to return _something_,
+but it leaves the decision to the implementation whether to:
+
+* return an error (or errors)
```
aphillips, line 60: I would probably say:

```suggestion
* return or emit an error or errors
```

One reason is that throwing an exception (in languages that do that) is not the same as returning a return value.

eemeli (changes requested):
As discussed on yesterday's call, I think we need to require two different things:
1. An implementation must provide a way for a message with a runtime error to be formatted in a way that provides the user with that error.
2. An implementation must provide a way for a message with a runtime error to be formatted using a fallback representation.

These separate requirements could be fulfilled by a single formatting method, or by more than one such method.

exploration/error-handling.md
```diff
@@ -0,0 +1,103 @@
...
+1. information about error(s)
+    - including, if relevant, the minimum number of errors for which such information is expected
+1. a fallback representation of the message
+
+## Background
+
+In practice,
+runtime errors happen when formatting messages.
+It is useful to provide information about any errors back to the callsite.
+It is useful to the end user to provide best effort fallback representation of the message.
```
catamorphism, line 32:
```suggestion
It is useful to the end user to provide a best effort fallback representation of the message.
```

catamorphism (commented):
I think this is good, if Addison's suggestion about eliminating the "MUST or MUST" phrasing is incorporated.

Per Eemeli's comment, maybe the idea of having the implementation provide two different methods should be listed as one of the alternatives? The current language mentions "the message formatter" and it's unclear from that whether the formatter has one formatting method or several.

exploration/error-handling.md
echeran: I used the word "signal" instead of "emit" because that better accomodates all of the alternatives in the solution space.

Note: I had to rewrite this section to reflect the extra constraint that I am taking on from your suggestion in the other [conversation thread](https://github.com/unicode-org/message-format-wg/pull/804#discussion_r1624685199), since this section existed to clarify the spec text concretely and in plain words.

echeran:
> As discussed on yesterday's call, I think we need to require two different things:
> 
> 1. An implementation must provide a way for a message with a runtime error to be formatted in a way that provides the user with that error.
> 2. An implementation must provide a way for a message with a runtime error to be formatted using a fallback representation.
> 
> These separate requirements could be fulfilled by a single formatting method, or by more than one such method.

@eemeli PTAL at the changes. See the discussion in [this conversation within this PR](https://github.com/unicode-org/message-format-wg/pull/804#discussion_r1630040749) where I address this. Also take a look at the wording and rationale for it based on [our discussion in Monday's meeting](https://github.com/unicode-org/message-format-wg/blob/main/meetings/2024/notes-2024-06-03.md#804), in particular my intent to be precise on what we're requiring & not, and whether that means the appropriate language is "returning"/"emitting"/"signalling" an error.

exploration/error-handling.md
```diff
@@ -0,0 +1,106 @@
...
+
+* Programming language: the language of the implementation informs idiomatic patterns of error handling.
+In Java, errors are thrown and subsequently caught in `try...catch` block.
+In Rust, fallible callsites (those which can return errors) should return a `Result<T, Err>` monad. 
+In both languages, built-in error handling assumes a singular error.
+* Environment constriants: as mentioned in [feedback from ICU4X](https://github.com/unicode-org/message-format-wg/issues/782#issuecomment-2103177417),
+ICU4X operates in low resource environments for which returning at most 1 error is desirable
+because returning more than 1 error would require heap allocation.
+* Programming conventions and idioms: in [feedback from ICU-TC](https://docs.google.com/document/d/11yJUWedBIpmq-YNSqqDfgUxcREmlvV0NskYganXkQHA/edit#bookmark=id.lx4ls9eelh99),
+they found over the 25 years of maintaining the library that there was more cost than benefit in providing a default best effort return value at the same time as providing error information.
```
eemeli, line 45: Do I understand right that this is referring to the following comments from the linked document?
> Rich: What Tim was describing was what we do in some places in ICU where we return a warning. Like our Missing Resource Warning.
> Markus: Right, and we’ve regretted that decision for years. We cooked up the warnings as a side channel so that we didn’t have to create APIs for various things. Ex: we went to a fallback language or default language instead of the request language. This is clunky because it doesn't survive nested functions or chained functions. It’s also confusing b/c we happily continue when there’s a warning code. You can have a function set a warning code, then you have a chain of 3-4 functions afterwards that are executed before we have code that looks at the error code, and then you don’t even know what the warning pertained to once you see it. And we keep Java & C++ in sync, but there’s no way to implement warnings in Java. Why do we have warnings? It’s so that the caller can decide to perform alternate behavior. However, we’ve been much more successful just giving a best effort return value that’s not exactly what the caller requested.

If that's the case, then isn't the feedback more appropriately represented as the inverse of what's summarised here?

```suggestion
* Programming conventions and idioms: in [feedback from ICU-TC](https://docs.google.com/document/d/11yJUWedBIpmq-YNSqqDfgUxcREmlvV0NskYganXkQHA/edit#bookmark=id.lx4ls9eelh99),
they found over the 25 years of maintaining the library that there was more cost than benefit in providing error information at the same time as providing a default best effort return value.
```

exploration/error-handling.md
```diff
@@ -0,0 +1,106 @@
...
+
+> In all cases, when encountering a runtime error,
+> a message formatter MUST provide some representation of the message.
+> An informative error or errors MUST also be separately provided.
+
+This alternative places constraints on implementations to provide multiple avenues of useful information (to the callsite and user).
+
+This alternative establishes constraints that would contravene the constraints that exist in projects that have implemented MF 2.0 (or likely will soon), based on:
+* programming language idioms/constraints
+* execution environment constraints
```
eemeli, line 83: What execution environment constraints does this alternative contravene? The only such mentioned in this document is that the cost of returning more than one error may be prohibitive in some cases, and the current text explicitly says "error or errors" to allow for an implementation signaling a single error to be valid.
```suggestion
```

exploration/error-handling.md
```diff
@@ -0,0 +1,106 @@
...
+* experience-based programming guidelines
```
eemeli, line 84: It is not clear to me what these guidelines are. Could they be explicitly spelled out?

eemeli (changes requested):
This document continues to not present and consider the alternative discussed on the call and in my preceding comment, of separating the two MUST requirements more clearly from each other.

exploration/error-handling.md
echeran: The text says "an informative error". As written, that implies to me an error object, like a `java.lang.Throwable` instance. Despite discussion in last week's meeting to interpret that phrase as equivalent to "whether or not an error occurred", it comes across distinctly as something more and thus should be rewritten if the intent is not so. There are applications like ICU and potentially browsers that might only want to provide a best effort message and signal an error, but not pay the cost of creating "informative error" objects each time.

As I mentioned in the 2024-04-09 meeting, another paradigm from which to look at this, besides "whether is returning an error possible", is "how actionable is returning the error object". ICU & browsers need to be performant and might not want to pay the cost of the creating a full error object.

exploration/error-handling.md
echeran: Done. Added text and a link to the same document as mentioned above.

echeran:
> This document continues to not present and consider the alternative discussed on the call and in my preceding comment, of separating the two MUST requirements more clearly from each other.

You did say that, although even after consulting the [notes from the meeting](https://github.com/unicode-org/message-format-wg/blob/main/meetings/2024/notes-2024-06-03.md#804), I couldn't understand what you meant by separating the 2 MUST requirements more clearly. If it's a matter of saying that you can satisfy the intended requirements in different ways, including having 2 APIs side-by-side in which one is stricter and signals an error while the other is permissive and always returns a best-effort value, the proposed alternative does that. 

The only thing that seemed close to matching your description of what you'd like is the following excerpt from the meeting's notes doc, which sounds much like the current text, so I didn't really see its distinguishing features enough to make it a new alternative:

> EAO: My position is that we should impose some requirements on implementations - what they must be able to provide. The way I would put the requirements - there must be a way to format the message that you get a fallback and that you must be able to format the message and get failure reported to you somehow. This would require behaviors from the implementer.

If you can write out the alternative you're thinking with pros & cons listed, I would be okay adding it.

eemeli:
> > This document continues to not present and consider the alternative discussed on the call and in my preceding comment, of separating the two MUST requirements more clearly from each other.
> 
> You did say that, although even after consulting the [notes from the meeting](https://github.com/unicode-org/message-format-wg/blob/main/meetings/2024/notes-2024-06-03.md#804), I couldn't understand what you meant by separating the 2 MUST requirements more clearly. If it's a matter of saying that you can satisfy the intended requirements in different ways, including having 2 APIs side-by-side in which one is stricter and signals an error while the other is permissive and always returns a best-effort value, the proposed alternative does that.

While the proposed alternative does allow for an API as you describe, it does not _require_ that a fallback representation is made available. The spec text that's currently proposed

> In all cases, when encountering an error,
> a message formatter MUST be able to signal an error or errors.
> It MAY also provide the appropriate fallback representation of the _message_ defined
> in this specification.

allows for a formatter that either returns the formatted result or `null` on error, as the latter can act as a signal that an error has occurred. This is not sufficient.

The requirement of a way to get a formatted representation of messages with runtime errors is important, because it needs to be accounted for in the formatter design: A formatter cannot just bail on the first error, but must have the capability of continuing to format a message past the first error.

The inclusion of "In all cases" also does not make it clear that an implementation providing two formatting methods (a strict thrower and a lenient fallback formatter) would be valid, given that formatting with the latter would not necessarily signal any error. This is an aspect of the current spec language that I believe for us to have consensus on fixing.

> If you can write out the alternative you're thinking with pros & cons listed, I would be okay adding it.

I'll see if I have time for that later this week, but it may take until next week.

exploration/error-handling.md
```diff
@@ -87,6 +99,26 @@ For example, in ICU,
...
+> In all cases, when encountering an error,
+> a message formatter MUST be able to signal an error or errors.
+> It MUST also provide the appropriate fallback representation of the _message_ defined
+> in this specification.
+
+This alternative maintains that an implementation must both return a best-effort message
+and must signal that an error has occurred.
+Compared to the current text,
+this alternative slightly loosens the constraint of 
+returning an error into only needing to signal an error.
```
aphillips, line 113: I think this is slightly misleading, in that it would be conformant to have two `format` methods, one that returns an error and one that returns the fallback message (cf. the conversation with @eemeli) 

Perhaps:

```suggestion
This alternative requires that an implementation provide both an error signal
and a means of accessing a "best-effort" fallback message.
This slightly relaxes the requirement of "returning" an error
(to allow a locally-appropriate signal of the error).

Under this alternative, implementations can be conformant by providing
two separate formatting methods or functions,
one of which returns the fallback string and one of which signals the error.
```

exploration/error-handling.md
```diff
@@ -83,6 +95,33 @@ This alternative establishes constraints that would contravene the constraints t
 * execution environment constraints
 * experience-based programming guidelines
 
+For example, in ICU, 
+[the suggested practice](https://docs.google.com/document/d/11yJUWedBIpmq-YNSqqDfgUxcREmlvV0NskYganXkQHA/edit#bookmark=id.lx4ls9eelh99)
+is to avoid additionally returning optional error codes when providing best-effort formatted results.
```
eemeli, line 100: This applies to all solutions that require a runtime error to be signalled in some way, right?

eemeli: By "solution" I meant the choice of spec language that we are debating here.

As I understand it (and please do correct me if I'm wrong), the ICU experience-based programming guideline in this case is roughly speaking, "You should not report formatting errors if you can fallback to some formatted result."

My point here is that this guideline applies to all spec text choices which require an error to be reported; for example, the current proposed design states:

> In all cases, when encountering an error,
> a message formatter MUST be able to signal an error or errors.

Therefore, that choice would also go against the above-mentioned experience-based programming guidelines. Is this not the case?

Just to be clear, I disagree about following this guideline for MF2, but here I'm trying to make sure that it's logically and consistently applied as it's being used as an argument in choosing between alternative designs.

exploration/error-handling.md
```diff
@@ -0,0 +1,145 @@
...
+* an API that throws or returns an Error object when encountering an error
+* an API that includes a read/write `ErrorCode` argument 
+* two APIs, a permissive one that always returns a best-effort formatted result
+and a stricter one that throws or returns an Error object when encountering an error
+* an API that always returns a best-effort formatted result
+and a global `boolean` values
+
+This does not give implementations full freedom to return _nothing_ or some other behavior.
+
+## Alternatives Considered
```
eemeli, line 81: As discussed in https://github.com/unicode-org/message-format-wg/pull/804#issuecomment-2158131626:

```suggestion
## Alternatives Considered

### Separate Requirements

The following spec text is proposed:

> When formatting a message with one or more errors:
> - An implementation MUST provide a way for a user to be informed
>   of the name of at least one of the errors,
>   either directly or via an identifying error code.
> - An implementation MUST provide a way for a message with one or more
>   _Resolution Errors_ or _Message Function Errors_ to be formatted
>   using a fallback representation.
>
> The two above requirements MAY be fulfilled by a single formatting method,
> or separately by more than one such method.

This alternative ensures that even messages with runtime errors
can be formatted with the same string representation in all implementations.

Compared to the current spec text,
the under-specified "informative error" is replaced
with a more concrete minimum requirement,
which can be fulfilled with a two-byte error code in very limited environments.

The phrase "In all cases" is left out,
and the ability to fulfill the requirements separately
rather than at the same time is explicitly called out.
```

exploration/error-handling.md
```diff
@@ -0,0 +1,145 @@
...
+> information about an error rather than "return" or "emit" when referring to
+> a requirement that an implementation must at least indicate that an error has
+> occurred.
+> The word "signal" better accomodates more alternatives in the solution space
+> like those that only choose to indicate that an error occurred,
+> while still including those that additionally prefer to return the error
+> itself as an error object.
+> (By contrast, "return an error" implies that an error object will be thrown or
+> returned, and "emit an error" is ambiguous as to what is or isn't performed.)
+
```
aphillips, line 59:
```suggestion
## Use Cases

As a software developer, I want calls to message format to signal runtime errors
in a manner consistent with my programming language/environment.
I would like error signals to include diagnostic information that allows me to debug errors.

As a software developer, I sometimes need to be able to emit a formatted message
even if a runtime error has occurred.

As a software developer, I sometimes want to avoid "fatal" error signals,
such as might occur due to unconstrained inputs,
due to errors in translation of the message,
or other reasons outside the developer's control.
For example, in Java, throwing an Exception is a common means of signaling an error.
However, `java.text.NumberFormat` provide both throwing and non-throwing
`parse` methods to allow developers to avoid a "fatal" throw of `ParseException`
(if the exception were uncaught).

As a MessageFormat implementer, I want to be able to signal errors in an idiomatic way
for my language and still be conformant with MF2 requirements.

```

aphillips:
In the 2024-06-24 call we agreed to accept this design with some changes. Waiting on changes/approvals to merge. @eemeli will then have an action to update the spec.

exploration/error-handling.md
```diff
@@ -0,0 +1,194 @@
...
+> occurred.
+> The word "signal" better accomodates more alternatives in the solution space
+> like those that only choose to indicate that an error occurred,
+> while still including those that additionally prefer to return the error
+> itself as an error object.
+> (By contrast, "return an error" implies that an error object will be thrown or
+> returned, and "emit an error" is ambiguous as to what is or isn't performed.)
+## Use Cases
+
+As a software developer, I want calls to message format to signal runtime errors
```
gibson042, line 61:
```suggestion
As a software developer, I want message formatting calls to signal runtime errors
```

exploration/error-handling.md
```diff
@@ -0,0 +1,194 @@
...
+in a manner consistent with my programming language/environment.
+I would like error signals to include diagnostic information that allows me to debug errors.
+
+As a software developer, I sometimes need to be able to emit a formatted message
+even if a runtime error has occurred.
+
+As a software developer, I sometimes want to avoid "fatal" error signals,
+such as might occur due to unconstrained inputs,
+due to errors in translation of the message,
+or other reasons outside the developer's control.
```
gibson042, line 71:
```suggestion
such as might occur due to unconstrained inputs,
errors in translation of the message,
or other reasons outside my control.
```

exploration/error-handling.md
```diff
@@ -0,0 +1,194 @@
...
+
+This solution requires implementations to be able to signal an error occurred,
+which can be accomplished in different ways. Ex:
+
+* an API that throws or returns an Error object when encountering an error
+* an API that includes a read/write `ErrorCode` argument 
+* two APIs, a permissive one that always returns a best-effort formatted result
+and a stricter one that throws or returns an Error object when encountering an error
+* an API that always returns a best-effort formatted result
+and a global `boolean` values
```
gibson042, line 99:
```suggestion
and sets a global `boolean` value
```

gibson042 (approved):
All suggestions are purely editorial.

aphillips (approved):
I think we should merge this (thank you @echeran!!!!) and I can do some cleanup when marking it approved. The main cleanup will be to point to the vote issue, notes, discussion, etc. so that someone reading this design doc will see how we got to spec text and to make the "proposed design" exactly match the text we're still working on in #816. Let's discuss in telecon (2024-07-29)

exploration/error-handling.md
```diff
@@ -0,0 +1,168 @@
+# Error Handling
+
+Status: **Proposed**
```
eemeli, line 3: I think this is good as is, but landing this as "proposed" would be fine as well.

```suggestion
Status: **Accepted**
```

exploration/error-handling.md
```diff
@@ -0,0 +1,168 @@
...
+	<summary>Metadata</summary>
+	<dl>
+		<dt>Contributors</dt>
+		<dd>@echeran</dd>
+		<dt>First proposed</dt>
+		<dd>2024-06-02</dd>
+        <dt>Issues</dt>
+		<dd><a href="https://github.com/unicode-org/message-format-wg/issues/782">#782</a></dd>
+		<dt>Pull Requests</dt>
+		<dd><a href="https://github.com/unicode-org/message-format-wg/pull/795">#795</a></dd>
```
eemeli, line 15:
```suggestion
		<dd><a href="https://github.com/unicode-org/message-format-wg/pull/795">#795</a></dd>
		<dd><a href="https://github.com/unicode-org/message-format-wg/pull/804">#804</a></dd>
```

eemeli (approved):

PR #811, opened 2024-06-24
[DESIGN] Implement changes to bidi to permit non-strict formation

aphillips:
This PR contains the design document for bidi handling at the syntax level. It contains a fairly comprehensive description of the problem space with examples.

### DIFF
--- a/exploration/bidi-usability.md
+++ b/exploration/bidi-usability.md
@@ -70,18 +70,23 @@ would enable tools, including translation tools, and users who are writing in RT
 to format a _message_ so that its plain-text representation and its function
 are unambiguous.
 
-The isolate controls are paired invisible control characters inserted around a portion of a string.
-The start of an isolate sequence is one of:
+The isolates are paired invisible characters inserted around a portion of a string.
+The start of an isolated sequence is one of:
 - U+2066 LEFT-TO-RIGHT ISOLATE (LRI)
 - U+2067 RIGHT-TO-LEFT ISOLATE (RLI)
 - U+2068 FIRST-STRONG ISOLATE (FSI)
 
-The end of an isolate sequence is U+2069 POP DIRECTIONAL ISOLATE (PDI).
+The end of an isolated sequence is U+2069 POP DIRECTIONAL ISOLATE (PDI).
 
-The characters inside an isolate sequence have the initial string (paragraph) direction
-corresponding to the starting control (LTR for LRI, RTL for RLI, auto for FSI).
-The isolate sequence is **isolated** from surrounding text.
-This means that the surrounding text treats it as-if the sequence were a single neutral character.
+The characters inside an isolated sequence have the initial string direction
+corresponding to the starting control (
+left-to-right for `LRI`, 
+right-to-left for `RLI`, 
+or <a href="https://www.w3.org/TR/i18n-glossary#auto-direction">auto</a> for `FSI`).
+The isolated sequence is **isolated** from surrounding text:
+it is processed using the Unicode Bidirectional Algorithm (UBA)
+separately from the rest of the string and
+the surrounding text treats the sequence as-if it were a single neutral character.
 
 > [!NOTE]
 > One of the side-effects of using `{`/`}` and `{{`/`}}` to delimit _expressions_
@@ -96,10 +101,21 @@ These include:
 - U+200F RIGHT-TO-LEFT MARK (RLM)
 - U+061C ARABIC LETTER MARK (ALM)
 
-These characters are invisible strongly-directional characters used in bidirectional
+These characters are invisible strongly-directional characters.
+They are used in bidirectional
 text to coerce certain directional behavior (usually to mark the end of 
 a sequence of characters that would otherwise be ambiguous or interact with
 neutrals or opposite direction runs in an unhelpful way).
+
+### Strictness
+
+We want the syntax to be somewhat permissive, particularly when it comes to paired isolates.
+The isolates and strongly-directional marks are invisble except in certain specialized editing environments.
+While users and tools should be strict about using well-formed isolate sequences,
+we don't want to invisble characters or whitespace to generate additional syntax errors except where
+absolutely necessary.
+Therefore, it should not be a syntax error if a user, editor, or tool fails to provide the
+opening or closing isolate.
 
 ## Use-Cases
 
@@ -253,7 +269,7 @@ Permit isolating bidi controls to be used on the **outside** of the following:
 - quoted literals
 - quoted patterns
 
-We permit any of the isolate starting controls (LRI, RLI, FSI) because we want to allow
+We permit any of the isolate starting characters (LRI, RLI, FSI) because we want to allow
 the user to set the base direction of a _literal_ or _pattern_ according to its respective 
 actual contents.
 
@@ -261,24 +277,24 @@ This would change the ABNF as follows:
 (Notice that this change includes a production `bidi` described further down
 in this document)
 ```abnf
-literal        = ( open-isolate (quoted / (unquoted [bidi])) close-isolate)
-               / (quoted / (unquoted [bidi]))
-quoted-pattern = ( open-isolate "{{" pattern "}}" close-isolate)
-               / ("{{" pattern "}}")
+literal        = [open-isolate] (quoted / (unquoted [bidi])) [close-isolate]
+quoted-pattern = [open-isolate] "{{" pattern "}}" [close-isolate]
 
 open-isolate   = %x2066-2068
 close-isolate  = %x2069
 ```
 
 > [!IMPORTANT]
-> The isolating controls go on the **_outside_** of the various _literal_ and _pattern_
+> The isolating characters go on the **_outside_** of the various _literal_ and _pattern_
 > productions because characters on the **_inside_** of these are part of the _literal_'s
 > or _pattern_'s textual content.
-> We need to allow users to include bidi controls in the output of MF2.
+> We need to allow users to include bidi characters, including isolates and strongly directional marks in the output of MF2.
 
-Permit **left-to-right** isolating bidi controls (`U+2066`...`U+2069`) to be used **immediately inside** the following:
+Permit **left-to-right** isolates (`U+2066`...`U+2069`) to be used **immediately inside** the following:
 - expressions
 - markup
+
+Permit isolates around any token inside of an expression or markup.
 
 We only permit the LTR isolates because the contents of an _expression_
 or _markup_ must be laid out left-to-right.
@@ -287,15 +303,12 @@ directional marks to ensure correct display.
 
 This would change the ABNF as follows (assuming the above changes are also incorporated):
 ```abnf
-expression            = "{" LRI (literal-expression / variable-expression / annotation-expression) close-isolate "}"
-                      / "{" (literal-expression / variable-expression / annotation-expression) "}"
+expression            = "{" [LRI] (literal-expression / variable-expression / annotation-expression) [close-isolate] "}"
 literal-expression    = [s] literal [s annotation] *(s attribute) [s]
 variable-expression   = [s] variable [s annotation] *(s attribute) [s]
 annotation-expression = [s] annotation *(s attribute) [s]
-markup = "{" [s] "#" identifier *(s option) *(s attribute) [s] ["/"] "}"                    ; open and standalone
-       / "{" [s] "/" identifier *(s option) *(s attribute) [s] "}"                          ; close
-       / "{" LRI [s] "#" identifier *(s option) *(s attribute) [s] ["/"] close-isolate "}"  ; open and standalone
-       / "{" LRI [s] "/" identifier *(s option) *(s attribute) [s] close-isolate "}"        ; close
+markup = "{" [LRI] [s] "#" identifier *(s option) *(s attribute) [s] ["/"] [close-isolate] "}"  ; open and standalone
+       / "{" [LRI] [s] "/" identifier *(s option) *(s attribute) [s] [close-isolate] "}"        ; close
 LRI = %x2066
 ```
 
--- a/exploration/bidi-usability.md
+++ b/exploration/bidi-usability.md
@@ -107,22 +107,19 @@ text to coerce certain directional behavior (usually to mark the end of
 a sequence of characters that would otherwise be ambiguous or interact with
 neutrals or opposite direction runs in an unhelpful way).
 
-### Strictness
+### Strictness and Abuse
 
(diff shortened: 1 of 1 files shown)


exploration/bidi-usability.md
```diff
@@ -253,49 +269,46 @@ Permit isolating bidi controls to be used on the **outside** of the following:
...
 This would change the ABNF as follows:
 (Notice that this change includes a production `bidi` described further down
 in this document)
 ```abnf
-literal        = ( open-isolate (quoted / (unquoted [bidi])) close-isolate)
-               / (quoted / (unquoted [bidi]))
-quoted-pattern = ( open-isolate "{{" pattern "}}" close-isolate)
-               / ("{{" pattern "}}")
+literal        = [open-isolate] (quoted / (unquoted [bidi])) [close-isolate]
+quoted-pattern = [open-isolate] "{{" pattern "}}" [close-isolate]
```
eemeli, line 281: This adds LRI, RLI & FSI to the content we need to look past at message start to determine if it's complex or not. That should be specifically called out.

exploration/bidi-usability.md
```diff
@@ -348,47 +407,50 @@ the results or debug what is wrong with their messages.
...
+**Pros**
+- Avoids problems with syntax errors that users and tools might find difficult to debug.
+- Effective if used carefully.
+- Addresses need to comply with UAX#31
 
-literal-expression    = "{" [LRI] [s] literal [s annotation] *(s attribute) [s] [close-isolate] "}"
-variable-expression   = "{" [LRI] [s] variable [s annotation] *(s attribute) [s] [close-isolate] "}"
-annotation-expression = "{" [LRI] [s] annotation *(s attribute) [s] [close-isolate] "}"
+**Cons**
+- Can be used irresponsibly, including enabling some Trojan Source cases (UAX#55)
```
macchiati, line 440:
```suggestion
    - However, linters could support "responsible" use of bidi characters. (It is likely that many implementations will end up needing to support linting, such as to verify the right number and datatype of input parameters, etc.
```

aphillips: This would be better as a pro perhaps? "Irresponsibly" might be the wrong word in the 'con', although it smacks of code attacks, which is the idea. This design enables the right thing by also enabling many wrong things. On the other hand, a lot of the time we'll be lucky if tools implement anything or users bother to insert the isolates/marks.

macchiati:
It is maybe a Pro (but more of a possibly-mitigating factor for a Con)

exploration/bidi-usability.md
aphillips: See most recent edits, which incorporate your suggestion into the 'con' text as well as adding more prose to the design option such that this is (hopefully) clear. I also added one more design, which tries to split-the-difference

exploration/bidi-usability.md
```diff
@@ -96,11 +104,26 @@ These include:
...
+These characters are invisible strongly-directional characters.
+They are used in bidirectional
 text to coerce certain directional behavior (usually to mark the end of 
 a sequence of characters that would otherwise be ambiguous or interact with
 neutrals or opposite direction runs in an unhelpful way).
 
+### Strictness and Abuse
+
+We want the syntax to be somewhat permissive, particularly when it comes to paired isolates.
+The isolates and strongly-directional marks are invisble except in certain specialized editing environments.
```
catamorphism, line 116:
```suggestion
The isolates and strongly-directional marks are invisible except in certain specialized editing environments.
```

exploration/bidi-usability.md
```diff
@@ -230,72 +259,90 @@ The workaround in #763 was to permit these characters _before_ or _after_ whites
...
-- _Placeholders_ and _markup_ are isolated (treated as neutrals) so that they appear
+Preferring LTR display is not the disadvantage to right-to-left languages that it might first appear:
+- Bidi inside of _patterns_ works normally (we go to great lengths to make the interior
+  of _patterns_ work as plain text)
+- _Placeholders_ and _markup_ can be isolated (treated as neutrals) so that they appear
   in the correct location in an RTL _pattern_
 - _Expressions_ use isolates and directional marks to display internal tokens in the
   correct order and without spillover effects
+- The syntax uses paired enclosing marks that the Unicode Bidirectional Algorithm pairs
+  for shaping purposes and these offer a poor person's form of isolation.
```
catamorphism, line 292: I can't tell whether to read this as "paired enclosing marks (that the Unicode Bidirectional Algorithm pairs for shaping purposes)" or "uses paired enclosing marks (that the Unicode Bidirectional Algorithm pairs) for shaping purposes."

exploration/bidi-usability.md
```diff
@@ -230,72 +259,90 @@ The workaround in #763 was to permit these characters _before_ or _after_ whites
...
-> The isolating controls go on the **_outside_** of the various _literal_ and _pattern_
+> The isolating characters go on the **_outside_** of the various _literal_ and _pattern_
 > productions because characters on the **_inside_** of these are part of the _literal_'s
 > or _pattern_'s textual content.
-> We need to allow users to include bidi controls in the output of MF2.
+> We need to allow users to include bidi characters, including isolates and strongly directional marks
+> in the output of MF2.
 
-Permit **left-to-right** isolating bidi controls (`U+2066`...`U+2069`) to be used **immediately inside** the following:
+Permit **left-to-right** isolates (`U+2066` and `U+2069`) to be used **immediately inside** the following:
```
catamorphism, line 327: Should this be part of a bullet list, along with the other "Permit..." imperatives that follow?

exploration/bidi-usability.md
```diff
@@ -348,47 +407,51 @@ the results or debug what is wrong with their messages.
...
+**Cons**
+- Syntax does not prevent poor display outcomes, including enabling some Trojan Source cases (UAX#55);
+  note that tooling or linting can help ameliorate these issues.
 
-markup = "{" [LRI] [s] "#" identifier *(s option) *(s attribute) [s] ["/"] [close-isolate] "}"
-       / "{" [LRI] [s] "/" identifier *(s option) *(s attribute) [s] [close-isolate] "}"
+### Strict isolation all the time
+
+Apply bidi isolates in a strict way.
+The main differences to the proposed solution is:
```
catamorphism, line 445: I only see one difference?

exploration/bidi-usability.md
aphillips: It's the former.

```suggestion
- The syntax uses enclosing marks (specifically curly brackets) which the Unicode Bidirectional Algorithm
  pairs up for shaping purposes, resulting in a weak form of isolation in the syntax itself.
```

exploration/bidi-usability.md
aphillips: The paragraph on line 331 should probably move below the paragraph "We only permit..." because it has a different meaning and refers to all isolates. Not sure about using a bullet list?

exploration/bidi-usability.md
aphillips: I should do the TODO, which makes it clearer. I'll do one here in the comment for clarity and then go back and fix the PR.

The current design has `expression` thusly:

```abnf
expression = "{" [LRI] (literal-expression / variable-expression / annotation-expression) [close-isolate] "}"
```

This alternative would turn that into:

```abnf
expression = "{" (literal-expression / variable-expression / annotation-expression) "}"
           / "{" LRI (literal-expression / variable-expression / annotation-expression) close-isolate "}"
```

In this formulation, you cannot have unpaired opening (or closing) isolates without a syntax error, nor can you have multiples of open or close.

Rinse and repeat for markup, option, attribute, and literals.

Make sense?

exploration/bidi-usability.md
```diff
@@ -348,47 +407,51 @@ the results or debug what is wrong with their messages.
...
+// TODO put ABNF here
 
-s = 1*( SP / HTAB / CR / LF [LRI] / %x3000 )
-LRI = %x2066
-open-isolate  = %x2066-2068
-close-isolate = %x2069
-```
 
 Isolating rather than marking `name` helps ensure
 that its directionality does not spill over to adjoining syntax.
```
eemeli, line 457: What follows was a part of the "Loose Isolation" alternative, parts of which are in this PR included in the proposed design; this is separate from the "Strict isolation all the time" alternative, and should have its own title as well as the ABNF that was previously included.

````suggestion

### Isolate `name` rather than `unquoted-literal`

Isolating rather than marking `name` helps ensure
that its directionality does not spill over to adjoining syntax.

The following replaces the proposed design's changes to `literal` and the `[bidi]` additions to
`variable-expression`, `function`, `option`, `attribute`, `markup`, and `ns-separator`:
```abnf
name = [open-isolate] name-start *name-char [close-isolate]
quoted-literal = [open-isolate] "|" *(quoted-char / quoted-escape) "|" [close-isolate]
```

````

**Review by eemeli — COMMENTED** [review]
I would be interested in hearing from @catamorphism, @mihnita, @lucacasonato and any other implementers about whether they would have any concerns about including bidi isolation characters within the `name` ABNF rule, as proposed in the alternative I'm naming here as "Isolate `name` rather than `unquoted-literal`".

This would not change the parsed meaning of a `name` in any way, but only include the bidi isolation or mark characters within the ABNF production so that they do not need to be mentioned in all the places where a `name` is used, as is the case in the current proposed solution.

@aphillips and I have discussed this, and as our opinions on this diverge, getting comments from other people who'll need to adjust their parsers due to the changes here would be helpful.

**Comment by lucacasonato** [comment]
Prefix: I am very unfamiliar with bidi - the linked introduction at the start of this doc was a great help. So thanks for linking that.

My impression is that the proposed ABNF seems reasonable. I do not think this would make my parser significantly more complex.

I make this statement on the following assumptions that I have based on my understanding of this document. Please correct them if they are wrong.
- isolation/bidi characters are in no way stored in an AST / data model
- isolation/bidi character placement is not meant to round-trip through parse-serialize
- a serializer can make a fully automatic determination based just on the AST / data model as to where what isolation/bidi chars need to be placed in the output
- a parser does not need to validate the correct pairing of bidi chars

---

I would be interested to understand further how likely it is that by editing a message that contains isolation/bidi chars, a user could produce a message that ends up being either:
- syntactically invalid
- has isolation/bidi chars in the middle of a pattern or quoted literal, only because they backspace / copy pasted etc

I guess what I am asking is how text (and code) editors usually deal with isolation/bidi chars: if I have a message that has an expresion containing a PDI just before the `}`, and I remove (backspace) the `}`, then manually move my cursor to the column trailing the last visible character in the expression, and type `}`, would the PDI now be in the pattern or would it be gone?

**Comment by catamorphism** [comment]
> I would be interested in hearing from @catamorphism, @mihnita, @lucacasonato and any other implementers about whether they would have any concerns about including bidi isolation characters within the `name` ABNF rule, as proposed in the alternative I'm naming here as "Isolate `name` rather than `unquoted-literal`".
> 
> This would not change the parsed meaning of a `name` in any way, but only include the bidi isolation or mark characters within the ABNF production so that they do not need to be mentioned in all the places where a `name` is used, as is the case in the current proposed solution.
> 
> @aphillips and I have discussed this, and as our opinions on this diverge, getting comments from other people who'll need to adjust their parsers due to the changes here would be helpful.

I don't have concerns about this... or maybe I will once I try implementing it, but I can't think of anything right now.

**Comment by aphillips** [comment]
@catamorphism Thanks. My disagreement with @eemeli about including isolates and directional marks inside names (that are not considered part of the name) is that this would require parsers to process each name to remove these characters. I think this is relatively high impact for very little reward.

We are discussing in another thread potentially doing NFC normalization for comparison purposes, but this doesn't require walking the buffer (and can be fast-checked and otherwise optimized). Prohibiting isolates and strong marks in names will permit a few corner cases in which the namespace and name display awkwardly but no cases in which false matches are produced. Users have a much better workaround ("choose unidirectional names") available.

Note: I would like to merge this PR and then have a technical discussion of the design.

**Comment by eemeli** [comment]
> Note: I would like to merge this PR and then have a technical discussion of the design.

I'd be fine with that, provided that the suggestion from https://github.com/unicode-org/message-format-wg/pull/811#discussion_r1691326748 or something like it is included to document the remaining aspect of this that we disagree on.

**Review by eemeli — APPROVED** [review]

**Comment by aphillips** [comment]
@eemeli I inserted the comment with edits and additions. Please have a look and see that I've captured the pros/cons to your satisfaction or if you have additions.

--

For the record, I think I'm leaning towards the "hybrid approaches" design. It would mean that MF2 messages might have spillover effects, but would encourage implementations to (re)serialize messages in ways that eliminate such effects. Looking forward to merge-and-discuss.

File: exploration/bidi-usability.md (hunk 4 above)
Comment (aphillips) [reply exploration/bidi-usability.md:292]: Done.

File: exploration/bidi-usability.md (hunk 5 above)
Comment (aphillips) [reply exploration/bidi-usability.md:327]: Done. Created the bullet list and reorganized.

File: exploration/bidi-usability.md [hunk 8]
```diff
@@ -348,69 +412,112 @@ the results or debug what is wrong with their messages.
...
+**Cons**
+- Implementations need to remove isolates from the `name` token before comparing
+  the value to other values (such as comparing `function` or `variable` names).
+  Because of namespacing, this requires looking _inside_ the token.
+- Implementations might need to insert isolates when generating names upon serialization.
+  The current data model does not separate `namespace` and `name`,
+  so this might be more complicated.
+- `unquoted-literal` values appear as keys, as operands, and as option values.
+  If not isolated, these can cause spillover effects, so we might need both `name`
+  and `unquoted-literal` isolation.
```
eemeli, line 516: Given how `unquoted-literal` is defined as https://github.com/unicode-org/message-format-wg/blob/01f2880bd709653ad850dc746f4a72fe1133c064/spec/message.abnf#L48 and `number-literal` only contains LTR characters https://github.com/unicode-org/message-format-wg/blob/01f2880bd709653ad850dc746f4a72fe1133c064/spec/message.abnf#L50 allowing `name` isolation should cover all the `unquoted-literal` cases that may need isolation.

eemeli (approved):
This looks fine to merge. One line comment below, but it's not a blocker.

exploration/bidi-usability.md
aphillips: Ur... this is something I don't like, since isolates are "not included" in the `name`, but are baked into the production. I'd rather do more with the ABNF to keep them separate (the isolates and directional characters are in the same places, but part of the surrounding goo instead of being part of the name-like token).

Note that `number-literal` contains only neutral and weakly directional characters (except for `e`, which is strongly LTR). Note that the leading `-` can switch sides in an RTL context (unless isolated or protected with LRM)

PR #812, opened 2024-07-01
Allow trailing whitespace for complex messages

eemeli:
Fix for a part of #809, CC @lucacasonato

When parsing a complex message, we should be lenient about trailing whitespace, and silently ignore it. It has no meaning, and should not be considered an error.

None of the parser complexity concerns around leading whitespace apply here.

### DIFF
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -168,9 +168,10 @@ and consists of:
 
 1. an optional list of _declarations_, followed by
 2. a _complex body_
+3. optional trailing whitespace
 
 ```abnf
-complex-message = *(declaration [s]) complex-body
+complex-message = *(declaration [s]) complex-body [s]
 ```
 
 ### Declarations
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -168,7 +168,6 @@ and consists of:
 
 1. an optional list of _declarations_, followed by
 2. a _complex body_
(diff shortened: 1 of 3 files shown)


lucacasonato (commented):
This is good, but I think we should also do leading whitespace. From my implementation experience, this is a trivial to implement and can be implemented to be as fast as the current spec behavior.

eemeli:
> This is good, but I think we should also do leading whitespace. From my implementation experience, this is a trivial to implement and can be implemented to be as fast as the current spec behavior.

I completely agree, but I'd like to ensure that we _at least_ get trailing whitespace rather than having this blocked by implementation or other concerns about mode detection with leading whitespace.

spec/syntax.md
```diff
@@ -168,9 +168,10 @@ and consists of:
 
 1. an optional list of _declarations_, followed by
 2. a _complex body_
+3. optional trailing whitespace
```
aphillips, line 171: I think this is distracting and unnecessary and would leave it out.

If we go on to allow preceding whitespace, we'd handle it differently.

aphillips (approved):

spec/syntax.md
eemeli:
```suggestion
```

mihnita (approved):

PR #814, opened 2024-07-01
Define function composition for date/time values

eemeli:
This is a follow-up to #798 and the conversations around that.

The intent with this PR is to define what happens when a `:datetime`, `:date`, or `:time` annotated value is used as an operand or option value. It might be best if we didn't initially nitpick on the exact verbiage used in this PR, but rather the overall behaviour: how operand options are merged with expression options, and what values are presented to the consumer.

Given how we've explicitly left out e.g. `calendar` and `timeZone`, the language needs to allow for those (and other implementation-defined options) to pass through while being far more opinionated about the style & field options.


spec/registry.md
```diff
@@ -944,6 +970,21 @@ The function `:time` has these _options_:
...
+#### Composition
+
+When an _operand_ or an _option_ value uses a _variable_ with a _declaration_
+using an _expression_ with a `:time` _annotation_,
+its resolved value contains only the implementation-defined date/time value
+of the _operand_ of the annotated _expression_ (with no date value),
+together with the resolved options values.
+
+Using a `:time` annotated value as the _operand_ of `:datetime` or `:date`
+produces a _Bad Operand_ error as it contains no date value.
```
aphillips, line 987: This seems incorrect. _Some_ Temporal values contain no date value (although most APIs can promote a time-only value to have a `0` date, e.g. 1970-01-01 epochal date). Annotating with the `:time` function shouldn't delete date information from an existing operand that uses a `Date` or any of the zoo of Temporal types. It certainly wouldn't affect the ability to use an incremental time value ("Instant", millis-since-epoch, `time_t` etc.).

I think making this paragraph operable would produce a lot of surprise for users.

```
.local $now = {$date :time timeZone=|America/Los_Angeles|}
{{It is now {$now} on {$now :date dateStyle=short}}}
```

spec/registry.md
```diff
@@ -921,6 +934,19 @@ The function `:date` has these _options_:
   - `medium`
   - `short` (default)
 
+If the _operand_ of the _expression_ is an implementation-defined date/time type,
+it can include other option values.
+Any _operand_ option values matching the `:datetime` _style options_ or _field options_ are ignored,
+as is any `style` option.
```
aphillips, line 940: I think it would be better to list the options that are *are* transitive rather than those that are not. I think the default probably should be that the operand and option list are not transitive, so that functions carefully list those that are.

eemeli: If we do that, then how do we avoid being proscriptive of the fields that an implementation might want to include in describing the locale options that are relevant to the date's formatting? Or would doing so be intentional and somehow beneficial?

spec/registry.md
aphillips: I think you're cutting corners, though.

If I pass in a `java.time.ZonedDateTime` value, it is weird to _force_ the `:time` annotation to turn that into an `OffsetTime` or `LocalTime` just because the formatter doesn't want the date parts. It's even sometimes bad, because, in many time zones at certain times of the year, it affects the presentation of the value. Consider `2024-03-10T10:30:00Z` in the `America/Los_Angeles` time zone (the time would normally be `2:30 a.m.` but the daylight savings transition makes it display as `3:30 a.m.`)

I think there is some margin around items like `{4.2 :integer}` because the operand begins life as a literal and we give some leeway for implementations to use an integer type to store the parsed value. 

Date/time values are more complicated, though, because there are a myriad of types (and variations in and between different programming languages). If I have `{|2024-07-06T13:14:15.666Z[America/Los_Angeles]| :time}` in my pattern, it is kind of weird to only parse out the date part. If my underlying time value is an incremental time type (think `java.util.Date` or `time_t`), the string parses to a specific number of clock ticks before/after the epoch date (1 Jan 1970). I think, in this case, it becomes an implementation-specific problem.

What I'm saying is that we should not **_require_** information to be removed from an operand, but we possibly should **_permit_** implementations to do so in very narrow ways.

spec/registry.md
aphillips: I think probably there are three things:
- options that MUST be passed (we want to list these)
- options that MUST NOT be passed (we want to list these)
- options that MAY be passed (implementation defined)

The last category mostly consists of (hopefully namespaced) implementation-specific options. Future additions to the standard fall into the first two categories.

So I guess I'm saying I retract my original comment, but am looking for us to be more formal about how we implement this. Every option should clearly say which of the three (MUST/MUST NOT/MAY) it is, with MAY being strongly discouraged (i.e. non-existant) in the default registry and options being considered for RGI.

spec/registry.md
eemeli: > If I pass in a `java.time.ZonedDateTime` value, it is weird to _force_ the `:time` annotation to turn that into an `OffsetTime` or `LocalTime` just because the formatter doesn't want the date parts.

Consider the value that in `.input {$t :time} .local {:foo bar=$t}` is passed to the `:foo` handler for its `bar` option: Does it hold a ZonedDateTime, OffsetTime, LocalTime, or potentially any of these? Or do we require the initial value of the `t` argument to always be a ZonedDateTime, even though we state we're only interested in its time? If we don't restrict the output to only be a time, 

> I think there is some margin around items like `{4.2 :integer}` because the operand begins life as a literal and we give some leeway for implementations to use an integer type to store the parsed value.

Does the type of the operand matter for this, though? I at least would presume that in the following `$a` and `$c` would be indistinguishable from each other, or from $d if the formatter is called with arguments like `{ d: 4.2 }`:

```
.local $a = {4.2 :integer}
.local $b = {4.2 :number}
.local $c = {$b :integer}
.input {$d :integer}
```

> What I'm saying is that we should not **_require_** information to be removed from an operand, but we possibly should **_permit_** implementations to do so in very narrow ways.

For the general case, I agree that implementations and custom functions must have the freedom to present the resolved value of their operand as they see fit, e.g. turning a string into a number or a datetime.

Specifically for the default registry, we do need to say what happens when you use a `:time` annotated value as a `:date` operand. If we do permit implementations to have narrowed the value to no longer carry the date (leaving perhaps only an OffsetTime or its equivalent), then I think we either need to say that this is an error, or that it leads to undefined behaviour. I don't see any benefit from picking the latter over the former.

spec/registry.md
```diff
@@ -796,6 +796,11 @@ output.
 If both are specified, a _Bad Option_ error MUST be emitted
 and a _fallback value_ used as the resolved value of the _expression_.
 
+If the _operand_ of the _expression_ is an implementation-defined date/time type,
+it can include _style options_, _field options_, or other option values.
```
stasm, line 800: I think we should say `MUST` here, or at least `SHOULD`, and be specific about which options are required to be present (all of them?). The purpose here is be to allow custom function authors to extend MF2 and be able to rely on the implementations exposing these options.

spec/registry.md
```diff
@@ -897,7 +902,15 @@ are encouraged to track development of these options during Tech Preview:
    - valid [Unicode Number System Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeNumberSystemIdentifier)
 - `timeZone` (default is system default time zone or UTC)
   - valid identifier per [BCP175](https://www.rfc-editor.org/rfc/rfc6557)
- 
+
+#### Composition
+
+When an _operand_ or an _option_ value uses a _variable_ with a _declaration_
+using an _expression_ with a `:datetime` _annotation_,
```
stasm, line 909: I find this sentence difficult to parse. Could we perhaps simplify it like so?

```suggestion
When an _operand_ or an _option_ value uses a _variable_ annotated,
directly or indirectly, by a `:datetime` _annotation_,
```

(Or: s/annotated/decorated/?)

spec/registry.md
```diff
@@ -944,6 +970,21 @@ The function `:time` has these _options_:
...
 
+If the _operand_ of the _expression_ is an implementation-defined date/time type,
+it can include other option values.
+Any _operand_ option values matching the `:datetime` _style options_ or _field options_ are ignored,
+as is any `style` option.
+
+#### Composition
+
+When an _operand_ or an _option_ value uses a _variable_ with a _declaration_
+using an _expression_ with a `:time` _annotation_,
```
eemeli, line 981:
```suggestion
When an _operand_ or an _option_ value uses a _variable_ annotated,
directly or indirectly, by a `:time` _annotation_,
```

spec/registry.md
aphillips: I don't think we can `MUST` or `SHOULD` that implementation-defined date/time types contain options. The text here needs to be implementer focused:

```suggestion
Implementations MAY permit callers to supply _style options_, _field options_, or other option values
as part of implementation-defined datetime types or values.
```

I use `MAY` here because `SHOULD` means that every implementation must do so unless they have a good reason not to. It will be difficult to test.

spec/registry.md
eemeli: To bring this thread to some conclusion, would this change be sufficient to the lines we're nominally talking about?

```suggestion
Using a `:time` annotated value as the _operand_ of `:datetime` or `:date`
produces a _Bad Operand_ error as it is not guaranteed to contain a date value.
```

aphillips: No. This phrasing _requires_ an error, when the actual data might not result in one. Why is the `:time` annotation so important to require it to be an error? It isn't the annotation that is the source of the error. Instead, `:date` or maybe `:datetime` will have an internal error, which they surface (probably as `bad-operand`), when the type provided is not one of the implementation-defined ones required by those functions.

```suggestion
Using a `:time` annotated value as the _operand_ of `:datetime` or `:date`
might result in a _Bad Operand_ error if the underlying implementation-defined type
does not contain appropriate date information.
```

What is the reason we have to require an error? Across every implementation?

eemeli: To remove unnecessary variance across implementations. There is no need for the date part of a `:time` value to be re-used when it's not clear what its date value is. The exact same results can and should be achieved with `:datetime`.

aphillips: Why does the `:time` function have to _change_ the operand value if the operand value contains date information? What happened to immutability? 

There are lots of cases where I might use `:time` (or `:date`) to annotate a time value that is used repeatedly in a set of variants, and still want both the other parts (date or time) later.

```
.input {$foo :date timezone=|America/Los_Angeles| dateStyle=medium}
{{Your delivery arrives on {$foo} by {$foo :time hour=numeric}}}
```

Some implementations might not even be able to do what you want. A lot of timekeeping is in terms of epoch incremental times, particularly in smaller languages.

eemeli: Immutability is still being fully maintained by e.g `$foo` in the above only ever having one value, which is set by the `foo` argument and the annotation of the `.input` expression.

For the two lines this thread is on, it's not really relevant whether the `:time` annotated value is of a different type than its operand, but that it might not contain the date parts, especially if its operand never had them.

I'd be willing to consider making the date-stripping optional, but that would not change still preferring an error here.

Your example could also be expressed as
```
.input {$foo :datetime timezone=|America/Los_Angeles|}
{{Your delivery arrives on {$foo :date style=medium} by {$foo :datetime hour=numeric}}}
```
though from your use of `:datetime timezone` and  `:time hour` I suspect you'd prefer we open up a separate discussion about the options on date/time annotations, but that should be separate from this thread.

spec/registry.md
```diff
@@ -907,7 +907,7 @@ are encouraged to track development of these options during Tech Preview:
 
 When an _operand_ or an _option_ value uses a _variable_ annotated,
 directly or indirectly, by a `:datetime` _annotation_,
-its resolved value contains only the implementation-defined datetime value
+its resolved value contains the implementation-defined datetime value
```
aphillips, line 910: as with the `:number` PR, I think "the implementation-defined" should consistently be "an implementation-defined" and for the same reason.

This one says "datetime", the others say "date/time"

spec/registry.md
```diff
@@ -943,7 +943,7 @@ as is any `style` option.
 
 When an _operand_ or an _option_ value uses a _variable_ annotated,
 directly or indirectly, by a `:date` _annotation_,
-its resolved value contains only the implementation-defined date/time value
+its resolved value contains the implementation-defined date/time value
 of the _operand_ of the annotated _expression_ (with no time value),
```
aphillips, line 947: I still do not agree with "(with no time value)"

spec/registry.md
```diff
@@ -979,7 +979,7 @@ as is any `style` option.
 
 When an _operand_ or an _option_ value uses a _variable_ annotated,
 directly or indirectly, by a `:time` _annotation_,
-its resolved value contains only the implementation-defined date/time value
+its resolved value contains the implementation-defined date/time value
 of the _operand_ of the annotated _expression_ (with no date value),
```
aphillips, line 983: I still do not agree with "(with no date value)"

aphillips (changes requested):

spec/registry.md
aphillips: I don't prefer an error here. Or, more specifically, I prefer _allowing_ an error when the user does something that the implementation finds illegal (`{$aLocalTime :date year=numeric} has no year`) but not when the user does something stupid looking (`.input {$instant :time}{{It is {$instant} in the year {$instant :date year=numeric}}}`). Working with datetime values is already complicated.

I agree that we should have a separate discussion about the options (we _really_ need to resolve the unincluded options... I should make a PR for that, eh?).

spec/registry.md
eemeli:
```suggestion
its resolved value contains an implementation-defined date/time value
```

spec/registry.md
```diff
@@ -897,7 +902,15 @@ are encouraged to track development of these options during Tech Preview:
    - valid [Unicode Number System Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeNumberSystemIdentifier)
 - `timeZone` (default is system default time zone or UTC)
   - valid identifier per [BCP175](https://www.rfc-editor.org/rfc/rfc6557)
- 
+
+#### Composition
+
+When an _operand_ or an _option_ value uses a _variable_ annotated,
+directly or indirectly, by a `:datetime` _annotation_,
+its resolved value contains an implementation-defined date/time value
```
eemeli, line 910:
```suggestion
its _resolved value_ contains an implementation-defined date/time value
```

spec/registry.md
```diff
@@ -921,6 +934,19 @@ The function `:date` has these _options_:
...
+If the _operand_ of the _expression_ is an implementation-defined date/time type,
+it can include other option values.
+Any _operand_ option values matching the `:datetime` _style options_ or _field options_ are ignored,
+as is any `style` option.
+
+#### Composition
+
+When an _operand_ or an _option_ value uses a _variable_ annotated,
+directly or indirectly, by a `:date` _annotation_,
+its resolved value contains the implementation-defined date/time value
```
eemeli, line 946:
```suggestion
its _resolved value_ contains the implementation-defined date/time value
```

spec/registry.md
```diff
@@ -944,6 +970,21 @@ The function `:time` has these _options_:
...
+If the _operand_ of the _expression_ is an implementation-defined date/time type,
+it can include other option values.
+Any _operand_ option values matching the `:datetime` _style options_ or _field options_ are ignored,
+as is any `style` option.
+
+#### Composition
+
+When an _operand_ or an _option_ value uses a _variable_ annotated,
+directly or indirectly, by a `:time` _annotation_,
+its resolved value contains the implementation-defined date/time value
```
eemeli, line 982:
```suggestion
its _resolved value_ contains the implementation-defined date/time value
```

eemeli:
Following Monday's discussion, I've updated the PR to make the function composition of expressions with `:date` or `:time` annotations implementation-defined, and explicitly call out the possibility that this can lead to emitting an error.

aphillips (approved):

PR #816, opened 2024-07-08
Refine error handling text

eemeli:
Updates to the spec that implement the design doc introduced in #804.

Filed initially as a draft pending the [previously discussed](https://github.com/unicode-org/message-format-wg/blob/main/meetings/2024/notes-2024-06-24.md#804) updates to the design doc, but happy to accept reviews and recommendations for the text already.

### DIFF
--- a/spec/errors.md
+++ b/spec/errors.md
@@ -24,10 +24,13 @@ or _Message Function Errors_ in _expressions_ that are not otherwise used by the
 such as _placeholders_ in unselected _patterns_
 or _declarations_ that are never referenced during _formatting_.
 
-When formatting a message with one or more errors:
-- An implementation MUST provide a way for a user to be informed
-  of the name of at least one of the errors,
-  either directly or via an identifying error code.
+When formatting a _message_ with one or more errors:
+- An implementation MUST provide a mechanism to discover and identify
+  at least one of the errors. 
+  The exact form of error signaling is implementation defined.
+  Some examples include throwing an exception,
+  returning an error code, 
+  or providing a function or method for enumerating any errors.
 - An implementation MUST provide a way for a message with one or more
   _Resolution Errors_ or _Message Function Errors_ to be formatted
   using a fallback representation.
--- a/spec/errors.md
+++ b/spec/errors.md
@@ -34,7 +34,9 @@ or providing a function or method for enumerating any errors.
 
 For all _messages_ without _Syntax Errors_ or _Data Model Errors_,
 an implementation MUST enable a user to get a formatted result.
-This result MAY include formatted _fallback values_.
+The formatted result might include _fallback values_ 
+such as when a _placeholder_'s _expression_ produced an error
+during formatting.
 
 The two above requirements MAY be fulfilled by a single formatting method,
 or separately by more than one such method.
--- a/spec/errors.md
+++ b/spec/errors.md
@@ -24,9 +24,16 @@ or _Message Function Errors_ in _expressions_ that are not otherwise used by the
 such as _placeholders_ in unselected _patterns_
 or _declarations_ that are never referenced during _formatting_.
 
-In all cases, when encountering a runtime error,
-a message formatter MUST provide some representation of the message.
-An informative error or errors MUST also be separately provided.
+When formatting a message with one or more errors:
+- An implementation MUST provide a way for a user to be informed
+  of the name of at least one of the errors,
+  either directly or via an identifying error code.
+- An implementation MUST provide a way for a message with one or more
+  _Resolution Errors_ or _Message Function Errors_ to be formatted
+  using a fallback representation.
+
+The two above requirements MAY be fulfilled by a single formatting method,
+or separately by more than one such method.
 
 When a message contains more than one error,
 or contains some error which leads to further errors,
--- a/spec/errors.md
+++ b/spec/errors.md
@@ -31,9 +31,9 @@ When formatting a _message_ with one or more errors:
   Some examples include throwing an exception,
   returning an error code, 
   or providing a function or method for enumerating any errors.
-- An implementation MUST provide a way for a message with one or more
-  _Resolution Errors_ or _Message Function Errors_ to be formatted
-  using a fallback representation.
+- For all messages without _Syntax Errors_ or _Data Model Errors_,
+  an implementation MUST enable a user to get a formatted result.
+  This result MAY include formatted _fallback values_.
 
 The two above requirements MAY be fulfilled by a single formatting method,
 or separately by more than one such method.
--- a/spec/errors.md
+++ b/spec/errors.md
@@ -24,16 +24,17 @@ or _Message Function Errors_ in _expressions_ that are not otherwise used by the
 such as _placeholders_ in unselected _patterns_
 or _declarations_ that are never referenced during _formatting_.
 
-When formatting a _message_ with one or more errors:
-- An implementation MUST provide a mechanism to discover and identify
-  at least one of the errors. 
-  The exact form of error signaling is implementation defined.
-  Some examples include throwing an exception,
-  returning an error code, 
-  or providing a function or method for enumerating any errors.
-- For all messages without _Syntax Errors_ or _Data Model Errors_,
-  an implementation MUST enable a user to get a formatted result.
-  This result MAY include formatted _fallback values_.
+When formatting a _message_ with one or more errors,
+an implementation MUST provide a mechanism to discover and identify
+at least one of the errors. 
+The exact form of error signaling is implementation defined.
+Some examples include throwing an exception,
+returning an error code, 
+or providing a function or method for enumerating any errors.
+
+For all _messages_ without _Syntax Errors_ or _Data Model Errors_,
+an implementation MUST enable a user to get a formatted result.
+This result MAY include formatted _fallback values_.
 
 The two above requirements MAY be fulfilled by a single formatting method,
 or separately by more than one such method.
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -735,9 +735,8 @@ each _text_ and _placeholder_ part of the selected _pattern_ is resolved and for
 
 Resolved values cannot always be formatted by a given implementation.
 When such an error occurs during _formatting_,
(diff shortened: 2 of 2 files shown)


spec/errors.md
```diff
@@ -24,9 +24,16 @@ or _Message Function Errors_ in _expressions_ that are not otherwise used by the
 such as _placeholders_ in unselected _patterns_
 or _declarations_ that are never referenced during _formatting_.
 
-In all cases, when encountering a runtime error,
-a message formatter MUST provide some representation of the message.
-An informative error or errors MUST also be separately provided.
+When formatting a message with one or more errors:
```
aphillips, line 27: Linkify terms.

```suggestion
When formatting a _message_ with one or more errors:
```

spec/errors.md
```diff
@@ -24,9 +24,16 @@ or _Message Function Errors_ in _expressions_ that are not otherwise used by the
...
+- An implementation MUST provide a way for a user to be informed
+  of the name of at least one of the errors,
+  either directly or via an identifying error code.
```
aphillips, line 30: I'm struggling with a few things here.

I think the "name of" part is a bit circuitous. I think the "identifying error code" part is just an example (out of many) that an implementation might employ. We don't require that implementations use the exact names, I think? That way, some implementations can have a `RESOLUTION_ERROR` and others a `ResolutionError` type, usw.

```suggestion
- An implementation MUST provide a mechanism to discover and identify
  at least one of the errors. 
  The exact form of error signaling is implementation defined.
  Some examples include throwing an exception,
  returning an error code, 
  or providing a function or method for enumerating any errors.
```

spec/errors.md
```diff
@@ -24,9 +24,16 @@ or _Message Function Errors_ in _expressions_ that are not otherwise used by the
...
+- An implementation MUST provide a way for a message with one or more
+  _Resolution Errors_ or _Message Function Errors_ to be formatted
+  using a fallback representation.
```
aphillips, line 33: I think this is set up wrong. The first requirement is about what must be provided to the user. This requires the "message... to be formatted"

Thought: what happens if the message contain a resolution error and also a syntax error? The syntax error is not a resolution or message function error, but this seems to require that the implementation provide fallback representation for the message because of the resolution error.

We do define a fallback representation for syntax and DM errors (it's the logo), so we could just require that there is always a way to get a string representation out.

Note that we prescribe the fallback representation elsewhere with normative language in [fallback values](https://github.com/unicode-org/message-format-wg/blob/main/spec/formatting.md#formatting-fallback-values) and in [fallback resolution](https://github.com/unicode-org/message-format-wg/blob/main/spec/formatting.md#fallback-resolution)

I think this might be better as:

```suggestion
- An implementation MUST provide a way for a user to obtain a [formatted fallback representation](formatting.md#formatting-fallback-values) of any _message_
  with one or more _Resolution Errors_ or _Message Function Errors_.
```

macchiati: >   with one or more _Resolution Errors_ or _Message Function Errors_.

"with" can be ambiguous. I suggest

that has one or more ...

That ties the "one or more ..." more tightly to the "any _message_

Even clearer would be to invert.

Whenever a message has one or more _Resolution Errors_ or _Message Function Errors_,
an implementation MUST provide a way for a user to obtain a [formatted fallback representation](formatting.md#formatting-fallback-values) of that _message_

aphillips: Translating @macchiati's suggestion. Note the addition of "combination of".

```suggestion
- Whenever a _message_ has one or more _Resolution Errors_ or _Message Function Errors_,
  or a combination of these types of errors,
  an implementation MUST provide a way for a user to obtain a 
  [formatted fallback representation](formatting.md#formatting-fallback-values) of that _message_.
```

spec/errors.md
eemeli: > We don't require that implementations use the exact names, I think?

We kind of do, because our conformance tests use [error codes](https://github.com/unicode-org/message-format-wg/blob/bcda078ecf59a78b7a02e8eeef1bb5406625c3b7/test/tests/functions/number.json#L25).

But I'm fine with the suggested change, which avoids the issue.

spec/errors.md
eemeli: > Thought: what happens if the message contain a resolution error and also a syntax error?

It can't. If a message has a syntax or data model error, then in pattern selection it resolves to a pattern with a single fallback element, and that can't fail to resolve or format.

> We do define a fallback representation for syntax and DM errors (it's the logo), so we could just require that there is always a way to get a string representation out.

This would make it significantly more difficult to implement a two-stage API for MF2 that treated syntax & runtime errors differently, as it would effectively require that an initial formatter construction would always succeed even if it's fed a bad message. In other words, the PR's current language explicitly allows for

```js
const msg = new Intl.MessageFormat('bad {horse')
msg.format()
```

to throw an error on the first line, rather than having the second line return `'{�}'`.

> an implementation MUST provide a way for a user to obtain a 
> [formatted fallback representation](formatting.md#formatting-fallback-values) of that _message_.

Linking to the formatting requirements for a concatenated string seems rather odd here, as this rule ought to apply to all formatting targets.

eemeli: How about something like this instead? Rather than enumerating the errors for which we need to be able to get a fallback result, rely on the formatting section to explain fallback values, and enumerate the cases when this MUST does not apply.

```suggestion
- For all messages without _Syntax Errors_ or _Data Model Errors_,
  an implementation MUST enable a user to get a formatted result.
  This result MAY include formatted _fallback values_.
```

aphillips: I like this approach very much, although I think this is repetitive to something we already have, specifically this paragraph in the start of Formatting:

> To start, we presume that a message has either been parsed from its syntax or created from a data model description. If this construction has encountered any Syntax Errors or Data Model Errors, an appropriate error MUST be emitted and a fallback value MAY be used as the formatting result.

We also mention previously (in the unchanged parts of the errors section) that what we're discussing here is the selection and formatting phases. As a result, no syntax or data model errors can apply here. We also want to ensure, I think, that non-conformant implementations (meaning, in this case, ones that emit some unenumerated type of error) still provide a fallback representation _during formatting_.

Question: do we require a fallback representation during pattern selection? Does it use the fallback pattern (that is, the one that has `*` keys all the way across the selectors)? Or just the logo?

```suggestion
- When an error occurs during _pattern selection_ the implementation MUST
  enable the user to get a formatted result.
  This result MAY include formatted _fallback values_.
- When an error occurs during _formatting_ the implementation MUST
  enable the user to get a formatted result.
  This result MAY include formatted _fallback values_.
```

eemeli: Eh, that probably works. My only concern would be that the vagueness in the language around eager/lazy processing allows for expression & markup resolution to take place separately or before either pattern selection or formatting, and we don't have a unique name for all of those steps together given that one of them is already "formatting". Also, why separate bullet points when the result is the same?

> Question: do we require a fallback representation during pattern selection? Does it use the fallback pattern (that is, the one that has `*` keys all the way across the selectors)? Or just the logo?

As the `*` variant is currently required (though #603 is still open), this is defined by the [Filter Variants](https://github.com/unicode-org/message-format-wg/blob/main/spec/formatting.md#filter-variants) step of the selection algorithm.

eemeli: > if we fail in selection, do we grab the * pattern for fallback or just the logo. The Filter Variants step might not be reached if we have a selection failure.

Added #828 to make it clear that never happens.

> If you encounter an error _before_ you select a pattern, you have no way to do fallback representation?

I mean that because we allow for eager processing, it's valid to resolve _all_ of the expressions before you start pattern selection. This means that if we only define what happens on errors specifically during pattern selection or formatting, we technically leave out what happens during that eager stage.

spec/errors.md
```diff
@@ -24,9 +24,19 @@ or _Message Function Errors_ in _expressions_ that are not otherwise used by the
...
+When formatting a _message_ with one or more errors:
+- An implementation MUST provide a mechanism to discover and identify
+  at least one of the errors. 
+  The exact form of error signaling is implementation defined.
+  Some examples include throwing an exception,
+  returning an error code, 
+  or providing a function or method for enumerating any errors.
+- For all messages without _Syntax Errors_ or _Data Model Errors_,
+  an implementation MUST enable a user to get a formatted result.
+  This result MAY include formatted _fallback values_.
```
eemeli, line 36:
```suggestion
When formatting a _message_ with one or more errors,
an implementation MUST provide a mechanism to discover and identify
at least one of the errors. 
The exact form of error signaling is implementation defined.
Some examples include throwing an exception,
returning an error code, 
or providing a function or method for enumerating any errors.

For all _messages_ without _Syntax Errors_ or _Data Model Errors_,
an implementation MUST enable a user to get a formatted result.
This result MAY include formatted _fallback values_.
```

eemeli:
I merged the fallback text suggestion from https://github.com/unicode-org/message-format-wg/pull/816#discussion_r1674418724, as it's at least a step forwards, and turned the two-item list into two separate paragraphs instead.

spec/errors.md
```diff
@@ -24,9 +24,20 @@ or _Message Function Errors_ in _expressions_ that are not otherwise used by the
...
+an implementation MUST provide a mechanism to discover and identify
+at least one of the errors. 
+The exact form of error signaling is implementation defined.
+Some examples include throwing an exception,
+returning an error code, 
+or providing a function or method for enumerating any errors.
+
+For all _messages_ without _Syntax Errors_ or _Data Model Errors_,
+an implementation MUST enable a user to get a formatted result.
+This result MAY include formatted _fallback values_.
```
aphillips, line 37: Line 738 below _requires_ ("MUST") that a fallback value be used. I think that could be read as conflicting with normative "MAY" here, when this is apparently just advice to the user?

Perhaps:

```suggestion
The formatted result might include _fallback values_ 
such as when a _placeholder_'s _expression_ produced an error
during formatting.
```

aphillips (commented):
Looking good. Minor comment about one of the normative statements keeps me from approving.

aphillips (approved):

catamorphism (approved):

mihnita (approved):

PR #823, opened 2024-07-15
Define function composition for :number and :integer values

eemeli:
A continuation for #798 and #814, defining the composition behaviour for numerical values.

### DIFF
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -428,6 +428,14 @@ it can include option values.
 These are included in the resolved option values of the _expression_,
 with _options_ on the _expression_ taking priority over any option values of the _operand_.
 
+> For example, the _placeholder_ in this _message_:
+> ```
+> .input {$n :number notation=scientific minimumFractionDigits=2}
+> {{{$n :number minimumFractionDigits=1}}}
+> ```
+> would be formatted with the resolved options
+> `{ notation: 'scientific', minimumFractionDigits: '1' }`.
+
 > [!NOTE]
 > The following options and option values are being developed during the Technical Preview
 > period.
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -499,7 +499,7 @@ The _function_ `:number` performs selection as described in [Number Selection](#
 
 When an _operand_ or an _option_ value uses a _variable_ annotated,
 directly or indirectly, by a `:number` _annotation_,
-its resolved value contains the implementation-defined numerical value
+its resolved value contains an implementation-defined numerical value
 of the _operand_ of the annotated _expression_,
 together with the resolved options values.
 
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -642,7 +642,7 @@ The _function_ `:integer` performs selection as described in [Number Selection](
 
 When an _operand_ or an _option_ value uses a _variable_ annotated,
 directly or indirectly, by a `:integer` _annotation_,
-its resolved value contains the implementation-defined integer value
+its _resolved value_ contains the implementation-defined integer value
 of the _operand_ of the annotated _expression_,
 together with the resolved options' values.
 
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -509,7 +509,7 @@ When an _operand_ or an _option_ value uses a _variable_ annotated,
 directly or indirectly, by a `:number` _annotation_,
 its resolved value contains an implementation-defined numerical value
 of the _operand_ of the annotated _expression_,
-together with the resolved options values.
+together with the resolved options' values.
 
 ### The `:integer` function
 
@@ -644,7 +644,7 @@ When an _operand_ or an _option_ value uses a _variable_ annotated,
 directly or indirectly, by a `:integer` _annotation_,
 its resolved value contains the implementation-defined integer value
 of the _operand_ of the annotated _expression_,
-together with the resolved options values.
+together with the resolved options' values.
 
 ### Number Operands
 
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -507,7 +507,7 @@ The _function_ `:number` performs selection as described in [Number Selection](#
 
 When an _operand_ or an _option_ value uses a _variable_ annotated,
 directly or indirectly, by a `:number` _annotation_,
-its resolved value contains an implementation-defined numerical value
+its _resolved value_ contains an implementation-defined numerical value
 of the _operand_ of the annotated _expression_,
 together with the resolved options' values.
 
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -422,6 +422,12 @@ The following options and their values are required to be available on the funct
 - `maximumSignificantDigits`
   - ([digit size option](#digit-size-options))
 
+If the _operand_ of the _expression_ is an implementation-defined type,
+such as the _resolved value_ of an _expression_ with a `:number` or `:integer` _annotation_,
+it can include option values.
+These are included in the resolved option values of the _expression_,
+with _options_ on the _expression_ taking priority over any option values of the _operand_.
+
 > [!NOTE]
 > The following options and option values are being developed during the Technical Preview
 > period.
@@ -489,6 +495,14 @@ MUST be multiplied by 100 for the purposes of formatting.
 
 The _function_ `:number` performs selection as described in [Number Selection](#number-selection) below.
 
+#### Composition
+
+When an _operand_ or an _option_ value uses a _variable_ annotated,
+directly or indirectly, by a `:number` _annotation_,
+its resolved value contains the implementation-defined numerical value
+of the _operand_ of the annotated _expression_,
+together with the resolved options values.
+
 ### The `:integer` function
 
 The function `:integer` is a selector and formatter for matching or formatting numeric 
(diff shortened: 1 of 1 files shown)


spec/registry.md
```diff
@@ -422,6 +422,12 @@ The following options and their values are required to be available on the funct
 - `maximumSignificantDigits`
   - ([digit size option](#digit-size-options))
 
+If the _operand_ of the _expression_ is an implementation-defined type,
+such as the _resolved value_ of an _expression_ with a `:number` or `:integer` _annotation_,
```
aphillips, line 426: We are defining `:number` and `:integer` here. Is "such as" needed?

spec/registry.md
```diff
@@ -538,6 +551,18 @@ function `:integer`:
...
+it can include option values.
```
aphillips, line 556: ibid.

spec/registry.md
```diff
@@ -605,6 +630,14 @@ MUST be multiplied by 100 for the purposes of formatting.
 
 The _function_ `:integer` performs selection as described in [Number Selection](#number-selection) below.
 
+#### Composition
+
+When an _operand_ or an _option_ value uses a _variable_ annotated,
```
aphillips, line 635: "uses" seems wrong to me?

```suggestion
When an _operand_ or the value of an _option_ consists of a _variable_ previously annotated,
```

spec/registry.md
eemeli: The "such as" is making sure that the _resolved value_ type is included in the "an implementation-defined type" set of possibilities.

spec/registry.md
```diff
@@ -422,6 +422,12 @@ The following options and their values are required to be available on the funct
 - `maximumSignificantDigits`
   - ([digit size option](#digit-size-options))
 
+If the _operand_ of the _expression_ is an implementation-defined type,
+such as the _resolved value_ of an _expression_ with a `:number` or `:integer` _annotation_,
+it can include option values.
```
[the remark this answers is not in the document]
eemeli: See the thread above; the language currently covers numerical types that are not _resolved values_. With that context, changing this to "it includes" seems a bit wrong?

spec/registry.md
```diff
@@ -422,6 +422,12 @@ The following options and their values are required to be available on the funct
...
+These are included in the resolved option values of the _expression_,
+with _options_ on the _expression_ taking priority over any option values of the _operand_.
```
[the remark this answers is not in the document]
eemeli:
```suggestion
These are included in the resolved option values of the _expression_,
with _options_ on the _expression_ taking priority over any option values of the _operand_.

> For example, the _placeholder_ in this _message_:
> ```
> .input {$n :number notation=scientific minimumFractionDigits=2}
> {{{$n :number minimumFractionDigits=1}}}
> ```
> would be formatted with the resolved options
> `{ notation: 'scientific', minimumFractionDigits: '1' }`.
```

spec/registry.md
eemeli: This language was suggested by @stasm in https://github.com/unicode-org/message-format-wg/pull/814#discussion_r1668948100, and is currently the same across the three PRs. Would it be ok to not nitpick on this now, and iterate on the exact language once these changes have first landed?

spec/registry.md
aphillips: Presumably it would be that, otherwise the result would be bad operand?

---

Why does the operand need to be of the implementation-defined variety? Aren't option values also applicable to, for example, accepted literal values?

```
{-1.234 :number maximumFractionDigits=2 signDisplay=always}
```

Why wouldn't the options "stick to" the value?

eemeli: I'm trying to follow our practice of not defining exactly what a "number" type is, while allowing for some implementation-defined number types, such as the resolved value coming from a `:number` annotation, to contain options within the value of the operand.

So this is covering each of the following use cases:

1. `{$x :number}`
where in a JS context that could be formatted with `{ x: BigInt(42) }`, i.e. using a language-specific numerical type.

2. `{$y :number style=currency}`
where in a JS context that could be formatted with
`{ y: Object.assign(new Number(42), { options: { currency: 'EUR' } }) }`,
i.e. an object that carries both a numerical value and a bag of options.

3. `.local $z = {42 :number notation=compact} {{{$z :number compactDisplay=long}}}`
so that in the _placeholder_ the options of the _operand_ `$z` can be combined with the _expression_ _options_.

aphillips: Right. We're on the same page technically. I'm bikeshedding the language.

I think the challenge for me is that "implementation-defined type" appears to exclude some values (such as literal representation). It's not clear to me if we require the implementation to resolve it to an implementation-defined type. For example, we say:

> The operand of a number function is either an implementation-defined type or a literal whose contents match the number-literal production in the [ABNF](https://github.com/unicode-org/message-format-wg/blob/main/spec/message.abnf). All other values produce a Bad Operand error.

... but we don't say that the literal _becomes_ one of the implementation-defined types (although it might).

eemeli: The thing is, though, that a literal representation of a number cannot include option values within the operand. The _annotation_ of a literally represented number can absolutely have options, but its _operand_ cannot.

aphillips: What about transitivity?

```
.local $a = {42 :number signDisplay=always}
.local $b = {$a :number minimumFractionDigits=1}
{{{$b}'s operand is an annotated literal??}}
```

eemeli: In that example, when resolving the _expression_ defining the value of `$b`, its _operand_ is a _resolved value_, an implementation-defined type. The numerical value contained in that implementation-defined type is the number 42, and it includes a bag of resolved options `{ signDisplay: 'always' }`.

When resolving the value of `$b` or formatting the placeholder `{$b}`, the only impact that the literal origin of the value should have is that the fallback string that's used for it is `|42|`, and not e.g. `$x`.

spec/registry.md
```diff
@@ -489,6 +503,14 @@ MUST be multiplied by 100 for the purposes of formatting.
 
 The _function_ `:number` performs selection as described in [Number Selection](#number-selection) below.
 
+#### Composition
+
+When an _operand_ or an _option_ value uses a _variable_ annotated,
+directly or indirectly, by a `:number` _annotation_,
+its resolved value contains an implementation-defined numerical value
```
catamorphism, line 510: Would it be more precise to say that the `getValue()` method of its resolved value returns an implementation-defined numeric value? (Assuming the definition of resolved values in #728.)

spec/registry.md
```diff
@@ -489,6 +503,14 @@ MUST be multiplied by 100 for the purposes of formatting.
...
+of the _operand_ of the annotated _expression_,
+together with the resolved options values.
```
catamorphism, line 512:
```suggestion
together with the resolved options' values.
```

spec/registry.md
```diff
@@ -538,6 +559,18 @@ function `:integer`:
 - `maximumSignificantDigits`
   - ([digit size option](#digit-size-options))
 
+If the _operand_ of the _expression_ is an implementation-defined type,
+such as the _resolved value_ of an _expression_ with a `:number` or `:integer` _annotation_,
+it can include option values.
+In general, these are included in the resolved option values of the _expression_,
+with _options_ on the _expression_ taking priority over any option values of the _operand_.
+Option values with the following names are however discarded if included in the _operand_:
```
catamorphism, line 567: Does this mean "discarded" as in "not included in the resolved option values of the expression", or "discarded" as in "ignored completely"? I think it's the first one, but it could be misread as meaning the second one.

spec/registry.md
```diff
@@ -605,6 +638,14 @@ MUST be multiplied by 100 for the purposes of formatting.
 
 The _function_ `:integer` performs selection as described in [Number Selection](#number-selection) below.
 
+#### Composition
+
+When an _operand_ or an _option_ value uses a _variable_ annotated,
+directly or indirectly, by a `:integer` _annotation_,
+its resolved value contains the implementation-defined integer value
+of the _operand_ of the annotated _expression_,
+together with the resolved options values.
```
catamorphism, line 647:
```suggestion
together with the resolved options' values.
```

catamorphism (changes requested):

spec/registry.md
eemeli: Yes.

spec/registry.md
eemeli: Is there some observable difference between the two interpretations? These are the `:number` options that are not available on `:integer`, so I would presume that they would not have any impact on the selection or formatting either?

catamorphism: Ah, I see -- I didn't catch that the list was listing options not available on `:integer`.

catamorphism (approved):

spec/registry.md
eemeli, line 510:
```suggestion
its _resolved value_ contains an implementation-defined numerical value
```

spec/registry.md
```diff
@@ -605,6 +638,14 @@ MUST be multiplied by 100 for the purposes of formatting.
 
 The _function_ `:integer` performs selection as described in [Number Selection](#number-selection) below.
 
+#### Composition
+
+When an _operand_ or an _option_ value uses a _variable_ annotated,
+directly or indirectly, by a `:integer` _annotation_,
+its resolved value contains the implementation-defined integer value
```
eemeli, line 645:
```suggestion
its _resolved value_ contains the implementation-defined integer value
```

PR #825, opened 2024-07-15
Rename `quoted` as `quoted-literal` and `unquoted` as `unquoted-literal`

eemeli:
Given that we also have `quoted-char` and `quoted-pattern`, it is somewhat confusing to have a bare `quoted` rule. It would be much clearer to call it `quoted-literal`, and correspondingly specify that `unquoted` is `unquoted-literal`.

Doing so would also fix at least two places in the formatting text which already include references to _quoted literal_ as a named thing.

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -361,7 +361,7 @@ An _expression_ fails to resolve when:
 
 The _fallback value_ depends on the contents of the _expression_:
 
-- _expression_ with any _literal_ _operand_:
+- _expression_ with a _literal_ _operand_ (either quoted or unquoted)
   U+007C VERTICAL LINE `|`
   followed by the value of the _literal_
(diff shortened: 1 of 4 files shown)


spec/formatting.md
```diff
@@ -361,7 +361,7 @@ An _expression_ fails to resolve when:
 
 The _fallback value_ depends on the contents of the _expression_:
 
-- _expression_ with _literal_ _operand_ (_quoted_ or _unquoted_):
+- _expression_ with any _literal_ _operand_:
```
aphillips, line 364: This one is tricky, since it eliminates what I thought was a good reminder about literals.

```suggestion
- _expression_ with a _literal_ _operand_ (either quoted or unquoted)
```

aphillips (approved):
I'm okay with this. It's an editorial change, but it **_is_** a change to the syntax...

catamorphism (approved):
Seems like a good clarification.

mihnita (approved):

PR #826, opened 2024-07-15
Apply minor test fixes

eemeli:
Discovered while updating the JS implementation to use the official test suite.

### DIFF
--- a/test/tests/functions/number.json
+++ b/test/tests/functions/number.json
@@ -27,7 +27,7 @@
       ]
     },
     {
-      "src": "invalid number literal {.1 :number}",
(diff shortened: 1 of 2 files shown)


aphillips (approved):
LGTM. Some of the number ones I glance at and wonder if we should permit?

eemeli:
> LGTM. Some of the number ones I glance at and wonder if we should permit?

Right now we only allow what JSON allows. Tbh I'd rather not relax that, as it's the one spec which is probably most well known that we could be using for this.

mihnita (approved):

PR #827, opened 2024-07-15
Test suite: Add `$schema` to scenarios, drop `expCleanSrc`

eemeli:
Two different changes here:
1. Add a `$schema` property to each scenario file, so that they can get autocompletion & validation automatically in editors.
2. Drop the `expCleanSrc` value from tests, as we don't have consensus on the canonical representation of each message.


aphillips:
> Drop the expCleanSrc value from tests, as we don't have consensus on the canonical representation of each message.

It's not (or not just) lack of consensus. We don't have a normative requirement here.

aphillips (approved):

aphillips:
fast-track per 2024-07-22 call. @mradbourne please review

mradbourne (approved):
Looks great.

You can also remove [test/.vscode/settings.json](https://github.com/unicode-org/message-format-wg/blob/main/test/.vscode/settings.json), which will be redundant after this change.

mradbourne:
@eemeli On reflection, I'm thinking that including the schema in the data might cause confusion for other usages of the data. [Adding example now...]

aphillips:
Fast track criteria were met. Thanks all!

PR #828, opened 2024-07-15
Clarify that calling MatchSelectorKeys always returns a list

eemeli:
Follow-up to https://github.com/unicode-org/message-format-wg/pull/816#discussion_r1677961448, as it was not previously clear that MatchSelectorKeys never fails.

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -549,6 +549,10 @@ The returned list MAY be empty.
 The most-preferred key is first,
 with each successive key appearing in order by decreasing preference.
 
+If calling MatchSelectorKeys encounters any error,
+a _Bad Selector_ error is emitted
+and an empty list is returned.
+
 ### Filter Variants
 
 Then, using the preferential key orders `pref`,


aphillips:
The description of this PR and the text seem different? `MatchSelectorKeys` "never fails" by emitting an error and returning the empty list?

spec/formatting.md
```diff
@@ -549,6 +549,10 @@ The returned list MAY be empty.
 The most-preferred key is first,
 with each successive key appearing in order by decreasing preference.
 
+If calling MatchSelectorKeys encounters any error,
+a _Bad Selector_ error is emitted
+and an empty list is returned.
```
aphillips, line 554:
```suggestion
and an empty list is returned.
The fallback representation of such a _message_ is 
the _fallback value_ for the _message_.
```

aphillips (approved):
Suggestion for the fallback handling.

spec/formatting.md
eemeli: That correction does not match what the algorithm results in. If we get an empty list here, we end up selecting the `*` variant.

catamorphism (approved):

PR #834, opened 2024-07-22
Update the stability policy

aphillips:
Based on discussion in the 2024-07-22 call and in PR #829, update the stability policy.

### DIFF
--- a/spec/README.md
+++ b/spec/README.md
@@ -79,7 +79,7 @@ A reference to a _term_ looks like this.
 > The provisions of the stability policy are not in effect until
 > the conclusion of the technical preview and adoption of this specification.
 
-Updates to this specification will not make any well-formed or valid _message_ invalid.
+Updates to this specification will not make any valid _message_ invalid.
 
 Updates to this specification will not remove any syntax provided in this version.
 
--- a/spec/README.md
+++ b/spec/README.md
@@ -79,12 +79,16 @@ A reference to a _term_ looks like this.
 > The provisions of the stability policy are not in effect until
 > the conclusion of the technical preview and adoption of this specification.
 
-Updates to this specification will not change
-the syntactical meaning, the runtime output, or other behaviour
-of valid messages written for earlier versions of this specification
-that only use functions defined in this specification.
-Updates to this specification will not remove any syntax provided in this version.
-Future versions MAY add additional structure or meaning to existing syntax.
+Updates to this specification shall not make any well-formed or valid _message_ invalid.
+Updates to this specification shall not remove any syntax provided in this version.
+Future versions might add additional structure or meaning to existing syntax.
+
+Updates to this specification shall not change the syntactical meaning
+of any syntax defined in this specification except for that syntax marked as
+"reserved for future standardization".
+
+Updates to this specification shall not assign any meaning or change the syntactical 
+requirements for any private-use annotation.
 
 Updates to this specification will not remove any reserved keywords or sigils.
 
@@ -95,9 +99,11 @@ Updates to this specification will not reserve or assign meaning to
 any character "sigils" except for those in the `reserved` production.
 
 Updates to this specification
-will not remove any functions defined in the default registry nor
-will they remove any options or option values.
-Additional options or option values MAY be defined.
+will not remove any functions defined in the default registry.
+
+Updates to this specification will not remove any options or option values
+defined in the default registry.
+Additional options or additional option values for existing options MAY be defined.
 
 > [!NOTE]
 > This does not guarantee that the results of formatting will never change.
--- a/spec/README.md
+++ b/spec/README.md
@@ -79,45 +79,39 @@ A reference to a _term_ looks like this.
 > The provisions of the stability policy are not in effect until
 > the conclusion of the technical preview and adoption of this specification.
 
-Updates to this specification shall not make any well-formed or valid _message_ invalid.
-Updates to this specification shall not remove any syntax provided in this version.
-Future versions might add additional structure or meaning to existing syntax.
+Updates to this specification will not make any well-formed or valid _message_ invalid.
 
-Updates to this specification shall not change the syntactical meaning
+Updates to this specification will not remove any syntax provided in this version.
+
+Updates to this specification will not change the syntactical meaning
 of any syntax defined in this specification except for that syntax marked as
 "reserved for future standardization".
 
-Updates to this specification shall not assign any meaning or change the syntactical 
+Updates to this specification will not assign any meaning to or change the syntactical 
 requirements for any private-use annotation.
 
 Updates to this specification will not remove any reserved keywords or sigils.
 
-> [!NOTE]
-> Future versions may define new keywords.
+Updates to this specification will not add any additional Unicode code points to 
+those in `reserved-annotation-start`.
 
-Updates to this specification will not reserve or assign meaning to
-any character "sigils" except for those in the `reserved` production.
-
-Updates to this specification
-will not remove any functions defined in the default registry.
+Updates to this specification will not remove any functions defined in the default registry.
 
 Updates to this specification will not remove any options or option values
 defined in the default registry.
-Additional options or additional option values for existing options MAY be defined.
 
 > [!NOTE]
-> This does not guarantee that the results of formatting will never change.
-> Even when the specification doesn't change,
+> The foregoing policies are _not_ a guarantee that the results of formatting will never change.
+> Even when this specification or its implementation do not change,
 > the functions for date formatting, number formatting and so on
-> will change their results over time.
-
-Later specification versions MAY make previously invalid messages valid.
+> can change their results over time or behave differently due to local runtime
+> differences in implementation or changes to locale data
+> (such as due to the release of new CLDR versions).
 
 Updates to this specification will not introduce message syntax that,
 when parsed according to earlier versions of this specification,
 would produce syntax or data model errors.
-Such messages MAY produce errors when formatted
-according to an earlier version of this specification.
+Such messages can produce errors when formatted according to an earlier version of this specification.
 
 From version 2.0, MessageFormat will only reserve, define, or require
 function names or function option names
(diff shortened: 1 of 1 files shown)


spec/README.md
```diff
@@ -79,12 +79,16 @@ A reference to a _term_ looks like this.
...
 > the conclusion of the technical preview and adoption of this specification.
 
-Updates to this specification will not change
-the syntactical meaning, the runtime output, or other behaviour
-of valid messages written for earlier versions of this specification
-that only use functions defined in this specification.
-Updates to this specification will not remove any syntax provided in this version.
-Future versions MAY add additional structure or meaning to existing syntax.
+Updates to this specification shall not make any well-formed or valid _message_ invalid.
+Updates to this specification shall not remove any syntax provided in this version.
```
eemeli, line 83: These two promises are already covered by
> Updates to this specification will not introduce message syntax that, when parsed according to earlier versions of this specification, would produce syntax or data model errors.

further down in the policy.

Also, the second one follows from the first.

spec/README.md
```diff
@@ -79,12 +79,16 @@ A reference to a _term_ looks like this.
...
+Future versions might add additional structure or meaning to existing syntax.
+
+Updates to this specification shall not change the syntactical meaning
+of any syntax defined in this specification except for that syntax marked as
+"reserved for future standardization".
```
eemeli, line 88: Isn't this already covered by the following?
> Updates to this specification will not reserve or assign meaning to
any character "sigils" except for those in the `reserved` production.

spec/README.md
```diff
@@ -95,9 +99,11 @@ Updates to this specification will not reserve or assign meaning to
...
 
 Updates to this specification
-will not remove any functions defined in the default registry nor
-will they remove any options or option values.
-Additional options or option values MAY be defined.
+will not remove any functions defined in the default registry.
+
+Updates to this specification will not remove any options or option values
+defined in the default registry.
+Additional options or additional option values for existing options MAY be defined.
```
eemeli, line 106: If we're this explicit about options & option values, we should note that new functions may be added.

eemeli (changes requested):
While I agree and appreciate that we should not promise the same output in future versions due to e.g. CLDR data changes, I think this PR goes too far in trimming back our promises: We should retain a promise that if you have a valid message that you can format without error using MF 2.0, you'll be able to format it without error also with MF 2.1.

I'm also concerned about the overlap between different parts of the stability policy. There was some before, but significant parts of the language added in this PR duplicate already existing promises, though with different words.

spec/README.md
eemeli: Eh, I confused myself. The text further down is promising things the other way than what's proposed here. Leaving this comment in, though, as these two parts should probably be merged into one.

spec/README.md
aphillips: I'm trying here to account for the original text's "syntactical meaning" promise, which I think is important (e.g. `.input` will not suddenly mean something different from an input declaration, `.match` will always be a selector, etc.). The `reserved-keyword` and `reserved-annotation` promises are separate and the promise not to define even more sigils (`reserved-annotation-start`) are yet more.

The one you quote should really mention the `reserved-annotation-start` production.

aphillips:
I did a thorough refactor of the stability policy.

spec/README.md
```diff
@@ -79,39 +79,39 @@ A reference to a _term_ looks like this.
...
-Updates to this specification will not change
-the syntactical meaning, the runtime output, or other behaviour
-of valid messages written for earlier versions of this specification
-that only use functions defined in this specification.
+Updates to this specification will not make any well-formed or valid _message_ invalid.
+
 Updates to this specification will not remove any syntax provided in this version.
-Future versions MAY add additional structure or meaning to existing syntax.
+
+Updates to this specification will not change the syntactical meaning
```
catamorphism, line 86:
```suggestion
Updates to this specification will not change the meaning
```

(Not sure how "syntactical meaning" is different from just "meaning" here)

spec/README.md
```diff
@@ -79,39 +79,39 @@ A reference to a _term_ looks like this.
...
-Additional options or option values MAY be defined.
+Updates to this specification will not remove any options or option values
+defined in the default registry.
 
 > [!NOTE]
-> This does not guarantee that the results of formatting will never change.
-> Even when the specification doesn't change,
+> The foregoing policies are _not_ a guarantee that the results of formatting will never change.
+> Even when this specification or its implementation do not change,
 > the functions for date formatting, number formatting and so on
```
catamorphism, line 106:
```suggestion
> Even when this specification does not change,
> the implementations of functions for date formatting, number formatting and so on
```

The functions are part of an implementation, so if the behavior of functions changed, it means that one or more implementations changed.

spec/README.md
```diff
@@ -79,39 +79,39 @@ A reference to a _term_ looks like this.
...
+> can change their results over time or behave differently due to local runtime
+> differences in implementation or changes to locale data
+> (such as due to the release of new CLDR versions).
 
 Updates to this specification will not introduce message syntax that,
 when parsed according to earlier versions of this specification,
 would produce syntax or data model errors.
-Such messages MAY produce errors when formatted
-according to an earlier version of this specification.
+Such messages can produce errors when formatted according to an earlier version of this specification.
```
catamorphism, line 114:
```suggestion
Messages that use syntax introduced in a future version of this specification
could produce resolution or message function errors
when formatted according to an earlier version of this specification.
```

It's not clear what "such messages" means. How about this?

catamorphism (approved):
LGTM with minor clarifications

spec/README.md
aphillips: Agreed, although I'm also trying to point out that the function implementation (the code) might also remain static and the output change (because your operating system, for example, got a new copy of CLDR data)

spec/README.md
aphillips: Syntactical meaning is: `.input` is an input declaration. 

"Meaning" can be read as "how a specific `.input` is understood", so `.input {$foo @bar=baz}` might "change meaning" if we allowed expression attributes and assigned "meaning" to the `@bar` one. That's why I used the circuitous and painful wording.

Perhaps:

```suggestion
Updates to this specification will not alter the role assigned to 
defined elements of the syntax, except to assign an interpretation to syntax marked as
"reserved for future standardization".
```

eemeli (commented):
Partial review; lots of terms should be linkified.

spec/README.md
```diff
@@ -79,39 +79,41 @@ A reference to a _term_ looks like this.
...
+
+Updates to this specification will not assign any meaning to or change the syntactical 
+requirements for any private-use annotation.
 
 Updates to this specification will not remove any reserved keywords or sigils.
 
-> [!NOTE]
-> Future versions may define new keywords.
+Updates to this specification will not add any additional Unicode code points to 
+those in `reserved-annotation-start`.
```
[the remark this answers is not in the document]
aphillips: That is an implication of that other promise. This is (a) a specific rewording of a promise we currently make to make it technically accurate and clear [which may be why we didn't notice that we'd already promised it by implication??] and (b) serves as a touchstone for "future us" when it is suggested that we create a new sigil not in the list [we explicitly promised not to].

spec/README.md
```diff
@@ -79,39 +79,41 @@ A reference to a _term_ looks like this.
 > The provisions of the stability policy are not in effect until
 > the conclusion of the technical preview and adoption of this specification.
 
-Updates to this specification will not change
-the syntactical meaning, the runtime output, or other behaviour
-of valid messages written for earlier versions of this specification
-that only use functions defined in this specification.
+Updates to this specification will not make any well-formed or valid _message_ invalid.
```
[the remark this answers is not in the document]
aphillips: In the refactor, I broke each of the promises apart.

This promise says that, if a message is well-formed, it will still be well-formed in the future and if a message is valid it will still be valid in the future. 

Well-formed is limited to the grammar, i.e. syntax errors. Valid extends to data model errors.

It's plausible to say that we will introduce new, tighter restrictions into the syntax in the future that would result in a message becoming a syntax error that is not currently a syntax error. The extension to well-formed is thus unreasonable to include and I will remove it now.

eemeli (changes requested):
It looks like my previous line-specific comments have been addressed, leaving only this blocker from my previous review:

> While I agree and appreciate that we should not promise the same output in future versions due to e.g. CLDR data changes, I think this PR goes too far in trimming back our promises: We should retain a promise that if you have a valid message that you can format without error using MF 2.0, you'll be able to format it without error also with MF 2.1.

spec/README.md
```diff
@@ -79,64 +79,59 @@ A reference to a _term_ looks like this.
...
 
-Updates to this specification will not change
-the syntactical meaning, the runtime output, or other behaviour
-of valid messages written for earlier versions of this specification
-that only use functions defined in this specification.
+Updates to this specification will not make any valid _message_ invalid.
+
 Updates to this specification will not remove any syntax provided in this version.
-Future versions MAY add additional structure or meaning to existing syntax.
+
```
eemeli, line 85: @aphillips This change would satisfy my remaining concerns.

```suggestion

Updates to this specification will not introduce new runtime error conditions that would
produce errors or require the use of fallback values when formatting messages
that formatted without error or fallback with an earlier version of this specification.

```

aphillips: Thanks for this. I think it could need tweaking?

```suggestion
Updates to this specification MUST NOT specify an error for any message
that previously did not specify an error.

Updates to this specification MUST NOT specify the use of a fallback value for any message
that previously did not specify a fallback value.
```

We need to guard against new error conditions (your text) but also assignment of existing errors to new criteria. Also, note that I focus on specifying, which is something _we_ control. Messages don't generate errors/fallbacks: implementations do. Non-conformant implementations should not constrain us.

eemeli: Sure, that works as well. I edited your suggestion to add in the leading and trailing empty lines; otherwise these will connect with other paragraphs.

eemeli (approved):

PR #839, opened 2024-07-26
Add LCA to delegates.txt

lucacasonato:


aphillips:
Actually, I'm going to eliminate this section of the delegates file. We don't need to maintenance chore. CLA is all that matters. Credit in the spec is currently derived from participation in github.

PR #840, opened 2024-07-28
Disallow "whitespace or special char" prefixed `.` in reserved-statement's body

lucacasonato:
This commit disallows using dots inside of `reserved-statement` bodies, outside of name-like strings.

This enables tools to perform better error recovery in case of incomplete reserved-statements.

## Background

Take for example the following incomplete message:

```mf2
.foobar $var =
.local $count = {1}
{{The count is {$count}.}}
```

It is relatively obvious here that the user did not intend to write a reserved-statement with keyword `.foobar` and body `$var = \n.local $count =` and expression `{1}`, but two declarations: a reserved statement and a `.local` declaration.

A parser that does error recovery, like is needed for in-editor tooling should be able to recover from the above message into the users' intended form: a local decl, a reserved statement, and a quoted body. But right now the syntax prohibits this error recovery, because the above message _is_ already valid, but parses into a single reserved-statement.

This PR attempts to resolve this behaviour by disallowing `.` in reserved-statement when it is not preceded by a `name-char`. The `.` is not banned outright, because I think that all valid declarations right now, should be valid reserved-statements, and we allow `.` in `.local` declarations inside of names: `.local $foo.bar = {123}`.

### DIFF
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -939,8 +939,8 @@ name-start = ALPHA / "_"
            / %x370-37D / %x37F-1FFF / %x200C-200D
            / %x2070-218F / %x2C00-2FEF / %x3001-D7FF
            / %xF900-FDCF / %xFDF0-FFFC / %x10000-EFFFF
-name-char  = name-start / DIGIT / "-" / %xB7 / %x300-36F
-           / %x203F-2040
+name-char  = name-start / DIGIT / "-"
+           / %xB7 / %x300-36F / %x203F-2040
 ```
 
 ### Escape Sequences
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -227,16 +227,16 @@ Any such future keyword must start with `.`,
 followed by two or more lower-case ASCII characters.
 
 The rest of the statement supports
-a similarly wide range of content as _reserved annotations_ (with the exception
-of U+002E FULL STOP `.`, which is only allowed following characters that are
-permissible in _name_), but it MUST end with one or more _expressions_.
+a similarly wide range of content as _reserved annotations_
+(with the exception of U+002E FULL STOP `.`,
+which is only allowed following characters that are permissible in _name_), but it MUST end with one or more _expressions_.
 
 ```abnf
 reserved-statement = reserved-keyword [s reserved-statement-body] 1*([s] expression)
 reserved-keyword   = "." name
 
 reserved-statement-body      = reserved-statement-body-part *([s] reserved-statement-body-part)
-reserved-statement-body-part = (name-char-no-dot 1*".") / content-char
+reserved-statement-body-part = (name-char 1*".") / content-char
                              / escaped-char / quoted-literal
 ```
 
@@ -246,10 +246,11 @@ reserved-statement-body-part = (name-char-no-dot 1*".") / content-char
 > `.input`, `.local`, or `.match`.
 
 > [!NOTE]
-> Only dots that are preceded by a valid name character are allowed in a
-> `reserved-statement-body`. This enables better parser error recovery for tools in the
-> case of incomplete reserved statements, like in the following example:
-> ```mf2
+> Only dots that are preceded by a valid name character are allowed in a  `reserved-statement-body`.
+> This enables better parser error recovery for tools
+> in the case of incomplete reserved statements,
+> like in the following example:
+> ```
 > .invalid $foo.bar =
 > .local $bar = {|baz|}
 > {{This is a message.}}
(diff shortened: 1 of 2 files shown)


spec/syntax.md
```diff
@@ -227,19 +227,34 @@ Any such future keyword must start with `.`,
...
 The rest of the statement supports
-a similarly wide range of content as _reserved annotations_,
-but it MUST end with one or more _expressions_.
+a similarly wide range of content as _reserved annotations_ (with the exception
+of U+002E FULL STOP `.`, which is only allowed following characters that are
+permissible in _name_), but it MUST end with one or more _expressions_.
 
 ```abnf
-reserved-statement = reserved-keyword [s reserved-body] 1*([s] expression)
+reserved-statement = reserved-keyword [s reserved-statement-body] 1*([s] expression)
```
eemeli, line 235: With the change proposed here, could we not also drop the requirement for one or more expressions? If a `.` preceded by whitespace or `}` is no longer allowed in a reserved statement, then we can allow for e.g. `.strict` just by itself to be a valid `reserved-statement`.

CC @gibson042 about this.

```suggestion
reserved-statement = reserved-keyword [s reserved-statement-body] *([s] expression)
```

spec/syntax.md
```diff
@@ -227,19 +227,34 @@ Any such future keyword must start with `.`,
...
 > [!NOTE]
 > The `reserved-keyword` ABNF rule is a simplification,
 > as it MUST NOT be considered to match any of the existing keywords
 > `.input`, `.local`, or `.match`.
 
+> [!NOTE]
+> Only dots that are preceded by a valid name character are allowed in a
+> `reserved-statement-body`. This enables better parser error recovery for tools in the
+> case of incomplete reserved statements, like in the following example:
+> ```mf2
```
eemeli, line 252: Semantic line breaks, and `mf2` isn't a recognised linguist syntax identifier at least yet. If we're to add that, let's do so systematically everywhere.

```suggestion
> Only dots that are preceded by a valid name character are allowed in a `reserved-statement-body`.
> This enables better parser error recovery for tools
> in the case of incomplete reserved statements,
> like in the following example:
> ```
```

spec/syntax.md
```diff
@@ -915,16 +930,17 @@ in this release.
...
-name-start = ALPHA / "_"
-           / %xC0-D6 / %xD8-F6 / %xF8-2FF
-           / %x370-37D / %x37F-1FFF / %x200C-200D
-           / %x2070-218F / %x2C00-2FEF / %x3001-D7FF
-           / %xF900-FDCF / %xFDF0-FFFC / %x10000-EFFFF
-name-char  = name-start / DIGIT / "-" / "."
-           / %xB7 / %x300-36F / %x203F-2040
+identifier       = [namespace ":"] name
+namespace        = name
+name             = name-start *name-char
```
eemeli, line 935: One option would be leaving out the `"."` from `name-char`, dropping `name-char-no-dot`, and including here something like:

```suggestion
name             = name-start *(name-char / ".")
```

That would somewhat imply our intent for `.` to be used as a separator between name parts without actually making it special in the syntax.

Please take this as purely a suggestion, I'm much more interested in the actual syntax rather than the rules we use to describe it.

eemeli (approved):
I support this, either with or without my comments below being taken into account. We'll also need `message.abnf` to be updated at the same time, but happy to work within `syntax.md` to first figure out the exact change we'll want to apply.

spec/syntax.md
lucacasonato: I would also be OK with this.

spec/syntax.md
lucacasonato: That seems reasonable.

lucacasonato:
Just want to clarify that I am opening this PR as both a "feedback with proposed solution". I feel a PR in draft mode is better suited to this than just an issue :)

I would like to discuss this at an upcoming meeting

aphillips:
(chair hat on)

You can add the Agenda+ label any issue or PR if you want to discuss something in a call. I have done that for you here.

Note that submitting a [design doc](https://github.com/unicode-org/message-format-wg/tree/main/exploration) is the preferred mechanism for this sort of thing, since it allows you to also present the technical argument and the results are kept for posterity.

spec/syntax.md
gibson042: Breadcrumb for when I get to this: #547

aphillips:
In [today](https://github.com/unicode-org/message-format-wg/blob/main/meetings/2024/notes-2024-09-09.md)'s call, we decided to _remove_ the reserved statement from the formal ABNF of a _message_ and instead modify our stability policy to allow future backward compatible additions to the syntax. As a result, this PR (and it's related issues) would not be relevant.

aphillips:
Closing this as out-of-scope. We'll revisit this if/when we revive reserved-statement or private-use.

PR #841, opened 2024-07-29
Require at least one keyword for complex messages

### LINKED ISSUE #787: [FEEDBACK] Isolating quoted patterns on the outside adds a lookahead to the syntax
An observation from [implementing](https://github.com/messageformat/messageformat/compare/mf2-source-bidi) bidi isolation as proposed in #781, but which also applies to the currently proposed design for [bidi usability](https://github.com/unicode-org/message-format-wg/blob/main/exploration/bidi-usability.md):

Isolating quoted patterns on the outside adds LRI, RLI & FSI to the set of characters (currently `{` and `.`) that could start a quoted message with no declarations, as in `\u2066{{hello}}\u2069`.

This doesn't make the syntax ambiguous as the `{{` isn't valid in a `simple-message`, but it does add a lookahead of one token to the parser.

The same lookahead is also required in `variant`, to determine whether a `\u2066` starts a `quoted` `key`, or a `quoted-pattern`.

The simplest change to avoid this lookahead would probably be to place the `open-isolate` and `close-isolate` between the braces, as in `{\u2066{hello}\u2069}`. In this position, it would also match what's proposed for `expression` and `markup`.

### LINKED ISSUE #837: [FEEDBACK] `.` can not be escaped
A simple message can not start with `.`. When one wants to create a message that starts with a `.`, the first logical thing to do is to try to escape it: `\.`. Unlike for `{`, this does not work. That means to create a message that starts with `.` you have two options:

```mf2
{{. foobar}}
```

(a complex message)

```mf2
{|.|} foobar
```

(wrapping `.` in a quoted literal in an expression)

That is kinda weird. It'd be nice if you could just escape `.` like you can escape `{`.

eemeli:
Fixes #787
Fixes #837 

We currently allow complex messages to start with `{{`, so `hello` and `{{hello}}` mean the exact same thing. As identified in #787, when we add [bidi isolation](https://github.com/unicode-org/message-format-wg/blob/main/exploration/bidi-usability.md) to the syntax (#811), this will mean that each of the following will be valid sequences starting a complex message:
- `.`
- `{{`
- `LRI{{` or `{LRI{`
- `RLI{{` or `{RLI{`
- `FSI{{` or `{FSI{`

where LRI, RLI, and FSI represent the U+2066..U+2068 characters.

This seems like a lot, esp. when the only reason why a complex message must be valid even if it has no keywords is to allow for messages like `{{.zip}}`, i.e. ones starting with a literal `.`.

If we were to instead allow for `\.` as an escape, we could _require_ complex messages to start with a keyword, and then the set of valid sequences starting a complex message reduces to only:
- `.`

### DIFF
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -5,9 +5,9 @@ simple-start      = simple-start-char / escaped-char / placeholder
 pattern           = *(text-char / escaped-char / placeholder)
 placeholder       = expression / markup
 
-complex-message   = *(declaration [s]) complex-body [s]
+complex-message   = 1*(declaration [s]) quoted-pattern [s]
+                  / 0*(declaration [s]) matcher [s]
 declaration       = input-declaration / local-declaration / reserved-statement
-complex-body      = quoted-pattern / matcher
 
 input-declaration = input [s] variable-expression
 local-declaration = local s variable [s] "=" [s] expression
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -163,14 +163,14 @@ simple-start   = simple-start-char / escaped-char / placeholder
 
 A **_<dfn>complex message</dfn>_** is any _message_ that contains _declarations_,
 a _matcher_, or both.
-A _complex message_ always begins with either a keyword that has a `.` prefix or a _quoted pattern_
-and consists of:
+A _complex message_ always begins with a keyword that has a `.` prefix.
 
-1. an optional list of _declarations_, followed by
-2. a _complex body_
+_Complex messages_ with multiple variants use a _matcher_,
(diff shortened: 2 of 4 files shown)


lucacasonato:
For the record, I am relatively strongly against this. It makes refactoring from simple -> complex message more annoying, as you have to transition through an invalid / confusing state:

---

```
Hello, World!
```

Renders as `Hello, World!`

---

```
{{Hello, World!}}
```

Renders as `{{Hello, World!}}`

---

```
.local $foo = {1}
{{Hello, World!}}
```

Renders as `Hello, World!`

eemeli:
Giving up on this one; not worth the fight.

PR #844, opened 2024-07-30
Add more exhaustive syntax tests (#843)

catamorphism:
Beginning to address #843 and the corresponding item under https://github.com/unicode-org/message-format-wg/wiki/Things-That-Need-Doing .

I tried to test each variation of each rule (up to but not including `identifier`.) I didn't try to test combinations of variations. For example, for this rule:

```
complex-message   = *(declaration [s]) complex-body [s]
```

the idea would be to add tests that exercise the following four (partial) derivations:
```
complex-message -> complex-body
complex-message -> complex-body s
complex-message -> declaration complex-body
complex-message -> declaration s complex-body
```

This isn't exhaustive by any means, but it was broad enough that I found one bug in the ICU4J parser (the `".input{$x}{{}}` test case) and two bugs in the ICU4C parser (the `\\\\` test case and the `{ #a}` test case).

### DIFF
--- a/test/tests/syntax-errors.json
+++ b/test/tests/syntax-errors.json
@@ -123,6 +123,9 @@
       "src": "bad {:placeholder @attribute=@foo}"
     },
     {
+      "src": "{ @misplaced = attribute }"
+    },
+    {
       "src": "no {placeholder end"
     },
     {
--- a/test/tests/syntax.json
+++ b/test/tests/syntax.json
@@ -581,13 +581,11 @@
     },
     {
       "src": "hello { world\t\n}",
-      "exp": "hello world",
-      "expCleanSrc": "hello {world}"
+      "exp": "hello world"
     },
     {
       "src": "hello {\u3000world\r}",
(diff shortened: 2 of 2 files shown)


catamorphism:
The validation CI check is failing because I used the `comment` field, which I added to [the version of the schema in the conformance repo](https://github.com/unicode-org/conformance/blob/main/schema/message_fmt2/testgen_schema.json) -- I guess I'll go ahead and update that in the same PR.

eemeli:
Tests may already include a `"description"`, why not use that? https://github.com/unicode-org/message-format-wg/blob/4514e880a8690ba2dd78a70696b4c89db93697ba/test/schemas/v0/tests.schema.json#L135-L142

The spec tests in this repo need to match the schema that's also in this repo.

catamorphism:
> Tests may already include a `"description"`, why not use that?

Ah, didn't notice that. Fixed.

eemeli (approved):

PR #845, opened 2024-07-30
Accept attributes design & remove spec note

### LINKED ISSUE #642: tech preview question about attribute like behavior
From 2024-02-15 discussion, we need to get feedback in tech preview about locale, id, and other attribute-like behavior

eemeli:
Closes #642 

Following the acceptance of the "Expression Attributes" design yesterday, this notes that approval in the doc, and updates the _attributes_ accordingly so that they have no runtime impact.

A separate PR adding contextual `u:` options will follow.

### DIFF
--- a/exploration/expression-attributes.md
+++ b/exploration/expression-attributes.md
@@ -16,6 +16,7 @@ Status: **Accepted**
 		<dd><a href="https://github.com/unicode-org/message-format-wg/pull/780">#780</a></dd>
 		<dd><a href="https://github.com/unicode-org/message-format-wg/pull/792">#792</a></dd>
 		<dd><a href="https://github.com/unicode-org/message-format-wg/pull/845">#845</a></dd>
+		<dd><a href="https://github.com/unicode-org/message-format-wg/pull/846">#846</a></dd>
 	</dl>
 </details>
 
--- a/exploration/expression-attributes.md
+++ b/exploration/expression-attributes.md
@@ -1,6 +1,6 @@
 # Expression Attributes
 
-Status: **Proposed**
+Status: **Accepted**
 
 <details>
 	<summary>Metadata</summary>
@@ -15,6 +15,7 @@ Status: **Proposed**
 		<dd><a href="https://github.com/unicode-org/message-format-wg/pull/772">#772</a></dd>
 		<dd><a href="https://github.com/unicode-org/message-format-wg/pull/780">#780</a></dd>
 		<dd><a href="https://github.com/unicode-org/message-format-wg/pull/792">#792</a></dd>
+		<dd><a href="https://github.com/unicode-org/message-format-wg/pull/845">#845</a></dd>
 	</dl>
 </details>
 
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -58,6 +58,8 @@ or even use them in their internal processing,
 as long as the final _formatting_ result is made available to users
 and the observable behavior of the formatter matches that described here.
 
+_Attributes_ MUST NOT affect the processing or output of a _message_.
+
 ## Formatting Context
 
 A message's **_<dfn>formatting context</dfn>_** represents the data and procedures that are required
@@ -761,10 +763,6 @@ When formatting to a string, the default representation of all _markup_
 MUST be an empty string.
 Implementations MAY offer functionality for customizing this,
 such as by emitting XML-ish tags for each _markup_.
-
-_Attributes_ are reserved for future standardization.
-Other than checking for valid syntax, they SHOULD NOT 
-affect the processing or output of a _message_.
 
 ### Examples
 
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -714,7 +714,8 @@ markup = "{" [s] "#" identifier *(s option) *(s attribute) [s] ["/"] "}"  ; open
 > {#button}Submit{/button} or {#img alt=|Cancel| /}.
 > ```
 
-> A _message_ with options on the closing tags:
+> A _message_ containing _markup_ that uses _options_ to pair 
+> two closing markup _placeholders_ to the one open markup _placeholder_:
 >
 > ```
 > {#ansi attr=|bold,italic|}Bold and italic{/ansi attr=|bold|} italic only {/ansi attr=|italic|} no formatting.}
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -90,7 +90,7 @@ Attempting to parse a _message_ that is not _well-formed_ will result in a _Synt
 A _message_ is **_<dfn>valid</dfn>_** if it is _well-formed_ and
 **also** meets the additional content restrictions
 and semantic requirements about its structure defined below for
-_declarations_, _matcher_ and _options_.
+_declarations_, _matcher_, _options_, and _attributes_.
 Attempting to parse a _message_ that is not _valid_ will result in a _Data Model Error_.
 
 ## The Message
@@ -154,7 +154,7 @@ message = simple-message / complex-message
 
 A **_<dfn>simple message</dfn>_** contains a single _pattern_,
 with restrictions on its first character.
-An empty string is a valid _simple message_.
+An empty string is a _valid_ _simple message_.
 
 ```abnf
 simple-message = [simple-start pattern]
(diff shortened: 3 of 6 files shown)


exploration/expression-attributes.md
```diff
@@ -15,6 +15,7 @@ Status: **Proposed**
 		<dd><a href="https://github.com/unicode-org/message-format-wg/pull/772">#772</a></dd>
 		<dd><a href="https://github.com/unicode-org/message-format-wg/pull/780">#780</a></dd>
 		<dd><a href="https://github.com/unicode-org/message-format-wg/pull/792">#792</a></dd>
+		<dd><a href="https://github.com/unicode-org/message-format-wg/pull/845">#845</a></dd>
```
eemeli, line 18:
```suggestion
		<dd><a href="https://github.com/unicode-org/message-format-wg/pull/845">#845</a></dd>
		<dd><a href="https://github.com/unicode-org/message-format-wg/pull/846">#846</a></dd>
```

spec/formatting.md
```diff
@@ -58,6 +58,8 @@ or even use them in their internal processing,
 as long as the final _formatting_ result is made available to users
 and the observable behavior of the formatter matches that described here.
 
+_Attributes_ MUST NOT affect the processing or output of a _message_.
```
aphillips, line 61: This might be too broadly stated, since the purpose of attributes might be the "processing" of a message (as, for example, in tools). 

If we wanted to prevent attributes from having an effect, we'd be better off to say that implementations must not/should not expose attribute values to functions (if functions can't see the attributes, they can't do anything with them)?

Perhaps:

```suggestion
_Attributes_ have no effect on selection or formatting in a _message_.
```

... and perhaps adding:

> Implementations MUST NOT pass _attributes_ to _functions_.

spec/syntax.md
```diff
@@ -714,7 +714,7 @@ markup = "{" [s] "#" identifier *(s option) *(s attribute) [s] ["/"] "}"  ; open
 > {#button}Submit{/button} or {#img alt=|Cancel| /}.
 > ```
 
-> A _message_ with attributes in the closing tag:
+> A _message_ with options on the closing tags:
```
aphillips, line 717: The opening tag has options too. Perhaps a bit more prose to show why we have this example?

```suggestion
> A _message_ containing _markup_ that uses _options_ to pair 
> two closing markup _placeholders_ to the one open markup _placeholder_:
```

Question: should this example include `u:id`?

spec/syntax.md
```diff
@@ -728,66 +728,25 @@ on the pairing, ordering, or contents of _markup_ during _formatting_.
...
 An **_<dfn>attribute</dfn>_** is an _identifier_ with an optional value
 that appears in an _expression_ or in _markup_.
+During formatting, _attributes_ have no effect,
+and they can be treated as code comments.
 
 _Attributes_ are prefixed by a U+0040 COMMERCIAL AT `@` sign,
 followed by an _identifier_.
-An _attribute_ MAY have a _value_ which is separated from the _identifier_
+An _attribute_ MAY have a _literal_ _value_ which is separated from the _identifier_
 by an U+003D EQUALS SIGN `=` along with optional whitespace.
```
aphillips, line 739: You removed the possibility that an attribute can have a variable as its value. Why? I agree that passing a variable might not be that useful (given the restrictions we are placing on attributes). But a use case might be found in tooling that uses the resolved value of some variable to assign an attribute value. For example, it might be used to set a debug level:

```
You have {$foo :function @debug=$debugLevel}
```

spec/syntax.md
```diff
@@ -728,66 +728,25 @@ on the pairing, ordering, or contents of _markup_ during _formatting_.
...
-The _value_ of an _attribute_ can be either a _literal_ or a _variable_.
 
 Multiple _attributes_ are permitted in an _expression_ or _markup_.
 Each _attribute_ is separated by whitespace.
+Each _attribute_'s _identifier_ MUST be unique within the _expression_ or _markup_:
+an _expression_ or _markup_ with duplicate _attribute_ _identifiers_ is not valid.
```
aphillips, line 744: ... and results in a Data Model Error?

spec/syntax.md
```diff
@@ -728,66 +728,25 @@ on the pairing, ordering, or contents of _markup_ during _formatting_.
...
 
 The order of _attributes_ is not significant.
 
-
 ```abnf
-attribute = "@" identifier [[s] "=" [s] (literal / variable)]
+attribute = "@" identifier [[s] "=" [s] literal]
```
aphillips, line 749: See above

aphillips (commented):
Good start. Thanks for saving me some work 😉 

A few comments.

spec/formatting.md
eemeli: I think we need a slightly wider ban than attributes' visibility to functions, to prohibit e.g. the formatted parts representation of a message from reflecting attribute values.

Would this work?

```suggestion
_Attributes_ MUST NOT have any effect on the formatted output of a _message_,
or be made available to function implementations.
```

spec/syntax.md
eemeli: Variable values are removed because that's a part of the design we accepted: https://github.com/unicode-org/message-format-wg/blob/main/exploration/expression-attributes.md#attributes

spec/syntax.md
eemeli: This is using the same language we have for options: https://github.com/unicode-org/message-format-wg/blob/4514e880a8690ba2dd78a70696b4c89db93697ba/spec/syntax.md?plain=1#L580-L581

The Data Model Error for invalid content is already mentioned here: https://github.com/unicode-org/message-format-wg/blob/4514e880a8690ba2dd78a70696b4c89db93697ba/spec/syntax.md?plain=1#L90-L94

But we do indeed need to add a note to the above pointing here as well.

spec/formatting.md
aphillips: I suggested a no-MUSTard version in part because I don't know all the use cases that an expression attribute might be put to and in part because I don't know how to test it. If an implementation decided to support some magical attributes, how would it be harmful if (for example) they affected the formatted parts? We don't define any expression attributes, so we could leave it up to implementations to decide what they mean. The statement I suggest should guide implementers away from using attributes as some sort of option factory.

spec/syntax.md
eemeli: Ah. That comes from the first paragraph of the [Proposed Design](https://github.com/unicode-org/message-format-wg/blob/main/exploration/expression-attributes.md#proposed-design):
> Provide separate solutions for attributes that impact formatting, and those which do not, so that their namespaces are not comingled.

From that starting point, the `u:` options have runtime impact, while `@` attributes do not. Therefore, if it's not possible for `@` attributes to do anything during formatting, it does not make sense to allow them to have variable values, reading which could be externally observable and could generate errors. Especially as the use cases `@` attributes are serving are all non-runtime ones, which do not have access to variable values.

catamorphism (approved):
Looks consistent with what we agreed on.

spec/formatting.md
eemeli: > If an implementation decided to support some magical attributes, how would it be harmful if (for example) they affected the formatted parts?

The MUST NOT establishes a promise that when a message is being processed for formatting only, all of the attributes can be trimmed out. This allows for potentially very verbose content to be included in attributes, given that doing so has no impact on the runtime size or parsing requirements.

Without the mustard, it is much more difficult to make this guarantee, as tools can't be certain about what they can and cannot trim out. It would also significantly fuzzify the boundary between "options" and "attributes", if they can have the same effect.

macchiati: So @locale=fr cannot be an attribute, right? Because it would affect the results of formatting.

It sounds like 'attributes' are really something like 'metadata'

macchiati: Ah, sounds like u: options now handle cases like locale settings?

eemeli: > So @Locale=fr cannot be an attribute, right? Because it would affect the results of formatting.

That's right.

> It sounds like 'attributes' are really something like 'metadata'

Yes. I would be entirely happy with renaming them as metadata, in case there's a sense that that would make it clearer.

> Ah, sounds like u: options now handle cases like locale settings?

Yes, PR #846 introduces a [`u:locale`](https://github.com/unicode-org/message-format-wg/blob/3a3d4abd84d074509eb181865e9fc0f1f596a33d/spec/registry/unicode.md#ulocale) option for just that.

aphillips (approved):

aphillips:
This was approved in the 2024-08-20 telecon, but not merged then due to a conflict resolved in f7b75fc. Merging now.

PR #846, opened 2024-07-30
Add u: options namespace

### LINKED ISSUE #452: Constraints on @locale values
In #450 we are considering the addition of a contextual _expression attribute_ for the locale. Currently this says that the values must be a BCP47 language tag (or comma-separated sequence of tags, e.g. a _language priority list_). We should agree on the level of validation for conformance.

Generally speaking, we should probably require well-formed tags (according to BCP47's understanding of well-formed), at least at the syntax level. 
We could require that the tags be valid (this requires checking that the subtags are in the registry)
We could require that the tags be valid Unicode Locale Identifiers (ULIs) (which has further canonicalization requirements)

Writing this issue to provide a separate place to discuss.

eemeli:
Closes #452
See also #845

Following the registry separations proposed in #634, this PR adds a new folder `spec/registry/` and moves the current default function definitions there as `default.md`, as well as adding a new `u:` Unicode Registry as `unicode.md`.

Initially, this includes only the `u:id`, `u:locale`, and `u:dir` definitions from the [design doc](https://github.com/unicode-org/message-format-wg/blob/main/exploration/expression-attributes.md).

A new **_function context_** definition is added to Function Resolution, to allow for it to be affected by the `u:` options. This is intended to encapsulate the parts of the _formatting context_ that are made available when calling functions.

Regarding the options, the text is almost directly as in the design doc. For `u:locale`, I specified that literal values matching the `langtag` rule from [RFC 5646](https://www.rfc-editor.org/rfc/rfc5646) are always supported, but that other tags and locale definitions MAY also be supported. This is roughly in line with what we have for e.g. `:datetime` operands, where we define something relatively strict that will work ~everywhere, while allowing other values beyond that definition to also be processed without error.

Note that `u:locale` and `u:dir` are ignored when used on markup.

A new test suite `unicode.json` is added; this also ends up testing our [Default Bidi Strategy](https://github.com/unicode-org/message-format-wg/blob/main/spec/formatting.md#handling-bidirectional-text).


spec/README.md
```diff
@@ -16,7 +16,8 @@
    1. [Data Model Errors](errors.md#data-model-errors)
    1. [Resolution Errors](errors.md#resolution-errors)
    1. [Message Function Errors](errors.md#message-function-errors)
-1. [Default Function Registry](registry.md)
+1. [Default Function Registry](registry/default.md)
+1. [`u:` Unicode Registry](registry/unicode.md)
```
aphillips, line 20: My design document is calling this the "Unicode Reserved Namespace" and "Unicode Reserved Namespace Registry", since, technically, all of the registries belong to Unicode.

spec/formatting.md
```diff
@@ -214,9 +214,21 @@ the following steps are taken:
...
+   This includes:
 
-   - The current _locale_.
+   - The current _locale_,
+     potentially including a fallback chain of locales.
+   - The base directionality of the _message_ and its _text_ tokens.
+
+   If the resolved mapping of _options_ includes any `u:` options
+   supported by the implementation,
+   process them as specified in the [Unicode Registry](/spec/registry/unicode.md).
```
aphillips, line 226: Unicode Registry => Unicode Reserve Namespace registry

spec/formatting.md
```diff
@@ -214,9 +214,21 @@ the following steps are taken:
...
-4. Call the function implementation with the following arguments:
+4. Determine the **_<dfn>function context</dfn>_** for calling the function implementation.
+   This includes:
 
-   - The current _locale_.
+   - The current _locale_,
+     potentially including a fallback chain of locales.
+   - The base directionality of the _message_ and its _text_ tokens.
+
+   If the resolved mapping of _options_ includes any `u:` options
```
aphillips, line 224: This seems like special pleading on our part. The `u:` namespace is "just a namespace"?

Perhaps:

```suggestion
Implementations are encouraged to support _options_ defined in
the Unicode Reserved Namespace (`u:`).
```

spec/formatting.md
```diff
@@ -344,6 +356,11 @@ The resolved value of _markup_ includes the following fields:
 - The _identifier_ of the _markup_
 - The resolved _options_ values after _option resolution_.
 
+If the resolved mapping of _options_ includes any `u:` options
+supported by the implementation,
+process them as specified in the [Unicode Registry](/spec/registry/unicode.md).
+Such `u:` options MAY be removed from the resolved mapping of _options_.
```
aphillips, line 362: See above

spec/registry/unicode.md
```diff
@@ -0,0 +1,63 @@
+# MessageFormat 2.0 Unicode Registry
```
aphillips, line 1:
```suggestion
# MessageFormat 2.0 Unicode Reserved Namespace Registry
```

spec/registry/unicode.md
```diff
@@ -0,0 +1,63 @@
...
+
+The `u:` namespace is reserved for use by the Unicode Consortium.
```
aphillips, line 3: Perhaps quote the design document here?

```suggestion
This registry is for items in the namespace `u:` 
which is reserved for use by the Unicode Consortium. 
This registry can contain _functions_ or _options_. 
Implementations are not required to implement entries found in this registry. 

Items in this registry are stable and subject to stability guarantees. 
This registry might sometimes be used to incubate functionality before promotion to the RGI or default registry in a future release. 
In such cases, the `u:` namespace version is retained, but deprecated.
```

spec/registry/unicode.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Options
+
+This section describes common options which each implementation SHOULD support
+for all _functions_ and _markup_.
```
aphillips, line 8: Perhaps:

```suggestion
This section describes _options_ which are intended to be common to all _functions_.
Implementations SHOULD support resolving and passing these _options_.
_Function_ authors SHOULD implement each _option_ as described here to ensure interoperability.
```

spec/registry/unicode.md
```diff
@@ -0,0 +1,63 @@
...
+
+### `u:id`
+
+A string value that is included as an `id` or other suitable value
+in the formatted parts for the _placeholder_,
+or any other structured formatted results.
+
+Ignored when formatting a message to a string.
+
+Accepts string values, or values which can be stringified without error.
```
aphillips, line 18: Should this say it must have a value? Is the empty literal, for example, acceptable?

spec/registry/unicode.md
```diff
@@ -0,0 +1,63 @@
...
+For other values, a _Bad Option_ error is emitted
+and the `u:id` option is ignored.
+
+### `u:locale`
+
+A comma-delimited list of BCP 47 language tags,
+or an implementation-defined list of such tags.
```
aphillips, line 25: Perhaps (because ULI's are a specific narrowing of the validity rules and we are, after all, the CLDR-TC):

```suggestion
A comma-delimited prioritized list of BCP47 language tags
which are expected to be Unicode Locale Identifiers,
or an implementation-defined list of locale identifiers.
```

spec/registry/unicode.md
```diff
@@ -0,0 +1,63 @@
...
+
+Replaces the _locale_ defined in the _function context_ for this _expression_.
+The value is ignored when set on _markup_.
```
aphillips, line 28: Why? Why not leave it up to the implementer?

spec/registry/unicode.md
```diff
@@ -0,0 +1,63 @@
...
+
+During processing, the `u:locale` option
+is always removed from the resolved mapping of _options_.
```
aphillips, line 31: Why? Because it replaces the value in the function context? Since nothing can overwrite it (it's namespaced), why not pass it?

spec/registry/unicode.md
```diff
@@ -0,0 +1,63 @@
...
+
+Values matching the following ABNF are always accepted:
+```abnf
+u-locale-option = langtag *([s] "," [s] langtag)
+```
+using `langtag` as defined in [RFC 5646](https://www.rfc-editor.org/rfc/rfc5646).
```
aphillips, line 37: So let's not link to a specific RFC! BCP47 only has the one `langtag` production in it. Link to BCP47.

spec/registry/unicode.md
```diff
@@ -0,0 +1,63 @@
...
+
+Implementations MAY support additional language tags,
+such as private-use or grandfathered tags,
+or tags using `_` instead of `-` as a separator.
+When the value of `u:locale` is set by a _variable_,
+implementations MAY support non-string values otherwise representing locales.
```
aphillips, line 43: Some problems here. Private-use and grandfathered tags are already covered by both the `langtag` production and BCP47 in general.

I think you're trying to do several things here and we should split them apart:

```suggestion
Implementations MAY process the list of locale identifiers,
including interpretation of language tags or canonicalization of values 
(such as mapping grandfathered tags to modern representations).

Implementations MAY convert or map the list to a prioritized list of
implementation-specific values.
For example, a JavaScript implementation might convert it into an `Intl.Locale` 
while a Java implementation might convert the list into an array of `java.util.Locale` objects.
Implementations MAY reject unrecognized or unsupported values.

Implementation MAY support proprietary or implementation-specific
locale identifiers.
```

spec/registry/unicode.md
```diff
@@ -0,0 +1,63 @@
...
+
+For unsupported values, a _Bad Option_ error is emitted
+and the value of the `u:locale` option is ignored.
+
+### `u:dir`
+
+Replaces the base directionality defined in
+the _function context_ for this _expression_.
+The value is ignored when set on _markup_.
```
aphillips, line 52: Why?

spec/registry/unicode.md
aphillips: (Note: this comment is about line 28)

spec/formatting.md
eemeli: It's not "just a namespace", because it needs special powers to affect the _function context_, though.

spec/registry/unicode.md
eemeli: The first paragraph is mostly fine, but as #634 is only a PR at this time, it's way too early to refer to the RGI, for instance, or talk about incubation. We can add this language later, when we get consensus on these processes and their names.

I would also find it a bit odd to say e.g. that this "can contain _functions_", when AFAIK no such `u:` function has been proposed by anyone.

spec/registry/unicode.md
eemeli: The proposed language doesn't quite match the [accepted design](https://github.com/unicode-org/message-format-wg/blob/main/exploration/expression-attributes.md#contextual-options), though. These are not intended to be options that are handled by each function, but by the implementation, and they e.g. apply changes to the function context, so that function authors explicitly do not need to do anything to enable their use.

spec/registry/unicode.md
aphillips: Permission to do something does not obligate us to actually do it. No one has proposed a `u:` function, but there's no reason to forbid them.

---

Are you disagreeing with the design in #634? More specifically, do you disagree with my thinking about how to organize the registries? I do think we should lock-step these efforts to some degree. We'll probably rename RGI somewhere along the way, for example. But I think we'll likely end up with the same concepts at least.

We could replace my line 9 above with something like:

> Items in this registry might be deprecated but will not be removed.

spec/registry/unicode.md
eemeli: An empty literal is an empty string, and is allowed by this language. This is intentional, as it's possible for a message containing a `u:id=$x` to sometimes need formatting without an identifier value, for which an empty string is a good stand-in.

spec/registry/unicode.md
eemeli: Unfortunately, ULI is not only a narrowing of the validity rules, it also explicitly allows for `_` as a separator, so e.g. `en_US` is a valid ULI but not a valid BCP47 tag. This is not universal, see e.g. tc39/ecma402#777 for an open discussion of supporting this in JS.

spec/registry/unicode.md
eemeli: Because that could introduce surprising differences in implementation behaviour and/or require us to define how markup open/close matching works. Consider this example, formatted as `en`:
```
French: {#span u:locale=fr}{$n :number}{/span}
English: {$n :number}
```

If we don't explicitly ignore `u:locale` and `u:dir` on markup, I would count it as likely that some implementations would format the first number in English, and others in French.

If that is desirable behaviour, then we would need to define open/close matching to ensure that the second number is still formatted in English.

spec/registry/unicode.md
eemeli:
```suggestion
using `langtag` as defined in [BCP 47](https://www.rfc-editor.org/rfc/bcp/bcp47.txt).
Note that `langtag` is the rule for "normal language tags",
and does not include private-use or grandfathered tags.
```

spec/registry/unicode.md
aphillips: Wouldn't the `u:locale` only affect the markup placeholder? Why would it affect the enclosed `$n` placeholder?

spec/registry/unicode.md
aphillips: https://www.rfc-editor.org/rfc/bcp/bcp47.txt should be https://www.rfc-editor.org/bcp/bcp47

spec/registry/unicode.md
eemeli: I updated the previous paragraph to note that as we're specifically referring to the `langtag` rule and not the `Language-Tag` rule, we're not requiring support for private-use or grandfathered tags.

The intended point of this passage is to make sure that values not matching `u-locale-option` may still be accepted, and explicitly calling out the extensions that are required for either BDP47 or ULI support.

spec/registry/unicode.md
eemeli: See thread above on `u:locale`.

spec/registry/unicode.md
eemeli: I am in general agreement about having definitions for non-required functions and options, but I'm honestly not really sure what the current state of #634 is, and whether e.g. the registry development process added to the PR last Friday is final, or whether it's being iterated on. When we last discussed this on [a call in May](https://github.com/unicode-org/message-format-wg/blob/4514e880a8690ba2dd78a70696b4c89db93697ba/meetings/2024/notes-2024-05-06.md#pr-634), it was "Still in progress".

I'm also not sure whether the current design doc proposal's language "Implementations are not required to implement any values found in this registry and may adopt or ignore registry entries at their discretion." ought to be strengthened into something more like a SHOULD.

My preference would be to not lockstep processes that do not strictly need lockstep, and to focus first on the concrete (i.e. this PR), and use that experience to inform the registry maintenance plan. On which we ought to have a separate discussion, if/once it's in a stable state.

spec/registry/unicode.md
eemeli: It's something that at least @mihnita has expressed some interest in, and if we leave it up to implementations, that becomes a possible interpretation of what happens when spanning markup has `u:` options set on it.

spec/formatting.md
mihnita: I think that the "function context" context distinction is an implementation detail.

In most formatters implementations the locale and the options on how to format are passed to the constructor as parameters, at the same time.

I would rather look at this as "universal function parameters" that might be recognized and honored by several / all functions.

spec/formatting.md
mihnita: > Such `u:` options MAY be removed from the resolved mapping of _options_.

?
Why?

spec/registry/unicode.md
mihnita: Why functions?

The idea for the parameters is "u: parameters are potentially recognized / honored by all functions"
What would be the meaning of a "u:" function?

If we think "u:function" is a Unicode function, it is redundant.
All functions in the standard registry are Unicode functions, since this is a Unicode spec.

spec/formatting.md
aphillips: I think @eemeli is thinking that they've "done their job", e.g. if you have:

```
.input {$foo :datetime u:locale=fr}
```

... then the implementation has read the `u:locale` option and set the locale of `$foo`'s `:datetime` invocation to French (in the formatting context). In such a case, `:datetime` doesn't need to know about this. In most cases, I think that's correct (if the function wants to know about the locale or direction or id, it can read it from the formatting context). But I could see a function or markup wanting access to the `u:id`. I can also imagine `u:xxx` options in the future where that might not be the right thing to do. Finally, this creates work for implementations (although it is only `MAY` and not required work).

spec/registry/unicode.md
aphillips: In your example messages, the difference between the two is some markup. The formatting function should still format using the invoking (contextual) locale. For example, the invoking locale might be `fr-CA-u-nu-cans` and you wouldn't want the markup's `lang` attribute to remove the additional formatting (the example is contrived: I don't believe that the `Cans` script has its own digits). If you wanted to trim the locale, you should put it on the `:number` invocation. Otherwise markup can cause the formatting to change, which is spooky--and unlike anything else in MF2.

spec/registry/unicode.md
aphillips: I don't think this is the right move. We don't need to get into the locale identifier game. I think referencing `langtag` instead of `Language-Tag` is a backwards step.

Do you not agree with the various changes I suggested?

spec/registry/unicode.md
aphillips: Shouldn't I be able to write:

```
I like {#html:strong u:dir=ltr}ASCII{/html:strong} as long as it's UTF-8.
```

And have it produce:

```html
I like <strong dir=ltr>ASCII</strong> as long as it's UTF-8.
```

eemeli: You should write instead:
```
I like {#html:strong dir=ltr}ASCII{/html:strong} as long as it's UTF-8.
```
as the direction you're seeking to control is the direction of the contents of the `<strong>` rather than the tag itself. This is rather similar to my example above in https://github.com/unicode-org/message-format-wg/pull/846#discussion_r1697345596:
```
{#span u:locale=fr}{$n :number}{/span}
```

If we allow `u:dir` and `u:locale` to be used as stand-ins for the HTML `dir` and `lang` attributes, we introduce unnecessary confusion about whether or not these options apply to the span between the open and close elements.

spec/registry/unicode.md
eemeli: I agree that we should not get into the locale identifier game. So much as we did with `:datetime` literals, I think we should identify a minimal definition that must be supported, and allow for implementations to go beyond that. And I think the `langtag` rule offers that minimum, i.e. something that can be supported by both a BCP47 and a ULI implementation, where both of those have features unsupported by the other.

spec/formatting.md
aphillips: I think the problem here is that the **_current set_** of `u:` options behaves that way, but we might introduce one that isn't function context affecting in the future? If our intention to is to _require_ that all such options be context-affecting, we should set that as a requirement in the design doc and elsewhere. We might need to contemplate an additional namespace in the future as well, although I can't think of any universal options just at the moment that we wouldn't just put in the default namespace.

Note: I am not disagreeing with doing this. Just making sure we're consistent and clear about it.

macchiati (changes requested):
We should be referencing the Unicode BCP 47 locale identifier, not [RFC 5646](https://www.rfc-editor.org/rfc/rfc5646). (In particular, we don't want to force people to support extlang.

aphillips:
@macchiati Can you be specific? I think the 5646 reference is to the `langtag` grammar. We can (and should) refer to Unicode Locale for locale stuff.

macchiati:
Here is a reference. https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#BCP_47_Language_Tag_Conversion

I've felt for a while that we need to hoist the Unicode BCP 47 locale identifier definition, I think we could do that in this release.

aphillips:
That's not what I mean. Where in the text does it need to change?

spec/registry/unicode.md
macchiati: It needs stronger language than "which are expected to be".

```suggestion
A comma-delimited prioritized list of BCP47 language tags that are also [valid Unicode CLDR locale identifiers](https://www.unicode.org/reports/tr35/#BCP_47_Conformance).
```

I think allowing "implementation-defined locale identifiers" is dangerous, and will lead to interoperability problems. What is the thinking behind allowing these? `locale:|The King's English|`

macchiati (changes requested):

aphillips:
@macchiati noted:
> I think allowing "implementation-defined locale identifiers" is dangerous,

The text doesn't allow that. It allows the list to be formatted in an implementation defined way, e.g. a `List<String>` or `Intl.Locale[]`. This might be phrased better to make it clear, though.

I think the are-also formulation is overkill. We can just say "valid CLDR locale identifiers", although that places some burden on implementations that can guarantee `langtag` but don't have a local implementation of the CLDR rules. Well-formed tags might be acceptable in certain cases.

eemeli:
Resolved conflicts with the `main` branch and renamed the files to not refer to a "registry", but to a "namespace".

spec/formatting.md
```diff
@@ -260,9 +260,20 @@ the following steps are taken:
...
+   This includes:
 
-   - The current _locale_.
+   - The current _locale_,
+     potentially including a fallback chain of locales.
+   - The base directionality of the _message_ and its _text_ tokens.
+
+   If the resolved mapping of _options_ includes any _`u:` options_
+   supported by the implementation, process them as specified.
+   Such `u:` options MAY be removed from the resolved mapping of _options_.
```
aphillips, line 272: I think we should be more disciplined about defining terms.

```suggestion
4. Determine the _function context_ for calling the function implementation.
   
   The **_<dfn>function context</dfn>_** contains the context necessary for
   the _function handler_ to perform formatting or selection. It includes:
   - The current _locale_,
     potentially including a fallback chain of locales or locale identifiers.
   - The base directionality of the _message_ and its _text_ tokens.

   If the resolve mapping of _options_ includes any _`u:` options_
   supported by the implementation, process them as specified.
   Such `u:` options MAY be removed from the resolve mapping of _options_.
```

Question: should "function implementation" be "function handler"?

spec/u-namespace.md
```diff
@@ -0,0 +1,65 @@
...
+### `u:locale`
+
+A comma-delimited list of BCP 47 language tags,
+or an implementation-defined list of such tags.
+
+Replaces the _locale_ defined in the _function context_ for this _expression_.
+The value is ignored when set on _markup_.
+
+During processing, the `u:locale` option
+is always removed from the resolved mapping of _options_.
```
aphillips, line 31: Tighten a bit?

```suggestion
is removed from the _resolved options_.
```

spec/u-namespace.md
```diff
@@ -0,0 +1,65 @@
...
+
+Values matching the following ABNF are always accepted:
+```abnf
+u-locale-option = langtag *([s] "," [s] langtag)
```
aphillips, line 35: `[s]` is the same as `o`. Use that.

I don't agree with using `langtag` over using `Language-Tag`. Grandfathered tags all have modern replacements (or are deprecated outright) and private use is the user's problem. There's no reason for us to inject additional checking gorp onto the language tag space.

spec/u-namespace.md
```diff
@@ -0,0 +1,65 @@
...
+Ignored when formatting a message to a string.
+
+Accepts string values, or values which can be stringified without error.
+For other values, a _Bad Option_ error is emitted
+and the `u:id` option is ignored.
+
+### `u:locale`
+
+A comma-delimited list of BCP 47 language tags,
+or an implementation-defined list of such tags.
```
aphillips, line 25: The below text around line 33 suggests that that the language tags are supposed to be well-formed.

```suggestion
A comma-delimited list consisting of
well-formed [BCP 47](https://www.rfc-editor.org/rfc/bcp/bcp47.txt)
language tags,
or an implementation-defined list of such tags.
```

spec/u-namespace.md
```diff
@@ -0,0 +1,65 @@
...
+implementations MAY support non-string values otherwise representing locales.
+
+For unsupported values, a _Bad Option_ error is emitted
+and the value of the `u:locale` option is ignored.
+
+### `u:dir`
+
+Replaces the base directionality defined in
+the _function context_ for this _expression_.
+The value is ignored when set on _markup_.
```
aphillips, line 54: This is probably a mistake. Instead of ignoring it, we should say the behavior is undefined (an implementation of markup might make use of it?)

aphillips (commented):
Looking good. Some suggestions.

spec/formatting.md
eemeli: > Question: should "function implementation" be "function handler"?

Yes, this text is from before the latter term was introduced.

> The **_<dfn>function context</dfn>_** contains the context necessary for
> the _function handler_ to perform formatting or selection.

This bit should perhaps be reworded a bit, as with our current language technically the _function handler_ doesn't do the formatting or selection, but returns a _resolved value_, which then either does or enables formatting and/or selection to happen.

> the resolve mapping of _options_

Is the suggestion to not say "resolved mapping" intentional? That phrasing seems a bit awkward.

spec/u-namespace.md
eemeli:
```suggestion
u-locale-option = langtag *(o "," o langtag)
```

spec/u-namespace.md
eemeli: Two thoughts here:
1. The "always" is intended to ensure that `u:locale` is removed from the options even if this bit from below applies:
   > For unsupported values, a _Bad Option_ error is emitted
   > and the value of the `u:locale` option is ignored.

   Without the "always", is it clear enough that it's always removed?
2. We don't define the term _resolved options_. This is what we have: https://github.com/unicode-org/message-format-wg/blob/e9fc1367cbf496897736e68b0a29fd3ace229759/spec/formatting.md?plain=1#L309-L310

spec/u-namespace.md
eemeli: > There's no reason for us to inject additional checking gorp onto the language tag space.

I agree with this sentiment, but I think the current language is effectively required to allow for that. Note that we also include this further below:

> Implementations MAY support additional language tags,
> such as private-use or grandfathered tags,
> or tags using `_` instead of `-` as a separator.

With `langtag` as a base, we allow but do not require grandfathered or private-use tags to be supported, while also allowing but not requiring Unicode locale identifiers to be supported.

The overall intent here is to require the intersection of BCP47 and ULI to be supported, and allow extension from there, so that each implementation can use whatever is its native language tag format, and doesn't need to inject additional checking gorp onto its language tag space.

spec/u-namespace.md
eemeli: I disagree; see my previous reply to this: https://github.com/unicode-org/message-format-wg/pull/846#discussion_r1721885800.

spec/formatting.md
aphillips: > Is the suggestion to not say "resolved mapping" intentional? That phrasing seems a bit awkward.

No, that's a typo. Apparently, I'm very consistent about that specific mistake. Not sure if we should use the term _resolved options_ instead of "resolved mapping of _options_"

spec/u-namespace.md
aphillips: WRT (1), I think the always is clear because "is removed" leaves no wiggle room. But I could live with always.

WRT (2), ah... I see. Perhaps:

```suggestion
MUST be removed from the _options_ in the _resolved value_.
```

That's stronger than _is_ and clarifies always. I do find that the phrasing is inelegant though and we could just ignore this comment.

spec/u-namespace.md
aphillips: My tendency has been to require "well-formed" language tags (including allowing grandfathered and private-use) but encourage "valid" tags. Does the following capture the intention:

Implementations:
- MUST support well-formed language tags
- SHOULD check for valid language tags
- SHOULD canonicalize language tags and SHOULD ensure that tags are ULIs
- MAY omit, reject, or skip individual invalid tags
- MAY support additional implementation-defined identifiers
- MAY support implementation-defined locale objects
- MUST omit language tags from the resolved value

Users:
- SHOULD use valid canonicalized ULIs

I notice in making this list that we don't say whether the `language` attribute of the formatted value should be modified to match the selected `u:locale`.

spec/u-namespace.md
aphillips: In your previous comment, note that it doesn't prevent people from using it. This just says that it's not effective. My thinking is that we don't have a ton of implementation experience in this space. By not forbidding it (making it a Bad Option) we leave it in a kind of nether space ("valid" but ineffectual and tempting to abuse).

spec/u-namespace.md
eemeli: > MUST support well-formed language tags

This is problematic at least in a JS context, as it would require that e.g. a grandfathered tag like `i-klingon` is supported, and it currently is not: https://tc39.es/ecma402/#sec-language-tags

It is not reasonable to require an implementation to extend or modify its language tag rules to accommodate MF2.

spec/u-namespace.md
eemeli: I'd be fine emitting a Bad Option if this or `u:locale` is set on markup.

spec/u-namespace.md
aphillips: You're right. Support is the wrong word, at least from the point of view of "locales that do something useful". My proposal should probably say "MUST accept well-formed language tags". (Note that canonicalization would turn `i-klingon` into `tlh`). The other SHOULD/MAY requirements allow well-formed but invalid tags to not be supported (consistent with EMCA-402). But a naive implementation could exist that accepted any seemingly-possible language tag (`xxx-Yyyy-ZZ-variant`)

macchiati: I would stongly suggest referencing instead https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#unicode_bcp47_locale_id

It has a slightly narrower syntax than BCP47; see https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#bcp-47-conformance

(Note: once the spec goes final, these links will be modified to be of the form https://www.unicode.org/reports/tr35/xxx/tr35.html#yyy

eemeli: Referencing [`unicode_bcp47_locale_id`](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#unicode_bcp47_locale_id) would be fine as well; it looks like it's effectively the same as the current `langtag`?

catamorphism (changes requested):

spec/u-namespace.md
aphillips: "MUST accept" means that any well-formed language tags can be passed in the language priority list. The range error in Intl happens later--it's allowed by "MAY omit, reject, or skip". The point is to be somewhat permissive about passing in things that might be language tags and then allowing implementations some leeway in disposing of them. In practice, no one is using `i-klingon` or `i-default` or `cel-gaulish`.

eemeli: Now I'm a bit confused. With what you propose, would it be acceptable for an implementation to emit a Bad Option error if `i-klingon` is used as a `u:locale` value?

aphillips: Yes, although possibly not fatally.

The point of these attributes is to set the locale for some formatting. Ideally, we should allow any value that can appear in e.g. HTML's `lang` attribute. But we can't enforce that such a locale "works" or has data installed on a given implementation or system. So if you pass `u:locale=cel-gaulish` (or `u:locale=$celGaulish`), you probably get Bad Option and a formatted message using whatever the contextual locale is. FWIW, an implementation might do the same for non-range-erroring `aaa-Yyyy-RR` (not valid) or even the valid tag `tlh-Cyrl-AQ`. All of the canonicalization stuff we point to is in the service of making tags around the margins marginally useful.

macchiati: What I propose is that 

1. There be a [ VC: ] constraint that the syntax for locale identifiers conforms to BCP47 so that we don't get complete garbage.
2. Implementations should disregard any locale identifier that they don't support (that is, not raise an error)
3. Implementations should follow the semantics of locale identifiers as in https://www.unicode.org/reports/tr35/tr35-general.html#unicode_bcp47_locale_id  
    - **NB: when moved to the CLDR spec, this should be a relative link tr35.html#unicode_bcp47_locale_id**

aphillips: I can live with that.

macchiati: @eemeli ?

eemeli: As I mentioned before, I'd be fine referencing `unicode_bcp47_locale_id` instead of `langtag`, but I'm not really sure how the further proposed changes would affect the text.

We already do not consider errors on options to be fatal, so I'd much rather not further restrict implementations from emitting errors for unsupported locales.

I'm going to apply an update, and then ask for another review pass.

spec/registry/unicode.md
eemeli: Replaced `langtag with `unicode_bcp47_locale_id`.

spec/registry/unicode.md
eemeli: Made using `u:dir` and `u:locale` on markup an error.

spec/u-namespace.md
eemeli: Changed this to emit an error.

spec/u-namespace.md
eemeli: Applied some MUSTard.

spec/registry/unicode.md
eemeli: This is now an error.

spec/u-namespace.md
```diff
@@ -0,0 +1,72 @@
+# MessageFormat 2.0 Unicode Namespace
+
+The `u:` namespace is reserved for use by the Unicode Consortium.
```
aphillips, line 3: Here's where the definition goes, I guess? Perhaps:

```suggestion
The `u:` _namespace_ is reserved for the definition of _options_
for use in assisting the translation process 
or which affect the _function context_ of the specific _expressions_
in which they appear.
The CLDR Technical Committee of the Unicode Consortium
manages the specification for this namespace, hence the name `u:`.
```

spec/u-namespace.md
```diff
@@ -0,0 +1,72 @@
...
+
+### `u:id`
+
+A string value that is included as an `id` or other suitable value
+in the formatted parts for the _placeholder_,
+or any other structured formatted results.
+
+Ignored when formatting a message to a string.
+
+Accepts string values, or values which can be stringified without error.
```
aphillips, line 18: The option is not a function, so it doesn't really "accept" anything?

```suggestion
The value of the `id` _option_ can be a _literal_ or a
_variable_ whose _resolved value_ is either a string
or can be resolved to a string without error.
```

macchiati (approved):

spec/u-namespace.md
```diff
@@ -0,0 +1,72 @@
...
+[Unicode Locale Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#unicode_bcp47_locale_id).
+
+Implementations MAY support additional language tags,
+such as private-use or grandfathered tags,
+or tags using `_` instead of `-` as a separator.
+When the value of `u:locale` is set by a _variable_,
+implementations MAY support non-string values otherwise representing locales.
+
+For unsupported values, a _Bad Option_ error is emitted
+and the value of the `u:locale` option is ignored.
```
aphillips, line 52: Unsupported is unclear here. We also need to resolve list handling.

```
{$foo :date u:locale=|en-US, tlh-Xsux-AQ, i-klingon, bad-Juju, American|}`
```

Does the above work because en-US is fine? Or emit Bad Option because:
- `tlh-Xsux-AQ` a valid tag for Klingon as written in cuneiform as used in Antarctica is not an installed locale
- `i-klingon` is grandfathered and this implementation doesn't support grandfathered tags
- `bad-Juju` is not a valid tag
- `American` is not well-formed

```suggestion
If any of the language tags in the value is unsupported,
invalid, or not well-formed, the implementation ignores that entry in the
list and emits a _Bad Option_ error.
If the resulting list is empty, a _Bad Option_ error is emitted
and the value of the `u:locale` option is ignored.
```

spec/u-namespace.md
```diff
@@ -0,0 +1,72 @@
...
+and the `u:id` option is ignored.
+
+### `u:locale`
+
+A comma-delimited list consisting of
+well-formed [BCP 47](https://www.rfc-editor.org/rfc/bcp/bcp47.txt)
+language tags,
+or an implementation-defined list of such tags.
+
+Replaces the _locale_ defined in the _function context_ for this _expression_.
```
aphillips, line 29: The other two say "what" first and then tackle the contents. Move this to line 23?

aphillips (changes requested):
Very close to ready. Comments are somewhat nitpicky.

spec/formatting.md
```diff
@@ -260,9 +260,22 @@ the following steps are taken:
...
+   The **_<dfn>function context</dfn>_** contains the context necessary for
+   the _function handler_ to resolve the _expression_. This includes:
+
+   - The current _locale_,
+     potentially including a fallback chain of locales.
+   - The base directionality of the _message_ and its _text_ tokens.
+
+   If the resolved mapping of _options_ includes any _`u:` options_
+   supported by the implementation, process them as specified.
+   Such `u:` options MAY be removed from the resolved mapping of _options_.
```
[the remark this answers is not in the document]
eemeli: Yes, with the qualifier that this does not mandate that all future `u:` options will behave like the current ones, but it allows for them to do so.

spec/u-namespace.md
eemeli: > for use in assisting the translation process

Isn't this handled by attributes?

spec/u-namespace.md
eemeli, line 29:
```suggestion
Replaces the _locale_ defined in the _function context_ for this _expression_.

A comma-delimited list consisting of
well-formed [BCP 47](https://www.rfc-editor.org/rfc/bcp/bcp47.txt)
language tags,
or an implementation-defined list of such tags.
```

spec/u-namespace.md
eemeli: Leaving that out:
```suggestion
The `u:` _namespace_ is reserved for the definition of _options_
which affect the _function context_ of the specific _expressions_
in which they appear.

The CLDR Technical Committee of the Unicode Consortium
manages the specification for this namespace, hence the name `u:`.
```

This doesn't cover `u:id` on markup, though. Drawing a bit of a blank on exactly how to include that.

spec/formatting.md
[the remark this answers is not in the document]
eemeli: I'm not sure that it's necessary to make a commitment. We should (as you propose below) define the namespace a bit better, which does some of the work, but we should not make it impossible for us to extend it later.

spec/u-namespace.md
eemeli: I fear that defining this too strictly mandate more than we'd like about how an implementation handles locales. Thinking about this some more, my preference would be to allow for each implementation to follow its own customs on locale fallback chain handling.

For instance JavaScript errors on not well-formed tags, but ignores invalid and unsupported ones:
```js
Intl.NumberFormat.supportedLocalesOf(['en-US', 'tlh-Xsux-AQ', 'i-klingon', 'bad-Juju'])
// Uncaught RangeError: invalid language tag: "i-klingon"

Intl.NumberFormat.supportedLocalesOf(['en-US', 'tok', 'bad-Juju', 'American'])
// [ "en-US" ]
```
but it's not reasonable to expect all other implementations to do the same thing.

Furthermore, referring to validity should probably be avoided as well, because in this context that would mean needing to carry around data that might not have any other use.

How about this?

```suggestion
Implementations MAY emit a _Bad Option_ error
and MAY ignore the value of the `u:locale` option
if it contains a locale that it does not support or
a language tag that is not well-formed.
```

aphillips: Mt examples were meant as rhetorical. Your suggestion is fine, except that I think we should allow for the list to be processed as I suggested. I intended that as MAY... basically the implementation can process the list as it likes. But if the list has one bad tag in it, should we ignore the whole thing? That's how it reads now.

eemeli: The current language is intended to allow for an implementation to ignore the whole value and emit an error if it contains a bad tag, or to not do so, or to decide on a case-by-case basis. Note that we also do not specify how and when locale negotiation is meant to happen; that could be handled by the implementation, but it could also be done separately in each function handler.

aphillips: (Back at a real computer)

Perhaps:

```suggestion
Implementations MAY emit a _Bad Option_ error
and MAY ignore the value of the `u:locale` _option_ as a whole
or any of the entries in the list of language tags.
This might be because the locale specified is not supported
or because the language tag is not well-formed,
not valid, or some other reason.
```

eemeli: Eh, seems a bit verbose but works.

aphillips (approved):

catamorphism (approved):

catamorphism:
Should this PR also include a change to [registry.md](https://github.com/unicode-org/message-format-wg/blob/main/spec/registry.md) explaining how the `:string` function _uses_ the value of `u:dir`? As it is, the tests are the only thing documenting how `u:dir` gets used, as far as I can see.

spec/u-namespace.md
```diff
@@ -0,0 +1,86 @@
+# MessageFormat 2.0 Unicode Namespace
+
+The `u:` _namespace_ is reserved for the definition of _options_
+which affect the _function context_ of the specific _expressions_
+in which they appear,
+or for the definition of _options_ that are universally applicable
+rather than function-specific.
```
macchiati, line 7: We also reserve `u:` functions (even if there are none currently).

macchiati (changes requested):

spec/u-namespace.md
eemeli: @aphillips Do you think we ought to call that out specifically?

aphillips: Yeah, I think we should mention it. Perhaps add a sentence to the end like:

> It might also be used to define _functions_ in a future release.

aphillips:
@catamorphism asked:
> Should this PR also include a change to [registry.md](https://github.com/unicode-org/message-format-wg/blob/main/spec/registry.md) explaining how the :string function uses the value of u:dir? As it is, the tests are the only thing documenting how u:dir gets used, as far as I can see.

We need to update and rename `registry.md` soon.

`u:dir` is used by any expression as part of the [default bidi isolation strategy](https://github.com/unicode-org/message-format-wg/blob/main/spec/formatting.md#handling-bidirectional-text).

spec/u-namespace.md
eemeli:
```suggestion
rather than function-specific.
It might also be used to define _functions_ in a future release.
```

aphillips:
In the 2024-10-21 call we merged this so that it is visible in the spec. Open issues against the results if iteration is needed.

PR #850, opened 2024-07-30
Add table of errors to test README (#706)

catamorphism:
Update README for tests directory to include a table relating error names in the test schema to error names in the spec.

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -26,22 +26,22 @@ A [JSON schema](./schemas/) is included for the test files in this repository.
 The following table relates the error names used in the [JSON schema](./schemas/)
 to the error names used in ["MessageFormat 2.0 Errors"](../spec/errors.md) in the spec.
 
-| Schema                      | Spec                        |
+| Spec                        | Schema                      |
 | --------------------------- | --------------------------- |
-| bad-operand                 | Bad Operand                 |
-| bad-option                  | Bad Option                  |
-| bad-selector                | Bad Selector                |
-| bad-variant-key             | Bad Variant Key             |
-| duplicate-declaration       | Duplicate Declaration       |
-| duplicate-option-name       | Duplicate Option Name       |
-| missing-fallback-variant    | Missing Fallback Variant    |
-| missing-selector-annotation | Missing Selector Annotation |
-| syntax-error                | Syntax Error                |
-| unknown-function            | Unknown Function            |
-| unresolved-variable         | Unresolved Variable         |
-| unsupported-expression      | Unsupported Expression      |
-| unsupported-statement       | Unsupported Statement       |
-| variant-key-mismatch        | Variant Key Mismatch        |
+| Bad Operand                 | bad-operand                 |
+| Bad Option                  | bad-option                  |
+| Bad Selector                | bad-selector                |
+| Bad Variant Key             | bad-variant-key             |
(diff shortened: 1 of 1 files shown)


test/README.md
```diff
@@ -24,6 +24,33 @@ A [JSON schema](./schemas/) is included for the test files in this repository.
 
 For users of Visual Studio Code, a [settings file](./.vscode/settings.json) is included that enables schema validation while editing the test files.
```
eemeli, line 26: This should've been dropped in #827.

```suggestion
```

eemeli (approved):

test/README.md
```diff
@@ -21,8 +21,32 @@ Some examples of test harnesses using these tests, from the source repository:
...
+| unknown-function            | Unknown Function            |
+| unresolved-variable         | Unresolved Variable         |
+| unsupported-expression      | Unsupported Expression      |
+| unsupported-statement       | Unsupported Statement       |
+| variant-key-mismatch        | Variant Key Mismatch        |
+
+The "Message Function Error" error name used in the spec
+is not included in the schema,
+as it is intended to be an umbrella category
+for implementation-specific errors.
```
aphillips, line 49: I can imagine writing tests for my own formatting functions that might use such an error. Would we be better off to define the schema name and then say:

```suggestion
> [!NOTE]
> The error name `message-function-error` does not appear in any of the
> tests found in this test suite because the "Message Function Error"
> category of errors is intended as an umbrella category 
> for implementation-specific errors.
```

test/README.md
```diff
@@ -21,8 +21,32 @@ Some examples of test harnesses using these tests, from the source repository:
 - [Formatting tests](https://github.com/messageformat/messageformat/blob/11c95dab2b25db8454e49ff4daadb817e1d5b770/packages/mf2-messageformat/src/messageformat.test.ts)
 
 A [JSON schema](./schemas/) is included for the test files in this repository.
-
-For users of Visual Studio Code, a [settings file](./.vscode/settings.json) is included that enables schema validation while editing the test files.
+## Error Codes
+
+The following table relates the error names used in the [JSON schema](./schemas/)
+to the error names used in ["MessageFormat 2.0 Errors"](../spec/errors.md) in the spec.
```
aphillips, line 27: Consider reversing the table? You're thinking about the errors from the point of view of the schema, but users of the spec will come here to find the ID

aphillips (commented):
Minor issues. Thanks for making this PR.

test/README.md
eemeli: The current schema is designed to specifically support our own tests, and it's not really extensible e.g. due to incorporating `"additionalProperties": false` requirements in many places. This means that adding `message-function-error` is beyond its current intended use cases.

As JSON Schema in general is not extensible (it has no concept of inheritance, for instance), other test suites that would have such a need probably need to copy our schema and patch it locally, or not rely on an explicit JSON Schema document.

I found the PR's proposed language appropriate.

test/README.md
catamorphism: Done in [c609a2f](https://github.com/unicode-org/message-format-wg/pull/850/commits/c609a2fc88c0c2bc2f5be415675ca06a3d84ec68)

aphillips (approved):

mradbourne (approved):

PR #853, opened 2024-08-05
Add Duplicate Variant error

eemeli:
See #847 for context and discussion.

This does not address the questions about identifier normalization also raised in the parent issue; that should be done in a separate PR.

### DIFF
--- a/spec/errors.md
+++ b/spec/errors.md
@@ -178,6 +178,27 @@ appears on the left-hand side of more than one _option_ in the same _expression_
 > {{This is {$foo}}}
 > ```
 
+### Duplicate Variant
+
+A **_<dfn>Duplicate Variant</dfn>_** error occurs when the same _keys_
+are used for more than one _variant_.
+
+> Examples of invalid messages resulting in a _Duplicate Variant_ error:
+>
+> ```
+> .match {$var :string}
+> * {{The first default}}
+> * {{The second default}}
+> ```
+>
+> ```
+> .match {$x :string} {$y :string}
+> *   foo   {{The first "foo" variant}}
+> bar *     {{The "bar" variant}}
+> *   |foo| {{The second "foo" variant}}
+> *   *     {{The default variant}}
+> ```
+
 ## Resolution Errors
 
 **_<dfn>Resolution Errors</dfn>_** occur when the runtime value of a part of a message
--- a/spec/errors.md
+++ b/spec/errors.md
@@ -180,8 +180,8 @@ appears on the left-hand side of more than one _option_ in the same _expression_
 
 ### Duplicate Variant
 
-A **_<dfn>Duplicate Variant</dfn>_** error occurs when the same _keys_
-are used for more than one _variant_.
+A **_<dfn>Duplicate Variant</dfn>_** error occurs when the
+same list of _keys_ is used for more than one _variant_.
 
 > Examples of invalid messages resulting in a _Duplicate Variant_ error:
 >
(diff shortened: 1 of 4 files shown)


spec/errors.md
```diff
@@ -178,6 +178,27 @@ appears on the left-hand side of more than one _option_ in the same _expression_
 > {{This is {$foo}}}
 > ```
 
+### Duplicate Variant
+
+A **_<dfn>Duplicate Variant</dfn>_** error occurs when the same _keys_
+are used for more than one _variant_.
```
aphillips, line 184: Nit

```suggestion
A **_<dfn>Duplicate Variant</dfn>_** error occurs when the
same list of _keys_ is used for more than one _variant_.
```

spec/errors.md
```diff
@@ -178,6 +178,27 @@ appears on the left-hand side of more than one _option_ in the same _expression_
...
+> .match {$var :string}
+> * {{The first default}}
+> * {{The second default}}
+> ```
+>
+> ```
+> .match {$x :string} {$y :string}
+> *   foo   {{The first "foo" variant}}
+> bar *     {{The "bar" variant}}
+> *   |foo| {{The second "foo" variant}}
```
aphillips, line 198: Put the foos next to each other?

spec/syntax.md
```diff
@@ -369,6 +369,8 @@ satisfied:
 - At least one _variant_ MUST exist whose _keys_ are all equal to the "catch-all" key `*`.
 - Each _selector_ MUST have an _annotation_,
   or contain a _variable_ that directly or indirectly references a _declaration_ with an _annotation_.
+- Every _variant_ MUST NOT use the same set of _keys_ in the same order as another _variant_.
```
aphillips, line 372: I suggested "list" above and would suggest it here as well. Sets, as you know, can be unordered, while lists generally are ordered.

```suggestion
Each _variant_ MUST use a list of _keys_ that is unique from that
of all other _variants_ in the _message_.
```

aphillips (commented):
Good start.

spec/errors.md
eemeli: I initially separated them to show that they don't need to be next to each other to be an error, but my sentiments here are not strong.

spec/syntax.md
eemeli: That makes sense; there's another "set" mention that ought to be enlisted as well. Fixed your indents:
```suggestion
- Each _variant_ MUST use a list of _keys_ that is unique from that
  of all other _variants_ in the _message_.
```

aphillips (approved):

catamorphism (approved):
Seems reasonable enough

spec/syntax.md
```diff
@@ -369,6 +369,9 @@ satisfied:
 - At least one _variant_ MUST exist whose _keys_ are all equal to the "catch-all" key `*`.
 - Each _selector_ MUST have an _annotation_,
   or contain a _variable_ that directly or indirectly references a _declaration_ with an _annotation_.
+- Each _variant_ MUST use a list of _keys_ that is unique from that
+  of all other _variants_ in the _message_.
+  _Literal_ _keys_ are compared by their contents, not their syntactical appearance.
```
macchiati, line 374:
```suggestion
  _Literal_ _keys_ are compared by their contents, not their syntactical appearance.
 Keys are considered the same if they are canonically equivalent.
 See [Canonical and Compatibility Equivalence](https://unicode.org/reports/tr15/#Canon_Compat_Equivalence)
```
Unicode conformance allows (and encourages) implementations to normalize canonically equivalent characters. This is a very common operation when text is input and/or processed. Without this step, there are cases where one implementation would return this error and another would not.

macchiati:
Makes sense

PR #854, opened 2024-08-06
Allow whitespace at complex-message start

### LINKED ISSUE #610: Complex message start syntax is too strict
While working on a presentation about the JS implementation, I encountered the following error when testing some of the code I'd written for the slides:

```js
const msg = `
  .local $currency = {EUR}
  .input {$price :number style=currency currency=$currency}
  {{Total price: {$price}}}`;
const mf = new MessageFormat(msg, 'en');
// Syntax parse error: parse-error at [the first { of '{{Total...']
```

After spending quite a while debugging the implementation code to determine the error here, I came to the conclusion that my parser was correct, the above is indeed a syntax error.

The problem with this message is that the leading whitespace before the first `.` means that it isn't special, and all of the declaration syntax parses as a valid simple pattern, up until the `{{` pattern start.

I [mentioned](https://unicode-org.slack.com/archives/C015V19EZEX/p1705586720310499) this problem on our Slack channel, where @stasm was also unable to initially see the problem. I think if _we_ can't see it, other developers won't either, and that this will become a significant stumbling block.

We should consider amending the ABNF to allow whitespace before the `.` at the beginning of a complex message.

### LINKED ISSUE #809: Allow leading and trailing whitespace in complex messages
Currenly, complex messages can not start or end with whitespace. The first character of a complex message must either be `.` or `{`, and the final character must be `}`. This is unfortunate, because this results in unnecessary complexity for users.

Example:

```js
const message = `
.input {$count :number}
.match {$count}
one {{You have {$count} notification.}}
*   {{You have {$count} notifications.}}
`
```

This does not work right now, because there is whitespace between the string start, and the first char of the complex message, and the trailing `}` and the end of the string.

This can not be remedied by just trimming all messages, because for simple messages, leading and trailing whitespace is significant.

For complex messages, leading and trailing whitespace is not significant, because all chars outside of a `quoted-pattern` are control characters and not relevant for display (they are not included in the output message).

To fix this, `complex-message` grammar can be changed as follows:

```diff
- complex-body = quoted-pattern / matcher
+ complex-body = [s] (quoted-pattern / matcher) [s]

```

### LINKED ISSUE #848: [^] ACTION-REQUIRED -- ballot of optional whitespace for complex messages
Per the discussion in the 2024-07-29 teleconference (along with preceding calls on the same topic), there has been a call for voting on the resolution of the issue raised in #809 as part of the LDML46 version of the specification.

# WORKING GROUP BALLOT

**_Please read the instructions CAREFULLY before responding._**

Please carefully read the design document before responding.

## Balloting Instructions

This is a simple vote. Please choose between options (A) or (B).

The **_deadline_** is **1700 (5 PM) in the `America/Los_Angeles` time zone on Sunday, 4 August 2024** Votes received after the deadline will be considered at the discretion of the chair.

- Anyone MAY submit exactly one vote.
  - Working group participants votes will be counted as a full vote.
  - Working group observers votes are not counted, but can be considered by the working group.
- You MAY edit, change, or delete your vote up until the deadline.
- Votes MUST be submitted as a comment on this github issue.
  Group members who cannot submit a comment on this issue should contact the chair (@aphillips) for assistance.
- Votes MUST contain only a vote for (A) or (B). Write in votes are not acceptable.
- You MUST NOT comment on the votes of others in this issue. "Electioneering" or non-voting commentary is not permitted in the issue **_except_** the chair may seek clarification of a vote. Use #849 to discuss.

## Candidates

There are two options. 

**(A)** _Complex messages MUST NOT start with whitespace_

This is the current WG consensus. Complex messages must start with a declaration (`.` followed by a keyword) or with pattern quotes (`{{`). Anything else is a simple pattern.

**(B)** _Complex messages MAY start with whitespace; the first non-whitespace character of a `simple-message` MUST NOT be `.`_

This would permit optional whitespace before the start of a complex message. Whitespace at the start of a simple message remains significant (part of the _pattern_). Note that the body of a well-formed complex message is not a valid simple message.

### LINKED ISSUE #849: Discussion issue for #848: balloting of optional whitespace at complex message start
Use this issue to discuss technical issues related to #848.

Note that there is already extensive discussion in:

* [Recent telecons](https://github.com/unicode-org/message-format-wg/blob/main/meetings/2024/notes-2024-07-29.md)
* #507 
* [PEWS design document](https://github.com/unicode-org/message-format-wg/blob/main/exploration/pattern-exterior-whitespace.md)

eemeli:
Closes #610 
Closes #809 
Closes #848 
Closes #849 
CC @lucacasonato

As agreed in the balloting under #848, complex messages should be allowed to start with whitespace.

### DIFF
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -1,11 +1,11 @@
 message           = simple-message / complex-message
 
-simple-message    = [simple-start pattern]
+simple-message    = [s] [simple-start pattern]
 simple-start      = simple-start-char / escaped-char / placeholder
 pattern           = *(text-char / escaped-char / placeholder)
 placeholder       = expression / markup
 
-complex-message   = *(declaration [s]) complex-body [s]
+complex-message   = [s] *(declaration [s]) complex-body [s]
 declaration       = input-declaration / local-declaration / reserved-statement
 complex-body      = quoted-pattern / matcher
 
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -86,7 +86,7 @@ name-char  = name-start / DIGIT / "-" / "."
            / %xB7 / %x300-36F / %x203F-2040
 
 ; Restrictions on characters in various contexts
-simple-start-char = content-char / s / "@" / "|"
+simple-start-char = content-char / "@" / "|"
 text-char         = content-char / s / "." / "@" / "|"
 quoted-char       = content-char / s / "." / "@" / "{" / "}"
 reserved-char     = content-char / "."
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -156,7 +156,7 @@ A **_<dfn>simple message</dfn>_** contains a single _pattern_,
 with restrictions on its first non-whitespace character.
 An empty string is a valid _simple message_.
 
-Whitespace at the start of a _simple message_ is significant,
+Whitespace at the start or end of a _simple message_ is significant,
 and a part of the _text_ of the _message_.
 
 ```abnf
@@ -172,7 +172,8 @@ and consists of:
 1. an optional list of _declarations_, followed by
 2. a _complex body_
 
-Whitespace at the start of a _complex message_ is ignored.
+Whitespace at the start or end of a _complex message_ is not significant,
+and does not affect the processing of the _message_.
 
 ```abnf
 complex-message = [s] *(declaration [s]) complex-body [s]
@@ -306,7 +307,7 @@ MUST be escaped as `\\`, `\{`, and `\}` respectively.
 
 In the ABNF, _text_ is represented by non-empty sequences of
 `simple-start-char`, `text-char`, `escaped-char`, and `s`.
-The first of these is used at the start of a _simple message_,
+The production `simple-start-char` represents the first non-whitespace in a _simple message_
 and matches `text-char` except for not allowing U+002E FULL STOP `.`.
 The ABNF uses `content-char` as a shared base for _text_ and _quoted literal_ characters.
 
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -153,11 +153,14 @@ message = simple-message / complex-message
 ```
 
 A **_<dfn>simple message</dfn>_** contains a single _pattern_,
-with restrictions on its first character.
+with restrictions on its first non-whitespace character.
 An empty string is a valid _simple message_.
 
+Whitespace at the start of a _simple message_ is significant,
+and a part of the _text_ of the _message_.
+
 ```abnf
-simple-message = [simple-start pattern]
+simple-message = [s] [simple-start pattern]
 simple-start   = simple-start-char / escaped-char / placeholder
 ```
 
@@ -169,8 +172,10 @@ and consists of:
 1. an optional list of _declarations_, followed by
 2. a _complex body_
 
+Whitespace at the start of a _complex message_ is ignored.
+
 ```abnf
-complex-message = *(declaration [s]) complex-body [s]
+complex-message = [s] *(declaration [s]) complex-body [s]
 ```
 
 ### Declarations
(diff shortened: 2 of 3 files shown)


spec/syntax.md
```diff
@@ -153,11 +153,14 @@ message = simple-message / complex-message
 ```
 
 A **_<dfn>simple message</dfn>_** contains a single _pattern_,
-with restrictions on its first character.
+with restrictions on its first non-whitespace character.
 An empty string is a valid _simple message_.
 
+Whitespace at the start of a _simple message_ is significant,
```
aphillips, line 159: while we at it...

```suggestion
Whitespace at the start or end of a _simple message_ is significant,
```

spec/syntax.md
```diff
@@ -169,8 +172,10 @@ and consists of:
 1. an optional list of _declarations_, followed by
 2. a _complex body_
 
+Whitespace at the start of a _complex message_ is ignored.
```
aphillips, line 175:
```suggestion
Whitespace at the start or end of a _complex message_ is not significant,
and does not affect the processing of the _message_.
```

spec/syntax.md
```diff
@@ -300,7 +305,7 @@ U+007B LEFT CURLY BRACKET `{`, and U+007D RIGHT CURLY BRACKET `}`
 MUST be escaped as `\\`, `\{`, and `\}` respectively.
 
 In the ABNF, _text_ is represented by non-empty sequences of
-`simple-start-char`, `text-char`, and `escaped-char`.
+`simple-start-char`, `text-char`, `escaped-char`, and `s`.
```
aphillips, line 308: You're adding this because simple messages can start with non-`simple-start-char` whitespace? Would we be better moving the optional whitespace into that construct?

spec/syntax.md
```diff
@@ -300,7 +305,7 @@ U+007B LEFT CURLY BRACKET `{`, and U+007D RIGHT CURLY BRACKET `}`
...
 The first of these is used at the start of a _simple message_,
 and matches `text-char` except for not allowing U+002E FULL STOP `.`.
```
aphillips, line 310: This actually needs revision, no?

```suggestion
The production `simple-start-char` represents the first non-whitespace in a _simple message_
and matches `text-char` except for not allowing U+002E FULL STOP `.`.
```

spec/message.abnf
```diff
@@ -1,11 +1,11 @@
 message           = simple-message / complex-message
 
-simple-message    = [simple-start pattern]
+simple-message    = [s] [simple-start pattern]
```
aphillips, line 3: I don't think this works because:

```abnf
```

This contains `s`, so allowing `simple-start-char` to match the optional spaces we just installed--and allowing access to `text-char` in `pattern`.

I think we might simplify by changing simple-start-char:

```abnf
simple-start-char = [s] (content-char / "@" / "|")
```

aphillips (changes requested):

spec/message.abnf
eemeli: Ah, you're right. Should be fixed now with `s` removed from `simple-start-char`.

spec/syntax.md
eemeli: I opted to instead remove `s` from `simple-start-char`; the overall effect should be the same, but it keeps the `-char` rule as a single character.

aphillips: Oh, it's better than that: `simple-start-char` was not necessarily a single char previously because `s` is `1*`. So I'd say it makes the production single-char as originally intended and makes the start char clear that `.` isn't included.

catamorphism (approved):
Looks consistent with what we voted on.

gibson042 (approved):
> Looks consistent with what we voted on.

Agreed.

lucacasonato (approved):
Looks good to me too

echeran (approved):
LGTM.

I like the extra callout of

> Whitespace at the start or end of a _simple message_ is significant,
and a part of the _text_ of the _message_.

and

> Whitespace at the start or end of a _complex message_ is not significant,
and does not affect the processing of the _message_.

mihnita (approved):

PR #858, opened 2024-08-11
Fix typo in example

### LINKED ISSUE #857: Possible typo in documentation `formatting.md`
The pattern's expression example mentioned on line 381 of `formatting.md` references undefined variable `$val`:
https://github.com/unicode-org/message-format-wg/blob/01f2880bd709653ad850dc746f4a72fe1133c064/spec/formatting.md?plain=1#L381

Possibly was meant to be `$var` as in `.local $var={|val|} {{{$var :func}}}`

aphillips:
Fixes #857 

Thanks to @mvilks for noticing.

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -378,7 +378,7 @@ The _fallback value_ depends on the contents of the _expression_:
 
   > Examples:
   > In a context where `:func` fails to resolve,
-  > the _pattern_'s _expression_ in `.local $var={|val|} {{{$val :func}}}`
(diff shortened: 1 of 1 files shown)


mihnita (approved):

mihnita:
Approved.

But to nitpick a bit, I would go with something like
```
.local $foo={|bar|} {{{$foo :func}}}
```
or

```
.local $variable={|value|} {{{$variable :func}}}
```


As it is, `var` is a keyword in several programming languages, so it might lead people to think there is something special about it here.
And the the fact that the name of the variable and the value are the same might also throw off some people.

echeran (approved):

PR #860, opened 2024-08-20
Add new selection-declaration alternative: Require annotation of selector variables in placeholders

eemeli:
As discussed in this week's call, I'd like to add a new alternative to the design doc. My own preferences as expressed in #824 have not changed, but this is a viable candidate for consideration.

### DIFF
--- a/exploration/selection-declaration.md
+++ b/exploration/selection-declaration.md
@@ -195,17 +195,27 @@ is expanded to also require later uses of a variable that's used as a selector t
 > each _placeholder_ _expression_ using the same _operand_ as a _selector_ MUST have an _annotation_,
 > or contain a _variable_ that directly or indirectly references a _declaration_ with an _annotation_.
 
-Example valid message:
+Example invalid message with this alternative:
 ```
+.match {$n :number minimumFractionDigits=2}
+* {{Data model error: {$n}}}
+```
+
+Valid, recommended form for the above message:
+```
+.input {$n :number minimumFractionDigits=2}
+.match {$n}
+* {{Formats '$n' as a number with fraction digits: {$n}}}
+```
+
+Technically valid but not recommended:
+```
+.match {$n :number minimumFractionDigits=2}
+* {{Formats '$n' as an integer: {$n :integer}}}
+
 .input {$n :integer}
 .match {$n :number minimumFractionDigits=2}
 * {{Formats '$n' as an integer: {$n}}}
-```
-
-Example invalid message:
-```
-.match {$n :integer}
-* {{If $n==1.2 formats {$n} as 1.2 in en-US}}
 ```
 
 **Pros**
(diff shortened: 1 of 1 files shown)


exploration/selection-declaration.md
```diff
@@ -182,6 +182,42 @@ Examples:
...
+> In a _complex message_,
+> each _placeholder_ _expression_ using the same _operand_ as a _selector_ MUST have an _annotation_,
+> or contain a _variable_ that directly or indirectly references a _declaration_ with an _annotation_.
+
+Example valid message:
+```
+.input {$n :integer}
+.match {$n :number minimumFractionDigits=2}
+* {{Formats '$n' as an integer: {$n}}}
+```
```
macchiati, line 203: The above is not only bizarre, but also error prone. I am dead-set against this.

A basic goal, set out a while ago, is to prevent errors like this one directly causes, where the plural category of the match differs from the plural category of the formatted number.

eemeli: I copied this example from the "Do Nothing" alternative, and I agree that it's bad practice, even if it is technically valid. This alternative does not make all bad ideas impossible, but at least it makes this invalid:
```
.match {$n :number minimumFractionDigits=2}
* {{Error due to missing annotation: {$n}}}
```
and that is the far more common error.

bearfriend: I also prefer other other alternatives over this, but I don't understand what's bizarre about the example. It's completely logical.

```javascript
n = Math.trunc(n);
switch (Intl.NumberFormat(locale, { minimumFractionDigits: 2 }).format(n)) {
  default:
  return `Format '$n' as an integer: ${n}`
}
```

No one would expect `n` in that template to be the formatted number. That said, I agree that for whatever reason it will result in some errors. I'm just not sure avoiding the option over documenting and educating is the right answer. If it's deemed the best option otherwise, of course.

macchiati: This is a toy example. A real one would have plural categories and possibly literals; and that's where the selection behavior matters.

bearfriend: Not really a hill worth dying on but I don't see how that changes anything. Perhaps you could expand on "bizarre".

eemeli:
@macchiati I clarified the examples a bit base on your feedback, noting that some of them are technically valid, but not recommended. PTAL?

exploration/selection-declaration.md
```diff
@@ -195,17 +195,27 @@ is expanded to also require later uses of a variable that's used as a selector t
...
-* {{If $n==1.2 formats {$n} as 1.2 in en-US}}
+.input {$n :number minimumFractionDigits=2}
+.match {$n}
+* {{Formats '$n' as a number with fraction digits: {$n}}}
+```
+
+Technically valid but not recommended:
+```
+.match {$n :number minimumFractionDigits=2}
+* {{Formats '$n' as an integer: {$n :integer}}}
```
aphillips, line 214: placeholder annotation is always valid like this, so this example is less helpful in illustrating the point. I'd move it lower, under the second example. We might want to add this example to the use case section as an illustration of something that we're not solving.

aphillips (approved):

PR #862, opened 2024-08-20
Miscellaneous test fixes

eemeli:
A few things noticed while updating my implementation:
- Some of the new syntax tests were missing expected errors (in particular, bad-selector) beyond the first.
- An unsupported-statement error has an incorrect expected-parts value
- We should test that declaration-like content in quoted patterns is fine, and that `*` and `|*|` are distinguished for duplicate-variant errors.

### DIFF
--- a/test/tests/data-model-errors.json
+++ b/test/tests/data-model-errors.json
@@ -180,6 +180,10 @@
           "type": "duplicate-variant"
         }
       ]
+    },
+    {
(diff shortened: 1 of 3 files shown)


aphillips (approved):
These look like good additions to me.

mradbourne (approved):

aphillips:
Merging on the fast-track per call of 2024-09-02. @mradbourne Consider a PR for your suggestion.

PR #863, opened 2024-08-21
Add tests for pattern selection

eemeli:
Currently, the tests we have for pattern selection rely on the behaviour of `:number`, and are somewhat incomplete. We should instead use `:test:select` and `:test:format` for this, and work to ensure more aspects are covered.

A new `fails` option is added to the the test functions, to allow for the changes made in #828 to be testable (i.e. what happens if a custom selector function throws an error during MatchSelectorKeys).

The `.match` tests on `:number` are all moved to the new test suite.


test/tests/functions/number.json
```diff
@@ -209,173 +209,6 @@
         }
       ]
     },
-    {
-      "src": ".match {$foo :number} one {{one}} * {{other}}",
```
aphillips, line 213: This file tests the function `:number` (and its friend `:integer`, which are specified as part of the standard function set*. Won't we need these tests for that purpose? The file is named appropriately for that.

test/tests/pattern-selection.json
```diff
@@ -0,0 +1,150 @@
+{
+  "$schema": "https://raw.githubusercontent.com/unicode-org/message-format-wg/main/test/schemas/v0/tests.schema.json",
+  "scenario": "Pattern selection",
+  "description": "Tests for the pattern selection parts of the specification",
```
aphillips, line 4:
```suggestion
  "description": "Tests for pattern selection",
```

I think this could be less verbose

test/tests/pattern-selection.json
```diff
@@ -0,0 +1,150 @@
...
+  "defaultTestProperties": {
+    "locale": "en-US"
```
aphillips, line 6: Does this need to be `en-US`? Should this be `und`? `zxx`? We shouldn't over-favor US English, especially when the tests themselves are not locale-affected.

test/tests/pattern-selection.json
```diff
@@ -0,0 +1,150 @@
...
+      "exp": "1.0"
+    },
+    {
+      "src": ".match {1 :test:select decimalPlaces=1} 1 {{1}} 1.0 {{1.0}} * {{other}}",
+      "exp": "1.0"
+    },
+    {
+      "src": ".match {1 :test:select decimalPlaces=9} 1.0 {{1.0}} 1 {{1}} * {{other}}",
+      "exp": "other",
+      "expErrors": [{ "type": "bad-option" }, { "type": "bad-selector" }]
```
aphillips, line 68: This one might have better self-documentation. The problem here is that the option _value_ `9` is invalid, a fact that requires knowledge of the `:test:select` function... which is fine. But we could add documentation here such as:

```
* {{bad-option-value}}",
"exp": "bad-option-value",
...
```

test/README.md
```diff
@@ -65,23 +65,34 @@ The function `:test:function` requires a [Number Operand](/spec/registry.md#numb
...
-a _digit size option_ for which only `0` and `1` are valid values.
+The following _options_ are available on `:test:function`:
+- `decimalPlaces`, a _digit size option_ for which only `0` and `1` are valid values.
+  - `0`
+  - `1`
+- `fails`
+  - `never` (default)
+  - `select`
+  - `format`
+  - `always`
```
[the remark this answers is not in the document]
eemeli: They're somewhat documented by the behaviour section below. Not explicitly documenting these is also in line with what we've done for all the other functions' option values.

test/tests/functions/number.json
eemeli: The tests removed from here rely on the category selection behaviour of `:number`, which is not strictly speaking required of an implementation, as we only [say](https://github.com/unicode-org/message-format-wg/blob/main/spec/registry.md#rule-selection) that "selection should be based on CLDR plural rule data". If that wiggle room is not intentional, then we should (separately from this PR) firm that up, and possibly introduce more specific tests for `:number` that check a wider range of locales.

These tests also mostly exercise the MF2 spec side of pattern selection, for which each case is now included in the `:test:select` tests.

test/tests/pattern-selection.json
eemeli: This can be anything.

```suggestion
    "locale": "und"
```

test/tests/pattern-selection.json
eemeli:
```suggestion
      "src": ".match {1 :test:select decimalPlaces=9} 1.0 {{1.0}} 1 {{1}} * {{bad-option-value}}",
      "exp": "bad-option-value",
      "expErrors": [{ "type": "bad-option" }, { "type": "bad-selector" }]
```

test/tests/pattern-selection.json
```diff
@@ -0,0 +1,150 @@
...
+    },
+    {
+      "src": ".input {$x :test:select decimalPlaces=1} .local $y = {$x :test:select} .match {$y} 1.0 {{1.0}} 1 {{1}} * {{other}}",
+      "params": [{ "name": "x", "value": 1 }],
+      "exp": "1.0"
+    },
+    {
+      "src": ".input {$x :test:select decimalPlaces=9} .match {$x :test:select decimalPlaces=1} 1.0 {{1.0}} 1 {{1}} * {{other}}",
+      "params": [{ "name": "x", "value": 1 }],
+      "exp": "other",
```
eemeli, line 93:
```suggestion
      "src": ".input {$x :test:select decimalPlaces=9} .match {$x :test:select decimalPlaces=1} 1.0 {{1.0}} 1 {{1}} * {{bad-option-value}}",
      "params": [{ "name": "x", "value": 1 }],
      "exp": "bad-option-value",
```

test/tests/functions/number.json
aphillips: We probably should fix these tests to use CLDR data, if that's what we want. My primary concern is not to lose tests for something that **_is_** part of our collection of specifications, even if it isn't part of MF2's core spec.

PR #870, opened 2024-08-26
Add hybrid option to selection-declaration.md

macchiati:


exploration/selection-declaration.md
```diff
@@ -356,3 +356,66 @@ and a data model error otherwise.
   Removes some self-documentation from the pattern.
 - Requires the pattern to change if the selectors are modified.
 - Limits number of referenceable selectors to 10 (in the current form)
+
+### Hybrid approach: Match may mutate, no duplicates
+
+In this alternative, in a `.match` statement:
+
+1. variables are mutated by options
+2. no variable can occur twice
```
aphillips, line 365: Jargon repair

```
1. variables are mutated by their annotation
2. no variable can occur twice in a selector
```

exploration/selection-declaration.md
```diff
@@ -356,3 +356,66 @@ and a data model error otherwise.
...
+
+This keeps most message more concise, producing the expected results in Example 1.
```
aphillips, line 367: Typo

```suggestion
This keeps most messages more concise, producing the expected results in Example 1.
```

exploration/selection-declaration.md
```diff
@@ -356,3 +356,66 @@ and a data model error otherwise.
...
+- No new syntax is required
+- Preserves immutability before and after the .match statement
+- Avoids the serious problem of mismatch of selector and format of option "Do Nothing"
+- Avoids the extra syntax of option "Allow both local and input declarative selectors with immutability"
+- Avoids the problem of multiple variables in "Allow immutable input declarative selectors"
+- Is much more consise than "Match on variables instead of expressions", since it doesn't require a .local or .input for every variable with options
+- Avoids the readability issues with "Provide a #-like Feature"
+
+**Cons**
+- Requires additional .local in cases where a variable would occur twice, such as .match {$date option=monthOnly}{$date option=full}
```
aphillips, line 419: Fixing the example to have a function

```suggestion
- Requires additional `.local` declarations in cases where a variable would occur twice
  such as `.match {$date :date option=monthOnly} {$date :date option=full}`
```

aphillips (commented):
Minor tweaks

aphillips (approved):

aphillips:
In the 2024-08-26 call we agreed to fast-track this item. I need one additional approval to merge.

exploration/selection-declaration.md
```diff
@@ -356,3 +356,66 @@ and a data model error otherwise.
   Removes some self-documentation from the pattern.
 - Requires the pattern to change if the selectors are modified.
 - Limits number of referenceable selectors to 10 (in the current form)
+
+### Hybrid approach: Match may mutate, no duplicates
```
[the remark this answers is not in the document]
macchiati: I was actually opening it up beyond that, that is

```
.input {$in :number}
.local $count = {$in :modify add=-1}
.match {$count :integer}
one {{You have {$count} whole apple.}}
* {{You have {$count} whole apples.}}
```

would be equivalent to:

```
.input {$in :number}
.local $count = {$in :modify add=-1}
.local $count2 = {$count :integer} // logically inserted
.match {$count2} // logically replaced
one {{You have {$count2} whole apple.}} // logically replaced
* {{You have {$count2} whole apples.}} // logically replaced
```
(And that's how it could be implemented internally, without internal mutation.)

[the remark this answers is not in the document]
aphillips: Using only functions in the standard set, would this be a valid restatement of that example, or an error?

```
.input {$in :number}
.local $count = {$in :number signDisplay=always} // affects the formatting
                                                 // but not selection on purpose
.match {$count :integer} // uses :integer for selection, might be a mutability error?
one {{You have {$count} whole apple}}
*   {{You have {$count} whole apples}}
```

I don't want to bring in the whole resolved value discussion, just the discussion of immutability, but I think we intended this to be an error currently. Mutator functions, such as your `:modify`, are super-interesting, in that they aren't really formatters or selectors.

Playing devil's advocate:

```
.input {$in :number}
.local $count = {$in :modify add=-1}
.local $count2 = {$count :integer}      // typed explicitly by the user
.match {$count2}                        // ... and used to select; can I annotate?
one {{You have {$count2} whole apple.}} // does this get implicitly changed?
* {{You have {$count2} whole apples.}}
```

Also, what about non-consonant selectors:

```
.input {$now :datetime}
.match {$now :before field=hour value=12} // is it morning? No effect on display.
true {{You still have time to do it before lunch.}}
*    {{It's already {$now :time} on {$now :date}}}
```

exploration/selection-declaration.md
```diff
@@ -356,3 +356,66 @@ and a data model error otherwise.
...
+```
+.match {$count :integer}
+one {{You have {$count} whole apple.}}
+* {{You have {$count} whole apples.}}
+```
+is precisely equivalent to:
+
+#### Example 2
+```
+.local $count2 {$count :integer}
```
eemeli, line 380:
```suggestion
.local $count2 = {$count :integer}
```

exploration/selection-declaration.md
```diff
@@ -356,3 +356,66 @@ and a data model error otherwise.
...
+
+#### Example 3
+```
+.match {$count <anything>}{$count <anything>}
+```
+It would need to be rewritten as something along the lines of:
+
+#### Example 4
+```
+.local $count3 {$count}
```
eemeli, line 401:
```suggestion
.local $count3 = {$count}
```

exploration/selection-declaration.md
```diff
@@ -356,3 +356,66 @@ and a data model error otherwise.
...
+**Pros**
+- No new syntax is required
+- Preserves immutability before and after the .match statement
+- Avoids the serious problem of mismatch of selector and format of option "Do Nothing"
+- Avoids the extra syntax of option "Allow both local and input declarative selectors with immutability"
+- Avoids the problem of multiple variables in "Allow immutable input declarative selectors"
+- Is much more consise than "Match on variables instead of expressions", since it doesn't require a .local or .input for every variable with options
+- Avoids the readability issues with "Provide a #-like Feature"
+
+**Cons**
```
eemeli, line 417: The cons from the "Allow _mutable_ input declarative selectors" alternative also apply here.

```suggestion
**Cons**
- Complexity: `.match` means more than one thing
- Complexity: `.match` implicitly creates a new lexical scope
- Violates immutability that we've established everywhere else
```

eemeli (approved):
This is fine to add, provided that the inline suggestions below are addressed.

exploration/selection-declaration.md
aphillips: The immutability violation might depend on @macchiati's response to my comment above. It's not a mutability violation if variables must not be twice annotated.

echeran (approved):
Ditto what @eemeli said -- it looks fine to add, but would like to see the suggested change re: cons addressed.

aphillips:
I am merging this without changing the title of the proposed option, the better to get the ballot out the door. The conversation @macchiati and I are having above remains interesting to any future technical discussion.

exploration/selection-declaration.md
```diff
@@ -356,3 +356,69 @@ and a data model error otherwise.
...
+
+1. variables are mutated by their annotation
+2. no variable can be the operand in two selectors
+
+This keeps most messages more concise, producing the expected results in Example 1.
+
+#### Example 1
+
+```
+.match {$count :integer}
```
mihnita, line 372: I know it is already submitted, and we are balloting on it.
But this is risky.

Think of a parameter of type `Person`, with first / last name, DOB, gender, etc.

```
.match {$host :gender}
female {You got an invitation from {$host} to her party.}
...
```

One would really expect something like "... from Maria to her ..."
But because the `.match` "binds" the parameter with `:gender` now we get "... from female to her ..."
Can be solved with a local declaration.
But surprising.

So we fix the plural (`:number`) surprise (select on something and render something else), and we replace it with a different surprise.

aphillips: (as individual)

Presumably `:gender` as a selector is not a formatting function. The default formatting function for `$host` of type "person" would be implementation defined and presumably not be the `:gender` selector, despite the annotation. `:number` or `:integer` are interesting because they are both selectors and formatters. I can imagine other values behaving like that as well (function defaulting is currently implementation defined)

Compound objects, such as person names or measurements, probably are complicated because of the reasonable desire to do query-style selection (e.g. the plural category of a currency value's _number_ or, conversely, select based on the currency value's _currencyCode_).

We haven't fully baked "resolved value" yet and this is part of that discussion.

One last thought: is this really that different from some of the other options being balloted? This consideration would go in the pro/con buckets for each.
---

(chair hat ON)

Note that we are balloting selection's relationship to declarations, but the voting is only advisory to a technical conversation. Also, you are encouraged to vote against solutions you don't like and to discuss them on the discussion thread.

PR #875, opened 2024-09-01
Update README.md

srl295:
site-wide updates per https://github.com/unicode-org/.github/issues/17

PR #876, opened 2024-09-01
Update contributing and joining info

aphillips:
This PR is intended to work in tandem with @srl295's PR #875 

- change the contributing page to point to Unicode's contributing page
- change the contributing page to point to the README.md file for joining instructions
- change the README.md to include joining instructions consistent with Unicode's actual policy. Previously we just asked people to join our mailing list and watch the repo. That's good for _following_ the work.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,9 +1,25 @@
 # Contributing to this project
 
-To join this Working Group, please read the information in the [README.md](./README.md) as well as the Contributor License Agreement information linked just below:
+To join this Working Group, please read the information in the [README.md](./README.md) as well as the Contributor License Agreement information just below:
+
+<!-- boilerplate follows - do not edit -->
 
 ## Contributor License Agreement
 
-In order to contribute to this project, the Unicode Consortium MUST have on file a Contributor License Agreement (CLA) covering your contributions, either an individual or a corporate CLA. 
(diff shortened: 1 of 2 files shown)


aphillips:
I definitely need to do something about delegates.md too, but am saving that for when I have time to do the work.

CONTRIBUTING.md
```diff
@@ -1,32 +1,9 @@
...
-## Joining the Working Group
-
-We are looking for participation from software developers, localization engineers and others with experience
-in Internationalization (I18N) and Localization (L10N). If you wish to contribute to this work, please review
-the information on the Contributor License Agreement below. In addition, you should:
-
-1. Apply to join our [mailing list](https://groups.google.com/a/chromium.org/forum/#!forum/message-format-wg)
-2. Watch this repository (use the "Watch" button in the upper right corner)
-
-<!-- boilerplate follows - do not edit -->
```
srl295, line 12: @aphillips note the 'do not edit' ^

srl295 (changes requested):
Actually, you shouldn't point to the other CONTRIBUTING.md - copy and paste is the right thing here, because we need this text to be in front of users right away.  This file is special and GitHub presents it to the users as part of the PR process.

I guess alternately you could *delete* the CONTRIBUTING.md file and have the readme point to the .github one <https://github.com/unicode-org/.github/blob/main/.github/CONTRIBUTING.md> - that works also.  Then GitHub presents the 'community health file' from the special .github repo in place of the local CONTRIBUTING.md file.

CONTRIBUTING.md
aphillips: I did. I removed it because I removed the boilerplate there.

annebright (approved):

aphillips:
@srl295 I went with a version of (A), tweaking the joining info in CONTRIBUTING.md to point to the revised text in the readme.

srl295 (approved):

PR #877, opened 2024-09-03
Match on variables instead of expressions

### LINKED ISSUE #872: Discussion thread for ballot: "effect of selectors on subsequent placeholders"
This is the discussion thread for the current ballot.

### LINKED ISSUE #873: [^] BALLOT: effect of selectors on subsequent placeholders
Per the discussion in the 2024-08-26 teleconference (along with preceding calls on the same topic), there has been a call for voting on the preferred design of those described in the Design Document [selection-declaration.md](https://github.com/unicode-org/message-format-wg/blob/main/exploration/selection-declaration.md). This is an advisory vote to be used in technical discussion.

> [!NOTE]
> The design document lists a "proposed design". Please ignore this.

> [!NOTE]
> The design document was merged from conversation on #824 and there may be valuable discussion on that thread.

# WORKING GROUP BALLOT

**_Please read the instructions CAREFULLY before responding._**

Please carefully read the design document before responding.

## Balloting Instructions

The voting methodology is single transferable vote.

The **_deadline_** is **1700 (5 PM) in the `America/Los_Angeles` time zone on Sunday, 1 September 2024** Votes received after the deadline will be considered at the discretion of the chair.

- Anyone MAY submit exactly one ballot.
  - Working group participants ballots will be counted as a full vote.
  - Working group observers ballots are not officially counted, but can be considered by the working group.
- You MAY edit, change, or delete your ballot up until the deadline.
- Ballots MUST be submitted as a comment on this github issue.
  Group members who cannot submit a comment on this issue should contact the chair (@aphillips) for assistance.
- Ballots MUST contain only a votes for candidates listed below. Write in votes are not acceptable.
- You MUST NOT comment on the ballots of others in this issue. "Electioneering" or non-voting commentary is not permitted in the issue **_except_** the chair may seek clarification of a vote. 
   - **Use #872 to discuss.**

Ballots should be cast as an ordered list of preferred candidates using the format specified below.
Do not include candidates into your ranked choices that you find unacceptable. 
You may indicate candidates you find unacceptable using `!` as showing in the example below.
Do not equate two items. The chair will treat `X = Y` as `X > Y` in all cases.

Example 1: `X > Y > Z ! Q = R` means:
- I prefer `X` to `Y` and `Y` to `Z`.
- Candidates `Q` and `R` are not acceptable to me.
- Candidates `S` and `T` are not listed. They receive no votes but are not unacceptable.

> [!IMPORTANT]
> Listing items as unacceptable has no effect on tallying. They are merely advisory to others in the WG.

## Candidates

**(A) Do nothing**

**(B) Require annotation of selector variables in placeholders**

**(C) Allow both local and input declarative selectors with immutability**

**(D) Allow _mutable_ input declarative selectors**

**(E) Allow _immutable_ input declarative selectors**

**(F) Match on variables instead of expressions**

**(G) Provide a #-like Feature**

**(H) Hybrid approach: Match may mutate, no duplicates**

eemeli:
Closes #872 
Closes #873 
~A rebase might be needed to deconflict test changes with #862.~ Edit: Done.

Submitting initially as a draft, pending final approval of the preliminary consensus noted in https://github.com/unicode-org/message-format-wg/issues/873#issuecomment-2325157889. The changes made here should be complete, though, and could be reviewed already. This PR also marks the design doc as accepted.

A consequence of changing the `selector` rule to be a `variable` rather than an `expression` is that the spaces around it need to now be required, rather than optional.

Another (mostly academic) interesting change is that a message starting with `.match`, such as
```
.match $x * {{...}}
```
is now _well-formed_, but not _valid_, as the missing annotation is considered a data model error. Complicating the syntax to make one or more declarations syntactically required for a message with variants seems unnecessary.


aphillips:
@eemeli Thanks for starting on this.

> Complicating the syntax to make one or more declarations syntactically required for a message with variants seems unnecessary.

While I tend to agree with you, the change isn't _that_ complicated. I think it might be:

```diff
- complex-message   = [s] *(declaration [s]) complex-body [s]
+ complex-message   = [s] (quoted-pattern / (1*(declaration [s]) complex-body)) [s]
```

TBH, I'd rather make it a data model error.

spec/errors.md
```diff
@@ -46,7 +46,7 @@ or contains some error which leads to further errors,
 an implementation which does not emit all of the errors
 SHOULD prioritise _Syntax Errors_ and _Data Model Errors_ over others.
 
-When an error occurs within a _selector_,
+When an error occurs while resolving a _selector_,
```
aphillips, line 49: Perhaps not "while resolving"? My thinking here is that selector resolution means determining what the selector consists of. The original text covered internal errors of the function itself. Perhaps:

```suggestion
When an error occurs during resolution of a _selector_ or within the _selector_,
```

spec/message.abnf
```diff
@@ -14,9 +14,9 @@ local-declaration = local s variable [s] "=" [s] expression
 
 quoted-pattern    = "{{" pattern "}}"
 
-matcher           = match-statement 1*([s] variant)
-match-statement   = match 1*([s] selector)
-selector          = expression
+matcher           = match-statement s variant *([s] variant)
+match-statement   = match 1*(s selector)
+selector          = variable
```
aphillips, line 19: Perhaps less complicated as:

```suggestion
matcher           = match-statement s 1*(variant [s])
match-statement   = match 1*(s selector)
selector          = variable
```

spec/registry.md
```diff
@@ -51,9 +51,9 @@ The function `:string` has no options.
 #### Selection
 
 When implementing [`MatchSelectorKeys(resolvedSelector, keys)`](/spec/formatting.md#resolve-preferences)
-where `resolvedSelector` is the resolved value of a _selector_ _expression_
+where `resolvedSelector` is the resolved value of a _selector_
 and `keys` is a list of strings,
-the `:string` selector performs as described below.
+the `:string` function performs as described below.
```
aphillips, line 56: I think this loses a bit of fidelity? Perhaps:

```suggestion
the `:string` selector function performs as described below.
```

spec/syntax.md
```diff
@@ -373,41 +374,41 @@ satisfied:
...
+  directly or indirectly references a _declaration_ with an _annotation_.
 - Each _variant_ MUST use a list of _keys_ that is unique from that
   of all other _variants_ in the _message_.
   _Literal_ _keys_ are compared by their contents, not their syntactical appearance.
 
 ```abnf
-matcher         = match-statement 1*([s] variant)
-match-statement = match 1*([s] selector)
+matcher         = match-statement s variant *([s] variant)
+match-statement = match 1*(s selector)
```
aphillips, line 385: Ibid if my proposed change is accepted

spec/syntax.md
```diff
@@ -373,41 +374,41 @@ satisfied:
...
 >
 > ```
-> .match {:platform} windows {{Settings}} * {{Preferences}}
+> .local $os = {:platform} .match $os windows {{Settings}} * {{Preferences}}
 > ```
 
 ### Selector
 
-A **_<dfn>selector</dfn>_** is an _expression_ that ranks or excludes the
+A **_<dfn>selector</dfn>_** is a _variable_ that ranks or excludes the
```
aphillips, line 405: This is doesn't make sense?

```suggestion
A **_<dfn>selector</dfn>_** is a _variable_ whose resolved value contains an _expression_
that ranks or excludes the
```

Or perhaps instead of "contains" say "is a _variable_ annotated with an _expression_"?

aphillips (commented):
First pass.

I'm still struggling with this vs. (E) [have not yet "disagree and committed" to the emerging consensus]. I guess I'm not thrilled to have to write declarations in order to use a selector, which is why I prefer option E in the voting.

There is a certain consistency to having to declare everything you use syntactically, but that consistency breaks down later because we allow annotations in pattern expressions without declaration. I would be super-unhappy to remove that capability, because historically this has provided a kind of pattern-local documentation for translators, e.g.

```
You have {$n} <- $n could be anything
```

vs.

```
You have {$n :number currency=USD} <- now I know how to translate this
```

spec/errors.md
aphillips: No, because parsing and resolution are not what the original problem was. The selector is a function call. If the selector blows up (signals an error, for example by throwing an exception), that has nothing to do with parsing or resolution. This text is meant to deal with that situation. Parsing errors are syntax errors (or data model errors) and happen long before calling the selector. Resolution errors are a potential problem, though.

bearfriend: Got it. Thanks. We're talking about [MatchSelectorKeys](https://github.com/unicode-org/message-format-wg/blob/ae120c0b22d8888bfb4df5b232bbe605f73d76f0/spec/formatting.md#resolve-preferences) here (or something very nearby it)?

eemeli: This needs to cover errors in both [Resolve Selectors](https://github.com/unicode-org/message-format-wg/blob/main/spec/formatting.md#resolve-selectors) (step 2.iii.) and in [Resolve Preferences](https://github.com/unicode-org/message-format-wg/blob/main/spec/formatting.md#resolve-preferences) (step 2.iv., calling calling MatchSelectorKeys). Hence also the two sorts of errors that are mentioned below, though that needs an update due to #828 -- I'll include that here.

I don't think "within" works anymore, as _selector_ is now a _variable_. Maybe something like this?

```suggestion
When an error occurs while resolving a _selector_
or calling MatchSelectorKeys with its resolved value,
```

spec/errors.md
```diff
@@ -46,7 +46,7 @@ or contains some error which leads to further errors,
...
 the _selector_ MUST NOT match any _variant_ _key_ other than the catch-all `*`
 and a _Resolution Error_ or a _Message Function Error_ MUST be emitted.
```
eemeli, line 51:
```suggestion
and a _Bad Selector_ error MUST be emitted.
```

spec/message.abnf
eemeli: While that works, it would mean that it's not clear if trailing spaces at the end of a complex message would belong to the `matcher` or the `complex-message`.

spec/syntax.md
eemeli: This could indeed be imprvoed, but I think including _expression_ here would be misleading, as the selection is ultimately done by the value coming from the function handling the annotation.
```suggestion
A **_<dfn>selector</dfn>_** is a _variable_ whose resolved value ranks or excludes the
```

spec/message.abnf
aphillips: The spaces are superfluous to both matcher and complex-message, so perhaps okay? But what you have works.

spec/syntax.md
aphillips: > as the selection is ultimately done by the value coming from the function handling the annotation

We sometimes talk as if this is the case, but it's not true. The function doesn't emit a value (unless that value is sorted-and-filtered list of keys from `MatchSelectorsKeys(r,k)`). The variable here is the operand being used by the selector _plus_ the annotation configuring the selector. A function name like `:number` has two separate interfaces: one for formatting and one for selection.

I agree that my formulation is kind of klunky, but one of the ramifications of Option F is that there MUST be an annotating expression in a declaration.

Maybe:

```suggestion
A **_<dfn>selector</dfn>_** is a _variable_ that appears in a
previous _declaration_ and whose resolved value ranks or excludes the
```

aphillips (approved):
Pending, of course, that we decide, as a working group, to accept this design as the WG consensus.

spec/message.abnf
```diff
@@ -14,9 +14,9 @@ local-declaration = local s variable [s] "=" [s] expression
 
 quoted-pattern    = "{{" pattern "}}"
 
-matcher           = match-statement 1*([s] variant)
-match-statement   = match 1*([s] selector)
-selector          = expression
+matcher           = match-statement s variant *([s] variant)
```
gibson042, line 17: I don't like the inconsistency of whitespace being required before the first match variant and optional before later ones, but it is necessary to support unquoted variant keys as in
```
.input{$count :number}.match $count␣1{{1 new message}}*{{{$count} new messages}}
```

gibson042 (approved):

aphillips:
@eemeli has updated this. Propose we merge this tomorrow (2024-09-10)

PR #879, opened 2024-09-08
Clarify error & fallback handling

eemeli:
This PR is seeking to fulfill the following [Things That Need Doing](https://github.com/unicode-org/message-format-wg/wiki/Things-That-Need-Doing#error-handling):

- Ensure that a complete enumeration and error types is in place (also discussed on [2024-07-22](https://github.com/unicode-org/message-format-wg/blob/main/meetings/2024/notes-2024-07-22.md#topic-result-of-balloting-for-error-handling-830))
- Ensure that error requirements are defined in the specification
- Ensure that error fallback format is defined for each error state

After a somewhat thorough review, the changes needed for the above are mostly but not completely editorial, adding explicit references, moving text around, removing unnecessary MUSTard, and making better use of our _well-formed_ and _valid_ definitions.

The one place where I'm proposing an actual change is in the treatment of attributes, about which we currently problematically say both https://github.com/unicode-org/message-format-wg/blob/aa8d29329a66a7e1399cf955a95c0b582cc1b0d5/spec/syntax.md?plain=1#L753-L754 and https://github.com/unicode-org/message-format-wg/blob/aa8d29329a66a7e1399cf955a95c0b582cc1b0d5/spec/formatting.md?plain=1#L61

To resolve this without needing to add something like a _Duplicate Attribute_ error or adding support for multiple attributes with the same identifier, I propose that we ignore all after the first. This removes a requirement for attributes to be included in the consideration of a message's validity, and allows for the statement in the Formatting section to hold more true.

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -10,8 +10,7 @@ or created from a data model description.
 If the resulting _message_ is not _well-formed_, a _Syntax Error_ is emitted.
 If the resulting _message_ is _well-formed_ but is not _valid_, a _Data Model Error_ is emitted.
 
-The formatting of a _message_ is defined by the following operations,
-starting with _Pattern Selection_:
+The formatting of a _message_ is defined by the following operations:
 
 - **_<dfn>Pattern Selection</dfn>_** determines which of a message's _patterns_ is formatted.
   For a message with no _selectors_, this is simple as there is only one _pattern_.
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -40,7 +40,7 @@ as long as the final _formatting_ result is made available to users
 and the observable behavior of the _formatting_ matches that described here.
 
 _Attributes_ MUST NOT have any effect on the formatted output of a _message_,
-or be made available to function implementations.
+nor be made available to function implementations.
 
 > [!IMPORTANT]
 >
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -7,10 +7,20 @@ when formatting a message for display in a user interface, or for some later pro
 
 To start, we presume that a _message_ has either been parsed from its syntax
 or created from a data model description.
-If this construction has encountered any _Syntax Errors_ or _Data Model Errors_,
-an appropriate error MUST be emitted and a _fallback value_ MAY be used as the formatting result.
+If the resulting _message_ is not _well-formed_, a _Syntax Error_ is emitted.
+If the resulting _message_ is _well-formed_ but is not _valid_, a _Data Model Error_ is emitted.
 
-Formatting of a _message_ is defined by the following operations:
+The formatting of a _message_ is defined by the following operations,
+starting with _Pattern Selection_:
+
+- **_<dfn>Pattern Selection</dfn>_** determines which of a message's _patterns_ is formatted.
+  For a message with no _selectors_, this is simple as there is only one _pattern_.
+  With _selectors_, this will depend on their resolution.
+
+- **_<dfn>Formatting</dfn>_** takes the resolved values of the selected _pattern_,
+  and produces the formatted result for the _message_.
+  Depending on the implementation, this result could be a single concatenated string,
+  an array of objects, an attributed string, or some other locally appropriate data type.
 
 - **_<dfn>Expression and Markup Resolution</dfn>_** determines the value of an _expression_ or _markup_,
   with reference to the current _formatting context_.
@@ -23,6 +33,14 @@ Formatting of a _message_ is defined by the following operations:
 
   The resolution of _text_ is rather straightforward,
   and is detailed under _literal resolution_.
+
+Formatter implementations are not required to expose
+the _expression resolution_ and _pattern selection_ operations to their users,
+or even use them in their internal processing,
+as long as the final _formatting_ result is made available to users
+and the observable behavior of the formatter matches that described here.
+
+_Attributes_ MUST NOT affect the processing or output of a _message_.
 
 > [!IMPORTANT]
 >
@@ -38,27 +56,6 @@ Formatting of a _message_ is defined by the following operations:
 > _declarations_ and _selectors_ affecting _variables_ referenced by that _expression_
 > have already been evaluated in the order in which the relevant _declarations_
 > and _selectors_ appear in the _message_.
-
-- **_<dfn>Pattern Selection</dfn>_** determines which of a message's _patterns_ is formatted.
-  For a message with no _selectors_, this is simple as there is only one _pattern_.
-  With _selectors_, this will depend on their resolution.
-  
-  At the start of _pattern selection_,
-  if the _message_ contains any _reserved statements_,
-  emit an _Unsupported Statement_ error.
-
-- **_<dfn>Formatting</dfn>_** takes the resolved values of the selected _pattern_,
-  and produces the formatted result for the _message_.
-  Depending on the implementation, this result could be a single concatenated string,
-  an array of objects, an attributed string, or some other locally appropriate data type.
-
-Formatter implementations are not required to expose
-the _expression resolution_ and _pattern selection_ operations to their users,
-or even use them in their internal processing,
-as long as the final _formatting_ result is made available to users
-and the observable behavior of the formatter matches that described here.
-
-_Attributes_ MUST NOT affect the processing or output of a _message_.
 
 ## Formatting Context
 
@@ -82,8 +79,7 @@ At a minimum, it includes:
 - The _function registry_,
   providing the implementations of the functions referred to by message _functions_.
 
-- Optionally, a fallback string to use for the message
-  if it contains any _Syntax Errors_ or _Data Model Errors_.
+- Optionally, a fallback string to use for the message if it is not _valid_.
 
 Implementations MAY include additional fields in their _formatting context_.
 
@@ -193,7 +189,7 @@ Otherwise, the _variable_ is an implicit reference to an input value,
 and its value is looked up from the _formatting context_ _input mapping_.
 
 The resolution of a _variable_ MAY fail if no value is identified for its _name_.
-If this happens, an _Unresolved Variable_ error MUST be emitted.
+If this happens, an _Unresolved Variable_ error is emitted.
 If a _variable_ would resolve to a _fallback value_,
 this MUST also be considered a failure.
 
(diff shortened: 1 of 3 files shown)


spec/formatting.md
```diff
@@ -7,10 +7,20 @@ when formatting a message for display in a user interface, or for some later pro
...
 
-Formatting of a _message_ is defined by the following operations:
+The formatting of a _message_ is defined by the following operations,
+starting with _Pattern Selection_:
+
+- **_<dfn>Pattern Selection</dfn>_** determines which of a message's _patterns_ is formatted.
+  For a message with no _selectors_, this is simple as there is only one _pattern_.
+  With _selectors_, this will depend on their resolution.
+
+- **_<dfn>Formatting</dfn>_** takes the resolved values of the selected _pattern_,
```
aphillips, line 20: Are _Formatting_ and _Expression and Markup Resolution_ out of order? I think one would need to do the resolution bits before performing the "formatting". The introductory text on line 13 suggests that this is an ordered list of operations, so maybe flip-flop the two?

spec/formatting.md
```diff
@@ -24,6 +34,14 @@ Formatting of a _message_ is defined by the following operations:
   The resolution of _text_ is rather straightforward,
   and is detailed under _literal resolution_.
 
+Formatter implementations are not required to expose
```
aphillips, line 37: Too specific?

```suggestion
Implementations are not required to expose
```

spec/formatting.md
```diff
@@ -24,6 +34,14 @@ Formatting of a _message_ is defined by the following operations:
...
+the _expression resolution_ and _pattern selection_ operations to their users,
+or even use them in their internal processing,
+as long as the final _formatting_ result is made available to users
+and the observable behavior of the formatter matches that described here.
```
aphillips, line 41: Consistent with the above comment...

```suggestion
and the observable behavior of the formatting matches that described here.
```

macchiati:
I agree with the direction, but it should be the last that wins. That
matches the expectations for maps, and of chained functions.

spec/formatting.md
```diff
@@ -24,6 +34,14 @@ Formatting of a _message_ is defined by the following operations:
...
+
+_Attributes_ MUST NOT affect the processing or output of a _message_.
```
aphillips, line 43: This is too strong, since "processing" might reasonably include that used in translation, which is the whole point of attributes. Perhaps this is closer to what is intended:

```suggestion
_Attributes_ MUST NOT affect the _pattern selection_, _expression resolution_, or _formatting_ of a _message_.
```

spec/formatting.md
```diff
@@ -193,7 +189,7 @@ Otherwise, the _variable_ is an implicit reference to an input value,
 and its value is looked up from the _formatting context_ _input mapping_.
 
 The resolution of a _variable_ MAY fail if no value is identified for its _name_.
```
aphillips, line 191: This is the one that is wrong. The "MAY" here is incorrectly normative.

```suggestion
The resolution of a _variable_ fails if no value is identified for its _name_.
```

spec/formatting.md
```diff
@@ -435,6 +431,18 @@ _Pattern selection_ is not supported for _fallback values_.
 
 ## Pattern Selection
 
+At the start of _pattern selection_,
+if the _message_ contains any _reserved statements_,
```
aphillips, line 435: Simplify?

```suggestion
If the _message_ contains any _reserved statements_,
```

spec/formatting.md
```diff
@@ -435,6 +431,18 @@ _Pattern selection_ is not supported for _fallback values_.
...
+emit an _Unsupported Statement_ error.
+
+If the _message_ being formatted is not _well-formed_ and _valid_,
+the result of pattern selection is a _pattern_ consisting of a single _fallback value_
+using the _message_'s fallback string defined in the _formatting context_
+or if this is not available or empty, the U+FFFD REPLACEMENT CHARACTER `�`.
+
+If the _message_ being formatted does not contain a _matcher_,
+the result of pattern selection is its _pattern_ value.
```
aphillips, line 444: I think there might be three cases here (for the sake of clarity)

```suggestion
If the _message_ is a _simple message_,
the result of _pattern selection_ is the _message_ itself.

If the _message_ is a _complex message_
and the _message_ contains no _matcher_,
the result of _pattern selection_ is the _pattern_ value of the _message_.
```

spec/formatting.md
```diff
@@ -787,8 +790,8 @@ the _fallback value_ as a string,
 and a U+007D RIGHT CURLY BRACKET `}`.
 
 > For example,
-> a message with a _Syntax Error_ and no fallback string
-> defined in the _formatting context_ would format to a string as `{�}`.
+> a _message_ that is not _well-formed_ would format to a string as `{�}`,
+> if no fallback string is defined in the _formatting context_.
```
aphillips, line 794: Nit.

```suggestion
> unless a fallback string is defined in the _formatting context_,
> in which case that string would be used instead.
```

spec/syntax.md
```diff
@@ -368,14 +368,18 @@ and at least one _variant_.
...
-- The number of _keys_ on each _variant_ MUST be equal to the number of _selectors_.
-- At least one _variant_ MUST exist whose _keys_ are all equal to the "catch-all" key `*`.
-- Each _selector_ MUST have an _annotation_,
+- _Variant Key Mismatch_:
+  The number of _keys_ on each _variant_ MUST be equal to the number of _selectors_.
+- _Missing Fallback Variant_:
+  At least one _variant_ MUST exist whose _keys_ are all equal to the "catch-all" key `*`.
+- _Missing Selector Annotation_:
+  Each _selector_ MUST have an _annotation_,
   or contain a _variable_ that directly or indirectly references a _declaration_ with an _annotation_.
```
aphillips, line 380: This text might change on Monday 😸

aphillips (commented):
Some comments. Good start on this.

spec/formatting.md
eemeli: The way I would put it, _Expression and Markup Resolution_ is a dependency for both _Pattern Selection_ and _Formatting_, hence putting them in this order here.

As a separate note, it might be a good idea to move the Pattern Selection examples somewhere else; they're pretty voluminous.

spec/formatting.md
eemeli: Sure; this paragraph was not actually modified, it just gets highlighted that way due to moving the note lower down.

spec/formatting.md
eemeli: We discussed this previously in https://github.com/unicode-org/message-format-wg/pull/845#discussion_r1697081284, but that wasn't conclusive before that PR was merged.

I'd still prefer my suggestion from https://github.com/unicode-org/message-format-wg/pull/845#discussion_r1697159755:
```suggestion
_Attributes_ MUST NOT have any effect on the formatted output of a _message_,
or be made available to function implementations.
```

That ought to have the same effect, while not requiring us to enumerate all the parts of this document.

Would that work?

spec/formatting.md
aphillips: Maybe restore line 13 then? More like

> The formatting of a _message_ includes these operations:

spec/formatting.md
eemeli: I would find it very confusing for pattern selection to result in either a _message_ or a _pattern_, esp. as we also say this: https://github.com/unicode-org/message-format-wg/blob/aa8d29329a66a7e1399cf955a95c0b582cc1b0d5/spec/syntax.md?plain=1#L155-L156

I know that we don't require a data model in the spec, but in places like this I think it's clearer if we don't use language that isn't necessarily related to any data model concepts, such as the difference between a _simple message_ and a _complex message_.

spec/syntax.md
eemeli: I'm aware; will resolve conflicts after whichever lands first.

spec/syntax.md
```diff
@@ -750,10 +755,10 @@ by an U+003D EQUALS SIGN `=` along with optional whitespace.
 
 Multiple _attributes_ are permitted in an _expression_ or _markup_.
 Each _attribute_ is separated by whitespace.
-Each _attribute_'s _identifier_ MUST be unique within the _expression_ or _markup_:
-an _expression_ or _markup_ with duplicate _attribute_ _identifiers_ is not _valid_.
 
-The order of _attributes_ is not significant.
+Each _attribute_'s _identifier_ SHOULD be unique within the _expression_ or _markup_:
+_attributes_ beyond the first with the same _identifier_ are ignored.
```
eemeli, line 760:
```suggestion
all but the last _attribute_ with the same _identifier_ are ignored.
```

spec/formatting.md
```diff
@@ -7,10 +7,20 @@ when formatting a message for display in a user interface, or for some later pro
...
 To start, we presume that a _message_ has either been parsed from its syntax
 or created from a data model description.
-If this construction has encountered any _Syntax Errors_ or _Data Model Errors_,
-an appropriate error MUST be emitted and a _fallback value_ MAY be used as the formatting result.
+If the resulting _message_ is not _well-formed_, a _Syntax Error_ is emitted.
+If the resulting _message_ is _well-formed_ but is not _valid_, a _Data Model Error_ is emitted.
 
-Formatting of a _message_ is defined by the following operations:
+The formatting of a _message_ is defined by the following operations,
+starting with _Pattern Selection_:
```
eemeli, line 14:
```suggestion
The formatting of a _message_ is defined by the following operations:
```

spec/formatting.md
aphillips: I agree, but also want to make life easy on readers of our spec.

Perhaps:

> If the _message_ is a _simple message_, the entire _message_ is the _pattern_.

eemeli: My concerns with your proposal are that
1. As we say in Syntax that the message "contains" a pattern, we should not say here that a message "is" a pattern.
2. By the time we get here, it's almost certain that an implementation is dealing with some data model representation of the pattern, and our own data model does not distinguish between simple and complex messages; both `foo` and `{{foo}}` map to the same construct. Therefore, adding even a reference to _simple message_ is really close to establishing a need for us to add that information to the data model, which we've managed to avoid so far.

aphillips: Fair enough

spec/formatting.md
aphillips: I think so. Do we need to mention markup in addition to function implementation?

eemeli: No, because markup processing can't be customized within MF2, and therefore it's covered by "any effect on the formatted output".

macchiati: > +_Attributes_ MUST NOT have any effect on the formatted output of a _message_,
> +or be made available to function implementations.

This is slightly ambiguous. If you mean:

Attributes_ MUST NOT have any effect on the formatted output of a _message_,
AND
Attributes_ MUST NOT be made available to function implementations.

Then it would be better to say:
> +_Attributes_ MUST NOT have any effect on the formatted output of a _message_,
> +NOR be made available to function implementations.

spec/formatting.md
```diff
@@ -24,6 +33,15 @@ Formatting of a _message_ is defined by the following operations:
...
   and is detailed under _literal resolution_.
 
+Implementations are not required to expose
+the _expression resolution_ and _pattern selection_ operations to their users,
+or even use them in their internal processing,
+as long as the final _formatting_ result is made available to users
+and the observable behavior of the _formatting_ matches that described here.
+
+_Attributes_ MUST NOT have any effect on the formatted output of a _message_,
+or be made available to function implementations.
```
eemeli, line 43: As suggested by @macchiati in https://github.com/unicode-org/message-format-wg/pull/879#discussion_r1749510245, but with lower-case "nor" as it's not an RFC 2119 term:
```suggestion
_Attributes_ MUST NOT have any effect on the formatted output of a _message_,
nor be made available to function implementations.
```

aphillips (approved):

PR #883, opened 2024-09-09
Remove forward-compatibility promise and all reserved & private syntax

### LINKED ISSUE #881: Should we drop private-use annotations?
During today's call, we resolved to drop forwards compatibility from the stability policy, and to remove all reserved syntax.

`private-use-annotation` was originally added to the spec in #404 at least in part due to us adding `reserved-annotation`. As we're now removing the latter, we should reconsider whether to remove the former as well.

Can we imagine use cases for private-use annotations that would not be equally well served by function annotations?

eemeli:
Closes #881 

As discussed today, we should remove the promise that a 2.0 parser will be able to deal with 2.1 syntax, and accordingly remove all reserved syntax (because with that stability policy change, all such reservations are unnecessary).

As noted in #881, this removes a major reason for private-use annotations, which I'm proposing to remove here as well.

As _annotation_ is now exactly the same as _function_, these two definitions are merged.

The Unsupported Statement and Unsupported Annotation errors are removed, and their corresponding tests are now syntax error tests.


aphillips (approved):

spec/README.md
```diff
@@ -90,16 +90,7 @@ Updates to this specification MUST NOT specify the use of a fallback value for a
...
 
 Updates to this specification will not change the syntactical meaning
-of any syntax defined in this specification except for that syntax marked as
-"reserved for future standardization".
-
-Updates to this specification will not assign any meaning to or change the syntactical 
-requirements for any private-use annotation.
-
-Updates to this specification will not remove any reserved keywords or sigils.
-
```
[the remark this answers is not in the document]
eemeli: Isn't that already pretty well covered by "not make any valid _message_ invalid" and "not remove any syntax provided in this version"?

spec/README.md
```diff
@@ -137,23 +121,13 @@ based on this version being invalid.
...
 > in future versions of this specification:
+> - Future versions may define new syntax and structures
+>   that would not be supported by this version of the specification.
 > - Future versions may add additional structure or meaning to existing syntax.
 > - Future versions may define new keywords.
-> - Future versions may define annotations that use portions of the `reserved-annotation`
->   syntax.
 > - Future versions may make previously invalid messages valid.
 > - Future versions may define additional functions in the default registry
 >   or may reserve the names of functions for the purposes of interoperability.
```
macchiati, line 133: Unicode specifications allow deprecation of a particular entity. However, deprecation does **not** mean removal; conformation implementations can (and sometimes should) continue to support that entity, but it is marked as being discouraged, eg resulting in lint warnings.

aphillips: @macchiati Do you have a suggestion here?

I think we say that we might deprecate (but never remove) elsewhere?

macchiati: Perhaps something like the following. It doesn't have to be here.

```suggestion
>   or may reserve the names of functions for the purposes of interoperability.
> - Future versions may deprecate elements (keywords, functions, options, etc.) but may not remove them. (Deprecation means that they are strongly discouraged; for example, linting implementations may issue warnings.)
```

eemeli: A few lines down we already have this:
> Future versions may deprecate functions, options, or option values.

Is that not sufficient?

PR #884, opened 2024-09-11
Add bidi support and address UAX31/UTS55 requirements

### LINKED ISSUE #661: [FEEDBACK] Improving bidi layout
In MF1.0, we allow RLM and LRM in any position between syntax tokens that is not part of the literal message; those characters are then ignorable in **parsing** the message. The advantage of this is that many people will be viewing the message in plaintext editors, and the default results for the bidi algorithm is designed for normal text, and can really mess up "code-like" text (as in MF1.0 or MF2.0).

Here is an example of what happens with BIDI text. This is using a tool (mentioned below) that has the convention that UPPERCASE stands for bidi characters so that people can read them, and I used ⁅...⁆  is used in place of {...} because of the tooling.

Example Source:
\* ⁅⁅YOU HAVE ⁅$count⁆ NOTIFICATION.⁆⁆

Resulting Display:
⁅⁅.NOITACIFITON ⁅count  :number$⁆ EVAH UOY⁆⁆ *

Notice that "YOU ARE" is reversed & NOTIFICATION is reversed (which is the correct order among the pieces), but the $ is jumbled.

but if the line started with 'one' instead of *, you'd get different results.

one ⁅⁅EVAH UOY ⁅$count :number⁆ NOITACIFITON.⁆⁆

If the syntax allows the insertion of an LRM in locations that are not part of the literal message and ignores those LRM, the source can be consistently left-to-right (so that the syntax is in the right order).

<LRM>* ⁅⁅YOU HAVE ⁅<LRM>$count :number⁆ NOTIFICATION.⁆⁆

Resulting Display:

\* ⁅⁅EVAH UOY ⁅$count :number⁆ NOITACIFITON.⁆⁆

That is still not good for BIDI languages: the ideal would be that the flow of the syntax was LTR, but the flow of the variant message would be consistently RTL, something like:

<RLM>* ⁅⁅YOU HAVE ⁅<RLM><LRM>$count :number⁆ NOTIFICATION.⁆⁆

Resulting Display:
⁅⁅.NOITACIFITON ⁅$count  :number⁆ EVAH UOY⁆⁆ *

A good "MF2" editor or localization tool would use something more sophisticated (we should call that to people's attention in the spec, pointing to https://www.unicode.org/reports/tr55, especially https://www.unicode.org/reports/tr55/#Ordering for applying HL4). 

But a lot of people will be using plain-text editors, and allowing the entry of LRM and RLM can make the syntax more readable. (And tooling can insert those characters in the appropriate places automatically, so that plain text looks better.)

The tool is:
https://util.unicode.org/UnicodeJsps/bidi.jsp?a=one+%E2%81%85%E2%81%85YOU+HAVE+%E2%81%85%24count%E2%81%86+NOTIFICATION.%E2%81%86%E2%81%86&p=Auto&hack=on

Notes:
- What it does not show is that the brackets ⁅...⁆ are mirrored if displayed R to L, so I did that manually above. 
- If you insert either x or X in the Source (and ignore them in the Reordered output) you can see what the ordering result would be.

aphillips:
Adds the bidi strong marks ALM, RLM, and LRM plus the bidi isolate controls LRI, RLI, FSI, and PDI to the syntax.

Formally defines optional vs. non-optional whitespace.

Non-optional whitespace must include at least one whitespace character. Optional whitespace may contain only bidi marks (which are invisible).

This replaces PR#673. This fixes #661, #847.

TODO: Add guidance on "strict" bidi.

### DIFF
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -12,12 +12,12 @@ complex-body      = quoted-pattern / matcher
 input-declaration = input owsp variable-expression
 local-declaration = local wsp variable owsp "=" owsp expression
 
-quoted-pattern    = "{{" pattern "}}"
+quoted-pattern    = owsp "{{" pattern "}}" owsp
 
 matcher           = match-statement wsp variant *(owsp variant)
 match-statement   = match 1*(wsp selector)
 selector          = variable
-variant           = key *(wsp key) owsp quoted-pattern
+variant           = owsp key *(wsp key) quoted-pattern
 key               = literal / "*"
 
 ; Expressions
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -1,41 +1,41 @@
 message           = simple-message / complex-message
 
-simple-message    = owsp [simple-start pattern]
+simple-message    = o [simple-start pattern]
 simple-start      = simple-start-char / escaped-char / placeholder
 pattern           = *(text-char / escaped-char / placeholder)
 placeholder       = expression / markup
 
-complex-message   = owsp *(declaration owsp) complex-body owsp
+complex-message   = o *(declaration o) complex-body o
 declaration       = input-declaration / local-declaration
 complex-body      = quoted-pattern / matcher
 
-input-declaration = input owsp variable-expression
-local-declaration = local wsp variable owsp "=" owsp expression
+input-declaration = input o variable-expression
+local-declaration = local s variable o "=" o expression
 
-quoted-pattern    = owsp "{{" pattern "}}" owsp
+quoted-pattern    = o "{{" pattern "}}" o
 
-matcher           = match-statement wsp variant *(owsp variant)
-match-statement   = match 1*(wsp selector)
+matcher           = match-statement s variant *(o variant)
+match-statement   = match 1*(s selector)
 selector          = variable
-variant           = owsp key *(wsp key) quoted-pattern
+variant           = o key *(s key) quoted-pattern
 key               = literal / "*"
 
 ; Expressions
 expression          = literal-expression
                     / variable-expression
                     / function-expression
-literal-expression  = "{" owsp literal [wsp function] *(wsp attribute) owsp "}"
-variable-expression = "{" owsp variable [wsp function] *(wsp attribute) owsp "}"
-function-expression = "{" owsp function *(wsp attribute) owsp "}"
+literal-expression  = "{" o literal [s function] *(s attribute) o "}"
+variable-expression = "{" o variable [s function] *(s attribute) o "}"
+function-expression = "{" o function *(s attribute) o "}"
 
-markup = "{" owsp "#" identifier *(wsp option) *(wsp attribute) owsp ["/"] "}"  ; open and standalone
-       / "{" owsp "/" identifier *(wsp option) *(wsp attribute) owsp "}"  ; close
+markup = "{" o "#" identifier *(s option) *(s attribute) o ["/"] "}"  ; open and standalone
+       / "{" o "/" identifier *(s option) *(s attribute) o "}"  ; close
 
 ; Expression and literal parts
-function       = ":" identifier *(wsp option)
-option         = identifier owsp "=" owsp (literal / variable)
+function       = ":" identifier *(s option)
+option         = identifier o "=" o (literal / variable)
 
-attribute      = "@" identifier [owsp "=" owsp (literal / variable)]
+attribute      = "@" identifier [o "=" o (literal / variable)]
 
 variable       = "$" name
 
@@ -66,8 +66,8 @@ name-char  = name-start / DIGIT / "-" / "."
 
 ; Restrictions on characters in various contexts
 simple-start-char = content-char / "@" / "|"
-text-char         = content-char / s / "." / "@" / "|"
-quoted-char       = content-char / s / "." / "@" / "{" / "}"
+text-char         = content-char / ws / "." / "@" / "|"
+quoted-char       = content-char / ws / "." / "@" / "{" / "}"
 content-char      = %x01-08        ; omit NULL (%x00), HTAB (%x09) and LF (%x0A)
                   / %x0B-0C        ; omit CR (%x0D)
                   / %x0E-1F        ; omit SP (%x20)
@@ -84,14 +84,14 @@ escaped-char = backslash ( backslash / "{" / "|" / "}" )
 backslash    = %x5C ; U+005C REVERSE SOLIDUS "\"
 
 ; Optional whitespace
-owsp = *(s / bidi)
+o = *(ws / bidi)
 
 ; Required whitespace
-wsp = (owsp) 1*s (owsp)
+s = *bidi ws *o
 
 ; Bidirectional marks and isolates
 ; ALM / LRM / RLM / LRI, RLI, FSI & PDI
 bidi = %x061C / %x200E / %x200F / %x2066-2069
 
 ; Whitespace characters
-s = SP / HTAB / CR / LF / %x3000
+ws = SP / HTAB / CR / LF / %x3000
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -17,7 +17,7 @@ quoted-pattern    = o "{{" pattern "}}"
 matcher           = match-statement s variant *(o variant)
 match-statement   = match 1*(s selector)
 selector          = variable
-variant           = o key *(s key) quoted-pattern
+variant           = key *(s key) quoted-pattern
 key               = literal / "*"
 
 ; Expressions
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -12,7 +12,7 @@ complex-body      = quoted-pattern / matcher
 input-declaration = input o variable-expression
 local-declaration = local s variable o "=" o expression
 
-quoted-pattern    = o "{{" pattern "}}" o
+quoted-pattern    = o "{{" pattern "}}"
 
 matcher           = match-statement s variant *(o variant)
 match-statement   = match 1*(s selector)
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -83,11 +83,11 @@ content-char      = %x01-08        ; omit NULL (%x00), HTAB (%x09) and LF (%x0A)
 escaped-char = backslash ( backslash / "{" / "|" / "}" )
 backslash    = %x5C ; U+005C REVERSE SOLIDUS "\"
 
-; Optional whitespace
-o = *(ws / bidi)
-
 ; Required whitespace
 s = *bidi ws o
+
+; Optional whitespace
+o = *(ws / bidi)
 
 ; Bidirectional marks and isolates
 ; ALM / LRM / RLM / LRI, RLI, FSI & PDI
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -1,41 +1,41 @@
 message           = simple-message / complex-message
 
-simple-message    = [s] [simple-start pattern]
+simple-message    = owsp [simple-start pattern]
 simple-start      = simple-start-char / escaped-char / placeholder
 pattern           = *(text-char / escaped-char / placeholder)
 placeholder       = expression / markup
 
-complex-message   = [s] *(declaration [s]) complex-body [s]
+complex-message   = owsp *(declaration owsp) complex-body owsp
 declaration       = input-declaration / local-declaration
 complex-body      = quoted-pattern / matcher
 
-input-declaration = input [s] variable-expression
-local-declaration = local s variable [s] "=" [s] expression
+input-declaration = input owsp variable-expression
+local-declaration = local wsp variable owsp "=" owsp expression
 
 quoted-pattern    = "{{" pattern "}}"
 
-matcher           = match-statement s variant *([s] variant)
-match-statement   = match 1*(s selector)
+matcher           = match-statement wsp variant *(owsp variant)
+match-statement   = match 1*(wsp selector)
 selector          = variable
-variant           = key *(s key) [s] quoted-pattern
+variant           = key *(wsp key) owsp quoted-pattern
 key               = literal / "*"
 
 ; Expressions
 expression          = literal-expression
                     / variable-expression
                     / function-expression
-literal-expression  = "{" [s] literal [s function] *(s attribute) [s] "}"
-variable-expression = "{" [s] variable [s function] *(s attribute) [s] "}"
-function-expression = "{" [s] function *(s attribute) [s] "}"
+literal-expression  = "{" owsp literal [wsp function] *(wsp attribute) owsp "}"
+variable-expression = "{" owsp variable [wsp function] *(wsp attribute) owsp "}"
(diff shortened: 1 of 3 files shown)


aphillips:
@eggrobin Pinging you to look at this, which reimplements the text you previously reviewed in our #673

eggrobin (approved):
This looks good. I see you are referring to UTS #&zwnj;55 in the PR title; in case you want to mention that you are following its recommendation in some informative part of the text, the relevant bit is _Section 3.2, [Whitespace and Syntax](https://www.unicode.org/reports/tr55/#Whitespace-Syntax)_.

spec/message.abnf
```diff
@@ -83,5 +83,15 @@ content-char      = %x01-08        ; omit NULL (%x00), HTAB (%x09) and LF (%x0A)
...
+
+; Required whitespace
+wsp = (owsp) 1*s (owsp)
+
+; Bidirectional marks and isolates
+; ALM / LRM / RLM / LRI, RLI, FSI & PDI
+bidi = %x061C / %x200E / %x200F / %x2066-2069
+
+; Whitespace characters
+s = SP / HTAB / CR / LF / %x3000
```
eemeli, line 97: I find replacing `[s]` and `s` with `owsp` and `wsp` to make the ABNF harder to read, as these longer names make it difficult to "see" them as representations of whitespace.

Could we do something like this instead?

```suggestion
; Optional whitespace
_ = *(ws / bidi)

; Required whitespace
s = _ ws _

; Bidirectional marks and isolates
; ALM / LRM / RLM / LRI, RLI, FSI & PDI
bidi = %x061C / %x200E / %x200F / %x2066-2069

; Whitespace characters
ws = SP / HTAB / CR / LF / %x3000
```

spec/syntax.md
```diff
@@ -803,24 +809,104 @@ and inside _patterns_ only escape `{` and `}`.
...
+following mechanisms to make messages display intelligibly in plain-text editors:
+
+1. Use paired isolating bidi controls `U+2066 LEFT-TO-RIGHT ISOLATE`
+   and `U+2069 POP DIRECTIONAL ISOLATE` as permitted by the ABNF around
+   parts of any _message_ containing RTL characters:
+   - _inside_ of _placeholder_ markers `{` and `}` 
+   - _outside_ _quoted-pattern_ markers `{{` and `}}`
+   - _identifiers_
+   - _literals_ (This is especially important for individual _keys_ in a _variant_)
+   - _option_ values
```
eemeli, line 844: Why should a message with RTL contents isolate them as LTR rather than RTL? And why do option values need a special call-out, are they not covered by identifiers and literals?

spec/message.abnf
```diff
@@ -83,5 +83,15 @@ content-char      = %x01-08        ; omit NULL (%x00), HTAB (%x09) and LF (%x0A)
...
-s = 1*( SP / HTAB / CR / LF / %x3000 )
+; Optional whitespace
+owsp = *(s / bidi)
+
+; Required whitespace
+wsp = (owsp) 1*s (owsp)
+
+; Bidirectional marks and isolates
+; ALM / LRM / RLM / LRI, RLI, FSI & PDI
+bidi = %x061C / %x200E / %x200F / %x2066-2069
```
eemeli, line 94: Note that ALM is also permitted in `name-start` and `name-char`. We need to either drop it from here or there.

eemeli (changes requested):
As discussed previously e.g. in https://github.com/unicode-org/message-format-wg/pull/871#discussion_r1734189401, I believe that we need to also allow for a bidi character before `name` and before the namespace delimiter.

Theses control characters would not be considered as a part of the parsed value of the `name`, much like the `|` delimiters are not a part of the parsed value of a `quoted-literal`.

```diff
-identifier = [namespace ":"] name
+identifier = [namespace [bidi] ":"] name
 namespace  = name
-name = name-start *name-char
+name = [bidi] name-start *name-char
```

spec/message.abnf
aphillips: `_` isn't a valid ABNF rule name:

```abnf
         rulename       =  ALPHA *(ALPHA / DIGIT / "-")
```

Alternatives might be `s` vs. `o` (optional) or perhaps `s` (required) and `s-` (optional)

spec/syntax.md
aphillips: LRI forces the neutral sigil to the left side of the token in items like `$مصر⁩` whereas RLI will force the neutral sigil to the right side. We separately provide the ability (as soon as I address your comment, that is...) to put a strong directional mark to fix neutrals and weaks that are part of the name itself.

You're correct that this list needs help.

aphillips: should be fixed.

spec/syntax.md
```diff
@@ -803,24 +809,104 @@ and inside _patterns_ only escape `{` and `}`.
...
+the following pattern whitespace characters are not allowed:
+`U+000B FORM FEED`, 
+`U+000C VERTICAL TABULATION`, 
+`U+0085 NEXT LINE`, 
+`U+2028 LINE SEPARATOR` and 
+`U+2029 PARAGRAPH SEPARATOR`;
+the character `U+3000 IDEOGRAPHIC SPACE`
+_is_ interpreted as whitespace,
+ and the directional isolates U+2066..U+2069
+ are treated as ignorable format controls.
```
eggrobin, line 896: So is U+061C, I think?

spec/message.abnf
eemeli: Ah, you're right. I think `s` and `o` is probably the least worst option here. `s-` looks like a typo.

eemeli: Also, to reduce nominal lookahead, I think the required whitespace rule ought to look like this: 

```abnf
s = *bidi ws o
```

spec/syntax.md
```diff
@@ -733,7 +733,13 @@ A **_<dfn>name</dfn>_** is a character sequence used in an _identifier_
 or as the name for a _variable_
 or the value of an _unquoted literal_.
 
-_Variable_ names are prefixed with `$`.
+A _name_ can be preceded or followed by bidirectional marks or isolating controls
+to aid in presenting names that contain right-to-left or neutral characters.
+These characters are **not** part of the _name_ and MUST be treated as if they were not present
```
eemeli, line 738:
```suggestion
These characters are **not** part of the value of the _name_ and MUST be treated as if they were not present
```

spec/syntax.md
```diff
@@ -733,7 +733,13 @@ A **_<dfn>name</dfn>_** is a character sequence used in an _identifier_
...
+when matching _name_ or _identifier_ strings or _unquoted literal_ values.
+Implementations MAY remove these characters from a _message_.
```
eemeli, line 740: What does this mean? When/why/how would these characters be removed?

spec/syntax.md
```diff
@@ -834,14 +840,14 @@ _option values_, and _keys_)
 _Messages_ that contain right-to-left (aka RTL) characters SHOULD use one of the 
 following mechanisms to make messages display intelligibly in plain-text editors:
```
eemeli, line 841: Before getting much deeper into discussing the mechanisms, could you clarify if this is intended to be the canonical/recommended way of isolating or marking messages with RTL contents that we've discussed, or is that something that'll be provided separately?

spec/syntax.md
```diff
@@ -834,14 +840,14 @@ _option values_, and _keys_)
...
+   and `U+2069 POP DIRECTIONAL ISOLATE` ("PDI") as permitted by the ABNF around
    parts of any _message_ containing RTL characters:
    - _inside_ of _placeholder_ markers `{` and `}` 
    - _outside_ _quoted-pattern_ markers `{{` and `}}`
-   - _identifiers_
-   - _literals_ (This is especially important for individual _keys_ in a _variant_)
-   - _option_ values
+   - _outside_ of _literals_, paying particular attention to _keys_ in a _variant_
+   - _outside_ of _variable_, _function_, _markup_, or _attribute_ _names_/_identifiers_,
+     including the identifying sigil (e.g. `<LRI>$var</PDI>` or `<LRI>:ns:name</PDI>`)
```
eemeli, line 850: I still don't understand a recommendation to isolate right-to-left content as left-to-right. It makes sense for stuff with prefixes (_variable_, _function_, _markup_, and _attribute_) if the surrounding context isn't LTR, but isn't it going to break the rendering of RTL _literals_ and _names_ that start with neutral characters?

spec/syntax.md
aphillips: Yes, and LRM/RLM

eggrobin: Right, but LRM and RLM are part of R3a-1, see the first note under https://www.unicode.org/reports/tr31/#R3a.

The profile includes adding U+061C to the ignorable format controls, not just the isolates.

eggrobin (approved):

spec/syntax.md
aphillips: > ... but isn't it going to break the rendering of RTL literals and names that start with neutral characters?

Yes, if it were used by itself. But (2) covers that. Using LRI/PDI gets the complete "token" isolated, with the sigil on the left side and the tokens themselves ordered LTR. Using strong marks _inside_ the token (before/after `name` or unquoted literals) then fixes problems with neutrals and weaks.

spec/syntax.md
aphillips: For example, an implementation might remove the strong marks from around names when storing the names (in order to facilitate faster comparison). An implementation might also remove them when serializing, at the cost of some legibility.

I could just get rid of this line.

spec/syntax.md
aphillips: No, that's separate.

spec/message.abnf
aphillips: The `o` would need a `*`, I think:

```abnf
s = *bidi ws *o
```

spec/syntax.md
aphillips: The second note talks about ALM, including:

> If it is added to the set of whitespace characters by a profile, it is interpreted as an ignorable format control. 

In any case, I now have a list of ignorable format controls, which might be overkill, but saves reading rule R3a 😉.

eggrobin: > If it is added to the set of whitespace characters by a profile, it is interpreted as an ignorable format control.

Indeed, but you still need to say that the profile adds it!

I agree that listing the set is probably better than the diff at this point.

spec/syntax.md
eemeli: I think we're already permitting that with the preceding sentence, so removing this would be clearer.

```suggestion
```

spec/message.abnf
eemeli: No need, an empty string satisfies `o` as it's got the `*` built into it.

spec/syntax.md
eemeli: We could also not allow ALM as a `bidi` character, as there should be no place in the syntax where an RLM couldn't be used just as well.

aphillips: I feel uncomfortable removing bidi marks. The ALM was added many years after RLM/LRM and its differences with RLM are minor. But we want bidi language users to have the tools they need to make things look right (and still be functional). I would hate to remove it because we need to add a couple of words to the spec.

spec/syntax.md
eemeli: I get that for the productions which include a leading sigil. But I do not understand the recommendation to add LRI..PDI isolation "_outside_ of _literals_" or "_outside_ of _variable_, _function_, _markup_, or _attribute_ _names_/_identifiers_". These productions do not include a leading sigil which is expected to stay on the left. They only include _literal_, _name_, or _identifier_.

I think we should recommend here something like this instead:
```suggestion
   - _outside_ of _variable_, _function_, _markup_, or _attribute_,
     including the identifying sigil (e.g. `<LRI>$var</PDI>` or `<LRI>:ns:name</PDI>`)
```

spec/syntax.md
eemeli: My main concern here is that treating it as an ignorable format control requires us to deviate further from the XML name production.

spec/syntax.md
aphillips: > "outside of variable, function, markup, or attribute names/identifiers"

This is `$foo`, `:foo`, `#foo`/`/foo`, and `@foo` values. They need an LRI to keep bidi from reordering them. Your suggestion is an improvement, though, as "identifier" confuses things.

spec/message.abnf
```diff
@@ -1,41 +1,41 @@
...
 declaration       = input-declaration / local-declaration
 complex-body      = quoted-pattern / matcher
 
-input-declaration = input [s] variable-expression
-local-declaration = local s variable [s] "=" [s] expression
+input-declaration = input o variable-expression
+local-declaration = local s variable o "=" o expression
 
-quoted-pattern    = "{{" pattern "}}"
+quoted-pattern    = o "{{" pattern "}}" o
```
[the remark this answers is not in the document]
aphillips: I agree that it is. This is more like giving explicit permission to attach bidi controls (and trailing whitespace) to the quoted pattern. But we can remove it.

spec/message.abnf
```diff
@@ -1,41 +1,41 @@
...
 
-matcher           = match-statement s variant *([s] variant)
+matcher           = match-statement s variant *(o variant)
 match-statement   = match 1*(s selector)
 selector          = variable
-variant           = key *(s key) [s] quoted-pattern
+variant           = o key *(s key) quoted-pattern
```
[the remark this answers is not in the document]
aphillips: This is the same thing.

spec/message.abnf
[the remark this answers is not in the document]
aphillips: I did remove it.

spec/syntax.md
aphillips: I think this change is the right thing to do.

ALM is an invisible, default-ignorable, non-spacing code point. As noted elsewhere, it was added to Unicode _after_ XML/XMLName were defined. According to XML's rules, an ALM all-by-itself is a valid identifier. That seems like a bug, not a feature. Maybe we should call out the deviation more clearly and maybe (wearing my _other_ chair hat) W3C should be called on to do an erratum. 

@macchiati Any thoughts?

macchiati: I strongly agree; it should be added.

spec/syntax.md
```diff
@@ -803,24 +814,112 @@ and inside _patterns_ only escape `{` and `}`.
...
 > compatibility with certain East Asian keyboards and input methods,
 > in which users might accidentally create these characters in a _message_.
 
 ```abnf
-s = 1*( SP / HTAB / CR / LF / %x3000 )
+; Optional whitespace
+o = *(s / bidi)
+
+; Required whitespace
+s = *bidi ws o
```
eemeli, line 915:
```suggestion
; Required whitespace
s = *bidi ws o

; Optional whitespace
o = *(s / bidi)
```

eemeli (approved):
See inline for a few final nitpicks, but I think this is good to go.

mihnita (approved):

PR #885, opened 2024-09-13
Address name and literal equality

aphillips:
This change defines equality as discussed in the 2024-09-09 teleconference in the following ways:

- It defines _name_ equality as being under NFC
- It defines _literal_ equality as explicitly **not** under NFC
- It moves _name_ before _identifier_ in that section of text to avoid a forward definition.

Note that this deviates from discussion in 2024-09-09's call in that we didn't discuss literals at length. It also doesn't discuss non-name/non-literal values, which I'll point out are limited to ASCII sequences such as keywords.

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -474,11 +474,6 @@ This selection method is defined in more detail below.
 An implementation MAY use any pattern selection method,
 as long as its observable behavior matches the results of the method defined here.
 
-The resolved value of each _key_ MUST be in Unicode Normalization Form C ("NFC"),
-even if the _literal_ for the _key_ is not.
-All comparisons of _keys_ MUST be done on the canonical, normalized values
-and the normalized value MUST be the value that is passed in the steps below.
-
 ### Resolve Selectors
 
 First, resolve the values of each _selector_:
@@ -520,6 +515,9 @@ The returned list MUST contain only unique elements of the input list `keys`.
 The returned list MAY be empty.
 The most-preferred key is first,
 with each successive key appearing in order by decreasing preference.
+
+The resolved value of each _key_ MUST be in Unicode Normalization Form C ("NFC"),
+even if the _literal_ for the _key_ is not.
 
 If calling MatchSelectorKeys encounters any error,
 a _Bad Selector_ error is emitted
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -474,6 +474,11 @@ This selection method is defined in more detail below.
 An implementation MAY use any pattern selection method,
 as long as its observable behavior matches the results of the method defined here.
 
+The resolved value of each _key_ MUST be in Unicode Normalization Form C ("NFC"),
+even if the _literal_ for the _key_ is not.
+All comparisons of _keys_ MUST be done on the canonical, normalized values
+and the normalized value MUST be the value that is passed in the steps below.
+
 ### Resolve Selectors
 
 First, resolve the values of each _selector_:
@@ -502,7 +507,7 @@ Next, using `res`, resolve the preferential order for all message keys:
       1. Let `key` be the `var` key at position `i`.
       1. If `key` is not the catch-all key `'*'`:
          1. Assert that `key` is a _literal_.
-         1. Let `ks` be the resolved value of `key`.
+         1. Let `ks` be the resolved value of `key` in Unicode Normalization Form C.
          1. Append `ks` as the last element of the list `keys`.
    1. Let `rv` be the resolved value at index `i` of `res`.
    1. Let `matches` be the result of calling the method MatchSelectorKeys(`rv`, `keys`)
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -753,7 +753,7 @@ that is, if they consist of the same sequence of Unicode code points after
 has been applied to both.
 
 > [!NOTE]
-> Implementations are not required to normalize _names_.
+> Implementations are not required to normalize all _names_.
 > Comparisons of _name_ values only need be done "as-if" normalization
 > has occured.
 > Since most text in the wild is already in NFC
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -684,7 +684,7 @@ except for U+0000 NULL or the surrogate code points U+D800 through U+DFFF.
 
 All code points are preserved.
 
-Two literals are considered equal if the consist of the same sequence of Unicode
+Two _literals_ are considered equal if they consist of the same sequence of Unicode
 code points.
 
 > [!IMPORTANT]
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -684,6 +684,17 @@ except for U+0000 NULL or the surrogate code points U+D800 through U+DFFF.
 
 All code points are preserved.
 
+Two literals are considered equal if the consist of the same sequence of Unicode
+code points.
+
+> [!IMPORTANT]
+> _Literal_ equality is different from _name_ equality in that
+> Unicode Normalization is not applied to _literal_ values before comparison.
+> Users are cautioned to ensure that they use the same character sequences
+> for equivalent values.
+> The use of [Normalization Form C]((https://unicode.org/reports/tr15/) for all
+> _literal_ values is RECOMMENDED.
+
 A **_<dfn>quoted literal</dfn>_** begins and ends with U+005E VERTICAL BAR `|`.
 The characters `\` and `|` within a _quoted literal_ MUST be
 escaped as `\\` and `\|`.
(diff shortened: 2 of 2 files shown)


spec/syntax.md
```diff
@@ -684,6 +684,17 @@ except for U+0000 NULL or the surrogate code points U+D800 through U+DFFF.
 
 All code points are preserved.
 
+Two _literals_ are considered equal if they consist of the same sequence of Unicode
+code points.
```
eemeli, line 688: This seems a bit excessive. While most _name_ comparisons are internal to the spec, AFAIK the only _literal_ comparison in the spec is for duplicate variant key lists, which tbh I'd prefer to be done with normalization.

All other _literal_ value handling is done by functions, which we should not restrict from applying normalization in their internal processing.

eemeli (changes requested):

spec/syntax.md
macchiati: I agree. I'd go a bit further in encouraging function specifiers to normalize when matching: and we make the standard functions do so. That is, something along the lines of:

----

An NFC comparison (aka Unicode canonically equivalent comparison)  produces the same results as if each string value being compared were converted to the Unicode Normalization Form C (NFC). For example, with an NFC comparison against the literal |U\x{3308}|, the same result is obtained as if the literal were |\x{DC}|. For more examples, see the [Unicode Standard](https://www.unicode.org/versions/latest/core-spec/chapter-2/#G287424).

When determining whether two variant key lists are duplicates, NFC comparison MUST be used for literals.

When a selector function evaluates matches to literal keys, the matches SHOULD use NFC comparison. Moreover, the implementation of the standard selector functions MUST use NFC comparison. Thus the standard :string selector function MUST match a string input parameter of "U\x{3308}" with the literal |\x{DC}|. 

----

BTW: Some selector functions, such as the standard numeric selectors, only match literals with all ASCII characters. ASCII literals never change when converted to NFC, and there are only 3 non ASCII characters that change to ASCII. So selector functions whose literals don't include ";", "`", or "K" don't need to use NFC comparison; that includes our numeric selectors.

Char. | Code Point | Name
-- | -- | --
; | U+037E | GREEK QUESTION MARK
` | U+1FEF | GREEK VARIA
K | U+212A | KELVIN SIGN

macchiati: I think it is far, far more likely that people will make mistakes with non-NFC literals (or input) than the really, really obscure edge case of someone wanting to match non-normalized text.

eemeli: > First, do we agree that literals MAY be non-normalized/denormalized?

As with pattern text, I agree that we should not require the normalization of literal values.

> Literal comparison is [...] also for matching between the sorted results of a selector and the keys in the message (the sorting is done by a function, but not the matching after sorting).

Regarding the latter, we say this: https://github.com/unicode-org/message-format-wg/blob/80bec5215a1a3fe10abdbd75a9489f9f42d470d7/spec/formatting.md?plain=1#L511-L514

That MUST is requiring the processing to not normalise any of the values, even if it did so for its internal processing.

I'd be completely fine with us normalising the keys before they're passed to the function, or at least allowing an implementation to do so.

> > When determining whether two variant key lists are duplicates, NFC comparison MUST be used for literals.
> 
> This is the opposite of what @eemeli is saying? If we allow normalization (but don't require it) we also allow the lack of it.

I'm aligned with @macchiati here. We don't _need_ to normalize key values, but we should do their comparison when checking for duplicate key lists as if they were normalized.

aphillips:
In the 2024-09-16 call we agreed to a variety of changes to this PR. Keys, option names, and attribute names to be NFC.

spec/syntax.md
```diff
@@ -438,6 +438,14 @@ A _key_ can be either a _literal_ value or the "catch-all" key `*`.
 The **_<dfn>catch-all key</dfn>_** is a special key, represented by `*`,
 that matches all values for a given _selector_.
 
+The value of each _key_ MUST be treated as if it were in
+[Unicode Normalization Form C](https://unicode.org/reports/tr15/) ("NFC").
+When _keys_ are passed during _pattern selection_, the _key_ values MUST
+be normalized into NFC.
```
eemeli, line 450: This should probably be in the relevant part of the formatting section, and not here.

spec/syntax.md
```diff
@@ -740,6 +747,20 @@ when matching _name_ or _identifier_ strings or _unquoted literal_ values.
 
 _Variable_ _names_ are prefixed with `$`.
 
+Two _names_ are considered equal if they are canonically equivalent strings,
+that is, if they consist of the same sequence of Unicode code points after
+[Unicode Normalization Form C](https://unicode.org/reports/tr15/) ("NFC")
+has been applied to both.
+
+> [!NOTE]
+> Implementations are not required to normalize _names_.
```
eemeli, line 756: Should clarify, if/as we're normalising key values.
```suggestion
> Implementations are not required to normalize all _names_.
```

eemeli (commented):
The bit about key values being normalised should be reflected here: https://github.com/unicode-org/message-format-wg/blob/95ec6d53220d7b61c70ebef5647d65592f3e2e62/spec/formatting.md?plain=1#L505-L506

Maybe with a change like this?
```diff
-         1. Let `ks` be the resolved value of `key`.
+         1. Let `ks` be the resolved value of `key` in Unicode Normalization Form C.
```

spec/syntax.md
aphillips: Keys are not names?

```abnf
key               = literal / "*"
```

spec/syntax.md
aphillips: I think maybe having it in both is a good idea?

spec/syntax.md
eemeli: They can be when you go down deeper.
```abnf
literal = quoted-literal / unquoted-literal
unquoted-literal = name / number-literal
```

spec/syntax.md
eemeli: At least it needs to be in Formatting. I don't object to it also being in Syntax, but this isn't really about syntax at all.

spec/syntax.md
aphillips: It doesn't really mean "name" there, though. I think that's us saving effort by defining what can be unquoted by recycling `name`. Ignoring the quoted/unquoted part syntactical representation, a key is just a literal value.

spec/syntax.md
aphillips: See the commit just below, which you'll probably have things to say about 🙈

spec/syntax.md
eemeli: Somewhat, yes, but we do say this: https://github.com/unicode-org/message-format-wg/blob/95ec6d53220d7b61c70ebef5647d65592f3e2e62/spec/syntax.md?plain=1#L732-L734

Rather than redefining _name_, I think it'd be easier to just include the "all" here. That way we don't need to explain how sometimes a `name` is not a _name_.

spec/formatting.md
```diff
@@ -474,6 +474,11 @@ This selection method is defined in more detail below.
 An implementation MAY use any pattern selection method,
 as long as its observable behavior matches the results of the method defined here.
 
+The resolved value of each _key_ MUST be in Unicode Normalization Form C ("NFC"),
+even if the _literal_ for the _key_ is not.
+All comparisons of _keys_ MUST be done on the canonical, normalized values
+and the normalized value MUST be the value that is passed in the steps below.
```
eemeli, line 480: This should probably be further below, after the paragraph starting with "The method MatchSelectorKeys is determined ...".

It's also very MUST...MUST...MUST. Is that really necessary? Does the second sentence really add anything beyond the first?

aphillips: Agreed. 

There are the resolved values passed to MatchSelectorKeys, but then there is the matching done checking for duplicates. I could leave out the third MUST... perhaps:

```suggestion
The resolved value of each _key_ MUST be in Unicode Normalization Form C ("NFC"),
even if the _literal_ for the _key_ is not, 
and all comparisons of _keys_ MUST be done on this canonical, normalized form.
```

eemeli: Better. Do note that the check for duplicates is a syntax/data model matter; this is about pattern selection. Any MUST about the former ought to be (and is already) under Syntax; it should not be here.

aphillips: Done

spec/syntax.md
eemeli: Drop these lines, given the Formatting additions?
```suggestion
```

spec/syntax.md
aphillips: My reluctance here is that "all" suggests that implementations might be required to normalize some names that aren't keys. In fact, not even keys are required to be stored in NFC. But it's such a small point that I'll commit your suggestion now.

eemeli (approved):
This is an improvement on what we have now. I'd be happier if the suggestion from https://github.com/unicode-org/message-format-wg/pull/885#discussion_r1763692401 was included, but not going to insist on it.

As discussed on the call, I think we should also normalise option names, but that doesn't need to be a part of this change.

spec/syntax.md
aphillips: Done

aphillips:
In spite of having only one approval, I'm going to merge this, based on consensus from 2024-09-16.

PR #894, opened 2024-09-19
Fix attribute value to be literal

### LINKED ISSUE #893: [FEEDBACK] incorrect syntax error test case?
The syntax looks good to me in the following test case.

https://github.com/unicode-org/message-format-wg/blob/57740d8fd9c63f62ae7c6f0de0b55da76dc526ea/test/tests/syntax-errors.json#L126

Why is it a syntax error?

aphillips:
Fixes #893

See discussion there.


macchiati (approved):

eemeli (approved):
Not sure how we missed this in #845.

PR #901, opened 2024-10-08
Clarify note about eager vs. lazy evaluation

### LINKED ISSUE #784: Spec language on eager vs. lazy evaluation should be clarified
In #753 , @echeran brought up the example of a custom function `getSystemTimeInMillisNow()` that would return the current time.

In the [spec intro](https://github.com/unicode-org/message-format-wg/blob/main/spec/formatting.md#introduction) we currently have the language:

"This specification does not require either eager or lazy expression resolution of message parts; do not construe any requirement in this document as requiring either."

Consider the following example (which formats to a string with the current time repeated twice... or does it?):

```
.local $x = {:getSystemTimeInMillisNow}
{{{$x} {$x}}}
```

In an eager implementation, this is guaranteed to print the same time twice.

In a lazy call-by-need implementation, where the right-hand side of `$x` is guaranteed to be evaluated at most once, it's also guaranteed to print the same time twice.

In a lazy call-by-name implementation, it may print out two different times.

In general, custom functions might have side effects observable by MF2, and getting the current time is a reasonable example of one. That means one can't assume that call-by-need and call-by-name are going to be equivalent. 

I think all that's needed is an editorial note pointing out that depending on the contents of the custom function registry, call-by-need and call-by-name may be observably different and the implementor should think about which one is the least surprising.

catamorphism:
Closes #784

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -58,14 +58,19 @@ nor be made available to _function handlers_.
 > have already been evaluated in the order in which the relevant _declarations_
 > appear in the _message_.
 >
-> Users or implementations can provide functions that have observable side effects
-> or whose results depend on external mutable state (for example, a function
-> that returns the current time and date.)
+> In some environments, users or implementations can provide functions
+> that mutate state that is external to the message formatter
+> (for example, deleting a file in the filesystem). This is not recommended.
+> Functions can also be written that depend on external mutable state
+> (for example, the current date and time).
 > Lazy evaluation might involve evaluating the same _expression_ multiple times
 > (call-by-name) or evaluating every expression at most once (call-by-need).
-> In the presence of custom functions with side effects, this implementation
+> In the presence of custom functions that either depend on or modify
+> mutable state external to the message formatter, this implementation
 > choice affects the result of formatting a message.
-> Functions that mutate external state are not recommended.
+> If all functions return a result that depends only on their _operand_
+> and _options_, then the choice of evaluation strategy has no effect
+> on the result of formatting a message.
 
 ## Formatting Context
 
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -57,6 +57,12 @@ nor be made available to _function handlers_.
 > _declarations_ affecting _variables_ referenced by that _expression_
 > have already been evaluated in the order in which the relevant _declarations_
 > appear in the _message_.
+>
+> Users may write custom functions that have observable side effects.
+> Lazy evaluation may involve evaluating the same expression multiple times
+> (call-by-name) or evaluating every expression at most once (call-by-need).
+> In the presence of custom functions with side effect, this implementation
+> choice affects the result of formatting a message.
 
 ## Formatting Context
 
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -58,19 +58,20 @@ nor be made available to _function handlers_.
 > have already been evaluated in the order in which the relevant _declarations_
 > appear in the _message_.
 >
-> In some environments, users or implementations can provide functions
-> that mutate state that is external to the message formatter
-(for example, setting or modifying the system locale).
-> Functions can also be written that depend on external mutable state
-> (for example, the current date and time).
-> Lazy evaluation might involve evaluating the same _expression_ multiple times
-> (call-by-name) or evaluating every expression at most once (call-by-need).
-> In the presence of custom functions that either depend on or modify
-> mutable state external to the message formatter, this implementation
-> choice affects the result of formatting a message.
-> If all functions return a result that depends only on their _operand_
-> and _options_, then the choice of evaluation strategy has no effect
-> on the result of formatting a message.
+> Implementations and users creating custom functions SHOULD avoid
+> creating _function handlers_ that mutate external program state.
+>
+>> [!NOTE]
+>> Function handlers can also be written that depend on external mutable state
+>> (for example, the current date and time).
+>> Lazy evaluation might involve evaluating the same _expression_ multiple times
+>> (call-by-name) or evaluating every expression at most once (call-by-need).
+>> In the presence of custom function handlers that either depend on or modify
+>> mutable state external to the message formatter, this implementation
+>> choice affects the result of formatting a message.
+>> If all function handlers return a result that depends only on their _operand_
+>> and _options_, then the choice of evaluation strategy has no effect
+>> on the result of formatting a message.
 
 ## Formatting Context
 
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -60,23 +60,23 @@ nor be made available to _function handlers_.
 > An implementation MUST ensure that every _expression_ in a _message_
 > is evaluated at most once.
 
+> [!NOTE]
+>
+> Implementations with lazy evaluation MUST NOT use a
+> call-by-name evaluation strategy. Instead, they must evaluate expressions
+> at most once ("call-by-need").
+> This is to prevent _expressions_ from having different values
+> when used in different parts of a given _message_.
+> _Function handlers_ are not necessarily pure: they can access
+> external mutable state such as the current system clock time.
+> Thus, evaluating the same _expression_ more than once
+> could yield different results. That behavior violates this specification.
+
 > [!IMPORTANT]
 > Implementations and users SHOULD NOT create _function handlers_
 > that mutate external program state,
 > particularly since such a _function handler_ can present a remote execution hazard.
 >
-
-> [!NOTE]
->
-> Function handlers can be written that depend on external mutable state
-> (for example, the current date and time).
-> Lazy evaluation might involve evaluating the same _expression_ multiple times
-> (call-by-name) or evaluating every expression at most once (call-by-need).
-> This justifies the requirement to ensure that every _expression_ is
-> evaluated at most once: different invocations of the same function handler
-> might return different results (functions are not pure).
-> This means that lazy implementations cannot use
-> a call-by-name evaluation strategy.
 
 ## Formatting Context
 
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -57,6 +57,8 @@ nor be made available to _function handlers_.
 > _declarations_ affecting _variables_ referenced by that _expression_
 > have already been evaluated in the order in which the relevant _declarations_
 > appear in the _message_.
+> An implementation MUST ensure that every _expression_ in a _message_
+> is evaluated at most once.
 >
 > Implementations and users SHOULD NOT create _function handlers_
 > that mutate external program state,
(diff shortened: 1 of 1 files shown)


spec/formatting.md
```diff
@@ -57,6 +57,12 @@ nor be made available to _function handlers_.
 > _declarations_ affecting _variables_ referenced by that _expression_
 > have already been evaluated in the order in which the relevant _declarations_
 > appear in the _message_.
+>
+> Users may write custom functions that have observable side effects.
+> Lazy evaluation may involve evaluating the same expression multiple times
```
aphillips, line 62: It's good practice to avoid the normative words like 'may' in non-normative contexts like this.

```suggestion
> Users or implementations can provide functions that have observable side effects.
> Lazy evaluation might involve evaluating the same _expression_ multiple times
```

Notice that I removed the word "custom" (which might be presumed?) and bring in implementations. 

Consider adding a health warning to this paragraph?

aphillips (commented):
Good start. A couple of minor comments.

spec/formatting.md
catamorphism: Removed the "may". What do you mean by "health warning"?

aphillips: By health warning, something like: It's a bad idea to create selectors or formatter functions with such side effects

catamorphism: I see. I'm not sure I would agree across the board. For example, you might want a `currentTime` function that returns the system clock time. This doesn't have "side effects" as such, but returns a different value when it's called at different times, so:

```
.local $time = {:currentTime}
{{ {$time} {$time} }}
```

could have a different result depending on whether the implementation evaluates the RHS of the variable `time` once or twice.

(The jargon is that `currentTime` is not referentially transparent.)

Perhaps some language saying that care should be taken when defining such functions and they should be avoided when possible?

eemeli: I thought we already had something in the spec about side effects, but the closest we come to is
> _Function handler_ access to the _formatting context_ MUST be minimal and read-only,
and execution time SHOULD be limited.

which relies on an implementation choosing to call "everything that a function can access" as its _formatting context_, because we've not explicitly prohibited access outside it.

That is to say, I agree with @aphillips that we ought to have stronger language to recommend against functions that mutate anything as a side effect.

catamorphism: I added some more language to (a) clarify that the problem also occurs with functions that don't mutate anything, but that depend on external mutable state; and (b) recommend against functions that mutate anything observably.

aphillips: Hmm...

side effect == touches the _resolved value_?

Lots of functions have "side effects" if that's the case. @eemeli your suggestion of limiting the _resolved value_ of `:date` and `:time` would be a side effect if we were to implement it. `currentTime` is a good example. `random` is another. Some operations, such as case-folding, alter the value in an irreversible way. Other operations, such as unfloating a time value, change the meaning.

Our design tries to be immutable, but we're actually mutable to the depth of one expression. Chaining `.local` or reannotating a placeholder can make these side-effects more visible.

Perhaps this paragraphs wants to be more like:

> Evaluating an _expression_ can affect the _resolved value_ associated with 
> a given _variable_.
> When combined with eager vs. lazy evaluation, 
> in which the same _expression_ might be evaluated multiple times,
> the side-effects of the _function_ on the _resolved value_
> might alter the result of formatting the _message_.

eemeli: > side effect == touches the _resolved value_?
>
> Lots of functions have "side effects" if that's the case. @eemeli your suggestion of limiting the _resolved value_ of `:date` and `:time` would be a side effect if we were to implement it.

Defining the _resolved value_ returned by a _function handler_ is not a [side effect](https://en.wikipedia.org/wiki/Side_effect_(computer_science)). Modifying the _operand_ would be a side effect. But a _function handler_ will quite rarely have a reason to return its input _operand_; in general, it will instead create a new _resolved value_ and return that.

> `currentTime` is a good example. `random` is another. Some operations, such as case-folding, alter the value in an irreversible way. Other operations, such as unfloating a time value, change the meaning.

A reasonable _function handler_ for any of these would not have any side effects, but would instead create a new _resolved value_ and return it.

> Our design tries to be immutable, but we're actually mutable to the depth of one expression. Chaining `.local` or reannotating a placeholder can make these side-effects more visible.

Being able to use a _resolved value_ as an _operand_ does not introduce mutability.

catamorphism: The phrase "side effect" is clearly confusing, so I rewrote the paragraph without using it. Please take another look!

spec/formatting.md
```diff
@@ -58,14 +58,19 @@ nor be made available to _function handlers_.
 > have already been evaluated in the order in which the relevant _declarations_
 > appear in the _message_.
 >
-> Users or implementations can provide functions that have observable side effects
-> or whose results depend on external mutable state (for example, a function
-> that returns the current time and date.)
+> In some environments, users or implementations can provide functions
+> that mutate state that is external to the message formatter
+> (for example, deleting a file in the filesystem). This is not recommended.
```
aphillips, line 63: "not recommended" turns out to be a 2119 keyword 😉 

Is this a likely example? Maybe a better example would be modifying the system or application locale (in some environments, setting the locale is not thread-safe)

```suggestion
(for example, setting or modifying the system locale).
```

macchiati:
I think a SHOULD NOT "provide functions that mutate" ... is warranted.

spec/formatting.md
aphillips: Actually... I think the problem here is the audience for your addition. This turns out not to be in a note block. It's in a normative "IMPORTANT" section. It should be written in a normative way or it should be in an informative NOTE inside the IMPORTANT block.

I'd actually do the former, replacing this paragraph with something like:

> Implementations and users creating custom functions SHOULD avoid 
> creating _function handlers_ that mutate external program state
> or which depend on external mutable state in a way that would
> result in different _resolved values_ for a _variable_ depending on 
> whether a given _expression_ is evaluated at most once (call-by-need)
> or multiple times (call-by-name).

The `:currentDateTime` is a good example of a function that implementation might very well want to provide and which produces a different result every time you call it (but at least a few ms, right?). Users of that function would need to know that they can't rely on it producing the exact same number each time they call it (the system clock is running) or that they can't rely on it to _not_ provide the same value every time in a given formatting call (the system clock is running, but the function only evaluates it once).

The warning about mutating program state doesn't seem to belong with eager/lazy? Can you think of an example of a real function that an implementation might provide (which isn't a security exploit)? I mean, my "set locale" suggestion above might quality in C, for example. That would certainly affect the formatting of the message (unless the "formatting context" was caching the locale). It would also be cause for spanking the developer who created/installed it.

macchiati:
+1

spec/formatting.md
catamorphism: I follow, but I'm a little unsure as to what you're suggesting the normative change should be. If it's "SHOULD avoid...", as in your suggested wording, doesn't that suggest there shouldn't be a `currentDateTime` or `random` function? But I can imagine these functions being useful, it's just that it may be surprising that different implementations with different evaluation strategies would handle them differently. Do we want to say to not provide functions like these, or just to be judicious about it?

catamorphism: I tried splitting the note into a normative part and an editorial "note" part, maybe that's closer to what you have in mind?

spec/formatting.md
```diff
@@ -57,6 +57,23 @@ nor be made available to _function handlers_.
...
+> Implementations and users creating custom functions SHOULD avoid
+> creating _function handlers_ that mutate external program state.
+>
+
+> [!NOTE]
+>
+> Function handlers can also be written that depend on external mutable state
+> (for example, the current date and time).
+> Lazy evaluation might involve evaluating the same _expression_ multiple times
+> (call-by-name) or evaluating every expression at most once (call-by-need).
```
gibson042, line 70: I really think it's a mistake to allow the same expression to ever evaluate to a different value. This is something that [CEL](https://github.com/google/cel-spec/blob/master/doc/langdef.md#evaluation) (to pick a roughly analogous technology) gets _very_ right.

spec/formatting.md
```diff
@@ -57,6 +57,23 @@ nor be made available to _function handlers_.
 > _declarations_ affecting _variables_ referenced by that _expression_
 > have already been evaluated in the order in which the relevant _declarations_
 > appear in the _message_.
+>
+> Implementations and users creating custom functions SHOULD avoid
+> creating _function handlers_ that mutate external program state.
```
aphillips, line 62: SHOULD avoid => SHOULD NOT

```suggestion
> Implementations and users SHOULD NOT create _function handlers_
> that mutate external program state,
> particularly since such a _function handler_ can present a remote execution hazard.
```

eemeli:
Should we include some language encouraging implementations to sandbox or otherwise restrict the access of function handlers to external state?

macchiati (approved):

catamorphism:
Per the 2024-10-21 call, I added stricter language saying that call-by-name implementations are not allowed because of the presence of impure functions. Please take another look!

eemeli (approved):

spec/formatting.md
```diff
@@ -57,6 +57,25 @@ nor be made available to _function handlers_.
 > _declarations_ affecting _variables_ referenced by that _expression_
 > have already been evaluated in the order in which the relevant _declarations_
 > appear in the _message_.
+> An implementation MUST ensure that every _expression_ in a _message_
+> is evaluated at most once.
+>
+> Implementations and users SHOULD NOT create _function handlers_
```
aphillips, line 63: This is a separate consideration. It should be a separate note.

```suggestion

> [!IMPORTANT]
> Implementations and users SHOULD NOT create _function handlers_
```

spec/formatting.md
```diff
@@ -57,6 +57,25 @@ nor be made available to _function handlers_.
...
+>
+> Function handlers can be written that depend on external mutable state
+> (for example, the current date and time).
+> Lazy evaluation might involve evaluating the same _expression_ multiple times
+> (call-by-name) or evaluating every expression at most once (call-by-need).
+> This justifies the requirement to ensure that every _expression_ is
+> evaluated at most once: different invocations of the same function handler
+> might return different results (functions are not pure).
+> This means that lazy implementations cannot use
+> a call-by-name evaluation strategy.
```
aphillips, line 78: I think this note is burying the lede. It should appear directly after the (first) "IMPORTANT", where we forbid multiple evaluations and I think it should start with the guidance. Something like:

> [!NOTE]
> Implementations that choose lazy evaluation cannot use a call-by-name
> evaluation strategy (instead of evaluating expressions at most once, or 
> "call-by-need").
> This is to prevent _function handlers_ that depend on external mutable state
> (for example, ones that obtain the current system clock time)
> from returning different values for the same _expression_ in different
> parts of a given _message_.

aphillips (commented):
Looks good, but some suggestions for organization and beautification.

spec/formatting.md
catamorphism: Took another pass at it in [fdbb48c](https://github.com/unicode-org/message-format-wg/pull/901/commits/fdbb48c57ca804e652902bb05951a98f64d29f6f) -- let me know what you think.

PR #903, opened 2024-10-10
Fix fallback value definition and use

eemeli:
This is a continuation for the changes of #728, defining a _fallback value_ explicitly as a _resolved value_, and rephrasing the parts of the spec that refer to resolution failures to instead test for whether a _resolved value_ is a _fallback value_.

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -377,7 +377,7 @@ For each _option_:
    1. Let `id` be the string value of the _identifier_ of the _option_.
    1. Let `rv` be the _resolved value_ of the _option_ value.
    1. If `rv` is a _fallback value_:
-      1. Emit a _Bad Option_ error.
+      1. If supported, emit a _Bad Option_ error.
    1. Else:
       1. Set `res[id]` to be `rv`.
 1. Return `res`.
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -336,7 +336,7 @@ Implementations MAY remove entries from `res` with unresolved values.
 
 The result of _option resolution_ MUST be a (possibly empty) mapping
 of string identifiers to values;
-that is, errors MAY be emitted, but such errors cannot be fatal.
+that is, errors MAY be emitted, but such errors MUST NOT be fatal.
 This mapping can be empty.
 
 ### Markup Resolution
@@ -406,6 +406,7 @@ The string representation of the _fallback value_ of an _expression_ depends on 
   > Examples:
   > In a context where `$var` fails to resolve, `{$var}` and `{$var :number}`
   > both resolve to the _fallback value_ `$var`.
+  > (even if `:number` fails to resolve).
 
 - _function_ _expression_ with no _operand_:
   U+003A COLON `:` followed by the _function_ _identifier_
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -334,8 +334,9 @@ determined as follows:
 
 Implementations MAY remove entries from `res` with unresolved values.
 
-Errors MAY be emitted during _option resolution_,
-but it always resolves to some mapping of string identifiers to values.
+The result of _option resolution_ MUST be a (possibly empty) mapping
+of string identifiers to values;
+that is, errors MAY be emitted, but such errors cannot be fatal.
 This mapping can be empty.
 
 ### Markup Resolution
@@ -353,7 +354,7 @@ The resolution of _markup_ MUST always succeed.
 ### Fallback Resolution
 
 A **_<dfn>fallback value</dfn>_** is the _resolved value_ for
-an _expression_ or _variable_ that fails to resolve.
+an _expression_ or _variable_ when that _expression_ or _variable_ fails to resolve.
 It has a string representation that is used for its formatting.
 
 The _resolved value_ of _text_, _literal_, and _markup_ MUST NOT be a _fallback value_.
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -259,6 +259,12 @@ The resolution of a _variable_ fails if no value is identified for its _name_.
 If this happens, an _Unresolved Variable_ error is emitted
 and a _fallback value_ is used as the _resolved value_ of the _variable_.
 
+If the _resolved value_ identified for the _variable_ _name_ is a _fallback value_,
+a _fallback value_ is used as the _resolved value_ of the _variable_.
+
+The _fallback value_ representation of a _variable_ has a string representation
+consisting of the U+0024 DOLLAR SIGN `$` followed by the _name_ of the _variable_.
+
 ### Function Resolution
 
 To resolve an _expression_ with a _function_,
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -211,7 +211,7 @@ whether its value was originally a _quoted literal_ or an _unquoted literal_.
 > For example,
 > the _option_ `foo=42` and the _option_ `foo=|42|` are treated as identical.
 
-> In a JavaScript formatter,
+> For example, in a JavaScript formatter,
 > the _resolved value_ of a _text_ or a _literal_ could have the following implementation:
 >
 > ```ts
@@ -237,7 +237,7 @@ and its value is looked up from the _formatting context_ _input mapping_.
 
 The resolution of a _variable_ fails if no value is identified for its _name_.
 If this happens, an _Unresolved Variable_ error is emitted
-and a _fallback value_ is used as _resolved value_ of the _variable_.
+and a _fallback value_ is used as the _resolved value_ of the _variable_.
 
 ### Function Resolution
 
@@ -246,13 +246,13 @@ the following steps are taken:
 
 1. If the _expression_ includes an _operand_, resolve its value.
    If this is a _fallback value_,
-   use it as the _resolved value_ of the _expression_.
+   return the _fallback value_ as the _resolved value_ of the _expression_.
 2. Resolve the _identifier_ of the _function_ and, based on the starting sigil,
    find the appropriate _function handler_ to call.
    If the implementation cannot find the _function handler_,
    or if the _identifier_ includes a _namespace_ that the implementation does not support,
    emit an _Unknown Function_ error
-   and use a _fallback value_ as the _resolved value_ of the _expression_.
+   and return a _fallback value_ as the _resolved value_ of the _expression_.
 
    Implementations are not required to implement _namespaces_ or installable
    _function registries_.
@@ -283,7 +283,7 @@ the following steps are taken:
    _operand_ did not match that expected by the _function_,
    the _function_ SHOULD cause a _Bad Operand_ error to be emitted.
   
-   In all failure cases, use a _fallback value_ as the _resolved value_ of the _expression_.
+   In all failure cases, return a _fallback value_ as the _resolved value_ of the _expression_.
 
 #### Function Handler
 
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -211,8 +211,7 @@ whether its value was originally a _quoted literal_ or an _unquoted literal_.
 > For example,
 > the _option_ `foo=42` and the _option_ `foo=|42|` are treated as identical.
 
-
-> For example, in a JavaScript formatter
+> In a JavaScript formatter,
 > the _resolved value_ of a _text_ or a _literal_ could have the following implementation:
 >
 > ```ts
@@ -237,9 +236,8 @@ Otherwise, the _variable_ is an implicit reference to an input value,
 and its value is looked up from the _formatting context_ _input mapping_.
 
 The resolution of a _variable_ fails if no value is identified for its _name_.
-If this happens, an _Unresolved Variable_ error is emitted.
-If a _variable_ would resolve to a _fallback value_,
-this MUST also be considered a failure.
+If this happens, an _Unresolved Variable_ error is emitted
+and a _fallback value_ is used as _resolved value_ of the _variable_.
 
 ### Function Resolution
 
@@ -247,13 +245,14 @@ To resolve an _expression_ with a _function_,
 the following steps are taken:
 
 1. If the _expression_ includes an _operand_, resolve its value.
-   If this fails, use a _fallback value_ for the _expression_.
+   If this is a _fallback value_,
+   use it as the _resolved value_ of the _expression_.
 2. Resolve the _identifier_ of the _function_ and, based on the starting sigil,
    find the appropriate _function handler_ to call.
    If the implementation cannot find the _function handler_,
    or if the _identifier_ includes a _namespace_ that the implementation does not support,
    emit an _Unknown Function_ error
-   and use a _fallback value_ for the _expression_.
+   and use a _fallback value_ as the _resolved value_ of the _expression_.
 
    Implementations are not required to implement _namespaces_ or installable
    _function registries_.
@@ -284,7 +283,7 @@ the following steps are taken:
    _operand_ did not match that expected by the _function_,
    the _function_ SHOULD cause a _Bad Operand_ error to be emitted.
   
-   In all failure cases, use the _fallback value_ for the _expression_ as its _resolved value_.
+   In all failure cases, use a _fallback value_ as the _resolved value_ of the _expression_.
 
 #### Function Handler
 
@@ -320,16 +319,20 @@ Implementation-defined _functions_ SHOULD use an implementation-defined _namespa
 
 #### Option Resolution
 
-The result of resolving _option_ values is an unordered mapping of string identifiers to values.
+The result of resolving _option_ values is an unordered mapping of string identifiers to values,
+determined as follows:
 
-For each _option_:
+1. Let `res` be a new empty mapping.
+1. For each _option_:
+   1. Let `id` be the string value of the _identifier_ of the _option_.
+   1. Let `rv` be the _resolved value_ of the _option_ value.
+   1. If `rv` is a _fallback value_:
+      1. Set the value bound to `id` in `res` to an unresolved value.
+   1. Else:
+      1. Set the value bound to `id` in `res` to `rv`.
+1. Return `res`.
 
-- Resolve the _identifier_ of the _option_.
-- If the _option_'s right-hand side successfully resolves to a value,
-   bind the _identifier_ of the _option_ to the _resolved value_ in the mapping.
-- Otherwise, bind the _identifier_ of the _option_ to an unresolved value in the mapping.
-   Implementations MAY later remove this value before calling the _function_.
-   (Note that an _Unresolved Variable_ error will have been emitted.)
+Implementations MAY remove entries from `res` with unresolved values.
 
 Errors MAY be emitted during _option resolution_,
 but it always resolves to some mapping of string identifiers to values.
@@ -349,17 +352,26 @@ The resolution of _markup_ MUST always succeed.
 
 ### Fallback Resolution
 
-A **_<dfn>fallback value</dfn>_** is the _resolved value_ for an _expression_ that fails to resolve.
+A **_<dfn>fallback value</dfn>_** is the _resolved value_ for
+an _expression_ or _variable_ that fails to resolve.
+It has a string representation that is used for its formatting.
+
+The _resolved value_ of _text_, _literal_, and _markup_ MUST NOT be a _fallback value_.
+
+A _variable_ fails to resolve when no value is identified for its _name_.
+The string representation of its _fallback value_ is
+U+0024 DOLLAR SIGN `$` followed by the _name_ of the _variable_.
 
 An _expression_ fails to resolve when:
 
-- A _variable_ used as an _operand_ (with or without a _function_) fails to resolve.
-  * Note that this does not include a _variable_ used as an _option_ value.
+- A _variable_ used as its _operand_ resolves to a _fallback value_.
+  Note that an _expression_ does not necessarily fail to resolve
+  if an _option_ resolves with a _fallback value_.
 - A _function_ fails to resolve.
 
-The _fallback value_ depends on the contents of the _expression_:
+The string representation of the _fallback value_ of an _expression_ depends on its contents:
 
-- _expression_ with a _literal_ _operand_ (either quoted or unquoted)
+- _expression_ with a _literal_ _operand_ (either quoted or unquoted):
   U+007C VERTICAL LINE `|`
   followed by the value of the _literal_
   with escaping applied to U+005C REVERSE SOLIDUS `\` and U+007C VERTICAL LINE `|`,
@@ -370,33 +382,29 @@ The _fallback value_ depends on the contents of the _expression_:
   > `{42 :func}` resolves to the _fallback value_ `|42|` and
   > `{|C:\\| :func}` resolves to the _fallback value_ `|C:\\|`.
 
-- _expression_ with _variable_ _operand_ referring to a local _declaration_ (with or without a _function_):
-  the _value_ to which it resolves (which may already be a _fallback value_)
+- _expression_ with a _variable_ _operand_ referring to a _declaration_:
+  the string representation of the _fallback value_ of the _expression_ of that _declaration_.
 
   > Examples:
   > In a context where `:func` fails to resolve,
-  > the _pattern_'s _expression_ in `.local $var={|val|} {{{$var :func}}}`
-  > resolves to the _fallback value_ `|val|` and the message formats to `{|val|}`.
-  > In a context where `:now` fails to resolve but `:datetime` does not,
-  > the _pattern_'s _expression_ in
+  > the _placeholder_ in `.local $var = {|val| :func} {{{$var}}}`
+  > resolves to the _fallback value_ `|val|`.
+  >
+  > In a context where `:pretty` fails to resolve but `:now` does not,
+  > the _placeholder_ in
   > ```
   > .local $t = {:now format=iso8601}
-  > .local $pretty_t = {$t :datetime}
-  > {{{$pretty_t}}}
+  > {{{$t :pretty}}}
   > ```
-  > (transitively) resolves to the _fallback value_ `:now` and
-  > the message formats to `{:now}`.
+  > resolves to the _fallback value_ `:now`.
 
-- _expression_ with _variable_ _operand_ not referring to a local _declaration_ (with or without a _function_):
+- _expression_ with _variable_ _operand_ not referring to a _declaration_:
+  the _fallback value_ representation of that _variable_,
   U+0024 DOLLAR SIGN `$` followed by the _name_ of the _variable_
 
   > Examples:
   > In a context where `$var` fails to resolve, `{$var}` and `{$var :number}`
   > both resolve to the _fallback value_ `$var`.
-  > In a context where `:func` fails to resolve,
-  > the _pattern_'s _expression_ in `.input $arg {{{$arg :func}}}`
-  > resolves to the _fallback value_ `$arg` and
-  > the message formats to `{$arg}`.
 
 - _function_ _expression_ with no _operand_:
   U+003A COLON `:` followed by the _function_ _identifier_
@@ -409,7 +417,7 @@ The _fallback value_ depends on the contents of the _expression_:
 
   This is not currently used by any expression, but may apply in future revisions.
 
-_Option_ _identifiers_ and values are not included in the _fallback value_.
+_Options_ and _attributes_ are not included in the _fallback value_.
 
 _Pattern selection_ is not supported for _fallback values_.
 
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -266,8 +266,9 @@ the following steps are taken:
 
 1. If the _expression_ includes an _operand_, resolve its value.
    If this is a _fallback value_,
-   return the _fallback value_ as the _resolved value_ of the _expression_.
-2. Resolve the _identifier_ of the _function_ and, based on the starting sigil,
+   return a _fallback value_ as the _resolved value_ of the _expression_.
+
+2. Resolve the _identifier_ of the _function_ and
    find the appropriate _function handler_ to call.
    If the implementation cannot find the _function handler_,
    or if the _identifier_ includes a _namespace_ that the implementation does not support,
@@ -292,7 +293,7 @@ the following steps are taken:
    supported by the implementation, process them as specified.
    Such `u:` options MAY be removed from the resolved mapping of _options_.
 
-5. Call the function implementation with the following arguments:
+5. Call the _function handler_ with the following arguments:
 
    - The _function context_.
    - The resolved mapping of _options_.
@@ -418,7 +419,8 @@ An _expression_ fails to resolve when:
 - A _variable_ used as its _operand_ resolves to a _fallback value_.
   Note that an _expression_ does not necessarily fail to resolve
   if an _option_ resolves with a _fallback value_.
-- A _function_ fails to resolve.
+- No _function handler_ is found for a _function_ _identifier_.
+- Calling a _function handler_ fails or does not return a valid value.
 
 The string representation of the _fallback value_ of an _expression_ depends on its contents:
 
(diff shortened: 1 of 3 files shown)


spec/formatting.md
```diff
@@ -211,8 +211,7 @@ whether its value was originally a _quoted literal_ or an _unquoted literal_.
 > For example,
 > the _option_ `foo=42` and the _option_ `foo=|42|` are treated as identical.
 
-
-> For example, in a JavaScript formatter
+> In a JavaScript formatter,
```
aphillips, line 214: I would put the "for example" back.

spec/formatting.md
```diff
@@ -284,7 +283,7 @@ the following steps are taken:
    _operand_ did not match that expected by the _function_,
    the _function_ SHOULD cause a _Bad Operand_ error to be emitted.
   
-   In all failure cases, use the _fallback value_ for the _expression_ as its _resolved value_.
+   In all failure cases, use a _fallback value_ as the _resolved value_ of the _expression_.
```
aphillips, line 286: I was going to comment on line 255 to the effect that "a _fallback value_" might want to be "the _fallback value_". Saying "the" is specific, even though the exact formulation of the fallback value is not (line 248 is different).

spec/formatting.md
```diff
@@ -320,16 +319,20 @@ Implementation-defined _functions_ SHOULD use an implementation-defined _namespa
 
 #### Option Resolution
 
-The result of resolving _option_ values is an unordered mapping of string identifiers to values.
+The result of resolving _option_ values is an unordered mapping of string identifiers to values,
```
aphillips, line 322: Is "unordered" too specific? Most developers know not to make an ordered map unless required to, but we do not need to require it be unordered (and I'm not sure how we'd test that it is unordered).

I'd also suggest tightening the definition:
```suggestion
Resolving _options_ results in a mapping of string _identifiers_ to _values_,
```

spec/formatting.md
```diff
@@ -320,16 +319,20 @@ Implementation-defined _functions_ SHOULD use an implementation-defined _namespa
...
+determined as follows:
 
-For each _option_:
+1. Let `res` be a new empty mapping.
+1. For each _option_:
+   1. Let `id` be the string value of the _identifier_ of the _option_.
```
aphillips, line 327: Mention NFC here?

Should we check if `id` exists (which is an error)?

Should there be a step for initializing `res` with options found in the operand's resolved value (enabling option transitivity)?

spec/formatting.md
```diff
@@ -320,16 +319,20 @@ Implementation-defined _functions_ SHOULD use an implementation-defined _namespa
...
+   1. Let `rv` be the _resolved value_ of the _option_ value.
+   1. If `rv` is a _fallback value_:
+      1. Set the value bound to `id` in `res` to an unresolved value.
```
aphillips, line 330: Is "unresolved value" defined somewhere? You're trying to avoid saying `null` right?

spec/formatting.md
```diff
@@ -320,16 +319,20 @@ Implementation-defined _functions_ SHOULD use an implementation-defined _namespa
...
+   1. Else:
+      1. Set the value bound to `id` in `res` to `rv`.
```
aphillips, line 332: See also line 330

```suggestion
      1. Bind the value `rv` to the ID `id` in `res`.
```

spec/formatting.md
```diff
@@ -320,16 +319,20 @@ Implementation-defined _functions_ SHOULD use an implementation-defined _namespa
...
-- Resolve the _identifier_ of the _option_.
-- If the _option_'s right-hand side successfully resolves to a value,
-   bind the _identifier_ of the _option_ to the _resolved value_ in the mapping.
-- Otherwise, bind the _identifier_ of the _option_ to an unresolved value in the mapping.
-   Implementations MAY later remove this value before calling the _function_.
-   (Note that an _Unresolved Variable_ error will have been emitted.)
+Implementations MAY remove entries from `res` with unresolved values.
 
 Errors MAY be emitted during _option resolution_,
 but it always resolves to some mapping of string identifiers to values.
```
aphillips, line 338: This is phrased wrong. It wants a MUST in it

```suggestion
The result of _option resolution_ MUST be a (possibly empty) mapping
of string identifiers to values;
that is, errors MAY be emitted, but such errors cannot be fatal.
```

spec/formatting.md
```diff
@@ -349,17 +352,26 @@ The resolution of _markup_ MUST always succeed.
 
 ### Fallback Resolution
 
-A **_<dfn>fallback value</dfn>_** is the _resolved value_ for an _expression_ that fails to resolve.
+A **_<dfn>fallback value</dfn>_** is the _resolved value_ for
+an _expression_ or _variable_ that fails to resolve.
```
aphillips, line 356: Although this is pretty clear, it could be read as the _resolved value_ being the thing that fails to resolve.

```suggestion
an _expression_ or _variable_ when that _expression_ or _variable_ fails to resolve.
```

spec/formatting.md
eemeli: On 248 the "the" is because we're re-using the operand's resolved value.

On 255 and 286 we need "a" because we need to create the value.

Is that ok, or is there a change here that you think ought to be made?

spec/formatting.md
eemeli: We agreed specifically about "unordered" in #751, though.

spec/formatting.md
eemeli: > Mention NFC here?

We should do that change separately, but I agree with it (as mentioned in https://github.com/unicode-org/message-format-wg/pull/885#pullrequestreview-2310712001).

> Should we check if `id` exists (which is an error)?

No, because that's a data model error and checked well before we get here.

> Should there be a step for initializing `res` with options found in the operand's resolved value (enabling option transitivity)?

No, that's something that a function handler needs to do internally. It needs to be able to separate the options that the operand carries and the options that are on the expression.

spec/formatting.md
eemeli: It's not defined, and yes, this ought to be represented by something like `undefined` or `None` or a null pointer or something similar.

spec/formatting.md
aphillips: I agree that 248 is different.

I guess the change would be necessary if we think that there will be a specific fallback value for every expression (that is, specific to that expression)--which I think is true. Using "a" implies you can use any old fallback value.......

But it's not that important.

spec/formatting.md
aphillips: You're right: we did discuss that. At that risk of discussing it again, I'd rather say it separately, e.g.

> Resolving _options_ results in a mapping of string _identifiers_ to _values.
> Note that this mapping MAY be unordered.

spec/formatting.md
```diff
@@ -237,23 +236,23 @@ Otherwise, the _variable_ is an implicit reference to an input value,
 and its value is looked up from the _formatting context_ _input mapping_.
 
 The resolution of a _variable_ fails if no value is identified for its _name_.
-If this happens, an _Unresolved Variable_ error is emitted.
-If a _variable_ would resolve to a _fallback value_,
-this MUST also be considered a failure.
+If this happens, an _Unresolved Variable_ error is emitted
+and a _fallback value_ is used as _resolved value_ of the _variable_.
```
catamorphism, line 240:
```suggestion
and a _fallback value_ is used as the _resolved value_ of the _variable_.
```

spec/formatting.md
```diff
@@ -237,23 +236,23 @@ Otherwise, the _variable_ is an implicit reference to an input value,
...
 
 ### Function Resolution
 
 To resolve an _expression_ with a _function_,
 the following steps are taken:
 
 1. If the _expression_ includes an _operand_, resolve its value.
-   If this fails, use a _fallback value_ for the _expression_.
+   If this is a _fallback value_,
+   use it as the _resolved value_ of the _expression_.
```
catamorphism, line 249: You might want to say "and omit step 2" here?

spec/formatting.md
aphillips: Would this step result in unbinding `id`'s value if it already existed? Put another way, is there value to adding `id` to the mapping if the value is blank? Some environments might not tolerate null values as well. We might permit binding the unresolved value or not binding at all.

catamorphism: I agree on avoiding "unresolved value". I would rather say something like "ensure no value is bound to `id` in `res`", because there are situations where there's no equivalent of null pointers/`undefined`. For example, in C++, the values in the map might not be pointers at all, or if they are, then binding a name to null might not be any different from not adding it to the map.

spec/formatting.md
aphillips: > >     Should we check if id exists (which is an error)?
> No, because that's a data model error and checked well before we get here.

This feels like something that should be mentioned here?

> >    Should there be a step for initializing res with options found in the operand's resolved value (enabling option transitivity)?
> No, that's something that a function handler needs to do internally. It needs to be able to separate the options that the operand carries and the options that are on the expression.

This also feels like something that should be mentioned. This is a case where the expression might overwrite an option, btw.

Perhaps after the algorithm:

> [!NOTE]
> Implementations are required to check for duplicate _options_ 
> during parsing or when the _message_ is created from a data model representation.

> [!NOTE]
> _Options_ (and their values) that are part of the _resolved value_ of an _operand_
> might be added to the _resolved options_ by the _function handler_.

spec/formatting.md
aphillips: For example:

```
.input {$foo :function option1=a option2=b}
.local $b = {$foo :function option1=$undeclaredVar}
{{{$b}}}
```

In think in this case what our spec has been trying to do is: the failure to resolve `$undeclaredVar` would result in `$b` having options `{option2: "b"}`, with `option1` not present (it's not `a` and it's not present-but-null). Either way you get a Bad Option error (if you want it). If `option1` were required, the _function handler_ would barf, but so to would it barf on a null value.

@eemeli are you thinking that you want to represent the "bad option" state into the data model?

spec/formatting.md
aphillips: Or... "return the _fallback value_ as the _resolved value_ of the _expression_" (fast fail)

spec/formatting.md
macchiati: > Note that this mapping MAY be unordered.

I don't think this is testable, not without extrapolation. I think it is cleaner to state this directly in terms of observable differences, something like:

The order among options MUST NOT produce different formatting results. For example,

`.input {$x option1=foo1 option2=foo2}`
must yield exactly the same resulting formats as:
`.input {$x option2=foo2 option1=foo1}`

spec/formatting.md
eemeli: The term was originally introduced from a suggestion by @gibson042 in https://github.com/unicode-org/message-format-wg/pull/524#discussion_r1408097009, and subsequently discussed on the [2023-12-04 call](https://github.com/unicode-org/message-format-wg/blob/main/meetings/2023/notes-2023-12-04.md#should-we-remove-unresolved-options-from-the-map-545) (based on my recollection, but the notes don't cover that) and then amended in #545.

I would prefer not re-opening the discussion about what we're doing in general, but I'd be very open for suggestions about some better language to use here instead of "unreseolved value".

Note that if we require a _fallback value_ to be able to show up as a resolved _option_ value in _function handler_ arguments, then that'll introduce a novel requirement on implementations, as the formatting behaviour otherwise does not give them any public visibility.

spec/formatting.md
eemeli:
```suggestion
> For example, in a JavaScript formatter,
```

spec/formatting.md
```diff
@@ -237,23 +236,23 @@ Otherwise, the _variable_ is an implicit reference to an input value,
...
 2. Resolve the _identifier_ of the _function_ and, based on the starting sigil,
    find the appropriate _function handler_ to call.
    If the implementation cannot find the _function handler_,
    or if the _identifier_ includes a _namespace_ that the implementation does not support,
    emit an _Unknown Function_ error
-   and use a _fallback value_ for the _expression_.
+   and use a _fallback value_ as the _resolved value_ of the _expression_.
```
eemeli, line 255:
```suggestion
   and return a _fallback value_ as the _resolved value_ of the _expression_.
```

spec/formatting.md
eemeli: The fallback value is specific to the _expression_ with respect to the current _formatting context_. I think it's best to leave it as "a", but modify this to "return" instead of "use".

```suggestion
   In all failure cases, return a _fallback value_ as the _resolved value_ of the _expression_.
```

spec/formatting.md
eemeli: That sounds like a requirement to introduce separately in the Function Handler section.

spec/formatting.md
eemeli: > > > Should we check if id exists (which is an error)?
> > 
> > No, because that's a data model error and checked well before we get here.
> 
> This feels like something that should be mentioned here?

We have discussed this previously, e.g. in https://github.com/unicode-org/message-format-wg/pull/577#discussion_r1439577540, and I continue to disagree that we need a specific callout here about this data model error.

> > > Should there be a step for initializing res with options found in the operand's resolved value (enabling option transitivity)?
> >
> > No, that's something that a function handler needs to do internally. It needs to be able to separate the options that the operand carries and the options that are on the expression.
> 
> This also feels like something that should be mentioned. This is a case where the expression might overwrite an option, btw.

This section is specifically about resolving the _option_ values on the current _expression_. I'm a bit concerned that adding a note about something that's ultimately internal to _function handlers_ seems like it would add confusion.

spec/formatting.md
eemeli: Is there an existing expression for this sort of record/mapping/object property value assignment that we could crib from another spec? I'd prefer not inventing new language, and I'm drawing a blank for this.

spec/formatting.md
gibson042: The purpose of "unresolved value" as originally suggested at https://github.com/unicode-org/message-format-wg/pull/524#discussion_r1408097009 is "a representation of failure to resolve [that is distinct from every possible resolved value]". If you're looking for a reasonable replacement, I would expect something like "a value uniquely representing failure to resolve" or "the _resolution failure value_" (with a corresponding definition).

gibson042: But that detail is very important IMO: it **must not** be possible for input to explicitly specify a value that the implementation would produce upon encountering an undeclared variable or similar, so that function implementations are guaranteed the ability to detect options suffering from such a state.

spec/formatting.md
gibson042: My best suggestion for a template would be the WHATWG Infra [Map](https://infra.spec.whatwg.org/#ordered-map).

spec/formatting.md
```diff
@@ -320,19 +319,24 @@ Implementation-defined _functions_ SHOULD use an implementation-defined _namespa
...
-- Otherwise, bind the _identifier_ of the _option_ to an unresolved value in the mapping.
-   Implementations MAY later remove this value before calling the _function_.
-   (Note that an _Unresolved Variable_ error will have been emitted.)
+Implementations MAY remove entries from `res` with unresolved values.
 
-Errors MAY be emitted during _option resolution_,
-but it always resolves to some mapping of string identifiers to values.
+The result of _option resolution_ MUST be a (possibly empty) mapping
+of string identifiers to values;
+that is, errors MAY be emitted, but such errors cannot be fatal.
```
gibson042, line 339:
```suggestion
that is, errors MAY be emitted, but such errors MUST NOT be fatal.
```

spec/formatting.md
```diff
@@ -370,33 +383,29 @@ The _fallback value_ depends on the contents of the _expression_:
...
+  > the _placeholder_ in
   > ```
   > .local $t = {:now format=iso8601}
-  > .local $pretty_t = {$t :datetime}
-  > {{{$pretty_t}}}
+  > {{{$t :pretty}}}
   > ```
-  > (transitively) resolves to the _fallback value_ `:now` and
-  > the message formats to `{:now}`.
+  > resolves to the _fallback value_ `:now`.
```
gibson042, line 400: This is a behaviorally distinct change that I hope is not intentional... the current example is attempting to demonstrate _transitive_ fallback such that an a variable defined as output from a defined function that uses as its operand a different variable defined as output from a base function that fails to resolve is the fallback value from that base resolution failure:
```
.local $t = {:now format=iso8601}
.local $pretty_t = {$t :datetime}
{{{$pretty_t}}}
```
formatting as `{:now}` to indicate the broken function reference _through_ the non-broken function expression, whereas the new scenario
```
.local $t = {:now format=iso8601}
{{{$t :pretty}}}
```
in which `:now` resolves but `:pretty` does not fail to capture those details, and AFAIK should format as the non-fallback output from `{:now format=iso8601}`.

At the root of it, I don't see why the fallback value of an expression with a declaration-referencing operand should be the _fallback_ value of that declaration rather than its _resolved_ value. So looking again at
```
.local $t = {:now format=iso8601}
{{{$t :pretty}}}
```
only if _`:now`_ failed to resolve would I expect the result to format as `{:now}`, because otherwise we have a better value for downstream fallbacks—the resolved value of `{:now format=iso8601}`.

spec/formatting.md
```diff
@@ -370,33 +383,29 @@ The _fallback value_ depends on the contents of the _expression_:
...
+  the _fallback value_ representation of that _variable_,
   U+0024 DOLLAR SIGN `$` followed by the _name_ of the _variable_
 
   > Examples:
   > In a context where `$var` fails to resolve, `{$var}` and `{$var :number}`
   > both resolve to the _fallback value_ `$var`.
-  > In a context where `:func` fails to resolve,
-  > the _pattern_'s _expression_ in `.input $arg {{{$arg :func}}}`
-  > resolves to the _fallback value_ `$arg` and
-  > the message formats to `{$arg}`.
```
gibson042, line 399: This example seems valuable to me; why is it being removed?

gibson042: I guess it could be rolled into the previous:
```suggestion
  > Examples:
  > In a context where `$var` fails to resolve, `{$var}` and `{$var :number}`
  > both resolve to the _fallback value_ `$var` (even if `:number` fails to
  > resolve).
```

spec/formatting.md
aphillips: We currently have this requirement (as you pointed out), but it is dimly and unclearly specified here. We already have a consensus for it, so, given where we are in the schedule, I would suggest:

- Add the text @machiatti suggests to Function Handler
- Add the first sentence of my text from the ["you're right" comment](https://github.com/unicode-org/message-format-wg/pull/903#discussion_r1796065115)
- Add a reminding note in place of my "MAY" text

```suggestion
Resolving _options_ results in a mapping of string _identifiers_ to values,
```

And later:

> [!NOTE]
> The order of the _options_ is not significant.

spec/formatting.md
aphillips: You're adding an algorithm here that was previously just some prose. Algorithms are also more fussy, since you have to specify the corner cases and such. A naive implementer would read this and just pick up the options specified in the expression--missing the fact that the operand might be bringing some to the party--especially because the first step is _Let `res` be a new empty mapping_ and does not contain a step "import the options from the resolved value (if any)"

I didn't want to make this more complicated than it is, because it's fine. But why not add the note about function handler behavior?

We don't need to include the other note if you feel strongly about it.

spec/formatting.md
aphillips: Infra Map is fine, but... ordered and not applicable in all programming environments.

Maybe just ignore my comment. My real comment was that "set the value bound to" is awkward.

spec/formatting.md
gibson042: > Finally, I'll note that "explicitly specified value that cannot be passed by the caller" is tricky, particularly in low-level language implementations. It suggests that the option values are boxed in some way so that the handler can check if it is an unresolved value (unresolved being disjoint from all data representations)

Absolutely, and my preference remains what it was in my original comment for #545—don't even let implementations see options whose value fails to resolve (_I accepted @stasm's argument about error messages to move things along, but I still think it would be better to rely only on the preceding Unresolved Variable error and present to functions only options that successfully resolve_).

But assuming options with unresolved values _are_ presented to functions, I don't consider the disjointness requirement to be negotiable.

spec/formatting.md
eemeli: I support Addison's suggestion, if done within a follow-on PR. This PR is only adding a comma to the end of this sentence, and the proposed action is introducing a new MUST NOT on function handlers, which are so far only governed by MAY and SHOULD text.

OTOH, if this PR is not merged and accepted on this Monday's call, then sure, let's allow this scope creep and see how far this one PR goes. For example, I note that we say that function handlers SHOULD emit a Bad Operand error, but we don't say the same about Bad Option.

spec/formatting.md
eemeli: How about this, combined with dropping the bit below about removing unresolved values?
```suggestion
      1. Emit a _Bad Option_ error.
```

spec/formatting.md
eemeli: > My real comment was that "set the value bound to" is awkward.

That's the problem I'm also seeking to solve. Would this work? We don't use `[]` access in the other algorithms, but I would presume that it'd be pretty clear.
```suggestion
      1. Set `res[id]` to be `rv`.
```

spec/formatting.md
eemeli: When reviewing #539, I had not noticed that this behaviour changed. Before that PR, we had this included in the spec: https://github.com/unicode-org/message-format-wg/blob/770cb66a29ee41ae8117cfeba732a266b569c915/spec/formatting.md?plain=1#L284-L288
(note that this was from before introducing input/local, so "local declaration" here meant all declarations)

The language proposed here would indeed revert to the old behaviour, where a failure in a later function would end up using the _fallback value_ of the earliest declaration that its operand is built from, as opposed to the _resolved value_ of its most recent declaration.

The rationalisation for this behaviour is presented in https://github.com/unicode-org/message-format-wg/pull/539#discussion_r1413036522: when any failure happens in the resolution of an _expression_, we should not trust that the resolved value of its operand is by default serialisable. For example, in:
```
.local $user = {$user_id :get-user}
{{Add {$user :first-nam} as friend}}
```
If `:get-user` succeeds but `:first-nam` is a typo that's not caught before we try to format the message, there is a risk that the default serialisation of the `$user` value would reveal information that is not intended to be public. Therefore, the safe formatting result here must be `Add {$user_id} as friend`.

If reverting to this behaviour is not acceptable, then we ought to add a requirement on all _resolved values_ to provide some safe serialization, in case their values are unintentionally included in the formatted result.

spec/formatting.md
```diff
@@ -370,33 +383,29 @@ The _fallback value_ depends on the contents of the _expression_:
...
+  > resolves to the _fallback value_ `:now`.
 
-- _expression_ with _variable_ _operand_ not referring to a local _declaration_ (with or without a _function_):
+- _expression_ with _variable_ _operand_ not referring to a _declaration_:
+  the _fallback value_ representation of that _variable_,
   U+0024 DOLLAR SIGN `$` followed by the _name_ of the _variable_
 
   > Examples:
   > In a context where `$var` fails to resolve, `{$var}` and `{$var :number}`
   > both resolve to the _fallback value_ `$var`.
```
eemeli, line 408:
```suggestion
  > both resolve to the _fallback value_ `$var`.
  > (even if `:number` fails to resolve).
```

spec/formatting.md
```diff
@@ -370,33 +383,30 @@ The _fallback value_ depends on the contents of the _expression_:
...
```
eemeli, line 408:
```suggestion
  > both resolve to the _fallback value_ `$var`
```

spec/formatting.md
eemeli: Added the note about function handler behavior.

spec/formatting.md
gibson042: That's a very reasonable concern that I find compelling. And (assuming agreement from others) the spec should be **very** clear with both text and examples about the consequences—that resolution failures propagate through operands, locking fallback values to the earliest such point in the causal chain regardless of success/failure of downstream functions.

aphillips: I don't get it? I mean, the default serialization of `$user` might reveal inner data of the user object. but the fallback value doesn't serialize that object. It just reveals the _name_ of that object is `$user`. The only time we expose a serialization is when the operand is a literal, in which case the value is already exposed in the message. So there is no danger of `$user`'s content being exposed, only its name.

I guess I'm also wondering why we walk back up the declaration chain to begin with. The example in the spec is:
```
.local $t = {:now format=iso8601}
.local $pretty_t = {$t :datetime}
{{{$pretty_t}}}
```
Shouldn't the fallback value be "at the point of failure"? What's wrong with `{$pretty_t}` coming out? Or `{$t}` if `:datetime` failed? Emiting `:now` doesn't tell you where the failure is and might point at the wrong element in the chain (imagine that `:datetime` were misspelled in the example)

spec/formatting.md
```diff
@@ -320,21 +319,32 @@ Implementation-defined _functions_ SHOULD use an implementation-defined _namespa
...
+      1. Set the value bound to `id` in `res` to `rv`.
+1. Return `res`.
 
-- Resolve the _identifier_ of the _option_.
-- If the _option_'s right-hand side successfully resolves to a value,
-   bind the _identifier_ of the _option_ to the _resolved value_ in the mapping.
-- Otherwise, bind the _identifier_ of the _option_ to an unresolved value in the mapping.
-   Implementations MAY later remove this value before calling the _function_.
-   (Note that an _Unresolved Variable_ error will have been emitted.)
+Implementations MAY remove entries from `res` with unresolved values.
```
eemeli, line 335:
```suggestion
```

spec/formatting.md
```diff
@@ -320,21 +319,32 @@ Implementation-defined _functions_ SHOULD use an implementation-defined _namespa
...
```
eemeli, line 336:
```suggestion
```

spec/formatting.md
```diff
@@ -320,21 +319,30 @@ Implementation-defined _functions_ SHOULD use an implementation-defined _namespa
...
+The result of _option resolution_ MUST be a (possibly empty) mapping
+of string identifiers to values;
+that is, errors MAY be emitted, but such errors MUST NOT be fatal.
 This mapping can be empty.
 
+> [!NOTE]
+> The _resolved value_ of a _function_ _operand_
+> can also include resolved option values.
+> These are not included in the _option resolution_ result,
+> and need to processed separately by a _function handler_.
```
catamorphism, line 344:
```suggestion
> and need to be processed separately by a _function handler_.
```

spec/formatting.md
```diff
@@ -320,21 +319,30 @@ Implementation-defined _functions_ SHOULD use an implementation-defined _namespa
...
+      1. Emit a _Bad Option_ error.
+   1. Else:
+      1. Set `res[id]` to be `rv`.
+1. Return `res`.
 
-Errors MAY be emitted during _option resolution_,
-but it always resolves to some mapping of string identifiers to values.
+The result of _option resolution_ MUST be a (possibly empty) mapping
+of string identifiers to values;
+that is, errors MAY be emitted, but such errors MUST NOT be fatal.
```
catamorphism, line 337: This is the first mention I've noticed in the spec of a non-fatal error, i.e. a warning. If _Bad Option_ is sometimes a fatal error and sometimes a warning, it seems like the spec needs some more language to clarify what implementations should do with it. Or if it's meant to always be a warning, maybe https://github.com/unicode-org/message-format-wg/blob/main/spec/errors.md should say something about that?

catamorphism: On second thought: from the perspective of an implementation can only signal errors or return usable output, but not both, I'm reading this as saying that the implementation can't signal a _Bad Option_ error. Which is fine, since it's a "MAY".

spec/formatting.md
```diff
@@ -259,6 +259,12 @@ The resolution of a _variable_ fails if no value is identified for its _name_.
 If this happens, an _Unresolved Variable_ error is emitted
 and a _fallback value_ is used as the _resolved value_ of the _variable_.
 
+If the _resolved value_ identified for the _variable_ _name_ is a _fallback value_,
+a _fallback value_ is used as the _resolved value_ of the _variable_.
```
macchiati, line 263: A general issue: The _resolved value_ is logically {datatype, optionMap}. So whenever we have a statement like this, that implies that a _fallback value_ is a _resolved value_. That means we **must** say what the optionMap is: I presume an empty map, but we need to be clear.

macchiati (changes requested):

spec/formatting.md
eemeli: Fallbacking is now set to use the fallback value of the last placeholder, rather than looking up a fallback value via its operand.

spec/formatting.md
eemeli: Good point. I've updated the text to make the references more explicit. PTAL?

macchiati: Excellent fixes to all the fallback value text, thanks!

macchiati (approved):

spec/formatting.md
aphillips: This is a good insight. It means that certain classes of errors will be silent in ICU4C. FWIW, ICU4J probably has similar issues. You can't `throw` _Bad Option_ without being fatal. Logging the error is about the best one can do.

spec/formatting.md
aphillips: This sentence reads as a tautology. Perhaps:

```suggestion
If the _resolved value_ associated with a _variable_ _name_ is a _fallback value_,
then the _resolved value_ of the _variable_ is that _fallback value_.
```

spec/formatting.md
```diff
@@ -257,23 +256,30 @@ Otherwise, the _variable_ is an implicit reference to an input value,
...
 ### Function Resolution
 
 To resolve an _expression_ with a _function_,
 the following steps are taken:
 
 1. If the _expression_ includes an _operand_, resolve its value.
-   If this fails, use a _fallback value_ for the _expression_.
-2. Resolve the _identifier_ of the _function_ and, based on the starting sigil,
+   If this is a _fallback value_,
+   return a _fallback value_ as the _resolved value_ of the _expression_.
```
aphillips, line 275:
```suggestion
   return that _fallback value_ as the _resolved value_ of the _expression_.
```

spec/formatting.md
```diff
@@ -366,17 +372,27 @@ The order of _options_ MUST NOT be significant.
...
-   bind the _identifier_ of the _option_ to the _resolved value_ in the mapping.
-- Otherwise, bind the _identifier_ of the _option_ to an unresolved value in the mapping.
-   Implementations MAY later remove this value before calling the _function_.
-   (Note that an _Unresolved Variable_ error will have been emitted.)
+1. Let `res` be a new empty mapping.
+1. For each _option_:
+   1. Let `id` be the string value of the _identifier_ of the _option_.
+   1. Let `rv` be the _resolved value_ of the _option_ value.
+   1. If `rv` is a _fallback value_:
+      1. Emit a _Bad Option_ error.
```
aphillips, line 380: This smells like a fatal, even though (see @catamorphism's comments below) we later say "MAY". Perhaps:

```suggestion
      1. If supported, emit a _Bad Option_ error.
```

spec/formatting.md
```diff
@@ -395,67 +411,73 @@ The resolution of _markup_ MUST always succeed.
...
 An _expression_ fails to resolve when:
 
-- A _variable_ used as an _operand_ (with or without a _function_) fails to resolve.
-  * Note that this does not include a _variable_ used as an _option_ value.
-- A _function_ fails to resolve.
+- A _variable_ used as its _operand_ resolves to a _fallback value_.
+  Note that an _expression_ does not necessarily fail to resolve
+  if an _option_ resolves with a _fallback value_.
+- No _function handler_ is found for a _function_ _identifier_.
+- Calling a _function handler_ fails or does not return a valid value.
```
aphillips, line 431: Nitpicky editing. 

```suggestion
- The _function handler_ fails, returns an error, or returns an invalid value.
```

aphillips (commented):
Nitpicks, Englische-type comments, some observations.

spec/formatting.md
eemeli: That's not quite right, though, because the fallback value needs to be different. Consider here:
```
.local $x = {42 :func-fails}
.local $y = {$x}
{{{$x}, {$y}}}
```
The above should format as "{$x}, {$y}", and that requires for the `$y` fallback value to be different from the `$x` fallback value.

spec/formatting.md
eemeli: See corresponding discussion above for variable resolution.

spec/formatting.md
eemeli: Would an error not count as an invalid value?

aphillips: Maybe? It's unclear. It could be "validly an error" 😃 

This comment isn't that important. I'm mostly trying to harmonize phrasing here.

eemeli: I would much rather not re-introduce the phrase "returns an error", as that's rather opinionated about the implementation's processing model. We also should not use the more generic "emits an error", as it's possible for a function handler to emit an Unsupported Operation error and still return a valid value.

aphillips: Fair enough.

aphillips (approved):

aphillips:
@catamorphism I'm ready to merge this, but checking with you before I do, since you previously had a comment.

aphillips:
Seeing no reply, I'm merging this. File issues as needed.

PR #904, opened 2024-10-10
Add tests for changes due to #885 (name/literal equality)

### LINKED ISSUE #900: Add tests for changes due to #885 (name/literal equality)
Add tests that test for correct implementation of this step in the spec:

"Let `ks` be the resolved value of `key` in Unicode Normalization Form C"

https://github.com/unicode-org/message-format-wg/pull/885

catamorphism:
Closes #900

### DIFF
--- a/test/tests/functions/string.json
+++ b/test/tests/functions/string.json
@@ -61,7 +61,12 @@
       "expErrors": [{"type": "duplicate-variant"}]
     },
     {
-      "description": "NFC: keys are normalized",
+      "description": "NFC: :string normalizes the comparison value (un-normalized selector, normalized key)",
+      "src": ".local $x = {\u1E0A\u0323 :string} .match $x \u1E0C\u0307 {{Right}} * {{Wrong}}",
+      "exp": "Right"
+    },
+    {
+      "description": "NFC: keys are normalized (normalized selector, un-normalized key)",
       "src": ".local $x = {\u1E0C\u0307 :string} .match $x \u1E0A\u0323 {{Right}} * {{Wrong}}",
       "exp": "Right"
     }
--- a/test/tests/functions/string.json
+++ b/test/tests/functions/string.json
@@ -44,6 +44,16 @@
           "type": "unresolved-variable"
         }
       ]
+    },
+    {
+      "description": "NFC: keys are normalized",
+      "src": ".local $x = {\u1E0A\u0323 :string} .match $x \u1E0A\u0323 {{Not normalized}} \u1E0C\u0307 {{Normalized}} * {{Wrong}}",
+      "expErrors": [{"type": "duplicate-variant"}]
+    },
+    {
+      "description": "NFC: keys are normalized",
+      "src": ".local $x = {\u1E0A\u0323 :string} .match $x \u1E0C\u0307 {{Right}} * {{Wrong}}",
+      "exp": "Right"
     }
   ]
 }
--- a/test/tests/functions/string.json
+++ b/test/tests/functions/string.json
@@ -46,8 +46,18 @@
       ]
     },
     {
-      "description": "NFC: keys are normalized",
+      "description": "NFC: keys are normalized (unquoted)",
       "src": ".local $x = {\u1E0A\u0323 :string} .match $x \u1E0A\u0323 {{Not normalized}} \u1E0C\u0307 {{Normalized}} * {{Wrong}}",
+      "expErrors": [{"type": "duplicate-variant"}]
+    },
+    {
+      "description": "NFC: keys are normalized (quoted)",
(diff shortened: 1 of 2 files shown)


test/tests/functions/string.json
```diff
@@ -44,6 +44,16 @@
           "type": "unresolved-variable"
         }
       ]
+    },
+    {
+      "description": "NFC: keys are normalized",
+      "src": ".local $x = {\u1E0A\u0323 :string} .match $x \u1E0A\u0323 {{Not normalized}} \u1E0C\u0307 {{Normalized}} * {{Wrong}}",
+      "expErrors": [{"type": "duplicate-variant"}]
+    },
```
eemeli, line 52: We should also validate that quoted keys are compared to unquoted keys according to their resolved value.

```suggestion
    {
      "description": "NFC: keys are normalized (unquoted)",
      "src": ".local $x = {\u1E0A\u0323 :string} .match $x \u1E0A\u0323 {{Not normalized}} \u1E0C\u0307 {{Normalized}} * {{Wrong}}",
      "expErrors": [{"type": "duplicate-variant"}]
    },
    {
      "description": "NFC: keys are normalized (quoted)",
      "src": ".local $x = {\u1E0A\u0323 :string} .match $x |\u1E0A\u0323| {{Not normalized}} |\u1E0C\u0307| {{Normalized}} * {{Wrong}}",
      "expErrors": [{"type": "duplicate-variant"}]
    },
    {
      "description": "NFC: keys are normalized (mixed)",
      "src": ".local $x = {\u1E0A\u0323 :string} .match $x \u1E0A\u0323 {{Not normalized}} |\u1E0C\u0307| {{Normalized}} * {{Wrong}}",
      "expErrors": [{"type": "duplicate-variant"}]
    },
```

test/tests/syntax.json
```diff
@@ -697,6 +697,33 @@
     {
       "src": "{{trailing whitespace}} \n",
       "exp": "trailing whitespace"
+    },
+    {
+      "description": "NFC: literals are not normalized",
```
eemeli, line 702:
```suggestion
      "description": "NFC: text is not normalized",
```

test/tests/functions/string.json
```diff
@@ -44,6 +44,16 @@
...
+    {
+      "description": "NFC: keys are normalized",
+      "src": ".local $x = {\u1E0A\u0323 :string} .match $x \u1E0C\u0307 {{Right}} * {{Wrong}}",
+      "exp": "Right"
     }
```
eemeli, line 57: We don't actually require this (matching keys to a non-normalized compare value). We should, but we don't. I think we need to update `:string` selection to also normalize the comparison value. Especially as the reverse works:
```suggestion
    {
      "description": "NFC: keys are normalized",
      "src": ".local $x = {\u1E0C\u0307 :string} .match $x \u1E0A\u0323 {{Right}} * {{Wrong}}",
      "exp": "Right"
    }
```

eemeli (changes requested):

test/tests/syntax.json
```diff
@@ -697,6 +697,33 @@
...
+    },
+    {
+      "description": "NFC: variables are compared to each other as-if normalized; decl is normalized, use isn't",
+      "src": ".input {$\u1E0c\u0307} {{{$\u0044\u0323\u0307}}}",
+      "params": [{"name": "\u1E0c\u0307", "value": "foo"}],
+      "exp": "foo"
+    },
+    {
+      "description": "NFC: variables are compared to each other as-if normalized; decl is non-normalized, use is",
+      "src": ".input {$\u0044\u0323\u0307} {{{$\u1E0c\u0307}}}",
```
aphillips, line 724: You're missing the reordering case (`\u0044\u0307\u0323`)

You might include a special case mapping. See [Charmod example](https://www.w3.org/TR/charmod-norm/#example-15), featuring `U+01FA`==`U+00C5 U+0301`==`U+212B U+0301`==`U+0041 U+030A U+0301`. (212B is the Angstrom sign)

catamorphism: Added these two tests in [80b73b7](https://github.com/unicode-org/message-format-wg/pull/904/commits/80b73b71dd30809709737a847bc9bea8037713b8)

eemeli (approved):

test/tests/functions/string.json
catamorphism: Added in [6bcd592](https://github.com/unicode-org/message-format-wg/pull/904/commits/6bcd5922e660244e2ce574d89375f99e8b15e0a6)

eemeli (approved):

aphillips (approved):

PR #906, opened 2024-10-11
Allow surrogates in content, issue #895

mihnita:
The goal was to allow surrogates in localizable content, but not in names or anything else that is "code".

This change reflects that agreement.

### DIFF
--- a/spec/appendices.md
+++ b/spec/appendices.md
@@ -23,11 +23,21 @@ that abnormally affect the display of the _message_
 when viewed as source code, or in resource formats or translation tools,
 but do not generate errors from MessageFormat parsers or processing APIs.
 
-The localizable elements of a message (text and string literals) allow the presence of
-unpaired surrogates (U+D800 to U+DFFF). This is for compatibility with existing formats
-that are agnostic about them. \
-But their presence of unpaired surrogates is likely an indication of mistakes or bad tooling.
-Their use is not recommended, and linting (if present) can be used to prevent them.
+> [!IMPORTANT]
+> _Text_ and _literals_ allow unpaired surrogate code points
+> (`U+D800` to `U+DFFF`).
+> This is for compatibility with formats or data structures 
+> that use the UTF-16 encoding 
+> and do not check for unpaired surrogates.
+> (Strings in Java or JavaScript are examples of this.)
+> These code points SHOULD NOT be used in a _message_.
+> Unpaired surrogate code points are likely an indication of mistakes
+> or errors in the creation, serialization, or processing of the _message_.
+> Many processes will convert them to 
+> &#xfffd; U+FFFD REPLACEMENT CHARACTER
+> during processing or display.
+> Implementations not based on UTF-16 might not be able to represent
+> a _message_ containing such code points.
 
 Bidirectional text containing right-to-left characters (such as used for Arabic or Hebrew) 
 also poses a potential source of confusion for users. 
--- a/spec/appendices.md
+++ b/spec/appendices.md
@@ -14,16 +14,20 @@ host environments, their serializations and resource formats,
 that might be sufficient to prevent most problems.
 However, MessageFormat itself does not supply such a restriction.
 
-MessageFormat _messages_ permit nearly all Unicode code points,
-with the exception of surrogates, 
+MessageFormat _messages_ permit nearly all Unicode code points
 to appear in _literals_, including the text portions of a _pattern_.
 This means that it can be possible for a _message_ to contain invisible characters
-(such as bidirectional controls, 
-ASCII control characters in the range U+0000 to U+001F,
+(such as bidirectional controls, ASCII control characters in the range U+0000 to U+001F,
 or characters that might be interpreted as escapes or syntax in the host format)
 that abnormally affect the display of the _message_
 when viewed as source code, or in resource formats or translation tools,
 but do not generate errors from MessageFormat parsers or processing APIs.
+
+The localizable elements of a message (text and string literals) allow the presence of
+unpaired surrogates (U+D800 to U+DFFF). This is for compatibility with existing formats
+that are agnostic about them. \
+But their presence of unpaired surrogates is likely an indication of mistakes or bad tooling.
+Their use is not recommended, and linting (if present) can be used to prevent them.
 
 Bidirectional text containing right-to-left characters (such as used for Arabic or Hebrew) 
 also poses a potential source of confusion for users. 
--- a/spec/appendices.md
+++ b/spec/appendices.md
@@ -23,22 +23,6 @@ that abnormally affect the display of the _message_
 when viewed as source code, or in resource formats or translation tools,
 but do not generate errors from MessageFormat parsers or processing APIs.
 
-> [!IMPORTANT]
-> _Text_ and _quoted literals_ allow unpaired surrogate code points
-> (`U+D800` to `U+DFFF`).
-> This is for compatibility with formats or data structures 
-> that use the UTF-16 encoding 
-> and do not check for unpaired surrogates.
-> (Strings in Java or JavaScript are examples of this.)
-> These code points SHOULD NOT be used in a _message_.
-> Unpaired surrogate code points are likely an indication of mistakes
-> or errors in the creation, serialization, or processing of the _message_.
-> Many processes will convert them to 
-> &#xfffd; U+FFFD REPLACEMENT CHARACTER
-> during processing or display.
-> Implementations not based on UTF-16 might not be able to represent
-> a _message_ containing such code points.
-
 Bidirectional text containing right-to-left characters (such as used for Arabic or Hebrew) 
 also poses a potential source of confusion for users. 
 Since MessageFormat 2.0's syntax makes use of 
--- a/spec/appendices.md
+++ b/spec/appendices.md
@@ -24,7 +24,7 @@ when viewed as source code, or in resource formats or translation tools,
 but do not generate errors from MessageFormat parsers or processing APIs.
 
 > [!IMPORTANT]
-> _Text_ and _literals_ allow unpaired surrogate code points
+> _Text_ and _quoted literals_ allow unpaired surrogate code points
 > (`U+D800` to `U+DFFF`).
 > This is for compatibility with formats or data structures 
 > that use the UTF-16 encoding 
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -115,6 +115,22 @@ A **_<dfn>local variable</dfn>_** is a _variable_ created as the result of a _lo
 > so that these do not need to be escaped in the body of a _message_.
 
 > [!NOTE]
+> _Text_ and _quoted literals_ allow unpaired surrogate code points
+> (`U+D800` to `U+DFFF`).
+> This is for compatibility with formats or data structures 
+> that use the UTF-16 encoding 
+> and do not check for unpaired surrogates.
+> (Strings in Java or JavaScript are examples of this.)
+> These code points SHOULD NOT be used in a _message_.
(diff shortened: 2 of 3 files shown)


mihnita:
Touching only `text-char` means that potentially localizable content does no allow surrogates.
For example:
```
... {img src=... alt=|This is localizable text|...}...
```

That would be a quoted literal.

Allowing surrogates in the quoted literal might also allow it in a couple of other places:
```
... {|allow it here| :string} ...
```
and
```
.match
|also here| {{ ... }}
* {{ ... }}
```

That is again non-code, and potentially localizable.

---

To help understanding what is going on I created a graph (click it for full size).
It represents the mf2 abnf as it is today, with colors added to the interesting nodes and all links to `ws`, `o`, `s`, `bidi` to reduce clutter.

mihnita:
> This would allow surrogates in both text and literals

I know. It is intentional, and I explained why.

aphillips:
I think this has to be discussed again if we propose to include `literal`. While I agree that the literal examples are valid, I think that makes my proposed solution more attractive (forbid surrogates in the grammar but permit implementations to not check for them (except in identifiers). Unpaired surrogates have no valid use and, as noted elsewhere, the first UTF-8 coder will eliminate any found in a message. 

Note that this PR is incomplete, since the syntax spec also has to be updated.

eemeli:
My understanding of the consensus reached during the 2024-10-07 call was that we'd be allowing unpaired surrogates in _text_ only.

If allowing surrogates in _text_ is not enough and we need to consider also allowing them in _quoted literal_ content, I think we need to see some examples of such use to consider their utility. In _text_ I can imagine them showing up from processes that are doing dumb concatenation or string slicing, but in _literals_ we're talking about content that needs to be processed within a framework that has a pretty good understanding of MF2, and content that is further processed within MF2 rather than only in a consumer of the formatted output of MF2.

I would also strongly prefer to see such examples or other rationalisation for unpaired surrogates in _quoted literals_ presented here, before we spend more meeting time on this topic.

My own base assumption is that any unpaired surrogate is an indicator of some broken localization process, and that silently ignoring them would be hiding an error.

mihnita:
> allowing unpaired surrogates in text only.

Literals are text. Sometimes localizable.

macchiati:
My thoughts are that we should allow unpaired surrogates in the text in *input
parameters*, but I don't think we need to allow them in the message itself
(literals, etc).

That is, in parsing a message format, we could throw an exception if an
unpaired surrogate were encountered, but the formatting software can accept
unpaired surrogates in data passed in, and treat them as if they were
reserved, or convert to U+FFFD "replacement character".

aphillips:
(chair hat)

In the 2024-10-14 call we agree to allow unpaired surrogates in text and literals, including changing mentions of characters to mentions of code points in the spec as needed and including a health warning.

mihnita:
Updated the non-abnf part of the spec.

In most cases there was no need to change code point to code units.

From the [Unicode spec, Chapter 3, section "3.8 Surrogates"](https://www.unicode.org/versions/Unicode16.0.0/core-spec/chapter-3/#G2630):
> _D71 High-surrogate code point:_ A Unicode code point in the range U+D800 to U+DBFF. \
> _D72 High-surrogate code unit:_ A 16-bit code unit in the range D80016 to DBFF16, used in UTF-16 as the leading code unit of a surrogate pair. \
> _D73 Low-surrogate code point:_ A Unicode code point in the range U+DC00 to U+DFFF. \
> _D74 Low-surrogate code unit:_ A 16-bit code unit in the range DC0016 to DFFF16, used in UTF-16 as the trailing code unit of a surrogate pair.

Just because in localizable elements we allow code points in the surrogates range it does not mean that all of our processing and spec are all of a sudden written in terms of code units.

I think that the changes I made are sufficient. \
I also included wording that unpaired surrogates are _"likely an indication of mistakes or bad tooling"_ and a recommendation the linting can be use to detect and prevent them.

spec/appendices.md
```diff
@@ -14,17 +14,21 @@ host environments, their serializations and resource formats,
...
 or characters that might be interpreted as escapes or syntax in the host format)
 that abnormally affect the display of the _message_
 when viewed as source code, or in resource formats or translation tools,
 but do not generate errors from MessageFormat parsers or processing APIs.
 
+The localizable elements of a message (text and string literals) allow the presence of
+unpaired surrogates (U+D800 to U+DFFF). This is for compatibility with existing formats
+that are agnostic about them. \
+But their presence of unpaired surrogates is likely an indication of mistakes or bad tooling.
+Their use is not recommended, and linting (if present) can be used to prevent them.
```
aphillips, line 30: Some grammar and typo issues. In addition, linkify terms, make into a note, and fix 2119 keywords.

```suggestion
> [!IMPORTANT]
> _Text_ and _literals_ allow unpaired surrogate code points
> (`U+D800` to `U+DFFF`).
> This is for compatibility with formats or data structures 
> that use the UTF-16 encoding 
> and do not check for unpaired surrogates.
> (Strings in Java or JavaScript are examples of this.)
> These code points SHOULD NOT be used in a _message_.
> Unpaired surrogate code points are likely an indication of mistakes
> or errors in the creation, serialization, or processing of the _message_.
> Many processes will convert them to 
> &#xfffd; U+FFFD REPLACEMENT CHARACTER
> during processing or display.
> Implementations not based on UTF-16 might not be able to represent
> a _message_ containing such code points.
```

spec/syntax.md
```diff
@@ -60,7 +60,8 @@ The syntax specification takes into account the following design restrictions:
    control characters such as U+0000 NULL and U+0009 TAB, permanently reserved noncharacters
    (U+FDD0 through U+FDEF and U+<i>n</i>FFFE and U+<i>n</i>FFFF where <i>n</i> is 0x0 through 0x10),
    private-use code points (U+E000 through U+F8FF, U+F0000 through U+FFFFD, and
-   U+100000 through U+10FFFD), unassigned code points, and other potentially confusing content.
+   U+100000 through U+10FFFD), unassigned code points, unpaired surrogates in messages and literals
+   only (U+D800 through U+DFFF), and other potentially confusing content.
```
aphillips, line 64: This is fine, but I think it would be okay to omit it too.

spec/syntax.md
```diff
@@ -274,8 +275,9 @@ A _quoted pattern_ MAY be empty.
 ### Text
 
 **_<dfn>text</dfn>_** is the translateable content of a _pattern_.
-Any Unicode code point is allowed, except for U+0000 NULL
-and the surrogate code points U+D800 through U+DFFF inclusive.
+Any Unicode code point is allowed, except for U+0000 NULL.
+Unpaired surrogates code points (U+D800 through U+DFFF inclusive) are allowed
+in localizable elements, but using them is likely a mistake and not recommended.
```
aphillips, line 280: I don't think introducing the phrase "localizable elements" is helpful here. The _text_ production **_is_** that. I think, instead, we might include a note reprising what was mentioned above:

```suggestion
> [!NOTE]
> Unpaired surrogate code points (`U+D800` through `U+DFFF` inclusive)
> are allowed for compatibility with UTF-16 based implementations
> that do not check for this encoding error.
```

spec/syntax.md
```diff
@@ -691,8 +693,7 @@ A _literal_ can appear
 as a _key_ value,
 as the _operand_ of a _literal-expression_,
 or in the value of an _option_.
-A _literal_ MAY include any Unicode code point
-except for U+0000 NULL or the surrogate code points U+D800 through U+DFFF.
+A _literal_ MAY include any Unicode code point except for U+0000 NULL.
```
aphillips, line 696: If we add the note for _text_ we should add it for _literal_ too?

spec/appendices.md
aphillips: I... don't think this is a security consideration? Maybe move this note to the section on _message_, just after the note that begins:

> This syntax is designed to be embeddable into many different programming languages...

aphillips (commented):
Thank you @mihnita for getting this done before our next call!!!!

I like the changes generally, but am making some suggestions below.

spec/syntax.md
mihnita: To be fair, I don't really see the benefit of all this duplication in syntax.md.
The abnf should be enough.
It only makes it difficult to keep them in sync.
I just found that `syntax.md` contains an exact copy of the abnf section I am changing.

Probably too late to refactor this though...

mihnita: I added the same note to the quoted literals.

And changed in a couple of places this is affects quoted literals, not all literals.
(see graph in the main comments section)

spec/appendices.md
```diff
@@ -14,17 +14,31 @@ host environments, their serializations and resource formats,
...
-ASCII control characters in the range U+0000 to U+001F,
+(such as bidirectional controls, ASCII control characters in the range U+0000 to U+001F,
 or characters that might be interpreted as escapes or syntax in the host format)
 that abnormally affect the display of the _message_
 when viewed as source code, or in resource formats or translation tools,
 but do not generate errors from MessageFormat parsers or processing APIs.
 
+> [!IMPORTANT]
+> _Text_ and _quoted literals_ allow unpaired surrogate code points
+> (`U+D800` to `U+DFFF`).
```
eemeli, line 28: How/why is this important to note as a security consideration?

spec/syntax.md
```diff
@@ -60,7 +60,8 @@ The syntax specification takes into account the following design restrictions:
    control characters such as U+0000 NULL and U+0009 TAB, permanently reserved noncharacters
    (U+FDD0 through U+FDEF and U+<i>n</i>FFFE and U+<i>n</i>FFFF where <i>n</i> is 0x0 through 0x10),
    private-use code points (U+E000 through U+F8FF, U+F0000 through U+FFFFD, and
-   U+100000 through U+10FFFD), unassigned code points, and other potentially confusing content.
+   U+100000 through U+10FFFD), unassigned code points, unpaired surrogates in messages and
+   quoted literals only (U+D800 through U+DFFF), and other potentially confusing content.
```
eemeli, line 64: The specifier about where surrogates are valid is unnecessary esp. in this context, where the other "potentially confusing content" is for the most part only allowed in the same.
```suggestion
   U+100000 through U+10FFFD), unassigned code points,
   unpaired surrogates (U+D800 through U+DFFF), and other potentially confusing content.
```

spec/syntax.md
```diff
@@ -274,8 +275,11 @@ A _quoted pattern_ MAY be empty.
 ### Text
 
 **_<dfn>text</dfn>_** is the translateable content of a _pattern_.
-Any Unicode code point is allowed, except for U+0000 NULL
-and the surrogate code points U+D800 through U+DFFF inclusive.
+Any Unicode code point is allowed, except for U+0000 NULL.
+> [!NOTE]
+> Unpaired surrogate code points (`U+D800` through `U+DFFF` inclusive)
+> are allowed for compatibility with UTF-16 based implementations
+> that do not check for this encoding error.
```
eemeli, line 282: This note would be better placed a bit further down in this section, and it needs to be separated from the surrounding content by an empty line. This would also be a more appropriate place for the concerns that are currently hidden away in the Security Considerations appendix.

eemeli (changes requested):

spec/appendices.md
aphillips: I previously suggested the same thing:

> I... don't think this is a security consideration? Maybe move this note to the section on message, just after the note that begins:
>
>>    This syntax is designed to be embeddable into many different programming languages...

I think that location is a good one because we put a whole wodge of notes there that are "read once and forget" about the content of messages.

aphillips (approved):

aphillips:
@mihnita Per today's call, ready to merge?

PR #909, opened 2024-10-19
Drop data model request for feedback on "name"

eemeli:
This note was added in #668, and we've discussed it on and off since then. My sense is that the data model is easier to work with and less error-prone if the namespace is kept as a part of the name.


aphillips:
We'll need to thoroughly review all of these special notes in 46.1.

aphillips:
Merged in the 2024-10-21 call. This is a non-normative change. We can still get feedback if anyone wants to provide it.

PR #911, opened 2024-10-19
Define locale options for `:datetime`, `:date` & `:time`

eemeli:
Adds a new "basket" of **_date/time locale options_** that override defaults set in the current locale or an implementation-defined date/time _operand_ value. These are optional, and can be made available on each of `:datetime`, `:date` & `:time`.

The `hourCycle` option is dropped from the `:datetime` _field options_ (where it never belonged), and its functionality is replaced by the `hour12=true|false`_date/time locale option_.

This PR explicitly does not address the concerns raised in #866; that is a separate discussion.

### DIFF
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -598,6 +598,9 @@ it can include _style options_, _field options_, or other option values.
 These are included in the resolved option values of the _expression_,
 with _options_ on the _expression_ taking priority over any option values of the _operand_.
 
+In addition to _style options_ and _field options_,
+an implementation MAY also support _date/time locale options_ for `:datetime`.
+
 > [!NOTE]
 > The names of _options_ and their _values_ were derived from the
 > [options](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/resolvedOptions#description)
@@ -671,11 +674,6 @@ The function `:datetime` has the following options:
   - `1`
   - `2`
   - `3`
-- `hourCycle` (default is locale-specific)
-  - `h11`
-  - `h12`
-  - `h23`
-  - `h24`
 - `timeZoneName`
   - `long`
   - `short`
@@ -683,20 +681,6 @@ The function `:datetime` has the following options:
   - `longOffset`
   - `shortGeneric`
   - `longGeneric`
-
-> [!NOTE]
-> The following options do not have default values because they are only to be used
-> as overrides for locale-and-value dependent implementation-defined defaults.
-
-The following date/time options are **not** part of the default registry.
-Implementations SHOULD avoid creating options that conflict with these, but
-are encouraged to track development of these options during Tech Preview:
-- `calendar` (default is locale-specific)
-  - valid [Unicode Calendar Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeCalendarIdentifier)
-- `numberingSystem` (default is locale-specific)
-   - valid [Unicode Number System Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeNumberSystemIdentifier)
-- `timeZone` (default is system default time zone or UTC)
-  - valid identifier per [BCP175](https://www.rfc-editor.org/rfc/rfc6557)
 
 #### Composition
 
@@ -728,6 +712,7 @@ The function `:date` has these _options_:
   - `long`
   - `medium` (default)
   - `short`
+- _date/time locale options_
 
 If the _operand_ of the _expression_ is an implementation-defined date/time type,
 it can include other option values.
@@ -764,6 +749,7 @@ The function `:time` has these _options_:
   - `long`
   - `medium`
   - `short` (default)
+- _date/time locale options_
 
 If the _operand_ of the _expression_ is an implementation-defined date/time type,
 it can include other option values.
@@ -826,3 +812,24 @@ For more information, see [Working with Timezones](https://w3c.github.io/timezon
 > The form of these serializations is known and is a de facto standard.
 > Support for these extensions is expected to be required in the post-tech preview.
 > See: https://datatracker.ietf.org/doc/draft-ietf-sedate-datetime-extended/
+
+### Date and Time Locale Options
+
+Implementations MAY support overriding some of the options set by the current locale
+or an implementation-defined date/time _operand_ value.
+To do so, the following **_<dfn>date/time locale options</dfn>_**
+can be made available on `:datetime`, `:date`, and `:time`:
+
+- `calendar`
+  - valid [Unicode Calendar Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeCalendarIdentifier)
+- `numberingSystem`
+   - valid [Unicode Number System Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeNumberSystemIdentifier)
+- `hour12`
+  - `true`
+  - `false`
+- `timeZone`
+  - valid identifier per [BCP175](https://www.rfc-editor.org/rfc/rfc6557)
+
+> [!NOTE]
+> These options do not have default values because they are only to be used
+> as overrides for locale-and-value dependent implementation-defined defaults.
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -1073,17 +1073,21 @@ override values set by the current locale,
 or provided by the _formatting context_ (such as the default time zone),
 or embedded in an implementation-defined date/time _operand_ value.
 
-The following options and their values are required to be available on
-the functions `:datetime`, `:date`, and `:time`:
+The following **standard** option and its values MUST be available on
+the functions `:datetime` and `:time`:
 
 - `hour12`
   - `true`
   - `false`
-- `calendar` \[OPTIONAL\]
+
+The following **optional** options and their values SHOULD be available on
+the functions `:datetime`, `:date`, and `:time`:
+
+- `calendar`
   - valid [Unicode Calendar Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeCalendarIdentifier)
-- `numberingSystem` \[OPTIONAL\]
+- `numberingSystem`
   - valid [Unicode Number System Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeNumberSystemIdentifier)
-- `timeZone` \[OPTIONAL\]
+- `timeZone`
   - valid identifier per [BCP175](https://www.rfc-editor.org/rfc/rfc6557)
 
 > [!NOTE]
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -595,7 +595,7 @@ All other _operand_ values produce a _Bad Operand_ error.
 The `:datetime` function can use either the appropriate _style options_ 
 or can use a collection of _field options_ (but not both) to control the formatted 
 output.
-_Date/time locale options_ can be combined with either _style options_ or _field options_.
+_Date/time override options_ can be combined with either _style options_ or _field options_.
 
 If both _style options_ and _field options_ are specified,
 a _Bad Option_ error is emitted
@@ -716,7 +716,7 @@ The function `:date` has these _options_:
   - `long`
   - `medium` (default)
   - `short`
-- _Date/time locale options_
+- _Date/time override options_
 
 If the _operand_ of the _expression_ is an implementation-defined date/time type,
 it can include other option values.
@@ -754,7 +754,7 @@ The function `:time` has these _options_:
   - `long`
   - `medium`
   - `short` (default)
-- _Date/time locale options_
+- _Date/time override options_
 
 If the _operand_ of the _expression_ is an implementation-defined date/time type,
 it can include other option values.
@@ -819,10 +819,12 @@ For more information, see [Working with Timezones](https://w3c.github.io/timezon
 > Support for these extensions is expected to be required in the post-tech preview.
 > See: https://datatracker.ietf.org/doc/draft-ietf-sedate-datetime-extended/
 
-### Date and Time Locale Options
+### Date and Time Override Options
 
-**_<dfn>Date/time locale options</dfn>_** override options set by the current locale
-or an implementation-defined date/time _operand_ value.
+**_<dfn>Date/time override options</dfn>_** are _options_ that allow an _expression_ to
+override values set by the current locale,
+or provided by the _formatting context_ (such as the default time zone),
+or embedded in an implementation-defined date/time _operand_ value.
 
 The following options and their values are required to be available on
 the functions `:datetime`, `:date`, and `:time`:
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -589,17 +589,16 @@ All other _operand_ values produce a _Bad Operand_ error.
 The `:datetime` function can use either the appropriate _style options_ 
 or can use a collection of _field options_ (but not both) to control the formatted 
 output.
+_Date/time locale options_ can be combined with either _style options_ or _field options_.
 
-If both are specified, a _Bad Option_ error MUST be emitted
+If both _style options_ and _field options_ are specified,
+a _Bad Option_ error is emitted
 and a _fallback value_ used as the _resolved value_ of the _expression_.
 
 If the _operand_ of the _expression_ is an implementation-defined date/time type,
 it can include _style options_, _field options_, or other option values.
 These are included in the resolved option values of the _expression_,
 with _options_ on the _expression_ taking priority over any option values of the _operand_.
-
-In addition to _style options_ and _field options_,
-an implementation MAY also support _date/time locale options_ for `:datetime`.
 
 > [!NOTE]
 > The names of _options_ and their _values_ were derived from the
@@ -632,7 +631,7 @@ and what format to use for that field.
 The _field options_ are defined as follows:
 
 > [!IMPORTANT]
-> The value `2-digit` for some _field options_ **must** be quoted
+> The value `2-digit` for some _field options_ MUST be quoted
 > in the MessageFormat syntax because it starts with a digit
 > but does not match the `number-literal` production in the ABNF.
 > ```
@@ -712,7 +711,7 @@ The function `:date` has these _options_:
   - `long`
   - `medium` (default)
   - `short`
-- _date/time locale options_
+- _Date/time locale options_
 
 If the _operand_ of the _expression_ is an implementation-defined date/time type,
 it can include other option values.
@@ -749,7 +748,7 @@ The function `:time` has these _options_:
   - `long`
   - `medium`
   - `short` (default)
-- _date/time locale options_
+- _Date/time locale options_
 
 If the _operand_ of the _expression_ is an implementation-defined date/time type,
 it can include other option values.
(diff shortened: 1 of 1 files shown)


spec/registry.md
```diff
@@ -826,3 +812,24 @@ For more information, see [Working with Timezones](https://w3c.github.io/timezon
...
+
+- `calendar`
+  - valid [Unicode Calendar Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeCalendarIdentifier)
+- `numberingSystem`
+   - valid [Unicode Number System Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeNumberSystemIdentifier)
+- `hour12`
+  - `true`
+  - `false`
+- `timeZone`
+  - valid identifier per [BCP175](https://www.rfc-editor.org/rfc/rfc6557)
```
aphillips, line 831: Don't link the RFC. Link the BCP (e.g. https://www.rfc-editor.org/bcp/bcp175)... but... time zone identifiers have a ton of quirks in them. Also, we almost certainly want to allow offset time zones (e.g. `GMT-01:23`) and we may want to allow special sauce like metazones. CLDR has a bunch of stuff about this, but I'm too busy this morning to look up the precise reference. It's somewhere near [here](https://www.unicode.org/reports/tr35/tr35-dates.html#Time_Zone_Names)

aphillips (commented):
I like some of the thought behind this, but want to think about it more.

spec/registry.md
eemeli: Happy to go with whatever; this reference was not changed from the earlier one we already included.

spec/registry.md
```diff
@@ -826,3 +812,24 @@ For more information, see [Working with Timezones](https://w3c.github.io/timezon
 > The form of these serializations is known and is a de facto standard.
 > Support for these extensions is expected to be required in the post-tech preview.
 > See: https://datatracker.ietf.org/doc/draft-ietf-sedate-datetime-extended/
+
+### Date and Time Locale Options
```
aphillips, line 816: Having thought about it, I think this is the wrong approach. The `MAY` means that these are optional options. We should just put them into the list for each function as **optional**

Doing so would solve the other problem I have, which is that `timeZone` is most definitely not part of locale (even if the `-u` extension provides a mechanism for transmitting the zone).

spec/registry.md
aphillips, line 831: I think we need to do a bit more here. The ability to float and unfloat time values is actually rather important, even if developers don't always understand it. See: [Working with Time and Timezones](https://w3c.github.io/timezone/#floating-times)

```suggestion
- `timeZone` **optional** (no default)
  - valid identifier per [BCP175](https://www.rfc-editor.org/bcp/bcp175)
  - `none` (removes the time zone, if present, to create a local/floating time)
```

aphillips (changes requested):
I think a refactor to "optional" is better than having them separate.

eemeli:
Updated as per discussion during this week's meeting.

spec/registry.md
```diff
@@ -833,3 +818,25 @@ For more information, see [Working with Timezones](https://w3c.github.io/timezon
...
+
+- `hour12`
+  - `true`
+  - `false`
+- `calendar` \[OPTIONAL\]
+  - valid [Unicode Calendar Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeCalendarIdentifier)
+- `numberingSystem` \[OPTIONAL\]
+  - valid [Unicode Number System Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeNumberSystemIdentifier)
+- `timeZone` \[OPTIONAL\]
+  - valid identifier per [BCP175](https://www.rfc-editor.org/rfc/rfc6557)
```
macchiati, line 838:
```suggestion
  - valid identifier per either [BCP175](https://www.rfc-editor.org/rfc/rfc6557) or [Unicode Timezone Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeTimezoneIdentifier)
```

ICU4X and some other implementations use the short timezone IDs, so we should allow them.

macchiati (changes requested):

spec/registry.md
aphillips: No, we shouldn't. This would effectively **require** every implementation to support them. I would rather permit them as implementation-defined.

spec/registry.md
```diff
@@ -833,3 +818,25 @@ For more information, see [Working with Timezones](https://w3c.github.io/timezon
...
+
+**_<dfn>Date/time locale options</dfn>_** override options set by the current locale
+or an implementation-defined date/time _operand_ value.
+
+The following options and their values are required to be available on
+the functions `:datetime`, `:date`, and `:time`:
+
+- `hour12`
+  - `true`
+  - `false`
```
aphillips, line 832: I thought we said that this option would _replace_ `hourCycle` in the `:time` and `:datetime` functions above. Having both seems bad.

It should also say that the default is locale-defined.

spec/registry.md
aphillips, line 838: Each of the optional options I think should have a value space of "valid identifier per [something] or an implementation-defined value"

---

It is important to be able to set `none` or blank as an option, as a way to allow developers to "float" a date/time value. See [floating-time](https://w3c.github.io/timezone/#dfn-float)

spec/registry.md
```diff
@@ -833,3 +818,25 @@ For more information, see [Working with Timezones](https://w3c.github.io/timezon
 > The form of these serializations is known and is a de facto standard.
 > Support for these extensions is expected to be required in the post-tech preview.
 > See: https://datatracker.ietf.org/doc/draft-ietf-sedate-datetime-extended/
+
+### Date and Time Locale Options
+
+**_<dfn>Date/time locale options</dfn>_** override options set by the current locale
+or an implementation-defined date/time _operand_ value.
```
aphillips, line 825: I think this sentence is incomplete. I'm also still not fond of saying "locale" here (because I spend a lot of my life saying "time zone is orthogonal to locale"). Perhaps call these "date/time override options"? Even if not, the sentence should be correct.

```suggestion
**_<dfn>Date/time override options</dfn>_** are _options_ that allow an _expression_ to
override values set by the current locale,
or provided by the _formatting context_ (such as the default time zone),
or embedded in an implementation-defined date/time _operand_ value.
```

aphillips (changes requested):

spec/registry.md
eemeli: This PR removes `hourCycle` from `:datetime`, and it was never available on `:time`.

The default being locale-defined is mentioned in the note below the list.

spec/registry.md
aphillips: Here's a stab at a full fix of `timeZone`. Note that I replaced `none` with `local` for consistency with HTML and Java Temporal. I would use `float` but want to reduce cognitive burden.

```suggestion
  - well-formed identifier per [BCP175](https://www.rfc-editor.org/bcp/bcp175)
  - an implementation-defined value or identifier 
    (such as a [Unicode Timezone Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeTimezoneIdentifier))
  - `local`
  
> [!NOTE]
> The value `local` permits a _message_ to convert a date/time value
> into a [floating](https://www.w3.org/TR/timezone/#floating) time value
> (sometimes called a _plain_ or _local_ time value) by removing
> the association with a specific time zone.

> [!NOTE]
> Implementations SHOULD check if identifiers for each of these _options_ are valid;
> they SHOULD ignore invalid or unknown values;
> and they SHOULD emit a _Bad Option_ error for invalid or unknown values.
```

spec/registry.md
```diff
@@ -833,3 +818,25 @@ For more information, see [Working with Timezones](https://w3c.github.io/timezon
...
+
+The following options and their values are required to be available on
```
aphillips, line 827: Is it weird to say "required to be available" and then have some listed as "\[OPTIONAL\]" ?

spec/registry.md
```diff
@@ -833,3 +818,25 @@ For more information, see [Working with Timezones](https://w3c.github.io/timezon
...
+the functions `:datetime`, `:date`, and `:time`:
+
+- `hour12`
+  - `true`
+  - `false`
+- `calendar` \[OPTIONAL\]
+  - valid [Unicode Calendar Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeCalendarIdentifier)
+- `numberingSystem` \[OPTIONAL\]
+  - valid [Unicode Number System Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeNumberSystemIdentifier)
```
aphillips, line 836: Fixing valid vs. well-formed

```suggestion
 - well-formed [Unicode Calendar Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeCalendarIdentifier) 
   or an implementation-defined value
- `numberingSystem` \[OPTIONAL\]
  - well-formed [Unicode Number System Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeNumberSystemIdentifier) 
    or an implementation-defined value
```

spec/registry.md
macchiati: > No, we shouldn't. This would effectively require every implementation to support them. I would rather permit them as implementation-defined.

In general, I really don't think we want to require every implementation to support **all** valid identifiers — whereby 'support' means that they are required to do something reasonable.

For example, we shouldn't require that u:locale=def "make a difference", rather than being ignored  if that language is not supported by the implementation.

With that in mind, adding (such as a [Unicode Timezone Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeTimezoneIdentifier)) doesn't have a cost, because people don't have to support all the values that we imply.

BTW, [BCP175](https://www.rfc-editor.org/bcp/bcp175) doesn't define either well-formedness or validity. For well-formedness, the best would be https://github.com/eggert/tz/blob/main/theory.html, but that is not very rigorous. For validity, that would be defined by scanning certain files, and looking for lines starting with Zone.

spec/registry.md
macchiati: I do think we need to make it clear somewhere (either with a general statement or multiple specific ones) that if the option for (say) numberingSystem is a **valid** Unicode Number System Identifier (eg Latn), it can be ignored, but can't be misinterpreted as something else (eg Hebrew numbering system); moreover, that those take precedence over an implementation-defined value.

macchiati: Maybe somewhere have a statement about 

### Implementation-defined values

Implementation-defined option values are permitted for many options, but if there is a valid standard option value then it takes precedence. Thus it is wisest for implementation-defined option values to use values that are not well-formed standard option values.

macchiati: Ah, here is the example I was thinking of, where "or an implementation-defined value" applied to a option value, and is wrong, for the reasons I state in https://github.com/unicode-org/message-format-wg/pull/938 (but had misapplied that  reasoning to **operand**, not **option values**).

macchiati: So the two statements "or an implementation-defined value" need to be removed.

spec/registry.md
```diff
@@ -595,8 +595,10 @@ All other _operand_ values produce a _Bad Operand_ error.
 The `:datetime` function can use either the appropriate _style options_ 
 or can use a collection of _field options_ (but not both) to control the formatted 
 output.
+_Date/time locale options_ can be combined with either _style options_ or _field options_.
```
eemeli, line 598:
```suggestion
_Date/time override options_ can be combined with either _style options_ or _field options_.
```

spec/registry.md
```diff
@@ -733,6 +716,7 @@ The function `:date` has these _options_:
   - `long`
   - `medium` (default)
   - `short`
+- _Date/time locale options_
```
eemeli, line 719:
```suggestion
- _Date/time override options_
```

spec/registry.md
```diff
@@ -770,6 +754,7 @@ The function `:time` has these _options_:
   - `long`
   - `medium`
   - `short` (default)
+- _Date/time locale options_
```
eemeli, line 757:
```suggestion
- _Date/time override options_
```

spec/registry.md
```diff
@@ -833,3 +818,25 @@ For more information, see [Working with Timezones](https://w3c.github.io/timezon
 > The form of these serializations is known and is a de facto standard.
 > Support for these extensions is expected to be required in the post-tech preview.
 > See: https://datatracker.ietf.org/doc/draft-ietf-sedate-datetime-extended/
+
+### Date and Time Locale Options
```
eemeli, line 822:
```suggestion
### Date and Time Override Options
```

spec/registry.md
aphillips: I agree about not requiring _support_ for all valid identifiers and agree that unknown values should be ignored (but not necessarily dropped, in case they are consumed downstream).

What I'm trying to guard against is _requiring_ (or implying that it is required) implementers to build support for parsing e.g. short timezone IDs. I think we want to required that they _accept_ Olson ids (for the definition of "accept" I explained elsewhere). They are not required to do anything with any value (although users will be unhappy if real time zone IDs don't work). The formulation you're suggesting would strongly suggest that the short IDs "have" to be supported.

> BTW, [BCP175](https://www.rfc-editor.org/bcp/bcp175) doesn't define either well-formedness or validity.

Yeah, I noticed that yesterday. I hadn't been into the tzinfo docs in a while.

spec/registry.md
aphillips: "... or an implementation-defined value" is meant to cover values such as `com.ibm.icu.text.NumberingSystem` or `java.time.chrono.Chronology`, in which the identifier has been converted to a local representation. This is the same argument as allowing the `currency` option in `:currency` to accept a `java.util.Currency` object in a Java implementation rather than requiring that the ISO4217 code be unboxed.

Implementation defined values of this sort can only be passed into options as variables, since there is no way to construct them using MF2's syntax.

The other sort of implementation defined value, of course, would be locally defined literals. Elsewhere in the PR you and I are discussing the use of UTI short timezone IDs, which I would classify as a sort of "locally defined literal" (vs. the Olson IDs). Implementers are responsible for not hanging themselves or their users by making locally-defined and standard values distinct. 

However, allowing such values permits implementers to adopt whatever local idiom exists (on their platform, programming languages, etc.) for locales, time zones, calendars, and so forth, with the caution that messages that use these IDs will not be portable and that tools/translators won't possibly recognize the values.

It's kind of surprising, but we don't say what almost any of the options do except by implication. A responsible implementer will probably do all the right things. But nothing forbids a lazy implementer from causing all invocations of `:date` to return the string "wildebeest".

macchiati: > "... or an implementation-defined value" is meant to cover values such as com.ibm.icu.text.NumberingSystem or java.time.chrono.Chronology, in which the identifier has been converted to a local representation.

If that's what it means, then that's what it has to say. Moreover, it should be in a note, not part of the definition.

So the following would be fine (bold for the addition):

> - `calendar` (default is locale-specific)
>    - well-formed [Unicode Calendar Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeCalendarIdentifier)
>    - **Note: The option value could be an implementation-defined value that denotes the same calendar as a   well-formed [Unicode Calendar Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeCalendarIdentifier), but with a different format.**

Bear with me for  a moment.

We have in [registry.md](https://github.com/unicode-org/message-format-wg/blob/main/spec/registry.md):

> Implementations MAY accept additional option values for options defined here. However, such values might become defined with a different meaning in the future, including with a different, incompatible name or using an incompatible value space. Supporting implementation-specific option values for standard or optional functions is NOT RECOMMENDED.

We also have
> option         = identifier o "=" o (literal / variable)

The implications are that conformant implementation can interpret **any** of:

```
{$x :currency compactDisplay=short} 
{$x :currency compactDisplay=medium} 
{$x :currency compactDisplay=μικρός}
{$x :currency compactDisplay=|🐭|}
{$x :currency compactDisplay=$myDisplay}
```

It can also interpret:

```
{$x :currency currency=CAD} 
{$x :currency currency=MyCurrency} 
{$x :currency currency=δολάριοΚαναδά}
{$x :currency currency=|¥|}
{$x :currency currency=|🐭|}
{$x :currency currency=$myCurrency}
```

So we could have _for every single option_ have that extra clause, as currently stated PR. 

- `select`
   -  `plural` (default; see [Default Value of `select` Option](#default-value-of-select-option) below)
   -  `ordinal`
   -  `exact`
   - or an implementation-defined value

aphillips: > If that's what it means, then that's what it has to say. Moreover, it should be in a note, not part of the definition.

I can make changes to that effect. Note that notes are non-normative, so it probably can't go in a note.

> So the following would be fine (bold for the addition):
>> Note: The option value could be an implementation-defined value that denotes the same calendar as a well-formed [Unicode Calendar Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeCalendarIdentifier), but with a different format.

No, I think this is too precise. It supposes that all sets of calendar models exactly mirror (or are a subset of) CLDRs, when the reality is likely to be just a bit more chaotic.

> Bear with me for a moment.

I follow your argument. The examples don't move me: they are what the quoted stuff starting "Implementations MAY..." mean. But it does convince me that we should remove implementation-defined value to a note here and elsewhere. It also convinces me that the "Implementations MAY..." text needs to spell out that it is referring to literals, with implementation-defined types handled separately.

macchiati: That makes sense to me.

aphillips: I added the MAY in #944 and I edited my proposed text above so that it can be used in place of "valid"

spec/registry.md
macchiati: Just for the record, a narrow-ish well-formed EBNF of current and past timezone IDs would be:

```
tzId   := tzPath | tzEtc | tzOld
tzPath := tzPart ("/" tzPart)+;
tzPart := tzWord ("_" tzWord)*;
tzWord := [A-Z][a-z]*;
tzEtc  := "Etc/" ("UTC" | "GMT" ([+\-] \d{1,2})?)
tzOld  := HST" | "PST8PDT" | "MST" | "MST7MDT" | "CST6CDT" | "EST" | "EST5EDT" | "WET" | "CET" | "MET" | "EET" | "Factory"
```

A very loose well-formed EBNF would be

```
tzId  := [a-zA-Z0-9+\-/_]+
```

aphillips: My tendency is to want a version of the narrow-ish EBNF (in ABNF, since that's the dialect our WG uses). It wouldn't want to support `tzOld` and would want to add support for "UTC" as an explicit zone name. See suggestion thread below.

eemeli:
I updated the structure a bit to get rid of the weird phrasing, and to not include `hour12` for `:date`.

I'm honestly not sure what the current request is regarding valid vs. well-formed, and which externals specs to reference, so leaving those unchanged for now.

spec/registry.md
eemeli: I'm having some trouble understanding exactly why it's worthwhile to establish a canonical definition of a well-formed timezone identifier within the MF2 spec for an optional `timeZone` option.

This does not seem like something that's fully baked yet.

On a deeper level, I'm no longer sure that it's a good idea to forbid a datetime formatter from emitting a Bad Option error and using fallback representation when it's told to override a timezone with a value that it doesn't support, in particular as the output might not include a timezone identifier. That seems highly likely to produce misleading results.

Also, the latest suggestion above includes normative language in a non-normative note.

aphillips: Yeah, I'm not excited about defining the ABNF here in the least. I _want_ to point to CLDR, but that would bring in all of the short identifier crud.

> On a deeper level, I'm no longer sure that it's a good idea to forbid a datetime formatter from emitting a Bad Option error and using fallback representation when it's told to override a timezone with a value that it doesn't support, in particular as the output might not include a timezone identifier. That seems highly likely to produce misleading results.

I agree that we don't just want to drop the zone on the floor. You might never notice that the option is not working (or only sometimes is not working). This is particularly true for `:date` formatting (where the time zone _does_ matter, but the zone ID is almost never displayed in the results).

> Also, the latest suggestion above includes normative language in a non-normative note.

We should just get rid of that note. The problem of well-formed-but-invalid values seems like a problem for the function handler implementation. We permit errors to come out in formatting.md. Any reason to prescribe what each function does with each option?

```suggestion
  - well-formed time zone identifier
    (see [BCP175](https://www.rfc-editor.org/bpc/bpc175))
  - `local`
  - `UTC`
  
 A time zone identifier is well-formed if it matches `tzId` in the following ABNF:
>```abnf
> tzId   = tzPath / tzEtc
> tzPath = tzPart 1*("/" tzPart)
> tzPart = tzWord *("_" tzWord)
> tzWord = (%x41-5A) *(%x61-7A) ; Uppercase ASCII letter followed by lowercase letters
> tzEtc  = ("Etc/" ("UTC" / "GMT" (("+" / "-") 1*2DIGIT))
>```
  
> [!NOTE]
> The value `local` permits a _message_ to convert a date/time value
> into a [floating](https://www.w3.org/TR/timezone/#floating) time value
> (sometimes called a _plain_ or _local_ time value) 
> by removing the association with a specific time zone.
```

PR #912, opened 2024-10-20
Fix `:integer` option `useGrouping` values

aphillips:
I noticed that `:integer` does not include the "never" value for the option `useGrouping`. This is a bug.

### DIFF
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -281,6 +281,7 @@ function `:integer`:
 - `useGrouping`
   - `auto` (default)
   - `always`
+  - `never`
   - `min2`
 - `minimumIntegerDigits`
   - ([digit size option](#digit-size-options), default: `1`)


mihnita:
Probably because the ECMAScript version does not have it:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#usegrouping

But that API (in ECMAScript) looks bad.
* `usegrouping`: `[ "always", "auto", "min2", true, false ]`, with `true` same as `"always"`
* `signDisplay`: `[ "auto", "always", "exceptZero", "negative", "never" ]`

So `usegrouping` has `"always"` and booleans, `signDisplay` has `"always"` and `"never"`


But this is good.
Approving :-)

Mihai

aphillips:
@mihnita I found the same weirdness when writing this fix--I first went to `:number` to check what we did there... and then went to MDN when my code didn't work. I wouldn't mind if true/false worked too, but we should at least be minimally consistent (ECMA-402 should too).

eemeli:
@sffc @ryzokuken Would you agree that introducing `'never'` as a valid `Intl.NumberFormat` `useGrouping` value would be appropriate? Without that, implementing this in MF2 without being compatible with ECMA-402 is not possible.

aphillips:
@eemeli noted:
> Without that, implementing this in MF2 without being compatible with ECMA-402 is not possible.

Not sure I entirely agree. It would be ideal if both 402 and MFWG used the same option values for the same options. But it would be okay too if the two are consistent with one-another (and it's just a mapping). The lack of internal consistency in `Intl.NumberFormat` is weird, but not a consideration for MF2.

sffc:
It was previously decided that `"never"` would not be allowed _in ECMA-402_ because the string `"never"` is truthy and would cause older implementations to show grouping separators. `false` was unambiguous. But, that is an ECMA-402 artifact, and I would encourage other clients unweighted by that legacy to use a pure string enumeration.

ryzokuken:
As @sffc points out, this is one of the rarer cases where being incompatible with this legacy behavior would actually be better for MF2.

PR #915, opened 2024-10-26
Implement `:currency` function in the default registry

aphillips:
Initial implementation of the `:currency` function.

This function will be **standard** in 2.0.

Note behavior differences from `:number` and `:integer`.

`:unit` is in #922.

### DIFF
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -454,12 +454,14 @@ The following options and their values are required to be available on the funct
   - ([digit size option](#digit-size-options), default: `1`)
 - `fractionDigits` (unlike number/integer formats, the fraction digits for currency formatting are fixed)
   - `auto` (default) (the number of digits used by the currency)
-  - `hideIfWhole` (see note above)
   - ([digit size option](#digit-size-options))
 - `minimumSignificantDigits`
   - ([digit size option](#digit-size-options))
 - `maximumSignificantDigits`
   - ([digit size option](#digit-size-options))
+- `trailingZeroDisplay`
+  - `auto` (default)
+  - `stripIfInteger`
 
 If the _operand_ of the _expression_ is an implementation-defined type,
 such as the _resolved value_ of an _expression_ with a `:currency` _annotation_,
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -340,6 +340,12 @@ Using this option in such a case results in a _Bad Option_ error.
 The value of the _operand_'s `currency` MUST be either a string containing a
 well-formed [Unicode Currency Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeCurrencyIdentifier)
 or an implementation-defined currency type.
+Although currency codes are expected to be uppercase,
+implementations SHOULD treat them in a case-insensitive manner.
+A well-formed Unicode Currency Identifier matches the production `currency_code` in this ABNF:
+```abnf
+currency_code = 3ALPHA
+```
 
 A [Number Operand](#number-operands) without a `currency` _option_ results in a _Bad Operand_ error.
 
@@ -353,7 +359,6 @@ A [Number Operand](#number-operands) without a `currency` _option_ results in a 
 > (there are many recent examples of this occuring).
 > It also avoids requiring implementations to check currency codes for validity,
 > although implementations are permitted to emit _Bad Option_ or _Bad Operand_ for invalid codes.
-> A well-formed currency code matches the regular expression `^[A-Z]{3}$`.
 
 > [!NOTE]
 > For runtime environments that do not provide a ready-made data structure,
@@ -363,7 +368,7 @@ A [Number Operand](#number-operands) without a `currency` _option_ results in a 
 > For example, such an implementation might define a "currency operand"
 > to include a key-value structure with specific keys to be the
 > local currency operand, which might look like the following:
-> ```json
+> ```
 > {
 >    "value": 123.45,
 >    "currency": "EUR"
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -324,7 +324,7 @@ together with the resolved options' values.
 
 ### The `:currency` function
 
-The function `:currency` is an OPTIONAL selector and formatter for currency values, 
+The function `:currency` is a selector and formatter for currency values, 
 which are a specialized form of numeric selection and formatting.
 
 #### Operands
@@ -373,9 +373,24 @@ the value of other options, or all of these.
 Fraction digits for currency values behave differently than for other numeric formatters.
 The number of fraction digits displayed is usually set by the currency used.
 For example, USD uses 2 fraction digits, while JPY uses none.
-Setting `fractionDigits` to `none` makes the display and selection work similar to `:integer`.
 Setting some other number of `fractionDigits` allows greater precision display
 (such as when performing currency conversions or other specialized operations)
+or disabling fraction digits if set to `0`.
+The special _option_ _value_ `hideIfWhole` is used to display values without
+fraction digits when the number of fraction digits is zero,
+or based on the currency when the number of fraction digits for the currency is non-zero.
+> For example, this _message_:
+> ```
+> The special price is {$price :currency fractionDigits=hideIfWhole}.
+> ```
+> When used with the value `5.00 USD` in the `en-US` locale displays as:
+> ```
+> The special price is $5.
+> ```
+> But like this when when value is `5.01 USD`:
+> ```
+> The special price is $5.01.
+> ```
 
 > [!NOTE]
 > Except where noted otherwise, the names of _options_ and their _values_ were derived from the
@@ -419,7 +434,7 @@ The following options and their values are required to be available on the funct
   - ([digit size option](#digit-size-options), default: `1`)
 - `fractionDigits` (unlike number/integer formats, the fraction digits for currency formatting are fixed)
   - `auto` (default) (the number of digits used by the currency)
-  - `none`
+  - `hideIfWhole` (see note above)
   - ([digit size option](#digit-size-options))
 - `minimumSignificantDigits`
   - ([digit size option](#digit-size-options))
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -464,9 +464,8 @@ The following options and their values are required to be available on the funct
   - `symbol` (default)
   - `name`
   - `code`
-  - `formal`
-  - `variant`
-  - `none` (this is called `hidden` in ICU)
+  - `formalSymbol`
+  - `never` (this is called `hidden` in ICU)
 - `useGrouping`
   - `auto` (default)
   - `always`
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -322,7 +322,7 @@ its _resolved value_ contains the implementation-defined integer value
 of the _operand_ of the annotated _expression_,
 together with the resolved options' values.
 
-## The `:currency` function
+### The `:currency` function
 
 The function `:currency` is an OPTIONAL selector and formatter for currency values, 
 which are a specialized form of numeric selection and formatting.
@@ -439,8 +439,8 @@ The _function_ `:currency` performs selection as described in [Number Selection]
 #### Composition
 
 When an _operand_ or an _option_ value uses a _variable_ annotated,
-directly or indirectly, by a `:number` _annotation_,
-its _resolved value_ contains an implementation-defined numerical value
+directly or indirectly, by a `:currency` _annotation_,
+its _resolved value_ contains an implementation-defined currency value
 of the _operand_ of the annotated _expression_,
 together with the resolved options' values.
 
@@ -482,13 +482,20 @@ All other values produce a _Bad Operand_ error.
 The _operand_ of the `:currency` function is an implementation-defined type that
 contains a number and a currency code
 or a map whose keys are `value` and `currency`.
+When the _operand_ is a map, the value of `value` MUST be a [Number Operand](#number-operands).
+When the _operand_ is a map, the value of `currency` MUST be either a 
+valid [Unicode Currency Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeCurrencyIdentifier)
+or an implementation-defined currency type.
 
 > [!NOTE]
 > For example, in ICU4J, the type `com.ibm.icu.util.CurrencyAmount` can be used
-> to set the currency.
+> to set the amount and currency.
 
 The _operand_ MAY be a [Number Operand](#number-operands), as long as the option `currency`
 is provided.
+The option `currency` MUST NOT be used to override the currency of an implementation-defined type
+or of a map containing a `currency` field.
+Using this option in such a case results in a _Bad Option_ error.
 
 ### Digit Size Options
 
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -327,10 +327,40 @@ together with the resolved options' values.
 The function `:currency` is an OPTIONAL selector and formatter for currency values, 
 which are a specialized form of numeric selection and formatting.
 
-#### `:currency` Operands
+#### Operands
 
-The function `:currency` requires a [Currency Operand](#currency-operands) as its _operand_
-or a [Number Operand](#number-operands), if used with the _option_ `currency`.
+The _operand_ of the `:currency` function can be one of any number of
+implementation-defined types,
+each of which contains a numerical `value` and a `currency`;
+or it can be a [Number Operand](#number-operands), as long as the option
+`currency` is provided.
+The option `currency` MUST NOT be used to override the currency of an implementation-defined type.
+Using this option in such a case results in a _Bad Option_ error.
+
+The value of the _operand_'s `currency` MUST be either a string containing a
+valid [Unicode Currency Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeCurrencyIdentifier)
+or an implementation-defined currency type.
+
+A [Number Operand](#number-operands) without a `currency` _option_ results in a _Bad Operand_ error.
+
+> [!NOTE]
+> For example, in ICU4J, the type `com.ibm.icu.util.CurrencyAmount` can be used
+> to set the amount and currency.
+
+> [!NOTE]
+> For runtime environments that do not provide a ready-made data structure,
+> class, or type for currency values, the implementation ought to provide
+> a data structure, convenience function, or documentation on how to encode
+> the value and currency code for formatting.
+> For example, such an implementation might define a "currency operand"
+> to include a key-value structure with specific keys to be the
+> local currency operand, which might look like the following:
+> ```json
+> "operandName": {
+>    "value": 123.45,
+>    "currency": "EUR"
+> }
+> ```
 
 #### Options
 
@@ -454,24 +484,6 @@ All other values produce a _Bad Operand_ error.
 > a value that includes a unit
 > or the type `com.ibm.icu.util.CurrencyAmount` can be used to set the currency and related
 > options (such as the number of fraction digits).
-
-### Currency Operands
-
-The _operand_ of the `:currency` function is one or more implementation-defined types
-each of which contains a numerical `value` and a `currency`
-or a [Number Operand](#number-operands), if used with the _option_ `currency`.
-The value of `currency` MUST be either a string containing a
-valid [Unicode Currency Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeCurrencyIdentifier)
-or an implementation-defined currency type.
-
-> [!NOTE]
-> For example, in ICU4J, the type `com.ibm.icu.util.CurrencyAmount` can be used
-> to set the amount and currency.
-
-The _operand_ MAY be a [Number Operand](#number-operands), as long as the option `currency`
-is provided.
-The option `currency` MUST NOT be used to override the currency of an implementation-defined type.
-Using this option in such a case results in a _Bad Option_ error.
 
 ### Digit Size Options
 
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -392,15 +392,22 @@ or based on the currency when the number of fraction digits for the currency is 
 > The special price is $5.01.
 > ```
 
+Implementations MAY internally alias option values that they do not have data or a backing implementation for.
+Notably, the `currencyDisplay` option has a rich set of values that mirrors developments in CLDR data.
+Some implementations might not be able to produce all of these formats for every currency.
+
 > [!NOTE]
 > Except where noted otherwise, the names of _options_ and their _values_ were derived from the
 > [options](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#options)
 > in JavaScript's `Intl.NumberFormat`.
 
+> [!NOTE]
+> The option `select` does not accept the value `ordinal` because selecting
+> currency values using ordinal rules makes no sense.
+
 The following options and their values are required to be available on the function `:currency`:
 - `select`
    -  `plural` (default)
-   -  `ordinal`
    -  `exact`
 - `currency`
    - valid [Unicode Currency Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeCurrencyIdentifier)
@@ -417,14 +424,14 @@ The following options and their values are required to be available on the funct
 - `currencySign`
   - `accounting`
   - `standard` (default)
-- `currencyDisplay` (this option's values are derived from those in ICU NumberFormatter)
-  - `auto` (default)
+- `currencyDisplay` (this option's values are derived from those in ICU [NumberFormatter.UnitWidth](https://unicode-org.github.io/icu-docs/apidoc/released/icu4j/com/ibm/icu/number/NumberFormatter.UnitWidth.html))
   - `narrow`
-  - `short`
-  - `iso`
+  - `short` (default)
   - `full`
+  - `iso`
   - `formal`
   - `variant`
+  - `none` (this is called `hidden` in ICU)
 - `useGrouping`
   - `auto` (default)
   - `always`
@@ -449,11 +456,11 @@ with _options_ on the _expression_ taking priority over any option values of the
 
 > For example, the _placeholder_ in this _message_:
 > ```
-> .input {$n :currency currency=USD fractionDigits=none}
+> .input {$n :currency currency=USD fractionDigits=hideIfWhole}
 > {{{$n :currency currencySign=accounting}}}
 > ```
 > would be formatted with the resolved options
-> `{ currencySign: 'accounting', fractionDigits: 'none', currency: 'USD' }`.
+> `{ currencySign: 'accounting', fractionDigits: 'hideIfWhole', currency: 'USD' }`.
 
 #### Selection
 
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -480,10 +480,8 @@ All other values produce a _Bad Operand_ error.
 ### Currency Operands
 
 The _operand_ of the `:currency` function is an implementation-defined type that
-contains a number and a currency code
-or a map whose keys are `value` and `currency`.
-When the _operand_ is a map, the value of `value` MUST be a [Number Operand](#number-operands).
-When the _operand_ is a map, the value of `currency` MUST be either a 
+contains a numerical `value` and a `currency` code.
+The value of `currency` MUST be either a 
 valid [Unicode Currency Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeCurrencyIdentifier)
 or an implementation-defined currency type.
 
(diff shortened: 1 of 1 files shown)


spec/registry.md
```diff
@@ -384,6 +322,128 @@ its _resolved value_ contains the implementation-defined integer value
...
+  - `accounting`
+  - `standard` (default)
+- `currencyDisplay` (this option's values are derived from those in ICU NumberFormatter)
+  - `auto` (default)
+  - `narrow`
+  - `short`
+  - `iso`
+  - `full`
+  - `formal`
+  - `variant`
```
eemeli, line 382: We should not mix-and-match sources like this, and here too use the Intl.NumberFormat values:
```suggestion
- `currencyDisplay`
  - `code`
  - `symbol` (default)
  - `narrowSymbol`
  - `name`
```

spec/registry.md
```diff
@@ -384,6 +322,128 @@ its _resolved value_ contains the implementation-defined integer value
...
+Some options do not have default values defined in this specification.
+The defaults for these options are implementation-dependent.
+In general, the default values for such options depend on the locale, 
+the currency,
+the value of other options, or all of these.
+
+Fraction digits for currency values behave differently than for other numeric formatters.
+The number of fraction digits displayed is usually set by the currency used.
+For example, USD uses 2 fraction digits, while JPY uses none.
+Setting `fractionDigits` to `none` makes the display and selection work similar to `:integer`.
```
eemeli, line 346: Given this, how would `fractionDigits=none` be different from `fractionDigits=0`? Why is a name needed in addition to the numerical value?

spec/registry.md
```diff
@@ -384,6 +322,128 @@ its _resolved value_ contains the implementation-defined integer value
...
+
+> For example, the _placeholder_ in this _message_:
+> ```
+> .input {$n :currency currency=USD fractionDigits=none}
+> {{{$n :currency currencySign=accounting}}}
+> ```
+> would be formatted with the resolved options
+> `{ currencySign: 'accounting', fractionDigits: 'none', currency: 'USD' }`.
+
+##### Default Value of `select` Option
```
eemeli, line 413: This section, and in fact all the text around selection, should be left out and replaced with a reference to `:number`. As is, having this text copied here is confusing.

spec/registry.md
```diff
@@ -417,6 +477,26 @@ All other values produce a _Bad Operand_ error.
...
 
+### Currency Operands
+
+The _operand_ of the `:currency` function is an implementation-defined type that
+contains a number and a currency code
+or a map whose keys are `value` and `currency`.
+When the _operand_ is a map, the value of `value` MUST be a [Number Operand](#number-operands).
+When the _operand_ is a map, the value of `currency` MUST be either a 
+valid [Unicode Currency Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeCurrencyIdentifier)
+or an implementation-defined currency type.
```
eemeli, line 488: We should not require support for such a map value, but leave it to implementations.
```suggestion
The _operand_ of the `:currency` function is an implementation-defined type that
contains a numerical `value` and a `currency` code.
The value of `currency` MUST be either a 
valid [Unicode Currency Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeCurrencyIdentifier)
or an implementation-defined currency type.
```

spec/registry.md
```diff
@@ -417,6 +477,26 @@ All other values produce a _Bad Operand_ error.
...
+
+> [!NOTE]
+> For example, in ICU4J, the type `com.ibm.icu.util.CurrencyAmount` can be used
+> to set the amount and currency.
+
+The _operand_ MAY be a [Number Operand](#number-operands), as long as the option `currency`
+is provided.
+The option `currency` MUST NOT be used to override the currency of an implementation-defined type
+or of a map containing a `currency` field.
```
eemeli, line 497:
```suggestion
The option `currency` MUST NOT be used to override the currency of an implementation-defined type.
```

eemeli (changes requested):

spec/registry.md
aphillips: I did this carefully and for a reason: there is no way in `Intl` to access important features, such as the alternate (`variant`) symbol (The Turkish lira is a problem here: the new `₺` symbol is not used for a lot of values vs. `TL`--and the symbols go on opposite sides of the number to boot) or for uses such as in-sentence presentation (JPY, CNY, TWD, and a few others have this, e.g. `￥123,457` vs. `123,457円`). I also omitted "hidden", which was not intentional.

We don't currently have an allowance for optional values. If "my" full list were mandated, Intl-based implementations could just alias the unsupported ones.

spec/registry.md
aphillips: My thinking is that `fractionDigits=auto` is the high-runner case and `fractionDigits=none` is in second place. It's semantically different than setting a specific number of digits?

There might be rounding mode differences, I suppose: I didn't deal with rounding mode. Otherwise, there is no difference.

spec/registry.md
aphillips: Fair enough. I'm going to merge your suggestion and then modify is slightly to allow multiple types (my--and thus your--current text implies only one type is allowed, which is not the intention)

spec/registry.md
eemeli: I would rather strongly prefer dropping `none`.

spec/registry.md
eemeli: I would not be comfortable adding option values here that are not supported by Intl.NumberFormat. I think it's quite important for an MF2 formatter to be implementable with the pre-existing TC39 Intl formatters, and that using a completely different -- and wider -- set of values like this seems rather arbitrary, even if done intentionally.

I also find the proposed ICU NumberFormatter values to be significantly harder to decipher without reference to documentation, compared to the TC39 option values.

> We don't currently have an allowance for optional values.

I would support allowing implementations to support option values beyond the set we define.

spec/registry.md
```diff
@@ -384,6 +322,106 @@ its _resolved value_ contains the implementation-defined integer value
...
+(such as when performing currency conversions or other specialized operations)
+
+> [!NOTE]
+> Except where noted otherwise, the names of _options_ and their _values_ were derived from the
+> [options](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#options)
+> in JavaScript's `Intl.NumberFormat`.
+
+The following options and their values are required to be available on the function `:currency`:
+- `select`
+   -  `plural` (default; see [Default Value of `select` Option](#default-value-of-select-option) below)
```
eemeli, line 357:
```suggestion
   -  `plural` (default)
```

spec/registry.md
```diff
@@ -417,6 +455,24 @@ All other values produce a _Bad Operand_ error.
 > or the type `com.ibm.icu.util.CurrencyAmount` can be used to set the currency and related
 > options (such as the number of fraction digits).
 
+### Currency Operands
```
eemeli, line 458: Having this section here seems odd. It would be better to have this text directly included in the "`:currency` Operands" section.

spec/registry.md
aphillips: I think we should discuss it. My tendency is to agree with you, but...

In the course of writing the text (and my replies) I found several bugs in ICU4J 76 RC. I'm more-than-averagely sensitive to currency formatting details and, thus, am concerned that we don't end up where parts of ICU have been: unable to reach the solutions already present in CLDR.

spec/registry.md
```diff
@@ -384,6 +322,136 @@ its _resolved value_ contains the implementation-defined integer value
...
+> [!NOTE]
+> For runtime environments that do not provide a ready-made data structure,
+> class, or type for currency values, the implementation ought to provide
+> a data structure, convenience function, or documentation on how to encode
+> the value and currency code for formatting.
+> For example, such an implementation might define a "currency operand"
+> to include a key-value structure with specific keys to be the
+> local currency operand, which might look like the following:
+> ```json
+> "operandName": {
```
eemeli, line 359:
```suggestion
> {
```

aphillips:
We agreed to take this in the 2024-11-04 call, provided the last changes were incorporated. This will fast-track, assuming I get positive reviews.

spec/registry.md
aphillips: It was removed.

spec/registry.md
```diff
@@ -384,6 +322,158 @@ its _resolved value_ contains the implementation-defined integer value
...
+
+Fraction digits for currency values behave differently than for other numeric formatters.
+The number of fraction digits displayed is usually set by the currency used.
+For example, USD uses 2 fraction digits, while JPY uses none.
+Setting some other number of `fractionDigits` allows greater precision display
+(such as when performing currency conversions or other specialized operations)
+or disabling fraction digits if set to `0`.
+The special _option_ _value_ `hideIfWhole` is used to display values without
+fraction digits when the number of fraction digits is zero,
+or based on the currency when the number of fraction digits for the currency is non-zero.
```
eemeli, line 381: This should be left out. This is novel currency formatting behaviour, and we should not be introducing it here via message formatting. Is this already somehow supported in ICU4J and ICU4C? If this makes sense for formatting currencies, wouldn't this also make sense for many other number formatting situations?

I do not object to a value like this in principle, but I object to it being introduced _here_.

spec/registry.md
```diff
@@ -384,6 +322,158 @@ its _resolved value_ contains the implementation-defined integer value
...
+  - `accounting`
+  - `standard` (default)
+- `currencyDisplay` (this option's values are derived from those in ICU [NumberFormatter.UnitWidth](https://unicode-org.github.io/icu-docs/apidoc/released/icu4j/com/ibm/icu/number/NumberFormatter.UnitWidth.html))
+  - `narrow`
+  - `short` (default)
+  - `full`
+  - `iso`
+  - `formal`
+  - `variant`
+  - `none` (this is called `hidden` in ICU)
```
eemeli, line 434: I continue to object to using a different source for this one single function option, when all the other options for all the other functions are using TC39 Intl as their source. Using this set of option values unnecessarily breaks from our otherwise well established patterns.

spec/registry.md
```diff
@@ -384,6 +322,158 @@ its _resolved value_ contains the implementation-defined integer value
...
+> The option `select` does not accept the value `ordinal` because selecting
+> currency values using ordinal rules makes no sense.
+
+The following options and their values are required to be available on the function `:currency`:
+- `select`
+   -  `plural` (default)
+   -  `exact`
+- `currency`
+   - valid [Unicode Currency Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeCurrencyIdentifier)
+     (no default)
```
eemeli, line 414: Just to check, an implementation that accepts all `currency` values that match `^[A-Z]{3}$` would be valid, yes? We should not be here requiring that each MF2 implementation carries with it a list of valid currency codes, and emits an error for currency codes not in that list.

eemeli (changes requested):

spec/registry.md
macchiati: ^[A-Z]{3}$ would be well-formed, but not necessarily valid. (We need to make sure that we preserve that distinction here and elsewhere — ill-formed must cause an error; invalid could be handled differently according to the implementation, eg distinguishing between deprecated and non-deprecated.

spec/registry.md
aphillips: This value is supported by ICU.

The use cases for currencies are more common than for other number values (with the possible exception of units). Normally one is trying to get consistent widths (or one doesn't care). With currency values you generally want a specific number of fractional digits or no fraction digits. Mutation of the value to produce this is undesirable.

However, I agree that this is "special sauce" and shouldn't be _required_ here. Should we introduce **optional** values?

spec/registry.md
aphillips: We discussed this in yesterday's call (which you weren't able to attend). The challenge here is that currency formatting has been greatly improved in CLDR (which is reflected by the changes in ICU) after `Intl.NumberFormat` was created. Those changes have not yet materialized in TC39, but really are needed. FWIW, there is continuing evolution here. To wit, there are a number of currencies with more than one symbol (some of which are `formal` and some of which are `variant`). These symbols need to be selectable and, when selected, can result in different formatting patterns.

The poster child for this is the Turkish Lira, which has a newfangled symbol, which displays on the left thusly: `₺123.456,79` , but which is very often rendered using the older ASCII symbol "TL":

image

The discussion yesterday recognized that there are more options here than in Intl, but that implementations are only required to accept all these keywords--not to provide different results based on them. Intl provides `code`, `symbol`, `narrowSymbol`, and `name`. We could reasonably rename thusly:

| Proposed Above | Intl Equivalent | Suggested | Comments |
|:---:|:---:|:---:|---|
| `narrow` | `narrowSymbol` | `narrow` | No need to be extra verbose |
| `short` | `symbol` | `symbol` | (default) `symbol` is consistent with user expectations |
| `full` | `name` | `name` | ICU's name is actually `FULL_NAME`. Can be misleading, e.g. JPY renders as `123,457円` |
| `iso` | `code` | `code` | `code` might be better because some codes are not ISO4217 |
| `formal` | n/a | `formal` | Variations such as `TRD` => `NT$` |
| `variant` | n/a | `variant` | Variations such as the `TRY` one illustrated above |
| `none` | n/a | `none` | Frequently used for columns of currency values |

Would that rename work for you?

spec/registry.md
aphillips: Should this say "well-formed" then? 

I note that ICU and Intl both accept (well-formed) garbage here without error, partly as a guard against future code assignments requiring global software updates to work.

aphillips: Actually, made the change and added a note.

spec/registry.md
aphillips: I did the rename just now (in bbc7091) but am looking for @eemeli's feedback

spec/registry.md
macchiati: Thanks, the change is good.

spec/registry.md
```diff
@@ -384,6 +322,166 @@ its _resolved value_ contains the implementation-defined integer value
...
+   - `standard` (default)
+   - `compact`
+- `numberingSystem`
+   - valid [Unicode Number System Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeNumberSystemIdentifier)
+     (default is locale-specific)
+- `currencySign`
+  - `accounting`
+  - `standard` (default)
+- `currencyDisplay` (this option's values are derived from those in ICU [NumberFormatter.UnitWidth](https://unicode-org.github.io/icu-docs/apidoc/released/icu4j/com/ibm/icu/number/NumberFormatter.UnitWidth.html))
+  - `narrow`
```
eemeli, line 436: @aphillips comments on this in https://github.com/unicode-org/message-format-wg/pull/915#discussion_r1829540496 with:
> No need to be extra verbose

The need here is for consistency. We should not be editorialising our source materials like this, and creating unnecessary inconsistencies.

I would also argue that "narrowSymbol" is much more clearly narrower than "symbol", and so without looking up the meanings in some source this makes it clearer which of "US$" and "$" is which, as otherwise a reader could more easily presume that the former is a "narrow" name, and the latter the "symbol".

```suggestion
  - `narrowSymbol`
```

spec/registry.md
eemeli: As we allow for implementations to support option values beyond the ones we explicitly define, the easiest way of making this optional is to leave it out for now. This would also more clearly allow for an implementation that did not support "hide if whole" fraction display to emit an error, rather than be required to fail silently.

> This value is supported by ICU.

Could you give me a pointer to how/where this is supported? I could not find this myself.

spec/registry.md
```diff
@@ -347,6 +347,14 @@ A [Number Operand](#number-operands) without a `currency` _option_ results in a
 > For example, in ICU4J, the type `com.ibm.icu.util.CurrencyAmount` can be used
 > to set the amount and currency.
 
+> [!NOTE]
+> The `currency` is only required to be well-formed rather than checked for validity.
+> This allows new currency codes to be defined 
+> (there are many recent examples of this occuring).
+> It also avoids requiring implementations to check currency codes for validity,
+> although implementations are permitted to emit _Bad Option_ or _Bad Operand_ for invalid codes.
+> A well-formed currency code matches the regular expression `^[A-Z]{3}$`.
```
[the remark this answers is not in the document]
aphillips: The source on 341 says:

> Codes consisting of 3 ASCII letters that are or have been valid in ISO 4217, plus certain additional codes that are or have been in common use. The list of countries and time periods associated with each currency value is available in [Supplemental Currency Data](https://cldr-smoke.unicode.org/spec/main/ldml/tr35-numbers.html#Supplemental_Currency_Data), plus the default number of decimals.
>
> The XXX code is given a broader interpretation as Unknown or Invalid Currency.

This doesn't distinguish valid and well-formed. I suppose we might link [ISO4217](https://www.iso.org/standard/64758.html)'s "about" page. Not sure anyone defines well-formed as a distinct entity.

In practice, this defines an alpha3 code. I should probably use ABNF (since we use that everywhere else in our spec) to define it. I'll provide an edit.

spec/registry.md
aphillips: I'll buy your argument and am committing your suggestion. I'll note that `narrowSymbol`/`symbol` don't necessarily produce `$` and `US$`---it really depends on the locale.

macchiati:
So to recap, https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeCurrencyIdentifier defines the validity precisely, but is squishy about the well-formedness.

- The valid values are those name attribute values in the type elements of key name="cu" in bcp47/[currency.xml](https://github.com/unicode-org/cldr/blob/main/common/bcp47/currency.xml).

As to well-formedness, instead of
> Codes consisting of 3 ASCII letters that are or have been valid ...
It should say:
> Well-formed codes are of the form `[A-Z]{3}`.
> The valid codes are or have been valid ...

(The second sentence is a description of how the valid values were derived, not the definition.)

Let me see about getting the TC to add this to the PR I already have for spec changes in 46.1

spec/registry.md
eemeli: Ah, thank you, I was looking for it among the precision options, as that's how it's presented here. Finding it for ICU let me also find the corresponding [JS option](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#trailingzerodisplay), which to no-one's surprise has a different spelling: `trailingZeroDisplay: 'stripIfInteger'`.

So let me amend my position here, and reiterate what I've stated previously: I'm fine with any subset of the Intl formatters and their options. In this case, I think we should add this as a separate option, following the TC39 nomenclature. This will also allow us to add it for `:number`.

I would be happy for us to include other rounding options for `:number` and related functions, which were previously left out.

spec/registry.md
```diff
@@ -384,6 +322,171 @@ its _resolved value_ contains the implementation-defined integer value
...
+- `minimumIntegerDigits`
+  - ([digit size option](#digit-size-options), default: `1`)
+- `fractionDigits` (unlike number/integer formats, the fraction digits for currency formatting are fixed)
+  - `auto` (default) (the number of digits used by the currency)
+  - `hideIfWhole` (see note above)
+  - ([digit size option](#digit-size-options))
+- `minimumSignificantDigits`
+  - ([digit size option](#digit-size-options))
+- `maximumSignificantDigits`
+  - ([digit size option](#digit-size-options))
```
eemeli, line 462:
```suggestion
- `fractionDigits` (unlike number/integer formats, the fraction digits for currency formatting are fixed)
  - `auto` (default) (the number of digits used by the currency)
  - ([digit size option](#digit-size-options))
- `minimumSignificantDigits`
  - ([digit size option](#digit-size-options))
- `maximumSignificantDigits`
  - ([digit size option](#digit-size-options))
- `trailingZeroDisplay`
  - `auto` (default)
  - `stripIfInteger`
```

spec/registry.md
aphillips: I can add the separate option using the Intl spelling. Do you have a list of other options to add?

spec/registry.md
aphillips: I'm going to add this, but it will need changes to the note text about digits. I'll push another change for that.

eemeli (commented):
My blocking objections to this PR have now been resolved. I would still prefer for the last three `currencyDisplay` option values to be left out, but I can live with them being included.

spec/registry.md
eemeli: We're missing the following, which could well be added in a separate PR:
- `roundingPriority`
  - `auto` (default)
  - `morePrecision`
  - `lessPrecision`
- `roundingIncrement`
  - 1 (default), 2, 5, 10, 20, 25, 50, 100, 200, 250, 500, 1000, 2000, 2500, and 5000
- `roundingMode`
  - `ceil`
  - `floor`
  - `expand`
  - `trunc`
  - `halfCeil`
  - `halfFloor`
  - `halfExpand` (default)
  - `halfTrunc`
  - `halfEven`

aphillips: I can't do this right this second, but will add these to this PR. Should they be optional or standard? I'm leaning towards standard. 

Almost all of our work between now and 46.1 appears to be fixing up the registry/function set, plus some housekeeping.

spec/registry.md
```diff
@@ -384,6 +325,176 @@ its _resolved value_ contains the implementation-defined integer value
...
+  - `accounting`
+  - `standard` (default)
+- `currencyDisplay` (this option's values are derived from those in ICU [NumberFormatter.UnitWidth](https://unicode-org.github.io/icu-docs/apidoc/released/icu4j/com/ibm/icu/number/NumberFormatter.UnitWidth.html))
+  - `narrowSymbol`
+  - `symbol` (default)
+  - `name`
+  - `code`
+  - `formal`
+  - `variant`
+  - `none` (this is called `hidden` in ICU)
```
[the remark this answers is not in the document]
eemeli: Is the fallback behaviour for formatting currencies with no variant documented somewhere?

spec/registry.md
eemeli: > Should they be optional or standard?

I'm fine either way.

> Almost all of our work between now and 46.1 appears to be fixing up the registry/function set, plus some housekeeping.

Good!

spec/registry.md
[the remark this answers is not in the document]
macchiati: Yes, the fallback for each element with `alt=...` is documented. It is normally the unmarked element (ie, without the `alt=...`), but in some documented cases (notably locale display names) it may be constructed.

macchiati (approved):

spec/registry.md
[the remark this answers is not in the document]
aphillips: Note that `variant` is restricted (currently) to currency symbols. 

The usability problem is that when a user writes the source message, she's not going to know about the need for `variant`. At Amazon, our currency formatter had enums for the presentation format ("in sentence", "standalone", "compact", "iso code", and such) that correspond to specific user interface use cases. Various locale/currency/site configurations then produced the correct symbol/pattern in each of these. This was helpful because it let us do one thing with (e.g.) `TRY` in the Türkiye site and different thing elsewhere without the developer (or translator) having to know whether a variant existed and, further, whether it was desired or not. Often "what to do" will depend entirely on which currency and which presentation case is wanted ☹️ 

The enums here are taken from `NumberFormatter` (and `Intl.NumberFormat`), which are procedural APIs for getting this effect. They are wrong for `MessageFormat`, I think, because MF is supposed to hide procedural gorp like that. As currently speced, the developer to know that they want the `variant` symbol all the time (unlikely) or the translator to put it there (possibly at odds with the display use case) for specific locales (and not generally knowing the currency). You don't want to have to ever write:

```
.local $curCode = {$amount :get field=currency}
.match $curCode
TRY {{You have {$amount :currency currencyDisplay=variant} left in your account.}}
UAH {{You have {$amount :currency currencyDisplay=variant} left in your account.}}
NTD {{You have {$amount :currency currencyDisplay=formal} left in your account.}}
*   {{You have {$mount :currency} left in your account.}}
```

spec/registry.md
aphillips: Added. I made them standard.

@mihnita @catamorphism Please note this (proposed) change.

spec/registry.md
```diff
@@ -381,6 +338,190 @@ together with the resolved options' values.
...
+> ```
+> When used with the value `5.00 USD` in the `en-US` locale displays as:
+> ```
+> The special price is $5.
+> ```
+> But like this when when value is `5.01 USD`:
+> ```
+> The special price is $5.01.
+> ```
+
```
eemeli, line 426: Nitpick: This seems a bit superfluous as we don't explain other options like this, but I don't object to including it.

```suggestion
```

eemeli (approved):
With the latest changes, this looks good, provided that changes in #925 will allow an implementation to emit some signal about unsupported option values.

spec/registry.md
aphillips: We could put this in the user guide. I included it here because it might otherwise be mysterious.

spec/registry.md
```diff
@@ -155,6 +155,25 @@ The following options and their values are required to be available on the funct
...
   - ([digit size option](#digit-size-options))
+- `trailingZeroDisplay`
+  - `auto` (default)
+  - `stripIfInteger`
+- `roundingPriority`
+  - `auto` (default)
+  - `morePrecision`
+  - `lessPrecision`
+- `roundingIncrement`
+  - 1 (default), 2, 5, 10, 20, 25, 50, 100, 200, 250, 500, 1000, 2000, 2500, and 5000
```
macchiati, line 166: For the future (not first release)

I know that this follows https://402.ecma-international.org/#:~:text=not%20the%20exponent.-,%5B%5BRoundingIncrement%5D%5D,-is%20an%20integer. 

That list is limited to powers of 10 from 1 to 4.

> evenly divides 10, 100, 1000, or 10000 into tenths, fifths, quarters, or halves

Given that currency values (due to hyperinflation) can range much higher than that, that seems like an arbitrary limit, and it should just be "a power of 10", and just limited by the implementation datatypes (as usual).

spec/registry.md
```diff
@@ -381,6 +338,190 @@ together with the resolved options' values.
...
+with _options_ on the _expression_ taking priority over any option values of the _operand_.
+
+> For example, the _placeholder_ in this _message_:
+> ```
+> .input {$n :currency currency=USD trailingZeroDisplay=stripIfInteger}
+> {{{$n :currency currencySign=accounting}}}
+> ```
+> would be formatted with the resolved options
+> `{ currencySign: 'accounting', trailingZeroDisplay: 'stripIfInteger', currency: 'USD' }`.
+
```
macchiati, line 513: Question.

It appears reasonable to allow composition of currency with numeric values, eg:

> .input {$n :number }
> {{{$n :currency currencySign=accounting}}}

or

> .input {$n :integer }
> {{{$n :currency currencySign=accounting}}}

whereby any options on the numeric functions are ignored for the currency formatting.

Not a blocker, since this can be done later.

macchiati (approved):

spec/registry.md
aphillips: Yeah, I closed my eyes while importing this.

spec/registry.md
```diff
@@ -381,6 +338,190 @@ together with the resolved options' values.
...
+#### Resolved Value
+
+The _resolved value_ of an _expression_ with a `:currency` _function_
+contains an implementation-defined currency value
+of the _operand_ of the annotated _expression_,
+together with the resolved options' values.
+
+#### Selection
+
+The _function_ `:currency` performs selection as described in [Number Selection](#number-selection) below.
```
eemeli, line 523: While considering #933, I realised that plural selection on currency values depends on being able to introspect the fraction digits which are by default locale-dependent. The ICU FormattedNumber might do that automatically, but in other APIs the operation can get a bit trickier.

spec/registry.md
macchiati: Sorry, I fat-fingered. I had copied the example, and meant to delete the currencySign option, and instead deleted the currencyCode option!

I had also meant just composition one way, from a number to a currency amount, as you have it:

```
.input {$n :number useGrouping=never numberingSystem=Latn}
.local $c = {$n :currency currencyCode=INR}
{{This prints in rupees without lakhs (or other groupings) and with Latin digits: {$c}}}
```

I hadn't meant from a currency amount to a number. 
```
.input {$c :currency}
.local $n = {$c :number useGrouping=never numberingSystem=Latn}
{{This prints just the number part of a currency amount: {$n}}}
```

spec/registry.md
aphillips: It's not locale-dependent. It's **currency** dependent. This is why fraction digits are different in this function!

macchiati: +1

macchiati:
I was just reviewing **[the registry document](https://github.com/unicode-org/message-format-wg/blob/main/spec/registry.md)**, and noticed that we are a bit inconsistent about when we use valid vs well-formed (and whether we mention either of them). 

for example:
> valid [Unicode Number System Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeNumberSystemIdentifier) (default is locale-specific)
> well-formed [Unicode Currency Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeCurrencyIdentifier)
> valid identifier per [BCP175](https://www.rfc-editor.org/rfc/rfc6557)
...

But the following note applies to other cases of "growing" identifiers (timezones, etc)

> The currency is only required to be well-formed rather than checked for validity. This allows new currency codes to be defined (there are many recent examples of this occuring). It also avoids requiring implementations to check currency codes for validity, although implementations are permitted to emit Bad Option or Bad Operand for invalid codes.

I jotted down principles in https://github.com/unicode-org/message-format-wg/issues/935 for a discussion on a more consistent approach

I don't want to derail our push to the end, so we could postpone this consistency work.

PR #917, opened 2024-10-27
Fix tests for bidirectional isolation

eemeli:
The current tests do not account for the [bidirectional isolation](https://github.com/unicode-org/message-format-wg/blob/main/spec/formatting.md#handling-bidirectional-text) of the formatted results. This adds an option `bidiIsolation` to the tests, with values `'default'` and `'none'` corresponding to the _Default BiDi Strategy_ and to a strategy of not applying any isolation.

Most of the existing tests follow the `'none'` strategy, and have been left unchanged except for noting this state of affairs. The `bidi.json` tests have been updated to use the `'default'` strategy, as that seemed appropriate.

A new `type: 'bidiIsolation'` part is added, and included in the `u-options.json` `expParts` results as appropriate; some bugs in those tests have also been addressed.

### DIFF
--- a/test/schemas/v0/tests.schema.json
+++ b/test/schemas/v0/tests.schema.json
@@ -118,6 +118,9 @@
         "src": {
           "$ref": "#/$defs/src"
         },
+        "bidiIsolation": {
+          "$ref": "#/$defs/bidiIsolation"
+        },
         "params": {
           "$ref": "#/$defs/params"
         },
@@ -146,6 +149,9 @@
         "src": {
           "$ref": "#/$defs/src"
         },
+        "bidiIsolation": {
+          "$ref": "#/$defs/bidiIsolation"
+        },
         "params": {
           "$ref": "#/$defs/params"
         },
@@ -171,6 +177,10 @@
     "src": {
       "description": "The MF2 syntax source.",
       "type": "string"
+    },
+    "bidiIsolation": {
+      "description": "The bidi isolation strategy.",
+      "enum": ["default", "none"]
     },
     "params": {
       "description": "Parameters to pass in to the formatter for resolving external variables.",
@@ -240,6 +250,23 @@
               },
               "value": {
                 "type": "string"
+              }
+            }
+          },
+          {
+            "description": "Bidi isolation part.",
+            "type": "object",
+            "additionalProperties": false,
+            "required": [
+              "type",
+              "value"
+            ],
+            "properties": {
+              "type": {
+                "const": "bidiIsolation"
+              },
+              "value": {
+                "enum": ["\u2066", "\u2067", "\u2068", "\u2069"]
               }
             }
           },
--- a/test/tests/bidi.json
+++ b/test/tests/bidi.json
@@ -2,6 +2,7 @@
   "scenario": "Bidi support",
   "description": "Tests for correct parsing of messages with bidirectional marks and isolates",
   "defaultTestProperties": {
+    "bidiIsolation": "default",
     "locale": "en-US"
   },
   "tests": [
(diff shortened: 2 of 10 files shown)


test/tests/u-options.json
```diff
@@ -120,7 +120,7 @@
     {
       "locale": "ar",
       "src": "أهلاً {بالعالم :string}",
-      "exp": "أهلاً \u2067بالعالم\u2069"
+      "exp": "أهلاً \u2068بالعالم\u2069"
     }
```
eemeli, line 124: We may need to drop this test, or more explicitly define whether `:string` applies introspection to its value to determine its directionality. As we don't forbid introspection, an implementation could use either an RLI or FSI here.

Presuming that the introspection would follow UAX#9, the user's perception of the results would be the same, so my preference would be to make the character-level output predictable by requiring that no introspection is done in the absence of a `u:dir` override or an operand that includes an explicit direction.

aphillips (approved):
I puzzled with it for a while but think this is right. The prevalence of FSI/PDI wrappers even in a purely LTR context makes me itch.

test/tests/bidi.json
```diff
@@ -114,7 +115,7 @@
     {
         "description":  " name... excludes U+FFFD and U+061C -- this pases as name -> [bidi] name-start *name-char",
         "src": ".local $\u061Cfoo = {1} {{ {$\u061Cfoo} }}",
-        "exp": " 1 "
+        "exp": " \u20681\u2069 "
```
catamorphism, line 118: I guess in this case, the direction of the message is determined to be `unknown`? Which is a little surprising to me, since the only bidi control character appears in one of the variable names, which I would expect not to have an effect on the formatted result of the message. Maybe what's throwing me off is that the explanation of the "Default Bidi Strategy" at the end of [this section](https://github.com/unicode-org/message-format-wg/blob/main/spec/formatting.md#handling-bidirectional-text) doesn't say how the message directionality is determined?

eemeli: The bidi controls in the syntax source don't matter here at all, and the message has LTR direction because of the `en-US` locale. It's the placeholder's direction that is unknown. For it to be known, we'd need to introspect the `1` string with something like [UAX#9 P2](https://unicode.org/reports/tr9/#P2), which will ultimately be done by the rendering engine, and even if we do, the direction would still be unknown because European numbers are only weakly directional. And so this is appropriately reflected by FSI/PDI isolating it.

This is much the same thing as I mention in https://github.com/unicode-org/message-format-wg/pull/917#discussion_r1818124771, but for LTR instead of RTL content.

> Maybe what's throwing me off is that the explanation of the "Default Bidi Strategy" at the end of [this section](https://github.com/unicode-org/message-format-wg/blob/main/spec/formatting.md#handling-bidirectional-text) doesn't say how the message directionality is determined?

Good point. We may want to add a mention under [Formatting Context](https://github.com/unicode-org/message-format-wg/blob/main/spec/formatting.md#formatting-context) about deducing the message direction from the locale if it's not explicitly given.

aphillips: If not explicitly provided, the direction of the message should be computed from the locale. As @eemeli notes, controls in the source don't matter. As noted elsewhere, you can't just look at the rendered placeholder to determine its desired direction--even strongly directional characters do not necessarily provide an accurate signal.

test/schemas/v0/tests.schema.json
```diff
@@ -118,6 +118,9 @@
         "src": {
           "$ref": "#/$defs/src"
         },
+        "bidiIsolation": {
```
catamorphism, line 121: https://github.com/unicode-org/conformance/blob/main/schema/message_fmt2/testgen_schema.json will need to be updated to reflect this change.

catamorphism (approved):

PR #921, opened 2024-10-28
Clean up `quoted-pattern` definition

lucacasonato:
`complex-body` was already always prefixed by optional space, so in that case there was effectively two subsequent optional spaces: that is unnecessary. It is much cleaner to remove the optional leading space from the `quoted-pattern` definition, and instead move it to explicitly before `quoted-pattern` in  `variant`.

This change is purely editorial and has no normative impact.

### DIFF
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -12,12 +12,12 @@ complex-body      = quoted-pattern / matcher
 input-declaration = input o variable-expression
 local-declaration = local s variable o "=" o expression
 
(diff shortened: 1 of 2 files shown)


aphillips:
This change looks like a good one, but the ABNF has to be changed to match. Can you add the change there also?

eemeli (approved):
Approving, provided that `message.abnf` is updated as well.

lucacasonato:
Whoops - absolutely

aphillips (approved):

PR #922, opened 2024-10-30
Implement `:unit` as Proposed RECOMMENDED in the registry

aphillips:
This introduces `:unit` as an optional function, per discussion elsewhere.

### DIFF
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -384,7 +384,6 @@ its _resolved value_ contains the implementation-defined integer value
 of the _operand_ of the annotated _expression_,
 together with the resolved options' values.
 
-====
 
 ### `:unit` function
 
@@ -424,7 +423,7 @@ A [Number Operand](#number-operands) without a `unit` _option_ results in a _Bad
 > For example, such an implementation might define a "unit operand"
 > to include a key-value structure with specific keys to be the
 > local unit operand, which might look like the following:
-> ```json
+> ```
 > {
 >    "value": 123.45,
 >    "unit": "kilometer",
@@ -451,6 +450,10 @@ The following options and their values are required to be available on the funct
 - `perUnit`
   - valid [Unit Identifier](https://www.unicode.org/reports/tr35/tr35-general.html#unit-identifiers)
     (no default)
+- `unitDisplay`
+  - `short` (default)
+  - `narrow`
+  - `long`
 - `compactDisplay` (this option only has meaning when combined with the option `notation=compact`)
    - `short` (default)
    - `long`
@@ -460,6 +463,12 @@ The following options and their values are required to be available on the funct
 - `numberingSystem`
    - valid [Unicode Number System Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeNumberSystemIdentifier)
      (default is locale-specific)
+- `signDisplay`
+   -  `auto` (default)
+   -  `always`
+   -  `exceptZero`
+   -  `negative`
+   -  `never`
 - `useGrouping`
   - `auto` (default)
   - `always`
@@ -524,7 +533,6 @@ together with the resolved options' values.
 > (for example, trying to convert meters to gallons produced a _Bad Option_)
 > nor will all implementations support conversion.
 
-====
 
 ### Number Operands
 
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -561,13 +561,10 @@ The _function_ `:currency` performs selection as described in [Number Selection]
 
 ### The `:unit` function
 
-The _function_ `:unit` is a selector and formatter for unitized values,
-that is, numeric values associated with a unit of measurement.
+The _function_ `:unit` is **Proposed** for inclusion in the next release of this specification but has not yet been finalized.
+The _function_ `:unit` is proposed to be a **RECOMMENDED** selector and formatter for unitized values,
+that is, for numeric values associated with a unit of measurement.
 This is a specialized form of numeric selection and formatting.
-
-> [!IMPORTANT]
-> Implementation of this function is **_OPTIONAL_**.
-> Any implementation of this function is strongly encouraged to follow this specification.
 
 #### Operands
 
@@ -622,10 +619,10 @@ The following options and their values are required to be available on the funct
    -  `exact`
 - `unit`
    - valid [Unit Identifier](https://www.unicode.org/reports/tr35/tr35-general.html#unit-identifiers)
-     (no default)
-- `usage` \[OPTIONAL\]
-    - Well-formed and valid usage identifiers are defined in [Unicode Preferences](https://www.unicode.org/reports/tr35/tr35-info.html#unit-preferences).
-    - (no default)
+     (no default, see [Unit Conversion](#unit-conversion) below)
+- `usage` \[RECOMMENDED\]
+    - valid [Unicode Unit Preference](https://www.unicode.org/reports/tr35/tr35-info.html#unit-preferences)
+      (no default)
 - `unitDisplay`
   - `short` (default)
   - `narrow`
@@ -691,12 +688,26 @@ with _options_ on the _expression_ taking priority over any _option_ values of t
 > would have the resolved options:
 > `{ unit: 'furlong', minimumFractionDigits: '2', minimumIntegerDigits: '1' }`.
 
-Some implementations support conversion to the locale's preferred units via the `usage` _option_.
+#### Resolved Value
+
+The _resolved value_ of an _expression_ with a `:unit` _function_
+consist of an implementation-defined unit value
+of the _operand_ of the annotated _expression_,
+together with the resolved _options_ and their resolved values.
+
+#### Selection
+
+The _function_ `:unit` performs selection as described in [Number Selection](#number-selection) below.
+
+#### Unit Conversion
+
+Implementations MAY support conversion to the locale's preferred units via the `usage` _option_.
 Implementing this _option_ is optional.
-Attempting to convert units produces an _Unsupported Operation_ error if such conversion is unsupported.
-It produces a _Bad Option_ error if the specified units are incompatible.
+Not all `usage` values are compatible with a given unit.
+Implementations SHOULD emit an _Unsupported Operation_ error if the requestion conversion is not supported.
+
 > For example, trying to convert a `length` unit such as meters
-> to a `volume` unit (such as "gallons") produces a _Bad Option_.
+> to a `volume` unit (such as "gallons") could produce an _Unsupported Operation_ error.
 
 Implementations MUST NOT substitute the unit without performing the associated conversion.
 
@@ -713,20 +724,7 @@ Implementations MUST NOT substitute the unit without performing the associated c
 > ```
 > This can produce "You have 405 feet to go."
 
-Not all `usage` values are compatible with a given unit.
-Implementations will produce a _Bad Option_ error for units 
-or combinations of units and usages that are not supported.
 
-#### Resolved Value
-
-The _resolved value_ of an _expression_ with a `:unit` _function_
-consist of an implementation-defined unit value
-of the _operand_ of the annotated _expression_,
-together with the resolved _options_ and their resolved values.
-
-#### Selection
-
-The _function_ `:unit` performs selection as described in [Number Selection](#number-selection) below.
 
 ### Number Operands
 
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -483,7 +483,8 @@ The following options and their values are required to be available on the funct
   - ([digit size option](#digit-size-options))
 - `maximumSignificantDigits`
   - ([digit size option](#digit-size-options))
-
+- `usage`
+    - Well-formed and valid usage identifiers are defined in [Unicode Preferences](https://www.unicode.org/reports/tr35/tr35-info.html#unit-preferences).
 If the _operand_ of the _expression_ is an implementation-defined type,
 such as the _resolved value_ of an _expression_ with a `:unit` _annotation_,
 it can include option values.
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -574,9 +574,9 @@ This is a specialized form of numeric selection and formatting.
 The _operand_ of the `:unit` function can be one of any number of
 implementation-defined types,
 each of which contains a numerical `value` plus a `unit`
-or it can be a [Number Operand](#number-operands), as long as the option
+or it can be a [Number Operand](#number-operands), as long as the _option_
 `unit` is provided.
-The option `unit` MAY be used to override the units of an implementation-defined type,
+The _option_ `unit` MAY be used to override the units of an implementation-defined type,
 provided the units are compatible and the implementation supports conversion.
 
 The value of the _operand_'s `unit` SHOULD be either a string containing a
@@ -679,9 +679,9 @@ The following options and their values are required to be available on the funct
 
 If the _operand_ of the _expression_ is an implementation-defined type,
 such as the _resolved value_ of an _expression_ with a `:unit` _annotation_,
-it can include option values.
-These are included in the resolved option values of the _expression_,
-with _options_ on the _expression_ taking priority over any option values of the _operand_.
+it can include _option_ values.
+These are included in the resolved _option_ values of the _expression_,
+with _options_ on the _expression_ taking priority over any _option_ values of the _operand_.
 
 > For example, the _placeholder_ in this _message_:
 > ```
@@ -693,9 +693,10 @@ with _options_ on the _expression_ taking priority over any option values of the
 
 Some implementations support conversion to the locale's preferred units via the `usage` _option_.
 Implementing this _option_ is optional.
-Attempting to convert units produces a _Bad Option_ error if such conversion is unsupported
-or if the specified units are incompatible.
-For example, trying to convert meters to a `volume` unit (such as "gallons") produces a _Bad Option_.
+Attempting to convert units produces an _Unsupported Operation_ error if such conversion is unsupported.
+It produces a _Bad Option_ error if the specified units are incompatible.
+> For example, trying to convert a `length` unit such as meters
+> to a `volume` unit (such as "gallons") produces a _Bad Option_.
 
 Implementations MUST NOT substitute the unit without performing the associated conversion.
 
@@ -712,7 +713,7 @@ Implementations MUST NOT substitute the unit without performing the associated c
 > ```
 > This can produce "You have 405 feet to go."
 
-Not all usages are compatible with the unit.
+Not all `usage` values are compatible with a given unit.
 Implementations will produce a _Bad Option_ error for units 
 or combinations of units and usages that are not supported.
 
@@ -721,7 +722,7 @@ or combinations of units and usages that are not supported.
 The _resolved value_ of an _expression_ with a `:unit` _function_
 consist of an implementation-defined unit value
 of the _operand_ of the annotated _expression_,
-together with the resolved options' values.
+together with the resolved _options_ and their resolved values.
 
 #### Selection
 
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -399,10 +399,10 @@ This is a specialized form of numeric selection and formatting.
 
 The _operand_ of the `:unit` function can be one of any number of
 implementation-defined types,
-each of which contains a numerical `value` plus a `unit` and optionally a `perUnit`.
-or it can be a [Number Operand](#number-operands), as long as at least the option
+each of which contains a numerical `value` plus a `unit`
+or it can be a [Number Operand](#number-operands), as long as the option
 `unit` is provided.
-The options `unit` and `perUnit` MAY be used to override the units of an implementation-defined type,
+The option `unit` MAY be used to override the units of an implementation-defined type,
 provided the units are compatible and the implementation supports conversion.
 
 The value of the _operand_'s `unit` SHOULD be either a string containing a
@@ -412,22 +412,21 @@ or an implementation-defined unit type.
 A [Number Operand](#number-operands) without a `unit` _option_ results in a _Bad Operand_ error.
 
 > [!NOTE]
-> For example, in ICU4J, the type `com.ibm.icu.util.Measure` can be used
-> to set the `value` and `unit` (and optionally the `perUnit`).
+> For example, in ICU4J, the type `com.ibm.icu.util.Measure` might be used
+> as an _operand_ for `:unit` because it contains the `value` and `unit`.
 
 > [!NOTE]
 > For runtime environments that do not provide a ready-made data structure,
 > class, or type for unit values, the implementation ought to provide
 > a data structure, convenience function, or documentation on how to encode
-> the value, unit, and optionally per-unit for formatting.
+> the value and unit for formatting.
 > For example, such an implementation might define a "unit operand"
 > to include a key-value structure with specific keys to be the
 > local unit operand, which might look like the following:
 > ```
 > {
 >    "value": 123.45,
->    "unit": "kilometer",
->    "perUnit": "hour"
+>    "unit": "kilometer-per-hour"
 > }
 > ```
 
@@ -447,8 +446,8 @@ The following options and their values are required to be available on the funct
 - `unit`
    - valid [Unit Identifier](https://www.unicode.org/reports/tr35/tr35-general.html#unit-identifiers)
      (no default)
-- `perUnit`
-  - valid [Unit Identifier](https://www.unicode.org/reports/tr35/tr35-general.html#unit-identifiers)
+- `usage`
+  - valid usage identifier (TBD)
     (no default)
 - `unitDisplay`
   - `short` (default)
@@ -500,7 +499,7 @@ with _options_ on the _expression_ taking priority over any option values of the
 > `{ unit: 'furlong', minimumFractionDigits: '2', minimumIntegerDigits: '1' }`.
 
 Some implementations support conversion between compatible units.
-Attempting to convert units (by specifying the `unit` option)
+Attempting to convert units (by specifying the `usage` option)
 produces a _Bad Option_ error if conversion is unsupported
 or if the specified units are incompatible.
 For example, trying to convert meters to gallons produces a _Bad Option_.
@@ -520,9 +519,9 @@ Implementations MUST NOT substitute the unit without performing the associated c
 > ```
 > This can produce "You have 405 feet to go."
 
-Not all per-units are compatible with the primary unit.
+Not all usages are compatible with the unit.
 Implementations will produce a _Bad Option_ error for units 
-or combinations of units and per-units that are not supported.
+or combinations of units and usages that are not supported.
 
 #### Selection
 
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -417,7 +417,7 @@ A [Number Operand](#number-operands) without a `unit` _option_ results in a _Bad
 
 > [!NOTE]
 > For runtime environments that do not provide a ready-made data structure,
-> class, or type for currency values, the implementation ought to provide
+> class, or type for unit values, the implementation ought to provide
 > a data structure, convenience function, or documentation on how to encode
 > the value, unit, and optionally per-unit for formatting.
 > For example, such an implementation might define a "unit operand"
@@ -439,7 +439,7 @@ In general, the default values for such options depend on the locale,
 the unit,
 the value of other options, or all of these.
 
-The following options and their values are required to be available on the function `:currency`:
+The following options and their values are required to be available on the function `:unit`:
 - `select`
    -  `plural` (default)
    -  `ordinal`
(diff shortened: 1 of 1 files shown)


spec/registry.md
```diff
@@ -384,6 +384,156 @@ its _resolved value_ contains the implementation-defined integer value
...
+> {
+>    "value": 123.5,
+>    "unit": "meter"
+> }
+> ```
+> The following _message_ might convert the formatted result to U.S. customary units:
+> ```
+> You have {$v :unit unit=foot} to go.
+> ```
+> Care has to be excersized with this type of operation.
```
catamorphism, line 531:
```suggestion
> Care has to be exercised with this type of operation.
```

spec/registry.md
```diff
@@ -384,6 +384,156 @@ its _resolved value_ contains the implementation-defined integer value
...
+#### Composition
+
+When an _operand_ or an _option_ value uses a _variable_ annotated,
+directly or indirectly, by a `:unit` _annotation_,
+its _resolved value_ contains an implementation-defined unit value
+of the _operand_ of the annotated _expression_,
+together with the resolved options' values.
+
+> [!NOTE]
+> Some implementations support conversation between compatible units.
```
catamorphism, line 519:
```suggestion
> Some implementations support conversion between compatible units.
```

spec/registry.md
```diff
@@ -384,6 +384,156 @@ its _resolved value_ contains the implementation-defined integer value
...
+> For example, consider the value:
+> ```
+> {
+>    "value": 123.5,
+>    "unit": "meter"
+> }
+> ```
+> The following _message_ might convert the formatted result to U.S. customary units:
+> ```
+> You have {$v :unit unit=foot} to go.
```
catamorphism, line 529: It seems like a pretty big footgun to me for some implementations to format this as "123.5 feet" while others format it as "405.2 feet".

aphillips: Yes, but not really what we're saying. It either formats as `405.2 feet` or emits _Bad Option_.

aphillips: I fixed up this text a bunch, made it normative, and moved it to the options section. Have a look.

catamorphism (approved):

spec/registry.md
```diff
@@ -384,6 +384,158 @@ its _resolved value_ contains the implementation-defined integer value
...
+
+> [!IMPORTANT]
+> Implementation of this function is **_OPTIONAL_**.
+> Any implementation of this function is strongly encouraged to follow this specification.
+
+#### Operands
+
+The _operand_ of the `:unit` function can be one of any number of
+implementation-defined types,
+each of which contains a numerical `value` plus a `unit` and optionally a `perUnit`.
```
macchiati, line 402: I don't understand this sentence. A perunit is a unit. 'unit' includes compounds, so:

kilowatt-hour-per-tonne is a unit, as is

- kilowatt-per-tonne
- per-tonne
- per-kilowatt-hour
- meter-watt-kilogram-per-square-second.

macchiati (changes requested):

spec/registry.md
```diff
@@ -384,6 +384,158 @@ its _resolved value_ contains the implementation-defined integer value
...
+  - ([digit size option](#digit-size-options), default: `1`)
+- `minimumFractionDigits`
+  - ([digit size option](#digit-size-options))
+- `maximumFractionDigits`
+  - ([digit size option](#digit-size-options))
+- `minimumSignificantDigits`
+  - ([digit size option](#digit-size-options))
+- `maximumSignificantDigits`
+  - ([digit size option](#digit-size-options))
+
```
macchiati, line 487: This also needs the **usage** setting. See https://unicode-org.github.io/icu-docs/apidoc/released/icu4j/com/ibm/icu/number/NumberFormatterSettings.html

That is what is used to signal that the unit **should be** converted to the local units if possible. Without it, the unit should **not** be converted.

spec/registry.md
```diff
@@ -384,6 +384,158 @@ its _resolved value_ contains the implementation-defined integer value
...
+> For example, the _placeholder_ in this _message_:
+> ```
+> .input {$n :unit unit=furlong minimumFractionDigits=2}
+> {{{$n :unit minimumIntegerDigits=1}}}
+> ```
+> would have the resolved options:
+> `{ unit: 'furlong', minimumFractionDigits: '2', minimumIntegerDigits: '1' }`.
+
+Some implementations support conversion between compatible units.
+Attempting to convert units (by specifying the `unit` option)
```
macchiati, line 503:
```suggestion
Attempting to convert units (by specifying the `usage` option)
```

spec/registry.md
aphillips: I'm mirroring what `MeasureFormat` currently does, which allows for a `perUnit` (so that one can modify a simple unit with a per-unit, like turning "furlongs" into "furlongs per fortnight". I can take this out and we can rely on the per-unit units already defined. It would make things _much_ simpler.

aphillips: For reference: https://unicode-org.github.io/icu-docs/apidoc/released/icu4j/com/ibm/icu/text/MeasureFormat.html#formatMeasurePerUnit-com.ibm.icu.util.Measure-com.ibm.icu.util.MeasureUnit-java.lang.StringBuilder-java.text.FieldPosition-

spec/registry.md
macchiati:
```suggestion
- `usage`
```

macchiati (changes requested):

spec/registry.md
macchiati: I checked and that is a convenience function. So instead of saying unit='per-tonne', you can say perUnit='tonne'. So if you also want to include it as a 'convenience option' that's fine. I would just define it using that equivalence.

aphillips: No, we're at the minimum requirements level here. I'll let @mihnita define that for ICU (if he wants to expose it) and will remove all the per-unit foolishness.

spec/registry.md
aphillips: Where are the valid values defined? Should I define a grammar for these (appears to be hyphen separated ASCII tokens??)?

I haven't worked with units in NumberFormatter. My experience is limited to MeasureFormat. I have a sense that I should push this to post-46.1 and we start with design (preferably from someone on the implementation side) rather than forcing it here.

macchiati: The grammar is defined in https://www.unicode.org/reports/tr35/tr35-general.html#unit-identifiers, with the EBNF at https://www.unicode.org/reports/tr35/tr35-general.html#syntax

aphillips: Those are unit identifiers. Where is the definition for **_usage_** identifiers?

macchiati: Sorry, the valid usage identifiers are `category` attribute values of the element `unitPreferences`, which are in the file [`supplemental/units.xml`](https://github.com/unicode-org/cldr/blob/main/common/supplemental/units.xml).

For example, `blood-glucose` because it is in:
```
<unitPreferences category="concentration" usage="blood-glucose">
  <unitPreference regions="AG AI AO AU BA BG BH BM BN BW BY CA CH CM CN CZ DK DM EE FI FJ GB GD HK HR HU IE IM IS KE KN KW KZ LC LI LT LU LV ME MG MK MO MS MT MU MY MZ NA NL NO NZ OM PG RS RU SE SG SI SK TC TO UA UG VC VG VN VU ZA">millimole-per-liter</unitPreference>
  <unitPreference regions="001">milligram-ofglucose-per-deciliter</unitPreference>
</unitPreferences>
```

Note: there is a ticket for creating a regular validity/ data file for the usage values. It would make it cleaner to reference that, so I'll look at accelerating that into v46.1 so that MF2.0 in 46.1 can use a more direct reference.

spec/registry.md
macchiati: sounds good!

spec/registry.md
macchiati:
```suggestion
- `usage`
    - Well-formed and valid usage identifiers are defined in [Unicode Preferences](https://www.unicode.org/reports/tr35/tr35-info.html#unit-preferences).
```

I have a PR to make the target of the link clearer, in https://github.com/unicode-org/cldr/pull/4179. I don't foresee any problem merging that into 46.1.

eemeli (changes requested):
As with `:currency currencyDisplay`, I am not comfortable with introducing functionality to the core function set that is not already available in the JS `Intl` formatters. Here this means that I think we should:
1. leave out unit conversion,
2. leave out `usage`, and
3. not require support for the full list of unit identifiers.

I'm quite open to discussing these aspects as iterations on top of initial proposals for `:unit` and `:currency`, but I find their inclusion to be a significant divergence from what we've done so far. As we want to conclude work on the initial release within a few weeks, we should aim for a smaller `:unit` that we can all agree to. We can and should improve the functions later, rather than trying to cram in all the features right away.

The basis for my position here is that we are defining a _message_ formatter, and that the original impetus for this work came from TC39 TG2, under which this group was initially organised. We should therefore be able to deliver a message formatter that does not impose additional requirements on _other_ formatters, such as unit or currency formatters, or introduce a deeper dependency on CLDR data than is minimally required.

I am aware that `:unit` is here proposed as optional, and that unit conversion is even more optional, but they're still all part of the core function set.

Otherwise, this proposal looks quite good.

macchiati:
> As with `:currency currencyDisplay`, I am not comfortable with introducing functionality to the core function set that is not already available in the JS `Intl` formatters. Here this means that I think we should:
> 
> 1. leave out unit conversion,
> 2. leave out `usage`, and
> 3. not require support for the full list of unit identifiers.
> 
> I'm quite open to discussing these aspects as iterations on top of initial proposals for `:unit` and `:currency`, but I find their inclusion to be a significant divergence from what we've done so far. As we want to conclude work on the initial release within a few weeks, we should aim for a smaller `:unit` that we can all agree to. We can and should improve the functions later, rather than trying to cram in all the features right away.
> 
> The basis for my position here is that we are defining a _message_ formatter, and that the original impetus for this work came from TC39 TG2, under which this group was initially organised. We should therefore be able to deliver a message formatter that does not impose additional requirements on _other_ formatters, such as unit or currency formatters, or introduce a deeper dependency on CLDR data than is minimally required.

While the original impetus was TC39, we need to make sure that it meets the needs of the broader community of implementers. The fact that it is optional means that it is not required of conformant implementations; thus a 'deeper dependency on CLDR' is also not required.

`usage` is a key requirement, because it allows implementers to format units in a customary way for a target locale, without having to investigate what all the customary units are for the target locale (country+region) and special-case each one themselves — just as number formatting doesn't force implementers to investigate what all the decimal/grouping separators and digits are needed for all the target locales. `units` is not internationalized (the main point of message formatting) without including that.

On the other hand, we should talk about your #3.

> 
> I am aware that `:unit` is here proposed as optional, and that unit conversion is even more optional, but they're still all part of the core function set.
> 
> Otherwise, this proposal looks quite good.

aphillips:
(as contributor)

I agree with @duerst that conversion is tricky. I think the `usage` stuff is supposed to make it "skeleton-like" in that locale data would be used to perform appropriate conversions without having to hardcode the mapping into the message. However, it strikes me as a complex mechanism that is implemented in ICU but not generally. I think users will find it more difficult to use, at least initially. But I see the value in providing the syntax and encouraging this part of CLDR to be adopted.

Elsewhere in the commentary with @macchiati, I note that I started from an API that I'm more personally familiar with and that said API in question has some quirks in it. Having looked into the more modern API, I was able to produce this version, but I'm reluctant to commit it. I think we should sideline this work until post-46.1 (focus on finishing standard parts of the spec--we have just a few weeks remaining in our schedule) and ensure that what we do produce is _fully_ baked. It could easily serve as an early instance of API function set management (to proof the process).

mihnita:
The `usage` stuff is a mixture of several things.
Part as a skeleton, sometimes determines the measurement unit.

If (for example) the usage is person height then the exact same measure amount (let's say 175 cm) would result in 5'9" for US (unit), or  "1.75m" or "175cm" or "1m 75cm" (skeleton-like).

Similar for temperature, if region is Canada the usage will make the result be Celsius for weather by Fahrenheit for cooking (unit).

---

Units are more complicated than they seem.

I think that leaving them post-46.1 would be a good idea.
Take them on only if we finish everything else :-)

macchiati:
I think adding units without adding usage is a mistake.

aphillips:
We agreed in the 2024-11-04 call to take this in 46.1 if-and-only-if it is ready for 46.1 and there is nothing else left to do. I made the changes discussed in telecon, notably removing `select=ordinal`.

sffc (commented):
This seems approximately aligned with my expectation. It should be conformant for an implementation to support only a subset of the units or usages. I defer on whether to require "usage" before it is available in ECMAScript.

aphillips:
@sffc You're right. I should have made the option `usage` optional.

macchiati (approved):

macchiati (approved):
Looks

macchiati (approved):

spec/registry.md
```diff
@@ -559,6 +559,175 @@ together with the resolved options' values.
...
+> The option `select` does not accept the value `ordinal` because selecting
+> unit values using ordinal rules makes no sense.
+
+The following options and their values are required to be available on the function `:unit`:
+- `select`
+   -  `plural` (default)
+   -  `exact`
+- `unit`
+   - valid [Unit Identifier](https://www.unicode.org/reports/tr35/tr35-general.html#unit-identifiers)
+     (no default)
```
eemeli, line 625: I continue to be uncomfortable with an effective requirement to emit Unsupported Operation errors for this one option instead of Bad Option, for values that are not supported by the implementation. That feels like a wart.

spec/registry.md
```diff
@@ -559,6 +559,175 @@ together with the resolved options' values.
...
+> .input {$n :unit unit=furlong minimumFractionDigits=2}
+> {{{$n :unit minimumIntegerDigits=1}}}
+> ```
+> would have the resolved options:
+> `{ unit: 'furlong', minimumFractionDigits: '2', minimumIntegerDigits: '1' }`.
+
+Some implementations support conversion to the locale's preferred units via the `usage` _option_.
+Implementing this _option_ is optional.
+Attempting to convert units produces an _Unsupported Operation_ error if such conversion is unsupported.
+It produces a _Bad Option_ error if the specified units are incompatible.
```
eemeli, line 697: Does this not conflict with our prior text about accepting the function and its options? https://github.com/unicode-org/message-format-wg/blob/f3344b5cea59e46aa8a0aa150904f53eb7690545/spec/registry.md?plain=1#L9-L13

eemeli (changes requested):
Apologies for leaving this until so late, but I simply hadn't had time for this PR. I've left a bunch of concerns inline, and I don't think this is ready yet.

sffc (commented):
The flow of options seems okay to me. I defer to others on error handling, etc.

macchiati:
Note, while doing https://github.com/unicode-org/message-format-wg/pull/949 I realized that we say that all RECOMMENDED functions/options need a maturity status. It sounds like (from the WG discussion) that it would be **Proposed** for now.

spec/registry.md
```diff
@@ -559,6 +559,175 @@ together with the resolved options' values.
...
+> ```
+> .input {$n :unit unit=furlong minimumFractionDigits=2}
+> {{{$n :unit minimumIntegerDigits=1}}}
+> ```
+> would have the resolved options:
+> `{ unit: 'furlong', minimumFractionDigits: '2', minimumIntegerDigits: '1' }`.
+
+Some implementations support conversion to the locale's preferred units via the `usage` _option_.
+Implementing this _option_ is optional.
+Attempting to convert units produces an _Unsupported Operation_ error if such conversion is unsupported.
```
macchiati, line 696:
```suggestion
Attempting to convert units produces an _Bad Option_ error if such conversion is unsupported.
```

spec/registry.md
macchiati: The definitions of [your engagement is non-provisioned](https://github.com/unicode-org/message-format-wg/blob/main/spec/errors.md#bad-option) and https://github.com/unicode-org/message-format-wg/blob/main/spec/errors.md#bad-option overlap. Thus if something is in the intersection one could go either way. This needs to be cleaned up in the future, but I added a suggestion to change to Bad Option to resolve this.

spec/registry.md
```diff
@@ -559,6 +559,175 @@ together with the resolved options' values.
...
+The following options and their values are required to be available on the function `:unit`:
+- `select`
+   -  `plural` (default)
+   -  `exact`
+- `unit`
+   - valid [Unit Identifier](https://www.unicode.org/reports/tr35/tr35-general.html#unit-identifiers)
+     (no default)
+- `usage` \[OPTIONAL\]
+    - Well-formed and valid usage identifiers are defined in [Unicode Preferences](https://www.unicode.org/reports/tr35/tr35-info.html#unit-preferences).
+    - (no default)
```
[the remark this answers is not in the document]
macchiati: If you'd be happier with the same locution used around timezone identifiers, that sounds good to me.

spec/registry.md
macchiati: I agree. The default if the usage is unsupported or doesn't apply is to ignore it.

aphillips: I agree about the _Bad Option_.

What I'm trying to do with the _Unsupported Operation_ thing, though, is give users feedback when the conversion doesn't work. Probably that's "SHOULD emit..."

spec/registry.md
aphillips: Just ⬆️ up there we said no Bad Option.

spec/registry.md
```diff
@@ -559,6 +559,175 @@ together with the resolved options' values.
...
+> ```
+> would have the resolved options:
+> `{ unit: 'furlong', minimumFractionDigits: '2', minimumIntegerDigits: '1' }`.
+
+Some implementations support conversion to the locale's preferred units via the `usage` _option_.
+Implementing this _option_ is optional.
+Attempting to convert units produces an _Unsupported Operation_ error if such conversion is unsupported.
+It produces a _Bad Option_ error if the specified units are incompatible.
+> For example, trying to convert a `length` unit such as meters
+> to a `volume` unit (such as "gallons") produces a _Bad Option_.
```
macchiati, line 699: Remove this; a usage isn't a unit, so this can't happen.

```suggestion
```

spec/registry.md
```diff
@@ -559,6 +559,175 @@ together with the resolved options' values.
...
+> ```
+> The following _message_ might convert the formatted result to U.S. customary units:
+> ```
+> You have {$v :unit unit=foot maximumFractionDigits=0} to go.
+> ```
+> This can produce "You have 405 feet to go."
+
+Not all `usage` values are compatible with a given unit.
+Implementations will produce a _Bad Option_ error for units 
+or combinations of units and usages that are not supported.
```
macchiati, line 718: No, the default is simply to leave the unit alone; it isn't a failure in that case.

```suggestion
If the `usage` values does not apply to the given unit, then it is ignored — without producing 
a _Bad Option_ error.
```

spec/registry.md
```diff
@@ -559,6 +559,175 @@ together with the resolved options' values.
...
+
+The following options and their values are required to be available on the function `:unit`:
+- `select`
+   -  `plural` (default)
+   -  `exact`
+- `unit`
+   - valid [Unit Identifier](https://www.unicode.org/reports/tr35/tr35-general.html#unit-identifiers)
+     (no default)
+- `usage` \[OPTIONAL\]
+    - Well-formed and valid usage identifiers are defined in [Unicode Preferences](https://www.unicode.org/reports/tr35/tr35-info.html#unit-preferences).
```
macchiati, line 627: I made this exactly parallel to timezone.
```suggestion
    (see [LDML](https://www.unicode.org/reports/tr35/tr35-info.html#unit-preferences)
    for information on identifiers)
```

spec/registry.md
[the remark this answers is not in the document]
macchiati: I made that suggestion.

macchiati (approved):
I made suggestions to address Eemeli's comments, and would approve with those changes.

spec/registry.md
```diff
@@ -559,6 +559,175 @@ together with the resolved options' values.
...
 The _function_ `:currency` performs selection as described in [Number Selection](#number-selection) below.
 
+### The `:unit` function
+
+The _function_ `:unit` is a selector and formatter for unitized values,
+that is, numeric values associated with a unit of measurement.
+This is a specialized form of numeric selection and formatting.
+
+> [!IMPORTANT]
+> Implementation of this function is **_OPTIONAL_**.
```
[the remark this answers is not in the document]
aphillips: We moved to REQUIRED/RECOMMENDED this past week. I implemented the change in the most recent push.

spec/registry.md
macchiati: Right!

spec/registry.md
aphillips: I addressed this by making _Unsupported Operation_ SHOULD for conversions that the implementation doesn't not handle, which might include irrational ones ("milliliters => furlongs"). I removed all other mention of Unsupported or Bad Option to allow the default stuff for functions to cover it.

macchiati (commented):
Small touchup

spec/registry.md
```diff
@@ -559,6 +559,173 @@ together with the resolved options' values.
...
+#### Unit Conversion
+
+Implementations MAY support conversion to the locale's preferred units via the `usage` _option_.
+Implementing this _option_ is optional.
+Not all `usage` values are compatible with a given unit.
+Implementations SHOULD emit an _Unsupported Operation_ error if the requestion conversion is not supported.
+
+> For example, trying to convert a `length` unit such as meters
+> to a `volume` unit (such as "gallons") could produce an _Unsupported Operation_ error.
+
```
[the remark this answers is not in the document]
aphillips: Actually, this is the one I would want an error for (see my comment elsewhere). The addition of "has no effect" is good, although it's really a MUST. I'll try again here.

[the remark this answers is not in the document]
macchiati: I'm OK with MUST. 

As to the error, I'm ok with MAY issue an _Unsupported Operation_; not sure it should be required. 

Side note that the usage also may have no effect if the size of the unit and the locale indicate that.

macchiati (approved):

macchiati (approved):

spec/registry.md
```diff
@@ -559,6 +559,174 @@ together with the resolved options' values.
...
+The _resolved value_ of an _expression_ with a `:unit` _function_
+consist of an implementation-defined unit value
+of the _operand_ of the annotated _expression_,
+together with the resolved _options_ and their resolved values.
+
+#### Selection
+
+The _function_ `:unit` performs selection as described in [Number Selection](#number-selection) below.
+
+#### Unit Conversion
```
eemeli, line 702: According to the text above, unit conversion is also possible via the `unit` option. That should be mentioned here beyond its errant use in the example below, possibly along with some description of what happens if you define both `unit` and `usage`.

eemeli (approved):
This is fine to land as proposed, because that does not imply any stability.

spec/registry.md
```diff
@@ -559,6 +559,174 @@ together with the resolved options' values.
...
+#### Resolved Value
+
+The _resolved value_ of an _expression_ with a `:unit` _function_
+consist of an implementation-defined unit value
+of the _operand_ of the annotated _expression_,
+together with the resolved _options_ and their resolved values.
+
+#### Selection
+
+The _function_ `:unit` performs selection as described in [Number Selection](#number-selection) below.
```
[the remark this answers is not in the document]
macchiati: Take 2 cases:

1. If I format {{unit=kilogram, operand=30}, usage=person, maximumFractionDigits=0} with {u:locale=|en-US|} I'll get "66 pounds".
2. If I format {{unit=pound, operand=66.1387}, maximumFractionDigits=0} I'll also get "66 pounds".

So what is happening in **both cases** is that the numeric value for selection is equivalent to what the formatted value has. And that is what needs to be selected on, namely 66.

The wording we have for that in [selection as described in](https://github.com/unicode-org/message-format-wg/blob/main/spec/registry.md#number-selection)

> where resolvedSelector is the _resolved value_ of a _selector_

So in both cases, the _resolved value_ has to be 66.

There are of course other cases: {{operand=66.1387}, maximumSignificantDigits=1} will format as "70" (In English), and "٧٠" in Arabic, and must match a selection key of 70.

So in this case, the resolved value has to be 70.

I think we haven't yet nailed 'resolved value', because the resolved value for selection and formatting ends up needing to be different than the resolved value for functional composition. But that we cannot resolve at this point in the release.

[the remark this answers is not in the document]
eemeli: I understand that what you describe is likely the intent here, but the _resolved value_ description talks of "an implementation-defined unit value of the _operand_ of the annotated _expression_", with no reference to any conversion possibly being made to it, and so it's reasonable to assume that e.g. your case 1 example would end up with the numerical value `30` being the value that selection is done on, and as the selection is the same as used for other numerical values, it doesn't know about needing to multiply the number by any such factor.

Note how in `:math`, the numeric value held in _resolved value_ is explicitly modified by the `add` and `subtract` option values. We need something like that here.

spec/registry.md
```diff
@@ -559,6 +559,174 @@ together with the resolved options' values.
...
+
+#### Operands
+
+The _operand_ of the `:unit` function can be one of any number of
+implementation-defined types,
+each of which contains a numerical `value` plus a `unit`
+or it can be a [Number Operand](#number-operands), as long as the _option_
+`unit` is provided.
+The _option_ `unit` MAY be used to override the units of an implementation-defined type,
+provided the units are compatible and the implementation supports conversion.
```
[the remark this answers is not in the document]
macchiati: Hmmm. I missed this.

I have never thought of 'unit=x' as involving or requiring conversion, but rather simply using the numeric value of the operand with the specified unit x. I think that you are right; this should take a bit of rewording. Perhaps:

```suggestion
The _option_ `unit` MAY be used to override the unit value of an implementation-defined type,
if the implementation supports unit conversion. 
However, if the units cannot be converted, then the unit value is not overridden and a Bad Option error must be emitted.
```

**Side issue**

Frankly, currency is even worse. Suppose we have input of a implementation-defined parameter $n = {value=300, currency=EUR} and we format or select {$n :currency currency=JPY}. That is really nasty, displaying 300¥. However, there is no stable conversion possible.

So for :currency, we should make it really clear that if there is an implementation-defined operand that contains a currency value, the currency option is invalid, no matter its value.

spec/registry.md
macchiati: Agreed.

macchiati (approved):
Since Eemeli is ok with this, I suggest we merge it, and then have a followup PR where we can discuss touchups.

spec/registry.md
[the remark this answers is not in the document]
aphillips: Currency explicitly says that this is an error.

The unit example is a leftover from early on. I will make it use `usage`. We should do the same thing with the `unit` option that we do with the `currency` option: you can use it to create a unit value with a number operand. Otherwise it is an error.

spec/registry.md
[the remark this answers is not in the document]
aphillips: If we permitted `unit` to convert, I agree that the resolved value should be modified. But I think we should do like `:currency` and only allow `unit` to compose with a numeric operand.

spec/registry.md
```diff
@@ -559,6 +559,174 @@ together with the resolved options' values.
...
+  - `trunc`
+  - `halfCeil`
+  - `halfFloor`
+  - `halfExpand` (default)
+  - `halfTrunc`
+  - `halfEven`
+
+If the _operand_ of the _expression_ is an implementation-defined type,
+such as the _resolved value_ of an _expression_ with a `:unit` _annotation_,
+it can include _option_ values.
```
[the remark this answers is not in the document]
aphillips: I'm not sure I understand. I don't think this needs to indicate that some types are not supported as operands.

This text is not about the _operand_. It's about the handling of _option_ values within a _resolved value_. It is nearly boilerplate, since most of our functions have a variation of this sentence. In practice, it might be better to say something like the following, which doesn't care what the operand is or is not:

```suggestion
The _resolved value_ of an _expression_ with a `:unit` _annotation_
includes the result of _option resolution_.
```

Such a boilerplate would be altered if some of the options were required to be filtered by the function. `unit` is candidate for this in `:unit` (since the `unit` presumably is now part of the resolved operand).

spec/registry.md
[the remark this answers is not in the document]
macchiati: That works for me, and is safe and far simpler.

It would still be possible for a future version to allow that, but I think one would have to make a good case for doing so.

spec/registry.md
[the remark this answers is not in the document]
macchiati: Sounds great.

macchiati (approved):
Latest changes look good!

PR #925, opened 2024-11-06
Provide normative guidance on function/option/option value impl.

aphillips:
See #924 for the origin of this PR.

### DIFF
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -22,7 +22,7 @@ Implementations MUST _accept_ each **standard** _function_ and
 MUST _accept_ all _options_ defined as **standard** for those _functions_.
 
 Implementations SHOULD _accept_ each **optional** _function_. 
-For each such _function_, the implementation MUST accept all _options_ and _option_ values 
+For each such _function_, the implementation MUST accept all _options_
 listed as **standard** for that _function_.
 
 Implementations SHOULD _accept_ _options_ and _option_ values that are marked as **optional**.
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -6,7 +6,7 @@ additional functionality.
 
 To **_<dfn>accept</dfn>_** a function means that an implementation MUST NOT
 emit an _Unknown Function_ error for that _function_'s _identifier_.
-To _accept_ an option means that an implementation MUST NOT
+To _accept_ an option means that a _function handler_ MUST NOT
 emit a _Bad Option_ error for that _option_'s _identifier_ when used with the _function_
 it is defined for
 and MUST NOT emit a _Bad Option_ error for any of the _option_ values
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -12,7 +12,7 @@ it is defined for
 and MUST NOT emit a _Bad Option_ error for any of the _option_ values
 (both **standard** and **optional**)
 defined for that _option_.
-Accepting a _function_ or its _options_ does not mean that any given output is produced.
+Accepting a _function_ or its _options_ does not mean that a particular output is produced.
 
 _Functions_ can define _options_. 
 An _option_ can be **standard** or **optional**.
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -4,15 +4,29 @@ This section defines the **standard** _functions_ which are REQUIRED for conform
 along with **optional** _functions_ that SHOULD be implemented to support
 additional functionality.
 
-Implementations MUST define and support each **standard** _function_ and
-MUST accept all _options_ and _option_ values defined as **standard** listed in this specification.
+To **_<dfn>accept</dfn>_** a function means that an implementation MUST NOT
+emit an _Unknown Function_ error for that _function_'s _identifier_.
+To _accept_ an option means that an implementation MUST NOT
+emit a _Bad Option_ error for that _option_'s _identifier_ when used with the _function_
+it is defined for
+and MUST NOT emit a _Bad Option_ error for any of the _option_ values
+(both **standard** and **optional**)
+defined for that _option_.
+Accepting a _function_ or its _options_ does not mean that any given output is produced.
 
-When defined and supported by an implementation, each **optional** _function_ 
-MUST accept all _options_ and _option_ values listed as **standard** in this specification.
+_Functions_ can define _options_. 
+An _option_ can be **standard** or **optional**.
 
-Implementations SHOULD implement _options_ and _option_ values that are marked as **optional**.
+Implementations MUST _accept_ each **standard** _function_ and
+MUST _accept_ all _options_ and _option_ values defined as **standard** for those _functions_.
 
-Implementations MAY implement _functions_ not defined in this specification.
+Implementations SHOULD _accept_ each **optional** _function_. 
+For each such _function_, the implementation MUST accept all _options_ and _option_ values 
+listed as **standard** for that _function_.
+
+Implementations SHOULD _accept_ _options_ and _option_ values that are marked as **optional**.
+
+Implementations MAY _accept_ _functions_ not defined in this specification.
 Such functions MUST use an implementation-defined _namespace_.
 In addition, implementations SHOULD provide mechanisms for users to
 register and use user-defined _functions_ and their associated _functional handlers_.
@@ -20,7 +34,7 @@ register and use user-defined _functions_ and their associated _functional handl
 Implementations MAY implement additional _options_ not defined here.  
 Such _options_ MUST use an implementation-specific _namespace_.
 
-Implementations MAY implement additional _option_ values for _options_ defined here,
+Implementations MAY _accept_ additional _option_ values for _options_ defined here,
 although care needs to be exercised to ensure interoperability
 and to avoid collisions with future standardization.
 There is no namespace mechanism for _option_ values,
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -27,7 +27,7 @@ listed as **standard** for that _function_.
 Implementations SHOULD _accept_ _options_ and _option_ values that are marked as **optional**.
 
 Implementations MAY _accept_ _functions_ not defined in this specification.
-Such functions MUST use an implementation-defined _namespace_.
+Functions not defined by any version of this specification MUST use an implementation-defined _namespace_.
 In addition, implementations SHOULD provide mechanisms for users to
 register and use user-defined _functions_ and their associated _functional handlers_.
 
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -1,11 +1,34 @@
 # MessageFormat 2.0 Default Function Registry
 
-This section describes the functions for which each implementation MUST provide
-a _function handler_ to be conformant with this specification.
+This section defines the **standard** _functions_ which are REQUIRED for conformance with this specification,
+along with **optional** _functions_ that SHOULD be implemented to support
+additional functionality.
 
-Implementations MAY implement additional _functions_ or additional _options_.
-In particular, implementations are encouraged to provide feedback on proposed
-_options_ and their values.
+Implementations MUST define and support each **standard** _function_ and
+MUST accept all _options_ and _option_ values defined as **standard** listed in this specification.
+
+When defined and supported by an implementation, each **optional** _function_ 
+MUST accept all _options_ and _option_ values listed as **standard** in this specification.
+
+Implementations SHOULD implement _options_ and _option_ values that are marked as **optional**.
+
+Implementations MAY implement _functions_ not defined in this specification.
+Such functions MUST use an implementation-defined _namespace_.
+In addition, implementations are encouraged to provide mechanisms for users to
+register and use user-defined _functions_ and their associated _functional handlers_.
+
+Implementations MAY implement additional _options_ not defined here.  
+Such _options_ MUST use an implementation-specific _namespace_.
+
+Implementations MAY implement additional _option_ values for _options_ defined here,
+although care needs to be exercised to ensure interoperability
+and to avoid collisions with future standardization.
+There is no namespace mechanism for _option_ values,
+however, the [stability policy](#stability-policy) for this specification 
+permits only the ranges a-z, A-Z, and 0-9 in _option_ values.
+Implementation-defined values SHOULD use a distinguishing character 
+or character sequence, such as by prefixing with a `_` U+005F LOW LINE,
+to ensure that they don't collide with future standardization.
 
 > [!NOTE]
 > The [Stability Policy](/spec#stability-policy) allows for updates to
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -33,7 +33,8 @@ register and use user-defined _functions_ and their associated _functional handl
 Functions not defined by any version of this specification SHOULD use 
 an implementation-defined or user-defined _namespace_.
 
-Implementations MAY implement additional _options_ not defined here
+Implementations MAY implement additional _options_ not defined
+by any version of this specification
 for **standard** and **optional** functions.
 Such _options_ MUST use an implementation-specific _namespace_.
 
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -14,7 +14,7 @@ Implementations SHOULD implement _options_ and _option_ values that are marked a
 
 Implementations MAY implement _functions_ not defined in this specification.
 Such functions MUST use an implementation-defined _namespace_.
-In addition, implementations are encouraged to provide mechanisms for users to
+In addition, implementations SHOULD provide mechanisms for users to
 register and use user-defined _functions_ and their associated _functional handlers_.
 
 Implementations MAY implement additional _options_ not defined here.  
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -13,6 +13,8 @@ and MUST NOT emit a _Bad Option_ error for any of the _option_ values
 (both **standard** and **optional**)
 defined for that _option_.
 Accepting a _function_ or its _options_ does not mean that a particular output is produced.
+Implementations MAY emit an _Unsupported Operation_ error for _options_
+or _option_ values that they cannot support.
 
 _Functions_ can define _options_. 
 An _option_ can be **standard** or **optional**.
@@ -36,24 +38,20 @@ Implementations MAY implement additional _options_ not defined here
 for **standard** and **optional** functions.
 Such _options_ MUST use an implementation-specific _namespace_.
 
-Implementations MAY _accept_ additional _option_ values for _options_ defined here,
-although care needs to be exercised to ensure interoperability
-and to avoid collisions with future standardization.
-There is no namespace mechanism for _option_ values,
-however, the [stability policy](#stability-policy) for this specification 
-permits only the ranges a-z, A-Z, and 0-9 in _option_ values.
-Implementation-defined values SHOULD use a distinguishing character 
-or character sequence, such as by prefixing with a `_` U+005F LOW LINE,
-to ensure that they don't collide with future standardization.
+Implementations MAY _accept_ additional _option_ values for _options_ defined here.
+However, such values might become defined with a different meaning in the future,
+including with a different, incompatible name
+or using an incompatible value space.
+Supporting implementation-specific _option_ values for **standard** or **optional** functions is NOT RECOMMENDED.
 
-> [!NOTE]
-> The [Stability Policy](/spec#stability-policy) allows for updates to
-> Default Registry functions to add support for new options.
-> As implementations are permitted to ignore options that they do not support,
-> it is possible to write messages using options not defined below
-> which currently format with no error, but which could produce errors
-> when formatted with a later edition of the Default Registry.
-> Therefore, using options not explicitly defined here is NOT RECOMMENDED.
+Future versions of this specification MAY define additional _options_ and _option_ values,
+subject to the rules in the [Stability Policy](#stability-policy),
+for _functions_ found in this specification.
+As implementations are permitted to ignore _options_ that they do not support,
+it is possible to write _messages_ using _options_ not defined below
+which currently format with no error, but which could produce errors
+when formatted with a later edition of this specification.
+Therefore, using _options_ not explicitly defined here is NOT RECOMMENDED.
 
 ## String Value Selection and Formatting
 
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -31,7 +31,8 @@ Functions not defined by any version of this specification MUST use an implement
 In addition, implementations SHOULD provide mechanisms for users to
 register and use user-defined _functions_ and their associated _functional handlers_.
 
-Implementations MAY implement additional _options_ not defined here.  
+Implementations MAY implement additional _options_ not defined here
+for **standard** and **optional** functions.
 Such _options_ MUST use an implementation-specific _namespace_.
 
 Implementations MAY _accept_ additional _option_ values for _options_ defined here,
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -10,7 +10,6 @@ To _accept_ an _option_ means that a _function handler_ MUST NOT
 emit a _Bad Option_ error for that _option_'s _identifier_ when used with the _function_
 it is defined for
 and MUST NOT emit a _Bad Option_ error for any of the _option_ values
-(both **standard** and **optional**)
 defined for that _option_.
 Accepting a _function_ or its _options_ does not mean that a particular output is produced.
 Implementations MAY emit an _Unsupported Operation_ error for _options_
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -6,7 +6,7 @@ additional functionality.
 
 To **_<dfn>accept</dfn>_** a function means that an implementation MUST NOT
 emit an _Unknown Function_ error for that _function_'s _identifier_.
-To _accept_ an option means that a _function handler_ MUST NOT
+To _accept_ an _option_ means that a _function handler_ MUST NOT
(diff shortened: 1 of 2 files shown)


macchiati (approved):

spec/registry.md
```diff
@@ -1,11 +1,34 @@
...
-
-Implementations MAY implement additional _functions_ or additional _options_.
-In particular, implementations are encouraged to provide feedback on proposed
-_options_ and their values.
+This section defines the **standard** _functions_ which are REQUIRED for conformance with this specification,
+along with **optional** _functions_ that SHOULD be implemented to support
+additional functionality.
+
+Implementations MUST define and support each **standard** _function_ and
+MUST accept all _options_ and _option_ values defined as **standard** listed in this specification.
```
eemeli, line 8: We ought to define somewhere what "accept" means.

Are there some conditions where an implementation could emit an error and use a fallback value for a placeholder which uses any set of options that it "MUST accept", or is it required that any perceived errors or unsupported values are silently ignored?

spec/registry.md
```diff
@@ -1,11 +1,34 @@
...
+
+When defined and supported by an implementation, each **optional** _function_ 
+MUST accept all _options_ and _option_ values listed as **standard** in this specification.
+
+Implementations SHOULD implement _options_ and _option_ values that are marked as **optional**.
+
+Implementations MAY implement _functions_ not defined in this specification.
+Such functions MUST use an implementation-defined _namespace_.
```
eemeli, line 16: A "MUST" here would make it impossible for an implementation to update a part of the core function set when it changes.

For example, a later LDML release could add some minor option to an existing formatter, and at the same time introduce a new `:list` formatter. If the implementation cannot immediately provide a `:list` formatter, it could not add the minor option either if there's no flexibility on non-namespaced function and option names.

Also, we can't really test for this.

```suggestion
Such functions SHOULD use an implementation-defined _namespace_.
```

spec/registry.md
```diff
@@ -1,11 +1,34 @@
...
+In addition, implementations are encouraged to provide mechanisms for users to
+register and use user-defined _functions_ and their associated _functional handlers_.
```
eemeli, line 18: "are encouraged to" feels a bit too vague.

```suggestion
In addition, implementations SHOULD provide mechanisms for users to
register and use user-defined _functions_ and their associated _functional handlers_.
```

spec/registry.md
```diff
@@ -1,11 +1,34 @@
...
+
+Implementations MAY implement additional _options_ not defined here.  
+Such _options_ MUST use an implementation-specific _namespace_.
```
eemeli, line 21: Again, I think this is too strong and limiting.

```suggestion
Such _options_ SHOULD use an implementation-specific _namespace_.
```

spec/registry.md
```diff
@@ -1,11 +1,34 @@
...
+
+Implementations MAY implement additional _option_ values for _options_ defined here,
+although care needs to be exercised to ensure interoperability
+and to avoid collisions with future standardization.
+There is no namespace mechanism for _option_ values,
+however, the [stability policy](#stability-policy) for this specification 
+permits only the ranges a-z, A-Z, and 0-9 in _option_ values.
```
eemeli, line 28: This is not the case. The stability policy only says this:
> Updates to this specification will only reserve, define, or require function names or function option names consisting of characters in the ranges a-z, A-Z, and 0-9. All other names in these categories are reserved for the use of implementations or users.

eemeli (changes requested):

spec/registry.md
macchiati: I'm having trouble understanding your reasoning. It seems to be something like:

- My implementation adds :foo:list with option backwards=
- LDML 50 introduces a :list formatter with different options.
- My implementation can't conformantly upgrade to LDML 50 by aliasing :list to my :foo:list order to add the :list formatter, because backwards= is not standard and doesn't have a namespace.

Is that the scenario you have in mind? I just can't make out what it is.

spec/registry.md
macchiati: SHOULD is probably ok.

spec/registry.md
macchiati: as above

spec/registry.md
macchiati: Right, that is for function names and option names, not option values. I have a suggestion below.

spec/registry.md
```diff
@@ -1,11 +1,34 @@
...
+Implementation-defined values SHOULD use a distinguishing character 
+or character sequence, such as by prefixing with a `_` U+005F LOW LINE,
+to ensure that they don't collide with future standardization.
```
macchiati, line 31:
```suggestion
There is no namespace mechanism for _option_ values.
Implementation-defined _enum_ values SHOULD use a distinguishing character 
or character sequence, such as by prefixing with a `_` U+005F LOW LINE,
to ensure that they don't collide with future standardization.

For example, the option `compactDisplay` for `:currency` has option values `short` and `long`.
If an implementation adds a medium value, it should be distinguished, eg `_medium`.
```

eemeli: Why not recommend that these be `foo:` namespaced? As in, `compactDisplay=icu:medium`.

spec/registry.md
macchiati: Ok, I understand what you mean now.

NOTE I think in the future we probably want to introduce any new functions/options/option values as _optional_ in their first release, and only make them _required_ in a subsequent release. That does not solve the problem you raise, but reduces the frequency of occurrence.

The downside of your approach is that it opens the door to collisions, and functions that are 'pretending' to be standard but aren't. I would rather see an **explicit** carve-out for that case.

```suggestion
Such functions MUST use an implementation-defined _namespace_. 

There is one exception to this policy, where the function is conformant to a later version of LDML. For example, suppose LDML 50 adds a `:list` function; in that case, an implementation that is conformant to LDML 49 can add that `:list` function and its options, while remaining conformant to LDML 49. 
```
I'll make the similar suggestion below, for options.

spec/registry.md
macchiati:
```suggestion

Such options MUST use an implementation-defined _namespace_. 

There is one exception to this policy, where the option is conformant to a later version of LDML. For example, suppose LDML 50 adds a `medium` option to `:number compactDisplay`; in that case, an implementation that is conformant to LDML 49 can add that `medium` option while remaining conformant to LDML 49. 
```

spec/registry.md
aphillips: To me, "accept" means that no error should be emitted just because a given function/option/option value appears in a placeholder. We say "accept" to avoid requiring that any given behavior attach to the f/o/v in question. However, it probably should work in some meaningful, recognizable manner.

spec/registry.md
macchiati: I didn't think that we had namespaced option **values**.

spec/registry.md
aphillips: I get it, but recall that we're going to break the functions apart into separate specs (according to our design doc). I can get behind SHOULD here to avoid versioning issues, but perhaps we could still say something like:

```suggestion
Functions not defined by any version of this specification MUST use an implementation-defined _namespace_.
```

This is still untestable, of course.

spec/registry.md
aphillips: We don't have namespaces option values and the ABNF does not permit a colon in the unquoted value (there can be one in a literal, but the literal has to be quoted to have a colon).

aphillips: > Implementation-defined _enum_ values

`enum` isn't a defined term. I get that you're trying to avoid making people put underscores on values such as our digit option value or other normal data serializations.

spec/registry.md
macchiati: Much pithier than my suggestion!

>This is still untestable, of course.

Right, it isn't _mechanically_ testable, but it is testable (in a broader sense) by inspection of implementation's documentation of what it accepts for functions, options, and option values.

spec/registry.md
macchiati: Maybe that could be omitted, or rephrased as something like:

Implementation-defined values that represent an enumeration of values such as {`short`, `medium`, `long`} SHOULD use a distinguishing character

eemeli: > We don't have namespaces option values and the ABNF does not permit a colon in the unquoted value (there can be one in a literal, but the literal has to be quoted to have a colon).

Oh, true. Then recommending a `_` prefix is probably a good idea.

spec/registry.md
aphillips: > Much pithier than my suggestion!

Earned my paycheck today. 😉

aphillips: I think I'll commit this one. I'm curious to hear @eemeli's opinion on options, tho... This formulation *could* apply there too.

spec/registry.md
aphillips: Done. See changes.

spec/registry.md
```diff
@@ -1,11 +1,48 @@
...
+register and use user-defined _functions_ and their associated _functional handlers_.
+
+Implementations MAY implement additional _options_ not defined here.  
+Such _options_ MUST use an implementation-specific _namespace_.
+
+Implementations MAY _accept_ additional _option_ values for _options_ defined here,
+although care needs to be exercised to ensure interoperability
+and to avoid collisions with future standardization.
+There is no namespace mechanism for _option_ values,
+however, the [stability policy](#stability-policy) for this specification
```
macchiati, line 41: I don't see that the stability policy that this points to only allows a-z, A-Z, and 0-9 in option values. Does that mean we couldn't have a _standard_ future function option that took fairly liberal string options like prefex=|β/γ| ?

macchiati (commented):
I left two comments. They are not showstoppers, but I wanted to get your read before approving.

spec/registry.md
```diff
@@ -1,11 +1,48 @@
...
+
+To **_<dfn>accept</dfn>_** a function means that an implementation MUST NOT
+emit an _Unknown Function_ error for that _function_'s _identifier_.
+To _accept_ an option means that an implementation MUST NOT
+emit a _Bad Option_ error for that _option_'s _identifier_ when used with the _function_
+it is defined for
+and MUST NOT emit a _Bad Option_ error for any of the _option_ values
+(both **standard** and **optional**)
+defined for that _option_.
+Accepting a _function_ or its _options_ does not mean that any given output is produced.
```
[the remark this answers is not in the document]
aphillips: No output is one possibility I entertained!

spec/registry.md
aphillips: The stability policy says specifically:

> Updates to this specification will only reserve, define, or require function names or function option names consisting of characters in the ranges a-z, A-Z, and 0-9. All other names in these categories are reserved for the use of implementations or users.

String options *can* include wide ranging literals. This is meant to restrict enumerated keywords. Note too that the restriction is to option/function names, not values. We should probably consider amending this policy in an appropriate way. I will add an issue and agenda item to that effect.

spec/registry.md
```diff
@@ -1,11 +1,48 @@
...
+Implementations SHOULD _accept_ each **optional** _function_. 
+For each such _function_, the implementation MUST accept all _options_ and _option_ values 
+listed as **standard** for that _function_.
+
+Implementations SHOULD _accept_ _options_ and _option_ values that are marked as **optional**.
+
+Implementations MAY _accept_ _functions_ not defined in this specification.
+Functions not defined by any version of this specification MUST use an implementation-defined _namespace_.
+In addition, implementations SHOULD provide mechanisms for users to
+register and use user-defined _functions_ and their associated _functional handlers_.
```
eemeli, line 32: Would this not mean that any user-defined function "MUST use an implementation-defined _namespace_"? That seems rather arbitrary.

I do not think that we ought to place any such restrictions on user-defined functions.

spec/registry.md
```diff
@@ -1,11 +1,48 @@
...
+
+Implementations MAY implement additional _options_ not defined here.
```
eemeli, line 34:
```suggestion
Implementations MAY implement additional _options_ not defined here
for **standard** and **optional** functions.
```

spec/registry.md
```diff
@@ -1,11 +1,48 @@
...
+
+Implementations MAY _accept_ additional _option_ values for _options_ defined here,
+although care needs to be exercised to ensure interoperability
+and to avoid collisions with future standardization.
+There is no namespace mechanism for _option_ values,
+however, the [stability policy](#stability-policy) for this specification 
+permits only the ranges a-z, A-Z, and 0-9 in _option_ values.
+Implementation-defined values SHOULD use a distinguishing character 
+or character sequence, such as by prefixing with a `_` U+005F LOW LINE,
+to ensure that they don't collide with future standardization.
```
eemeli, line 45: If this is desirable, then we should update the stability policy to ensure that future standardization won't be able to use a `_` in option values.

spec/registry.md
```diff
@@ -1,11 +1,48 @@
...
-Implementations MAY implement additional _functions_ or additional _options_.
-In particular, implementations are encouraged to provide feedback on proposed
-_options_ and their values.
+This section defines the **standard** _functions_ which are REQUIRED for conformance with this specification,
+along with **optional** _functions_ that SHOULD be implemented to support
+additional functionality.
+
+To **_<dfn>accept</dfn>_** a function means that an implementation MUST NOT
+emit an _Unknown Function_ error for that _function_'s _identifier_.
+To _accept_ an option means that an implementation MUST NOT
```
eemeli, line 9:
```suggestion
To _accept_ an option means that a _function handler_ MUST NOT
```

eemeli (changes requested):
I continue to think that the proposed text is too prescriptive.

Thus far, we've been describing what future versions of the spec may do, as in the stability policy part that's been quoted in the discussion here a few times:
> Updates to this specification will only reserve, define, or require function names or function option names consisting of characters in the ranges a-z, A-Z, and 0-9. All other names in these categories are reserved for the use of implementations or users.

Note the second sentence there, which is specifically reserving a part of the non-namespaced function and option names for implementations and users, beyond even what's left to them in the namespaced identifiers. Yes, this text was originally written before we'd defined `:` namespaces, but it also was not modified in #834, where much of the text around it was refactored.

This PR presents a different type of requirement on implementations: We're no longer stating what future spec versions might do, but mandating how they MUST behave and what they might allow their users to do.

The proposed text also requires the swallowing of errors even in places where errors are non-fatal, disallowing them from being emitted even via a side-channel. This does not match our existing behaviour e.g. in option resolution, where we specifically allow for emitting errors for unresolved values while still requiring that the option resolution as a whole succeeds. We ought to allow an implementation to signal that it doesn't support an option or option value while still emitting _some_ formatted output, if we want to include the options currently proposed in #915.

spec/registry.md
eemeli: For clarity, we should probably update the stability policy text to refer to "identifiers" rather than "names" in that phrase.

spec/registry.md
aphillips: So... SHOULD use an implementation defined namespace, right?

spec/registry.md
aphillips: Not sure why this change is needed?

spec/registry.md
aphillips: Agreed. I will make a PR for that.

aphillips:
@eemeli 

> Note the second sentence there, which is specifically reserving a part of the non-namespaced function and option names for implementations and users, beyond even what's left to them in the namespaced identifiers. Yes, this text was originally written before we'd defined : namespaces, but it also was not modified in https://github.com/unicode-org/message-format-wg/pull/834, where much of the text around it was refactored.

We explicitly want to allow user-defined functions to have non-namespaced options and non-weird option values. That's why that text was (and is) like that. _This_ section is saying what users/implementations can do to **standard** and **optional** functions in the registry. If my _proposed_ text is too strong and you think we should SHOULD instead of MUST, suggest that?

> This PR presents a different type of requirement on implementations: We're no longer stating what future spec versions might do, but mandating how they MUST behave and what they might allow their users to do.

I think the requirements here were implicit in our stance about the standard function registry: you can't claim conformance if you don't implement these functions with (at least) these options. The stability policy says that we will never remove any of these. But we might add to the list over time. We don't require that implementations provide additional functions or that they support user-defined ones.

> The proposed text also requires the swallowing of errors even in places where errors are non-fatal, disallowing them from being emitted even via a side-channel. This does not match our existing behaviour e.g. in option resolution, where we specifically allow for emitting errors for unresolved values while still requiring that the option resolution as a whole succeeds. 
The proposed text does not require errors to be swallowed. It requires that specific ones not be produced. That's **_instead_** of trying to require that each and every functions and each and every option be positively implemented or have some specific effect. 

Notice that these requirements are highly targeted. For standard functions we prohibit _Unknown Function_. That is, the function must exist. For standard options we prohibit _Bad Option_ for that identifier. We don't prohibit _Bad Option_ for a bad value (this might be clarified). 

> We ought to allow an implementation to signal that it doesn't support an option or option value while still emitting some formatted output, if we want to include the options currently proposed in https://github.com/unicode-org/message-format-wg/pull/915.

This might be possible? I would tend to add an _Unsupported Operation_ error instead of trying to smoosh everything into _Bad Option_. It's the difference between searching high-and-low for why my bog standard message doesn't work ("I'm sure I spelled `currencyDisplay` correctly!!") and finding out that the value `variant` is unimplemented on this platform.

spec/registry.md
eemeli: That's weird too; the namespace used by a user-defined function should be entirely user-definable, and we should continue to allow for user functions using no namespace at all.

spec/registry.md
eemeli: This is meant to make it clearer that the additional options this is talking about are the ones definable on the core functions, rather than _all_ functions. The important context here is the next line, which requires the options to use a namespace. Without the limitation I propose, this could be read as requiring implementations to namespace all the options they add on all functions, even implementation-defined namespaced functions.

spec/registry.md
eemeli: See #929.

spec/registry.md
aphillips: I don't see that, but I'm happy to make the change.

spec/registry.md
aphillips: > Functions not defined by any version of this specification SHOULD use an implementation-defined or user-defined _namespace_.

Does that fix it? (I added _or user-defined_)

I agree with you that impls/users can do whatever they want, but we should provide guidance not to use un-namespaced functions that we don't define.

aphillips: (mutated the text: have a look)

aphillips:
I'm toying with adding:

> [!NOTE]
> An implementation can emit a _Bad Option_ error for an _option_ value
> that is ill-formed, unsupported, or violates some constraint

The potential problem is the word "unsupported"

macchiati:
I think some examples would help. 

In all cases, we want to very significantly reduces the chances of collisions. We could explore — in the future — reducing it further by putting a requirement like java does, based on domain names, but that is clearly way beyond what we should be discussing now. 

Non-standard function
- We don't want this to be :list, but rather mozilla:list
- It doesn't need namespaced options or option values, since it defines all of them.

Non-standard option on standard function
- We don't want this to be :number invert, but rather :number mozilla:invert
- It doesn't need namespaced option values, since it defines all of them

Non-standard option value on standard function standard option 
- We don't want* this to be :number style=rational, but rather :number style=mozilla:rational

\* A key feature for me is that we should distinguish enum options (whose option values are expected to be one of a fixed set of identifiers) from non-enum options, like regex=|[a-z]{3,8}(-[a-z]{3,8})*|. The former would be easy to namespace option values for, while the latter get trickier. I think this distinction is too last-minute for making into 46.1, but would make our approach more cohesive.

aphillips:
I agree with most of your comment @macchiati except:

> Non-standard option value on standard function standard option
>
> - We don't want* this to be :number style=rational, but rather :number style=mozilla:rational

An option value can be a `literal` or a `variable`. 

We don't permit `:` in **_unquoted_** literals nor do we have namespacing in option values as a feature. I don't think we need to introduce namespacing either. Non-standard option _values_ will produce _Bad Option_ errors on implementations that don't recognize the value. The suggestion of introducing a character like underscore that we (the Unicode "we") promise not to use in standard/optional (reserved name) values could help, e.g. `:number style=mozilla_rational` or `:number style=_moz_rational`. But I don't think we need to overbake it here.

macchiati:
Sorry, I should have written style=|mozilla:rational|

However, you have a good point. Having a namespace option in option values is unnecessary, _because_ and implementation can always use a namespaced option instead. 

That is, instead of `:number style=rational`, they can use `:number mozilla:style=rational`

That is much cleaner, and means they are not polluting the standard option's values. What we can the do is simply impose the requirement that:

OLD

~Implementations MAY _accept_ additional _option_ values for _options_ defined here,
although care needs to be exercised to ensure interoperability
and to avoid collisions with future standardization.
There is no namespace mechanism for _option_ values,
however, the [stability policy](#stability-policy) for this specification~

NEW
Implementations MUST not implement additional option values for standard options of standard functions. Instead, the implementations can use a standard function with a namespaced option that has the additional option values. For example:

```
1) :number style=rational // must no be used unless some version of this standard has rational as an option value
2) :number foo:style=rational // legitimate, since there is a different namespace.
```
This lets any parser of the message containing (2) know that there is a special option supported by 'foo'.

aphillips:
@eemeli can you check and see if the changes reflect your understanding?

macchiati (approved):

spec/registry.md
```diff
@@ -1,20 +1,57 @@
...
+along with **optional** _functions_ that SHOULD be implemented to support
+additional functionality.
+
+To **_<dfn>accept</dfn>_** a function means that an implementation MUST NOT
+emit an _Unknown Function_ error for that _function_'s _identifier_.
+To _accept_ an option means that a _function handler_ MUST NOT
+emit a _Bad Option_ error for that _option_'s _identifier_ when used with the _function_
+it is defined for
+and MUST NOT emit a _Bad Option_ error for any of the _option_ values
+(both **standard** and **optional**)
```
eemeli, line 13:
```suggestion
```

spec/registry.md
```diff
@@ -1,20 +1,57 @@
...
+defined for that _option_.
+Accepting a _function_ or its _options_ does not mean that a particular output is produced.
+Implementations MAY emit an _Unsupported Operation_ error for _options_
+or _option_ values that they cannot support.
+
+_Functions_ can define _options_. 
+An _option_ can be **standard** or **optional**.
+
+Implementations MUST _accept_ each **standard** _function_ and
+MUST _accept_ all _options_ and _option_ values defined as **standard** for those _functions_.
```
eemeli, line 23:
```suggestion
MUST _accept_ all _options_ defined as **standard** for those _functions_.
```

spec/registry.md
```diff
@@ -1,20 +1,57 @@
...
+
+Implementations SHOULD _accept_ each **optional** _function_. 
+For each such _function_, the implementation MUST accept all _options_ and _option_ values
```
eemeli, line 26:
```suggestion
For each such _function_, the implementation MUST accept all _options_
```

spec/registry.md
```diff
@@ -1,20 +1,57 @@
...
+An _option_ can be **standard** or **optional**.
+
+Implementations MUST _accept_ each **standard** _function_ and
+MUST _accept_ all _options_ and _option_ values defined as **standard** for those _functions_.
+
+Implementations SHOULD _accept_ each **optional** _function_. 
+For each such _function_, the implementation MUST accept all _options_ and _option_ values 
+listed as **standard** for that _function_.
+
+Implementations SHOULD _accept_ _options_ and _option_ values that are marked as **optional**.
```
eemeli, line 29:
```suggestion
Implementations SHOULD _accept_ _options_ that are marked as **optional**.
```

spec/registry.md
```diff
@@ -1,20 +1,57 @@
...
+
+Implementations MAY _accept_ _functions_ not defined in this specification.
+In addition, implementations SHOULD provide mechanisms for users to
+register and use user-defined _functions_ and their associated _functional handlers_.
+Functions not defined by any version of this specification SHOULD use 
+an implementation-defined or user-defined _namespace_.
+
+Implementations MAY implement additional _options_ not defined here
```
eemeli, line 37: Should match the text in the preceding paragraph about functions.

```suggestion
Implementations MAY implement additional _options_ not defined
by any version of this specification
```

spec/registry.md
aphillips: I'm nervous about this one, although I agree with your argument about standard/options for values. However, the requirement is meant to be that you cannot emit _Bad Option_ for a valid option value, e.g.:

```
.input {$n number maximumFractionDigits=2}
```

But you may do so for misspellings or otherwise bad values:

```
.input {$n :number maximumFActionDigit=2.5}
```

eemeli: We still allow emitting Bad Option for `2.5`, though? That's covered by the definition above:
> To _accept_ an _option_ means that a _function handler_ MUST NOT emit a _Bad Option_ error for that _option_'s _identifier_ when used with the _function_ it is defined for and MUST NOT emit a _Bad Option_ error for any of the _option_ values defined for that _option_.

Note that the final requirement only applies "for any of the _option_ values defined for that _option_", and `2.5` is not a valid _digit size option_ value.

eemeli (approved):

aphillips:
Merging per agreement in 2024-11-11 call.

PR #927, opened 2024-11-07
Replace Composition with Resolved Value sections for default functions

eemeli:
As we've now defined _resolved value_ as a term, and e.g. formatting refers to "resolving" placeholders, we should replace/expand the "Composition" sections of default function descriptions with "Resolved Value" sections.

This should help provide a clearer basis for the text around `:number offset` proposed in #926, which is currently proposing two alternative _resolved value_ definitions for `:number`.

### DIFF
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -48,6 +48,13 @@ The function `:string` has no options.
 > Proposals for string transformation options or implementation
 > experience with user requirements is desired during the Tech Preview.
 
+#### Resolved Value
+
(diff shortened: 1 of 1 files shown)


aphillips (approved):
LGTM

aphillips:
The `:date` and `:time` sections make me itch a bit. I want to say more than just "implementation-defined type" but don't really want to reawaken to issue about whether (or not) the type "trims" the other part of the value (e.g. `:date` removes time units or `:time` removes y-m-d units)

PR #932, opened 2024-11-12
Add a :math function, for addition and subtraction

### LINKED ISSUE #701: [FEEDBACK] add `:number` offset option
I investigated using a mock implementation, and there is nothing really standing in the way of adding an offset option to :number.  This could be done as an icu:offset option, but it should be noted that one wouldn't be able to have MF1 compatibility without something like it.

The key to implementing is that:

- The formatting is done with the offset applied.
- The matching is done depending on the type of key:
    - for literals the offset is **not** applied.
    - for plural/ordinal categories the offset **is** applied.

This only required the addition of a few lines of code.

I compared against the equivalent MF1 pattern. Note that MF1 has the special # symbol to indicate the formatted value with offset applied: that just turned into a regular placeholder. Here are the results.

MF1 Pattern:
```
	{gender_of_host, select, 
	 female {
	  {num_guests, plural, offset:1 
	   =0 {{host} does not give a party.}
	   =1 {{host} invites {guest} to her party.}
	   =2 {{host} invites {guest} and one other person to her party.}
	   other {{host} invites {guest} and # other people to her party.}}}
	 male {
	  {num_guests, plural, offset:1 
	   =0 {{host} does not give a party.}
	   =1 {{host} invites {guest} to his party.}
	   =2 {{host} invites {guest} and one other person to his party.}
	   other {{host} invites {guest} and # other people to his party.}}}
	 other {
	  {num_guests, plural, offset:1 
	   =0 {{host} does not give a party.}
	   =1 {{host} invites {guest} to their party.}
	   =2 {{host} invites {guest} and one other person to their party.}
	   other {{host} invites {guest} and # other people to their party.}}}}
```
MF2Pattern:
```
	.input {$num_guests :number u:offset=1}
	.match {$gender_of_host}{$num_guests}
	 female 0 {{{$host} does not give a party.}}
	 female 1 {{{$host} invites {$guest} to her party.}}
	 female 2 {{{$host} invites {$guest} and one other person to her party.}}
	 female * {{{$host} invites {$guest} and {$num_guests} other people to her party.}}
	 male 0 {{{$host} does not give a party.}}
	 male 1 {{{$host} invites {$guest} to his party.}}
	 male 2 {{{$host} invites {$guest} and one other person to his party.}}
	 male * {{{$host} invites {$guest} and {$num_guests} other people to his party.}}
	 * 0 {{{$host} does not give a party.}}
	 * 1 {{{$host} invites {$guest} to their party.}}
	 * 2 {{{$host} invites {$guest} and one other person to their party.}}
	 * * {{{$host} invites {$guest} and {$num_guests} other people to their party.}}
```
Test results:
```
OK with input: {$gender_of_host=female, $guest=Mike, $host=Sarah, $num_guests=0}
	MF1: Sarah does not give a party.
	MF2: Sarah does not give a party.
OK with input: {$gender_of_host=female, $guest=Mike, $host=Sarah, $num_guests=1}
	MF1: Sarah invites Mike to her party.
	MF2: Sarah invites Mike to her party.
OK with input: {$gender_of_host=female, $guest=Mike, $host=Sarah, $num_guests=2}
	MF1: Sarah invites Mike and one other person to her party.
	MF2: Sarah invites Mike and one other person to her party.
OK with input: {$gender_of_host=female, $guest=Mike, $host=Sarah, $num_guests=3}
	MF1: Sarah invites Mike and 2 other people to her party.
	MF2: Sarah invites Mike and 2 other people to her party.
```

eemeli:
Fixes #701.
An alternative to #926.

Adds `:math` as a new numeric function that can `add` or `subtract` from the operand.

Note that the option values are more strict than for other functions; option values beyond "digit size option" are considered errors.

### DIFF
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -381,6 +381,69 @@ together with the resolved options' values.
 
 The _function_ `:integer` performs selection as described in [Number Selection](#number-selection) below.
 
+### The `:math` function
+
+The function `:math` is a selector and formatter for matching or formatting
+numeric values to which a mathematical operation has been applied.
+
+> This function is useful for plural selection and formatting of an offset of an input value.
+> For example, it can be used in a message such as this:
+> ```
+> .input {$like_count :integer}
+> .local $others_count = {$like_count :math subtract=1}
+> .match $like_count $others_count
+> 0 * {{Your post has no likes.}}
+> 1 * {{{$name} liked your post.}}
+> * 1 {{{$name} and one other person liked your post.}}
+> * * {{{$name} and {$others_count} other people liked your post.}}
+> ```
+
+#### Operands
+
+The function `:math` requires a [Number Operand](#number-operands) as its _operand_.
+
+#### Options
+
+The options on `:math` are exclusive with each other,
+and exactly one option is always required.
+The options do not have default values.
+
+The following options and their values are
+required in the default registry to be available on the function `:math`:
+- `add`
+  - ([digit size option](#digit-size-options))
+- `subtract`
+  - ([digit size option](#digit-size-options))
+
+If no options or more than one option is set,
+or if an _option_ value is not a [digit size option](#digit-size-options),
+a _Bad Option_ error is emitted
+and a _fallback value_ used as the _resolved value_ of the _expression_.
+
+#### Resolved Value
+
+The _resolved value_ of an _expression_ with a `:math` _function_
+contains the implementation-defined numeric value
+of the _operand_ of the annotated _expression_.
+
+If the `add` option is set,
+the numeric value of the _resolved value_ is formed by incrementing
+the numeric value of the _operand_ by the integer value of the digit size option value.
+
+If the `subtract` option is set,
+the numeric value of the _resolved value_ is formed by decrementing
+the numeric value of the _operand_ by the integer value of the digit size option value.
+
+If the _operand_ of the _expression_ is an implementation-defined numeric type,
+such as the _resolved value_ of an _expression_ with a `:number` or `:integer` _annotation_,
+it can include option values.
+These are included in the resolved option values of the _expression_.
+The `:math` _options_ are not included in the resolved option values.
+
+#### Selection
+
+The _function_ `:math` performs selection as described in [Number Selection](#number-selection) below.
+
 ### Number Operands
 
 The _operand_ of a number function is either an implementation-defined type or
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -438,6 +438,13 @@ it can include option values.
 These are included in the resolved option values of the _expression_.
 The `:math` _options_ are not included in the resolved option values.
 
+> [!NOTE]
+> Implementations can encounter practical limits with `:math` _expressions_,
+> such as the result of adding two integers exceeding 
+> the storage or precision of some implementation-defined number type.
+> In such cases, implementations can emit an _Unsupported Operation_ error
+> or they might just silently overflow the underlying data value.
+
 #### Selection
 
 The _function_ `:math` performs selection as described in [Number Selection](#number-selection) below.
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -386,7 +386,8 @@ The _function_ `:integer` performs selection as described in [Number Selection](
 The function `:math` is a selector and formatter for matching or formatting
 numeric values to which a mathematical operation has been applied.
 
-> This function is useful for plural selection and formatting of an offset of an input value.
+> This function is useful for selection and formatting of values that
+> differ from the input value by a specified amount.
 > For example, it can be used in a message such as this:
 > ```
 > .input {$like_count :integer}
--- a/test/tests/functions/math.json
+++ b/test/tests/functions/math.json
@@ -1,0 +1,77 @@
+{
+  "$schema": "https://raw.githubusercontent.com/unicode-org/message-format-wg/main/test/schemas/v0/tests.schema.json",
+  "scenario": "Math function",
+  "description": "The built-in formatter and selector for addition and subtraction.",
+  "defaultTestProperties": {
+    "bidiIsolation": "none",
+    "locale": "en-US"
+  },
+  "tests": [
+    {
+      "src": "{:math add=13}",
+      "expErrors": [{ "type": "bad-operand" }]
+    },
+    {
+      "src": "{foo :math add=13}",
+      "expErrors": [{ "type": "bad-operand" }]
+    },
+    {
+      "src": "{42 :math}",
+      "expErrors": [{ "type": "bad-option" }]
+    },
+    {
+      "src": "{42 :math add=foo}",
+      "expErrors": [{ "type": "bad-option" }]
+    },
+    {
+      "src": "{42 :math subtract=foo}",
+      "expErrors": [{ "type": "bad-option" }]
+    },
+    {
+      "src": "{42 :math foo=13}",
+      "expErrors": [{ "type": "bad-option" }]
+    },
+    {
+      "src": "{42 :math add=13 subtract=13}",
+      "expErrors": [{ "type": "bad-option" }]
+    },
+    {
+      "src": "{41 :math add=1}",
+      "exp": "42"
+    },
+    {
+      "src": "{52 :math subtract=10}",
+      "exp": "42"
+    },
+    {
+      "src": "{41 :math add=1 foo=13}",
+      "exp": "42"
+    },
+    {
+      "src": ".local $x = {41 :integer signDisplay=always} {{{$x :math add=1}}}",
+      "exp": "+42"
+    },
+    {
+      "src": ".local $x = {52 :number signDisplay=always} {{{$x :math subtract=10}}}",
+      "exp": "+42"
+    },
+    {
+      "src": "{$x :math add=1}",
+      "params": [{ "name": "x", "value": 41 }],
+      "exp": "42"
+    },
+    {
+      "src": "{$x :math subtract=10}",
+      "params": [{ "name": "x", "value": 52 }],
+      "exp": "42"
+    },
+    {
+      "src": ".local $x = {1 :math add=1} .match $x 1 {{=1}} 2 {{=2}} * {{other}}",
+      "exp": "=2"
+    },
+    {
+      "src": ".local $x = {10 :integer} .local $y = {$x :math subtract=6} .match $y 10 {{=10}} 4 {{=4}} * {{other}}",
+      "exp": "=4"
+    }
+  ]
+}


spec/registry.md
```diff
@@ -381,6 +381,69 @@ together with the resolved options' values.
 
 The _function_ `:integer` performs selection as described in [Number Selection](#number-selection) below.
 
+### The `:math` function
+
+The function `:math` is a selector and formatter for matching or formatting
```
aphillips, line 386: It's not a formatting function, is it? A formatting function would require all of `:number` or `:integer`'s options, e.g. to control zero digits, separators, and the like.

spec/registry.md
```diff
@@ -381,6 +381,69 @@ together with the resolved options' values.
...
+numeric values to which a mathematical operation has been applied.
+
+> This function is useful for plural selection and formatting of an offset of an input value.
```
aphillips, line 389: Clarity

```suggestion
This function is useful for selection and formatting of values that
differ from the input value by a specified amount.
```

spec/registry.md
```diff
@@ -381,6 +381,69 @@ together with the resolved options' values.
...
+
+The following options and their values are
+required in the default registry to be available on the function `:math`:
+- `add`
+  - ([digit size option](#digit-size-options))
+- `subtract`
+  - ([digit size option](#digit-size-options))
+
+If no options or more than one option is set,
+or if an _option_ value is not a [digit size option](#digit-size-options),
```
aphillips, line 419: is not a => does not resolve to?

we want to allow `{$n :math subract=$x}` where _$x_ contains a digit size option literal

spec/registry.md
```diff
@@ -381,6 +381,69 @@ together with the resolved options' values.
...
+
+If the _operand_ of the _expression_ is an implementation-defined numeric type,
+such as the _resolved value_ of an _expression_ with a `:number` or `:integer` _annotation_,
+it can include option values.
+These are included in the resolved option values of the _expression_.
+The `:math` _options_ are not included in the resolved option values.
+
+#### Selection
+
+The _function_ `:math` performs selection as described in [Number Selection](#number-selection) below.
```
aphillips, line 445: Should this make it clear that the selection is done after the math is performed?

spec/registry.md
```diff
@@ -381,6 +381,69 @@ together with the resolved options' values.
...
+  - ([digit size option](#digit-size-options))
+- `subtract`
+  - ([digit size option](#digit-size-options))
+
+If no options or more than one option is set,
+or if an _option_ value is not a [digit size option](#digit-size-options),
+a _Bad Option_ error is emitted
+and a _fallback value_ used as the _resolved value_ of the _expression_.
+
+#### Resolved Value
```
aphillips, line 423: Do we need a note about implementation specific boundary conditions, e.g.

```
{$n :math add=$maxIntegerOnMyPlatform}
```

spec/registry.md
eemeli: As proposed, you would be able to format a `:math` annotated value, so doesn't that make it a formatting function? If you need the options, you can set them in a separate declaration.

spec/registry.md
eemeli: Yes; that's covered by this in the _digit size option_ definition:
> Implementations MAY also accept implementation-defined types as the value.

spec/registry.md
eemeli: I don't think so; that's also covered under _digit size option_:
> Implementations MAY define an upper limit on the _resolved value_ of a digit size option option consistent with that implementation's practical limits.

spec/registry.md
eemeli: We could, but it's not technically necessary, as Number Selection includes this:
> When implementing `MatchSelectorKeys(resolvedSelector, keys)` where `resolvedSelector` is the _resolved value_ of a _selector_ and `keys` is a list of strings, numeric selectors perform as described below.

which ends up referring back to the preceding Resolved Value section, which clarifies when the modifications are done on the _resolved value_.

spec/registry.md
```diff
@@ -381,6 +381,70 @@ together with the resolved options' values.
...
+If the _operand_ of the _expression_ is an implementation-defined numeric type,
+such as the _resolved value_ of an _expression_ with a `:number` or `:integer` _annotation_,
+it can include option values.
+These are included in the resolved option values of the _expression_.
+The `:math` _options_ are not included in the resolved option values.
+
+#### Selection
+
+The _function_ `:math` performs selection as described in [Number Selection](#number-selection) below.
+
```
macchiati, line 447: I see some significant problems in these formulations.

>The _resolved value_ of an _expression_ with a `:number` _function_
contains an implementation-defined numerical value
of the _operand_ of the annotated _expression_,

If that is all the resolved value is, then there is no difference in resolved value between the following two.
```
.local $x1 = {$x :integer}
.local $x2 = {$x :number}
```

But of course there is a big difference: the resolved value of $x1 contains a :integer annotation and the resolved value of $x2 contains a :number annotation, as in:
```
such as the _resolved value_ of an _expression_ with a `:number` or `:integer` _annotation_,
```

And it has to be that way, otherwise no .local setting later on could tell the difference between them.

So I think L142 of formatting.md is insufficient:

> ```ts
> interface MessageValue {
>   formatToString(): string
>   formatToX(): X // where X is an implementation-defined type
>   getValue(): unknown
>   selectKeys(keys: string[]): string[]
> }
> ```

It also needs do have something like: getFunctionAnnotation();

In shorthand, the resolved value needs to logically contain a triple {datatype, functionAnnotation, optionMap}. Otherwise, a subsequent .local statement using a resolved value wouldn't have enough information to know how to compose.

So let's look at the following:
```
The _operand_ of a number function is either an implementation-defined type or
a literal whose contents match the `number-literal` production in the [ABNF](/spec/message.abnf).
All other values produce a _Bad Operand_ error.
```

But the operand can also be a resolved value, which is neither a literal nor just an implementation datatype; it has additional information.

And in multiple cases in this file, we have wording like:

>Resolved Value
>The resolved value of an expression with a :number function contains an implementation-defined numerical value of the operand of the annotated expression, together with the resolved options' values.

Now, if somewhere we say that the resolved value of any expression with a function contains that function as an annotation, then we could get away with this shorthand (although we also need to fix the wording for options). 

Otherwise, we need to say:
>The resolved value of an expression with a :number function contains an implementation-defined numerical value of the operand of the annotated expression, together with **the :number function annotation, and** the **options with their respective** resolved option~'s~ values.

macchiati (changes requested):
Whoa, I must of missed this earlier. It is seriously wrong.

```

**Comment by mihnita** [comment]
I really don't understand how this is supposed to work.

It looks very close to a :number with offset... except that the exact selection, the keyword selection, and the formatting are all done both on the same value.

Which seem cleaner, but less intuitive for a user:

```
.local $countDelta = {$guestCount :math sub=2}
.match $countDelta
   -2 {{No guest}}
   -1 {{{$guest1}}}"
    0 {{{$guest1} and {$guest2}}}"
  one {{{$guest1}, {$guest2}, and {$countDelta} other guest.}}"
    * {{{$guest1}, {$guest2}, and {$countDelta} other guests.}}"
```

As a user is not intuitive at all why would I select on `-2` to get `"No guest"`, or why a `:math` does something that looks like plural selection.

File: spec/registry.md (hunk 5 above)
Comment (aphillips) [reply spec/registry.md:424]: Right, although you can get problems with overflow without approaching the "practical limits" for the option value, e.g.

`{2147483646 :math add=2}  // exceeds Java int by 1`

I think a note is called for. I'll propose one when I do a full review.

File: spec/registry.md (hunk 6 above)
Comment (aphillips) [reply spec/registry.md:447]: I don't get it. 

The composition model is that functions can expose specific options, but the identity of the function itself is not passed. There is no difference between the resolved values of `{$x :number}` and `{$x :integer}` with regard to the receiving function: it's just a numeric type. A subsequent local declaration doesn't need to know which annotation was previously applied: the preceding function handler modified the resolved value and/or passed the relevant options to communicate what happened.

It has to be that way, because otherwise each function would need to understand what all of the other functions (including namespaced ones) "mean".

If `{$x :math subtract=2}` doesn't return `$x - 2` as its resolved value, `:number` isn't going to know what to do.

There are places where I have actively resisted "doing harm" to the resolved value based on the annotation. That's an argument we can have. Mostly I'm concerned that we don't strip information from the resolved value because of some formatting instructions:

```
.input {$now :date style=medium}
.local $time = {$now :time style=short}
{{"Today is {$now} and it is {$time}" only works if :date doesn't remove the time.
    I assume that is okay because immutability says that if $now is a Date object or ZonedDateTime
    there is no need to convert it to just a ZonedDate and make the time midnight.}}
```

We could change `:integer` to make the resolved value the floor value of the operand or to make it the implementation-defined integer-type casting of the operand. That doesn't break immutability because the resolved value is only exposed in a declaration. I think that's the better course than what you're proposing.

Comment (macchiati) [reply spec/registry.md:447]: I think I conflated too many thoughts in my message. So put on hold the notion of the function annotation (not needed for release) and I focus on just 2 issues.

1. **The operand for a function may also be a resolved value** (which includes an optionMap). That is not the same as an implementation defined type (eg Date); a resolved value can **have an** implementation-defined type. Thus 

OLD
> The _operand_ of a number function is either an implementation-defined type or
a literal whose contents match the `number-literal` production in the [ABNF](/spec/message.abnf).
All other values produce a _Bad Operand_ error.

needs to be changed to:

NEW
> The _operand_ of a number function is either an implementation-defined type or
a literal whose contents match the `number-literal` production in the [ABNF](/spec/message.abnf),
**or a resolved value.**
All other values produce a _Bad Operand_ error.

(with similar changes for other definitions of _operand_ of an X function.)

2. **The expression "together with the resolved options' values." is not accurate.** For {|3| :number roundingIncrement=5 trailingZeroDisplay=auto}, that would imply a logical result of 

`{operand=3, resolvedOptionValues=[5, auto]} 
`
rather than the logical result of

`{operand=3, resolvedOptionMap={roundingIncrement=5, trailingZeroDisplay=auto}} 
`
So we need to have:

OLD
> ... value of the operand of the annotated expression, together with the resolved options' values.

changed to something like:

NEW
> ... value of the operand of the annotated expression, together **with a map from options to their resolved** values

Comment (aphillips) [reply spec/registry.md:447]: The problem with the first item is that the resolved value might not be of the correct implementation-defined type. The current text accounts for this by not mentioning. We could say instead:

> The _operand_ of a number _function_ is either an implementation-defined type or
> a _literal_ whose contents match the `number-literal` production in the [ABNF](message.abnf)
> (either of which can be the _resolved value_ of a previous _declaration_).
> All other values produce a _Bad Operand_ error.

The second suggestion we already discussed. Saying "map" was thought too prescriptive (ditto "array", "list", etc.). I can see how one might read "options' values" to mean **only** the values and not the k-v pairing. We might say:

> ... value of the _operand_ of the annotated _expression_, together with _options_ and their resolved values.

File: spec/registry.md (hunk 5 above)
Comment (eemeli) [reply spec/registry.md:424]: Tbf, an integer overflow does seem like a practical limit to me.

File: spec/registry.md (hunk 6 above)
Comment (macchiati) [reply spec/registry.md:447]: From Addison

> The problem with the first item is that the resolved value might not be of the correct implementation-defined type. The current text accounts for this by not mentioning. We could say instead:
> 
> > The _operand_ of a number _function_ is either an implementation-defined type or
> > a _literal_ whose contents match the `number-literal` production in the [ABNF](message.abnf)
> > (either of which can be the _resolved value_ of a previous _declaration_).
> > All other values produce a _Bad Operand_ error.

Wouldn't quite work. Again, we have defined the resolved value of {|3| :number style=percent} to be **NOT** the same as |3| — since the resolved value includes options. I think what we could say:

> > (either of which can be from the _resolved value_ of a previous _declaration_).

I think that is at least headed in the right direction, though not as correct as 

> or a resolved value with those as an operand.

> 
> The second suggestion we already discussed. Saying "map" was thought too prescriptive (ditto "array", "list", etc.). I can see how one might read "options' values" to mean **only** the values and not the k-v pairing. We might say:
> 
> > ... value of the _operand_ of the annotated _expression_, together with _options_ and their resolved values.

That sounds perfect.

File: spec/registry.md (hunk 5 above)
Comment (aphillips) [reply spec/registry.md:424]: Integer overflow is a "practical limit" but the text only limits the "resolved value of a digit size option", not problems like overflow due to the operation itself. 

I'll suggest some text. It would just be a note, so not material to merging this.

**Review by aphillips — APPROVED** [review]

File: spec/registry.md (hunk 6 above)
Comment (aphillips) [reply spec/registry.md:444]: @macchiati 
> Wouldn't quite work. Again, we have defined the resolved value of {|3| :number style=percent} to be NOT the same as |3| — since the resolved value includes options. I think what we could say:
>> (either of which can be from the resolved value of a previous declaration).

I think this formulation is actually counterproductive. The _operand_ is not extracted from the _resolved value_. The _operand_ **includes** the accumulated options (which might affect the processing of the _function_ or which might be filtered or ignored).

I agree that the _operands_ in the two expressions in your comments have different _resolved values_, but one of the differences is that `{|3| :number style=percent}`'s _resolved value_ is likely to be an implementation-defined numeric type (such as Java `BigInteger` or `BigDecimal`) and `:math` wouldn't have to parse the literal `|3|` again. 

In any case, my formulation does not say "from" on purpose.

The later "perfect" quote relates to the resolved value _of the `:math` function_, which is the **output** of the function. My clarification moved the word 'resolved' to try to make clear that the options and their values were both resolved. But this is a different context from the _operand_ discussion just above.

- Operand: a naked argument or a _resolved value_ (which is...)
- Resolved value: a resolved (and possibly modified) operand plus resolved options and their resolved values

Oddly, this discussion appears to be about a stock formulation in all of our functions. Can we merge `:math` and then make any changes globally? Can I get your approval on this PR? I filed #937 to track this conversation so we make any agreed changes.

Comment (macchiati) [reply spec/registry.md:444]: Sure.

**Review by macchiati — APPROVED** [review]

**Comment by aphillips** [comment]
I'm going to merge this. @mihnita (or anyone else) If you have concerns about the resulting text, please raise PRs or issues.

File: spec/registry.md (hunk 4 above)
Comment (macchiati) [reply spec/registry.md:450]: The following still bothers me:

> If the _operand_ of the _expression_ is an implementation-defined numeric type,
> such as the _resolved value_ of an _expression_ with a `:number` or `:integer` _annotation_,
> it can include option values.

I think we should replace that by:

> If the _operand_ of the _expression_ is an implementation-defined numeric type,
> such as the _resolved value_ of an _expression_ with a `:number` or `:integer` _annotation_,
> it includes their options.

I really don't think we want the following two to differ in selection behavior, formatting, etc.

```
.local $y = {|3| :number maximumSignificantDigits=2}
.local $x = {$y :math subtract=1}
```

```
.local $x = {|2| :number maximumSignificantDigits=2}
```

So I'd recommend that the resolved value of :math inherit both the 'datatype' and the options from its operand.

**Review by macchiati — CHANGES_REQUESTED** [review]

**Comment by mihnita** [comment]
> I'm going to merge this. @mihnita (or anyone else) If you have concerns about the resulting text, please raise PRs or issues.

I commented 3 days ago, with no answer.

But I would like to see how the most common use case looks like. The one in my example or (more complicated) the one in issue #701, which this PR tries to solve.

For now I don't see how this is solving the problem.

So yes, I DO have concerns. Unanswered.

---

It also makes things less intuitive. Select on math is not something people do.
One would expect math to take one or more numeric values, do some operation(s), and return some more numeric values.
Not to format, and not to select.

If this proposal makes `:math` format and select, it is not much different than :integer with `offset`.

It also hijacks a very good name (`math`) that would probably be better use later for, you know, math things?

**Comment by macchiati** [comment]
I commented seconds too late. Please look at my comment.

**Comment by macchiati** [comment]
Mihai, I view :math as basically producing an altered version *of a numeric
resolved value*, and thus having the same "features" (data type and
options) as its operand. Thus using it in selection and/or formatting is
not surprising.

**Comment by aphillips** [comment]
@macchiati 
> I think we should replace that by:
>> If the operand of the expression is an implementation-defined numeric type,
>>  such as the resolved value of an expression with a :number or :integer annotation,
>>  it includes their options.

The reason it does not say "their options" is that some options are not passed through. This isn't necessarily the case for `:number` or `:integer` currently, but it might in the future. Notice that the resolved value examples are "such as", but might contain e.g. `:unit` or `:currency` or some other numeric type expression. Only the passed options are replicated down.

Otherwise fully agree with your observations.

**Comment by aphillips** [comment]
... and a second later it occurs to me to point out that `:math`'s options are **_not_** passed.

**Comment by mihnita** [comment]
> Mihai, I view :math as basically producing an altered version *of a numeric
resolved value*, and thus having the same "features" (data type and
options) as its operand. 

But then what's the difference between this and :integer with offset?
It does the exact same things.

---

I still want to see an example of how this looks like for the most common use case.

---

Just curious, why was this merged? Accident?

**Comment by macchiati** [comment]
Yes, I agree with :math's options not being included in its resolved
values. The numeric options that are passed through can be enumerated
explicitly OR say that all of them are currently passed through, and if and
when at some future version adds an option that shouldn't be passed
through, then an explicit mention can be made as to the exception.

(I think the latter practice is far easier to read, since the former always
forces the reader to go back to the numeric functions to see which things
are *not* passed through. Moreover, it gives the reader the confidence that
:math is basically producing an altered version *of a numeric
resolved value*, and thus having the same "features" (data type and
options) as its operand. Thus using it in selection and/or formatting is
not surprising.)

**Comment by aphillips** [comment]
@mihnita I apologize. I didn't mean to overlook your comment.

I merged this because I didn't see further commentary and thought we'd agreed to pursue this in the WG call. I'm trying to tie up loose ends before this coming week's call.

There are three possibilities here, I think. I'm sympathetic to not making `:math` a formatter for the reason you cite. The options seem to be:

- Neither a selector nor a formatter. This would make `:math` a purely "transform" declaration and not valid to use in a pattern. This has a certain appeal. However, it would require care when doing declarations:

```
.input {$x :integer}
.local $y = {$x :math subtract=1}
.local $z = {$y :integer}
.match $z * {{{$z} is required because selecting on {$y} is an error}}
```

- Only a selector. This fixes the double-declaration, but requires annotation in a pattern.
```
.input {$x :integer}
.local $y = {$x :math subtract=1}
.match $y
* {{{$y :integer} works but {$y} produces fallback}}
```

- Both, as now. This produces the problems you enumerated, but makes the resolved value of `:math` behave normally otherwise.

As I said, arguments can be made for each of these. I have added this to the agenda for Monday.

**Comment by mihnita** [comment]
> I merged this because I didn't see further commentary

Yes, but you announced "I'm going to merge this." 52 minutes ago

Mark answered 51 minutes ago, and I answered 51 minutes ago.

Why even ask if there is no time allowed for answers?

> I merged this because I didn't see further commentary

I didn't see a point to comment more if a 3 days comment was unaddressed?

> I have added this to the agenda for Monday.

Can someone have some example of what the equivalent of this MF1 looks like?

```
{guestCount, plural, offset:2
     =0 {No guest}}
     =1 {{$guest1}}
     =2 {{$guest1} and {$guest2}}
    one {{$guest1}, {$guest2}, and # other guest.}
  other {{$guest1}, {$guest2}, and # other guests.}
}
```

Because that is what issue https://github.com/unicode-org/message-format-wg/issues/701 is about.

So that we know what we will discuss on Monday?

**Comment by aphillips** [comment]
```
.input {$guestCount :integer}
.local $others = {$guestCount :math subtract=2}
.match $guestCount $others
0 * {{No guests}}
1 * {{guest {$guest1}}}
2 * {{guests {$guest1} and {$guest2}}}
* one {{guests {$guest1}, {$guest2}, and {$others} more guest}}
* * {{... {$others} more guests}}
```

Double selectors are needed because you need plural handling for the back part of the list (imagine Polish). 

Note that I'm working on a phone so the example is ugly.

eemeli:
@mihnita In addition to the example given by @aphillips above, the merged text also includes this example: https://github.com/unicode-org/message-format-wg/blob/b6ba4bf977be30ec51dfd3e05aec3dd22bbc07dc/spec/registry.md?plain=1#L386-L397

mihnita:
> Double selectors are needed because you need plural handling for the back part of the list (imagine Polish).

Got you.

So why can't this be done with :integer and offset?
Why do we need this `math` thing?

Functionality similar to MF1 is implementable without problems with `:number`.

See it here: 
https://github.com/unicode-org/icu/blob/06a23f8d37ce0c3a985bb676cbc05eb7bd5f3573/icu4j/main/core/src/test/java/com/ibm/icu/dev/test/message2/MessageFormat2Test.java#L334

Submitted 2 years ago:
https://github.com/unicode-org/icu/pull/2170/files#diff-d51d5c796bff35d045a8d917f5bcef8f9a1f353184125c903003c4c0dbb14d99R333

eemeli:
> So why can't this be done with :integer and offset?
> Why do we need this `math` thing?

I presented some arguments originally here: https://github.com/unicode-org/message-format-wg/pull/926#issuecomment-2466228706

mihnita:
I understand and somewhat agree with
"the weirdness of having the =1 and one operate on different values."

But I think that having to use a "double selection" for the same
functionality is weirder.

I suspect that most people would have to look up some example so figure out
how to build the same functionality.


But ok, maybe "weird" is in the eyes of the beholder.

Then why not use the exact same "trick" with two selectors, both :integer?
Why do we have to invent another function just for that?
Basically replace :math with :integer in the example submitted?


Mihai

PR #934, opened 2024-11-14
Remove tech preview note from `:string`

aphillips:
Remove a note that is no longer relevant

### DIFF
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -81,6 +81,13 @@ All other values produce a _Bad Operand_ error.
 
 The function `:string` has no _options_.
 
+> [!NOTE]
+> While `:string` has no built- in _options_,
+> _options_ in the `u:` _namespace_ can be used. 
+> For example: 
+>```
+> {$s :string u:dir=ltr u:locale=fr-CA}
+>```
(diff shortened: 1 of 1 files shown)


macchiati (approved):

eemeli (approved):
Removing the note is fine; I'd be a bit tempted to clarify the text a bit as we separately define the meaning of `u:` options:
```
{foo :string u:dir=ltr u:locale=en-US}
```

spec/registry.md
```diff
@@ -79,11 +79,7 @@ All other values produce a _Bad Operand_ error.
 
 #### Options
 
-The function `:string` has no options.
-
-> [!NOTE]
-> Proposals for string transformation options or implementation
-> experience with user requirements is desired during the Tech Preview.
+The function `:string` has no _options_.
```
aphillips, line 83:
```suggestion

> [!NOTE]
> While `:string` has no built- in _options_,
> _options_ in the `u:` _namespace_ can be used. 
> For example: 
>```
> {$s :string u:dir=ltr u:locale=fr-CA}
>```

```

macchiati (approved):

PR #938, opened 2024-11-15
Remove misleading terms/clauses for `:currency` options and operands

macchiati:
We already state near the top of the document that:

>Implementations MAY _accept_ additional _option_ values for _options_ defined here.
However, such values might become defined with a different meaning in the future,
including with a different, incompatible name
or using an incompatible value space.
Supporting implementation-specific _option_ values for **standard** or **optional** functions is NOT RECOMMENDED.

So there is no need to have a separate statement for particular options saying that X must be either Y or an implementation defined value. Moreover, by calling it out on *one* option on a function, it makes the readers mistrust whether is actually the case for _other_ options on the function.

Eg, "berries can be added to any desserts" at the top of the menu, then later 
"desserts are pancakes (optionally with berries), or waffles". 

It makes people question whether you can have waffles with berries even though you say at the top you can.

### DIFF
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -393,9 +393,8 @@ or it can be a [Number Operand](#number-operands), as long as the option
 The option `currency` MUST NOT be used to override the currency of an implementation-defined type.
 Using this option in such a case results in a _Bad Option_ error.
 
-The value of the _operand_'s `currency` MUST be either a string containing a
-well-formed [Unicode Currency Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeCurrencyIdentifier)
-or an implementation-defined currency type.
+The value of the _operand_'s `currency` is a
+well-formed [Unicode Currency Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeCurrencyIdentifier).
 Although currency codes are expected to be uppercase,
 implementations SHOULD treat them in a case-insensitive manner.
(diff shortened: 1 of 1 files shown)


aphillips:
You're changing the _operand_ definition, which is not the same thing as an _option_ value. All of our functions **explicitly** allow implementation defined types in the _operand_ section.

Similarly, _option_ values covered by the text you cite (which might not be clear enough) provides permission to use a different value space (which mostly means literal values, which we caution against for good reason). The text you're eliding here permits the same implementation defined types to be used as an option value.

What we're trying to permit is this:

```java
ImmutableMap<String,Object> args = ImmutableMap.of(
   "currency", Currency.getInstance("USD"),
   "value", 123.45
);
MessageFormatter mf2 = MessageFormatter.builder()
    .setPattern("You have: {$value :currency currency=$currency}").build();
System.out.println(mf2.formatToString(args));
```

We want `$currency` to be allowed to be a `java.util.Currency` (or `com.ibm.icu.util.Currency`) here, not just a literal `USD`.

Similarly, as we explicitly call out, want to be able to use a specific boxing value (here, ICU4J's `CurrencyAmount`) for the operand `$value`:

```java
ImmutableMap<String,Object> args = ImmutableMap.of(
   "value", new CurrencyAmount( 123.45, Currency.getInstance("USD"))
);
MessageFormatter mf2 = MessageFormatter.builder()
    .setPattern("You have: {$value :currency}").build();
System.out.println(mf2.formatToString(args));
```

macchiati:
Makes sense, I'll close this.

PR #939, opened 2024-11-16
Drop exp value for :number tests with bad option values

eemeli:
When an option value does not match the required option values, function behaviour is left as implementation-defined. Therefore, we should not specify an expected formatted result for them.


aphillips (approved):

aphillips:
@eemeli We have a merge conflict here. I think removing the `exp` is expected (per the PR description).

macchiati (approved):

mradbourne (approved):

PR #941, opened 2024-11-16
Linkify stability policy

### LINKED ISSUE #930: Linkify the stability policy
In reviewing #929 I notice that we don't linkify that section and we should.

aphillips:
Fixes #930 

Also clarifies that [!IMPORANT] blocks are normative.

### DIFF
--- a/spec/README.md
+++ b/spec/README.md
@@ -72,7 +72,11 @@ A **_term_** looks like this when it is defined in this specification.
 
 A reference to a _term_ looks like this.
 
+> [!NOTE]
 > Examples are non-normative and styled like this.
+
+> [!IMPORTANT]
+> Text marked "Important" like this are normative.
 
 ### Stability Policy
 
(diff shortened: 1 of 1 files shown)


spec/README.md
```diff
@@ -84,42 +88,42 @@ Updates to this specification will not make any valid _message_ invalid.
...
 Updates to this specification will not change the syntactical meaning
 of any syntax defined in this specification.
 
-Updates to this specification will not remove any functions defined in the default registry.
+Updates to this specification will not remove any _functions_ defined in the default function registry.
 
-Updates to this specification will not remove any options or option values
-defined in the default registry.
+Updates to this specification will not remove any _options_ or _option_ values
+defined in the default function registry.
```
macchiati, line 104: One additional item. We should add the following clause for clarity.

```suggestion

Updates to this specification may **deprecate** _functions_, _options_, and _operands_ defined in the default registry, as is standard practice in Unicode specifications. This is to indicate that the use of those _functions_, _options_, or _operands_ is discouraged. However, they are never removed, and implementations can continue to support them.
```

macchiati (changes requested):
Just one item

spec/README.md
aphillips: We already say this (line 146), except we don't mention _operands_ in the stability policy and, of course, it's much less explanatory. Note that the stability policy is about what we *won't* do.

Making this change would be normative and require WG consensus, so I'll change the tags and remove fast-track until we discuss Monday.

macchiati: Sounds good

spec/README.md
```diff
@@ -72,8 +72,12 @@ A **_term_** looks like this when it is defined in this specification.
 
 A reference to a _term_ looks like this.
 
+> [!NOTE]
 > Examples are non-normative and styled like this.
```
eemeli, line 76: This change ought to be reflected in the text. We have _many_ `>` indented blocks without a `[!NOTE]` specifier that are currently non-normative, and which would be made maybe-normative by this change.

eemeli (changes requested):

spec/README.md
eemeli: Eh, this is a completely separate change from this PR. It really should be handled separately, but we're going to feature-creep it into this same change once again, yes?

Something like this was previously discussed here: https://github.com/unicode-org/message-format-wg/pull/883#discussion_r1751083540, and the spec text already includes this, which is in an `[!IMPORTANT]` block and therefore normative: https://github.com/unicode-org/message-format-wg/blob/f3344b5cea59e46aa8a0aa150904f53eb7690545/spec/README.md?plain=1#L142

aphillips: > Eh, this is a completely separate change from this PR. It really should be handled separately, but we're going to feature-creep it into this same change once again, yes?

Under normal circumstances, I would insist on it. But I'm hoping to be done tomorrow.

spec/README.md
aphillips: Agreed. I dithered about adding the `[!NOTE]` notation. I will remove it and add a separate bit about notes.

eemeli (approved):
The current text looks good. I don't think the added deprecation item is necessary, as we already mention it later in the text. If there's something missing from that (such as explicitly allowing for deprecating operands), it ought to be modified in place.

PR #942, opened 2024-11-17
Always apply isolation when `u:dir` is set

eemeli:
See #916 for some of the preceding discussion.

In addition to changing the directionality of a placeholder, we should treat an explicitly set `u:dir` option as a request to isolate the placeholder, even if the direction matches the message's direction.

This requires a small change to our default bidi strategy.

The inheritance of the option is not specified when it's set on a value that's used as a function operand, so e.g. a message like
```
.local $world = {world :string u:dir=ltr}
{{hello {$world :string}}}
```
might or might not have the placeholder isolated (depending on the implementation), while it would always be isolated in messages like
```
hello {world :string u:dir=ltr}
```
and
```
.local $world = {world :string u:dir=ltr}
{{hello {$world}}}
```

### DIFF
--- a/spec/u-namespace.md
+++ b/spec/u-namespace.md
@@ -84,8 +84,9 @@ or a _variable_ whose _resolved value_ is one of these _literals_:
 - `ltr`: left-to-right directionality
 - `rtl`: right-to-left directionality
 - `auto`: directionality determined from _expression_ contents
-- `inherit` (default): directionality inherited from the _message_,
-  without isolation of the _expression_ value
+- `inherit` (default): directionality inherited from the _message_
+   or from the _resolved value_ of the _operand_ without
+   requiring isolation of the _expression_ value.
 
 For other values, a _Bad Option_ error is emitted
 and the value of the `u:dir` option is ignored.
--- a/spec/u-namespace.md
+++ b/spec/u-namespace.md
@@ -68,7 +68,8 @@ not valid, or some other reason.
 ### `u:dir`
 
 Replaces the base directionality defined in
-the _function context_ for this _expression_.
+the _function context_ for this _expression_
+and applies bidirectional isolation to it.
 
 If this option is set on _markup_, a _Bad Option_ error is emitted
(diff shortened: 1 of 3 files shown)


test/tests/u-options.json
```diff
@@ -102,6 +105,25 @@
         { "type": "bidiIsolation", "value": "\u2069" }
       ]
     },
+    {
+      "src": ".local $world = {world :string u:dir=ltr u:id=foo} {{hello {$world}}}",
```
aphillips, line 109: This is intended as the transitivity text, I think? Should we have one with mulitple `.local` declarations? or `.input` followed by `.local`?

aphillips (approved):
I think this is an important and quite useful improvement.

test/tests/u-options.json
eemeli: Eh, not sure if it's necessary. I think the important part is making sure that moving the option from the placeholder to a declaration works.

spec/u-namespace.md
```diff
@@ -68,20 +68,24 @@ not valid, or some other reason.
...
 before calling the _function handler_.
+Its value is retained in the _resolved value_ of the _expression_.
 
 The value of the `u:dir` _option_ MUST be one of the following _literal_ values
 or a _variable_ whose _resolved value_ is one of these _literals_:
 - `ltr`: left-to-right directionality
 - `rtl`: right-to-left directionality
 - `auto`: directionality determined from _expression_ contents
+- `inherit` (default): directionality inherited from the _message_,
+  without isolation of the _expression_ value
```
aphillips, line 88: This is a good add, but should also allow the value. Also, note that "without isolation" actually means without _requiring_ isolation (isolation can still happen if the directions don't match)

```suggestion
- `inherit` (default): directionality inherited from the _message_
   or from the _resolved value_ of the _operand_ without
   requiring isolation of the _expression_ value.
```

aphillips (changes requested):

PR #943, opened 2024-11-17
Rename `:currency` option `currencySign` as `sign`

eemeli:
The option name is inherited from ECMA-402, but as we've already opted to go with `:currency` instead of `:number style=currency`, we don't need the prefix for `currencySign`, but can use just `sign` for it.

The same does not quite apply to `currencyDisplay`, as we also have `compactDisplay`, and that option is specifically about the display of the "currency" part of the formatted value.

### DIFF
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -564,7 +564,7 @@ The following options and their values are required to be available on the funct
 - `numberingSystem`
    - valid [Unicode Number System Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeNumberSystemIdentifier)
      (default is locale-specific)
-- `currencySign`
+- `sign`
   - `accounting`
   - `standard` (default)
 - `currencyDisplay`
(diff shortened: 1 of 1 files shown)


spec/registry.md
```diff
@@ -564,7 +564,7 @@ The following options and their values are required to be available on the funct
 - `numberingSystem`
    - valid [Unicode Number System Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeNumberSystemIdentifier)
      (default is locale-specific)
-- `currencySign`
+- `sign`
```
aphillips, line 567: It's not really "sign" in the usual sense, since the value `accounting` often means this sort of format: `($123.45)` and because it's easy to be confused about `sign` vs. `signDisplay`.

If we just want to get rid of the spurious `currency` in the name, I'd probably look to change `currencyDisplay` to `symbol` with options `auto` (or perhaps `regular`), `narrow`, `name`, `code`, `formal`, and `never`.

aphillips (commented):
🤷‍♂️ 

I like the usability improvement, but wish we had this conversation previously 😄

spec/registry.md
eemeli: Would `signDisplay` be better?

aphillips: We already have `signDisplay` with a long list of values, so it can't be that--unless we want to merge them?

mihnita: `symbol` sounds good.
The Unicode block is "Currency Symbols".

Wikipedia says
"A currency symbol or currency sign is a graphic symbol used..."
https://en.wikipedia.org/wiki/Currency_symbol

So both would be fine. But sign is more confusing.

PR #944, opened 2024-11-17
Address "implementation-defined" literal and type values

aphillips:
In [this thread](https://github.com/unicode-org/message-format-wg/pull/911#discussion_r1845265757) on #911, @macchiati and I discussed the handling of implementation-defined literal values and implementation-defined types.

This change splits the "MAY _accept_" for these two cases, permitting both and saving us having to say "... or an implementation-defined value..." in lots of places.

### DIFF
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -38,13 +38,15 @@ by any version of this specification
 for **standard** and **optional** functions.
 Such _options_ MUST use an implementation-specific _namespace_.
 
-Implementations MAY _accept_ additional literal _option_ values for _options_ defined here.
+Implementations MAY _accept_, for _options_ defined in this specification,
+_option_ values which are not defined in this specification.
 However, such values might become defined with a different meaning in the future,
 including with a different, incompatible name
 or using an incompatible value space.
-Supporting implementation-specific literal _option_ values for **standard** or **optional** functions is NOT RECOMMENDED.
+Supporting implementation-specific _option_ values for **standard** or **optional** functions is NOT RECOMMENDED.
 
-Implementations MAY _accept_ implementation-defined types for _operands_ and _option_ values defined here.
+Implementations MAY _accept_, for _operands_ and _option_ values defined in this specification,
+implementation-defined types.
 Such values can be useful to users in cases where local usage and support exists
 (including cases in which details vary from those defined by Unicode and CLDR).
 
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -45,7 +45,7 @@ including with a different, incompatible name
 or using an incompatible value space.
 Supporting implementation-specific _option_ values for **standard** or **optional** functions is NOT RECOMMENDED.
 
-Implementations MAY _accept_, for _operands_ and _option_ values defined in this specification,
+Implementations MAY _accept_, for _operands_ or _option_ values defined in this specification,
 implementation-defined types.
 Such values can be useful to users in cases where local usage and support exists
 (including cases in which details vary from those defined by Unicode and CLDR).
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -45,8 +45,8 @@ including with a different, incompatible name
 or using an incompatible value space.
 Supporting implementation-specific _option_ values for **standard** or **optional** functions is NOT RECOMMENDED.
 
-Implementations MAY _accept_, for _operands_ or _option_ values defined in this specification,
-implementation-defined types.
+Implementations MAY _accept_, for _operands_ or _options_ defined in this specification,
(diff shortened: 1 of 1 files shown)


spec/registry.md
```diff
@@ -38,11 +38,23 @@ by any version of this specification
 for **standard** and **optional** functions.
 Such _options_ MUST use an implementation-specific _namespace_.
 
-Implementations MAY _accept_ additional _option_ values for _options_ defined here.
+Implementations MAY _accept_ additional literal _option_ values for _options_ defined here.
```
eemeli, line 41: Using "literal" here is confusing, to the extent that I'm not sure what this sentence now means. The confusion stems from our definition of the word as a part of our syntax: https://github.com/unicode-org/message-format-wg/blob/f3344b5cea59e46aa8a0aa150904f53eb7690545/spec/syntax.md?plain=1#L706-L707

An implementation (or, more accurately, a _function handler_) really should not care at all whether a value is set by a _literal_, a _variable_, or anything else.

If the intent is to refer to some other meaning of "literal", then a different word should be used.

eemeli (changes requested):
In addition to the inline comment, this text ought to more clearly set limits on _function handlers_ rather than implementations (which I note we never actually _**define**_).

spec/registry.md
aphillips: I do actually mean our definition of literal here. Of necessity, all values defined in the registry are string literals, either enumerated values or a reference to some serialization (like digit options or the various Unicode XXX Identifiers). The only way to have a non-literal type is by reference through a variable reference (which the second part of this PR addresses).

Would it be enough to linkify the term `literal`? Or do I need to expand on this, perhaps with an example?

aphillips:
@eemeli notes:
> In addition to the inline comment, this text ought to more clearly set limits on function handlers rather than implementations (which I note we never actually define).

This is a good point. I need to read the section and think about how/whether to modify here.

spec/registry.md
eemeli: I don't think we actually have a necessity of only defining _literal_ values here.

I find it much more useful to think of these restrictions and requirements as being placed on _function handlers_, which are processing the resolved values of operands and options.

With this viewpoint, the values that a function handler processes are resolved values that may contain a character sequence value, and we're requiring that certain character sequences are accepted for certain options. This would also match our spec text on literals:
https://github.com/unicode-org/message-format-wg/blob/f3344b5cea59e46aa8a0aa150904f53eb7690545/spec/formatting.md?plain=1#L222-L224

I find this distinction rather important, even if nitpicky. It also avoids introducing confusion about the _literals_ `short` and `|short|` needing to be treated equally, as they resolve to the same value.

If "character sequence" is a bit too verbose, we could define _string_ as a term.

aphillips: So I recast the requirements here, eliminating the "literal" part altogether. I think this actually does a better job of it now. See what you think.

spec/registry.md
```diff
@@ -38,12 +38,26 @@ by any version of this specification
...
-Implementations MAY _accept_ additional _option_ values for _options_ defined here.
+Implementations MAY _accept_, for _options_ defined in this specification,
+_option_ values which are not defined in this specification.
 However, such values might become defined with a different meaning in the future,
 including with a different, incompatible name
 or using an incompatible value space.
 Supporting implementation-specific _option_ values for **standard** or **optional** functions is NOT RECOMMENDED.
 
+Implementations MAY _accept_, for _operands_ or _option_ values defined in this specification,
+implementation-defined types.
```
eemeli, line 49:
```suggestion
Implementations MAY _accept_, for _operands_ or _options_ defined in this specification,
values with implementation-defined types.
```

spec/registry.md
```diff
@@ -38,12 +38,26 @@ by any version of this specification
...
+Such values can be useful to users in cases where local usage and support exists
+(including cases in which details vary from those defined by Unicode and CLDR).
+
+> For example, implementations are encouraged to _accept_ a native representation
+> for currency amounts as the _operand_ in the _function_ `:currency`.
+> Or a Java implementation might _accept_ a `java.time.chrono.Chronology` object
+> as a value for the date/time option `calendar`
```
eemeli, line 56: This is probably missing some word before the initial "Or".

macchiati (approved):

eemeli (approved):
Looks fine now. Added a couple of grammary nitpicks.

spec/registry.md
aphillips: See refactor of the example

mihnita (approved):

PR #945, opened 2024-11-18
Internationalize the `:math` example

aphillips:
Let's not embarrass ourselves, eh? 

The trailing selector should use the keyword `one` instead of the exact match `1`, in part because other languages need to explode (or contract, e.g. in languages such as `ja` or `zh`) the matrix.


macchiati (approved):

eemeli (approved):

mihnita (approved):
:-)

PR #946, opened 2024-11-18
Change standard/optional to required/recommended

aphillips:
Per 2024-11-18 teleconference


eemeli (approved):
The styling in places as `RECOMMENDED` or **RECOMMENDED** is probably unnecessary, as just RECOMMENDED ought to be enough.

macchiati (approved):

PR #948, opened 2024-11-18
Add hour12 date/time override option

eemeli:
This is the part of the changes originally proposed in #911 that was deemed as immediately mergeable on today's call.

Perhaps a little pre-emptively, it adopts the nomenclature to be introduced in #946.

This PR is expected to be followed up with separate PRs for the RECOMMENDED override options.

### DIFF
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -857,8 +857,10 @@ All other _operand_ values produce a _Bad Operand_ error.
 The `:datetime` function can use either the appropriate _style options_ 
 or can use a collection of _field options_ (but not both) to control the formatted 
 output.
+_Date/time override options_ can be combined with either _style options_ or _field options_.
 
-If both are specified, a _Bad Option_ error MUST be emitted
+If both _style options_ and _field options_ are specified,
+a _Bad Option_ error is emitted
 and a _fallback value_ used as the _resolved value_ of the _expression_.
 
 If the _operand_ of the _expression_ is an implementation-defined date/time type,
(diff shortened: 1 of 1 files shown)


aphillips (approved):

mihnita:
Looking at the whole file, it looks like the way "Date/time override options" is added to `:datetime` and `:time` is inconsistent.

`:datetime` has it in text, but not in the list of options:

> The :datetime function can use either the appropriate style options or can use a collection of field options (but not both) to control the formatted output. **Date/time override options** can be combined with either style options or field options.

Which is kind of hard to discover.

`:time` has it in the bulleted list of options


`:date` has it in the bulleted list of options

But it should not have it at all.
Because right now the only Date/time override option is "hour12"
Which does not apply to dates (they don't have an hour field).

macchiati (approved):

aphillips:
This meets fast-track requirements and was approved in the 2024-11-18 call previously.

mihnita:
> was approved in the 2024-11-18 call previously

Allowing `hour12` on `:date` is clearly wrong.
We do that just because was approved for fast-track?

The fix for that would be simple: remove it from `:date`

---

I can live with the option "hidden in text" for `:datetime`
It is a bit hard to discover, but not wrong.

spec/registry.md
```diff
@@ -995,6 +992,7 @@ The function `:date` has these _options_:
   - `long`
   - `medium` (default)
   - `short`
+- _Date/time override options_
```
mihnita, line 995: Remove this line.
The overrides right now only support `hour12`, which is not valid on dates.

PR #949, opened 2024-11-19
Add calendar and numberingSystem options to date/time

macchiati:
This is as per the discussion in the WG, separating them out from https://github.com/unicode-org/message-format-wg/pull/911

Two changes from that:
1. Used RECOMMENDED instead of Optional, as per https://github.com/unicode-org/message-format-wg/pull/946
2. I noticed also there that RECOMMENDED needs a maturity status, so added a placeholder for that, with **Proposed**. Since that is only in exploration/maintaining-registry.md, I presume it would be removed for release.

### DIFF
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -1112,7 +1112,6 @@ the functions `:datetime` and `:time`:
 
 The following _options_ and their values are RECOMMENDED to be available on
 the functions `:datetime`, `:date`, and `:time`.
-The maturity level for each is **Proposed**.
 
 - `calendar`
   - valid [Unicode Calendar Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeCalendarIdentifier)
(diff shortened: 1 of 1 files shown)


aphillips:
I resolved the conflict. In doing so, I reordered the proposed options below the REQUIRED one and rephrased the text slightly to take advantage of the fact that REQUIRED and RECOMMENDED are 2119 keywords. Please have a look.

eemeli (changes requested):
At least the `calendar` and `numberingSystem` mentions should also be removed from this note: https://github.com/unicode-org/message-format-wg/blob/960f8003cc9d725b4b473877e3391abf101aafe4/spec/registry.md?plain=1#L952-L964

aphillips:
@eemeli I will make a separate PR to remove the note.

mihnita (commented):
For some reason I can review, but can't approve.
My only button is "Submit review"

So: approved

PR #950, opened 2024-11-19
Create option `timezone` as Proposed REQUIRED option

aphillips:
This still has the ABNF for time zone ID, which should disappear as soon as I can replace it with something formal.

It's fair game to discuss whether this should be RECOMMENDED instead.

### DIFF
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -1115,18 +1115,11 @@ This option currently has a Maturity Level of **Proposed**.
 
 - `timeZone`
   - A valid time zone identifier
-    (see [BCP175](https://www.rfc-editor.org/bpc/bpc175))
+    (see [TZDB](https://www.iana.org/time-zones)
+    and [LDML](https://www.unicode.org/reports/tr35/tr35-dates.html#Time_Zone_Names)
+    for information on identifiers)
   - `local`
   - `UTC`
-  
-A time zone identifier is well-formed if it matches `tzId` in the following ABNF:
-```abnf
-tzId   = tzPath / tzEtc
-tzPath = tzPart 1*("/" tzPart)
-tzPart = tzWord *("_" tzWord)
-tzWord = (%x41-5A) *(%x61-7A) ; Uppercase ASCII letter followed by lowercase letters
-tzEtc  = "Etc/GMT" ("+" / "-") 1*2DIGIT
-```
   
 > [!NOTE]
 > The value `local` permits a _message_ to convert a date/time value
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -1114,9 +1114,19 @@ the functions `:datetime` and `:time`:
   - `true`
   - `false`
 
-The following _option_ and its values will be REQUIRED to be available on
+The following _options_ and their values are RECOMMENDED to be available on
 the functions `:datetime`, `:date`, and `:time`.
-This option currently has a Maturity Level of **Proposed**.
+
+- `calendar`
+  - valid [Unicode Calendar Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeCalendarIdentifier)
+- `numberingSystem`
+  - valid [Unicode Number System Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeNumberSystemIdentifier)
+
+The following _option_ and its values are **Proposed** for
+inclusion in the next release of this specification but have not yet been
+finalized.
+If accepted, implementations could be REQUIRED to make this _option_
+available in the functions `:datetime`, `:date`, and `:time`.
 
 - `timeZone`
   - A valid time zone identifier
@@ -1131,13 +1141,3 @@ This option currently has a Maturity Level of **Proposed**.
 > into a [floating](https://www.w3.org/TR/timezone/#floating) time value
 > (sometimes called a _plain_ or _local_ time value) by removing
 > the association with a specific time zone.
-
-The following _options_ and their values are RECOMMENDED to be available on
-the functions `:datetime`, `:date`, and `:time`.
-
-- `calendar`
-  - valid [Unicode Calendar Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeCalendarIdentifier)
-- `numberingSystem`
-  - valid [Unicode Number System Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeNumberSystemIdentifier)
-
-
--- a/spec/registry.md
+++ b/spec/registry.md
@@ -1109,6 +1109,31 @@ the functions `:datetime` and `:time`:
   - `true`
   - `false`
 
+The following _option_ and its values will be REQUIRED to be available on
+the functions `:datetime`, `:date`, and `:time`.
+This option currently has a Maturity Level of **Proposed**.
+
+- `timezone`
+  - A valid time zone identifier
+    (see [BCP175](https://www.rfc-editor.org/bpc/bpc175))
+  - `local`
+  - `UTC`
+  
+A time zone identifier is well-formed if it matches `tzId` in the following ABNF:
+```abnf
+tzId   = tzPath / tzEtc
+tzPath = tzPart 1*("/" tzPart)
+tzPart = tzWord *("_" tzWord)
+tzWord = (%x41-5A) *(%x61-7A) ; Uppercase ASCII letter followed by lowercase letters
+tzEtc  = ("Etc/" ("UTC" / "GMT" (("+" / "-") 1*2DIGIT))
+```
+  
+> [!NOTE]
+> The value `local` permits a _message_ to convert a date/time value
+> into a [floating](https://www.w3.org/TR/timezone/#floating) time value
+> (sometimes called a _plain_ or _local_ time value) by removing
+> the association with a specific time zone.
+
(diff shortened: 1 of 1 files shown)


spec/registry.md
```diff
@@ -1109,6 +1109,31 @@ the functions `:datetime` and `:time`:
...
+  - `local`
+  - `UTC`
+  
+A time zone identifier is well-formed if it matches `tzId` in the following ABNF:
+```abnf
+tzId   = tzPath / tzEtc
+tzPath = tzPart 1*("/" tzPart)
+tzPart = tzWord *("_" tzWord)
+tzWord = (%x41-5A) *(%x61-7A) ; Uppercase ASCII letter followed by lowercase letters
+tzEtc  = ("Etc/" ("UTC" / "GMT" (("+" / "-") 1*2DIGIT))
```
macchiati, line 1128: If you leave the ABNF in, then you should should adjust to remove the UTC here
```suggestion
tzEtc  = "Etc/GMT" ("+" / "-") 1*2DIGIT
```

spec/registry.md
```diff
@@ -1109,6 +1109,31 @@ the functions `:datetime` and `:time`:
   - `true`
   - `false`
 
+The following _option_ and its values will be REQUIRED to be available on
+the functions `:datetime`, `:date`, and `:time`.
+This option currently has a Maturity Level of **Proposed**.
+
+- `timezone`
+  - A valid time zone identifier
+    (see [BCP175](https://www.rfc-editor.org/bpc/bpc175))
```
eemeli, line 1118: This link is broken (should probably be https://www.rfc-editor.org/info/bcp175), but looking at the linked text (which in #911 I'd just copied from the previous), it states this:
> The normative reference for [time zone names] is the TZ Database [TZDB].

where the onward reference is to
> Eggert, P. and A. Olson, "Sources for Time Zone and Daylight Saving Time Data", 1987, <ftp://ftp.iana.org/tz/code/tz-link.htm>.

As accessing FTP links isn't as easy as it once was, I may only presume that https://data.iana.org/time-zones/tz-link.html is a valid current corresponding reference, but it does not establish anything clear about how time zone names are defined.

For comparison, the JS reference is more current and much more clear in its definition of the appropriate authority (IANA), and rather exactly defines how to look up the available names: https://tc39.es/ecma402/#sec-use-of-iana-time-zone-database

spec/registry.md
```diff
@@ -1109,6 +1109,31 @@ the functions `:datetime` and `:time`:
...
+  - `local`
+  - `UTC`
+  
+A time zone identifier is well-formed if it matches `tzId` in the following ABNF:
```
eemeli, line 1122: As discussed during the call, it is not appropriate for us to include this ABNF, as we cannot require that IANA or any other external authority will fit their valid values within it.

spec/registry.md
```diff
@@ -1109,6 +1109,31 @@ the functions `:datetime` and `:time`:
...
+tzPart = tzWord *("_" tzWord)
+tzWord = (%x41-5A) *(%x61-7A) ; Uppercase ASCII letter followed by lowercase letters
+tzEtc  = ("Etc/" ("UTC" / "GMT" (("+" / "-") 1*2DIGIT))
+```
+  
+> [!NOTE]
+> The value `local` permits a _message_ to convert a date/time value
+> into a [floating](https://www.w3.org/TR/timezone/#floating) time value
+> (sometimes called a _plain_ or _local_ time value) by removing
+> the association with a specific time zone.
```
eemeli, line 1135: I find this very confusing. From the name, my presumption would be that "local" means the formatter's local system time, but that's a time with an explicit timezone, while this note is indicating that it ought to mean a datetime not associated with any timezone.

With something like a literal date as operand, I presume that these would give the same results, but if the operand is an implementation-defined datetime with a timezone different from the local system time, it's not at all clear what date might get formatted.

I would find it very useful to have an example provided for a use case where `timeZone=local` would be required to get desirable behaviour, as opposed to just leaving out the option entirely.

Also, just for clarity, are there any use cases for floating datetimes that include anything more precise than a day being formatted?

eemeli (changes requested):

spec/registry.md
aphillips: The link has a typo (it should s/bpc/bcp/). We could link the TZDB directly and that's probably better than the BCP. As Mark pointed out, the real reference is buried inside the datafiles.

The JS reference is a good one and describes a similar workaround for IANA's cross-border zone merging scheme to that of others.

As mentioned elsewhere, I don't want to get into the business of defining all these things. We have the problem that CLDR/ICU "do stuff" with time zones and identifiers and that stuff, while not incompatible with what JS does, is more complex. 

I think what I'll do here is reference IANA's page and also LDML. In practice, we don't need to define "well-formed" or "valid" directly.

spec/registry.md
aphillips: Agreed.

spec/registry.md
```diff
@@ -1109,6 +1109,31 @@ the functions `:datetime` and `:time`:
   - `true`
   - `false`
 
+The following _option_ and its values will be REQUIRED to be available on
+the functions `:datetime`, `:date`, and `:time`.
+This option currently has a Maturity Level of **Proposed**.
```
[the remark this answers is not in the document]
aphillips: The point of that particular design document is to address maintenance of the ~~registry~~ function sets, which we suddenly find ourselves in the midst of doing 😉. 

I can remove this line. A better course of action would be to add a legend to the top of registry.md that says what the maturity levels are. Probably the only maturity levels that should show are "Proposed" and "Deprecated" ("Accepted" applies to a proposal that hasn't been released yet, i.e. you're not reading it).

spec/registry.md
```diff
@@ -1109,6 +1109,31 @@ the functions `:datetime` and `:time`:
   - `true`
   - `false`
 
+The following _option_ and its values will be REQUIRED to be available on
```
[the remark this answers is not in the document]
aphillips: What's the reason to make this option merely recommended?

aphillips:
Note: in resolving the conflict with the calendar/numberingSystem PR, I moved the note about default values above the options.

spec/registry.md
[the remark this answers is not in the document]
eemeli: It introduces a dependency on timezone data. Without the option, a `:datetime` formatter doesn't need to know how timezones work, as it can "just" format the input values. With the option, the formatter is expected to know how timezones relate to each other, and when e.g. daylight savings changes happen.

This is particularly relevant as the formatted output often does not include a timezone indicator.

I'm concerned here about the bar we're setting for a minimal valid MF2 implementation, which I don't want to burden with timezone data, esp. given how frequently and unpredictably it changes. Marking this as RECOMMENDED would give a signal to implementers and users that this should not be relied on without checking the capabilities of the MF2 implementation.

spec/registry.md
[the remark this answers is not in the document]
eemeli: Until/unless we have an internal reference for what "Proposed" means, this statement ought to be replaced with text that describes its relevant effects on `timeZone`. If/when later adding a general definition, we can replace the text here with a reference.

eemeli (commented):
This is getting really close. To be explicit, with the following changes I'd be happy to approve this:
1. Make it recommended rather than required.
2. Drop the maturity level reference.
3. Drop `local` as a value. Yes, it's important, but it's also confusing and we should take some time to iterate on it until we add it in later.

spec/registry.md
```diff
@@ -1103,13 +1103,35 @@ override values set by the current locale,
...
+the functions `:datetime`, `:date`, and `:time`.
+This option currently has a Maturity Level of **Proposed**.
+
+- `timeZone`
+  - A valid time zone identifier
+    (see [TZDB](https://www.iana.org/time-zones)
+    and [LDML](https://www.unicode.org/reports/tr35/tr35-dates.html#Time_Zone_Names)
+    for information on identifiers)
+  - `local`
+  - `UTC`
```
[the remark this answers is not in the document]
aphillips: I call it out for the reason JS calls it out: _technically_ it is not a valid value and is added by exception. `UTC` is also special, in that it is generally the pivot for other zones as well as the zone that applies to most canonical incremental time values (`Date` objects in many languages and platforms is in terms of millis or nanos from the UNIX epoch _in UTC_).

By listing it here, we require that all implementations accept that specific literal. All of the other literals are up to the implementation and its capabilities.

spec/registry.md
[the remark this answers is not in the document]
aphillips: Fair enough. I will replace this with text that describes the specific status.

spec/registry.md
```diff
@@ -1118,7 +1122,22 @@ the functions `:datetime`, `:date`, and `:time`.
 - `numberingSystem`
   - valid [Unicode Number System Identifier](https://cldr-smoke.unicode.org/spec/main/ldml/tr35.html#UnicodeNumberSystemIdentifier)
 
+The following _option_ and its values are **Proposed** for
+inclusion in the next release of this specification but have not yet been
+finalized.
+If accepted, implementations could be REQUIRED to make this _option_
+available in the functions `:datetime`, `:date`, and `:time`.
```
eemeli, line 1129: With this clarification, I'm fine with merging the currently proposed text as it does not impose any restrictions or requirements on implementations, effectively kicking the can down the road.

eemeli (approved):

macchiati (approved):

PR #965, opened 2024-11-28
[editorial] Reorganise Readmes & spec introduction

eemeli:
While looking for a place for an MF2 explainer and looking at what text we had already available, I think the editorial reorg proposed here would clarify things quite a bit. The end result here would be (links to this PR's versions):
- [`/README.md`](https://github.com/unicode-org/message-format-wg/blob/editorial-text-moves/README.md) describes the working group and its processes
- [`/spec/README.md`](https://github.com/unicode-org/message-format-wg/blob/editorial-text-moves/spec/README.md) holds the ToC for the spec, along with a general introduction to MF2
- [`/spec/intro.md`](https://github.com/unicode-org/message-format-wg/blob/editorial-text-moves/spec/intro.md) is the actual "Introduction" section of the spec

I believe this arrangement would be clearer to a newcomer, and keep separate concepts more clearly separate. Note that as a result, the `/spec/README.md` file would not be considered a part of the actual spec.

The only changes to the text that are made here are fixes to links due to moved contents, and a small cleanup of the ToC (which had bad links and was missing the appendices), so I'm marking this as `editorial` and `fast-track`.

If this rearrangement is accepted, it becomes much easier to iterate on `/spec/README.md` as an explainer for the spec.

### DIFF
--- a/spec/README.md
+++ b/spec/README.md
@@ -7,19 +7,15 @@
    1. [Terminology and Conventions](intro.md#terminology-and-conventions)
    1. [Stability Policy](intro.md#stability-policy)
 1. [Syntax](syntax.md)
-   1. [Productions](syntax.md#productions)
-   1. [Tokens](syntax.md#tokens)
    1. [`message.abnf`](message.abnf)
+1. [Formatting](formatting.md)
 1. [Errors](errors.md)
-   1. [Error Handling](errors.md#error-handling)
-   1. [Syntax Errors](errors.md#syntax-errors)
-   1. [Data Model Errors](errors.md#data-model-errors)
-   1. [Resolution Errors](errors.md#resolution-errors)
-   1. [Message Function Errors](errors.md#message-function-errors)
 1. [Default Function Registry](registry.md)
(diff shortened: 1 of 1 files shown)


ryzokuken (approved):

README.md
```diff
@@ -11,7 +11,7 @@ The Message Format Working Group (MFWG) is tasked with developing an industry st
 
 ## MessageFormat 2 Technical Preview
 
-The MessageFormat 2 specification is a new part of
+The [MessageFormat 2 specification](./spec/) is a new part of
```
aphillips, line 14: Technically `/spec/` is not the thing in tech preview. Instead, it's actually in the TR35 report.

aphillips (commented):
I like this in principle, although I am reticent about putting anything under the spec directory that isn't the specification. Let's discuss in the upcoming call.

aphillips:
I have to prepare the spec and the repo for 46.1. These changes are all editorial. I'm going to approve and merge them so that I can make a separate PR with my necessary changes.

aphillips (approved):

PR #968, opened 2024-12-06
Clarification to default bidi strategy

catamorphism:
I'm submitting this PR to test my understanding of the spec. As I understand it (having read https://github.com/unicode-org/message-format-wg/issues/918 ), the directionality of `fmt` _could_ be determined by introspection on `fmt`, but the implementation is not required to do that introspection. Rather, the resolved value of the expression has both an "input directionality" (determined by the `u:dir` attribute) and an "output directionality" (determined by the function handler for the annotation, or with a default value for placeholders without annotation.) The "output directionality" is what is meant by "the directionality of `fmt`" in the original text. 

I was confused because point 5 of  https://github.com/unicode-org/message-format-wg/issues/918 implies that no introspection is necessary, if I'm understanding correctly; but "the directionality of `fmt`" seems to imply that the directionality has to be determined from the formatted output, rather than from the metadata of that formatted output that is created by a function (that is, the "resolved value").

If I'm still confused, hopefully there's another edit that would make it clearer.

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -941,7 +941,7 @@ The _Default Bidi Strategy_ is defined as follows:
       1. Append `fmt` to `out`.
    1. Else:
       1. Let `fmt` be the formatted string representation of the _resolved value_ of `part`.
-      1. Let `dir` be the directionality of `fmt`,
+      1. Let `dir` be `DIR(part)`,
          one of « `'LTR'`, `'RTL'`, `'unknown'` », with the same meanings as for `msgdir`.
          > [!NOTE]
          > _Resolved values_ need to track two different pieces of metadata about directionality:
@@ -972,5 +972,21 @@ The _Default Bidi Strategy_ is defined as follows:
          1. Append `fmt` to `out`.
          1. Append U+2069 POP DIRECTIONAL ISOLATE to `out`.
 1. Emit `out` as the formatted output of the message.
+
+The auxiliary function `DIR`, which maps an _expression_ to one of
+« `'LTR'`, `'RTL'`, `'unknown'` », is expected to be defined using the
+_resolved value_ of the _expression_. An implementation can use
+a representation of _resolved values_ that tracks two different
+pieces of metadata about directionality: the `isolate` flag used in step 2(iii),
+and the directionality itself. Each _function handler_ can have its own means
+for determining the directionality annotation on the _resolved value_ it returns.
+Alternately, an implementation could simply compute `DIR(exp)` based on the
+locale.
+
+> [!NOTE]
+> `DIR(exp)` SHOULD NOT be determined by introspecting
+> the character sequence in the formatted string representation
+> of the resolved value of `exp`.
+
 
 
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -153,7 +153,7 @@ and different implementations MAY choose to perform different levels of resoluti
 >   getValue(): unknown
 >   resolvedOptions(): { [key: string]: MessageValue }
 >   selectKeys(keys: string[]): string[]
->   directionality() : Dir // where Dir is an enum type, one of « `'LTR'`, `'RTL'`, `'unknown'` »
+>   directionality(): 'LTR' | 'RTL' | 'unknown'
 >   isolate(): boolean
 > }
 > ```
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -131,6 +131,13 @@ identifies not only the name of the external input value,
 but also the _variable_ to which the _resolved value_ of the _variable-expression_ is bound.
 
 In a _pattern_, the _resolved value_ of an _expression_ or _markup_ is used in its _formatting_.
+To support the _Default Bidi Strategy_,
+the _resolved value_ of each _expression_ _placeholder_
+SHOULD include information about the directionality
+of its formatted string representation,
+as well as a flag to indicate whether
+its formatted representation requires isolation
+from the surrounding text.
 
 The form that _resolved values_ take is implementation-dependent,
 and different implementations MAY choose to perform different levels of resolution.
@@ -965,18 +972,12 @@ The _Default Bidi Strategy_ is defined as follows:
 1. Emit `out` as the formatted output of the message.
 
 > [!NOTE]
-> A _resolved value_ is derived from an _expression_
-> together with a _formatting context_.)
-> An implementation can use
-> a representation of _resolved values_ that tracks
-> everything needed to determine the directionality
+> As mentioned in the "Resolved Values" section,
+> the representation of a _resolved value_
+> can track everything needed
+> to determine the directionality
 > of the formatted string representation
 > of a _resolved value_.
-> This can be accomplished by incorporating
-> the `isolate` flag used in step 2(iii),
-> as well as the separate directionality annotation
-> (one of « `'LTR'`, `'RTL'`, `'unknown'` »),
-> into the representation of _resolved values_.
 > Each _function handler_ can have its own means
 > for determining the directionality annotation
 > on the _resolved value_ it returns.
@@ -984,7 +985,7 @@ The _Default Bidi Strategy_ is defined as follows:
 > determine directionality
 > based on the locale.
 
-> [!NOTE]
+> [!IMPORTANT]
 > Directionality SHOULD NOT be determined by introspecting
 > the character sequence in the formatted string representation
 > of `resval`.
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -940,19 +940,10 @@ The _Default Bidi Strategy_ is defined as follows:
          Note that this is normally the empty string.
       1. Append `fmt` to `out`.
    1. Else:
-      1. Let `fmt` be the formatted string representation of the _resolved value_ of `part`.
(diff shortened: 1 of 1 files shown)


aphillips:
> I was confused because point 5 of https://github.com/unicode-org/message-format-wg/issues/918 implies that no introspection is necessary, if I'm understanding correctly; but "the directionality of fmt" seems to imply that the directionality has to be determined from the formatted output, rather than from the metadata of that formatted output that is created by a function (that is, the "resolved value").

Introspection of the formatted string makes no sense, because introspection would take the form of scanning for the first strong character (plus a bunch of details from UAX9) and this _exactly_ identical to what using FSI gets you for free. Only now you are responsible for bugs. We don't prohibit your doing this, but there is no value to be derived from doing so.

What the strategy allows is for the function handler to have some say over the direction used. This will usually be the result of the function handler calling a function like `uloc_isRightToLeft` \[[doc](https://unicode-org.github.io/icu-docs/apidoc/released/icu4c/uloc_8h.html#aacb7279682b9b063122bfcdbb4cab28d)\] on the resolved locale of the formatter.

There's a fallback for what the direction of the placeholder is:

`u:dir` > `fmt` dir > `msgdir`

The `fmt` directionality might be considering the directionality of the resolved value of the operand (if there is an operand and if said operand has a direction) or it might just compute it as mentioned above.

spec/formatting.md
```diff
@@ -936,7 +936,7 @@ The _Default Bidi Strategy_ is defined as follows:
    right-to-left directionality, and to the message's directionality not being known.
 1. For each _expression_ `exp` in _pattern_:
    1. Let `fmt` be the formatted string representation of the _resolved value_ of `exp`.
-   1. Let `dir` be the directionality of `fmt`,
+   1. Let `dir` be the directionality of the _resolved value_ of `exp`,
       one of « `'LTR'`, `'RTL'`, `'unknown'` », with the same meanings as for `msgdir`.
```
aphillips, line 940:
```suggestion
      1. Let `dir` be the directionality of `fmt`,
         one of « `'LTR'`, `'RTL'`, `'unknown'` », with the same meanings as for `msgdir`.
      > [!NOTE]
      > `dir` is usually computed from the locale of the formatter
      > or from other information, such as the directionality of the _resolved value_ of `exp`,
      > and not by introspecting the character sequence in `fmt`
```

macchiati (approved):

spec/formatting.md
catamorphism: I didn't take your suggestion exactly as-is, but let me know what you think about [58afcfa](https://github.com/unicode-org/message-format-wg/pull/968/commits/58afcfa7a2acc42e54a9341db473e0e1b94e3391).

spec/formatting.md
```diff
@@ -936,7 +936,7 @@ The _Default Bidi Strategy_ is defined as follows:
    right-to-left directionality, and to the message's directionality not being known.
 1. For each _expression_ `exp` in _pattern_:
    1. Let `fmt` be the formatted string representation of the _resolved value_ of `exp`.
-   1. Let `dir` be the directionality of `fmt`,
+   1. Let `dir` be `DIR(exp)`,
```
eemeli, line 939: The direction cannot be defined only from `exp`, as its _resolved value_ also depends on the formatting context.

I'm not convinced that defining `DIR()` makes this algorithm clearer, as in the immediate context we're also reading "the formatted string representation of the _resolved value_ of `exp`" as well as "the `u:dir` _option_ of the _resolved value_ of `exp`"

Something like this might be better:
```
   1. Let `resval` be the _resolved value_ of `exp`.
   1. Let `fmt` be the formatted string representation of `resval`.
   1. Let `dir` be the directionality of `resval`,
      one of « `'LTR'`, `'RTL'`, `'unknown'` », with the same meanings as for `msgdir`.
   1. Let the boolean value `isolate` be
      True if the `u:dir` _option_ of `resval` has a value other than `'inherit'`,
      or False otherwise.
```

And then reactoring the "The auxiliary function `DIR`..." paragraph into a non-normative note.

spec/formatting.md
```diff
@@ -957,4 +957,20 @@ The _Default Bidi Strategy_ is defined as follows:
          prefix `fmt` with U+2068 FIRST STRONG ISOLATE
          and postfix it with U+2069 POP DIRECTIONAL ISOLATE.
 
+The auxiliary function `DIR`, which maps an _expression_ to one of
+« `'LTR'`, `'RTL'`, `'unknown'` », is expected to be defined using the
+_resolved value_ of the _expression_. An implementation can use
```
eemeli, line 962: Please use [semantic line breaks](https://sembr.org/), they make reviewing and editing markdown much easier.

eemeli (changes requested):

spec/formatting.md
catamorphism: Tried to refine the placement of the line breaks; if there's anything specific, let me know.

spec/formatting.md
```diff
@@ -963,4 +964,30 @@ The _Default Bidi Strategy_ is defined as follows:
...
+> the `isolate` flag used in step 2(iii),
+> as well as the separate directionality annotation
+> (one of « `'LTR'`, `'RTL'`, `'unknown'` »),
+> into the representation of _resolved values_.
+> Each _function handler_ can have its own means
+> for determining the directionality annotation
+> on the _resolved value_ it returns.
+> Alternately, an implementation could simply
+> determine directionality
+> based on the locale.
```
eemeli, line 985: On further consideration, does this perhaps belong more appropriately in the section defining _resolved values_?

Specifically, I think we ought to expand a bit the paragraph starting on [line 133](https://github.com/unicode-org/message-format-wg/blob/a19b3df18c3802bd13c1dde077efd29a95339e7a/spec/formatting.md?plain=1#L133) with something like this:

```diff
 In a _pattern_, the _resolved value_ of an _expression_ or _markup_ is used in its _formatting_.
+To account for the _Default Bidi Strategy_,
+the _resolved value_ of each _expression_ _placeholder_
+SHOULD include information about the directionality of its formatted string representation,
+as well as information about whether its formatted representation
+requires isolation from the surrounding text.
```

With that in place, this note could be simplified quite a bit, or even left out entirely.

spec/formatting.md
```diff
@@ -963,4 +964,30 @@ The _Default Bidi Strategy_ is defined as follows:
...
+
+> [!NOTE]
+> Directionality SHOULD NOT be determined by introspecting
+> the character sequence in the formatted string representation
+> of `resval`.
+
+
```
eemeli, line 992: We should not say SHOULD NOT in a non-normative note. If this is intended to be normative, it should be marked as `[!IMPORTANT]`, or not be a note.

```suggestion
Directionality SHOULD NOT be determined by introspecting
the character sequence in the formatted string representation
of `resval`.
```

eemeli (changes requested):

spec/formatting.md
aphillips: This is pretty good, but I wouldn't limit it to the default bidi strategy, nor would I necessarily limit it to placeholders (other expressions might have a directionality so that the direction is commutative across assignments). Perhaps:

> The _resolved value_ of each _expression_
> SHOULD include the [string direction](https://www.w3.org/TR/i18n-glossary/#dfn-string-direction)
>  of its formatted string representation,
> as well as information about whether the message author
> wants to suppress or require isolation from the surrounding _message_ text,
> such as is found in the _Default Bidi Strategy_.

eemeli: Note that _resolved value_ is an implementation-internal thing, and is therefore not expected to have public visibility during formatting. This means that attaching any normative requirements on it ought to be explicitly connected to some spec-defined consumer of that information, i.e. the _Default Bidi Strategy_.

catamorphism: I made these changes. Also modified the example `MessageValue` definition to show what the directionality methods could look like.

spec/formatting.md
catamorphism: Changed to `[!IMPORTANT]`.

spec/formatting.md
```diff
@@ -131,6 +131,13 @@ identifies not only the name of the external input value,
 but also the _variable_ to which the _resolved value_ of the _variable-expression_ is bound.
 
 In a _pattern_, the _resolved value_ of an _expression_ or _markup_ is used in its _formatting_.
+To support the _Default Bidi Strategy_,
+the _resolved value_ of each _expression_ _placeholder_
```
eemeli, line 135: As per @aphillips's suggestion.
```suggestion
the _resolved value_ of each _expression_
```

spec/formatting.md
```diff
@@ -146,6 +153,8 @@ and different implementations MAY choose to perform different levels of resoluti
 >   getValue(): unknown
 >   resolvedOptions(): { [key: string]: MessageValue }
 >   selectKeys(keys: string[]): string[]
+>   directionality() : Dir // where Dir is an enum type, one of « `'LTR'`, `'RTL'`, `'unknown'` »
```
eemeli, line 156: We don't need to define a new type in order to communicate the intent here; this is TypeScript after all.
```suggestion
>   directionality(): 'LTR' | 'RTL' | 'unknown'
```

eemeli (approved):

PR #969, opened 2024-12-06
In bidi default strategy, make steps consistent with each other

catamorphism:
I'm not sure what "If `msgdir` is `LTR` in the formatted output" meant, but this is my best guess; "let `fmt` be itself" was inconsistent with the other steps, which refer to "In the formatted output..." rather than setting `fmt`.

I think this is a normative change, since the spec was basically saying ~~there is no formatted output~~ that no change is made to the formatted output if step 2(iv)(a) applies.


spec/formatting.md
```diff
@@ -942,9 +942,9 @@ The _Default Bidi Strategy_ is defined as follows:
       True if the `u:dir` _option_ of the _resolved value_ of `exp` has a value other than `'inherit'`,
       or False otherwise.
    1. If `dir` is `'LTR'`:
-      1. If `msgdir` is `'LTR'` in the formatted output
+      1. If `msgdir` is `'LTR'`
          and `isolate` is False,
-         let `fmt` be itself
+         let the formatted output be `fmt` itself.
```
aphillips, line 947: This change implies replacing _all_ of the formatted output with `fmt`, not just replacing `exp` with the contents of `fmt`. So I would not make this change.

spec/formatting.md
```diff
@@ -942,9 +942,9 @@ The _Default Bidi Strategy_ is defined as follows:
       True if the `u:dir` _option_ of the _resolved value_ of `exp` has a value other than `'inherit'`,
       or False otherwise.
    1. If `dir` is `'LTR'`:
-      1. If `msgdir` is `'LTR'` in the formatted output
+      1. If `msgdir` is `'LTR'`
```
aphillips, line 945: I think this is a good change, in that it is clearer. `msgdir` is whatever it is.

aphillips:
Adding `Agenda+` so that we can discuss bidi on Monday.

spec/formatting.md
catamorphism: Ah, yes. Fixed in [39e9056](https://github.com/unicode-org/message-format-wg/pull/969/commits/39e9056592d0aa4bfe8da978ee646f0f6737e751)

spec/formatting.md
```diff
@@ -942,9 +942,9 @@ The _Default Bidi Strategy_ is defined as follows:
...
          and `isolate` is False,
-         let `fmt` be itself
+         append `fmt` to the formatted output.
```
aphillips, line 947: The other steps don't do append. They talk about what `fmt` contains (e.g. "prefix `fmt` with...")

The change we should be making should actually be earlier. Something like:

> 2. For each expression `exp` in _pattern_:
>    i. Let `fmt` be the formatted string representation of the _resolved value_ of `exp` to be appended to the formatted output.

aphillips (changes requested):

catamorphism:
As discussed in today's meeting, I rewrote the text to specify a function from expressions to formatted strings.

spec/formatting.md
```diff
@@ -928,7 +928,8 @@ Implementations MAY provide other _bidirectional isolation strategies_.
 
 Implementations MAY supply a _bidirectional isolation strategy_ that performs no processing.
 
-The _Default Bidi Strategy_ is defined as follows:
+The _Default Bidi Strategy_ is defined as a function `B` from expressions
```
aphillips, line 931: Linkify _expressions_.

This is weird? Where does `B` come from? Why is it a "function"? The text above this talks about "bidirectional isolation strategies". I think this needs to be different.

Perhaps:

> The _Default Bidi Strategy_ is defined as follows:
>
> 1. Let `out` be an empty string to contain the formatted output of the _message_
> 1. Let `msgdir`...
> 1. For each _expression_ `exp` in the _pattern_:
>    1. If `exp` is a plain literal (text), append `exp` to `out`.
>    1. If `exp` is an _placeholder_, let `fmt` be the formatted string representation of
>       the _resolved value_ of `exp`.
>    1. Let `dir` be the directionality of `fmt`...
>     // etc.
>     1. Append `fmt` to `out`.
> 1. Emit `out` as the formatted output of the _message_.

Note that we don't have a name for the non-placeholder bits of text in a pattern. I call them expressions here, but they aren't really expressions.

It would be even better if the section on formatting just above here defined how to make the string output. The "bidi strategy" then could be reference in that set of instructions (_"apply the appropriate bidirectional isolation strategy to the contents of expression"_)

catamorphism: I rewrote this imperatively instead of as a function. I also did a rewrite to call the components of a pattern a "part", and distinguish between text parts and placeholder parts.

spec/formatting.md
```diff
@@ -930,31 +930,38 @@ Implementations MAY supply a _bidirectional isolation strategy_ that performs no
...
-         and postfix it with U+2069 POP DIRECTIONAL ISOLATE.
+1. For each part `part` in _pattern_:
+   1. If `part` is a plain literal (text) part, append `part` to `out`.
    1. Else:
-      1. In the formatted output,
-         prefix `fmt` with U+2068 FIRST STRONG ISOLATE
-         and postfix it with U+2069 POP DIRECTIONAL ISOLATE.
+       1. Assert `part` is a _placeholder_.
+       1. Let `exp` be `part`.
+       1. Let `fmt` be the formatted string representation of the _resolved value_ of `exp`.
```
aphillips, line 943: This exposes a problem, in that a _placeholder_ can be an _expression_ or _markup_. We need to handle _markup_ here. Perhaps:

```suggestion
       1. Assert `part` is a _placeholder_.
       1. If `part` is _markup_, append the _resolved value_ of `part` to `out`.
          Note that this is normally the empty string.
       1. Else:
          1. Let `fmt` be the formatted string representation of the _resolved value_ of `part`.
```

spec/formatting.html
```diff
@@ -0,0 +1,1048 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
aphillips, line 1: Remove this file.

aphillips (changes requested):

spec/formatting.html
catamorphism: Oops, fixed.

aphillips (approved):

spec/formatting.md
```diff
@@ -930,31 +930,40 @@ Implementations MAY supply a _bidirectional isolation strategy_ that performs no
...
+1. For each part `part` in _pattern_:
+   1. If `part` is a plain literal (text) part, append `part` to `out`.
    1. Else:
-      1. In the formatted output,
-         prefix `fmt` with U+2068 FIRST STRONG ISOLATE
-         and postfix it with U+2069 POP DIRECTIONAL ISOLATE.
+       1. Assert `part` is a _placeholder_.
+       1. If `part` is _markup_, append the _resolved value_ of `part` to `out`.
+          Note that this is normally the empty string.
+       1. Else:
```
eemeli, line 944: This could be structured a bit differently to specify the assertion, and drop one indentation level from everything that follows. Also, let's not presume that the resolved value of markup is directly a string.
```
    1. Else if `part` is a _markup_ _placeholder_:
        1. Let `fmt` be the formatted string representation of the _resolved value_ of `part`.
           Note that this is normally the empty string.
        1. Append `fmt` to `out`.
    1. Else:
        1. Assert that `part` is an _expression_.
        1. ...
```

spec/formatting.md
```diff
@@ -930,31 +930,40 @@ Implementations MAY supply a _bidirectional isolation strategy_ that performs no
...
+          1. Let `fmt` be the formatted string representation of the _resolved value_ of `part`.
+          1. Let `dir` be the directionality of `fmt`,
+             one of « `'LTR'`, `'RTL'`, `'unknown'` », with the same meanings as for `msgdir`.
+          1. Let the boolean value `isolate` be
+             True if the `u:dir` _option_ of the _resolved value_ of `exp` has a value other than `'inherit'`,
+          or False otherwise.
+          1. If `dir` is `'LTR'`:
+              1. If `msgdir` is `'LTR'`
+                 and `isolate` is False,
+                 append `fmt` to `out`.
```
eemeli, line 954: It's clearer if we stick to the same style.

```suggestion
              1. If `msgdir` is `'LTR'` and `isolate` is False:
                  1. Append `fmt` to `out`.
```

eemeli (approved):
This is an improvement. See inline for a few further nitpicky suggestions.

spec/formatting.md
catamorphism: Done as part of [e9afb9d](https://github.com/unicode-org/message-format-wg/pull/969/commits/e9afb9d592c8983028977f3e3570f1a97222e1f6)

spec/formatting.md
catamorphism: Done as part of [e9afb9d](https://github.com/unicode-org/message-format-wg/pull/969/commits/e9afb9d592c8983028977f3e3570f1a97222e1f6)

aphillips (approved):

PR #971, opened 2024-12-08
Add namespaces to example non-default functions

eemeli:
Given that our spec includes this: https://github.com/unicode-org/message-format-wg/blob/7c1f1a4af4aead387ca6aec48fffc2c6e2191c4d/spec/registry.md?plain=1#L33-L34

We should apply that same guideline to our own examples which use non-default functions.

### DIFF
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -814,14 +814,14 @@ Examples:
 > This has an {42 :integer}
 > ```
 >
-> A function from the `icu` namespace:
+> A function from the `ns` namespace:
 > ```
-> This has a {:icu:function}
+> This has a {:ns:function}
 > ```
 >
 > Options with and without a namespace:
 > ```
-> This has {:icu:function option=value ns:option=add_on}
+> This has {:ns:function option=value ns:option=value}
 > ```
 
 Support for _namespaces_ and their interpretation is implementation-defined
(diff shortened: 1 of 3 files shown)


spec/syntax.md
```diff
@@ -808,17 +808,20 @@ Examples:
...
+> A default function:
 > ```
-> This has a {:function}
+> This has an {42 :integer}
 > ```
-> An add-on function from the `icu` namespace:
+>
+> A function from the `icu` namespace:
 > ```
 > This has a {:icu:function}
```
aphillips, line 819: Probably better to use the examples namespace than `icu`, since `icu` doesn't have a `function` function.

```suggestion
> A function from the `ns` namespace:
> ```
> This has a {:ns:function}
```

spec/syntax.md
```diff
@@ -808,17 +808,20 @@ Examples:
...
 > ```
-> An option and an add-on option:
+>
+> Options with and without a namespace:
 > ```
-> This has {:options option=value icu:option=add_on}
+> This has {:icu:function option=value ns:option=add_on}
```
aphillips, line 824: Ibid.

```suggestion
> This has {:ns:function option=value ns:option=value}
```

PR #974, opened 2024-12-20
Split `spec/registry.md` into multiple files under `spec/functions/`

eemeli:
This PR splits the current `spec/registry.md` into four files, all normative:
- `spec/functions/README.md` -- the overall presentation of default functions, now including a TOC that enumerates the default functions and establishes their spec order
- `spec/functions/string.md` -- String Value Selection and Formatting (`:string`)
- `spec/functions/number.md` -- Numeric Value Selection and Formatting (`:number`, `:integer`, `:math`, `:currency`, and `:unit`)
- `spec/functions/datetime.md` -- Date and Time Value Formatting (`:datetime`, `:date`, and `:time`)

These files correspond to the top-level spec sections, and each contains a number of internal references to shared definitions, such as "Digit Size Options" for numeric functions. Because of those references, I don't think it makes sense to split each function into its own file.


spec/functions/README.md
```diff
@@ -0,0 +1,87 @@
...
+>   for the option `numberingSystem` in _functions_ such as `:number` or `:integer`.
+
+Future versions of this specification MAY define additional _options_ and _option_ values,
+subject to the rules in the [Stability Policy](#stability-policy),
+for _functions_ found in this specification.
+As implementations are permitted to ignore _options_ that they do not support,
+it is possible to write _messages_ using _options_ not defined below
+which currently format with no error, but which could produce errors
+when formatted with a later edition of this specification.
+Therefore, using _options_ not explicitly defined here is NOT RECOMMENDED.
```
aphillips, line 87: Probably we should have a section here describing the registration process and maturity level of the default function registry. We should also have a specific version number somewhere. These comments are not for this specific PR, but for _after_ we make the changes here.

spec/functions/datetime.md
```diff
@@ -0,0 +1,313 @@
+## Date and Time Value Formatting
+
+This subsection describes the functions and options for date/time formatting.
+Selection based on date and time values is not required in this release.
+
+> [!NOTE]
+> Selection based on date/time types is not required by MF2.
+> Implementations should use care when defining selectors based on date/time types.
```
aphillips, line 8: This uses a normative word ("should") as if it were not normative... in a non-normative note and dances around the word "required". I'd suggest instead:

```suggestion
> This version of MF2 does not define selection on date/time types.
> Use care if trying to provide implementation-specific support for this feature.
```

spec/intro.md
```diff
@@ -10,7 +10,7 @@
    1. [`message.abnf`](message.abnf)
 1. [Formatting](formatting.md)
 1. [Errors](errors.md)
-1. [Default Function Registry](registry.md)
+1. [Default Function Registry](functions/README.md)
```
aphillips, line 13: In some teleconferences in October/November, we discussed changing from using the term "registry" to using the term "function set". I made changes _outside_ the spec, but didn't change the spec to implement that. We should probably adopt the terminology we intend to use longer-term, e.g.

```suggestion
1. [Default Function Set](functions/README.md)
```

spec/intro.md
```diff
@@ -10,7 +10,7 @@
...
 1. [`u:` Namespace](u-namespace.md)
```
aphillips, line 14: Should the `u:` namespace be in the functions directory? Or in its own subdirectory?

spec/syntax.md
```diff
@@ -539,7 +539,7 @@ whether an _operand_ is required,
 what form the values of an _operand_ can take,
 what _options_ and _option_ values are acceptable,
 and what outputs might result.
-See [function registry](./registry.md) for more information.
+See [function registry](./functions/README.md) for more information.
```
aphillips, line 542:
```suggestion
See the [default function set](./functions/README.md) for more information.
```

aphillips (commented):
Thank you! Nice start!

See comments about whether we make the "registry"->"function set" change now (an open topic probably).

eemeli:
@aphillips I have a follow-up PR nearly ready that drops the "function registry" term, but that ends up being less clearly editorial because it modifies e.g. our stability policy. 

Hence not changing any words in this PR, so we can consider the slightly more substantive chances separately.

aphillips (approved):

PR #983, opened 2025-01-20
Drop references to a "function registry"

eemeli:
As discussed previously (most recently in https://github.com/unicode-org/message-format-wg/pull/974#discussion_r1894125386), we should update our terminology to no longer refer to a "function registry", a term that we don't actually define.

This PR applies that change, along with some incidental cleanup. The term _**default function**_ is introduced, and used where appropriate.

The GitHub label `registry` should probably also be renamed as `default-functions`.

### DIFF
--- a/spec/functions/README.md
+++ b/spec/functions/README.md
@@ -1,4 +1,4 @@
-# MessageFormat 2.0 Default Function Registry
+# MessageFormat 2.0 Default Functions
 
 ## Table of Contents
 
@@ -18,8 +18,9 @@
 
 ## Introduction
 
-This section defines the **REQUIRED** _functions_ which are REQUIRED for conformance with this specification,
-along with **RECOMMENDED** _functions_ that SHOULD be implemented to support
+This section defines the **_<dfn>default functions</dfn>_**
+which are REQUIRED for conformance with this specification,
+along with _default functions_ that SHOULD be implemented to support
 additional functionality.
 
 To **_<dfn>accept</dfn>_** a function means that an implementation MUST NOT
@@ -34,26 +35,25 @@ Implementations MAY emit an _Unsupported Operation_ error for _options_
 or _option_ values that they cannot support.
 
 _Functions_ can define _options_. 
-An _option_ can be **REQUIRED** or **RECOMMENDED**.
+An _option_ can be REQUIRED or RECOMMENDED.
 
-Implementations MUST _accept_ each **REQUIRED** _function_ and
-MUST _accept_ all _options_ defined as **REQUIRED** for those _functions_.
+Implementations MUST _accept_ each REQUIRED _default function_ and
+MUST _accept_ all _options_ defined as REQUIRED for those _functions_.
 
-Implementations SHOULD _accept_ each **RECOMMENDED** _function_. 
+Implementations SHOULD _accept_ each RECOMMENDED _default function_. 
 For each such _function_, the implementation MUST accept all _options_
-listed as **REQUIRED** for that _function_.
+listed as REQUIRED for that _function_.
 
-Implementations SHOULD _accept_ _options_ that are marked as **RECOMMENDED**.
+Implementations SHOULD _accept_ _options_ that are marked as RECOMMENDED.
 
 Implementations MAY _accept_ _functions_ not defined in this specification.
 In addition, implementations SHOULD provide mechanisms for users to
-register and use user-defined _functions_ and their associated _functional handlers_.
+register and use user-defined _functions_ and their associated _function handlers_.
 Functions not defined by any version of this specification SHOULD use 
 an implementation-defined or user-defined _namespace_.
 
 Implementations MAY implement additional _options_ not defined
-by any version of this specification
-for **REQUIRED** and **RECOMMENDED** functions.
+by any version of this specification for _default functions_.
 Such _options_ MUST use an implementation-specific _namespace_.
 
 Implementations MAY _accept_, for _options_ defined in this specification,
@@ -61,7 +61,7 @@ _option_ values which are not defined in this specification.
 However, such values might become defined with a different meaning in the future,
 including with a different, incompatible name
 or using an incompatible value space.
-Supporting implementation-specific _option_ values for **REQUIRED** or **RECOMMENDED** functions is NOT RECOMMENDED.
+Supporting implementation-specific _option_ values for _default functions_ is NOT RECOMMENDED.
 
 Implementations MAY _accept_, for _operands_ or _options_ defined in this specification,
 values with implementation-defined types.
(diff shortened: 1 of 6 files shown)


spec/functions/README.md
```diff
@@ -18,8 +18,9 @@
 
 ## Introduction
 
-This section defines the **REQUIRED** _functions_ which are REQUIRED for conformance with this specification,
-along with **RECOMMENDED** _functions_ that SHOULD be implemented to support
+This section defines the **_<dfn>default functions</dfn>_**
+which are REQUIRED for conformance with this specification,
+along with _default functions_ that SHOULD be implemented to support
```
aphillips, line 23: This dilutes the nice parallel between the name **REQUIRED**/**RECOMMENDED** and the 2119 associated keywords REQUIRED/RECOMMENDED for the various "default functions", although I diluted it for clarity by using SHOULD instead of RECOMMENDED when I created this section.

Note that we previously discussed name this the "Default Function Set" with "function set" being a term. I'm not sure that "function set" is useful as a term?

I am also not sure if "default" is the right term, if we include the recommended items in it.

Finally, I note that "This section..." might not be right, since we broke it up into a list of documents.

Perhaps:

```suggestion
This section defines the **_<dfn>default functions</dfn>_**.
This includes the **REQUIRED** _functions_,
which are required for conformance with this specification,
along with the **RECOMMENDED** _functions_
that SHOULD be implemented to support
```

spec/functions/README.md
```diff
@@ -34,34 +35,33 @@ Implementations MAY emit an _Unsupported Operation_ error for _options_
 or _option_ values that they cannot support.
 
 _Functions_ can define _options_. 
-An _option_ can be **REQUIRED** or **RECOMMENDED**.
+An _option_ can be REQUIRED or RECOMMENDED.
 
-Implementations MUST _accept_ each **REQUIRED** _function_ and
-MUST _accept_ all _options_ defined as **REQUIRED** for those _functions_.
+Implementations MUST _accept_ each REQUIRED _default function_ and
```
aphillips, line 40: I wouldn't make this change: I don't think it is necessary.

spec/functions/README.md
```diff
@@ -34,34 +35,33 @@ Implementations MAY emit an _Unsupported Operation_ error for _options_
...
+MUST _accept_ all _options_ defined as REQUIRED for those _functions_.
 
-Implementations SHOULD _accept_ each **RECOMMENDED** _function_. 
+Implementations SHOULD _accept_ each RECOMMENDED _default function_.
```
aphillips, line 43: Ibid.

spec/functions/README.md
```diff
@@ -34,34 +35,33 @@ Implementations MAY emit an _Unsupported Operation_ error for _options_
 or _option_ values that they cannot support.
 
 _Functions_ can define _options_. 
-An _option_ can be **REQUIRED** or **RECOMMENDED**.
+An _option_ can be REQUIRED or RECOMMENDED.
```
aphillips, line 38: The formatting here is intentional. Many such below. Won't repeat.

spec/functions/README.md
```diff
@@ -34,34 +35,33 @@ Implementations MAY emit an _Unsupported Operation_ error for _options_
...
-listed as **REQUIRED** for that _function_.
+listed as REQUIRED for that _function_.
 
-Implementations SHOULD _accept_ _options_ that are marked as **RECOMMENDED**.
+Implementations SHOULD _accept_ _options_ that are marked as RECOMMENDED.
 
 Implementations MAY _accept_ _functions_ not defined in this specification.
 In addition, implementations SHOULD provide mechanisms for users to
-register and use user-defined _functions_ and their associated _functional handlers_.
+register and use user-defined _functions_ and their associated _function handlers_.
```
aphillips, line 51: Good catch

spec/syntax.md
```diff
@@ -532,14 +532,11 @@ values during formatting.
...
-whether an _operand_ is required,
-what form the values of an _operand_ can take,
-what _options_ and _option_ values are acceptable,
-and what outputs might result.
-See [function registry](./functions/README.md) for more information.
+The resolution of a _function_ relies on an implementation-defined _function handler_.
+Some _functions_ can be used both as a _selector_ as well as in a _placeholder_;
+others are only valid in one of these positions.
+_Functions_ also differ in their requirements on the _operand_ and _options_ that they accept.
+See _Function Resolution_ and _default functions_ for more information.
```
aphillips, line 539: Maybe keep the link to the default function document instead of relying on the linkifier? That makes the document easier to use from GH.

aphillips (commented):
Some food for thought comments. Otherwise looks good.

aphillips:
About the label: I was waiting to rename it based on discussion of what the name should be 😉

spec/functions/README.md
eemeli: Now I'm confused: Are the bold **REQUIRED** and **RECOMMENDED** intended to be understood as RFC 2119 key words, or not? I'm also confused by your suggestion including "required for conformance" about the former and "SHOULD be implemented" about the latter.

I would find it easier to understand if we had only one set of _default functions_, which includes both those that are REQUIRED and those that are RECOMMENDED for implementation, without creating separate category labels for those subsets.

spec/syntax.md
eemeli:
```suggestion
See [Function Resolution](./spec/formatting.md#function-resolution)
and [Default Functions](./spec/functions/README.md) for more information.
```

spec/functions/README.md
aphillips: The idea is to clearly identify which functions and which options are required vs. recommended. The words are both 2119 keywords and "categories" (There is also the "proposed" state, for items not yet finalized that composes with REQUIRED/RECOMMENDED). We can discuss if the additional level of formality is needed in the call later today.

spec/functions/number.md
```diff
@@ -517,7 +520,13 @@ In general, the default values for such _options_ depend on the locale,
 the unit,
 the value of other _options_, or all of these.
 
-The following options and their values are required to be available on the function `:unit`:
+> [!NOTE]
+> The option `select` does not accept the value `ordinal` because selecting
+> unit values using ordinal rules makes no sense.
```
aphillips, line 525: See above comment on line 391

spec/functions/number.md
```diff
@@ -388,7 +386,11 @@ Some implementations might not be able to produce all of these formats for every
 > [options](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#options)
 > in JavaScript's `Intl.NumberFormat`.
 
-The following options and their values are required to be available on the function `:currency`:
+> [!NOTE]
+> The option `select` does not accept the value `ordinal` because selecting
+> currency values using ordinal rules makes no sense.
```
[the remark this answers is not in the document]
eemeli: I removed these notes, as the `select` option was removed from `:currency` and `:unit` functions in #991.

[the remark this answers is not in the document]
eemeli: Actually, needed to force-push a fix, as this and a few other problems were due to the preceding merge applying incorrect changes.

aphillips (approved):

aphillips:
Per discussion in the 2025-02-10 teleconference, I am merging these changes. Others might be needed to meet v47 release criteria.

PR #986, opened 2025-01-21
Deduplicate the section "Default Value of `select` Option"

eemeli:
This is a purely editorial change that should be fast-tracked: the exact same section was included under both `:number` and `:integer`, and it makes more sense under "Number Selection". It should also be referenced in all the places where the default value of `select` is set.

### DIFF
--- a/spec/functions/number.md
+++ b/spec/functions/number.md
@@ -97,29 +97,6 @@ with _options_ on the _expression_ taking priority over any option values of the
 > would be formatted with the resolved options
 > `{ notation: 'scientific', minimumFractionDigits: '1' }`.
 
-##### Default Value of `select` Option
-
-The value `plural` is the default for the option `select`
-because it is the most common use case for numeric selection.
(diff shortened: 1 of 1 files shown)


bearfriend:
I realize this is just an editorial move, but I'm not sure how clear this section is.

Is the developer still naive for not including `two`, `few`, and `many`? Why not? If not, why are they for excluding `one` if they're writing in English where it will never be used?

Something that has frustrated me with MF documentation is the dearth of acknowledgment of the actual translation process, or at least how one might reasonably/typically expect that to go. Who is responsible for adding what to source vs target languages? Clearly that's not set in stone but I've always thought some guidance would be helpful.

Anyway, I mention it only because this section starts to scratch the surface of doing that but cuts well short of being ultimately helpful, in my opinion.

I'm happy to expand on this a bit further elsewhere if we think it's worth it.

macchiati (approved):

PR #988, opened 2025-01-21
Remove `style=percent` from :number and :integer

### LINKED ISSUE #956: [FEEDBACK] Please move style percent to its own function
Recently, currency and unit formatting were moved to their own registry functions. However, there is still `style` that is either "decimal" or "percent".

This does not make sense. Please either remove "percent" (since a unit formatter should be able to handle it) or move it to its own function.

eemeli:
Fixes #956 

A separate `:percent` function is added, and the `style` option dropped from `:number` and `:integer`.

As proposed in #987, selection is not enabled for `:percent`.

I've also left out the `notation` and `compactDisplay` options, as I'm not aware of any actual use cases combining them with percent formatting.

### DIFF
--- a/spec/functions/number.md
+++ b/spec/functions/number.md
@@ -43,9 +43,6 @@ The following options and their values are required to be available on the funct
   - `exceptZero`
   - `negative`
   - `never`
-- `style`
-  - `decimal` (default)
-  - `percent` (see [Percent Style](#percent-style) below)
 - `useGrouping`
   - `auto` (default)
   - `always`
@@ -97,21 +94,6 @@ with _options_ on the _expression_ taking priority over any option values of the
 > would be formatted with the resolved options
 > `{ notation: 'scientific', minimumFractionDigits: '1' }`.
 
-##### Percent Style
-
-When implementing `style=percent`, the numeric value of the _operand_
-MUST be multiplied by 100 for the purposes of formatting.
-
-> For example,
->
-> ```
-> The total was {0.5 :number style=percent}.
-> ```
->
-> should format in a manner similar to:
->
-> > The total was 50%.
-
 #### Resolved Value
 
 The _resolved value_ of an _expression_ with a `:number` _function_
@@ -160,9 +142,6 @@ function `:integer`:
   - `exceptZero`
   - `negative`
   - `never`
-- `style`
-  - `decimal` (default)
-  - `percent` (see [Percent Style](#percent-style) below)
 - `useGrouping`
   - `auto` (default)
   - `always`
@@ -185,21 +164,6 @@ Option values with the following names are however discarded if included in the 
 - `minimumFractionDigits`
 - `maximumFractionDigits`
 - `minimumSignificantDigits`
-
-##### Percent Style
-
-When implementing `style=percent`, the numeric value of the _operand_
-MUST be multiplied by 100 for the purposes of formatting.
-
-> For example,
->
-> ```
-> The total was {0.5 :number style=percent}.
-> ```
->
-> should format in a manner similar to:
->
-> > The total was 50%.
 
 #### Resolved Value
 
@@ -817,7 +781,6 @@ representing its decimal value:
 - `minimumSignificantDigits`
 - `maximumSignificantDigits`
 - `notation`
-- `style`
 
 ```abnf
 integer = "0" / ["-"] ("1"-"9") *DIGIT


spec/functions/number.md
```diff
@@ -526,6 +490,117 @@ together with the resolved options' values.
...
+
+The function `:percent` is a formatter for percentage values,
+which are a specialized form of numeric formatting.
+
+#### Operands
+
+The function `:percent` requires a [Number Operand](#number-operands) as its _operand_.
+
+For the purposes of formatting,
+the numeric value of the _operand_ is multiplied by 100.
```
aphillips, line 503: Since we have a separate function, purpose built for it, should we offer a pre-computed mode?

spec/functions/number.md
```diff
@@ -526,6 +490,117 @@ together with the resolved options' values.
...
+>
+> would be formatted with the resolved options
+> `{ minimumFractionDigits: '1', roundingMode: 'floor' }`.
+
+#### Resolved Value
+
+The _resolved value_ of an _expression_ with a `:percent` _function_
+contains an implementation-defined numerical value
+of the _operand_ of the annotated _expression_ multiplied by 100,
+together with the resolved options' values.
```
aphillips, line 599: For clarity, and since "resolved value" and "options" are the terms, perhaps:

```suggestion
together with the _resolved values_ of the _options_.
```

spec/functions/number.md
eemeli: You mean an option toggling if the input is multiplied or not?

spec/functions/number.md
eemeli: Variants of this same snippet are used in nearly all the functions. This change would be better made separately, affecting all of them at once.

spec/functions/number.md
aphillips: Yes, exactly.

spec/functions/number.md
aphillips: Perhaps "_resolved values_ of the resolved list of _options_"? Oh the pedantry...

spec/functions/number.md
sffc: ICU4C calls it "scale" or "multiplier"

ICU4X calls it "multiply_pow10"

I don't think we've had a proper bikeshedding session on this before. I've sort-of named it what I felt like naming it and no one ever pushed back with better names.

spec/functions/number.md
```diff
@@ -526,6 +490,117 @@ together with the resolved options' values.
...
+If the _operand_ of the _expression_ is an implementation-defined type,
+such as the _resolved value_ of an _expression_ with a `:number` or `:integer` _annotation_,
+it can include option values.
+In general, these are included in the resolved option values of the _expression_,
+with _options_ on the _expression_ taking priority over any option values of the _operand_.
+Option values with the following names are however discarded if included in the _operand_:
+
+- `compactDisplay`
+- `notation`
+- `select`
```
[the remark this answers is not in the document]
eemeli: That would be different from what we do with other functions, such as `:integer`, and it would make extensibility difficult.

An implementation could support additional options that are set in the calling code on the value being formatted, and are understood by the formatter. For a somewhat silly example, an implementation could support colour as a property of numbers, and format a "green 0.42" so that the % sign was also green in the output. If we here said that all options not explicitly listed are discarded, this kind of extensibility would not be possible.

spec/functions/number.md
eemeli: Given that we'll be defaulting to a value 1 getting formatted as something like "100%", I think `scale` makes the most sense, with a default value of `1` and supporting `100` as an alternative. With this approach,
```
{42 :percent scale=100}
```
would then be formatted as "42%".

Or is the ICU4C meaning of this option the exact opposite?

sffc: `scale=100` means "multiply by 100" so your code would result in `4200%`.

sffc: See here: https://unicode-org.github.io/icu/userguide/format_parse/numbers/skeletons.html#scale

sffc: So if you decide to use the name `scale`, then I would suggest that it defaults to `100` and allows setting it to `1` to be consistent with how ICU4C handles it.

sffc: Another thought: an enumeration might be appropriate here. Something like:

1. `scalingMode: "scale"` scales the number: 0.5 => 50%
2. `scalingMode: "passThrough"` passes through the number: 50 => 50%

(sorry for the quad-post)

eemeli: In that case, `scale` is a rather sub-optimal choice, as it demonstrably can be understood as either "scale the number by this much" or "the scale extends to this value as the maximum".

sffc: More suggestions, with some help from an LLM

```
enum PercentageScale { Fraction, Percentage }
enum InputMode { Decimal, Scaled }
enum ValueType { Normalized, Preformatted }
enum PercentMode { Ratio, Percent }
enum ScaleType { UnitInterval, WholePercent }
enum InputScale { PartsPerHundred, FinalPercent }
enum Scale { Raw, Scaled }
```

eemeli: Maybe `style=raw`, as an alternative to a default `style=scaled`?

sffc: I would like to avoid any option with the unqualified `style` name.

eemeli:
During today's discussion, we noted that percent formatting is already available via `:unit unit=percent`.

This means that introducing `:percent` or retaining support for `:number style=percent` would be duplicating nearly the same functionality by two different means, and I don't think we should do that.

To still support the multiplicative behaviour (which `:unit` doesn't have), it might be best to add a `:math multiply` operator. @sffc raised some performance concerns about allowing multiplication by any number, and the digit-size option range (used by `:math add` and `:math subtract`) caps out at `99` for literals. Therefore, I'm considering refactoring this PR to:
- Drop `:number style` and `:integer style`.
- Add `:math multiply` with supported values `10`, `100`, `1000`.

aphillips:
> Add :math multiply with supported values 10, 100, 1000.

I'm dubious that this is a good idea. The `:math` function has a suggestive name. It's reasonable to assume that it will, over time, be extended to other purposes, similar to math functionality in programming languages (such as `java.lang.Math` in Java). It's _convenient_ to solve specific short term problems using this function now, but we should avoid binding our hands too much.

Note that this would mean that `unit=percent` would become unscaled (`{50 :unit unit=percent}` formats as `50%`). If we want to provide a `:math` modifier, perhaps `scale` or `exp`.

Also, there are other formats that we're ignoring (because they are much less common) which are supported by CLDR data, which is per-mille and per-myriad formatting. We should just solve the problem. The existence of these suggests a convenience function or maybe a convenience _unit_. 

Consider:

```
.local $val = {|50|}
.local $pct = {$val :unit unit=percent-scaled}
{{I print {$pct} as 5000% but {$val :unit unit=percent} as 50%}}
```

sffc:
I still have the weakly held position that :percent carries its own weight, and we should just add it and move on. 

It doesn't _completely_ duplicate functionality: you use :percent if you want scaling, and :unit unit=percent if not.

sffc:
Another thought. If we want to try and be more general, I'd be okay with a function such as `:dimensionless` which supports all dimensionless units _and_ automatically scales values.

This could also be an option on `:unit` where the only two valid options are "scale dimensionless units" and "do not scale dimensionless units" (not an arbitrary multiplier). Maybe return an error if the option is set on a non-dimensionless unit.

But I still prefer a clean `:percent`.

eemeli:
Updated as per discussion during today's meeting.

aphillips (approved):

aphillips:
This change was approved in the 2025-02-10 call.

PR #989, opened 2025-01-21
Simplify syntax character definitions, dropping content-char

eemeli:
With the simplifications we've made to the syntax, in particular allowing unpaired surrogates and dropping the unsupported constructs, the character definitions are no longer benefiting from building on a common `content-char` definition, and it's simpler to not refer to it.

This change is purely editorial, as the final meaning of the ABNF stays just the same as before.


aphillips:
approved in 2025-02-10 call

PR #990, opened 2025-01-27
Allow name-char as first character of unquoted-literal

### LINKED ISSUE #724: [FEEDBACK] Rationalize name-char
This was "[FEEDBACK] Message Format Unquoted Literals #724", but was changed to focus on name-char. See later comments.
----
OLD
## Summary

Consider relaxing constraints on literals, after v45

## Background

Right now, unquoted literals are fairly narrowly constrained by
[<u>message.abnf</u>](https://github.com/unicode-org/message-format-wg/blob/main/spec/message.abnf)
; here are the relevant lines:

```
unquoted = name / number-literal

; number-literal matches JSON number
(https://www.rfc-editor.org/rfc/rfc8259#section-6)

number-literal = \["-"\] (%x30 / (%x31-39 \*DIGIT)) \["." 1\*DIGIT\]
\[%i"e" \["-" / "+"\] 1\*DIGIT\]

; name matches https://www.w3.org/TR/REC-xml-names/#NT-NCName

name = name-start \*name-char

name-start = ALPHA / "\_"

/ %xF900-FDCF / %xFDF0-FFFC / %x10000-EFFFF

name-char = name-start / DIGIT / "-" / "."

```

### Reason for reconsidering

However, for functions outside of the standard registry, this forces
many natural literals to use quotes. Here is an example from a function
that would handle MF1’s choice format:

```
|[0,1)| {{{\$count} is zero or fraction}}
```
The natural literals to use would be intervals, which use \[,(,),\]
characters for ranges (the choice format would require some recasting
because it depends on ordering of variants. It currently uses \>.) So
that would require

```
|[0,1)| {{{\$count} is zero or fraction}}
```
Many Unicode symbols are *included* by XML’s **NT-NCName** (about 6,000
currently), while many are excluded (about 2,600 currently). But these
are **literals**, **not identifiers**, which is what **name** is
intended for. By expanding beyond identifier usage, it allows functions
to avoid requiring quoting in many cases. It also allows us to dispense
with the special formulation for number-literal.

The literals for number, date, etc could be specified elsewhere, but
wouldn’t have to be in the ABNF.

That would allow for various registries to have more sophisticated
literal without requiring quoting, and without privileging the
structured literals that we know about now.

### Requirements

So, what restrictions on characters for a broadened definition of
**unquoted** literals would be required by a revised ABNF?

1.  No ‘}’, because it would make .local \$x = {literal} fail.

2.  No ‘\|’, because an initial one would conflict with quoting
    1.  While it would be possible to just disallow initial ones, but for clarity best to always forbid.

3.  No ‘:' or '$', because an initial one would indicate a function or variable, which would conflict in expressions starting with one (and initial '$' would conflict in the value of an option).
    * While it would be possible to just disallow initial ones, but for clarity best to always forbid.

4.  No ‘{’. Not strictly required, but for clarity best to always forbid.

5.  None of the big blocks of ‘strange’ code points that XML forbids: controls, (unpaired) surrogates, private-use, noncharacters.
    * These are all immutable ([<u>Unicode Character Encoding Stability</u>](https://www.unicode.org/policies/stability_policy.html#Property_Value)).

    * This also disallows the noncharacters that XML didn’t know about yet, before the noncharacter property was made immutable.

6.  No whitespace, since **variant** uses that for separators between  keys, and expressions use it to separate various components.

    * This could be done by just disallowing the “**s**” production characters, but that could be very confusing. {a b} looks too  much like two items (the space is an A0 NO-BREAK SPACE). So it should be broadened to the Unicode Whitespace characters.

    * Unicode Whitespace is not guaranteed immutable, but has not changed for over a decade. Anyway, we would derive the code points as of now, so everything would be stable into the future.

## Detailed Proposal

This would result in the following change:

### OLD

```
unquoted-literal = name / number-literal
; number-literal matches JSON number (https://www.rfc-editor.org/rfc/rfc8259#section-6)
number-literal   = ["-"] (%x30 / (%x31-39 *DIGIT)) ["." 1*DIGIT] [%i"e" ["-" / "+"] 1*DIGIT]

// The characters include the following (though name-char and
number-literal additions are positional):

// name-start is \[\\: A-Z \_ a-z \x{C0}-\x{D6} \x{D8}-\x{F6}
\x{F8}-\x{2FF} \x{370}-\x{37D} \x{37F}-\x{1FFF} \x{200C}-\x{200D}
\x{2070}-\x{218F} \x{2C00}-\x{2FEF} \x{3001}-\x{D7FF} \x{F900}-\x{FDCF}
\x{FDF0}-\x{FFFD} \x{10000}-\x{EFFFF}\]

// name-char adds \[\\- . 0-9 \x{B7} \x{0300}-\x{036F}
\x{203F}-\x{2040}\]

// number-literal adds \[+ e\]
```

### NEW
This changes just the first line above, `unquoted-literal = name / number-literal` — the rest of the above would remain the same.

```
unquoted-literal = literal-char+

// Then down in ; Restrictions on characters in various contexts

literal-char = _all **but** the following list; simpler to leave in this format until after feedback._
```
Needed to avoid syntax conflicts
```
U+0024	DOLLAR SIGN
U+003A	COLON
U+007B LEFT CURLY BRACKET
U+007C VERTICAL LINE
U+007D RIGHT CURLY BRACKET
```
Whitespace
```
U+0020 SPACE
U+00A0 NO-BREAK SPACE
U+1680 OGHAM SPACE MARK
U+2000 - U+200A EN QUAD .. HAIR SPACE
U+2028 LINE SEPARATOR
U+2029 PARAGRAPH SEPARATOR
U+202F NARROW NO-BREAK SPACE
U+205F MEDIUM MATHEMATICAL SPACE
U+3000 IDEOGRAPHIC SPACE
```
Controls
```
U+0000 - U+001F
U+007F - U+009F
```
Surrogates
```
U+D800 - U+DFFF
```
Private Use
```
U+E000 - U+F8FF U+F0000 - U+FFFFD U+100000 - U+10FFFD
```
Noncharacters
```
U+FDD0 - U+FFFE U+FFFF U+1FFFE U+1FFFF U+2FFFE U+2FFFF U+3FFFE U+3FFFF
U+4FFFE U+4FFFF U+5FFFE U+5FFFF U+6FFFE U+6FFFF U+7FFFE U+7FFFF U+8FFFE
U+8FFFF U+9FFFE U+9FFFF U+AFFFE U+AFFFF U+BFFFE U+BFFFF U+CFFFE U+CFFFF
U+DFFFE U+DFFFF U+EFFFE U+EFFFF U+FFFFF U+FFFFE U+FFFFF U+10FFFE
U+10FFFF
```

### Notes
1. We could allow a non-initial ':' and '$'
2. We could reserve a few additional initial ASCII characters, eg. '!', '%', '*', '/', '?', '@'

eemeli:
Fixes #724 

As discussed today, our `name` definition is a slightly restricted variant of the XML `NCName`, which has not been updated to account for developments in the last 15 years or so that have e.g. included combining marks and the ALM in the `name-start` range.

So as the `name` production is already messy, `unquoted-literal` is not really made _much worse_ by allowing `name-char` as its first character. Doing so also allows us to drop the `number-literal` rule from the syntax.

The `number-literal` rule is still needed for number functions, hence its re-definition in the Number Operands section. Note that I've left out the exponenetial part from it, as the `+` character is no longer valid without quoting.

This change also removes the need to quote `2-digit`, a valid literal value for some `:datetime` options. Previously, it was the only default function option value that needed quoting.

Filing initially as a draft, as I want to implement this to make sure it works as expected, and to make sure that all test suite changes have been accounted for.

### DIFF
--- a/spec/functions/number.md
+++ b/spec/functions/number.md
@@ -655,8 +655,12 @@ Implementations MUST NOT substitute the unit without performing the associated c
 ### Number Operands
 
 The _operand_ of a number function is either an implementation-defined type or
-a literal whose contents match the `number-literal` production in the [ABNF](/spec/message.abnf).
+a literal whose contents match the following `number-literal` production.
 All other values produce a _Bad Operand_ error.
+
+```abnf
+number-literal = ["-"] (%x30 / (%x31-39 *DIGIT)) ["." 1*DIGIT]
+```
 
 > For example, in Java, any subclass of `java.lang.Number` plus the primitive
 > types (`byte`, `short`, `int`, `long`, `float`, `double`, etc.)
@@ -667,7 +671,7 @@ All other values produce a _Bad Operand_ error.
 > [!NOTE]
 > String values passed as variables in the _formatting context_'s
 > _input mapping_ can be formatted as numeric values as long as their
-> contents match the `number-literal` production in the [ABNF](/spec/message.abnf).
+> contents match the `number-literal` production.
 >
 > For example, if the value of the variable `num` were the string
 > `-1234.567`, it would behave identically to the local
--- a/spec/functions/number.md
+++ b/spec/functions/number.md
@@ -659,7 +659,7 @@ a literal whose contents match the following `number-literal` production.
 All other values produce a _Bad Operand_ error.
 
 ```abnf
-number-literal = ["-"] (%x30 / (%x31-39 *DIGIT)) ["." 1*DIGIT]
+number-literal = ["-"] (%x30 / (%x31-39 *DIGIT)) ["." 1*DIGIT] [%i"e" ["-" / "+"] 1*DIGIT]
 ```
 
 > For example, in Java, any subclass of `java.lang.Number` plus the primitive
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -41,9 +41,7 @@ variable       = "$" name
 
 literal          = quoted-literal / unquoted-literal
 quoted-literal   = "|" *(quoted-char / escaped-char) "|"
-unquoted-literal = name / number-literal
-; number-literal matches JSON number (https://www.rfc-editor.org/rfc/rfc8259#section-6)
-number-literal   = ["-"] (%x30 / (%x31-39 *DIGIT)) ["." 1*DIGIT] [%i"e" ["-" / "+"] 1*DIGIT]
+unquoted-literal = 1*name-char
 
 ; Keywords; Note that these are case-sensitive
 input = %s".input"
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -735,19 +735,17 @@ escaped as `\\` and `\|`.
 An **_<dfn>unquoted literal</dfn>_** is a _literal_ that does not require the `|`
 quotes around it to be distinct from the rest of the _message_ syntax.
 An _unquoted literal_ MAY be used when the content of the _literal_
-contains no whitespace and otherwise matches the `unquoted` production.
+contains no whitespace and otherwise matches the `unquoted-literal` production.
 Implementations MUST NOT distinguish between _quoted literals_ and _unquoted literals_
 that have the same sequence of code points.
 
-_Unquoted literals_ can contain a _name_ or consist of a _number-literal_.
-A _number-literal_ uses the same syntax as JSON and is intended for the encoding 
-of number values in _operands_ or _options_, or as _keys_ for _variants_.
+_Unquoted literals_ can contain any characters also valid in _name_,
+but with none of its additional restrictions on its first character.
 
 ```abnf
 literal          = quoted-literal / unquoted-literal
 quoted-literal   = "|" *(quoted-char / escaped-char) "|"
-unquoted-literal = name / number-literal
-number-literal   = ["-"] (%x30 / (%x31-39 *DIGIT)) ["." 1*DIGIT] [%i"e" ["-" / "+"] 1*DIGIT]
+unquoted-literal = 1*name-char
 ```
 
 ### Names and Identifiers
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -740,7 +740,7 @@ Implementations MUST NOT distinguish between _quoted literals_ and _unquoted lit
 that have the same sequence of code points.
 
 _Unquoted literals_ can contain any characters also valid in _name_,
-but with none of its additional restrictions on its first character.
+less _name_'s additional restrictions on the first character.
 
 ```abnf
 literal          = quoted-literal / unquoted-literal
--- a/test/tests/functions/datetime.json
+++ b/test/tests/functions/datetime.json
@@ -45,7 +45,7 @@
       "src": "{|2006-01-02T15:04:06| :datetime}"
     },
     {
-      "src": "{|2006-01-02T15:04:06| :datetime year=numeric month=|2-digit|}"
+      "src": "{|2006-01-02T15:04:06| :datetime year=numeric month=2-digit}"
     },
     {
       "src": "{|2006-01-02T15:04:06| :datetime dateStyle=long}"
--- a/spec/functions/datetime.md
+++ b/spec/functions/datetime.md
@@ -74,19 +74,7 @@ and what format to use for that field.
 > [!NOTE] > _Field options_ do not have default values because they are only to be used
 > to compose the formatter.
 
-The _field options_ are defined as follows:
-
-> [!IMPORTANT]
-> The value `2-digit` for some _field options_ MUST be quoted
-> in the MessageFormat syntax because it starts with a digit
-> but does not match the `number-literal` production in the ABNF.
->
-> ```
-> .local $correct = {$someDate :datetime year=|2-digit|}
-> .local $syntaxError = {$someDate :datetime year=2-digit}
-> ```
-
-The function `:datetime` has the following options:
+The function `:datetime` has the following _field options_:
 
 - `weekday`
   - `long`
--- a/test/tests/functions/integer.json
+++ b/test/tests/functions/integer.json
@@ -16,8 +16,8 @@
       "exp": "hello -4"
     },
     {
-      "src": "hello {0.42e+1 :integer}",
-      "exp": "hello 4"
+      "src": "hello {0.42 :integer}",
+      "exp": "hello 0"
     },
     {
       "src": ".input {$foo :integer} .match $foo 1 {{=1}} * {{other}}",
--- a/test/tests/functions/integer.json
+++ b/test/tests/functions/integer.json
@@ -20,6 +20,10 @@
       "exp": "hello 0"
     },
     {
+      "src": "hello {|0.42e+1| :integer}",
+      "exp": "hello 4"
+    },
+    {
       "src": ".input {$foo :integer} .match $foo 1 {{=1}} * {{other}}",
       "params": [{ "name": "foo", "value": 1.2 }],
       "exp": "=1"
--- a/test/tests/functions/number.json
+++ b/test/tests/functions/number.json
@@ -20,6 +20,10 @@
       "exp": "hello 0.42"
     },
     {
+      "src": "hello {|0.42e+1| :number}",
+      "exp": "hello 4.2"
+    },
+    {
       "src": "hello {foo :number}",
       "exp": "hello {|foo|}",
       "expErrors": [
--- a/test/tests/functions/number.json
+++ b/test/tests/functions/number.json
@@ -16,8 +16,8 @@
       "exp": "hello -4.2"
     },
     {
-      "src": "hello {0.42e+1 :number}",
-      "exp": "hello 4.2"
+      "src": "hello {0.42 :number}",
+      "exp": "hello 0.42"
     },
     {
       "src": "hello {foo :number}",
--- a/test/tests/syntax.json
+++ b/test/tests/syntax.json
@@ -421,72 +421,72 @@
       ]
     },
     {
-      "description": "... literal -> quoted-literal -> \"|\" \"|\" ...",
+      "description": "... quoted-literal",
       "src": "{||}",
       "exp": ""
     },
     {
-      "description": "... quoted-literal -> \"|\" quoted-char \"|\"",
+      "description": "... quoted-literal",
       "src": "{|a|}",
       "exp": "a"
     },
     {
-      "description": "... quoted-literal -> \"|\" escaped-char \"|\"",
+      "description": "... quoted-literal",
       "src": "{|\\\\|}",
       "exp": "\\"
     },
     {
-      "description": "... quoted-literal -> \"|\" quoted-char 1*escaped-char \"|\"",
+      "description": "... quoted-literal",
       "src": "{|a\\\\\\{\\|\\}|}",
       "exp": "a\\{|}"
     },
     {
-      "description": "... unquoted-literal -> number-literal -> %x30",
+      "description": "... unquoted-literal",
       "src": "{0}",
       "exp": "0"
     },
     {
-      "description": "... unquoted-literal -> number-literal -> \"-\" %x30",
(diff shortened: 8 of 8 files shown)


test/tests/functions/datetime.json
```diff
@@ -45,7 +45,7 @@
       "src": "{|2006-01-02T15:04:06| :datetime}"
     },
     {
-      "src": "{|2006-01-02T15:04:06| :datetime year=numeric month=|2-digit|}"
+      "src": "{|2006-01-02T15:04:06| :datetime year=numeric month=2-digit}"
```
mihnita, line 48: With this change the parsers now need a pretty big lookahead.

Before you it was enough to look at the first character:
`|` => quoted string
`0`-`9` or '-' > try to get a number literal
starting char => literal
anything else => Error

spec/syntax.md
```diff
@@ -735,19 +735,17 @@ escaped as `\\` and `\|`.
...
-A _number-literal_ uses the same syntax as JSON and is intended for the encoding 
-of number values in _operands_ or _options_, or as _keys_ for _variants_.
+_Unquoted literals_ can contain any characters also valid in _name_,
+but with none of its additional restrictions on its first character.
 
 ```abnf
 literal          = quoted-literal / unquoted-literal
 quoted-literal   = "|" *(quoted-char / escaped-char) "|"
-unquoted-literal = name / number-literal
-number-literal   = ["-"] (%x30 / (%x31-39 *DIGIT)) ["." 1*DIGIT] [%i"e" ["-" / "+"] 1*DIGIT]
```
mihnita, line 750: Yet another step towards making everything a string :-(

test/tests/functions/datetime.json
gibson042: How does that change? Anywhere that literal is valid, single-character lookahead still suffices.
* `|` → quoted-literal
* `$` → variable
* `:` → function
* `*` → key
* `#` or `/` → (markup)
* name-char → unquoted-literal
* anything else → error

gibson042 (commented):
> The `number-literal` rule is still needed for number functions, hence its re-definition in the Number Operands section. Note that I've left out the exponenetial part from it, as the `+` character is no longer valid without quoting.

This seems like a rather large drawback... don't we still want to allow something like `{|1e+12| :currency currency=JPY}`? Or for that matter, just allow `+` to be unquoted like e.g. `unquoted-literal = 1*(name-char / "+" / …)`.

> This change also removes the need to quote `2-digit`, a valid literal value for some `:datetime` options. Previously, it was the only default function option value that needed quoting.

That's nice.

test/tests/functions/datetime.json
macchiati: I agree with @gibson042. 

It is trivial to check the first character. A number literal can only start with one of 11 characters: - or 0..9. All of these are contained in `name-char`:

```
```

So the first character is enough to send you down the path.

Of course, once you start down the path to any of these outcomes, you could end up with something bogus

aphillips:
> > The number-literal rule is still needed for number functions, hence its re-definition in the Number Operands section. Note that I've left out the exponenetial part from it, as the + character is no longer valid without quoting.
>
> This seems like a rather large drawback... don't we still want to allow something like {|1e+12| :currency currency=JPY}? Or for that matter, just allow + to be unquoted like e.g. unquoted-literal = 1*(name-char / "+" / …).

We've... stalled on expanding number literals to non-integer values. I think this is something that will be wanted, but attempts to address this so far have not succeeded. I do hope that they will in the future.

Note that the `number-literal` rule doesn't have to appear in the ABNF if the `name` rule or one of the literal rules already covers it. It is _convenient_ that it is used in both, but it would be equally valid to just have the `number.md`  define number literals in the section in digit size options. It might even be better to do that, so that other functions could follow that model for subsetting names (cf. @macchiati's discussion of valid vs well-formed).

So, yeah, we could consider allowing the ASCII character `+` in an unquoted.

Is the meta-goal to allow any literal that does not contain whitespace, bidi formatting characters, or syntax-meaningful "sigils" to be unquoted? If that's the case, maybe we restore a small production for `reserved` starter characters?

eemeli:
> This seems like a rather large drawback... don't we still want to allow something like `{|1e+12| :currency currency=JPY}`? Or for that matter, just allow `+` to be unquoted like e.g. `unquoted-literal = 1*(name-char / "+" / …)`.

Is there really that much utility in supporting literal numerical values with exponents? I would claim that this is exceedingly rare, and could be left out initially. If someone does present a real-world use case, we could extend support later for `+` in `unquoted-literal` and for exponents in `number-literal`.

aphillips:
@eemeli mentions:
> I would claim that this is exceedingly rare, and could be left out initially. If someone does present a real-world use case, we could extend support later for + in unquoted-literal and for exponents in number-literal.

I agree that it's rare and note that we have already left it out. However, it would be good if we had enough foresight to minimize changes of this sort to the syntax in the future. Allowing `+` into unquoted does not mean that we "support exponents". It means that labels such as `goofy+lingonberry` or `1e+03` are valid unquoted strings.

@mihnita noted:
> It needs more than one if I still want to keep the number-literal:
> 12345678.9 => number literal
> 12345678Z9 => string literal

I think this is a good callout, from the point of view that the current syntax suggests we accept `name` or `number-literal` tokens, when the intention is that `number-literal` expands the set of starter characters. Elsewhere, in @macchiati's proposal, we do away with this (note that `12345678Z9` is _not_ a valid `name`, even though we all seem to be presuming that it is a valid `unquoted`!!)

I think that moving to a strictly character production would allow single-lookahead to capture `unquoted` (good efficiency) without incorrectly suggesting that the processor capture a number or a name. Interpretation of a given option or key **value** as a number by a functional handler is still permitted.

gibson042:
> Is there really that much utility in supporting literal numerical values with exponents? I would claim that this is exceedingly rare, and could be left out initially. If someone does present a real-world use case, we could extend support later for `+` in `unquoted-literal` and for exponents in `number-literal`.

Exponential notation seems important to me for big numbers and small numbers, such as are encountered in finance and physics. It's also present in most if not all of the languages in which one can already NumberFormat (cf. [MDN ```numberFormat.format(`${bigNum}E-6`)```](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/format#using_format_with_a_string)), and required in `:number` _output_ for notation "scientific" and "engineering".

Given all that, I would ask what is the benefit of **not** supporting exponential notation in Number Operands? If it is just "_it should be possible to express any number operand as an unquoted literal_", then I think the same arguments apply to "+" for exponents as applied to "-" for "2-digit". But for the record, I don't consider quoting to be particularly special or onerous, and would be fine if it is required for `|1.416784e+32|` but not for `1.416784e32` or `1.616255e-35`.

> I agree that it's rare and note that we have already left it out.

@aphillips what do you mean by "have already left it out"? Exponential notation is certainly a part of the current [`number-literal`](https://github.com/unicode-org/message-format-wg/blob/12ffb9b551925388787cf61ccd1b77f9f8301778/spec/message.abnf#L45-L46), which is documented to match [JSON number](https://www.rfc-editor.org/rfc/rfc8259#section-6).

> note that `12345678Z9` is _not_ a valid `name`, even though we all seem to be presuming that it is a valid `unquoted`!!

Well, yeah. Numeric literals were never valid names, and expanding `unquoted-literal` to remove the first-character constraints inherently includes such garden-path cases.

eemeli:
> It needs more than one if I still want to keep the number-literal:
> 12345678.9 => number literal
> 12345678Z9 => string literal

As this PR makes it so that a `number-literal` is only valid as an operand of a numerical function, I don't see how it'd be possible to write a parser that didn't initially parse a literal operand as an `unquoted-literal` or `quoted-literal`, given how the function is _after_ the operand.

Pre-emptively parsing the operand of `{1.0 :string}` as a number would definitely be a mistake.

aphillips:
@gibson042 asked:
> what do you mean by "have already left it out"? 

I mean that, while `number-literal` includes it, we don't normatively require implementations to parse anything except integers (for _digit size option_ in `number.md`). Implementations can, of course, interpret a number literal as a number (including exponentiation), but that's their business.

spec/functions/number.md
```diff
@@ -655,9 +655,13 @@ Implementations MUST NOT substitute the unit without performing the associated c
 ### Number Operands
 
 The _operand_ of a number function is either an implementation-defined type or
-a literal whose contents match the `number-literal` production in the [ABNF](/spec/message.abnf).
+a literal whose contents match the following `number-literal` production.
 All other values produce a _Bad Operand_ error.
 
+```abnf
+number-literal = ["-"] (%x30 / (%x31-39 *DIGIT)) ["." 1*DIGIT]
```
aphillips, line 662: This is different from the `number-literal` in the current ABNF and inconsistent with @gibson042's comment (and my reply to it) in that it leaves out the exponent part. If we leave out exponents in this ABNF, we should probably permit implementations to support them. Given that the original ABNF came from JSON, is there any reason to deviate from it?

spec/message.abnf
```diff
@@ -41,9 +41,7 @@ variable       = "$" name
 
 literal          = quoted-literal / unquoted-literal
 quoted-literal   = "|" *(quoted-char / escaped-char) "|"
-unquoted-literal = name / number-literal
-; number-literal matches JSON number (https://www.rfc-editor.org/rfc/rfc8259#section-6)
-number-literal   = ["-"] (%x30 / (%x31-39 *DIGIT)) ["." 1*DIGIT] [%i"e" ["-" / "+"] 1*DIGIT]
+unquoted-literal = 1*name-char
```
aphillips, line 44: This change loses the optional `bidi` productions around unquoted (because they are in `name`). This could be fixed and is probably cleaner if we add the `o` (optional whitespace) production to the end of `attribute` and `option` (all of the other functions already have optional/required whitespace after a literal).

macchiati: Agreed.

spec/functions/number.md
macchiati: I agree, we should keep the optional exponent here.

spec/syntax.md
macchiati: I think that comment is a confusion about thinking that quoted literal mean string, when they are completely separate.

The main syntax shouldn't talk about what format literal numbers are, or what format literal dates are, or what format literal units are; that's up to the functions.

spec/functions/number.md
eemeli: Done.

spec/message.abnf
eemeli: That should not be necessary, as whitespace is always allowed after `option` and `attribute`.

aphillips:
@eemeli Ready for review?

spec/syntax.md
```diff
@@ -735,19 +735,17 @@ escaped as `\\` and `\|`.
...
-contains no whitespace and otherwise matches the `unquoted` production.
+contains no whitespace and otherwise matches the `unquoted-literal` production.
 Implementations MUST NOT distinguish between _quoted literals_ and _unquoted literals_
 that have the same sequence of code points.
 
-_Unquoted literals_ can contain a _name_ or consist of a _number-literal_.
-A _number-literal_ uses the same syntax as JSON and is intended for the encoding 
-of number values in _operands_ or _options_, or as _keys_ for _variants_.
+_Unquoted literals_ can contain any characters also valid in _name_,
+but with none of its additional restrictions on its first character.
```
aphillips, line 743: Very slightly unclear? Perhaps:

```suggestion
less _name_'s additional restrictions on the first character.
```

aphillips (approved):
Very slight nit, otherwise LGTM

macchiati (approved):

macchiati:
It'd be nice to get this merged so I can have the other literal PR based off of head.

aphillips:
Based on previous discussion in the 2025-02-10 call, merging this change.

PR #991, opened 2025-01-27
Drop selection from :currency and :unit

eemeli:
As discussed today, we should leave out selection from `:currency` and `:unit` for now, as it's not clear that data is available to select correct plural values for them. For example, consider how the English phrase "$3 is your total price" refers to the currency value in the singular.

If/once selection on these values is better defined, we should consider adding support for that in MF2.

Where necessary, plural selection on the numerical values of currencies and unitized values is still possible via `:number` and `:integer`, which do support selection.

### DIFF
--- a/spec/functions/number.md
+++ b/spec/functions/number.md
@@ -287,8 +287,8 @@ The _function_ `:math` performs selection as described in [Number Selection](#nu
 
 ### The `:currency` function
 
-The function `:currency` is a selector and formatter for currency values,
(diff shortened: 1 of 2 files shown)


sffc (approved):
If something doesn't work even in English, then it most likely is even worse in other languages. This is also a compelling reason to keep these functions separate from :number. Thank you.

mihnita (approved):

PR #994, opened 2025-01-30
Fix #993, make expErrors array only (not boolean)

### LINKED ISSUE #993: Update the test schema to make expErrors "array" only
Right now that is an array or a boolean:

```json
"expErrors": {
  "description": "The runtime errors expected to be emitted when formatting the message. If expErrors is either absent or empty, the message must be formatted without errors.",
  "type": [
    "array",
    "boolean"
  ],
  ...
```

Why:

1. Something that can be one of two different types makes it harder to consume the json from statically typed languages
2. It is not useful in any way.
  The description already says _"If expErrors is either absent or empty, the message must be formatted without errors"_. That covers the possible `expErrors = false`. \
  So we already have 2 ways to say "nothing fails", don't need 3 ways. \
  And if we expect an error, then we should say what that error is, not just say "yes, we expect errors" (`expErrors = true`).

The only use of `expErrors` as boolean right now is in `test/tests/functions/currency.json`, and all instances are `"expErrors": false`. 
Which is not needed (as per description missing `expErrors` means no errors).
And that is what we do in all other tests, we don't specify `expErrors` when none are expected.

mihnita:
I hope this is not controversial.

### DIFF
--- a/test/schemas/v0/tests.schema.json
+++ b/test/schemas/v0/tests.schema.json
@@ -353,10 +353,7 @@
     },
     "expErrors": {
       "description": "The runtime errors expected to be emitted when formatting the message. If expErrors is either absent or empty, the message must be formatted without errors.",
-      "type": [
-        "array",
-        "boolean"
-      ],
+      "type": "array",
       "items": {
         "type": "object",
         "additionalProperties": false,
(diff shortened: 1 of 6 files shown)


mihnita:
Drop for now, it looks like some tests files use `expError = true`

PR #1003, opened 2025-02-10
Drop message.dtd data model definition

eemeli:
As discussed during today's call, we have not identified an actual use case for the `message.dtd` variant of the data model definition. In its absence, let's drop it for now, and only publish the TypeScript & JSON Schema definitions.

The data model section is also more clearly titled "Interchange Data Model", as it actually already was in the ToC.


aphillips (approved):
Seems to accomplish the request. LGTM.

aphillips:
@macchiati I need one more approval to merge this?

macchiati (approved):

PR #1007, opened 2025-02-15
Preparing the specification for LDML47 release

### LINKED ISSUE #1001: Fix spec details for pourover to v47 (and future)
I discussed with @srl295 the changes needed to make our spec flow more easily into TR35 when a release takes place. This issue is to track the needed _editorial_ changes to ensure our spec is crisp and clean for that process. A partial list is:

- Ensure all heads are at least H2 (e.g. \#\# not \#)
- Ensure all heads link uniquely when the separate .md files are merged into a single document
- Ensure heads don't have "2.0" in them where possible
- Ensure inter-file links transform properly
- Embed the ABNF into the spec properly
- Review tr-archive scripts and provide automation as possible

aphillips:
This PR includes the changes necessary to prepare the specification for LDML47, which is the first Stable release of MF2.

Normative changes:
- Activate the stability policy

Non-normative changes:
- Fix header indentation to support ToC generation
- Disambiguate headers where necessary

Fixes #1001

### DIFF
--- a/spec/data-model/README.md
+++ b/spec/data-model/README.md
@@ -1,4 +1,4 @@
-# Interchange Data Model
+## Interchange Data Model
 
 This section defines a data model representation of MessageFormat 2 _messages_.
 
@@ -47,7 +47,7 @@ declarations, options, and attributes to be optional rather than required proper
 > sometimes uses the production `identifier`.
 > This happens when the named item, such as a _function_, supports namespacing.
 
-## Messages
+### Messages Model
 
 A `SelectMessage` corresponds to a syntax message that includes _selectors_.
 A message without _selectors_ and with a single _pattern_ is represented by a `PatternMessage`.
@@ -114,7 +114,7 @@ interface CatchallKey {
 }
 ```
 
-## Patterns
+### Pattern Model
 
 Each `Pattern` contains a linear sequence of text and placeholders corresponding to potential output of a message.
 
@@ -160,7 +160,7 @@ interface FunctionExpression {
 }
 ```
 
-## Expressions
+### Expression Model
 
 The `Literal` and `VariableRef` correspond to the the _literal_ and _variable_ syntax rules.
 When they are used as the `body` of an `Expression`,
@@ -200,7 +200,7 @@ interface FunctionRef {
 type Options = Map<string, Literal | VariableRef>;
 ```
 
-## Markup
+### Markup Model
 
 A `Markup` object has a `kind` of either `"open"`, `"standalone"`, or `"close"`,
 each corresponding to _open_, _standalone_, and _close_ _markup_.
@@ -218,7 +218,7 @@ interface Markup {
 }
 ```
 
-## Attributes
+### Attribute Model
 
 `Attributes` is a key-value mapping
 used to represent the _expression_ and _markup_ _attributes_.
@@ -229,7 +229,7 @@ _Attributes_ with no value are represented by `true` here.
 type Attributes = Map<string, Literal | true>;
 ```
 
-## Extensions
+### Model Extensions
 
 Implementations MAY extend this data model with additional interfaces,
 as well as adding new fields to existing interfaces.
--- a/spec/data-model/README.md
+++ b/spec/data-model/README.md
@@ -47,7 +47,7 @@ declarations, options, and attributes to be optional rather than required proper
 > sometimes uses the production `identifier`.
 > This happens when the named item, such as a _function_, supports namespacing.
 
-### Messages Model
+### Message Model
 
 A `SelectMessage` corresponds to a syntax message that includes _selectors_.
 A message without _selectors_ and with a single _pattern_ is represented by a `PatternMessage`.
--- a/spec/functions/README.md
+++ b/spec/functions/README.md
@@ -1,4 +1,4 @@
-## MessageFormat 2.0 Default Functions
+## Default Functions
 
 ### Table of Contents
 
--- a/spec/functions/README.md
+++ b/spec/functions/README.md
@@ -20,6 +20,17 @@ which are REQUIRED for conformance with this specification,
 along with _default functions_ that SHOULD be implemented to support
 additional functionality.
 
+> [!IMPORTANT]
+> > _Functions_ that are not marked **Draft** are **Stable** and subject to
+> the provisions of the [stability policy](../intro.md#stability-policy).
+>
+> _Functions_ and _options_ marked as **Draft** are not stable.
+> Their name, _operands_, and _options_/_values_, and other requirements
+> might change before being declared **Stable** in a future release.
+>
+> Some **Draft** _functions_ or _options_ are also marked as **Proposed**.
+> These are additions to the specification that have not yet reached Technical Preview.
+
 To **_<dfn>accept</dfn>_** a function means that an implementation MUST NOT
 emit an _Unknown Function_ error for that _function_'s _identifier_.
 To _accept_ an _option_ means that a _function handler_ MUST NOT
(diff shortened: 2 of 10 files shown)


spec/functions/README.md
```diff
@@ -16,13 +15,22 @@
...
 
+> [!IMPORTANT]
+> _Functions_ that are not marked **Draft** are **Stable** and subject to
+> the provisions of the [stability policy](../intro.md#stability-policy).
+>
+> _Functions_ and _options_ marked as **Draft** are not stable.
+> Their name, _operands_, and _options_/_values_, and other requirements
+> might change or be removed before being declared **Stable** in a future release.
+>
+> Some **Draft** _functions_ or _options_ are also marked as **Proposed**.
```
macchiati, line 31: you use 'or' here, but 'and' 5 lines above. Do you mean to contrast them?

macchiati: The difference between 'Draft' and 'Proposed' is unclear.

spec/intro.md
```diff
@@ -78,10 +78,6 @@ A reference to a _term_ looks like this.
 
 ### Stability Policy
 
-> [!IMPORTANT]
-> The provisions of the stability policy are not in effect until
-> the conclusion of the technical preview and adoption of this specification.
-
 Updates to this specification will not make any valid _message_ invalid.
 
 Updates to this specification will not remove any syntax provided in this version.
```
macchiati, line 83: "Remove" any syntax is ambiguous. That would appear to prevent us from changing %XX / %YY / %ZZ to %XX-%ZZ.

It should be something like (but with better wording):
```suggestion
Updates to this specification will not make any message not match, where that message had matched a previous version's syntax.
```

spec/data-model/README.md
```diff
@@ -47,7 +47,7 @@ declarations, options, and attributes to be optional rather than required proper
 > sometimes uses the production `identifier`.
 > This happens when the named item, such as a _function_, supports namespacing.
 
-## Messages
+### Messages Model
```
eemeli, line 50:
```suggestion
### Message Model
```

spec/functions/README.md
```diff
@@ -1,8 +1,7 @@
-# MessageFormat 2.0 Default Functions
+## MessageFormat 2.0 Default Functions
```
eemeli, line 1:
```suggestion
## Default Functions
```

spec/u-namespace.md
```diff
@@ -1,4 +1,4 @@
-# MessageFormat 2.0 Unicode Namespace
+## MessageFormat 2.0 Unicode Namespace
```
eemeli, line 1:
```suggestion
## Unicode Namespace
```

eemeli (changes requested):

spec/functions/README.md
eemeli: Do we really need two different statuses for non-Stable things at this point? Especially as either Draft or Proposed status enables anything about them to change.

aphillips: @eemeli Actually, I think we might... not so much in this release as in the future.

The WG has formally approved a number of these, including their text, options, values, and such. These are effectively in (extended) Technical Preview and presumably lowercase-s stable. 

Items that the WG has not accepted yet need to be differentiated. The one example in this release is the option `timeZone`.

That said, I could live with everything being Draft and can make that change.

aphillips: @macchiati I meant to use `or` since there are options inside stable functions that are draft.

spec/intro.md
macchiati: ok.

> Updates to this specification will not change the syntactical meaning
of any syntax defined in this specification.	of any syntax defined in this specification.

We need room to be able to extend and/or clarify the meaning. And how do we even test whether 'syntactical meaning' have changed.  I think it is redundant. If it remains, it needs to be restated in terms of compatibility. Maybe like: 
> Updates to this specification will not change the syntactical meaning
of any syntax defined in this specification in a way that makes messages valid according to previous versions invalid.

aphillips: I propose we kill this item.

macchiati: Agreed!

macchiati (approved):

macchiati:
Another copy editing issue:

1. Sometimes the wording is "MessageFormat 2.0" and sometimes "MessageFormat 2". It should be consistent.
2. I think it is vital in the intro to use the term, and contrast it with "MessageFormat 1.0". We might even expand on the following (either in place or in an appendix) to provide context and contrast with MF 1.0. 
> The document is the successor to ICU MessageFormat,
> henceforth called ICU MessageFormat 1.0.
3. However, I think that in the body of the document simply using "MessageFormat" is cleaner (especially since going forward, it would be confusing to have LDML 49 MessageFormat 2.0 (or MessageFormat 2.2).

spec/functions/README.md
eemeli: I'd prefer to go with a single Draft definition for now, and then revisit when we know better what the '48 spec will look like.

spec/intro.md
```diff
@@ -78,10 +78,6 @@ A reference to a _term_ looks like this.
 
 ### Stability Policy
```
eemeli, line 79: There's a NOTE in this section that includes a SHOULD; that ought to be fixed as well.

eemeli: I filed #1010 for that.

spec/functions/README.md
aphillips: Fair enough. I'll make that change now.

spec/functions/datetime.md
```diff
@@ -280,11 +283,12 @@ the functions `:datetime`, `:date`, and `:time`.
...
-inclusion in the next release of this specification but have not yet been
-finalized.
-If accepted, implementations could be REQUIRED to make this _option_
-available in the functions `:datetime`, `:date`, and `:time`.
+> [!IMPORTANT]
+> The following _option_ and its values are **Draft** and proposed for
+> inclusion in a future release of this specification but have not yet been
+> finalized.
+> If accepted, implementations could be REQUIRED to make this _option_
+> available in the functions `:datetime`, `:date`, and `:time`.
```
eemeli, line 291: Everything here is draft, so we no longer need to note this as anything special.

```suggestion
The following _option_ and its values are REQUIRED to be available on
the functions `:datetime`, `:date`, and `:time`.
```

spec/functions/README.md
```diff
@@ -16,13 +15,19 @@
...
 This section defines the **_<dfn>default functions</dfn>_**
 which are REQUIRED for conformance with this specification,
 along with _default functions_ that SHOULD be implemented to support
 additional functionality.
 
+> [!IMPORTANT]
+> _Functions_ that are not marked **Draft** are **Stable** and subject to
+> the provisions of the [stability policy](../intro.md#stability-policy).
+>
+> _Functions_ and _options_ marked as **Draft** are not stable.
```
eemeli, line 27: After the other change I suggest here, we only have Draft functions.

```suggestion
> _Functions_ marked as **Draft** are not stable.
```

eemeli (changes requested):

spec/functions/README.md
aphillips: That's true **now** but might not be in the future. We shouldn't have to modify the introduction if/when we introduce draft options in the future.

It should be "or" though:

```suggestion
> _Functions_ or _options_ marked as **Draft** are not stable.
```

eemeli: Eh, sure. I still think it'd be easier to just redefine this yet again later, but it's not important.

eemeli (approved):

aphillips:
Merging these changes. Additional fast-tracks may be used for v47 cleanup.

PR #1008, opened 2025-02-16
Rationalize name-char

macchiati:
https://github.com/unicode-org/message-format-wg/issues/724 Rationalize name-char

### DIFF
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -53,38 +53,39 @@ identifier = [namespace ":"] name
 namespace  = name
 name       = [bidi] name-start *name-char [bidi]
 name-start = ALPHA
-                  / %x2B          ; 【+】     omit Cc %x0-1F, Whitespace %20, Ascii 【!"#$%&'()*】
-                  / %x5F          ; 【_】     omit Ascii 【,-./0123456789:;<=>?@】 【[\]^】
-                  / %xA1-61B      ;          omit Cc %x7F-9F, Whitespace %xA0, Ascii 【`】 【{|}~】
-                  / %x61D-167F    ;          omit BidiControl %x61C
-                  / %x1681-1FFF   ;          omit Whitespace %x1680
-                  / %x200B-200D   ;          omit Whitespace %x2000-200A
-                  / %x2010-2027   ;          omit BidiControl %x200E-200F
-                  / %x2030-205E   ;          omit Whitespace %x2028-2029 %x202F, BidiControl %x202A-202E
-                  / %x2060-2065   ;          omit Whitespace %x205F
-                  / %x206A-2FFF   ;          omit BidiControl %x2066-2069
-                  / %x3001-D7FF   ;          omit Whitespace %x3000
-                  / %xF900-FDCF   ;          omit Cs %xD800-DFFF, Co %xE000-F8FF
-                  / %xFDF0-FFFD   ;          omit NChar %xFDD0-FDEF
-                  / %x10000-1FFFD ;          omit NChar %xFFFE-FFFF
-                  / %x20000-2FFFD ;          omit NChar %x1FFFE-1FFFF
-                  / %x30000-3FFFD ;          omit NChar %x2FFFE-2FFFF
-                  / %x40000-4FFFD ;          omit NChar %x3FFFE-3FFFF
-                  / %x50000-5FFFD ;          omit NChar %x4FFFE-4FFFF
-                  / %x60000-6FFFD ;          omit NChar %x5FFFE-5FFFF
-                  / %x70000-7FFFD ;          omit NChar %x6FFFE-6FFFF
-                  / %x80000-8FFFD ;          omit NChar %x7FFFE-7FFFF
-                  / %x90000-9FFFD ;          omit NChar %x8FFFE-8FFFF
-                  / %xA0000-AFFFD ;          omit NChar %x9FFFE-9FFFF
-                  / %xB0000-BFFFD ;          omit NChar %xAFFFE-AFFFF
-                  / %xC0000-CFFFD ;          omit NChar %xBFFFE-BFFFF
-                  / %xD0000-DFFFD ;          omit NChar %xCFFFE-CFFFF
-                  / %xE0000-EFFFD ;          omit NChar %xDFFFE-DFFFF,
-                                  ;          omit NChar %xEFFFE-EFFFF %xFFFFE-FFFFF %x10FFFE-10FFFF,
-                                  ;          omit Co %xF0000-FFFFD %x100000-10FFFD
+                  / %x2B            ; «+»      omit Cc: %x0-1F, Whitespace: « », Ascii: «!"#$%&'()*»
+                  / %x5F            ; «_»      omit Ascii: «,-./0123456789:;<=>?@» «[\]^»
+                  / %xA1-61B        ;          omit Cc: %x7F-9F, Whitespace: %xA0, Ascii: «`» «{|}~»
+                  / %x61D-167F      ;          omit BidiControl: %x61C
+                  / %x1681-1FFF     ;          omit Whitespace: %x1680
+                  / %x200B-200D     ;          omit Whitespace: %x2000-200A
+                  / %x2010-2027     ;          omit BidiControl: %x200E-200F
+                  / %x2030-205E     ;          omit Whitespace: %x2028-2029 %x202F, BidiControl: %x202A-202E
+                  / %x2060-2065     ;          omit Whitespace: %x205F
+                  / %x206A-2FFF     ;          omit BidiControl: %x2066-2069
+                  / %x3001-D7FF     ;          omit Whitespace: %x3000
+                  / %xE000-FDCF     ;          omit Cs: %xD800-DFFF
+                  / %xFDF0-FFFD     ;          omit NChar: %xFDD0-FDEF
+                  / %x10000-1FFFD   ;          omit NChar: %xFFFE-FFFF
+                  / %x20000-2FFFD   ;          omit NChar: %x1FFFE-1FFFF
+                  / %x30000-3FFFD   ;          omit NChar: %x2FFFE-2FFFF
+                  / %x40000-4FFFD   ;          omit NChar: %x3FFFE-3FFFF
+                  / %x50000-5FFFD   ;          omit NChar: %x4FFFE-4FFFF
+                  / %x60000-6FFFD   ;          omit NChar: %x5FFFE-5FFFF
+                  / %x70000-7FFFD   ;          omit NChar: %x6FFFE-6FFFF
+                  / %x80000-8FFFD   ;          omit NChar: %x7FFFE-7FFFF
+                  / %x90000-9FFFD   ;          omit NChar: %x8FFFE-8FFFF
+                  / %xA0000-AFFFD   ;          omit NChar: %x9FFFE-9FFFF
+                  / %xB0000-BFFFD   ;          omit NChar: %xAFFFE-AFFFF
+                  / %xC0000-CFFFD   ;          omit NChar: %xBFFFE-BFFFF
+                  / %xD0000-DFFFD   ;          omit NChar: %xCFFFE-CFFFF
+                  / %xE0000-EFFFD   ;          omit NChar: %xDFFFE-DFFFF
+                  / %xF0000-FFFFD   ;          omit NChar: %xEFFFE-EFFFF
+                  / %x100000-10FFFD ;          omit NChar: %xFFFFE-FFFFF
+                                    ;          omit NChar: %x10FFFE-10FFFF
 
 name-char  = name-start / DIGIT
-                  / %x2D-2E       ; 【-.】    omit Cc %x0-1F, Whitespace 【 】, Ascii 【!"#$%&'()*+,】
+                  / %x2D-2E         ; «-.»     omit Cc: %x0-1F, Whitespace: « », Ascii: «!"#$%&'()*+,»
 
 ; Restrictions on characters in various contexts
 simple-start-char = %x01-08        ; omit NULL (%x00), HTAB (%x09) and LF (%x0A)
--- a/spec/message.abnf
+++ b/spec/message.abnf
@@ -53,39 +53,39 @@ identifier = [namespace ":"] name
 namespace  = name
 name       = [bidi] name-start *name-char [bidi]
 name-start = ALPHA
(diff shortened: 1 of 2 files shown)


macchiati:
Here is a first cut at changes for name-char. Feedback welcome

eemeli (changes requested):
A rather cursory initial pass.

In addition to line comments, this PR needs to be accompanied by some test cases.

spec/message.abnf
```diff
@@ -54,13 +54,39 @@ match = %s".match"
...
-           / %xF900-FDCF / %xFDF0-FFFC / %x10000-EFFFF
-name-char  = name-start / DIGIT / "-" / "."
-           / %xB7 / %x300-36F / %x203F-2040
+name-start = ALPHA
+                  / %x2B          ; 【+】     omit Cc %x0-1F, Whitespace %20, Ascii 【!"#$%&'()*】
+                  / %x5F          ; 【_】     omit Ascii 【,-./0123456789:;<=>?@】 【[\]^】
+                  / %xA1-61B      ;          omit Cc %x7F-9F, Whitespace %xA0, Ascii 【`】 【{|}~】
+                  / %x61D-167F    ;          omit BidiControl %x61C
+                  / %x1681-1FFF   ;          omit Whitespace %x1680
+                  / %x200B-200D   ;          omit Whitespace %x2000-200A
```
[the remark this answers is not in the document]
aphillips: We still have `name-char`. Why not put the joiners in there?

I kind of also question ZWSP

spec/message.abnf
```diff
@@ -54,13 +54,39 @@ match = %s".match"
...
+                  / %x2010-2027   ;          omit BidiControl %x200E-200F
+                  / %x2030-205E   ;          omit Whitespace %x2028-2029 %x202F, BidiControl %x202A-202E
+                  / %x2060-2065   ;          omit Whitespace %x205F
+                  / %x206A-2FFF   ;          omit BidiControl %x2066-2069
+                  / %x3001-D7FF   ;          omit Whitespace %x3000
+                  / %xF900-FDCF   ;          omit Cs %xD800-DFFF, Co %xE000-F8FF
```
aphillips, line 69: This omits the PUA characters (later we omit the PU planes). Is there a reason to avoid private use in names?

spec/syntax.md
```diff
@@ -843,15 +843,62 @@ option     = identifier o "=" o (literal / variable)
...
+> Syntactically, the definitions of `identifier` and `name-char` provide backwards compatibility over time by allowing a stable,
+> wide range of characters.
+> So when there is a new character in a version of Unicode, it can be used in any conformant implementation of Message Format.
+> The definition currently excludes:
+> * Most ASCII except for letters and characters used for numbers
+>    * This avoids conflicts with syntax characters, and reserves some characters for future syntax.
+> * Bidirectional controls (`Bidi_C`)
+> * Control characters (`GC=Cc`, but not Format characters: `GC=Cf`)
+> * Whitespace characters (`WSpace`)
+> * Isolated Surrogate characters (`GC=Cs`)
```
aphillips, line 890: We don't allow them in pairs either. A UTF-16 encoded message might have surrogate code units in it, but they are interpreted as the encoded code point. So:

```suggestion
> * Surrogate code points (`GC=Cs`)
```

spec/syntax.md
```diff
@@ -843,15 +843,62 @@ option     = identifier o "=" o (literal / variable)
...
+> * Private use characters (`GC=Co`)
+> * Non-Characters (`NChar`)
+>
+> Although syntactically a wide range of characters are included,
+> when function and implementations and message authors are creating new identifiers (for functions, options, variables, …),
+> it is strongly recommended that they conform to the following to minimize confusion.
+> These are also recommended for Message Format linter implementations.
```
aphillips, line 897: This is normative language in a "NOTE", which is verboten in our specification. There is no need to make this a note, since this is the specification and is meant to describe what is going on. Something like:

```suggestion
This syntax allows a wide range of characters in _names_ and _identifiers_.
Implementers and authors of _functions_ and _messages_,
including _functions_, _options_, and _operands_ (variable names),
SHOULD avoid creating _names_ that could produce confusion or harm usability
by choosing names consistent with the following guidelines.
MessageFormat tools, such as linters, SHOULD warn when _names_ chosen by users
violate these constraints.
```

(Note that it is `MessageFormat` not `Message Format`)

aphillips (changes requested):
A good start. Some small things.

spec/message.abnf
```diff
@@ -54,13 +54,39 @@ match = %s".match"
...
-           / %x2070-218F / %x2C00-2FEF / %x3001-D7FF
-           / %xF900-FDCF / %xFDF0-FFFC / %x10000-EFFFF
-name-char  = name-start / DIGIT / "-" / "."
-           / %xB7 / %x300-36F / %x203F-2040
+name-start = ALPHA
+                  / %x2B          ; 【+】     omit Cc %x0-1F, Whitespace %20, Ascii 【!"#$%&'()*】
+                  / %x5F          ; 【_】     omit Ascii 【,-./0123456789:;<=>?@】 【[\]^】
+                  / %xA1-61B      ;          omit Cc %x7F-9F, Whitespace %xA0, Ascii 【`】 【{|}~】
+                  / %x61D-167F    ;          omit BidiControl %x61C
+                  / %x1681-1FFF   ;          omit Whitespace %x1680
```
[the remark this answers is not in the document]
macchiati: I don't see a difference on my screen. Do you want an additional space before or after 'omit', or a space deleted before or after 'omit'.

spec/message.abnf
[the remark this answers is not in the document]
macchiati: There is no particular utility to having ZWSP start a name-char, nor any utility to having it end a name-char. (

It doesn't hurt to move that one (ZWSP) to name-char, but it doesn't really make a dent either — and we really wouldn't want to go too far down the very long and slippery slope. That's for linters and guidance.

spec/message.abnf
macchiati: They are fundamentally not interoperable.

That being said, I have no strong objections.

spec/message.abnf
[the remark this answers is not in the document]
aphillips: The question here is why put these characters in `name-start`, where they have no utility? At least in `name-char` they would be enclosed or at the end?

spec/message.abnf
aphillips: I agree that people shouldn't use them, but we otherwise put minimal restrictions on names. Do we care enough to prohibit them at the syntax level? I could be talked into it 😀

macchiati: I'm ok adding them back

spec/message.abnf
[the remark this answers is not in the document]
macchiati: What I'm afraid of is that if we move that to name-char, it will just open it up to people endlessly complaining that:
> "ZWSP" is in name-chart instead of name-start: why is XXX in name-start when it should also be just be in name-chart???:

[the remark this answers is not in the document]
macchiati: That is, the basic difference between name-char and name-start is that 

- name is used in identifiers and variables, and can't start with a digit, -, or .
- name-char is used in literals, and can start with digit, -, .

The syntactic motivation is clear: to make sure that identifiers and variables are distinguishable from numbers. That is a clear syntactic need.

ZWSP certainly isn't _needed_ at the start of an identifier or variable, but there is an large and complicated list of characters that are _also_ not needed at start of identifiers and variables, and plucking just one of those characters out, without any syntactic need, doesn't actually provide much value.

spec/message.abnf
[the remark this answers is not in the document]
macchiati: I changed the EA brackets to guillemets, since they line up better for monospace.

spec/message.abnf
macchiati: Added back.

spec/message.abnf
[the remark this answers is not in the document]
eemeli: I meant offset in a vertical direction, so a comment like "omit BidiControl %x61C" should follow the range `%xA1-61B`, rather than the range `%x61D-167F`.

spec/message.abnf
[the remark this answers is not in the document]
eemeli: Given that we do not allow space characters or control characters, I'd prefer not allowing zero-width spaces in names or unquoted literals.

eemeli:
This also needs to be fixed: https://github.com/unicode-org/message-format-wg/blob/1fccd1ee734808c02f7b65c2f4c877b4c2fde1f1/spec/syntax.md?plain=1#L789-L793

spec/message.abnf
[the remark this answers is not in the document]
macchiati: Ok, will work on that.

spec/message.abnf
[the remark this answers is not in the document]
macchiati: A zero-width space is **not** a space; that is just a name used for familiarity. It is a Format character, like many others.

https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5Cp%7Bgc%3Dformat%7D&g=&i=

spec/syntax.md
```diff
@@ -786,11 +786,9 @@ has been applied to both.
...
 > to _name_ values.
 
-Valid content for _names_ is based on <cite>Namespaces in XML 1.0</cite>'s 
-[NCName](https://www.w3.org/TR/xml-names/#NT-NCName).
-This is different from XML's [Name](https://www.w3.org/TR/xml/#NT-Name)
-in that it MUST NOT contain a U+003A COLON `:`.
-Otherwise, the set of characters allowed in a _name_ is large.
+The _names_ are [immutable identifiers](https://www.unicode.org/reports/tr31/#Immutable_Identifier_Syntax).
+They are similar to <cite>Namespaces in XML 1.0</cite>'s [NCName](https://www.w3.org/TR/xml-names/#NT-NCName),
+but have been updated to be more consistent.
```
eemeli, line 791: Given how far we've departed from that spec, this note seems only relevant within the history of the spec, and not with where we're ending up with this PR. I'd prefer dropping it, and moving the preceding sentence (if keeping) above the preceding note.

macchiati: ok.

spec/message.abnf
```diff
@@ -49,18 +49,43 @@ local = %s".local"
...
 name       = [bidi] name-start *name-char [bidi]
-name-start = ALPHA / "_"
-           / %xC0-D6 / %xD8-F6 / %xF8-2FF
-           / %x370-37D / %x37F-61B / %x61D-1FFF / %x200C-200D
-           / %x2070-218F / %x2C00-2FEF / %x3001-D7FF
-           / %xF900-FDCF / %xFDF0-FFFC / %x10000-EFFFF
-name-char  = name-start / DIGIT / "-" / "."
-           / %xB7 / %x300-36F / %x203F-2040
+name-start = ALPHA
+                                    ;          omit Cc: %x0-1F, Whitespace: « », Ascii: «!"#$%&'()*»
```
aphillips, line 56: This looks weird: Whitespace: « »

```suggestion
                                    ;          omit Cc: %x0-1F, Whitespace: SPACE, Ascii: «!"#$%&'()*»

**Review by aphillips — APPROVED** [review]

File: spec/message.abnf [hunk 8]
```diff
@@ -49,18 +49,43 @@ local = %s".local"
...
+                  / %xA0000-AFFFD   ;          omit NChar: %xAFFFE-AFFFF
+                  / %xB0000-BFFFD   ;          omit NChar: %xBFFFE-BFFFF
+                  / %xC0000-CFFFD   ;          omit NChar: %xCFFFE-CFFFF
+                  / %xD0000-DFFFD   ;          omit NChar: %xDFFFE-DFFFF
+                  / %xE0000-EFFFD   ;          omit NChar: %xEFFFE-EFFFF
+                  / %xF0000-FFFFD   ;          omit NChar: %xFFFFE-FFFFF
+                  / %x100000-10FFFD ;          omit NChar: %x10FFFE-10FFFF
+
+name-char  = name-start / DIGIT
+                  / %x2D-2E         ; «-.»
```
eemeli, line 88: This would be easier to read, and it's what we currently have:
```suggestion
name-char  = name-start / DIGIT / "-" / "."
```

spec/syntax.md
```diff
@@ -843,15 +839,66 @@ option     = identifier o "=" o (literal / variable)
...
+                  / %xA0000-AFFFD   ;          omit NChar: %xAFFFE-AFFFF
+                  / %xB0000-BFFFD   ;          omit NChar: %xBFFFE-BFFFF
+                  / %xC0000-CFFFD   ;          omit NChar: %xCFFFE-CFFFF
+                  / %xD0000-DFFFD   ;          omit NChar: %xDFFFE-DFFFF
+                  / %xE0000-EFFFD   ;          omit NChar: %xEFFFE-EFFFF
+                  / %xF0000-FFFFD   ;          omit NChar: %xFFFFE-FFFFF
+                  / %x100000-10FFFD ;          omit NChar: %x10FFFE-10FFFF
+
+name-char  = name-start / DIGIT
+                  / %x2D-2E         ; «-.»
```
eemeli, line 875:
```suggestion
name-char  = name-start / DIGIT / "-" / "."
```

eemeli (approved):

spec/message.abnf
```diff
@@ -49,18 +49,41 @@ local = %s".local"
...
 name       = [bidi] name-start *name-char [bidi]
-name-start = ALPHA / "_"
-           / %xC0-D6 / %xD8-F6 / %xF8-2FF
-           / %x370-37D / %x37F-61B / %x61D-1FFF / %x200C-200D
-           / %x2070-218F / %x2C00-2FEF / %x3001-D7FF
-           / %xF900-FDCF / %xFDF0-FFFC / %x10000-EFFFF
+name-start = ALPHA
+                                    ;          omit Cc: %x0-1F, Whitespace: SPACE, Ascii: «!"#$%&'()*»
+                  / %x2B            ; «+»      omit Ascii: «,-./0123456789:;<=>?@» «[\]^»
+                  / %x5F            ; «_»      omit Cc: %x7F-9F, Whitespace: %xA0, Ascii: «`» «{|}~»
```
gibson042, line 58: Why are there two sets of guillemet-quoted ASCII characters here?
```suggestion
                  / %x2B            ; «+»      omit Ascii: «,-./0123456789:;<=>?@[]^» and REVERSE SOLIDUS "\"
                  / %x5F            ; «_»      omit Ascii: «`{|}~», Cc: %x7F-9F, Whitespace: %xA0
```

PR #1010, opened 2025-02-17
Fix normative language in notes

eemeli:
I found two notes with normative language; this fixes one to be IMPORTANT instead (and removes the character range comment, as we've since added namespaces that should be used), and drops the unnecessary phrase from the other.

### DIFF
--- a/spec/intro.md
+++ b/spec/intro.md
@@ -119,7 +119,8 @@ which satisfy either of the following two requirements:
 All other _identifiers_ in these categories are reserved for the use of implementations or users.
 
 > [!IMPORTANT]
-> Users defining custom _identifiers_ SHOULD use a _namespace_
+> Implementation-defined or user-defined _functions_ and _function_ _options_
+> SHOULD use a _namespace_ as part of their _identifiers_
 > to avoid collisions with other implementations.
 
 Future versions of this specification will not introduce changes
--- a/spec/intro.md
+++ b/spec/intro.md
@@ -118,10 +118,9 @@ which satisfy either of the following two requirements:
 
 All other _identifiers_ in these categories are reserved for the use of implementations or users.
 
-> [!NOTE]
-> Users defining custom _identifiers_ SHOULD include at least one character outside these ranges
-> to ensure that they will be compatible with future versions of this specification.
-> They SHOULD also use the _namespace_ feature to avoid collisions with other implementations.
+> [!IMPORTANT]
+> Users defining custom _identifiers_ SHOULD use a _namespace_
+> to avoid collisions with other implementations.
 
 Future versions of this specification will not introduce changes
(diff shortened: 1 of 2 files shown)


spec/intro.md
```diff
@@ -118,10 +118,9 @@ which satisfy either of the following two requirements:
 
 All other _identifiers_ in these categories are reserved for the use of implementations or users.
 
-> [!NOTE]
-> Users defining custom _identifiers_ SHOULD include at least one character outside these ranges
-> to ensure that they will be compatible with future versions of this specification.
-> They SHOULD also use the _namespace_ feature to avoid collisions with other implementations.
+> [!IMPORTANT]
+> Users defining custom _identifiers_ SHOULD use a _namespace_
```
macchiati, line 122: It is not clear exactly what 'custom identifiers' means. Does that mean "all other identifiers"? If so, the text should say so

What does "users" mean in this context? Authors of messages? Authors of functions?

identifiers are used in functions, options, option values, and attributes.

The above implies that I (a message writter) should use a namespace in an option value that is a free-form string wit non-ascii. So instead of:

... name=René ...

I should write: name=med:René ...

Moreover, "to avoid collisions with other implementations." needs to be toned down, to "to help avoid collisions with other implementations.". Using a namespace is no guarantee of avoidance: there could easily be another 'med' namespace.

eemeli: Note that identifiers are not used in option values.

The point about attributes is valid; this needs to be fixed for that. If the following addresses your concerns, please merge this suggestion?

```suggestion
> Users defining custom _identifiers_ for _functions_ and _function_ _options_ SHOULD use a _namespace_
```

aphillips: A better overall phrasing would be more like:

```suggestion
Implementation-defined or user-defined _functions_ and _function_ _options_ SHOULD use a _namespace_
as part of their _identifiers_
```

eemeli: Merging with editorial edits.

spec/intro.md
```diff
@@ -118,10 +118,10 @@ which satisfy either of the following two requirements:
...
 All other _identifiers_ in these categories are reserved for the use of implementations or users.
 
-> [!NOTE]
-> Users defining custom _identifiers_ SHOULD include at least one character outside these ranges
-> to ensure that they will be compatible with future versions of this specification.
-> They SHOULD also use the _namespace_ feature to avoid collisions with other implementations.
+> [!IMPORTANT]
+> Implementation-defined or user-defined _functions_ and _function_ _options_
+> SHOULD use a _namespace_ as part of their _identifiers_
+> to avoid collisions with other implementations.
```
aphillips, line 124:
```suggestion
> to help avoid collisions with other implementations.
```

PR #1011, opened 2025-02-17
Require prioritising syntax & data model errors

eemeli:
We should tighten up our requirement about error priority; there's no reason to keep this as a SHOULD instead of a MUST. The syntax & data model errors are also determinable before runtime.

### DIFF
--- a/spec/errors.md
+++ b/spec/errors.md
@@ -44,7 +44,7 @@ or separately by more than one such method.
 When a message contains more than one error,
 or contains some error which leads to further errors,
 an implementation which does not emit all of the errors
-SHOULD prioritise _Syntax Errors_ and _Data Model Errors_ over others.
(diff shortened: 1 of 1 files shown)


spec/errors.md
```diff
@@ -44,7 +44,7 @@ or separately by more than one such method.
 When a message contains more than one error,
 or contains some error which leads to further errors,
 an implementation which does not emit all of the errors
-SHOULD prioritise _Syntax Errors_ and _Data Model Errors_ over others.
+MUST prioritise _Syntax Errors_ and _Data Model Errors_ over others.
```
aphillips, line 47: I don't think we should take this today. We can make it MUST in 48.

macchiati: I agree, for discussion for v48

PR #1012, opened 2025-02-17
Define optionality separately for each u: option

eemeli:
This clarifies the requirement status for each `u:` option separately, as `u:id` only has an effect if the output can be something other than a single string, and concerns have been raised about `u:locale`.

I would be very happy to say that `u:id` and `u:dir` MUST be supported, if the rest of the WG agrees.

This should also clarify that an implementation might choose to support a subset of the `u:` options.

### DIFF
--- a/spec/u-namespace.md
+++ b/spec/u-namespace.md
@@ -12,10 +12,13 @@ manages the specification for this namespace, hence the _namespace_ `u:`.
 
 ## Options
 
-This section describes common **_<dfn>`u:` options</dfn>_** which each implementation SHOULD support
+This section describes common **_<dfn>`u:` options</dfn>_** that apply
 for all _functions_ and _markup_.
 
 ### `u:id`
+
+Implementations providing a formatting target other than a concatenated string
+SHOULD support this option.
 
 A string value that is included as an `id` or other suitable value
 in the formatted parts for the _placeholder_,
@@ -30,6 +33,8 @@ For other values, a _Bad Option_ error is emitted
 and the `u:id` option is ignored.
 
 ### `u:locale`
+
+Implementations MAY support this option.
 
 Replaces the _locale_ defined in the _function context_ for this _expression_.
 
@@ -66,6 +71,8 @@ or because the language tag is not well-formed,
 not valid, or some other reason.
 
 ### `u:dir`
+
+Implementations SHOULD support this option.
 
 Replaces the base directionality defined in
 the _function context_ for this _expression_
--- a/spec/u-namespace.md
+++ b/spec/u-namespace.md
@@ -12,8 +12,9 @@ manages the specification for this namespace, hence the _namespace_ `u:`.
 
 ## Options
 
-This section describes common **_<dfn>`u:` options</dfn>_** that apply
-for all _functions_ and _markup_.
+This section describes **_<dfn>`u:` options</dfn>_**.
+When implemented, they apply to all _functions_ and _markup_,
+including user-defined _functions_ in that implementation.
 
 ### `u:id`
 
--- a/spec/u-namespace.md
+++ b/spec/u-namespace.md
@@ -1,4 +1,4 @@
-# MessageFormat 2.0 Unicode Namespace
+# MessageFormat Unicode Namespace
 
 The `u:` _namespace_ is reserved for the definition of _options_
 which affect the _function context_ of the specific _expressions_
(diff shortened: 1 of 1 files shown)


spec/u-namespace.md
```diff
@@ -12,11 +12,14 @@ manages the specification for this namespace, hence the _namespace_ `u:`.
 
 ## Options
 
-This section describes common **_<dfn>`u:` options</dfn>_** which each implementation SHOULD support
+This section describes common **_<dfn>`u:` options</dfn>_** that apply
 for all _functions_ and _markup_.
 
 ### `u:id`
 
+Implementations providing a formatting target other than a concatenated string
```
macchiati, line 20: What does it imply for implementations to say that u:id is supported? We should say _something_ about the intended usage.

spec/u-namespace.md
```diff
@@ -12,11 +12,14 @@ manages the specification for this namespace, hence the _namespace_ `u:`.
 
 ## Options
 
-This section describes common **_<dfn>`u:` options</dfn>_** which each implementation SHOULD support
+This section describes common **_<dfn>`u:` options</dfn>_** that apply
 for all _functions_ and _markup_.
```
aphillips, line 16: I don't like this phrasing, as it changes the meaning, not just the normativity.

```suggestion
This section describes **_<dfn>`u:` options</dfn>_**.
When implemented, they apply to all _functions_ and _markup_,
including user-defined _functions_ in that implementation.
```

spec/u-namespace.md
aphillips: We should insert the examples from the design doc

spec/u-namespace.md
```diff
@@ -31,6 +34,8 @@ and the `u:id` option is ignored.
 
 ### `u:locale`
 
+Implementations MAY support this option.
```
aphillips, line 37: I think this should be SHOULD.

I know we're discussing the ICU4X objection. The common use case for this option is is basically "I18N demos", which I write a lot of 😉:
```
The default date format in locale: {$locale :x:localeGetDisplayName} is {$now :date u:locale=$locale}
```

There are cases in which users wish to tailor the locale, for example using the `-u` extension for options but do not wish to transmit the "decorated" locale everywhere. Some of these items are option values on built-in functions, but overriding (for example) the digit shaping in the locale works for functions that have no such options.

A fairly common case is wanting to override the locale to be `und`/`Locale.ROOT` or some POSIX variant to get a locale-neutral representation. 

Encouraging implementations to provide this feature with the stronger normative SHOULD does not mean that implementations that cannot support it are "bad".

spec/u-namespace.md
```diff
@@ -67,6 +72,8 @@ not valid, or some other reason.
 
 ### `u:dir`
 
+Implementations SHOULD support this option.
```
aphillips, line 75: I don't think we can or should make any `u:` option MUST in this release, but really everyone should implement this feature. Otherwise string metadata about direction has nowhere to go. For those playing at home, my [UTW presentation](https://www.youtube.com/watch?v=eJEKbUx3B0w) explains why.

spec/u-namespace.md
eemeli: I don't think i18n demos is a use case that we ought to support if/as it adds complexity to an implementation.

If the numbering system is desirable to fix for a single function in a single message, wouldn't it also be desirable to fix for all messages, and therefore be encoded in the locale that's used to format the message or messages?

For ensuring that something is formatted in a locale-neutral representation, wouldn't a programmer want to submit that to the message formatter in a pre-formatted form, to guarantee that localization will not affect it at all?

aphillips: Consider a case like:

```
The message shouted: {$text :transform type=uppercase u:locale=$textLocale} <- what if the string is Turkish?
```

There are other cases. I think that on-going discussion suggests moving `u:locale` to draft and that we rebuild a design document rather than including it in 47 as stable.

sffc: Thank you very much for illuminating a use case; this is more than I've seen before.

However, I'm not convinced this rises to the level of a SHOULD.

Here's my litmus test: if you are in a situation where you are able to choose your MF implementation, then your use case is a MAY. If you are in a situation where your environment chooses your MF implementation for you, then your use case is a SHOULD.

Given that Addison is writing his own strings for his own personal use, I think this lands in the MAY category.

aphillips: > Given that Addison is writing his own strings for his own personal use, I think this lands in the MAY category.

I don't understand this statement? Every MF2 user is "writing their own strings for their own personal use"? Do you mean "you're writing your own function `:transform`"? 

Most users will be in situations in which their MF2 implementation is the one in their platform or in a library such as ICU. Their goal and task is to write messages, not to cherry-pick different choices.

I don't think we have time this morning to make `u:locale` satisfy everyone and I think we should employ our WG process--which requires a design doc. We have ample time in v48 to do this correctly and zero time to get it right just now. Hence, I propose moving `u:locale` to DRAFT.

PR #1013, opened 2025-02-17
Require digit size to support values 0-99

eemeli:
Digit size option values MAY have an implementation-defined upper limit, but let's make sure that does not exclude anything allowed by the `digit-size-option` rule for literal values.

### DIFF
--- a/spec/functions/number.md
+++ b/spec/functions/number.md
@@ -640,7 +640,8 @@ such as the number of fraction, integer, or significant digits.
 A "digit size option" is an _option_ value that the _function_ interprets
 as a small integer value greater than or equal to zero.
 Implementations MAY define an upper limit on the _resolved value_
-of a digit size option option consistent with that implementation's practical limits.
+of a digit size option option consistent with that implementation's practical limits,
+as long as all integer values from 0 to 99 (inclusive) are supported.
 
 In most cases, the value of a digit size option will be a string that
 encodes the value as a non-negative integer.
--- a/spec/functions/number.md
+++ b/spec/functions/number.md
@@ -640,7 +640,7 @@ such as the number of fraction, integer, or significant digits.
 A "digit size option" is an _option_ value that the _function_ interprets
 as a small integer value greater than or equal to zero.
 Implementations MAY define an upper limit on the _resolved value_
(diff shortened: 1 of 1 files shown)


spec/functions/number.md
```diff
@@ -640,7 +640,8 @@ such as the number of fraction, integer, or significant digits.
 A "digit size option" is an _option_ value that the _function_ interprets
 as a small integer value greater than or equal to zero.
 Implementations MAY define an upper limit on the _resolved value_
-of a digit size option option consistent with that implementation's practical limits.
+of a digit size option option consistent with that implementation's practical limits,
+as long as all integer values from 0 to 99 (inclusive) are supported.
```
macchiati, line 644: 99 is way beyond the significant digits supported by an IEEE754 double. If we are going to require minimal values, it shouldn't be more than needed with double, which means it may vary according to the option.

aphillips: I agree that this isn't the right fix. I think what you're trying to say is maybe:

```suggestion
of a digit size option consistent with that implementation's practical limits.
Implementations that do so MUST emit an _Unsupported Operation_ error.
The literal values for integers from 0 to 99 (inclusive) MUST NOT produce
a _Bad Option_ error.
```

(Note that your text has "option option")

spec/functions/number.md
```diff
@@ -640,7 +640,8 @@ such as the number of fraction, integer, or significant digits.
 A "digit size option" is an _option_ value that the _function_ interprets
 as a small integer value greater than or equal to zero.
 Implementations MAY define an upper limit on the _resolved value_
-of a digit size option option consistent with that implementation's practical limits.
+of a digit size option option consistent with that implementation's practical limits,
```
eemeli, line 643:
```suggestion
of a digit size option consistent with that implementation's practical limits,
```

spec/functions/number.md
eemeli: My intent here was to clarify the general case, but arguably that does complicate matters; the current language allows more clearly for an implementation to define a different upper bound for each digit size option.

The "option option" text is currently in the spec; we should fix that separately if discarding this change.

aphillips:
closing per 2025-02-17 call discussion. typo to be fixed in 1007

PR #1014, opened 2025-02-17
Drop the u:locale option

eemeli:
I'm somewhat ambivalent on this myself, but filing this as a draft so that a WG decision becomes a little easier to enact.

I'd be fine leaving this out of the '47 release, as we can add it in if/once user need for it can be demonstrated.

There's a minor overlap with #1012, which should be easy to resolve.


aphillips:
we went another direction in the 2025-02-17 call

PR #1015, opened 2025-02-17
Drop the notation, compactDisplay, & numberingSystem options

eemeli:
This fixes some of the issues identified in the ICU4X TC in the doc linked from https://github.com/unicode-org/message-format-wg/issues/1006#issuecomment-2659730990 by removing the notation, compactDisplay, and numberingSystem options from all functions that currently have them.

This is not a complete solution, nor is it the one preferred by the ICU4X TC, but for at least these options it's the simplest thing we can do, and as @aphillips notes in a comment on the doc:
> Compact display strikes me as a candidate for removal to another function, since not all platforms have this feature in their low-level I18N libraries. We hived off percent. Compact display seems more optional.

The removal of `notation` and `compactDisplay` was previously considered in #664, but dismissed on [2024-02-26](https://github.com/unicode-org/message-format-wg/blob/main/meetings/2024/notes-2024-02-26.md#664).

Regarding `numberingSystem`, we should take it out for now and re-evaluate what shape its option ought to take after the '47 release. Again, quoting @aphillips:
> [Toggling between Latin and native digit shaping] is a VERY common thing to want to do. Forcing Latin doesn't require a variable, but setting native digits does with the current design, since you have to specify the script. We might provide a boolean option to turn "digit shaping" on/off to reduce the incidence of needing `numberingSystem` as an option. I would even support removing this option in favor of such a switch. If you want arbitrary digit shaping, you are doing something super-special and probably not I18N best practices.

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -116,10 +116,10 @@ which makes it available for use in later _expressions_ and _markup_ _options_.
 > For example, in
 > ```
 > .input {$a :number minimumFractionDigits=3}
-> .local $b = {$a :integer notation=compact}
+> .local $b = {$a :integer useGrouping=never}
 > .match $a
 > 0 {{The value is zero.}}
-> * {{In compact form, the value {$a} is rendered as {$b}.}}
+> * {{Without grouping separators, the value {$a} is rendered as {$b}.}}
 > ```
 > the _resolved value_ bound to `$a` is used as the _operand_
 > of the `:integer` _function_ when resolving the value of the _variable_ `$b`,
@@ -991,6 +991,4 @@ The _Default Bidi Strategy_ is defined as follows:
 > Directionality SHOULD NOT be determined by introspecting
 > the character sequence in the formatted string representation
 > of `resval`.
-
-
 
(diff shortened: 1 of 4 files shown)


spec/formatting.md
```diff
@@ -116,10 +116,10 @@ which makes it available for use in later _expressions_ and _markup_ _options_.
 > For example, in
 > ```
 > .input {$a :number minimumFractionDigits=3}
-> .local $b = {$a :integer notation=compact}
+> .local $b = {$a :integer useGrouping=never}
```
aphillips, line 119: We should make this change regardless

spec/functions/README.md
```diff
@@ -73,9 +73,6 @@ Such values can be useful to users in cases where local usage and support exists
 >   for currency amounts as the _operand_ in the _function_ `:currency`.
 > - A Java implementation might _accept_ a `java.time.chrono.Chronology` object
 >   as a value for the _date/time override option_ `calendar`
-> - ICU4J's implementation might _accept_ a `com.ibm.icu.text.NumberingSystem` object
->   instead of using a [Unicode Numbering System Identifier](https://unicode.org/reports/tr35/tr35.html#UnicodeNumberSystemIdentifier)
->   for the option `numberingSystem` in _functions_ such as `:number` or `:integer`.
```
aphillips, line 78: In an ideal world we would find a replacement here, but not this morning.

aphillips (approved):
I thought about suggesting keeping these in Draft, but suspect that we'll do something different. No reason for us to mislead implementers.

sffc:
Sure, if you want to do this and punt on the problem, fine with me.

I'm not convinced compact notation should be moved to a separate function, but that's a discussion we can have at another time.

PR #1016, opened 2025-02-17
Require select option to be set by a literal value

eemeli:
For context, see #1006 

As pointed out by @aphillips in a comment on the ICU4X-TC feedback doc:
> [It's] nearly impossible to localize a message in which [the `select`] option is passed by reference, since tools and translators can't tell how to translate the messages or how to explode (expand/contract) the list of keywords in a given locale.

We can't really make using a variable for `select` a syntax or data model error unless we add a really special case for it in the Syntax part, but this is _nearly_ that.

Note that as shown in the test cases, formatting the following message would result in `bad-option` and `bad-selector` errors:

```
.local $sel = {1 :number select=exact}
.local $bad = {$sel :number}
.match $bad
1 {{ONE}}
* {{operand select {$bad}}}
```

### DIFF
--- a/spec/functions/number.md
+++ b/spec/functions/number.md
@@ -658,7 +658,7 @@ or any option-specific lower limit, a _Bad Option Error_ is emitted.
 ### Number Selection
 
 The value of the `select` _option_ MUST be set by a _literal_,
-as otherwise the message might not be translatable.
+as otherwise the _message_ might not be translatable.
 If this value is set by a _variable_ or
 the option value of an implementation-defined type used as an _operand_,
 a _Bad Option Error_ is emitted and
--- a/spec/functions/number.md
+++ b/spec/functions/number.md
@@ -22,8 +22,8 @@ the value of other options, or both.
 
 The following options and their values are REQUIRED to be available on the function `:number`:
 
-- `select`
-  - `plural` (default; see [Default Value of `select` Option](#default-value-of-select-option) below)
+- `select` (see [Number Selection](#number-selection) below)
+  - `plural` (default)
   - `ordinal`
   - `exact`
 - `compactDisplay` (this option only has meaning when combined with the option `notation=compact`)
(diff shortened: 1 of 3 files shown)


spec/functions/number.md
```diff
@@ -657,6 +657,14 @@ or any option-specific lower limit, a _Bad Option Error_ is emitted.
 
 ### Number Selection
 
+The value of the `select` _option_ MUST be set by a _literal_,
+as otherwise the message might not be translatable.
+If this value is set by a _variable_ or
+the option value of an implementation-defined type used as an _operand_,
+a _Bad Option Error_ is emitted and
+the _resolved value_ of the expression MUST NOT support selection.
```
aphillips, line 665: link cleanup

```suggestion
as otherwise the _message_ might not be translatable.
If this value is set by a _variable_ or
the _option_ _value_ of an implementation-defined type used as an _operand_,
a _Bad Option Error_ is emitted and
the _resolved value_ of the expression MUST NOT support selection.
```

aphillips (approved):

spec/functions/number.md
eemeli: The first fix is valid, but _option_ _value_ is incorrect when referring to an implementation-defined type, because its values are not syntax values.
```suggestion
as otherwise the _message_ might not be translatable.
If this value is set by a _variable_ or
the option value of an implementation-defined type used as an _operand_,
a _Bad Option Error_ is emitted and
the _resolved value_ of the expression MUST NOT support selection.
```

test/tests/functions/number.json
```diff
@@ -185,6 +185,36 @@
         }
       ]
     },
+    {
+      "src": "literal select {1 :number select=exact}",
+      "exp": "literal select {1}"
+    },
+    {
+      "src": ".local $bad = {exact} {{variable select {1 :number select=$bad}}}",
```
sffc, line 193: Suggestion: add a test similar to this one but with .input

PR #1017, opened 2025-02-17
Require u:locale option to be set by a literal value

eemeli:
This is an alternative to #1014, and requires the `u:locale` to have a literal value using similar spec language as proposed for `select` in #1016.

Filing this too as a draft, to more clearly show what this change would look like and to make it easier to adopt if we so choose.


aphillips:
we went another direction in the 2025-02-17 call

PR #1019, opened 2025-02-17
Include explicit text on literal option origin

eemeli:
On today's call, we decided to merge #1016, and to add text to the main part of the spec clarifying the need to carry the information about an option value's origin in its _resolved value_. That information is here added in three places:
- The _resolved value_ definition, with a normative MUST requirement.
- The _resolved value_ example.
- The _option resolution_ algorithm.

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -138,8 +138,8 @@ its formatted representation requires isolation
 from the surrounding text.
 
 To allow for _function handlers_ to ensure that certain _option_ values are set by _literals_,
-the _resolved value_ of each _option_ value MUST include information about its source
-as either a _literal_ or _variable_.
+the _resolved value_ of each _option_ value MUST include information about 
+whether the _option_ value is a _literal_ or a _variable_.
 
 The form that _resolved values_ take is implementation-dependent,
 and different implementations MAY choose to perform different levels of resolution.
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -137,6 +137,10 @@ as well as a flag to indicate whether
 its formatted representation requires isolation
 from the surrounding text.
 
+To allow for _function handlers_ to ensure that certain _option_ values are set by _literals_,
+the _resolved value_ of each _option_ value MUST include information about its source
+as either a _literal_ or _variable_.
+
 The form that _resolved values_ take is implementation-dependent,
 and different implementations MAY choose to perform different levels of resolution.
 
@@ -153,6 +157,7 @@ and different implementations MAY choose to perform different levels of resoluti
 >   selectKeys(keys: string[]): string[]
 >   directionality(): 'LTR' | 'RTL' | 'unknown'
 >   isolate(): boolean
+>   isLiteralOptionValue(): boolean
 > }
 > ```
 >
@@ -369,7 +374,7 @@ Implementation-defined _functions_ SHOULD use an implementation-defined _namespa
 
 **_<dfn>Option resolution</dfn>_** is the process of computing the _options_
 for a given _expression_. 
-_Option resolution_ results in a mapping of string _identifiers_ to _values_.
+_Option resolution_ results in a mapping of string _identifiers_ to _resolved values_.
 The order of _options_ MUST NOT be significant.
 
 > For example, the following _message_ treats both both placeholders identically:
@@ -386,6 +391,8 @@ For each _option_:
    1. If `rv` is a _fallback value_:
       1. If supported, emit a _Bad Option_ error.
    1. Else:
+      1. If the _option_ value was set by a _literal_:
+         1. Include that information in `rv`.
       1. Set `res[id]` to be `rv`.
 1. Return `res`.
 


spec/formatting.md
```diff
@@ -369,7 +374,7 @@ Implementation-defined _functions_ SHOULD use an implementation-defined _namespa
 
 **_<dfn>Option resolution</dfn>_** is the process of computing the _options_
 for a given _expression_. 
-_Option resolution_ results in a mapping of string _identifiers_ to _values_.
+_Option resolution_ results in a mapping of string _identifiers_ to _resolved values_.
```
gibson042, line 377: :+1:

spec/formatting.md
```diff
@@ -386,6 +391,8 @@ For each _option_:
    1. If `rv` is a _fallback value_:
       1. If supported, emit a _Bad Option_ error.
    1. Else:
+      1. If the _option_ value was set by a _literal_:
+         1. Include that information in `rv`.
       1. Set `res[id]` to be `rv`.
 1. Return `res`.
```
gibson042, line 397: I think we should probably take greater pains to explain transitivity. For example, consider
```
.input {$x :number}
.local $a = {|EUR|}
.local $b = {$a}
{{Balance: {$x :currency currency=$b}.}}
```
1. Per [Literal Resolution](https://github.com/unicode-org/message-format-wg/blob/2b3b5706cd7319e73119f7d4422e4aabcbbe6a6c/spec/formatting.md#literal-resolution), the resolved value of $a is "EUR" (and per the new text in [Resolved Values](https://github.com/unicode-org/message-format-wg/blob/2b3b5706cd7319e73119f7d4422e4aabcbbe6a6c/spec/formatting.md#resolved-values), includes its literal nature).
2. Per [Variable Resolution](https://github.com/unicode-org/message-format-wg/blob/2b3b5706cd7319e73119f7d4422e4aabcbbe6a6c/spec/formatting.md#variable-resolution), the resolved value of $b is just the resolved value of $a (i.e., a literal with contents "EUR")—although I cannot determine whether or not implementations are given freedom to drop that propagation by «An implementation MAY perform additional processing when resolving the value of an <em>expression</em> that consists only of a <em>variable</em>».
3. These new steps in [Option Resolution](https://github.com/unicode-org/message-format-wg/blob/2b3b5706cd7319e73119f7d4422e4aabcbbe6a6c/spec/formatting.md#option-resolution) don't _add_ literalness to the resolved value of $b, but it's already there so the `:currency` function will still see it.

In fact, it seems like the new steps here cannot ever have an effect, because the resolved value of a literal option is marked as such in [Literal Resolution](https://github.com/unicode-org/message-format-wg/blob/2b3b5706cd7319e73119f7d4422e4aabcbbe6a6c/spec/formatting.md#literal-resolution) and the resolved value of a variable referencing another variable is just the resolved value of its target. So rather than changing this algorithm, the PR should instead accompany it with note(s) explaining the above.

However, more normative text _is_ required is for explaining a variant where `.local $b = {$a}` is replaced by `.local $b = {$a :string}`, in which it is not currently clear whether or not [Function Resolution](https://github.com/unicode-org/message-format-wg/blob/2b3b5706cd7319e73119f7d4422e4aabcbbe6a6c/spec/formatting.md#function-resolution) grants implementations the freedom to return a result that is marked as a literal. I strongly believe that such freedom should **not** exist, and that literalness be preserved **only** by direct isolated variable references.

eemeli: The intent of the changes here is make it so that
```
{{Balance: {$x :currency currency=EUR}.}}
```
would have the value of the currency option marked as coming from a literal, while
```
.local $a = {|EUR|}
{{Balance: {$x :currency currency=$a}.}}
```
would not.

To make that clearer, maybe the new normative text in Resolved Values needs some work? I'll add a suggestion there.

spec/formatting.md
```diff
@@ -137,6 +137,10 @@ as well as a flag to indicate whether
 its formatted representation requires isolation
 from the surrounding text.
 
+To allow for _function handlers_ to ensure that certain _option_ values are set by _literals_,
+the _resolved value_ of each _option_ value MUST include information about its source
+as either a _literal_ or _variable_.
```
eemeli, line 142: Ping @gibson042 

```suggestion
To allow for _function handlers_ to ensure that certain _option_ values are set by _literals_,
the _resolved value_ of each _option_ value MUST include information about 
whether the _option_ value is a _literal_ or a _variable_.
```

spec/formatting.md
macchiati: I agree with @eemeli that what he has is sufficient for now.

The stated problems were where the resolved value of a currency option depended on an input parameter, directly or indirectly. However, requiring detection of that finer-grained information, while possible, seems like a lot of overkill. For the use cases we are concerned about in v47, it is sufficient _to enable and allow functions to decline variables, but allow literals_.

We can examine later whether we need to refine that further, _after_ 47.

macchiati (approved):

spec/formatting.md
aphillips: I think the text could be misinterpreted. Is there any reason not to say:

```
      1. If the _option_ value consists of a _literal_:
         1. Include that information in `rv`.
 ```

Comment (eemeli) [reply spec/formatting.md:397]: That works for me.

@gibson042 Would this and the change suggested for line 142 address your concerns?

```suggestion
   1. If `rv` is a _fallback value_:
      1. If supported, emit a _Bad Option_ error.
   1. Else:
      1. If the _option_ value consists of a _literal_:
         1. Include that information in `rv`.
      1. Set `res[id]` to be `rv`.
1. Return `res`.
```

gibson042: I think the algorithm also needs an "Else" to _strip_ any "consists of a literal" information from **non**-literal option values.

spec/formatting.md
gibson042: This suggestion looks good to me.

spec/formatting.md
gibson042:
```suggestion
   1. If `rv` is a _fallback value_:
      1. If supported, emit a _Bad Option_ error.
   1. Else, if the _option_ value consists of a _literal_:
      1. Set `res[id]` to be `rv`.
   1. Else:
      1. Let `variableRV` be a copy of `rv` which identifies its source as a _variable_.
      1. Set `res[id]` to be `variableRV`.
1. Return `res`.
```

eemeli: I don't see how that's better than encoding the information "this _option_ value consists of a _literal_".

spec/formatting.md
```diff
@@ -137,6 +137,10 @@ as well as a flag to indicate whether
 its formatted representation requires isolation
 from the surrounding text.
 
+To allow for _function handlers_ to ensure that certain _option_ values are set by _literals_,
+the _resolved value_ of each _option_ value MUST include information about 
+whether the _option_ value is a _literal_ or a _variable_.
```
aphillips, line 142: Reading the below commentary thread and thinking about the problem more, I'm going to suggest that we solve the problem here rather than trying to spruce up the algorithm. By better explaining the situation, the algorithm might be clearer?

```suggestion
To allow _functions_ to limit _option_ values to _literal_ values explicitly assigned in the _message_,
the _resolved value_ of each _option_ value MUST include information about
whether the _option_ value consists of a _literal_ or, at any point in the resolution process, 
consisted of a _variable_.

> [!NOTE]
> An example of this is the `select` _option_ in the _default functions_ `:number` or `:integer`. 
> The `plural` and `ordinal` selection methods not only affect how number _operands_ are matched, 
> but also influence the translation of the _message_,
> such as the expansion or contraction of the list of _keys_.
> Such adjustment cannot be sensibly done if the _selector_ is chosen at runtime.
> The _function_, therefore, does not permit the use of a _variable_ as the value
> of the _option_ `select`.
```

aphillips (commented):
See if this suggestion solves the problems with the algorithm.

spec/formatting.md
eemeli: This proposed language would make things more confusing. Note that _literal_ and _variable_ are defined in the syntax, and that e.g. Variable Resolution will never make a _variable_ be a _literal_.

An extended explanation like the note here would probably be better placed in Number Selection, where currently we only say "the _message_ might not be translatable".

aphillips: Okay, never mind.

aphillips (approved):

spec/formatting.md
macchiati: I think Eemeli's wording is fine.
Moreover, this is something that can be refined after review by the CLDR TC — which is to happen tomorrow. So let's just merge this.

macchiati (approved):

aphillips:
I am merging this now. @gibson042 I will open an issue to track improving this text.

PR #1023, opened 2025-02-18
Fix term link errors

aphillips:
Fix errors in term highlighting. The missing terms are:

- annotation (not fixed, no longer used formally in the spec)
- bad-option-error (fixed here)
- error  (not fixed)
- expression-resolution (fixed here)
- field-option (fixed here)
- formatter (not fixed)
- function-resolution (fixed here)
- literal-resolution (fixed here)
- namespace (fixed here)
- option-value (not fixed)
- style-option (fixed here)
- variable-resolution (fixed here)
- whitespace (not fixed)

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -20,7 +20,7 @@ The formatting of a _message_ is defined by the following operations:
   Depending on the implementation, this result could be a single concatenated string,
   an array of objects, an attributed string, or some other locally appropriate data type.
 
-- **_<dfn>Expression Resolution</dfn>_** determines the value of an _expression_ or _markup_,
+- **_<dfn>Expression and Markup Resolution</dfn>_** determines the value of an _expression_ or _markup_,
   with reference to the current _formatting context_.
   This can include multiple steps,
   such as looking up the value of a variable and calling formatting functions.
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -20,7 +20,7 @@ The formatting of a _message_ is defined by the following operations:
   Depending on the implementation, this result could be a single concatenated string,
   an array of objects, an attributed string, or some other locally appropriate data type.
 
-- **_<dfn>Expression and Markup Resolution</dfn>_** determines the value of an _expression_ or _markup_,
+- **_<dfn>Expression Resolution</dfn>_** determines the value of an _expression_ or _markup_,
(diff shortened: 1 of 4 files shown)


spec/formatting.md
```diff
@@ -20,7 +20,7 @@ The formatting of a _message_ is defined by the following operations:
   Depending on the implementation, this result could be a single concatenated string,
   an array of objects, an attributed string, or some other locally appropriate data type.
 
-- **_<dfn>Expression and Markup Resolution</dfn>_** determines the value of an _expression_ or _markup_,
+- **_<dfn>Expression Resolution</dfn>_** determines the value of an _expression_ or _markup_,
```
eemeli, line 23: I don't think it's right to say that "Expression Resolution determines the value of [...] markup".

aphillips: I didn't change that part of the sentence ;-)

The problem here is that the term as built is "expression and markup resolution", but we refer to it literally _everywhere_ as "expression resolution". I could do:

```suggestion
**_Expression and Markup Resolution_** ("<dfn>expression resolution</dfn>")
determines the value of an _expression_ or _markup_,
```

eemeli: I would prefer keeping this as is, and defining _**Expression Resolution**_ immediately before line 191, like this:

```md
### Expression and Markup Resolution

_Expressions_ are used in _declarations_ and _patterns_.
_Markup_ is only used in _patterns_.

#### Expression Resolution

**_<dfn>Expression Resolution</dfn>_** determines the value of an _expression_.
Depending on the presence or absence of a _variable_ or _literal_ operand and a _function_,
the _resolved value_ of the _expression_ is determined as follows:
```

The subsequent text there is specifically about expressions, and we already have a Markup Resolution subsection further down in the Expression and Markup Resolution section.

aphillips: That's a better solution. I'm trying to make minimal changes here, but your proposal is a non-normative change, so works for me.

eemeli:
```suggestion
- **_<dfn>Expression and Markup Resolution</dfn>_** determines the value of an _expression_ or _markup_,
```

eemeli (approved):
Approving, provided that the suggestion restoring "and Markup" is accepted.

PR #1025, opened 2025-02-20
Allow valid message invalidation with new default functions

aphillips:
Fixes (part of) ~~#1024.~~ #1033 

@macchiati Demonstrated that the stability policy could logically conflict with locally-assigned default-namespace functions or options. This fixes the policy to allow us to assign **_new_** values that conflict with such.

I believe the change is consistent with working group consensus.

### DIFF
--- a/spec/intro.md
+++ b/spec/intro.md
@@ -78,17 +78,8 @@ A reference to a _term_ looks like this.
 
 ### Stability Policy
 
-Updates to this specification will not make any valid _message_ invalid,
-with the exception that such updates can introduce new _default functions_
-or add new _options_ or additional values to the existing _options_ of existing _default functions_.
-
-> [!NOTE]
-> The exception in the above policy would only make an existing valid _message_ invalid
-> if the existing _message_ has used an implementation or user-defined _function_,
-> _option_, or _option_ value that conflicts or whose values conflict with the new
-> _default function_ assignment.
-> Implementers and users are strongly cautioned to namespace all custom _functions_
-> or the addition of _options_ to _default functions_.
+Updates to this specification will not result in a _valid_ _message_ emitting a _Syntax Error_
+or _Data Model Error_.
 
 Updates to this specification will not specify an _error_ for any _message_
 that previously did not specify an _error_.
--- a/spec/intro.md
+++ b/spec/intro.md
@@ -78,7 +78,17 @@ A reference to a _term_ looks like this.
 
 ### Stability Policy
 
-Updates to this specification will not make any valid _message_ invalid.
+Updates to this specification will not make any valid _message_ invalid,
+with the exception that such updates can introduce new _default functions_
+or add new _options_ or additional values to the existing _options_ of existing _default functions_.
+
+> [!NOTE]
+> The exception in the above policy would only make an existing valid _message_ invalid
+> if the existing _message_ has used an implementation or user-defined _function_,
+> _option_, or _option_ value that conflicts or whose values conflict with the new
+> _default function_ assignment.
+> Implementers and users are strongly cautioned to namespace all custom _functions_
+> or the addition of _options_ to _default functions_.
 
 Updates to this specification will not specify an _error_ for any _message_
 that previously did not specify an _error_.
--- a/spec/intro.md
+++ b/spec/intro.md
@@ -78,8 +78,7 @@ A reference to a _term_ looks like this.
 
 ### Stability Policy
 
-Updates to this specification will not result in a _valid_ _message_ emitting a _Syntax Error_
-or _Data Model Error_.
+Updates to this specification will not make any _valid_ _message_ become not _valid_.
 
 Updates to this specification will not specify an _error_ for any _message_
 that previously did not specify an _error_.


gibson042 (approved):

spec/intro.md
```diff
@@ -78,7 +78,17 @@ A reference to a _term_ looks like this.
...
+with the exception that such updates can introduce new _default functions_
+or add new _options_ or additional values to the existing _options_ of existing _default functions_.
+
+> [!NOTE]
+> The exception in the above policy would only make an existing valid _message_ invalid
+> if the existing _message_ has used an implementation or user-defined _function_,
+> _option_, or _option_ value that conflicts or whose values conflict with the new
+> _default function_ assignment.
+> Implementers and users are strongly cautioned to namespace all custom _functions_
+> or the addition of _options_ to _default functions_.
```
macchiati, line 91: For completeness and clarity, here's what I think we need to do.

### OLD
Updates to this specification will only reserve, define, or require function identifiers and function option identifiers which satisfy either of the following two requirements:

- Includes no namespace, and has a name consisting of characters in the ranges a-z, A-Z, and 0-9, and the characters U+002E FULL STOP ., U+002D HYPHEN-MINUS -, and U+005F LOW LINE _.
- Uses a namespace consisting of a single character in the ranges a-z and A-Z.

All other identifiers in these categories are reserved for the use of implementations or users.

### NEW

_Reserved identifiers_ are those that satisfy at least one of the following conditions:

- Includes no namespace, and has a name consisting of characters in the ranges a-z, A-Z, and 0-9, and the characters U+002E FULL STOP ., U+002D HYPHEN-MINUS -, and U+005F LOW LINE _.
- Uses a namespace consisting of a single character in the ranges a-z and A-Z.

Updates to this specification will only reserve, define, or require _reserved identifiers_ . All identifiers other than reserved identifiers are available for the use of implementations or users.

All other identifiers in these categories are reserved for the use of implementations or users.

(I would have made the above a suggestion at the right place, but github only lets me do that on lines you've changed.)

```suggestion
Updates to this specification will not make any valid _message_ invalid,
with the exception that such updates can introduce new _default functions_
or add new _options_ or additional values to the existing _options_ of existing _default functions_.

> [!NOTE]
> The exception in the above policy would only make an existing valid _message_ invalid
> if the existing _message_ has used reserved identifiers.
> Implementers and users are strongly advised to namespace all identifiers used in custom _functions_,
> _options_ to _default functions_, _attributes_, and _markup_.

.
```

macchiati (changes requested):

macchiati:
I retired the issue that had multiple sub-issues, so the description should reference https://github.com/unicode-org/message-format-wg/issues/1033

spec/intro.md
aphillips: @macchiati 

I don't think your version makes sense? Its words can be interpreted as "if you use a _reserved identifier_ (which includes the default function values such as `:number` or number's option `select`) in your _message_ we are permitted to break you".

Actually, I think this whole change is wrong. Here's what _valid_ means:

> A message is valid if it is well-formed and also meets the additional content restrictions and semantic requirements about its structure defined below for declarations, matcher, and options. Attempting to parse a message that is not valid will result in a Data Model Error.

The _matcher_ validity requirements are:

> A message can only be considered valid if the following requirements are satisfied; otherwise, a corresponding Data Model Error will be produced during processing:
> - Variant Key Mismatch: The number of keys on each variant MUST be equal to the number of selectors.
> - Missing Fallback Variant: At least one variant MUST exist whose keys are all equal to the "catch-all" key *.
> - Missing Selector Annotation: Each selector MUST be a variable that directly or indirectly references a declaration with a function.
> - Duplicate Variant: Each variant MUST use a list of keys that is unique from that of all other variants in the message. Literal keys are compared by their contents, not their syntactical appearance.

The _options_ validity requirement is:
> Each option's identifier MUST be unique within the function: a function with duplicate option identifiers is not valid and will produce a Duplicate Option Name error during processing.

(declarations isn't important here)

You are thinking of a different level of "valid" that we don't currently have in the spec--one in which we validate option values.

What we should do is link _valid_ and clarify what the real promise is:

```suggestion
Updates to this specification will not result in a _valid_ _message_ emitting a _Syntax Error_
or _Data Model Error_.
```

We can handle _default function_ stability separately (as we already do). We can also, using the above formulation, strengthen _valid_ in the future to encompass what you've proposed for validation of options/values, because those cannot result in syntax or data model errors (those are bad option or unsupported operation errors)

aphillips:
I merged my suggestion (you can uncollapse @macchiati's comment to see the discussion). Please comment if you want the note back or some other wording.

macchiati (approved):

macchiati:
The change is good, but just to note that it doesn't address https://github.com/unicode-org/message-format-wg/issues/1033

spec/intro.md
macchiati: Our stability policy is pretty vacuous if we don't cover custom functions, custom options for standard functions, custom attributes, and custom markup.

Consider another case: An implementation creates a custom function `:unit`, which formats a currency-unit; not a number, just the unit. It has an option second that takes a second currency. An old message has `{USD :unit}` and formats it (in English) as "US Dollar".

That message is going to fail spectacularly when* we add `:unit` in v48 that takes a operand like `|3 meter|`, and barfs on an operand like `USD`. It will return bad operand errors, whereas it used to work fine.

In no way can we consider that to be good policy for backwards compatibility, _which is the goal for stability_. 

If they had defined `amzn:datetime`, they would be safe. The same would go for `amzn:img` markup and `@amzn:translate`. But because we didn't warn them very clearly about the need for that, they didn't use a namespace.

My preference was to go the clearer 'forbid' route: to tell people that only the spec can define identifiers that match reserved identifiers (with 1-letter namespace or limited to [a-zA-Z0-9.\-_]). Custom functions (and custom options for standard functions), markup, and attributes can't define identifiers that collide. That gives people a very clear bright line to follow.

You suggested a weaker constraint; that we don't forbid people from defining custom functions, markup, and/or attributes that use reserved identifiers, but instead warn them that under our stability policies future versions can them if they do, so they are best off never doing that — to avoid that possibility they can use identifiers that can't collide. That would also work, so that's what I put together.

spec/intro.md
```diff
@@ -78,7 +78,8 @@ A reference to a _term_ looks like this.
 
 ### Stability Policy
 
-Updates to this specification will not make any valid _message_ invalid.
+Updates to this specification will not result in a _valid_ _message_ emitting a _Syntax Error_
+or _Data Model Error_.
```
eemeli, line 82: I don't understand the change as it's currently proposed. We have a fine [definition](https://github.com/unicode-org/message-format-wg/blob/main/spec/syntax.md#well-formed-vs-valid-messages) of _valid_ in the spec, and we should not effectively re-define it here. If the issue is about linkifying the term and "invalid" not being defined, that we can fix:

```suggestion
Updates to this specification will not make any _valid_ _message_ become not _valid_.
```

eemeli (changes requested):

spec/intro.md
aphillips: 💯 I agree with your statement, but you have to read the previous thread. @macchiati proposed elsewhere (#935) to add to the definition of "valid" in a future release (e.g. v48), to include option validation as a consideration for "valid"... and we've discussed this a lot recently. His issue #1033 points out that new function/option assignments could result in a _valid_ message failing. You and I agree that "valid" has a specific meaning (syntax or DM error)

aphillips: @macchiati said:
> The change is good, but just to note that it doesn't address https://github.com/unicode-org/message-format-wg/issues/1033

I think it **does** address part of #1033. That issue is about the stability policy. We have multiple things going on there. One is about which identifiers we'll assign or not assign--we are not _materially_ changing that part of the policy, btw. The other (which is PR deals with) is what we are permitted to do in future specs.

The error introduced by creating a new reserved identifier (of whatever sort) that conflicts with an existing message's use of said identifier is not a syntax error (it can never be) and is not a data model error (the data model can represent lots of stuff). Such a message might start to emit a Message Function Error or (squinting) maybe a Bad Selector (a Resolution Error).

You created #1033 because (I assume) your thinking about #935 (as cited above). This PR doesn't completely fix #1033, I guess, except that, from the point of view of the stability policy, it does the job.

aphillips:
I merged @eemeli's text. I think that's an improvement. If we needed to, we could add a note calling out that a _valid_ _message_ can start to have errors if users or implementers have abused a reserved identifier--that should go in the stability policy section about identifiers.

eemeli (approved):
I think it's important for us to have a term like the current _valid_ that a _messsage_ can be verified to pass with no reference to anything outside the _message_, including functions.

I would not be opposed to defining a name for a message that we know is going to produce an error during formatting because it has operand or option values that don't pass muster, but I'd also like to note that that check is in most cases not going to prove that the message will not produce an error, because formatting relies on external values that are not necessarily known ahead of time.

aphillips:
Merging this. Other changes to follow for the name collision parts of #1033

PR #1027, opened 2025-02-20
Clarify option literalness details

### LINKED ISSUE #1026: Option resolution does not correctly handle transitivity
As described at https://github.com/unicode-org/message-format-wg/pull/1019#discussion_r1958859549 .

> ```
> .input {$x :number}
> .local $a = {|EUR|}
> .local $b = {$a}
> {{Balance: {$x :currency currency=$b}.}}
> ```
> 
> 1. Per [Literal Resolution](https://github.com/unicode-org/message-format-wg/blob/9c9ecfcc292fe8c0fbf3973b4e5755c69ff08ff7/spec/formatting.md#literal-resolution), the resolved value of $a is "EUR" (and per the new [«the <em>resolved value</em> of each <em>option</em> value MUST include information about whether the <em>option</em> value is a <em>literal</em> or a <em>variable</em>.»] text in [Resolved Values](https://github.com/unicode-org/message-format-wg/blob/9c9ecfcc292fe8c0fbf3973b4e5755c69ff08ff7/spec/formatting.md#resolved-values), includes its literal nature).
> 2. Per [Variable Resolution](https://github.com/unicode-org/message-format-wg/blob/9c9ecfcc292fe8c0fbf3973b4e5755c69ff08ff7/spec/formatting.md#variable-resolution), the resolved value of $b is just the resolved value of $a (i.e., a literal with contents "EUR")—although I cannot determine whether or not implementations are given freedom to drop that propagation by «An implementation MAY perform additional processing when resolving the value of an _expression_ that consists only of a _variable_».
> 3. These new steps in [Option Resolution](https://github.com/unicode-org/message-format-wg/blob/9c9ecfcc292fe8c0fbf3973b4e5755c69ff08ff7/spec/formatting.md#option-resolution) don't _add_ literalness to the resolved value of $b, but it's already there so the `:currency` function will still see it.
> 
> In fact, it seems like the new steps here cannot ever have an effect, because the resolved value of a literal option is marked as such in [Literal Resolution](https://github.com/unicode-org/message-format-wg/blob/9c9ecfcc292fe8c0fbf3973b4e5755c69ff08ff7/spec/formatting.md#literal-resolution) and the resolved value of a variable referencing another variable is just the resolved value of its target. So rather than changing this algorithm, the PR should instead accompany it with note(s) explaining the above.
> 
> However, more normative text _is_ required is for explaining a variant where `.local $b = {$a}` is replaced by `.local $b = {$a :string}`, in which it is not currently clear whether or not [Function Resolution](https://github.com/unicode-org/message-format-wg/blob/9c9ecfcc292fe8c0fbf3973b4e5755c69ff08ff7/spec/formatting.md#function-resolution) grants implementations the freedom to return a result that is marked as a literal. I strongly believe that such freedom should **not** exist, and that literalness be preserved **only** by direct isolated variable references.

gibson042:
Fixes #1026

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -137,10 +137,17 @@ as well as a flag to indicate whether
 its formatted representation requires isolation
 from the surrounding text.
 
-To allow for _function handlers_ to ensure that certain _option_ values are set by _literals_,
-the _resolved value_ of each _option_ value MUST include information about 
-whether the _option_ value is a _literal_ or a _variable_.
-Note that this information is irrelevant for a _resolved value_ that is not used as the value of an _option_.
+For each _option_ value, the _resolved value_ MUST indicate if the value
+was directly set with a _literal_, as opposed to being resolved from a _variable_.
+This is to allow _functions handlers_ to require specific _options_ to be set using _literals_.
+
+> For example, the _default functions_ `:number` and `:integer` require that the _option_
+> `select` be set with a _literal_ (`plural`, `ordinal`, or `exact`). 
+> Allowing a _variable_ in this _option_ would produce a _message_ that 
+> is difficult to translate because the set of _keys_ is tied to the _selector_ chosen.
+
+> [!NOTE]
+> Such information is irrelevant for _resolved values_ not used as the value of an _option_.
 
 The form that _resolved values_ take is implementation-dependent,
 and different implementations MAY choose to perform different levels of resolution.
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -144,7 +144,7 @@ This is to allow _functions handlers_ to require specific _options_ to be set us
 > For example, the _default functions_ `:number` and `:integer` require that the _option_
 > `select` be set with a _literal_ (`plural`, `ordinal`, or `exact`). 
 > Allowing a _variable_ in this _option_ would produce a _message_ that 
-> is difficult to translate because the set of _keys_ is tied to the _selector_ chosen.
+> is impossible to translate because the set of _keys_ is tied to the _selector_ chosen.
 
 > [!NOTE]
 > Such information is irrelevant for _resolved values_ not used as the value of an _option_.
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -140,6 +140,7 @@ from the surrounding text.
 To allow for _function handlers_ to ensure that certain _option_ values are set by _literals_,
 the _resolved value_ of each _option_ value MUST include information about 
 whether the _option_ value is a _literal_ or a _variable_.
+Note that this information is irrelevant for a _resolved value_ that is not used as the value of an _option_.
 
 The form that _resolved values_ take is implementation-dependent,
 and different implementations MAY choose to perform different levels of resolution.
@@ -329,6 +330,7 @@ the following steps are taken:
 
 6. If the call succeeds,
    resolve the value of the _expression_ as the result of that function call.
+   The value MUST NOT be marked as a literal option value.
 
    If the call fails or does not return a valid value,
    emit the appropriate _Message Function Error_ for the failure.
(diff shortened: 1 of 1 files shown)


spec/formatting.md
```diff
@@ -395,7 +397,7 @@ For each _option_:
       1. If supported, emit a _Bad Option_ error.
    1. Else:
       1. If the _option_ value consists of a _literal_:
-         1. Include that information in `rv`.
+         1. Mark `rv` as a literal option value.
```
aphillips, line 400: I like this wording change very much

eemeli: This change does seem good.

spec/formatting.md
```diff
@@ -140,6 +140,7 @@ from the surrounding text.
 To allow for _function handlers_ to ensure that certain _option_ values are set by _literals_,
 the _resolved value_ of each _option_ value MUST include information about 
 whether the _option_ value is a _literal_ or a _variable_.
+Note that this information is irrelevant for a _resolved value_ that is not used as the value of an _option_.
```
eemeli, line 143: Hmm. Is it perhaps not clear that "_option_ value" is a syntax feature? This phrase seems to imply that a _resolved value_ could be an _option_ value, and that doesn't really make sense.

We probably ought to explicitly define _option value_ as a term in the Syntax section.

macchiati: > We probably ought to explicitly define option value as a term in the Syntax section.

Absolutely!

aphillips: Let's do that in a separate PR.

aphillips:
Does this need to be in 47? It looks 47-ish, as I believe it is non-normative and small. Changing to _option value_ as a term is also a candidate.

spec/formatting.md
aphillips, line 143: I would do this differently?

```suggestion
For each _option_ value, the _resolved value_ MUST indicate if the value
was directly set with a _literal_, as opposed to being resolved from a _variable_.
This is to allow _functions handlers_ to require specific _options_ to be set using _literals_.

> For example, the _default functions_ `:number` and `:integer` require that the _option_
> `select` be set with a _literal_ (`plural`, `ordinal`, or `exact`). 
> Allowing a _variable_ in this _option_ would produce a _message_ that 
> is difficult to translate because the set of _keys_ is tied to the _selector_ chosen.

> [!NOTE]
> Such information is irrelevant for _resolved values_ not used as the value of an _option_.
```

macchiati:
Non-normative changes: typos, clarifications, even new definitions, etc we can do; we just need to notify the TC.

Normative changes that are important need approval by the TC.

aphillips (approved):
I am okay with the wording proposed, pace we work in _option value_. I made a wording suggestion, however, that might improve the paragraph?

spec/formatting.md
macchiati: > is difficult to translate because the set of _keys_ is tied to the _selector_ chosen.
=>
is impossible to translate because the set of _keys_ is tied to the _selector_ chosen.

gibson042: @aphillips Accepted suggestion.

spec/formatting.md
```diff
@@ -137,9 +137,17 @@ as well as a flag to indicate whether
...
-the _resolved value_ of each _option_ value MUST include information about 
-whether the _option_ value is a _literal_ or a _variable_.
+For each _option_ value, the _resolved value_ MUST indicate if the value
+was directly set with a _literal_, as opposed to being resolved from a _variable_.
+This is to allow _functions handlers_ to require specific _options_ to be set using _literals_.
+
+> For example, the _default functions_ `:number` and `:integer` require that the _option_
+> `select` be set with a _literal_ (`plural`, `ordinal`, or `exact`). 
+> Allowing a _variable_ in this _option_ would produce a _message_ that 
+> is difficult to translate because the set of _keys_ is tied to the _selector_ chosen.
```
macchiati, line 147:
```suggestion
> is impossible to translate because the set of _keys_ is tied to the _selector_ chosen.
```

macchiati (approved):
I suggest one further change, but not a blocker.

macchiati (approved):

PR #1038, opened 2025-02-21
Define reserved identifiers

### LINKED ISSUE #1033: [FEEDBACK] Stability policies conflict with allowing all identifiers
Note: for functions and function option identifiers, we sort-of half-reserve certain identifiers: 

> Updates to this specification will only reserve, define, or require function identifiers and function option identifiers which satisfy either of the following two requirements:
> 
> - Includes no namespace, and has a name consisting of characters in the ranges a-z, A-Z, and 0-9, and the characters U+002E FULL STOP ., U+002D HYPHEN-MINUS -, and U+005F LOW LINE _.
> - Uses a namespace consisting of a single character in the ranges a-z and A-Z.
> 
> All other identifiers in these categories are reserved for the use of implementations or users.

_For the purposes of this discussion, let's call identifiers satisfying either one of the bullets, **X identifiers**._

The reason I say "half-reserve" is that we only say that LDML **won't** define identifiers that are **not** X identifiers. There is nothing that says that an implementation can't use X identifiers. And we don't reserve markup X identifiers or attribute X identifiers in any way.

And because of the 'half-reservation', we can't live up to our stability policy that "Updates to this specification will not make any valid message invalid.", as the following example shows.

Suppose after release:

1. Implementation X defines its own function :datetime, with option skeleton=abc.
2. They also define a literal operand of :datetime to allow the form 31-1-2026 (meaning Jan 1, 2026).
3. So `{31-1-2026 :datetime skeleton=abc}` is valid.
4. We don't know that X does this (since we won't know of all the possible implementations).
5. In v48 we define :datetime so that it takes ISO 8601-format literals, and skeleton options that preclude 'abc'. 
6. That makes X's `{31-1-2026 :datetime skeleton=abc}` be invalid

I think here are the options:
1. We amend the spec to reserve all X identifiers for the specification, into the future, and also cover markup and attributes. Implementations must not support them unless they are specified in some version of LDML<br>OR
2. We amend the stability policy to add an exception clause, such as:> Updates to this specification will not make any valid message invalid, unless if they use  X identifiers.

I think Option 1 is preferable.

eemeli:
Fixes #1033 with the text proposed there by @macchiati. The changes to the making-messages-invalid line proposed in #1025 are left out.

### DIFF
--- a/spec/intro.md
+++ b/spec/intro.md
@@ -110,20 +110,19 @@ defined for _default functions_.
 > differences in implementation or changes to locale data
 > (such as due to the release of new CLDR versions).
 
-Updates to this specification will only reserve, define, or require
-_function_ _identifiers_ and _function_ _option_ _identifiers_
-which satisfy either of the following two requirements:
+**_<dfn>Reserved identifiers</dfn>_** are those that satisfy at least one of the following conditions:
+
 - Includes no _namespace_,
   and has a _name_ consisting of characters in the ranges a-z, A-Z, and 0-9,
   and the characters U+002E FULL STOP `.`, U+002D HYPHEN-MINUS `-`, and U+005F LOW LINE `_`.
 - Uses a _namespace_ consisting of a single character in the ranges a-z and A-Z.
 
-All other _identifiers_ in these categories are reserved for the use of implementations or users.
+Updates to this specification will only reserve, define, or require _reserved identifiers_.
+All _identifiers_ other than _reserved identifiers_ are available for the use of implementations or users.
 
-> [!IMPORTANT]
-> Implementation-defined or user-defined _functions_ and _function_ _options_
-> SHOULD use a _namespace_ as part of their _identifiers_
-> to help avoid collisions with other implementations.
+> [!NOTE]
+> Implementers and users are strongly advised to namespace all _identifiers_ used in custom _functions_,
+> _options_ to _default functions_, _attributes_, and _markup_.
 
 Future versions of this specification will not introduce changes
 to the data model that would result in a data model representation


macchiati (approved):
I don't think this completely solves problem, but this is big step in the right direction.

spec/intro.md
```diff
@@ -110,20 +110,19 @@ defined for _default functions_.
...
+Updates to this specification will only reserve, define, or require _reserved identifiers_.
+All _identifiers_ other than _reserved identifiers_ are available for the use of implementations or users.
 
-> [!IMPORTANT]
-> Implementation-defined or user-defined _functions_ and _function_ _options_
-> SHOULD use a _namespace_ as part of their _identifiers_
-> to help avoid collisions with other implementations.
+> [!NOTE]
+> Implementers and users are strongly advised to namespace all _identifiers_ used in custom _functions_,
+> _options_ to _default functions_, _attributes_, and _markup_.
```
aphillips, line 125: This is a normative change, since it removes a normative SHOULD (albeit to function authors and which cannot be tested). Why not keep the normative level intact? We can do a wording change, if desired.

I'm not sure about "markup", since we don't define markup.

We could make it a bit tighter:

```suggestion
> [!NOTE]
> Use a namespace for all _identifiers_ defined by an implementation
> or in the _names_ of custom _functions_, _attributes_, or _markup_ defined by users.
```

spec/intro.md
```diff
@@ -110,20 +110,19 @@ defined for _default functions_.
...
-All other _identifiers_ in these categories are reserved for the use of implementations or users.
+Updates to this specification will only reserve, define, or require _reserved identifiers_.
+All _identifiers_ other than _reserved identifiers_ are available for the use of implementations or users.
 
-> [!IMPORTANT]
-> Implementation-defined or user-defined _functions_ and _function_ _options_
-> SHOULD use a _namespace_ as part of their _identifiers_
-> to help avoid collisions with other implementations.
+> [!NOTE]
+> Implementers and users are strongly advised to namespace all _identifiers_ used in custom _functions_,
```
macchiati, line 124:
```suggestion
> Implementers and users are strongly advised to only use _custom identifiers_ in custom _functions_,
```

spec/intro.md
macchiati: The fuzziness of 'implementation' and 'user' gets in the way here. I read 'custom' as 'not defined by this spec. I read 'users' as 'users of an implementation', including that implementation's custom functions, etc.

spec/intro.md
```diff
@@ -110,20 +110,19 @@ defined for _default functions_.
...
 
-Updates to this specification will only reserve, define, or require
-_function_ _identifiers_ and _function_ _option_ _identifiers_
-which satisfy either of the following two requirements:
+**_<dfn>Reserved identifiers</dfn>_** are those that satisfy at least one of the following conditions:
+
 - Includes no _namespace_,
   and has a _name_ consisting of characters in the ranges a-z, A-Z, and 0-9,
   and the characters U+002E FULL STOP `.`, U+002D HYPHEN-MINUS `-`, and U+005F LOW LINE `_`.
 - Uses a _namespace_ consisting of a single character in the ranges a-z and A-Z.
```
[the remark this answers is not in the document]
aphillips: This stuff seems to belong better in syntax.md than in the body of the stability policy. We have a whole section about names and identifiers there and that's where most readers would expect to find definitions like these. Then the stability policy can focus on it's job of forbidding/permitting certain things _found in the specification_

spec/intro.md
eemeli: Given that the SHOULD is effectively untestable, I don't think changing it to "strongly advised" is that significant a change.

I'd be fine with leaving un-namespaced markup as explicitly not reserved.

spec/intro.md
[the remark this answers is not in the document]
eemeli: I agree with Addison. Defining _reserved identifier_ here makes sense, but going beyond that would be a bit much.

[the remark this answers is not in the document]
macchiati: I agree, in the sense that "Includes no _namespace_,
  and has a _name_ consisting of characters in the ranges a-z, A-Z, and 0-9,
  and the characters U+002E FULL STOP `.`, U+002D HYPHEN-MINUS `-`, and U+005F LOW LINE `_`." should never have been just in the body; it should have had a syntax definition.

That is, I think we need to define the 'reserved identifiers' and the 'custom identifiers', and ideally they would be in the abnf and not here.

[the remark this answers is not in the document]
eemeli: The `message.abnf` only includes rules that are required for the definition of the `message` rule. We already have some ABNF rules sprinkled elsewhere in the spec:
- `currency_code`
- `number-literal`
- `digit-size-option`
- `integer`
- `u-locale-option`

spec/intro.md
aphillips: This changes the meaning. There are two things going on here:

- Non-default functions should use a namespace. <- this is not normative but is good advice

They can omit the namespace by choosing names that include non-reserved identifier characters too, so separately:

- Non-default functions should use custom identifiers

We also have to be careful to say that we mean for the function **name**.

```
{$foo :custom:function option=|option has a 'reserved' identifier which we want to encourage|}
```

spec/intro.md
macchiati: > I'd be fine with leaving un-namespaced markup as explicitly not reserved.

I disagree: we currently don't define any standard markup, but we could in the future (same with attributes). So the same principles for functions and attributes apply.

macchiati (approved):

aphillips:
Closing this in favor of #1040

PR #1036, opened 2025-02-21
Literal keys contents / values

macchiati:
See https://github.com/unicode-org/message-format-wg/issues/1028

PR #1039, opened 2025-02-21
Define _option value_ and _digit size option_

### LINKED ISSUE #1034: Make _option value_ a term
In #1027 we discussed making _option value_ a term. Making an issue to track that.

eemeli:
Fixes #1034

Mostly the changes are straightforward and completely editorial, though I did opt to make a few further changes as I went:

- Defining also _digit size option_, which we effectively already use as a term
- Some of the explanatory text about `:number select` added in #1027 is moved from Resolved Values to Number Selection
- Harmonize the introductions to each function's options and make sure that each required option is indeed REQUIRED.

I left out a definition of _resolved options_ that I think we probably ought to add as well, but thought that the above could be done as a first step.

### DIFF
--- a/spec/errors.md
+++ b/spec/errors.md
@@ -347,13 +347,13 @@ for that specific _function_.
 #### Bad Option
 
 A **_<dfn>Bad Option</dfn>_** error is an error that occurs when there is
-an implementation-defined error with an _option_ or its value.
+an implementation-defined error with an _option_ or an _option value_.
 These might include:
 - A required _option_ is missing.
 - Mutually exclusive _options_ are supplied.
-- An _option_ value provided to a _function_ during _function resolution_
+- An _option value_ provided to a _function_ during _function resolution_
    does not match one of the implementation-defined types or values for that _function_;
-   or in which the literal _option_ value does not have the required format
+   or in which the _literal_ _option value_ does not have the required format
    and thus cannot be processed into one of the expected
    implementation-defined types for that specific _function_.
 
@@ -386,5 +386,5 @@ does not match the expected implementation-defined format.
 #### Unsupported Operation
 
 A **_<dfn>Unsupported Operation</dfn>_** error is an implementation-specific error
-that occurs when a given _option_, _option_ value, _operand_ value, or some combination
+that occurs when a given _option_, _option value_, _operand_, or some combination
 of these are incompatible or not supported by a given _function_ or its _function handler_.
--- a/spec/errors.md
+++ b/spec/errors.md
@@ -353,7 +353,7 @@ These might include:
 - Mutually exclusive _options_ are supplied.
 - An _option value_ provided to a _function_ during _function resolution_
    does not match one of the implementation-defined types or values for that _function_;
-   or in which the _literal_ _option value_ does not have the required format
+   or in which the _string value_ of an _option_ does not have the required format
    and thus cannot be processed into one of the expected
    implementation-defined types for that specific _function_.
 
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -434,8 +434,8 @@ The resolution of _markup_ MUST always succeed.
 
 A **_<dfn>fallback value</dfn>_** is the _resolved value_ for
 an _expression_ or _variable_ when that _expression_ or _variable_ fails to resolve.
-It contains a string representation that is used for its formatting,
-and no resolved option values.
+It contains a string representation that is used for its formatting.
+All _options_ are removed.
 
 The _resolved value_ of _text_, _literal_, and _markup_ MUST NOT be a _fallback value_.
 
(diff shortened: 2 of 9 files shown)


spec/functions/number.md
```diff
@@ -610,34 +610,35 @@ number-literal = ["-"] (%x30 / (%x31-39 *DIGIT)) ["." 1*DIGIT] [%i"e" ["-" / "+"
...
 #### Digit Size Options
 
-Some _options_ of number _functions_ are defined to take a "digit size option".
+Some _options_ of number _functions_ are defined to take a _digit size option_.
 The _function handlers_ for number _functions_ use these _options_ to control aspects of numeric display
 such as the number of fraction, integer, or significant digits.
 
-A "digit size option" is an _option_ value that the _function_ interprets
+A **_<dfn>digit size option</dfn>_** has an _option value_ that the _function_ interprets
 as a small integer value greater than or equal to zero.
```
aphillips, line 618: This change makes it not a proper term definition. "is" was better here.

```suggestion
A **_<dfn>digit size option</dfn>_** is an _option_ 
whose _option value_ is interpreted by the _function_
as a small integer greater than or equal to zero.
```

spec/u-namespace.md
```diff
@@ -34,11 +34,11 @@ or any other structured formatted results.
 
 Ignored when formatting a message to a string.
 
-The value of the `u:id` _option_ MUST be a _literal_ or a
+The `u:id` _option value_ MUST be a _literal_ or a
 _variable_ whose _resolved value_ is either a string
 or can be resolved to a string without error.
 For other values, a _Bad Option_ error is emitted
-and the `u:id` option is ignored.
+and the `u:id` _option_ and its _option value_ are ignored.
```
aphillips, line 41: There's no need to add option value here, is there?

```suggestion
and the `u:id` _option_ is ignored.
```

aphillips (changes requested):
Great start.

spec/u-namespace.md
eemeli: I was thinking that explicitly ignoring the _option value_ would allow for an implementation to not emit any further errors that might arise during its resolution.

Similar phrasing is used further down in this section, so if you think that we ought to change this, we ought to change all of them.

spec/errors.md
```diff
@@ -347,13 +347,13 @@ for that specific _function_.
...
-an implementation-defined error with an _option_ or its value.
+an implementation-defined error with an _option_ or an _option value_.
 These might include:
 - A required _option_ is missing.
 - Mutually exclusive _options_ are supplied.
-- An _option_ value provided to a _function_ during _function resolution_
+- An _option value_ provided to a _function_ during _function resolution_
    does not match one of the implementation-defined types or values for that _function_;
-   or in which the literal _option_ value does not have the required format
+   or in which the _literal_ _option value_ does not have the required format
```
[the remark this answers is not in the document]
eemeli: Going to accept this with the presumption that #1036 will introduce the _string value_ term.

spec/formatting.md
```diff
@@ -441,7 +435,7 @@ The resolution of _markup_ MUST always succeed.
 A **_<dfn>fallback value</dfn>_** is the _resolved value_ for
 an _expression_ or _variable_ when that _expression_ or _variable_ fails to resolve.
 It contains a string representation that is used for its formatting,
```
eemeli, line 437:
```suggestion
It contains a string representation that is used for its formatting.
```

spec/formatting.md
```diff
@@ -441,7 +435,7 @@ The resolution of _markup_ MUST always succeed.
...
-and no option values.
+and no resolved option values.
```
eemeli, line 438:
```suggestion
All _options_ are removed.
```

spec/u-namespace.md
aphillips: That seems appropriately cautious.

aphillips (approved):

macchiati (approved):

spec/formatting.md
```diff
@@ -137,17 +137,12 @@ as well as a flag to indicate whether
...
 This is to allow _functions handlers_ to require specific _options_ to be set using _literals_.
 
 > For example, the _default functions_ `:number` and `:integer` require that the _option_
-> `select` be set with a _literal_ (`plural`, `ordinal`, or `exact`). 
-> Allowing a _variable_ in this _option_ would produce a _message_ that 
-> is impossible to translate because the set of _keys_ is tied to the _selector_ chosen.
-
-> [!NOTE]
-> Such information is irrelevant for _resolved values_ not used as the value of an _option_.
+> `select` be set with a _literal_ _option value_ (`plural`, `ordinal`, or `exact`).
```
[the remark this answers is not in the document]
aphillips: It's not removed: it's moved to number selection.

aphillips:
Meets fast-track. Merging.

PR #1040, opened 2025-02-21
Define reserved and custom identifiers

aphillips:
Moves these definitions from the stability policy into the section on identifiers in syntax.

### DIFF
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -886,16 +886,19 @@ name-char  = name-start / DIGIT / "-" / "."
 A **_<dfn>reserved identifier</dfn>_** is one that satisfies the following conditions:
 - Includes no _namespace_ or uses a _namespace_ consisting of a single letter
   in the ranges a-z and A-Z.
-- Has a _name_ consisting only of characters in the ranges a-z, A-Z, and 0-9,
-  and the characters U+002E FULL STOP `.`, U+002D HYPHEN-MINUS `-`, and U+005F LOW LINE `_`.
+- Has a _name_ that matches the following ABNF:
+```abnf
+reserved-identifier = ALPHA *[ALPHA / DIGIT / "." / "-" / "_"]
+```
 
 A **_<dfn>custom identifier</dfn>_** is any _identifier_ that is not a _reserved identifier_.
 
-Always use a _namespace_ to identify a _function_ that is not a _default function_
-or when defining a custom _option_ for a _default function_.
-Choose a _custom identifier_ for any _functions_, _markup_, or _attributes_ not defined by this specification.
-
-_Operand_ (variable) _names_ and _option_ _names_ for custom _functions_ are encouraged to use _reserved identifiers_.
+> [!NOTE]
+> Always use a _namespace_ to identify a _function_ that is not a _default function_
+> or when defining a custom _option_ for a _default function_.
+> Choose a _custom identifier_ for any _functions_, _markup_, or _attributes_ not defined by this specification.
+>
+> _Operand_ (variable) _names_ and _option_ _names_ for custom _functions_ are encouraged to use _reserved identifiers_.
 
 The syntax allows a wide range of characters in _names_ and _identifiers_.
 Implementers and authors of _functions_ and _messages_,
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -898,7 +898,7 @@ A **_<dfn>custom identifier</dfn>_** is any _identifier_ that is not a _reserved
 > or when defining a custom _option_ for a _default function_.
 > Choose a _custom identifier_ for any _functions_, _markup_, or _attributes_ not defined by this specification.
 >
-> _Operand_ (variable) _names_ and _option_ _names_ for custom _functions_ are encouraged to use _reserved identifiers_.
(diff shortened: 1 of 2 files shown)


spec/syntax.md
```diff
@@ -883,11 +883,23 @@ name-char  = name-start / DIGIT / "-" / "."
...
+- Includes no _namespace_ or uses a _namespace_ consisting of a single letter
+  in the ranges a-z and A-Z.
+- Has a _name_ consisting only of characters in the ranges a-z, A-Z, and 0-9,
+  and the characters U+002E FULL STOP `.`, U+002D HYPHEN-MINUS `-`, and U+005F LOW LINE `_`.
+
+A **_<dfn>custom identifier</dfn>_** is any _identifier_ that is not a _reserved identifier_.
+
+Always use a _namespace_ to identify a _function_ that is not a _default function_
+or when defining a custom _option_ for a _default function_.
+Choose a _custom identifier_ for any _functions_, _markup_, _options_, or _attributes_ not defined by this specification.
```
macchiati:
```suggestion
Always use a _custom identifier_ for any _functions_, _markup_, _options_, or _attributes_ not defined by this specification.
```
Rationale: the two sentences give somewhat different advice.

macchiati (approved):
Looks good. One additional suggestion, for better wording.

spec/syntax.md
eemeli: The proposed change (as well as the variant suggested above) are in a different tone than the rest of the spec, using an imperative form to instruct the reader directly. I'd rather not introduce such a discontinuity.

aphillips: I make the text into a note. It's user guidance anyway.

@macchiati The two sentences _do_ give different advice. Once signals that namespaces are good specifically for functions and for additions to default functions. The other says to choose custom identifiers for a bunch of stuff.

My comment on the other PR applies here: options in particular do not want "custom identifiers" when defined on user-defined functions.

spec/syntax.md
```diff
@@ -883,11 +883,28 @@ name-char  = name-start / DIGIT / "-" / "."
...
+```
+
+A **_<dfn>custom identifier</dfn>_** is any _identifier_ that is not a _reserved identifier_.
+
+> [!NOTE]
+> Always use a _namespace_ to identify a _function_ that is not a _default function_
+> or when defining a custom _option_ for a _default function_.
+> Choose a _custom identifier_ for any _functions_, _markup_, or _attributes_ not defined by this specification.
+>
+> _Operand_ (variable) _names_ and _option_ _names_ for custom _functions_ are encouraged to use _reserved identifiers_.
```
eemeli, line 901: The "operand" bit here is not really relevant.

```suggestion
> _Variable_ _names_ and _option_ _names_ for custom _functions_ are encouraged to use _reserved identifiers_.
```

spec/syntax.md
```diff
@@ -883,11 +883,28 @@ name-char  = name-start / DIGIT / "-" / "."
...
+
+The syntax allows a wide range of characters in _names_ and _identifiers_.
 Implementers and authors of _functions_ and _messages_,
-including _functions_, _options_, and _operands_ (variable names),
+including _functions_, _options_, and _operands_ (variable names)
```
eemeli, line 905:
```suggestion
including _functions_, _options_, and _variables_,
```

eemeli (approved):
Approving, provided that the inline suggestions below are resolved.

spec/syntax.md
macchiati: I agree with your comment on not needing custom identifiers for options on custom functions.

So it would be:

Always use a _custom identifier_ for any _functions_, _markup_, _options_ on _default functions_, or _attributes_ not defined by this specification.

The point I had about the language you had was that "namespaced identifier" is different, and not in a good way for this paragraph. If you want to replace "Always" by "Choose", that's fine.

| Identifiers | u:abc | abc | walmart.com:abc | π:abc | μικρό |
| :---- | :---- | :---- | :---- | :---- | :---- |
| *Reserved* | ✅ | ✅ | ❌ | ❌ | ❌ |
| *Custom* | ❌ | ❌ | ✅ | ✅ | ✅ |
| *Namespaced* | ✅ | ❌ | ✅ | ✅ | ❌ |

aphillips: So `μικρό` is a custom identifier, and, while that's sufficient for avoiding conflicts with future spec additions, it ignores the fact that others might choose to create their own `μικρό` function (or whatever). We can move namespacing after custom identifier.

aphillips: Check the change I made in response?

macchiati (approved):

spec/syntax.md
macchiati: Yes, that works.

For the future, here are my thoughts.

- Using a custom identifier protects against collisions in future versions of our spec.
- Using a custom identifier with a namespace helps protect against collisions with *other implementers*.

I say "helps", because it is still possible for Oracle Medical Supplies to define :oracle:date and for Oracle (database company) to define :oracle:date.

- We could drastically reduce the opportunity for collisions with various strategies for uniqueness (or probably uniqueness), but we'd have to consider the likelihood and priority of something like that.

PR #1044, opened 2025-02-26
Clarify meaning of `fails` option for test functions

catamorphism:
[The spec for `:test:function` says](https://github.com/unicode-org/message-format-wg/blob/main/test/README.md#test-functions):

"If `FailsFormat` is true, attempting to format the _placeholder_ to any formatting target will fail."

This doesn't say anything about what the output should be for `:test:function fails=format` with any operand. However, the first two tests in fallback.json impose requirements on the output. This is stronger than what the spec requires. Hence I suggest removing these tests.

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -126,7 +126,8 @@ the behaviour of calling it as the `rv` value of MatchSelectorKeys(`rv`, `keys`)
 depends on its `Input`, `DecimalPlaces` and `FailsSelect` values.
 
 - If `FailsSelect` is `true`,
-  calling the method will fail and not return any value.
+  calling the method will emit a _Message Function Error_
+  and not return any value.
 - If the `Input` is 1 and `DecimalPlaces` is 1,
   the method will return some slice of the list « `'1.0'`, `'1'` »,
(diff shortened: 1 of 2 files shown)


eemeli (changes requested):
This output (using the fallback value) is defined by the second paragraph of [this section](https://github.com/unicode-org/message-format-wg/blob/main/spec/formatting.md#formatting-of-the-selected-pattern). If that's not clear, the test function description ought to be improved rather than removing the tests.

catamorphism:
> This output (using the fallback value) is defined by the second paragraph of [this section](https://github.com/unicode-org/message-format-wg/blob/main/spec/formatting.md#formatting-of-the-selected-pattern). If that's not clear, the test function description ought to be improved rather than removing the tests.

I see. I was confused because it wasn't clear to me what "fails" meant in the descriptions of what `fails=format` and `fails=select` do. Instead of changing the tests, I repurposed the PR to change the test README to make that clearer.

eemeli (approved):

aphillips (approved):
Good documentation fix!

PR #1046, opened 2025-02-28
Tests: fix 'literal select' tests for :number and :integer

catamorphism:
Unless I'm misunderstanding something, `literal select {1 :integer select=exact}` should format to `literal select 1` (with no fallback output), since `exact` is a literal.

### DIFF
--- a/test/tests/functions/integer.json
+++ b/test/tests/functions/integer.json
@@ -64,7 +64,7 @@
     {
       "src": ".local $sel = {1 :integer select=$bad} .match $sel 1 {{ONE}} * {{variable select {$sel}}}",
       "params": [{ "name": "bad", "value": "exact" }],
-      "exp": "variable select {|1|}",
+      "exp": "variable select {$sel}",
       "expErrors": [{ "type": "bad-option" }, { "type": "bad-selector" }]
     }
   ]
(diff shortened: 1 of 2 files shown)


test/tests/functions/integer.json
```diff
@@ -54,7 +54,7 @@
     },
     {
       "src": ".local $sel = {1 :integer select=exact} .match $sel 1 {{literal select {$sel}}} * {{OTHER}}",
-      "exp": "literal select {1}"
+      "exp": "literal select 1"
```
aphillips, line 57: This is right, although don't we try to avoid depending on the specific output of a formatter? Perhaps make the pattern `{{literal select 1}}`

aphillips (commented):
Looks good.

test/tests/functions/integer.json
eemeli: No, we need to validate that the formatting of `$sel` emits no error here, unlike `$bad` in the next two tests.

eemeli (approved):
Thank you for fixing these! Not sure what I was thinking; they're almost certainly copy-paste errors.

catamorphism:
Added one more commit to change `{1}` to `{|1|}` (etc.) in the expected output.

PR #1048, opened 2025-03-01
Fix select tests to not presume fallback for formatting

eemeli:
The changes in #1046 weren't quite enough, as they don't account for this part of the spec: https://github.com/unicode-org/message-format-wg/blob/662076e128887cad9caa5502b4d4601c5c9bb231/spec/functions/number.md?plain=1#L645

That means that the formatting of a placeholder with `select=$bad` does not experience fallback, even if an error is emitted.

### DIFF
--- a/test/tests/functions/integer.json
+++ b/test/tests/functions/integer.json
@@ -43,13 +43,13 @@
     },
     {
       "src": ".local $bad = {exact} {{variable select {1 :integer select=$bad}}}",
-      "exp": "variable select {|1|}",
+      "exp": "variable select 1",
       "expErrors": [{ "type": "bad-option" }]
     },
     {
       "src": "variable select {1 :integer select=$bad}",
       "params": [{ "name": "bad", "value": "exact" }],
-      "exp": "variable select {|1|}",
+      "exp": "variable select 1",
       "expErrors": [{ "type": "bad-option" }]
     },
     {
@@ -58,13 +58,13 @@
     },
     {
       "src": ".local $sel = {1 :integer select=exact} .local $bad = {$sel :integer} .match $bad 1 {{ONE}} * {{operand select {$bad}}}",
-      "exp": "operand select {$sel}",
+      "exp": "operand select 1",
       "expErrors": [{ "type": "bad-option" }, { "type": "bad-selector" }]
     },
     {
       "src": ".local $sel = {1 :integer select=$bad} .match $sel 1 {{ONE}} * {{variable select {$sel}}}",
       "params": [{ "name": "bad", "value": "exact" }],
-      "exp": "variable select {$sel}",
+      "exp": "variable select 1",
       "expErrors": [{ "type": "bad-option" }, { "type": "bad-selector" }]
     }
   ]
--- a/test/tests/functions/number.json
+++ b/test/tests/functions/number.json
@@ -290,13 +290,13 @@
     },
     {
       "src": ".local $bad = {exact} {{variable select {1 :number select=$bad}}}",
-      "exp": "variable select {|1|}",
+      "exp": "variable select 1",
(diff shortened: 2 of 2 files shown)


test/tests/functions/integer.json
```diff
@@ -58,13 +58,13 @@
...
+      "exp": "operand select 1",
       "expErrors": [{ "type": "bad-option" }, { "type": "bad-selector" }]
     },
     {
       "src": ".local $sel = {1 :integer select=$bad} .match $sel 1 {{ONE}} * {{variable select {$sel}}}",
       "params": [{ "name": "bad", "value": "exact" }],
-      "exp": "variable select {$sel}",
+      "exp": "variable select 1",
       "expErrors": [{ "type": "bad-option" }, { "type": "bad-selector" }]
     }
```
aphillips, line 69: shouldn't the result here be `nomatch` (we don't say what happens with `nomatch`??) and thus result in an empty pattern (and the logo)?

test/tests/functions/integer.json
```diff
@@ -58,13 +58,13 @@
     },
     {
       "src": ".local $sel = {1 :integer select=exact} .local $bad = {$sel :integer} .match $bad 1 {{ONE}} * {{operand select {$bad}}}",
-      "exp": "operand select {$sel}",
+      "exp": "operand select 1",
       "expErrors": [{ "type": "bad-option" }, { "type": "bad-selector" }]
     },
```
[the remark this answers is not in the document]
eemeli: The bad-option and failure to select is because the `select` is set in the operand of the `$bad` expression, and that's not allowed: https://github.com/unicode-org/message-format-wg/blob/662076e128887cad9caa5502b4d4601c5c9bb231/spec/functions/number.md?plain=1#L641-L644

test/tests/functions/integer.json
eemeli: No, `nomatch` is not possible with MF 2.0 because we've not reached consensus on #603. If we end up with a selector that doesn't support selection (as happens here), we format the `*` pattern.

test/tests/functions/integer.json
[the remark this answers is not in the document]
eemeli: Added descriptions.

test/tests/functions/integer.json
eemeli: That section is actually really excplicit, it's just complicated because we ended up doing best-match rather than first-match selection. The bad-selector error is coming from Resolve Selectors step 2.iii.c., selecting the `*` variant is due to Filter Variants step 2.i.b.a.

aphillips (approved):

catamorphism:
I can't think of any other cases where an error is emitted but the output doesn't contain any fallback strings (or fallback parts, in the case of `formatToParts`). This means that in an environment where lenient error handling can be enabled (i.e. no errors, only best-effort output), it might not be obvious from looking at the output that an error occurred. I don't think that's a big problem, but I wanted to point it out.

catamorphism (approved):

eemeli:
> I can't think of any other cases where an error is emitted but the output doesn't contain any fallback strings (or fallback parts, in the case of `formatToParts`).

Selection errors can work this way: if such happens, we emit an error and select the `*` pattern, which may well format without any error.

aphillips:
@eemeli noted:
> Selection errors can work this way: if such happens, we emit an error and select the * pattern, which may well format without any error.

Yes, we just discussed this. This feels sort of wrong? I'm not sure what the utility of emitting a non-fallback formatted pattern is. Note that with multiple selectors, the `*` pattern (the all-`*` one) may be a terrible fit. I am not arguing for a change in the tests in the circa-47 timeframe, but maybe we need to discuss it as a WG a bit more. I'll file an issue to track it.

PR #1050, opened 2025-03-04
Add test tags, initially for u: options

eemeli:
After #1012, the `u:locale` option has Draft status, while `u:id` and `u:dir` are RECOMMENDED. However, the tests for all three are in the same file, and indistinguishable from eahc other.

~We should for now drop the rather minimal tests for `u:locale`, and add more comprehensive tests when the status changes.~

See latest comments, as the PR now keeps the tests, but adds an array of tags to them.

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -50,6 +50,20 @@ is not included in the schema,
 as it is intended to be an umbrella category
 for implementation-specific errors.
 
+## Test Tags
+
+Some of the tests are for functionality that is not stable,
+i.e. is marked RECOMMENDED, OPTIONAL, or DRAFT.
+Tests for such features have a `tags` array attached to them
+to mark the features that they rely on.
+This may include one or more of the following:
+
+| Tag        | Feature                                               |
+| ---------- | ----------------------------------------------------- |
+| `u:dir`    | The [u:dir](../spec/u-namespace.md#udir) option       |
+| `u:id`     | The [u:id](../spec/u-namespace.md#uid) option         |
+| `u:locale` | The [u:locale](../spec/u-namespace.md#ulocale) option |
+
 ## Test Functions
 
 As the behaviour of some of the default registry _functions_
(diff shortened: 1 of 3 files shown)


catamorphism (approved):
The change is fine, but why not add a new file with the `u:locale` tests rather than removing them? (I realize it's just two tests, but it might still be good to keep what we have.)

eemeli:
They'll still be here in the git commit history, and these tests are really rather simple.

aphillips:
Stepping back for a moment... we need to develop a way to manage items in draft. We should have tests for new functions and options. People who implement these should be able to test them. So dumpstering these tests feels bad. Can we add a test status like "draft" so users can omit the things they didn't implement?

eemeli:
A better solution would be one that separately identifies each of the features that are not REQUIRED, as we ought to also enable the testing and validation of features that are RECOMMENDED or OPTIONAL.

aphillips:
Draft should be separate from MUST/SHOULD/MAY. We need both.

eemeli:
The PR has been rebased and refactored, such that it now keeps the tests, and add `tags` for them. The `tests/README.md` is updated with an explanation for them, along with a listing of available tags.

For an example use case, see [this JS implementation commit](https://github.com/messageformat/messageformat/commit/debdba82c7a65dec2acd0dd695cbb05c0324c341) which skips all tests with a `'u:locale'` tag.

catamorphism (approved):
Good idea with the tags!

test/README.md
```diff
@@ -50,6 +50,20 @@ is not included in the schema,
 as it is intended to be an umbrella category
 for implementation-specific errors.
 
+## Test Tags
+
+Some of the tests are for functionality that is not stable,
```
aphillips, line 55: This is not correct. Items that are not MUST are stable, just not required.

PR #1054, opened 2025-03-06
Clarify option resolution regarding fallback option values

### LINKED ISSUE #1053: Bug in option resolution
The TAG review comments, @jyasskin notes:

> In [option resolution](https://www.unicode.org/reports/tr35/dev/tr35-messageFormat.html#option-resolution), "If rv is a fallback value: If supported, emit a Bad Option error." doesn't say what to do if the BadOption error isn't supported.

The refers to this algorithm:

```
1. For each _option_:
   1. Let `id` be the string value of the _identifier_ of the _option_.
   1. Let `rv` be the _resolved value_ of the _option value_.
   1. If `rv` is a _fallback value_:
   1. Else:
      1. If the _option value_ consists of a _literal_:
         1. Mark `rv` as a _literal_ _option value_.
      1. Set `res[id]` to be `rv`.
1. Return `res`.
```

I think the desired algorithm is:

```
1. For each _option_:
   1. Let `id` be the string value of the _identifier_ of the _option_.
   1. Let `rv` be the _resolved value_ of the _option value_.
   1. If `rv` is a _fallback value_:
   1. Else:
      1. If the _option value_ consists of a _literal_:
         1. Mark `rv` as a _literal_ _option value_.
   1. Set `res[id]` to be `rv`.
1. Return `res`.
```

aphillips:
Fixes #1053.

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -395,12 +395,16 @@ For each _option_:
    1. Let `id` be the string value of the _identifier_ of the _option_.
    1. Let `rv` be the _resolved value_ of the _option value_.
    1. If `rv` is a _fallback value_:
-      1. If supported, emit a _Bad Option_ error.
+      1. Emit a _Bad Option_ error, if supported.
    1. Else:
       1. If the _option value_ consists of a _literal_:
          1. Mark `rv` as a _literal_ _option value_.
-   1. Set `res[id]` to be `rv`.
+      1. Set `res[id]` to be `rv`.
 1. Return `res`.
+
+> [!NOTE]
+> If the _resolved value_ of an _option value_ contains a _fallback value_,
+> the _option_ is intentionally omitted from the list of _options_.
 
 The result of _option resolution_ MUST be a (possibly empty) mapping
 of string identifiers to values;
@@ -409,7 +413,7 @@ This mapping can be empty.
 
 > [!NOTE]
 > The _resolved value_ of a _function_ _operand_
-> can also include resolved option values.
+> can also include resolved _option values_.
 > These are not included in the _option resolution_ result,
 > and need to be processed separately by a _function handler_.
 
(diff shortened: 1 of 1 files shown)


eemeli (changes requested):
The current indentation is intentional, and options resolving to fallback values are meant to be left out of `res`. This was most recently discussed in https://github.com/unicode-org/message-format-wg/pull/903#discussion_r1795984280.

Rather than changing the algorithm, we should probably add a note about this.

aphillips:
@eemeli I added the note you suggest. I also reworded the step in question to not make it look like an "if XXX" step in the algorithm (leading to the commenter's query about "where's the else"). See what you think.

spec/formatting.md
```diff
@@ -395,21 +395,25 @@ For each _option_:
...
+      1. Emit a _Bad Option_ error, if supported.
    1. Else:
       1. If the _option value_ consists of a _literal_:
          1. Mark `rv` as a _literal_ _option value_.
       1. Set `res[id]` to be `rv`.
 1. Return `res`.
 
+> [!NOTE]
+> If the _resolved value_ of an _option value_ contains a _fallback value_,
+> the _option_ is intentionally omitted from the list of _options_.
```
eemeli, line 407:
```suggestion
> If the _resolved value_ of an _option value_ is a _fallback value_,
> the _option_ is intentionally omitted from the mapping of resolved options.
```

spec/formatting.md
```diff
@@ -395,21 +395,25 @@ For each _option_:
...
+
 The result of _option resolution_ MUST be a (possibly empty) mapping
 of string identifiers to values;
 that is, errors MAY be emitted, but such errors MUST NOT be fatal.
 This mapping can be empty.
 
 > [!NOTE]
 > The _resolved value_ of a _function_ _operand_
-> can also include resolved option values.
+> can also include resolved _option values_.
```
eemeli, line 416: An _option value_ is a syntax construct, and therefore cannot appear in a mapping of resolved values.

```suggestion
> can also include resolved option values.
```

eemeli (approved):
See inline for a few necessary fixes; otherwise good.

PR #1057, opened 2025-03-08
Fix markup examples to show that literals work normally

### LINKED ISSUE #1029: [FEEDBACK] Use of | in markup
> **{\#button}Submit{/button} or {\#img alt=|Cancel| /}.**

This is misleading. Someone might think that the |​ characters are corresponding to ' or " when they don't really represent that, or think they are needed when they are not. Recommend removing the | characters or adding a note.

aphillips:
Fixes #1029

Removed unnecessary literal quotes from examples and ensured that there was also an option that required pipes so both were visible. Did not add any notes to explain this.

### DIFF
--- a/spec/syntax.md
+++ b/spec/syntax.md
@@ -635,14 +635,14 @@ markup = "{" o "#" identifier *(s option) *(s attribute) o ["/"] "}"  ; open and
 > A _message_ with one `button` markup span and a standalone `img` markup element:
 >
 > ```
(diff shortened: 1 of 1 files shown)


macchiati (approved):

aphillips:
Meets fast track, but will save to merge in the 2025-03-10 call

mihnita (approved):

PR #1059, opened 2025-03-17
Add requirement and stability level to the tests schema

mihnita:


eemeli (changes requested):
I would prefer labelling each separate non-required non-stable part of functionality with a unique identifier, but I'm not opposed to the tagging proposed here -- I just don't find it particularly useful.

I do, however, think that any such tags should not only be introduced to the schema, but also applied to the test data.

aphillips:
> I do, however, think that any such tags should not only be introduced to the schema, but also applied to the test data.

Once we agreed on the design, then we should modify all of the tests to match.

> I would prefer labelling each separate non-required non-stable part of functionality with a unique identifier, but I'm not opposed to the tagging proposed here -- I just don't find it particularly useful.

I'm not sure I follow. Can you give an example? I can see having an identifier for each test, but their use would be separate. These tags would allow one to select (or de-select) tests based on stability level or based on requirement level. The latter isn't *that* useful, since presumably most implementations would implement most optional behavior and only need to de-select specific tests.

aphillips:
closing in 2025-03-24 call in favor of a revision of #1050

PR #1060, opened 2025-03-22
In tests, use "text" rather than "literal" as the type for formatted-parts text parts

eemeli:
The test suite `expParts` should not expect message text parts to be formatted as `"literal"`, but as `"text"`. The current type is misleading, given that in our spec _literals_ are a different thing.

### DIFF
--- a/test/schemas/v0/tests.schema.json
+++ b/test/schemas/v0/tests.schema.json
@@ -237,7 +237,7 @@
       "items": {
         "oneOf": [
           {
-            "description": "Message literal part.",
+            "description": "Message text part.",
             "type": "object",
             "additionalProperties": false,
             "required": [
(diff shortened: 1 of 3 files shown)


aphillips (approved):

gibson042 (approved):
The use of "literal" in this spec is already at odds with every ECMA-402 `formatToParts` method, since the [PartitionPattern](https://tc39.es/ecma402/#sec-partitionpattern) operation supporting them uses type "literal" for each returned part that is not from a placeholder (such as the parts of Intl.DateTimeFormat output that are not time scale components, e.g. punctuators like "-"/":" or affix characters like "年"/"月"/"日"). However, I agree that the sequences of non-placeholder text in a MessageFormat pattern are distinguished from those implementation-introduced parts (_MessageFormat itself operating one level up, since the patterns are supplied by developers rather than internal to the implementation_), and support their representation with type "text" as proposed here.

PR #1061, opened 2025-03-25
In test suite, drop source from non-fallback formatted parts

eemeli:
While reviewing the messageformat.dev documentation, I started to re-think the `source` value that's currently included in formatted parts.

I think we should drop it, as it makes a part of what ought to be considered implementation details into a public API (such as the name of a message variable or function). If a user does have a need to identify a specific part in the output, that ought to be explicit, and it's what we have `u:id` for.

For markup in particular it adds no value, as the `source` value is constructable from the `name` and the `kind`.

CC @ryzokuken

### DIFF
--- a/test/schemas/v0/tests.schema.json
+++ b/test/schemas/v0/tests.schema.json
@@ -290,9 +290,6 @@
                   "close"
                 ]
               },
-              "source": {
-                "type": "string"
-              },
               "name": {
                 "type": "string"
               },
@@ -308,8 +305,7 @@
             "description": "Message expression part.",
             "type": "object",
             "required": [
-              "type",
-              "source"
+              "type"
             ],
             "not": {
               "required": [
@@ -319,9 +315,6 @@
             },
             "properties": {
               "type": {
-                "type": "string"
-              },
-              "source": {
                 "type": "string"
               },
               "locale": {
@@ -334,11 +327,7 @@
                   "properties": {
                     "type": {
                       "type": "string"
-                    },
-                    "source": {
-                      "type": "string"
-                    },
-                    "value": {}
+                    }
                   },
                   "required": [
                     "type"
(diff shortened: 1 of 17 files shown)


test/schemas/v0/tests.schema.json
```diff
@@ -347,6 +342,23 @@
               },
               "value": {}
             }
+          },
+          {
+            "description": "Fallback part.",
```
aphillips, line 347: Can we be more verbose with the descriptions? They're here as documentation, after all...

test/tests/u-options.json
```diff
@@ -102,13 +89,7 @@
...
         { "type": "text", "value": "hello " },
         { "type": "bidiIsolation", "value": "\u2066" },
-        {
-          "type": "string",
-          "source": "|world|",
-          "dir": "ltr",
-          "id": "foo",
-          "value": "world"
-        },
+        { "type": "string", "dir": "ltr", "id": "foo", "value": "world" },
```
aphillips, line 92: any reason to lose the formatting?

test/schemas/v0/tests.schema.json
eemeli: This is literally a fallback part, i.e. the result of formatting a value that experienes fallback. I don't really know what else to say about it?

test/tests/u-options.json
eemeli: Do you mean that this is on a single line, rather than the multiple lines it had previously? It's because with the change, it's now short enough to easily fit on a single line.

catamorphism:
I'm not sure of the reasoning behind still including the `source` in fallback parts.

eemeli:
> I'm not sure of the reasoning behind still including the `source` in fallback parts.

It's needed to hold the representation that's also used in the string output, i.e. `source: '$x'` when the part would be formatted to a string as `{$x}`.

catamorphism (approved):

PR #1067, opened 2025-04-06
Semantic skeletons design

aphillips:
This design document contains the proposed design for including semantic skeletons into Unicode MessageFormat's functions.

This PR is based on @sffc's [doc](https://docs.google.com/document/d/1s7GeN5V0cnw9B1erfMHTWwmnxz0EJkq2v78ZjzRq2t8/edit)

### DIFF
--- a/exploration/semantic-skeletons.md
+++ b/exploration/semantic-skeletons.md
@@ -129,22 +129,32 @@ I should trust that the placeholder will produce appropriate results for my lang
 
 _What properties does the solution have to manifest to enable the use-cases above?_
 
-1. It should be possible to format operands consisting of common incremental time types
-   (e.g. milliseconds since epoch times such as `java.util.Date`, `time_t`, JS `Date`, etc.)
-2. It should be possible to format operands consisting of field-based time types
-   (e.g. those that contain seperate values per field type in a date/time, such as a year-month)
-3. It should be possible to format [floating time](https://www.w3.org/TR/timezone/#dfn-floating-time) values
-   (e.g. those that are not tied to a specific time zone, variously called local/plain/civil times)
-4. Date/time formatters should not permit users to format fields that don't exist in the value
+1. It should be possible to format operands consisting of locally-relevant date/time types, including:
+   - Temporal values such as `java.time` or JS `Temporal` values,
+   - incremental time types ("timestamps")
+     (e.g. milliseconds since epoch times such as `java.util.Date`, `time_t`, JS `Date`, etc.),
+   - field-based time types
+     (e.g. those that contain seperate values per field type in a date/time, such as a year-month),
+   - [floating time](https://www.w3.org/TR/timezone/#dfn-floating-time) values
+     (e.g. those that are not tied to a specific time zone, variously called local/plain/civil times),
+   -  or other local exotica (Java `Calendar`, C `tm` struct, etc.)
+1. Date/time formatters should not permit users to format fields that don't exist in the value
    (e.g. the "month" of a time, the "hour" of a date)
-5. Date/time formatters should not permit users to format bad combinations of fields
+1. Date/time formatters should not permit users to format bad combinations of fields
    (e.g. `MMMMmm` (month-minute), `yyyyjm` (year-hour-minute), etc.)
-6. Date/time formatters should permit users to control or influence the width of indvidual fields
-   in a manner similar to classical skeletons
-   (e.g. `yMMd` vs. `yMMMd` vs. `yMMMMd` => 04/06/2025 vs. Apr 6, 2025 vs. April 6, 2025)
-7. Developers, translators, and UI designers should only have to learn a single "microsyntax" or set of options for date formatting.
-   Such a syntax or option set should be easy to understand only from the placeholder.
-   Such a syntax or option set should not require translators to alter the values in most or all locales.
+1. Date/time formatters should permit users to specify the desired width of indvidual fields
+   in a manner similar to classical skeletons,
+   while relying on locale data to prevent undesirable results.
+   For example:
+   | Classical Skeleton | `en-US` Output | 
+   |---|---|
+   | `yMd` | 04/06/2025 |
+   | `yMMMd` | Apr 6, 2025 |
+   | `yMMMMd` | April 6, 2025 |
+1. Developers, translators, and UI designers should not have to learn
+   multiple new microsyntaxes or multiple different sets of options for date/time value formatting.
+1. Any microsyntax or option set specified should be easy to understand only from the placeholder.
+1. Any microsyntax or option set specified should not _require_ translators to alter the values in most or all locales.
 
 ## Constraints
 
--- a/exploration/semantic-skeletons.md
+++ b/exploration/semantic-skeletons.md
@@ -139,9 +139,10 @@ such as `java.text.SimpleDateFormat`.
 This section considers some potential arguments against the design
 or captures frequently asked questions about it.
 
-**What if semantic skeletons doesn't support the format I want?
-Unlike picture strings or classical skeletons, semantic skeletons do not allow unrestricted
-composition of date/time formats.**
+**What if semantic skeletons don't support the format I want?**
+
+Unlike picture strings or classical skeletons,
+semantic skeletons do not allow unrestricted composition of date/time formats.
 
 If there were an overlooked format, it could be added in a future release.
 However, the designers of semantic skeletons have considered the breadth of use cases.
--- a/exploration/semantic-skeletons.md
+++ b/exploration/semantic-skeletons.md
@@ -30,11 +30,10 @@ Previously, ICU MessageFormat provided support for "classical skeletons",
 using a microsyntax derived from familiar 'picture strings' (see below)
 combined with code in ICU (`DateTimePatternGenerator`) to produce the desired date/time value format.
 
-`Intl.DateTimeFormat` provided options to provide a similar capability.
-These weren't "skeletons" from the point of view that they didn't use a dedicated microsyntax
-similar to 'picture strings',
-but the effect was the same:
-users specified which fields they wanted with which display options (such as width).
+`Intl.DateTimeFormat` provides options that provide a similar capability.
+These aren't "skeletons" from the point of view that they don't use a dedicated microsyntax similar to 'picture strings',
+but the effect is the same:
+users specify which fields they want with which display options (such as width).
 For example:
 ```javascript
 options = {
--- a/exploration/semantic-skeletons.md
+++ b/exploration/semantic-skeletons.md
@@ -11,7 +11,7 @@ Status: **Proposed**
 		<dt>First proposed</dt>
 		<dd>2024-04-06</dd>
 		<dt>Pull Requests</dt>
-		<dd>#000</dd>
+		<dd><a href="https://github.com/unicode-org/message-format-wg/pull/1067">#1067</a></dd>
 	</dl>
 </details>
 
--- a/exploration/semantic-skeletons.md
+++ b/exploration/semantic-skeletons.md
@@ -68,7 +68,7 @@ Many date/time formatting regimes provide for "picture strings".
 A "picture string" is a pattern using a microsyntax in which the user (developer, translator, UX designer)
 exactly specifies the desired format of the date/time value.
 In a picture string, separators, spaces, and other formatting are explicitly specified.
-This provides a lot of power to the devleoper or user experience designer, in terms of specifying formatting.
+This provides a lot of power to the developer or user experience designer, in terms of specifying formatting.
 For example: `MMM dd, yyyy` or `yyyy-dd-MM'T'HH:mm:ss`
   
 Picture strings require translators to interact with and "translate" the picture string
--- a/exploration/semantic-skeletons.md
+++ b/exploration/semantic-skeletons.md
@@ -280,6 +280,21 @@ what is the format?
 - `Jan 1, 1970, 12:00:00 AM UTC` (value wins; user cannot adjust)
 - `Jan 1, 1970, 2:00:00 AM EET` (option wins; inconsistent with some Temporal behaviors on Java and JS)
 
+### Inflection
+
+```
+new Intl.DateTimeFormat('fi', { month: 'long' }).format()
+// 'toukokuu'
+
+new Intl.DateTimeFormat('fi', { day: 'numeric', month: 'long' }).format()
+// '2. toukokuuta'
+```
+
+As a developer, I need the formatted values to inflect fields and formats correctly,
+depending on the combination chosen for display.
+For example, as shown above, 
+I want the month name to be inflected for standalone when used by itself,
+but properly inflected when used in a month-day combination.
 
 ## Requirements
 
@@ -407,7 +422,8 @@ Such a list might look like:
 
 - Standalone fields
   - `:day`, `:weekday`, `:month`, `:year`, `:hour`, `:minute`, `:second`, `:zone`, `:era`
-  - plus the common combo `:time` and `:date`
+- Common combinations
+  - `:time` and `:date`
 - Date Field Sets
   - `:day-weekday`, `:month-day`, `:month-day-weekday`, `:year-month-day`, `:year-month-day-weekday`
 - Composite field sets
--- a/exploration/semantic-skeletons.md
+++ b/exploration/semantic-skeletons.md
@@ -34,7 +34,7 @@ combined with code in ICU (`DateTimePatternGenerator`) to produce the desired da
 These weren't "skeletons" from the point of view that they didn't use a dedicated microsyntax
 similar to 'picture strings',
 but the effect was the same:
-user's specified which fields they wanted with which display options (such as width).
+users specified which fields they wanted with which display options (such as width).
 For example:
 ```javascript
 options = {
--- a/exploration/semantic-skeletons.md
+++ b/exploration/semantic-skeletons.md
@@ -169,11 +169,12 @@ such as "11 PM April" (`jjMMMM` or `HHaMMMM`) or "2 2025" (`dyyyy`).
 
 **What about specialized formats, such as ISO8601?**
 
-These should be provided via other means that requiring a specialized pattern 
-and the (optional) `@locale` attribute. Do you really want to support this,
+These should be provided via other means than requiring a specialized pattern 
+and the (optional) `u:locale` option.
+Do you really want to support this,
 given that it can then be used for other things:
 ```
-{$now :datetime pattern=|yyyy-MM-dd'T'HH:mm:ss.sssz| @locale=und timezone=UTC}
+{$now :datetime pattern=|yyyy-MM-dd'T'HH:mm:ss.sssz| u:locale=und timezone=UTC}
 ```
 
 **What do we call a [floating time value](https://www.w3.org/TR/timezone/#dfn-floating-time)?**
--- a/exploration/semantic-skeletons.md
+++ b/exploration/semantic-skeletons.md
@@ -380,7 +380,7 @@ _Pros_
 _Cons_
 - Not fully type-safe.
 
-#### Use separate semantic functions
+#### Use separate typed functions
 
 `:date`, `:time`, `:datetime`, `:zoneddatetime`, *maybe* `:zoneddate`, `:zonedtime`, `:timezone`
 
@@ -395,7 +395,45 @@ _Pros_
 - Better at documenting the message author's intention
 
 _Cons_
-- _Lots_ of functions
+- More functions
+
+
+#### Use separate skeleton functions
+
+Define functions according to the available "field sets" or fields in the semantic skeletons spec.
+Use options to handle field widths.
+Such a list might look like:
+
+- Standalone fields
+  - `:day`, `:weekday`, `:month`, `:year`, `:hour`, `:minute`, `:second`, `:zone`, `:era`
+  - plus the common combo `:time` and `:date`
+- Date Field Sets
+  - `:day-weekday`, `:month-day`, `:month-day-weekday`, `:year-month-day`, `:year-month-day-weekday`
+- Composite field sets
+  - `:date-zone`, `:date-time-zone`, `:time-zone` (note: `time-zone` is two fields, not "timezone")
+
+Examples:
+```
+Your package arrived on {$d :date} at {$d :time}.
+    Your package arrived on Apr 27, 2025 at 10:50 AM.
+Your package arrived on {$d :month-day-weekday} at {$d :time-zone}.
+    Your package arrived on Sunday, April 27th at 10:50 AM PDT.
+Your package arrived on {$d :month-day-weekday month=medium weekday=full} at {$d :time-zone display=full zone=short}.
+    Your package arrive on Sunday, Apr 27th at 10:50 PDT.
+```
+
+
+_Pros_
+- Options focused on field widths
+- Self-documenting: obvious what fields are shown
+- Reserves options for field width
+
+_Cons_
+- _Nineteen_ (or so) functions
+- Names appear generative, but aren't actually
+- Options might be baroque. Some option values might not be compatible with one another.
+  - Expression-wide options (short/medium/long/full) and field-level options might both be needed together
+
 
 ---
 
--- a/exploration/semantic-skeletons.md
+++ b/exploration/semantic-skeletons.md
@@ -24,11 +24,27 @@ _What is this proposal trying to achieve?_
 "Semantic skeletons" are a method introduced in CLDR 46 for programmatically selecting a datetime pattern for formatting. 
 There is a fixed set of acceptable semantic skeletons.
 
+#### Avoid 'classical skeletons'
+
 Previously, ICU MessageFormat provided support for "classical skeletons",
-using a microsyntax derived from familiar picture strings (see below)
+using a microsyntax derived from familiar 'picture strings' (see below)
 combined with code in ICU (`DateTimePatternGenerator`) to produce the desired date/time value format.
-`Intl.DateTimeFormat` uses "option bags" to provide a similar capability.
-A classical skeleton allowed users to express the desired fields and field widths in a formatted date/time value.
+
+`Intl.DateTimeFormat` provided options to provide a similar capability.
+For example:
+```javascript
+options = {
+  hour: "numeric",
+  minute: "numeric",
+  second: "numeric",
+  timeZone: "Australia/Sydney",
+  timeZoneName: "short",
+};
+console.log(new Intl.DateTimeFormat("en-AU", options).format(date));
+// "2:00:00 pm AEDT"
+```
+
+Classical skeleton solutions allow users to express the desired fields and field widths in a formatted date/time value.
 The runtime uses locale data to determine minutiae such as 
 field-order, 
 separators, 
@@ -39,11 +55,12 @@ etc. to produce the desired output.
 Advantages of semantic skeletons over classical skeletons:
 
 - Provides all and only those combinations that make sense
-   - Allows for more efficient implementation, since there is no need to support "crazy" combinations like "month-hour"
+   - Allows for more efficient implementation, since there is no need to support combinations like "month-hour"
+     that are not useful to users
 - Allows for a more clear, ergonomic placeholder syntax, since the number of options can be limited
 - Easier for user experience designers to specify, developers to implement, and translators to interpret
 
-### Avoid 'picture strings'
+#### Avoid 'picture strings'
 
 The MFWG early on considered including support for "picture strings" in the formatting of date/time values.
 There is a Working Group consensus **_not_** to support picture strings in Unicode MessageFormat, if possible.
@@ -129,6 +146,9 @@ I should trust that the placeholder will produce appropriate results for my lang
 
 _What properties does the solution have to manifest to enable the use-cases above?_
 
+Users want the most intuitive formats and outcomes to be the defaults.
+They should not have to coerce or convert normal date/time types in order to do simple operations.
+
 1. It should be possible to format operands consisting of locally-relevant date/time types, including:
    - Temporal values such as `java.time` or JS `Temporal` values,
    - incremental time types ("timestamps")
@@ -180,17 +200,17 @@ In this section, we use a scheme similar to `FieldSetBuilder` linked earlier.
 Options:
 
 ```
-{$date :datetime dateFields="YMD"}  
-{$date :datetime date="YMD"}  
-{$date :datetime fields="YMD"}
+{$date :datetime dateFields=YMD}  
+{$date :datetime date=YMD}  
+{$date :datetime fields=YMD}
 ```
 
 #### TimePrecision
 
 Options:
 ```
-{$date :datetime timePrecision="minute"}  
-{$date :datetime time="minute"}
+{$date :datetime timePrecision=minute}  
+{$date :datetime time=minute}
 ```
 (TODO: Add others)
 
@@ -198,12 +218,46 @@ Options:
 
 Some choices:
 
-1. A single :datetime function  
-   1. Pro: All in one place  
-   2. Con: More combinations of options that form invalid skeletons  
-2. :date, :time, and :datetime  
-   1. Pro: More tailored and type-safe  
-   2. Con: Not fully type-safe  
-3. :date, :time, :datetime, :zoneddatetime, *maybe* :zoneddate, :zonedtime, :timezone  
-   1. Pro: Most type-safe  
-   2. Con: Lots of functions
+#### A single `:datetime` function
+
+_Pros_
+- Everything is in one place
+
+_Cons_
+- More combinations of options that can form invalid skeletons
+- Can require verbose placeholders
+
+#### Use separate `:date`, `:time` and `:datetime` functions
+
+_Pros_
+- More tailored to specific user requirements
+- `:date` and `:time` are somewhat self-documenting about the message author's formatting intention
+
+_Cons_
+- Not fully type-safe.
+
+#### Use separate semantic functions
+
+`:date`, `:time`, `:datetime`, `:zoneddatetime`, *maybe* `:zoneddate`, `:zonedtime`, `:timezone`
+
+Problem: Most users are likely to prefer date/time/datetime to zoneddate/zonedtime/zoneddatetime
+as the default formatting functions.
+Most "classical" time types are timestamps.
+User intentions might not be met by these names.
+
+Problem: Different platforms cannot agree on what to call a Floating Time Value: HTML and Java use `LocalXXX`,
+JavaScript has adopted `PlainXXX`,
+some others use different terms, such as `CivilXXX`.
+
+_Pros_
+- Helps users get the right results
+- Could be less verbose
+- Better at documenting the message author's intention
+
+_Cons_
+- _Lots_ of functions
+
+---
+
+
+
--- a/exploration/semantic-skeletons.md
+++ b/exploration/semantic-skeletons.md
@@ -31,7 +31,35 @@ Advantages of semantic skeletons over classical skeletons:
 
 Advantages of semantic skeletons over other mechanisms:
 
-- Unlike "picture strings" (`MMM dd, yyyy`), does not require translation
+- Unlike "picture strings" (`MMM dd, yyyy`), skeletons do not require translation.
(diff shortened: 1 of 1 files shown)


exploration/semantic-skeletons.md
```diff
@@ -0,0 +1,180 @@
...
+dd-MM-yyyy=21-04-2025
+21 avr. 2025
+---
+ja-JP
+MMM dd, yyyy=4月 21, 2025
+dd MMM, yyyy=21 4月, 2025
+MM/dd/yyyy=04/21/2025
+dd-MM-yyyy=21-04-2025
+2025年4月21日
+```
```
eemeli, line 62: This should really be a table.

exploration/semantic-skeletons.md
```diff
@@ -0,0 +1,180 @@
...
+<details>
+	<summary>Metadata</summary>
+	<dl>
+		<dt>Contributors</dt>
+		<dd>@sffc</dd>
+    <dd>@aphillips</dd>
+		<dt>First proposed</dt>
+		<dd>2024-04-06</dd>
+		<dt>Pull Requests</dt>
+		<dd>#000</dd>
```
eemeli, line 14:
```suggestion
		<dd><a href="https://github.com/unicode-org/message-format-wg/pull/1067">#1067</a></dd>
```

exploration/semantic-skeletons.md
```diff
@@ -0,0 +1,180 @@
...
+Advantages of semantic skeletons over classical skeletons:
+
+- A smaller set of acceptable options focused on producing outputs that are sensical
+- Allows for a more efficient implementation
+- Allows for a more clear, ergonomic API
+- More future-proof since CLDR recently added them (??)
+
+Advantages of semantic skeletons over other mechanisms:
+
+- Unlike "picture strings" (`MMM dd, yyyy`), skeletons do not require translation.
```
eemeli, line 34: This is a straw-man argument, as picture strings have not at any point been even suggested for MF2. That should at least be noted somewhere.

exploration/semantic-skeletons.md
```diff
@@ -0,0 +1,180 @@
...
+
+"Semantic skeletons" are a method introduced in CLDR 46 for programmatically selecting a datetime pattern for formatting. 
+There is a fixed set of acceptable semantic skeletons.
+
+Advantages of semantic skeletons over classical skeletons:
+
+- A smaller set of acceptable options focused on producing outputs that are sensical
+- Allows for a more efficient implementation
+- Allows for a more clear, ergonomic API
+- More future-proof since CLDR recently added them (??)
```
eemeli, line 30: This line of argumentation makes no sense. The new shiny thing being new and shiny does not mean that it'll last as long as tried and tested solutions.

exploration/semantic-skeletons.md
```diff
@@ -0,0 +1,180 @@
...
+As a translator, I don't want to have to "translate" or modify a date/time placeholder to suit
+my language's needs.
+I should trust that the placeholder will produce appropriate results for my language.
+
+## Requirements
+
+_What properties does the solution have to manifest to enable the use-cases above?_
+
+1. It should be possible to format common incremental time types
+   (e.g. milliseconds since epoch times)
```
eemeli, line 114: This is an odd one to have as the very first requirement. I would challenge that it's _at all_ common to need to format epoch milliseconds in a localizable manner.

exploration/semantic-skeletons.md
```diff
@@ -0,0 +1,180 @@
...
+2. It should be possible to format field-based time types
+   (e.g. those that contain seperate values per field type in a date/time, such as a year-month)
+3. It should be possible to format [floating time](https://www.w3.org/TR/timezone/#dfn-floating-time) values
+   (e.g. those that are not tied to a specific time zone)
```
eemeli, line 118: Why do we care about the input value's type? Are these not implementation concerns?

exploration/semantic-skeletons.md
```diff
@@ -0,0 +1,180 @@
...
+4. Date/time formatters should not permit users to format fields that don't exist in the value
+   (e.g. the "month" of a time, the "hour" of a date)
```
eemeli, line 120: Again, isn't this an implementation concern? I don't see what this has to do with the shape of the formatting options.

exploration/semantic-skeletons.md
```diff
@@ -0,0 +1,180 @@
...
+5. Date/time formatters should not permit users to format bad combinations of fields
+   (e.g. `MMMMmm` (month-minute), `yyyyjm` (year-hour-minute), etc.)
+6. Date/time formatters should permit users to control or influence the width of indvidual fields
+   in a manner similar to classical skeletons
+   (e.g. `yMMd` vs. `yMMMd` vs. `yMMMMd` => 04/06/2025 vs. Apr 6, 2025 vs. April 6, 2025)
```
eemeli, line 125: Influence, maybe; control, no. For example, not all locales have a sensible "short" form for months that makes sense when formatting as a full date, and enforcing its use can produce bad output.

exploration/semantic-skeletons.md
```diff
@@ -0,0 +1,180 @@
...
+7. Developers, translators, and UI designers should only have to learn a single "microsyntax" or set of options for date formatting.
+   Such a syntax or option set should be easy to understand only from the placeholder.
+   Such a syntax or option set should not require translators to alter the values in most or all locales.
```
eemeli, line 128: Developers, translators, and UI designers do not work in a vacuum outside MF2, and they presumably are already familiar with some microsyntax or date formatting options. Given that, this requirement seems to be asking for the most common choice to be selected, and that seems surprising.

exploration/semantic-skeletons.md
aphillips: I can note that. Actually, picture strings were considered in the way-long ago. This text was added to acknowledge their existence (especially in MF1!) and document why we never consider them. There was a discussion of them in yesterday (2025-04-22)'s call which prompted these latest edits. 

I will add a note to point out the WG consensus not to support picture strings.

exploration/semantic-skeletons.md
aphillips: "Incremental time types" are things like Javascript `Date`, Java `java.util.Date`, `time_t`, etc.

I would find it extremely odd if such types could not be formatted. I should change the parenthetical to be clearer. Formatting random `long` values is not what I meant.

exploration/semantic-skeletons.md
aphillips: These are implementation concerns. But they generate requirements for date/time formatting options.

exploration/semantic-skeletons.md
eemeli: Ah; this is meant to be about input, rather than output. The phrasing confused me.

exploration/semantic-skeletons.md
eemeli: Really? Could you give an example, because I don't see how.

exploration/semantic-skeletons.md
sffc: I find this to be a requirement looking for a problem. _Of course_ if a language has a data type that represents a date, it should be able to be formatted with a MessageFormat date formatter.

exploration/semantic-skeletons.md
sffc: This is the main benefit of the semantic skeleton design, and I think points 1-4 can all be merged into this one. The point is that not all date types are the same and not all formatters are the same, and the date type needs to be able to expose all fields that the formatter needs.

exploration/semantic-skeletons.md
aphillips: done

exploration/semantic-skeletons.md
aphillips: Done

exploration/semantic-skeletons.md
aphillips: Removed the last argument. Clarified the earlier ones.

exploration/semantic-skeletons.md
aphillips: @sffc It might be possible to interpret this as only supporting Temporal values. It does no harm to spell out the need to format the various kinds of "classical" timekeeping.

exploration/semantic-skeletons.md
aphillips: Control is probably too strong and influence is perhaps too weak. See what you think of the updated...

exploration/semantic-skeletons.md
```diff
@@ -0,0 +1,209 @@
...
+field-order, 
+separators, 
+spacing, 
+field-length, 
+etc. to produce the desired output.
+
+Advantages of semantic skeletons over classical skeletons:
+
+- Provides all and only those combinations that make sense
+   - Allows for more efficient implementation, since there is no need to support "crazy" combinations like "month-hour"
```
catamorphism, line 42:
```suggestion
   - Allows for more efficient implementation, since there is no need to support unlikely combinations like "month-hour"
```

Avoid quoted adjectives in favor of a more precise description, if possible. Is it really true that _no one_ would want to use these combinations, or (my guess) is it more likely to be an error than not?

exploration/semantic-skeletons.md
```diff
@@ -0,0 +1,209 @@
...
+
+### Avoid 'picture strings'
+
+The MFWG early on considered including support for "picture strings" in the formatting of date/time values.
+There is a Working Group consensus **_not_** to support picture strings in Unicode MessageFormat, if possible.
+Many date/time formatting regimes provide for "picture strings".
+A "picture string" is a pattern using a microsyntax in which the user (developer, translator, UX designer)
+exactly specifies the desired format of the date/time value.
+In a picture string, separators, spaces, and other formatting are explicitly specified.
+This provides a lot of power to the devleoper or user experience designer, in terms of specifying formatting.
```
catamorphism, line 54:
```suggestion
This provides a lot of power to the developer or user experience designer, in terms of specifying formatting.
```

exploration/semantic-skeletons.md
```diff
@@ -0,0 +1,209 @@
...
+## Requirements
+
+_What properties does the solution have to manifest to enable the use-cases above?_
+
+1. It should be possible to format operands consisting of locally-relevant date/time types, including:
+   - Temporal values such as `java.time` or JS `Temporal` values,
+   - incremental time types ("timestamps")
+     (e.g. milliseconds since epoch times such as `java.util.Date`, `time_t`, JS `Date`, etc.),
+   - field-based time types
+     (e.g. those that contain seperate values per field type in a date/time, such as a year-month),
```
catamorphism, line 137:
```suggestion
     (e.g. those that contain separate values per field type in a date/time, such as a year-month),
```

exploration/semantic-skeletons.md
```diff
@@ -0,0 +1,209 @@
...
+In this section, we use a scheme similar to `FieldSetBuilder` linked earlier.
+
+#### DateTime fields
+
+Options:
+
+```
+{$date :datetime dateFields="YMD"}  
+{$date :datetime date="YMD"}  
+{$date :datetime fields="YMD"}
```
catamorphism, line 185: Are `dateFields`, `date` and `fields` just different possible names for the same option? Or do they mean different things? (Same question about `timePrecision` vs. `time` below.)

sffc: Different possible names for the same option.

exploration/semantic-skeletons.md
```diff
@@ -0,0 +1,209 @@
...
+```
```
catamorphism, line 186:
```suggestion
```

The names `dateFields`, `date`, and `fields` are candidate names for the option that specifies the semantic skeleton string to be used for formatting the date/time value.
```

File: exploration/semantic-skeletons.md [hunk 15]
```diff
@@ -0,0 +1,209 @@
...
+
+#### TimePrecision
+
```
catamorphism, line 189: I think this needs a little more explanation (it should be possible to follow this doc without reading the other linked-to docs).

catamorphism (changes requested):

exploration/semantic-skeletons.md
duerst: I'm not sure why all the skeletons have 'yyyy' at the end even for Japanese. In Japanese, the year always comes at the start. Something like "4月 21, 2025" looks like a mess, and 04/21/2025 or 21-04-2025 will most probably be misunderstood one way or another. "2025年4月21日" or "2025/4/21" or something similar will all be easily understood.

aphillips: @duerst That's the point: they are *not* skeletons in the table. They are picture strings. Picture string produce wrong outcomes.

exploration/semantic-skeletons.md
aphillips: done

exploration/semantic-skeletons.md
aphillips: This part of the spec is basically empty and in need of work. I started to noodle around with it today. Each one of the options will need quasi-complete descriptions in order to see how they'll work and the relative usability of each.

exploration/semantic-skeletons.md
```diff
@@ -0,0 +1,266 @@
...
+   - [floating time](https://www.w3.org/TR/timezone/#dfn-floating-time) values
+     (e.g. those that are not tied to a specific time zone, variously called local/plain/civil times),
+   -  or other local exotica (Java `Calendar`, C `tm` struct, etc.)
+1. Date/time formatters should not permit users to format fields that don't exist in the value
+   (e.g. the "month" of a time, the "hour" of a date)
+1. Date/time formatters should not permit users to format bad combinations of fields
+   (e.g. `MMMMmm` (month-minute), `yyyyjm` (year-hour-minute), etc.)
+1. Date/time formatters should permit users to specify the desired width of indvidual fields
+   in a manner similar to classical skeletons,
+   while relying on locale data to prevent undesirable results.
```
sffc, line 169: I want to be more clear about what is "required" versus "wanted".

I think "required" is that users should specify an overall length, and "wanted" to hint at the width of an individual field independently of the overall length.

exploration/semantic-skeletons.md
```diff
@@ -0,0 +1,266 @@
...
+JavaScript has adopted `PlainXXX`,
+some others use different terms, such as `CivilXXX`.
+
+_Pros_
+- Helps users get the right results
+- Could be less verbose
+- Better at documenting the message author's intention
+
+_Cons_
+- _Lots_ of functions
```
sffc, line 261: But only 1 more than the previous option. The only one this adds is `:zoneddatetime`.

exploration/semantic-skeletons.md
```diff
@@ -0,0 +1,266 @@
...
+- `:date` and `:time` are somewhat self-documenting about the message author's formatting intention
+
+_Cons_
+- Not fully type-safe.
+
+#### Use separate semantic functions
+
+`:date`, `:time`, `:datetime`, `:zoneddatetime`, *maybe* `:zoneddate`, `:zonedtime`, `:timezone`
+
+Problem: Most users are likely to prefer date/time/datetime to zoneddate/zonedtime/zoneddatetime
```
sffc, line 246: Great, so if naming is the only problem here, let's bikeshed the names and make another set of names. Let's try not to deviate _too much_ from java.time and Temporal. :)

sffc (approved):
The more I think about it, the more I think we should either go all-in on 4 functions or all-in on 1 function. The "middle ground" `:date`, `:time`, `:datetime` just seems flawed to me: it is trying to be type-safe but it fails at being type-safe.

exploration/semantic-skeletons.md
```diff
@@ -0,0 +1,266 @@
...
+   - Temporal values such as `java.time` or JS `Temporal` values,
+   - incremental time types ("timestamps")
+     (e.g. milliseconds since epoch times such as `java.util.Date`, `time_t`, JS `Date`, etc.),
+   - field-based time types
+     (e.g. those that contain separate values per field type in a date/time, such as a year-month),
+   - [floating time](https://www.w3.org/TR/timezone/#dfn-floating-time) values
+     (e.g. those that are not tied to a specific time zone, variously called local/plain/civil times),
+   -  or other local exotica (Java `Calendar`, C `tm` struct, etc.)
+1. Date/time formatters should not permit users to format fields that don't exist in the value
+   (e.g. the "month" of a time, the "hour" of a date)
```
eemeli, line 164: This presumably should not be understood to mean that e.g. epoch milliseconds ought not be formattable as year-month-date, given that the value only contains milliseconds, yes?

```suggestion
1. Date/time formatter options should be able to impose restrictions on acceptable values,
   such as formatting a month requiring a date, and formatting an hour requiring a time.
```

exploration/semantic-skeletons.md
```diff
@@ -0,0 +1,266 @@
...
+   while relying on locale data to prevent undesirable results.
+   For example:
+   | Classical Skeleton | `en-US` Output | 
+   |---|---|
+   | `yMd` | 04/06/2025 |
+   | `yMMMd` | Apr 6, 2025 |
+   | `yMMMMd` | April 6, 2025 |
+1. Developers, translators, and UI designers should not have to learn
+   multiple new microsyntaxes or multiple different sets of options for date/time value formatting.
+1. Any microsyntax or option set specified should be easy to understand only from the placeholder.
```
eemeli, line 178:
```suggestion
1. Any microsyntax or option set specified should be easy to understand only from the expression.
```

exploration/semantic-skeletons.md
```diff
@@ -0,0 +1,266 @@
...
+`:date`, `:time`, `:datetime`, `:zoneddatetime`, *maybe* `:zoneddate`, `:zonedtime`, `:timezone`
+
+Problem: Most users are likely to prefer date/time/datetime to zoneddate/zonedtime/zoneddatetime
+as the default formatting functions.
+Most "classical" time types are timestamps.
+User intentions might not be met by these names.
+
+Problem: Different platforms cannot agree on what to call a Floating Time Value: HTML and Java use `LocalXXX`,
+JavaScript has adopted `PlainXXX`,
+some others use different terms, such as `CivilXXX`.
```
eemeli, line 253: I don't understand the relevance of this. Isn't this purely an implementation concern, which has no visibility in the MF2 syntax? As I understand it, the idea is for the function names to describe the formatted output, not the input.

exploration/semantic-skeletons.md
aphillips: One such problem is the need to manage an input value's relationship to the timeline. While temporal types trying to pack all of the information into strongly typed objects, classical time values are usually timestamps (seconds or millis of epoch time). Formatting them requires a time zone (or at least an offset). Which means some _external_ way of expressing the zone when it is different from the runtime's default time zone.
```
Delivery time local to you and not this web server will be: {$timestamp :datetime timezone=$usersTimeZone}
```

There is also a need to be able to "float" a value: remove it from the timeline so that field values stay constant regardless of time zone. This involves _removing_ the time zone from the value, the equivalent of `ZonedDateTime.toLocalDateTime()` in Java:
```
Your birthday is: {$birthday :datetime offset=none} (option name made up on the spot)
```

exploration/semantic-skeletons.md
```diff
@@ -0,0 +1,266 @@
...
+#### Avoid 'picture strings'
+
+The MFWG early on considered including support for "picture strings" in the formatting of date/time values.
+There is a Working Group consensus **_not_** to support picture strings in Unicode MessageFormat, if possible.
+Many date/time formatting regimes provide for "picture strings".
+A "picture string" is a pattern using a microsyntax in which the user (developer, translator, UX designer)
+exactly specifies the desired format of the date/time value.
+In a picture string, separators, spaces, and other formatting are explicitly specified.
+This provides a lot of power to the developer or user experience designer, in terms of specifying formatting.
+For example: `MMM dd, yyyy` or `yyyy-dd-MM'T'HH:mm:ss`
```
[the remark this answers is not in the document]
aphillips: Fair enough. I moved the WG bits on top to make them visible, but this works.

exploration/semantic-skeletons.md
aphillips: They're not really "formatter options" though, are they? `{$timeOnlyValue :date}` == Bad Operand?

```suggestion
1. Date/time formatting functions and their options should be able to impose restrictions on acceptable values,
   such as formatting a month requiring a value that includes the month 
   or formatting an hour requiring requiring a time.
1. Date/time formatting functions and their options should be able to emit errors 
   when trying to format fields that cannot be computed from the operand value
   such as the "month" of a time-only value (`{|03:45Z| :date format=YMD}`)
   or the "hour" of a date-only value (`{|2025-04-23| :time format=HM}`).
```

exploration/semantic-skeletons.md
aphillips: A reasonable approach would be to name the functions consistent with usage. The point here is that there is not agreement on what to call these.

exploration/semantic-skeletons.md
```diff
@@ -31,6 +31,10 @@ using a microsyntax derived from familiar 'picture strings' (see below)
 combined with code in ICU (`DateTimePatternGenerator`) to produce the desired date/time value format.
 
 `Intl.DateTimeFormat` provided options to provide a similar capability.
+These weren't "skeletons" from the point of view that they didn't use a dedicated microsyntax
+similar to 'picture strings',
+but the effect was the same:
+users specified which fields they wanted with which display options (such as width).
```
eemeli, line 37: `Intl.DateTimeFormat` is still very much a thing.

```suggestion
`Intl.DateTimeFormat` provides options to provide a similar capability.
These aren't "skeletons" from the point of view that they don't use a dedicated microsyntax
similar to 'picture strings',
but the effect is the same:
users specify which fields they want with which display options (such as width).
```

exploration/semantic-skeletons.md
```diff
@@ -129,6 +134,47 @@ selecting locale-appropriate separators,
...
 
+### FAQ
+
+This section considers some potential arguments against the design
+or captures frequently asked questions about it.
+
+**What if semantic skeletons doesn't support the format I want?
+Unlike picture strings or classical skeletons, semantic skeletons do not allow unrestricted
+composition of date/time formats.**
+
```
eemeli, line 145:
```suggestion
**What if semantic skeletons don't support the format I want?**

Unlike picture strings or classical skeletons,
semantic skeletons do not allow unrestricted composition of date/time formats.

```

exploration/semantic-skeletons.md
```diff
@@ -129,6 +134,47 @@ selecting locale-appropriate separators,
...
+such as "11 PM April" (`jjMMMM` or `HHaMMMM`) or "2 2025" (`dyyyy`).
+
+**What about specialized formats, such as ISO8601?**
+
+These should be provided via other means that requiring a specialized pattern 
+and the (optional) `@locale` attribute. Do you really want to support this,
+given that it can then be used for other things:
+```
+{$now :datetime pattern=|yyyy-MM-dd'T'HH:mm:ss.sssz| @locale=und timezone=UTC}
+```
```
eemeli, line 176:
````suggestion
These should be provided via other means than requiring a specialized pattern 
and the (optional) `u:locale` option.
Do you really want to support this,
given that it can then be used for other things:
```
{$now :datetime pattern=|yyyy-MM-dd'T'HH:mm:ss.sssz| u:locale=und timezone=UTC}
```
````

File: exploration/semantic-skeletons.md (hunk 21 above)
Comment (eemeli) [reply exploration/semantic-skeletons.md:299]: I still don't understand how naming concerns about an input value type would transmit to the syntax. For example, I would expect an expression for a time with hours and minutes to look something like
```
{$t :datetime hour=numeric minute=numeric}
```
or maybe
```
{$t :time fields=hm}
```
irrespective of whether `$t` held a floating time value, or one with a timezone attached, or any other representation of the value to be formatted.

Or do you think we ought to have different functions for each input datetime type supported by an implementation?

File: exploration/semantic-skeletons.md (hunk 19 above)
Comment (eemeli) [reply exploration/semantic-skeletons.md:210]: Your suggested second requirement smells like scope creep to me. I understand this design doc to be about function options, not about enabling or restricting a function handler from emitting errors.

We don't actually impose any restriction on when a function handler can emit an error (just on what errors may be emitted when), we already allow an implementation with an options-basket or even classical-skeleton option style to reject option combinations that could not be expressed as a semantic skeleton.

This design doc should be focused on the ergonomics and readability of the different options styles, and not on implementation details.

File: exploration/semantic-skeletons.md (hunk 21 above)
Comment (sffc) [reply exploration/semantic-skeletons.md:299]: By "type safe" I mean that there should be well-defined semantics around when implementations should emit a Bad Operand error. For example, if you pass an implementation-specific "time-only" type to `:date`, you should get a Bad Operand error.

File: exploration/semantic-skeletons.md [hunk 26]
```diff
@@ -0,0 +1,442 @@
...
+- Options focused on field widths
+- Self-documenting: obvious what fields are shown
+- Reserves options for field width
+
+_Cons_
+- _Nineteen_ (or so) functions
+- Names appear generative, but aren't actually
+- Options might be baroque. Some option values might not be compatible with one another.
+  - Expression-wide options (short/medium/long/full) and field-level options might both be needed together
+
```
eemeli, line 437: For example:
```js
new Intl.DateTimeFormat('fi', { month: 'long' }).format()
// 'toukokuu'

new Intl.DateTimeFormat('fi', { day: 'numeric', month: 'long' }).format()
// '2. toukokuuta'
```

```suggestion
- Leads to localization failures similar to picture strings,
  as combinations of standalone fields are formatted incorrectly in inflected languages.
```

exploration/semantic-skeletons.md
```diff
@@ -0,0 +1,442 @@
...
+   | `yMMMMd` | April 6, 2025 |
+1. Developers, translators, and UI designers should not have to learn
+   multiple new microsyntaxes or multiple different sets of options for date/time value formatting.
+1. Any microsyntax or option set specified should be easy to understand only from the expression.
+1. Any microsyntax or option set specified should not _require_ translators to alter the values in most or all locales.
+1. User's should be able to specify the time zone or local time offset used in formatting
+   an incremental or timestamp value
+   because many date/time formatting processes are
+   running in a different environment/offset/zone from where the value will be seen
+   or are formatting a value for a specific offset or zone.
```
eemeli, line 321: I agree with being able to specify the timezone, either while constructing the formattable value, or possibly within the _options_, but I don't see the use case for explicit offset support. Could we drop that?
```suggestion
1. Users should be able to specify the timezone used in formatting
   an incremental or timestamp value
   because many date/time formatting processes are
   running in a different environment or timezone from where the value will be seen
   or are formatting a value for a specific timezone.
```

exploration/semantic-skeletons.md
aphillips: I don't understand this addition. I mean, I understand what you mean, but not why this is specific to this solution. Any skeleton solution should produce the correct output as a result of running the date/time pattern generator on the selected combination of fields and field widths. FWIW, all solutions can produce incorrect results if the information about the desired format is incomplete.

eemeli: My concern is about the affordances we provide. Something like
```
{$d :month}/{$d :day}
```
is much likelier to get actual use than
```
{$d :datetime month=numeric}/{$d :datetime day=numeric}
```
or
```
{$d :date fields=month}/{$d :date fields=day}
```
because it's much shorter, and doesn't immediately communicate to the reader how hacky it is. Meanwhile, the two latter choices hint much more that this isn't good prcatice.

aphillips: Okay, that's a good callout. I think we might make standalone fields less convenient across the board for similar reasons. We enable them, of course, but when you need them they should require more work?

exploration/semantic-skeletons.md
aphillips: As I said, it's not optimal to use offset. But offsets are common. If we made an optional option for them, that's less interoperable. Documentation should definitely discourage offsets vs. true zones.  But their prevalence is high enough that I think support might end up being necessary. Note that temporal has offset types.

An alternative would be using offset zones?  (I know... ick...)

We should start to put option sets together to compare, much as you've been doing in some comments. I can't do that from my phone.

exploration/semantic-skeletons.md
eemeli: I wouldn't say that we "make them less convenient", but more like we don't specifically make them excessively convenient, but this is semantics and I think we agree.

exploration/semantic-skeletons.md
eemeli: Offset zones aren't a full solution, because they're only available with integer hour values.

Regarding "necessary", note that AFAIK no extant message formatting syntax allows for setting the timezone or offset within the message syntax, and that all message formatting happens in the context of some programming language environment. In other words, we are here introducing a novel capability, and none of this is strictly speaking necessary -- it would be perfectly valid for us to rely solely on the timezone or offset to be set in the code that sets the datetime variable value that the formatter is called with.

Also, we don't need to introduce all capabilities at once. If we leave out something e.g. from the 48 release, it can be added in later. The reverse is not true.

aphillips: > offset zones aren't a full solution, because they're only available with integer hour values.

Offset zones are not restricted to integer hour values. The `Etc/GMT` zones built into systems are often integer or half-integer hour zones or such, but this is not a universal restriction. I'm not really talking about using constructed zones, in any case.

> note that AFAIK no extant message formatting syntax allows for setting the timezone or offset within the message syntax

Most message formatters require the caller to manage the time zone in code and there exist API interfaces. You might recall that I suggested that time zone be part of the formatting context. Lots of the problems that developers have with floating time values stem from the fact that there is poor control at the message level and developers have to cobble together solutions.

> Also, we don't need to introduce all capabilities at once. If we leave out something e.g. from the 48 release, it can be added in later. The reverse is not true.

My "necessary" sentence was fairly _suppositional_. I am leery of incomplete solutions because they lead to bad coding or formatting patterns. I agree that we can omit things that are not well-baked or for which the syntax and features are not yet well-understood (or are _suppositional_, if you prefer 🤣). I'd rather not deprecate things later/ever, if we can avoid it.

However, my point was that we have a good bit of experience formatting date/time values. In this document we're assembling requirements and trying to define the set of functions and options. It's good to at least look at the use cases--to put all of the utensils on the table, not just the fish forks and teaspoons--before deciding a design.

> To clarify, I'm not objecting at all to supporting the formatting of input values that include an offset, but explicitly setting an offset within an MF2 function call seems like a capability that ought to be left to implementations that choose to support it, rather than something to require of all implementations.

In any case, your original suggestion is about requirement text. I can see how we might make that clearer without the requirement requiring offsets be supported. Perhaps:

```suggestion
1. Message authors and developers should be able to manage 
   the formatting of date/time values such that
   processes running in an environment with a different time zone or local time offset
   from that needed to express or format the value
   can get the necessary results.
```

eemeli: That suggestion looks good; you'll probably need to re-apply it separately as the typo fix has made this thread "outdated".

exploration/semantic-skeletons.md
eemeli:
```suggestion
- Providing easy-to-use standalone fields may lead to localization failures similar to picture strings,
  as their combinations are formatted incorrectly in inflected languages.
```

aphillips: I'm going to modify this slightly to include the examples for posterity or future use in our own documentation.

```suggestion
- Providing easy-to-use standalone field functions 
  could lead to localization failures similar to picture strings,
  as naive users might use combinations of standalone functions to concatenate date/time values
  that are inappropriate in various other locales and require translator intervention.
> For example, this message hard codes separators, field order, and
> uses standalone representation of the fields:
> ```
> Today is {$d :month} {$d :day}, {$d :year} and it is {$d :hour}:{$d :minute} o'clock.
>```
```

aphillips: Also, note that this is an objection to **any** standalone formats and might be better located in requirements/use cases with a short con for this item (and other designs that bump up against it)

eemeli: Modified suggestion looks good; I edited it to indent the `>` block.

eemeli (approved):
This seems fine. The document still covers input datetime values and their details in way more detail than I understand to be relevant to the consideration of choosing between something like "semantic skeleton" and "field options".

Currently, my preferred option would be for us to end up with the functions `:date`, `:time`, and `:datetime`, each of which would format what they describe. This would mean that it would not be possible to use `:datetime` to output only a time or a date; both would always be present.

With this approach, we could have each of the following:
```
{$t :time precision=second}

{$t :date fields=month-day-weekday}

{$t :datetime dateFields=month-day-weekday timePrecision=second}
```

I find the single-letter abbreviations of the date fields mostly workable, except `E` standing for "weekday" is completely opaque. Using the fields' spelled-out names is longer, but would allow a translator to understand what they represent.

I don't think the inclusion or not of the timezone in the output merits separate `:zoned*` variants of the functions; that should instead be controlled by a `timeZoneName` option [as in JS](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat#timezonename). That option should be supported on `:datetime` and `:time`, but probably not on `:date`.

I think the inclusion (or not) of input override fields like `timeZone` and `offset` should be considered as a separate discussion from this one, which is focusing on how to represent the formatted output.

I don't think we should allow for a resolved `:date` value to be used as the operand of `:time` or `:datetime`, or a resolved `:time` value to be used as the operand of `:date` or `:datetime`, even if that might be technically possible in some implementations. I would, however, allow a resolved `:datetime` value to be used as the operand of `:date` or `:time`, with each only caring about the options relevant to it. This would allow for a message to use a single basket of options to e.g. apply an appropriate timezone, while avoiding ambiguities about what it means to format the `:date` of a `:time`.

sffc:
I agree with @eemeli's comment (😃) except I think we need more discussion on

> I don't think the inclusion or not of the timezone in the output merits separate :zoned* variants of the functions; that should instead be controlled by a timeZoneName option as in JS. That option should be supported on :datetime and :time, but probably not on :date.

I've stated previously that I don't understand the mental model for having :date, :time, and :datetime _but not_ :datetimezone (or :zoneddatetime). Just as a date is fundamentally different from a time, and a datetime is fundamentally different from a date or a time, a zoned datetime is fundamentally different from a date, time, or datetime. If we split date, time, and datetime into individual functions, why would we not do the same for the zoned versions?

eemeli:
> If we split date, time, and datetime into individual functions, why would we not do the same for the zoned versions?

I'd say it's because the timezone is a part of the time specifier. I do not see including or not the timezone as introducing a fundamental difference _in the formatted output_, though it's obviously a significant part of the input value that's being formatted.

sffc:
1. `May`
2. `May 15`
3. `May 15, 10 AM`
4. `May 15, 10:01 AM`
5. `May 15, 10:01 AM CEST`

Can you explain a mental model for why you draw a line between only 2 and 3?

eemeli:
Without a time, a date is describing a whole day or an even wider range. With a time, we're describing a specific instant. We speak of them differently.

So for example we don't think of "May 15" as implicitly meaning midnight or noon or any other time, but the whole day when multiple things could happen one after the other. Meanwhile, "May 15, 10 AM" conceptually expands to mean 10:00:00, a single moment in time. If it's not implicitly enough tied to a timezone, we might want to include that for disambiguation, but we're still speaking about a specific instant of time.

aphillips:
> From the MF WG meeting this week, the following design seems worth exploring:
> 
> ```
> # A date formatter
> {:datetime dateFields="YMD" }
> 
> # A time formatter
> {:datetime timePrecision="minute" }
> 
> # A datetime formatter
> {:datetime dateFields="YMD" timePrecision="minute" }
> 
> # A zoned datetime formatter
> {:datetime dateFields="YMD" timePrecision="minute" zoneStyle="generic" }
> ```

I will add this to the design document under the `:datetime` design option.

macchiati:
With and without the timezone does make a difference of some 26 hours.

1. If I write that X happened on 2025 May 15 CEST, that gives you a specific 24 hour interval in which it happened. (Near a daylight savings transition, can be 23 or 25 hours.)
2. If I write that X happened on 2025 May 15, unless you assume a particular timezone, it gives me a roughly 50 hour interval in which it happened, from the first moment on earth that had May 15 00:00:00 (Kiribati), to the last moment on earth that had May 15 11:59:59.999... (Baker Island)

PR #1068, opened 2025-04-07
Design document for percent formatting

aphillips:
This document now includes the proposed design discussed in the (poorly documented, sparsely attended) 2025-05-19 call. The emerging consensus appears to be:

- `:number`/`:integer` with option `style=percent` (this is the current design) with scaling
- `:unit` with unit `percent` is REQUIRED but other units are not required, this function has NO scaling

The most recent commit posits that `:number`/`:integer` select _after_ scaling because we don't support fraction selection currently.

> [!NOTE]
> All previous conversations were marked resolved on purpose and not because their content was in any way deficient, off topic, or necessarily addressed. Please comment on the proposed design.

### DIFF
--- a/exploration/percent-format.md
+++ b/exploration/percent-format.md
@@ -181,7 +181,7 @@ _Pros_
 
 _Cons_
 - `:unit` won't be REQUIRED, so percentage format will not be guaranteed across implementations.
-  Requiring `:unit type=percent` would be complicated at best.
+  Requiring `:unit unit=percent` would be complicated at best.
 - Implementation of `:unit` in its entirely requires significantly more data than implementation of
   percentage formatting.
 - More verbose placeholder
@@ -275,18 +275,23 @@ The value `0.5` formats as `50%`
 > Example.
 > ```
 > .local $pctSaved = {50}
-> {$pctSaved :percent}
+> {{{$pctSaved :percent}}}
 > ```
 > Prints as `5,000%`.
 
 #### Optional Scaling
-Implementation automatically does (or does not) scale.
-There is an option to switch to the other behavior.
+Function automatically does (or does not) scale,
+but there is an option to switch to the non-default behavior.
+Such an option might be:
+- An option with a name like `scaling` with boolean-like values `true` and `false`
+- An option with a name like `scale` with 
+  digit size option values
+  limited to a small set of supported values (possibly only `1` and `100`)
 
 > Example. Note that `scale=false` is only to demonstrate switching.
 >```
 > .local $pctSaved = {50}
-> {$pctSaved :percent} {$pctSaved :percent scale=false}
+> {{{$pctSaved :percent} {$pctSaved :percent scale=false}}}
 >```
 > Prints as `5,000% 50%` if `:percent` is autoscaling by default
 
@@ -300,7 +305,7 @@ Extension of `:math` to support other mathematical capabilities would allow for 
 >```
 > .local $pctSaved = {0.5}
 > .local $pctScaled = {$pctSaved :math exp=2}
-> {$pctSaved :percent} {$pctScaled :unit unit=percent}
+> {{{$pctSaved :percent} {$pctScaled :unit unit=percent}}}
 >```
 > Prints as `50% 50%` if `:percent` is autoscaling by default and `:unit` is not.
 
--- a/exploration/percent-format.md
+++ b/exploration/percent-format.md
@@ -137,11 +137,11 @@ You saved {$savings :unit unit=percent} on your order today!
 
 #### Use `:number`/`:integer` with `type=percent`
 
-Use the existing functions for number formatting with a separate `type` option for `percent`.
+Use the existing functions for number formatting with a separate `style` option for `percent`.
 (This was previously the design)
 
 ```
-You saved {$savings :number type=percent} on your order today!
+You saved {$savings :number style=percent} on your order today!
 ```
 
 **Pros**
--- a/exploration/percent-format.md
+++ b/exploration/percent-format.md
@@ -50,8 +50,9 @@ Developers need to know which behavior will occur so that they can adjust the va
 > - `Intl.NumberFormat(locale, { style: 'percent' })` scales
 > - `Intl.NumberFormat(locale, { style: 'unit', unit: 'percent' })` does not scale
 
-It is also possible for Unicode MessageFormat to provide support for scaling in the message itself,
-perhaps by extending the `:math` function.
+It is also possible for Unicode MessageFormat to provide support for scaling in the message itself.
+Since we've removed the `:math` function (at least for now), this would have to be through either
+the re-introduction of `:math` or through a specialized scaling function.
 
 An addition concern is whether to add a dedicated `:percent` function,
 use one of the existing number-formatting functions `:number` and `:integer` with an option `type=percent`,
@@ -295,13 +296,15 @@ Such an option might be:
 >```
 > Prints as `5,000% 50%` if `:percent` is autoscaling by default
 
-#### Provide scaling via additions to `:math`
+#### Provide scaling via a function
 Regardless of the scaling done by the percent formatting function, 
 there might need to be an in-message mechanism for scaling/descaling values.
-The (currently DRAFT) function `:math` was added to support offsets in number matching/formatting.
-Extension of `:math` to support other mathematical capabilities would allow for scaling.
+The function `:math` was originally proposed to support offsets in number matching/formatting,
+although the WG removed this proposal and replaced with with an `:offset` function in May 2025.
+Reintroducing `:math` to support scaling
+or the proposal of a new function dedicated to scaling might address the need for value adjustment.
 
-> Example. 
+> Example using `:math` as a placeholder function name
 >```
 > .local $pctSaved = {0.5}
 > .local $pctScaled = {$pctSaved :math exp=2}
@@ -322,9 +325,9 @@ _Cons_
   instability into the message regime as new options are introduced over time.
   Compare with `java.lang.Math`
 
-Two proposals exist for using `:math`:
+Two proposals exist for `:math`-like scaling:
 
-##### Use `:math exp` to scale
+##### Use `:math exp` (`:exp`??) to scale
 Provide functionality to scale numbers with integer powers of 10 using the `:math` function.
 
 Examples using `:unit`, each of which would format as "Completion: 50%.":
@@ -344,7 +347,7 @@ _Cons_
 - Cannot use _digit size option_ as the `exp` option value type, since negative exponents are a Thing
 
 
-##### Use `:math multiply` to scale
+##### Use `:math multiply` (`:multiply`??) to scale
 Provide arbitrary integer multiplication functionality using the `:math` function.
 
 Examples using `:unit`, each of which would format as "Completion: 50%.":
--- a/exploration/percent-format.md
+++ b/exploration/percent-format.md
@@ -197,7 +197,7 @@ You saved {$savings :scaled per=100} on your order today!
 #### No Scaling
 User has to scale the number. The value `0.5` formats as `0.5%`
 
-#### Scaling
+#### Always Scale
 Implementation always scales the number. The value `0.5` formats as `50%`
 
 #### Optional Scaling
--- a/exploration/percent-format.md
+++ b/exploration/percent-format.md
@@ -58,6 +58,36 @@ use one of the existing number-formatting functions `:number` and `:integer` wit
 or use the proposed _optional_ function `:unit` with an option `unit=percent`.
 Combinations of these approached might also be used.
 
+### Unit Scaling
+
+There is a difference between _input_ scaling and _output_ scaling in `MeasureFormat`,
+which is the model for the `:unit` function in Unicode MessageFormat.
+
+For example, an input of <3.5, `meter`> with `meter` as the output unit doesn't scale.
+
+If one supplies <0.35 `percent`> as the input and the output unit were `percent`, 
+`MeasureFormat` would format as 0.35%. 
+Just like `meter` ==> `meter` doesn't scale.
+
+However, if one supplies a different input unit, then percent does scale 
+(just like `meter` ==> `foot`). 
+The base unit is for such dimensionless units is 'part'.
+In MF, a bare number literal, such as `.local $foo = {35}`
+or an implementation-specific number type (such as an `int` in Java)
+might be considered to use the input unit of `part`
+unless we specified that the `percent` unit value or `:percent` function overrode the `part` unit with `percent`.
+ 
+With <0.35 `part`> as the input and the output unit of `percent`, the format is "35%".
+
+| Amount | Input Unit | Formatted Value with... | Unit |
+|---|---|---|---|
+| 0.35 | part | 0.35 | part |
+| 0.35 | part | 35.0 | percent |
+| 0.35 | part | 350.0 | permille |
+| 0.35 | part | 3500.0 | permyriad |
+| 0.35 | part | 350000.0 | part-per-1e6 |
+| 0.35 | part | 3.5E8 | part-per-1e9 |
+
 ## Use-Cases
 
 _What use-cases do we see? Ideally, quote concrete examples._
--- a/exploration/percent-format.md
+++ b/exploration/percent-format.md
@@ -71,7 +71,7 @@ Just like `meter` ==> `meter` doesn't scale.
 
 However, if one supplies a different input unit, then percent does scale 
 (just like `meter` ==> `foot`). 
-The base unit is for such dimensionless units is 'part'.
+The base unit for such dimensionless units is called 'part'.
 In MF, a bare number literal, such as `.local $foo = {35}`
 or an implementation-specific number type (such as an `int` in Java)
 might be considered to use the input unit of `part`
--- a/exploration/percent-format.md
+++ b/exploration/percent-format.md
@@ -204,7 +204,28 @@ Implementation always scales the number. The value `0.5` formats as `50%`
 Implementation automatically does (or does not) scale.
 There is an option to switch to the other behavior.
 
-#### Use `:math exp` to scale
+#### Provide scaling via additions to `:math`
+Regardless of the scaling done by the percent formatting function, 
+there might need to be an in-message mechanism for scaling/descaling values.
+The (currently DRAFT) function `:math` was added to support offsets in number matching/formatting.
+Extension of `:math` to support other mathematical capabilities would allow for scaling.
+
+**Pros**
+- Users may find utility in performing math transforms in messages rather than in business logic.
+- Should be easy to implement, given that basic math functionality is common
+ 
+**Cons**
+- Implementation burden, especially when providing generic mathematical operations
+- Designs should be generic and extensible, not tied to short term needs of a given formatter.
+- Potential for abuse and misuse is higher.
+- "Real" math utilities or classes tend to have a long list of functions with many capabilities.
+  A complete implementation would require a lot of design work and effort or introduce
+  instability into the message regime as new options are introduced over time.
+  Compare with `java.lang.Math`
+
+Two proposals exist:
+
+##### Use `:math exp` to scale
 Provide functionality to scale numbers with integer powers of 10 using the `:math` function.
 
 Examples using `:unit`, each of which would format as "Completion: 50%.":
@@ -216,7 +237,7 @@ Examples using `:unit`, each of which would format as "Completion: 50%.":
 {{Completion: {$n :unit unit=percent}.}}
 ```
 
-#### Use `:math multiply` to scale
+##### Use `:math multiply` to scale
 Provide arbitrary integer multiplication functionality using the `:math` function.
 
 Examples using `:unit`, each of which would format as "Completion: 50%.":
--- a/exploration/percent-format.md
+++ b/exploration/percent-format.md
@@ -264,7 +264,7 @@ The value `0.5` formats as `0.5%`
 > Example.
 > ```
 > .local $pctSaved = {50}
-> {$pctSaved :percent}
+> {{{$pctSaved :percent}}}
 > ```
 > Prints as `50%`.
 
--- a/exploration/percent-format.md
+++ b/exploration/percent-format.md
@@ -7,6 +7,7 @@ Status: **Proposed**
 	<dl>
 		<dt>Contributors</dt>
 		<dd>@aphillips</dd>
+		<dd>@eemeli</dd>
 		<dt>First proposed</dt>
 		<dd>2025-04-07</dd>
 		<dt>Pull Requests</dt>
@@ -135,30 +136,7 @@ _What prior decisions and existing conditions limit the possible design?_
 
 _Describe the proposed solution. Consider syntax, formatting, errors, registry, tooling, interchange._
 
-Support the formatting of percent values as follows:
-
-- REQUIRE the `:unit` function for all implementations
-  - Only specific `unit` option values are required, initially the unit `percent`.
-  - The function `:unit unit=percent` does not scale the operand, e.g. `{5 :unit unit=percent}` formats as `5%`.
-- REQUIRE the `:number` and `:integer` functions to support `style=percent` as an option
-  - The functions `:number` and `:integer` scale the operand, e.g. `{5 :integer style=percent}` formats as `500%`.
-    Note that the selector selects on the scaled value
-    (selectors currently cannot select fractional parts)
-
-> Examples. These are equivalent **except** that `:unit` does NOT scale.
->```
-> {{You have {$pct :number style=percent} remaining.}}
-> {{You have {$pct :unit unit=percent} remaining.}}
-> {{You have {$pct :integer style=percent} remaining.}}
->```
-> Selector example:
->```
-> .local $pct = {0.05 :number style=percent}
-> .match $pct
-> 5   {{This pattern is selected}}
-> one {{You have {$pct} left.}}
-> *   {{You have {$pct} left.}}
->``` 
+TBD
 
 ## Alternatives Considered
 
@@ -387,3 +365,32 @@ _Pros_
 
 _Cons_
 - Brings in multiplication
+
+---
+
+### Why not both?
+
+Rather than choosing only one option, choose multiple parallel solutions:
+
+- REQUIRE the `:unit` function for all implementations
+  - Only specific `unit` option values are required, initially the unit `percent`.
+  - The function `:unit unit=percent` does not scale the operand, e.g. `{5 :unit unit=percent}` formats as `5%`.
+- REQUIRE the `:number` and `:integer` functions to support `style=percent` as an option
+  - The functions `:number` and `:integer` scale the operand, e.g. `{5 :integer style=percent}` formats as `500%`.
+    Note that the selector selects on the scaled value
+    (selectors currently cannot select fractional parts)
+
+> Examples. These are equivalent **except** that `:unit` does NOT scale.
+>```
+> {{You have {$pct :number style=percent} remaining.}}
+> {{You have {$pct :unit unit=percent} remaining.}}
+> {{You have {$pct :integer style=percent} remaining.}}
+>```
+> Selector example:
+>```
+> .local $pct = {0.05 :number style=percent}
+> .match $pct
+> 5   {{This pattern is selected}}
+> one {{You have {$pct} left.}}
+> *   {{You have {$pct} left.}}
+>``` 
--- a/exploration/percent-format.md
+++ b/exploration/percent-format.md
@@ -94,7 +94,7 @@ _Describe the proposed solution. Consider syntax, formatting, errors, registry, 
 - Use a dedicated function `:percent` that scales by default.
 - Provide an option `scaling` with values `true` and `false` and defaulting to `true`.
 - Provide all options identical to `:number` _except_ that `select` does not provide `ordinal` value.
-- Allow `unit=percent` in `:unit` that is identical to `:percent` in formatting performance,
+- Allow `unit=percent` in `:unit` that is identical to `:percent` in formatting capabilities,
   for compatibility with CLDR units,
   but document that this usage is not preferred.
 
--- a/exploration/percent-format.md
+++ b/exploration/percent-format.md
@@ -142,15 +142,25 @@ although this does not have to be the default behavior of UMF's percent unit for
 You saved {$savings :unit unit=percent} on your order today!
 ```
 
-**Pros**
+The `:unit` alternative could also support other unit-like alternatives, such as
+_per-mille_ and _per-myriad_ formatting.
+It doesn't fit as cleanly with other notational variations left out of v47, such as
+compact notation (1000000 => 1M, 1000 => 1K),
+or scientific notation (1000000 => 1.0e6).
+
+_Pros_
 - Uses an existing set of functionality
+- Might provide a more consistent interface for formatting "number-like" values
 - Removes percentages from `:number` and `:integer`, making those functions more "pure"
 
-**Cons**
+_Cons_
 - `:unit` won't be REQUIRED, so percentage format will not be guaranteed across implementations.
   Requiring `:unit type=percent` would be complicated at best.
+- Implementation of `:unit` in its entirely requires significantly more data than implementation of
+  percentage formatting.
 - More verbose placeholder
-- Could require a separate scaling mechanism
+
+---
 
 #### Use `:number`/`:integer` with `type=percent`
 
@@ -161,11 +171,14 @@ Use the existing functions for number formatting with a separate `style` option 
 You saved {$savings :number style=percent} on your order today!
 ```
 
-**Pros**
+_Pros_
 - Consistent with ICU MessageFormat
 
-**Cons**
-- It's the only special case remaining in these functions. Why?
+_Cons_
+- It's the only special case remaining in these functions,
+  unless we also restore compact, scientific, and other notational variations.
+
+---
 
 #### Use a dedicated `:percent` function
 
@@ -179,13 +192,16 @@ You saved {$savings :percent} on your order today!
 > @sffc suggested that we should consider other names for `:percent`.
 > The name shown here could be considered a placeholder pending other suggestions.
 
-**Pros**
+_Pros_
 - Least verbose placeholder
-- Clear what the placeholder does; self-documenting?
+- Clear what the placeholder does; self-documenting
+- Consistent with separating `:currency`
 
-**Cons**
+_Cons_
 - Adds to a (growing) list of functions
 - Not "special enough" to warrant its own formatter?
+
+---
 
 #### Use a generic scaling function
 
@@ -197,14 +213,14 @@ You saved {$savings :dimensionless unit=percent} on your order today!
 You saved {$savings :scaled per=100} on your order today!
 ```
 
-**Pros**
+_Pros_
 - Could be used to support non-percent/non-permille scales that might exist in other cultures
 - Somewhat generic
 - Unlike currency or unit values, "per" units do not have to be stored with the value to prevent loss of fidelity,
   since the scaling is done to a plain old number.
   This would not apply if the values are not scaled.
 
-**Cons**
+_Cons_
 - Only percent and permille are backed with CLDR data and symbols.
   Other scales would impose an implementation burden.
 - More verbose. Might be harder for users to understand and use.
@@ -215,15 +231,29 @@ You saved {$savings :scaled per=100} on your order today!
 User has to scale the number. 
 The value `0.5` formats as `0.5%`
 
+> Example.
+> ```
+> .local $pctSaved = {50}
+> {$pctSaved :percent}
+> ```
+> Prints as `50%`.
+
 #### Always Scale
 Implementation always scales the number. 
 The value `0.5` formats as `50%`
+
+> Example.
+> ```
+> .local $pctSaved = {50}
+> {$pctSaved :percent}
+> ```
+> Prints as `5000%`.
 
 #### Optional Scaling
 Implementation automatically does (or does not) scale.
 There is an option to switch to the other behavior.
 
-> Example.
+> Example. Note that `scale=false` is only to demonstrate switching.
 >```
 > .local $pctSaved = {50}
 > {$pctSaved :percent} {$pctSaved :percent scale=false}
@@ -236,11 +266,19 @@ there might need to be an in-message mechanism for scaling/descaling values.
 The (currently DRAFT) function `:math` was added to support offsets in number matching/formatting.
 Extension of `:math` to support other mathematical capabilities would allow for scaling.
 
-**Pros**
+> Example. 
+>```
+> .local $pctSaved = {0.5}
+> .local $pctScaled = {$pctSaved :math exp=2}
+> {$pctSaved :percent} {$pctScaled :unit unit=percent}
+>```
+> Prints as `50% 50%` if `:percent` is autoscaling by default and `:unit` is not.
+
+_Pros_
 - Users may find utility in performing math transforms in messages rather than in business logic.
 - Should be easy to implement, given that basic math functionality is common
  
-**Cons**
+_Cons_
 - Implementation burden, especially when providing generic mathematical operations
 - Designs should be generic and extensible, not tied to short term needs of a given formatter.
 - Potential for abuse and misuse is higher.
@@ -263,6 +301,14 @@ Examples using `:unit`, each of which would format as "Completion: 50%.":
 {{Completion: {$n :unit unit=percent}.}}
 ```
 
+_Pros_
+- Avoids multiplication of random values
+- Useful for other scaling operations
+
+_Cons_
+- Might require changes to digit size options, since negative exponents are a Thing
+
+
 ##### Use `:math multiply` to scale
 Provide arbitrary integer multiplication functionality using the `:math` function.
 
@@ -274,3 +320,9 @@ Examples using `:unit`, each of which would format as "Completion: 50%.":
 .local $n = {0.5 :math multiply=100}
 {{Completion: {$n :unit unit=percent}.}}
 ```
+
+_Pros_
+- Can be used for other general purpose math
+
+_Cons_
+- Brings in multiplication
--- a/exploration/percent-format.md
+++ b/exploration/percent-format.md
@@ -327,7 +327,7 @@ although the WG removed this proposal and replaced with with an `:offset` functi
 Reintroducing `:math` to support scaling
 or the proposal of a new function dedicated to scaling might address the need for value adjustment.
 
-> Example using `:math` as a placeholder function name
+> Example using `:math` as a hypothetical function name
 >```
 > .local $pctSaved = {0.5}
 > .local $pctScaled = {$pctSaved :math exp=2}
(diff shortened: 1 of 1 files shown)


exploration/percent-format.md
```diff
@@ -0,0 +1,169 @@
...
+
+**Pros**
+- Consistent with ICU MessageFormat
+
+**Cons**
+- It's the only special case remaining in these functions. Why?
+
+#### Use a dedicated `:percent` function
+
+Use a new function `:percent` dedicated to percentages.
```
sffc, line 141: Should we consider names besides `:percent`?

The function could apply to all dimensionless units including permille, permillion, perbillion, etc.

For example: `{$var :dimensionless unit=permillion}`

exploration/percent-format.md
```diff
@@ -0,0 +1,169 @@
...
+- Adds to a (growing) list of functions
+- Not "special enough" to warrant its own formatter?
+
+### Scaling Alternatives
+
+#### No Scaling
+User has to scale the number. The value `0.5` formats as `0.5%`
+
+#### Scaling
+Implementation always scales the number. The value `0.5` formats as `50%`
```
sffc, line 161: If we go with the "always scales" approach, `:unit` can still be used to not scale.

sffc (approved):

exploration/percent-format.md
aphillips: We should consider other names. I'll add that option.

I'm not wild about `unit=percent` (mille, billion, etc. etc.). It's verbose and the other uses seems rare. Really only percent and permille are backed by CLDR data. The others strike me as special uses for unit or number formatting.

exploration/percent-format.md
```diff
@@ -0,0 +1,204 @@
...
+- Could require a scaling mechanism
+
+#### Use `:number`/`:integer` with `type=percent`
+
+Use the existing functions for number formatting with a separate `type` option for `percent`.
+(This was previously the design)
+
+```
+You saved {$savings :number type=percent} on your order today!
+```
```
eemeli, line 140: Previously, the option was `style`.

````suggestion
Use the existing functions for number formatting with a separate `style` option for `percent`.
(This was previously the design)

```
You saved {$savings :number style=percent} on your order today!
```
````

File: exploration/percent-format.md [hunk 4]
```diff
@@ -0,0 +1,204 @@
...
+- Only percent and permille are backed with CLDR data and symbols.
+  Other scales would impose an implementation burden.
+- More verbose. Might be harder for users to understand and use.
+
+### Scaling Alternatives
+
+#### No Scaling
+User has to scale the number. The value `0.5` formats as `0.5%`
+
+#### Scaling
```
eemeli, line 195:
```suggestion
#### Always Scale
```

exploration/percent-format.md
```diff
@@ -0,0 +1,204 @@
...
+Implementation always scales the number. The value `0.5` formats as `50%`
+
+#### Optional Scaling
+Implementation automatically does (or does not) scale.
+There is an option to switch to the other behavior.
```
eemeli, line 200:
```suggestion
Formatter automatically does (or does not) scale.
There is an option to switch to the other behavior.
The option here may be:
- An option `scaling` with boolean values `true` and `false`.
- An option `scale` with a small set of supported integer values, possibly only `1` and `100`.
```

exploration/percent-format.md
```diff
@@ -0,0 +1,204 @@
...
+
+#### Use `:math` to scale
+Provide functionality to scale numbers arbitrarily using the `:math` function.
+This alternative can be used with scaling/no scaling to fix the passed value appropriately without altering userland code.
```
eemeli, line 204:
````suggestion
#### Use `:math exp` to scale
Provide functionality to scale numbers with integer powers of 10 using the `:math` function.

Examples using `:unit`, each of which would format as "Completion: 50%.":
```
.local $n = {50}
{{Completion: {$n :unit unit=percent}.}}

.local $n = {0.5 :math exp=2}
{{Completion: {$n :unit unit=percent}.}}
```

#### Use `:math multiply` to scale
Provide arbitrary integer multiplication functionality using the `:math` function.

Examples using `:unit`, each of which would format as "Completion: 50%.":
```
.local $n = {50}
{{Completion: {$n :unit unit=percent}.}}

.local $n = {0.5 :math multiply=100}
{{Completion: {$n :unit unit=percent}.}}
```
````

File: exploration/percent-format.md [hunk 7]
```diff
@@ -0,0 +1,204 @@
...
+## Proposed Design
+
+_Describe the proposed solution. Consider syntax, formatting, errors, registry, tooling, interchange._
+
+- Use a dedicated function `:percent` that scales by default.
+- Provide an option `scaling` with values `true` and `false` and defaulting to `true`.
+- Provide all options identical to `:number` _except_ that `select` does not provide `ordinal` value.
+- Allow `unit=percent` in `:unit` that is identical to `:percent` in formatting performance,
+  for compatibility with CLDR units,
+  but document that this usage is not preferred.
```
eemeli, line 94: Would this mean that `:unit unit=percent` would or would not apply scaling? And why do we need or benefit from compatibility with CLDR units here?

eemeli (changes requested):
A first pass. I don't agree with the currently proposed design, but let's first get the available options better presented.

It would be nice to have more example code for the options.

exploration/percent-format.md
aphillips: > why do we need or benefit from compatibility with CLDR units here?

We don't have to be compatible, except that currently the definition of the `unit` option values is completely delegated to the unit identifiers found [here](https://www.unicode.org/reports/tr35/tr35-general.html#unit-identifiers) in TR35. It would be unfortunate to say "unit identifiers except this one specific one"

> Would this mean that :unit unit=percent would or would not apply scaling?

That's a good question. In the most recent WG discussion, there was a sentiment that we should make them behave identically to avoid confusion. There's an equal sentiment that they should be opposite each other (for utility). Here I'm trying to express equivalent performance without binding to a specific scaling/not-scaling choice (since that is separate).

exploration/percent-format.md
```diff
@@ -0,0 +1,229 @@
...
+**Pros**
+- Could be used to support non-percent/non-permille scales that might exist in other cultures
+- Somewhat generic
+- Unlike currency or unit values, "per" units do not have to be stored with the value to prevent loss of fidelity,
+  since the scaling is done to a plain old number.
+  This would not apply if the values are not scaled.
+
+**Cons**
+- Only percent and permille are backed with CLDR data and symbols.
+  Other scales would impose an implementation burden.
```
sffc, line 192: CLDR has data for other scales, too, via `portion` units.

exploration/percent-format.md
```diff
@@ -0,0 +1,229 @@
...
+Examples using `:unit`, each of which would format as "Completion: 50%.":
+```
+.local $n = {50}
+{{Completion: {$n :unit unit=percent}.}}
+
+.local $n = {0.5 :math exp=2}
+{{Completion: {$n :unit unit=percent}.}}
+```
+
+#### Use `:math multiply` to scale
```
sffc, line 219: Please note my concern about implementation burden due to having to support a more general function than we actually need.

exploration/percent-format.md
aphillips: That's not really percent/per mille type scaling though, is it?

exploration/percent-format.md
aphillips: The `:math` function, currently in draft and originally included to support plural offsets from MF1, is certainly a potential "slippery slope". 

Many programming languages have math-related classes or function sets with many different operators in them. The existence of a `:math` function in MF would certainly invite proposals for many of these to migrate into messages, regardless of utility. This in a specification that is strongly **_UN_**typed. 

If we go down the `:math` route, I would suggest that we write a full design document, including considerations for what our policies would be about future expansion. We should also consider whether `math` is the right name or different design strategies, such as unbundling functionality into separate functions (is it a better imposition of burden to have separate required `:add` and `:subtract` functions than a required function that has addition, subtraction, scaling, etc. into which we might add hard-to-achieve functionality? There is also the question of versioning the `:math` function if we add new operations to it over time, creating a portability hazard)

exploration/percent-format.md
bearfriend: Currently, my agreement with the proposal is dependent on `:unit unit=percent` _not_ scaling. So, I see specifying that here one way or another as important.

aphillips: >  So, I see specifying that here one way or another as important.

👍 It is imperative that we specify one or the other.

> Currently, my agreement with the proposal is dependent on :unit unit=percent not scaling.

The proposal is to make `:percent` and `:unit unit=percent` perform identically, so both would scale by default. Is your opposition to `:unit` scaling so that message writers could get access to both behaviors without having to use a `scale` option? Articulating your reasoning will help me improve the design doc to include that as a design we considered (and perhaps sway consensus).

exploration/percent-format.md
```diff
@@ -0,0 +1,250 @@
...
+Use the existing functions for number formatting with a separate `style` option for `percent`.
+(This was previously the design)
+
+```
+You saved {$savings :number style=percent} on your order today!
+```
+
+**Pros**
+- Consistent with ICU MessageFormat
+
```
macchiati, line 149:
```suggestion
- Widely view as a number format option in spreadsheets and other contexts, so simplest for people to understand.
```

exploration/percent-format.md
macchiati: There is a misunderstanding of unit scaling.

For unit formatting, CLDR has both an input unit and an output unit, where the output unit typically depends on the unit preferences. For example, <3.5, `meter`> input with `foot` output formats as "11.5 feet" (in English). There *is* scaling involved, in the conversion of 3.5 to 11.5. If there is no specified output unit, or the output unit is explicitly the same as the input unit, then there is no scaling. Thus:

<3.5, `meter`> input with `meter` output doesn't scale.

* If I supply <0.35 `percent`> as the input and the output unit were `percent`, it would format as 0.35%. Just like meter ==> meter doesn't scale.

However, if I supply the right input unit, then percent does scale (just like meter ==> foot). And the base unit is for such dimensionless units is 'part'.

With <0.35 `part`> as the input and the output unit of `percent`, the format is  "35%". 

Here are sample conversions that I just generated (no formatting)
```
0.35	part	0.35	part
0.35	part	35.0	percent
0.35	part	350.0	permille
0.35	part	3500.0	permyriad
0.35	part	350000.0	part-per-1e6
0.35	part	3.5E8	part-per-1e9
```

exploration/percent-format.md
aphillips: Perhaps debatable? 

It's certainly proximate to numeric formats, at least in some spreadsheets. FWIW, we do group it into the number _functions_ and it certainly takes a numeric operand. But I think a case can be made that `:number type=percent` or `:percent` are both intuitive--and the latter becomes maybe a bit more obvious given `:currency`. 

The meta debate we're having is a classic in the I18N space: split or lump? Should we prefer functions that do many things with lots of options? Or should we prefer functions that do roughly one thing with minimal options (and lots and lots of functions)?

---

Note that the "123" button is "More Formats" in Google sheets:

image

Excel puts percent after date/time:

image

macchiati: A fair number of the other Pros/Cons are debatable... But I'll tweak my suggested change.

I'm not wild about :percent as a separate function; nor wild about :scientific or :engineering or :compact or even :integer. Just the sheer volume of duplicated options gets to be very daunting.

exploration/percent-format.md
aphillips: I understood that to be the case.

`:unit` can override the unit, in which case scaling occurs. The question is what happens when there is no other unit? Using `MeasureFormat` in ICU4J can only be an approximation, since the only way to call it is with a `Measure` object. Presumably a bare number operand in MF would, behind the scenes, be packaged with the unit. 

I'm not suggesting that `:unit` does not convert. Only that the default behavior of `unit=percent` is unscaled given a numeric operand. This is different from MF1's handling of `operand,number,percent` formatting and the _proposed_ performance of `:percent`. Do you disagree?

exploration/percent-format.md
aphillips: > Just the sheer volume of duplicated options gets to be very daunting.

The worst of all worlds would be lots of functions each of which has lots of options and where some functions are general purpose and overlap with special purpose ones. With support for custom functions, that will sometimes be unavoidable. But for the default function set we should have a clear policy/design philosophy. The meta debate is, in many ways, _more important_, than the concrete decision of what to name the percent formatting function (but percent is as good a trial horse, I think, as we're going to get). Note that the discussion about semantic skeletons _also_ is considering the problem of function packaging.

macchiati:
I agree that unless there is a way to distinguish the source and target
unit, it wouldn't scale. I was just pointing that out for people unfamiliar
with the way that units work.

(And I'm not a fan of using :unit unit=percent at all to solve the problem
of producing "10%" from 0.1, for a number of reasons).

exploration/percent-format.md
bearfriend: > Is your opposition to :unit scaling so that message writers could get access to both behaviors without having to use a scale option?

Sorry I wasn't clear on that. Yes, I want them to act differently, so I guess that just isn't this proposal but an alternative. The reason, though, is not just to have access to both behaviors (though it's an excellent side benefit) but because it makes semantic/intuitive sense to me.

I see them as for different purposes, where the input value to `:percent` is for (or from) some computation which results in a ~number, and the input to `:unit` would be roughly a string (or semantically equivalent in its static intent, if that makes sense).

`1/10 = .1` -> format via `:percent` -> "You've completed 10% of your tasks"
vs
A user inputs into a marketing tool a discount value of "10" and selects "%" (as opposed to "$", "lbs", "items" etc.), and that ends up rendering things like "10% off", "$10 off", "Get 10 lbs free...", "Buy 10 get 1 free" or similar.

This is how I, presumptuously, think most people would expect each to work. Happy to be wrong about that, though.

I'm not terribly familiar with the input -> output scaling Mark mentioned, so I'll try to digest that a bit more and see if it changes my perspective. It doesn't initially seem problematic, though.

exploration/percent-format.md
```diff
@@ -0,0 +1,250 @@
...
+- maximum number of fractional digits
+- minimum number of decimal digits
+- group used (for very large percentages, i.e. > 999%)
+- etc.
+
+## Requirements
+
+_What properties does the solution have to manifest to enable the use-cases above?_
+
+
```
eemeli, line 84: I think it's important for us not to special-case percentage formatting too much. It might be useful to determine the extent of appropriate use cases for scaling. `:math` applies to numbers, but is (or should) scaling be restricted to only unit formatting?

```suggestion
## Requirements

_What properties does the solution have to manifest to enable the use-cases above?_

- As percentages are not the only units for which scaling may occur,
  the solution needs to be extensible or applicable for other units as well.
```

exploration/percent-format.md
aphillips: @bearfriend I think those are both great use cases and will add them to the document.

Note that the "proposed solution" is a strawman. The alternatives considered are what is important. We'll see if a consensus emerges--or vote on which technical decisions to make.

exploration/percent-format.md
eemeli: I was mostly thinking of `permille`, actually. A relevant question here that I'd like us to answer is whether we envision any valid _non-unit_ scaling. Do we want to allow for e.g. currency values expressed in integer cents?

If we can't point at a potential non-unit use for scaling, then we probably ought to restrict it to only units.

eemeli:
I just realised that this whole discussion is also related to https://github.com/unicode-org/message-format-wg/pull/1015#pullrequestreview-2621386507, which we probably ought to address as well.

In other words, as we currently don't have `:number notation`, we probably ought to figure out how we're going to do its style of scaling as well.

exploration/percent-format.md
aphillips: Actually, I think your proposed requirement only makes sense if we've decided that there needs to be a scale that applies to non-percents(milles/myriads). What it should say is more like:

```suggestion
## Requirements

_What properties does the solution have to manifest to enable the use-cases above?_

- **Be consistent** Any solution for scaling percentages should be a model for other, similar scaling operations,
  such as _per-mille_ or _per-myriad_,
  as well as other, non-percent or even non-unit scaling.
  This does not mean that a scaling mechanism or any particular scaling mechanism itself is a requirement.
```

exploration/percent-format.md
aphillips: @macchiati 
> There is a misunderstanding of unit scaling.

I added your example (suitably edited and expanded) at length to the document. Check for veracity.

catamorphism (approved):

exploration/percent-format.md
```diff
@@ -58,15 +58,47 @@ use one of the existing number-formatting functions `:number` and `:integer` wit
 or use the proposed _optional_ function `:unit` with an option `unit=percent`.
 Combinations of these approached might also be used.
 
+### Unit Scaling
+
+There is a difference between _input_ scaling and _output_ scaling in `MeasureFormat`,
+which is the model for the `:unit` function in Unicode MessageFormat.
```
eemeli, line 64: This is not the whole story, though, as almost all of the currently proposed `:unit` options are based on the JS `Intl.NumberFormat` options.

exploration/percent-format.md
```diff
@@ -58,15 +58,47 @@ use one of the existing number-formatting functions `:number` and `:integer` wit
...
+
+For example, an input of <3.5, `meter`> with `meter` as the output unit doesn't scale.
+
+If one supplies <0.35 `percent`> as the input and the output unit were `percent`, 
+`MeasureFormat` would format as 0.35%. 
+Just like `meter` ==> `meter` doesn't scale.
+
+However, if one supplies a different input unit, then percent does scale 
+(just like `meter` ==> `foot`).
```
eemeli, line 73: The `:unit` spec text doesn't actually say what happens if the `unit` is defined both in the _operand_ and in an option.

exploration/percent-format.md
```diff
@@ -58,15 +58,47 @@ use one of the existing number-formatting functions `:number` and `:integer` wit
...
+
+For example, an input of <3.5, `meter`> with `meter` as the output unit doesn't scale.
+
+If one supplies <0.35 `percent`> as the input and the output unit were `percent`, 
+`MeasureFormat` would format as 0.35%. 
+Just like `meter` ==> `meter` doesn't scale.
+
+However, if one supplies a different input unit, then percent does scale 
+(just like `meter` ==> `foot`). 
+The base unit is for such dimensionless units is 'part'.
```
eemeli, line 74:
```suggestion
The base unit for such dimensionless units is called 'part'.
```

exploration/percent-format.md
```diff
@@ -58,15 +58,47 @@ use one of the existing number-formatting functions `:number` and `:integer` wit
...
+In MF, a bare number literal, such as `.local $foo = {35}`
+or an implementation-specific number type (such as an `int` in Java)
+might be considered to use the input unit of `part`
+unless we specified that the `percent` unit value or `:percent` function overrode the `part` unit with `percent`.
```
eemeli, line 78: Does `MeasureFormat` or any other unit formatter implementation ever make such an assumption, of having a numerical input value not match the formatter's output units?

exploration/percent-format.md
```diff
@@ -125,15 +172,25 @@ although this does not have to be the default behavior of UMF's percent unit for
...
+The `:unit` alternative could also support other unit-like alternatives, such as
+_per-mille_ and _per-myriad_ formatting.
+It doesn't fit as cleanly with other notational variations left out of v47, such as
+compact notation (1000000 => 1M, 1000 => 1K),
+or scientific notation (1000000 => 1.0e6).
+
+_Pros_
 - Uses an existing set of functionality
+- Might provide a more consistent interface for formatting "number-like" values
 - Removes percentages from `:number` and `:integer`, making those functions more "pure"
```
eemeli, line 184: We already removed `:number style=percent`.
```suggestion
- Keeps percentage formatting out of `:number` and `:integer`, making those functions more "pure"
```

exploration/percent-format.md
```diff
@@ -162,14 +222,17 @@ You saved {$savings :percent} on your order today!
 > @sffc suggested that we should consider other names for `:percent`.
 > The name shown here could be considered a placeholder pending other suggestions.
 
-**Pros**
+_Pros_
 - Least verbose placeholder
-- Clear what the placeholder does; self-documenting?
+- Clear what the placeholder does; self-documenting
+- Consistent with separating `:currency`
```
eemeli, line 228: I don't see this consistency. If it applies to `:percent`, does that it also applies to `:permille`? How about `:meter` or `:radian`?

`:currency` makes sense to keep separate because currency formatting depends on currency codes, which in turn impact default fraction digits, and other presentation details. Nothing like that applies to percentage formatting.

exploration/percent-format.md
```diff
@@ -180,41 +243,72 @@ You saved {$savings :dimensionless unit=percent} on your order today!
...
-Implementation always scales the number. The value `0.5` formats as `50%`
+Implementation always scales the number. 
+The value `0.5` formats as `50%`
+
+> Example.
+> ```
+> .local $pctSaved = {50}
+> {$pctSaved :percent}
+> ```
+> Prints as `5000%`.
```
eemeli, line 280:
```suggestion
> Prints as `5,000%`.
```

exploration/percent-format.md
```diff
@@ -180,41 +243,72 @@ You saved {$savings :dimensionless unit=percent} on your order today!
...
 #### Optional Scaling
 Implementation automatically does (or does not) scale.
 There is an option to switch to the other behavior.
 
+> Example. Note that `scale=false` is only to demonstrate switching.
+>```
+> .local $pctSaved = {50}
+> {$pctSaved :percent} {$pctSaved :percent scale=false}
+>```
+> Prints as `5000% 50%` if `:percent` is autoscaling by default
```
eemeli, line 291:
```suggestion
> Prints as `5,000% 50%` if `:percent` is autoscaling by default
```

exploration/percent-format.md
```diff
@@ -237,6 +331,14 @@ Examples using `:unit`, each of which would format as "Completion: 50%.":
 {{Completion: {$n :unit unit=percent}.}}
 ```
 
+_Pros_
+- Avoids multiplication of random values
+- Useful for other scaling operations
+
+_Cons_
+- Might require changes to digit size options, since negative exponents are a Thing
```
eemeli, line 339: Rather, this would not be a _digit size option_.
```suggestion
- Cannot use _digit size option_ as the `exp` option value type, since negative exponents are a Thing
```

exploration/percent-format.md
aphillips: This is documenting `MeasureFormat`'s behavior. We should consider whether to extend `:unit`

exploration/percent-format.md
eemeli: I don't know about ICU, but CLDR does appear to treat `part` as a [unit](https://github.com/unicode-org/cldr/blob/faa2818490a34a2288f307720ab0abdab8c91ff7/common/validity/unit.xml#L200).

aphillips: Not exactly. There are various measures (such as `parts-per-million`, which is a flavor of `concentr-part` \[concentration\]) that have "parts", but not a measurement `part` all by itself. ICU represents the various measures in `MeasureUnit` and it doesn't have a standalone `PART` member.

Which is beside the point. The percent format in `NumberFormatter` is unscaled when working on a raw number.

exploration/percent-format.md
```diff
@@ -0,0 +1,358 @@
...
+> Prints as `50%`.
+
+#### Always Scale
+Implementation always scales the number. 
+The value `0.5` formats as `50%`
+
+> Example.
+> ```
+> .local $pctSaved = {50}
+> {$pctSaved :percent}
```
eemeli, line 278:
```suggestion
> {{{$pctSaved :percent}}}
```

exploration/percent-format.md
eemeli: I think this suggestion is still relevant.

exploration/percent-format.md
```diff
@@ -0,0 +1,358 @@
...
+or scientific notation (1000000 => 1.0e6).
+
+_Pros_
+- Uses an existing set of functionality
+- Might provide a more consistent interface for formatting "number-like" values
+- Keeps percentage formatting out of `:number` and `:integer`, making those functions more "pure"
+
+_Cons_
+- `:unit` won't be REQUIRED, so percentage format will not be guaranteed across implementations.
+  Requiring `:unit type=percent` would be complicated at best.
```
eemeli, line 184:
```suggestion
  Requiring `:unit unit=percent` would be complicated at best.
```

exploration/percent-format.md
```diff
@@ -0,0 +1,358 @@
...
+- Implementation of `:unit` in its entirely requires significantly more data than implementation of
+  percentage formatting.
```
eemeli, line 186: I think we're going to end up with a scale (heh) of unit support no matter what, as JS supports a [subset](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/supportedValuesOf#supported_unit_identifiers) of the CLDR units.

We're also going to make the `:unit usage` option OPTIONAL, possibly along with allowing but not reuiring fallback when operand and option `unit` values differ.

Given that, I would not object to making `:unit` as a whole REQUIRED with perhaps only `unit=percent` also REQUIRED, and all other option values as RECOMMENDED or OPTIONAL.

exploration/percent-format.md
```diff
@@ -0,0 +1,358 @@
...
+| 0.35 | part | 3.5E8 | part-per-1e9 |
+
+## Use-Cases
+
+_What use-cases do we see? Ideally, quote concrete examples._
+
+Developers wish to write messages that format a numeric value as a percentage in a locale-sensitive manner.
+
+The numeric value of the operand is not pre-scaled because it is the result of a computation, 
+e.g. `var savings = discount / price`.
```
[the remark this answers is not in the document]
aphillips: We could add other uses cases. A user story might be:

> - As a UX designer or developer, I want to create a message that correctly formats a percent value
>   that I have expressed as a whole number (i.e. pre-scaled) because that is the most natural way to
>   think about such values. For example, a message like:
>```
>Act now to receive a {5 :percent} discount on your next order!
>```

I agree that we *can* impose scaling on calculated values. The question, obviously, is whether we should.

I agree that MF1 compatibility is a consideration. Calling code will generally always have to change to call MF2, but the value marshaling part of that code is what should concern us.

exploration/percent-format.md
aphillips: I wrote "implementation", which isn't quite right, because "formatter" seemed too specific. Percent functions might also be selectors.

```suggestion
Function automatically does (or does not) scale,
but there is an option to switch to the non-default behavior.
Such an option might be:
- An option with a name like `scaling` with boolean-like values `true` and `false`
- An option with a name like `scale` with 
  digit size option values
  limited to a small set of supported values (possibly only `1` and `100`)
```

eemeli (approved):

exploration/percent-format.md
```diff
@@ -135,7 +135,30 @@ _What prior decisions and existing conditions limit the possible design?_
...
 _Describe the proposed solution. Consider syntax, formatting, errors, registry, tooling, interchange._
 
-(Awaiting WG consensus)
+Support the formatting of percent values as follows:
+
+- REQUIRE the `:unit` function for all implementations
+  - Only specific `unit` option values are required, initially the unit `percent`.
+  - The function `:unit unit=percent` does not scale the operand, e.g. `{5 :unit unit=percent}` formats as `5%`.
+- REQUIRE the `:number` and `:integer` functions to support `style=percent` as an option
+  - The functions `:number` and `:integer` scale the operand, e.g. `{5 :integer style=percent}` formats as `500%`.
```
eemeli, line 144: I find `:integer style=percent` to be confusing, and I don't think we should support it. Even though it may be well specified, it's not obvious to a reader whether the scaling happens before the rounding, or after it. We should avoid introducing unnecessary uncertainty, esp. as it's likely that implementations will choose to default to `maximumFractionDigits=0` for percent formatting.

Already now, the options supported by `:integer` are a subset of the options supported by `:number`, so this would only extend the set of excluded options.

```suggestion
- REQUIRE the `:number` function to support `style=percent` as an option
  - The function `:number` scales the operand, e.g. `{5 :number style=percent}` formats as `500%`.
```

exploration/percent-format.md
```diff
@@ -135,7 +135,30 @@ _What prior decisions and existing conditions limit the possible design?_
...
+    Note that the selector selects on the scaled value
+    (selectors currently cannot select fractional parts)
```
eemeli, line 146: Highlighting for @sffc and @ryzokuken that this may require supporting something like `style` as an Intl.PluralRules option, or being ok with Intl.MessageFormat offering a capability beyond Intl.PluralRules.

Is plural category selection on percent-formatted values already well supported by the ICU libraries?

exploration/percent-format.md
```diff
@@ -135,7 +135,30 @@ _What prior decisions and existing conditions limit the possible design?_
...
+
+> Examples. These are equivalent **except** that `:unit` does NOT scale.
+>```
+> {{You have {$pct :number style=percent} remaining.}}
+> {{You have {$pct :unit unit=percent} remaining.}}
+> {{You have {$pct :integer style=percent} remaining.}}
+>```
```
eemeli, line 153:
```suggestion
> Examples. These are equivalent **except** that `:unit` does NOT scale.
>```
> {{You have {$pct :number style=percent} remaining.}}
> {{You have {$scaledPct :unit unit=percent} remaining.}}
>```
```

exploration/percent-format.md
eemeli: Also, if selection makes sense for `:number style=percent`, is there a reason to not allow selection for `:unit unit=percent` as well? And if the latter is fine, then why not allow selection for all of `:unit`.

My preference would be for us to not allow for selection with percent formatting at this time.

exploration/percent-format.md
```diff
@@ -0,0 +1,389 @@
...
+  - Any solution for formatting percentages should be a model for solving related problems with:
+    - per-mille
+    - per-myriad
+    - compact notation
+    - scientific notation
+    - (others??)
+
+## Constraints
+
+_What prior decisions and existing conditions limit the possible design?_
```
sffc, line 132: I think we should mention that since MF1 scales, we want MF2 to be able to scale?

sffc (changes requested):
I'm not happy with the Proposed Design: Let's not add the `style` option back to `:number`. It was productive when we removed it, and I don't want to go back. Among other things, it causes problems when mixed with other options, questionable plural selection behavior as @eemeli noted, and inconsistent with Intl.NumberFormat which uses `style` for currency and unit as well.

I also do not like requiring `:unit unit=percent`, because we want percent formatting to not pull in all unit formatting data, but this is unavoidable in this design for reasons the ICU4X TC has expressed previously.

exploration/percent-format.md
```diff
@@ -0,0 +1,389 @@
...
+- Clear what the placeholder does; self-documenting
+- Consistent with separating specialized formats from `:number`/`:integer`
+  as was done with `:currency`
+
+_Cons_
+- Adds to a (growing) list of functions
+- Not "special enough" to warrant its own formatter?
+- Unlike `:currency`, because currency formatting depends on currency codes,
+  which in turn impact default fraction digits, and other presentation details.
+  Nothing like that applies to percents.
```
sffc, line 256: I don't buy any of these cons.

1. There is no set of rules I've seen to dictate when or when not to add new functions. Apparently we are okay with `:integer` and `:number` being separate even though one is a superset of another. And we seem to have been okay with `:date`, `:time`, and `:datetime` being different. So I don't think "we don't want to grow the list of functions" is sufficient justification to not add this function.
2. This document is self-contradictory in percent formatting being "important enough". Apparently the author thinks that it is "important enough" to muddy `:number` by adding back the `style=percent` option, and that it is important enough to make `:unit unit=percent` be REQUIRED.
3. We have evidence of at least one option being different, a potential `scaling` option. But, besides this, I point back to `:integer` which is just a subset of `:number`.

exploration/percent-format.md
aphillips: Fair enough. In this commit I was trying to capture the WG telecon conversation and I understood it to include `:integer`.

I don't think `:integer style=percent` is confusing, but it does add more surface to number formatting. Do I suppose correctly that you'd oppose adding other stylistic options (compact, scientific, etc.) to `:integer` as well?

exploration/percent-format.md
aphillips: > this may require supporting something like style as an Intl.PluralRules option

I don't think this necessarily follows? The function handler would scale the value before using plural rules, just as the function handler calls `Intl.NumberFormat` to format the number later.

> My preference would be for us to not allow for selection with percent formatting at this time.

This would require a separate placeholder just to perform message selection on what is fundamentally a number.

exploration/percent-format.md
```diff
@@ -0,0 +1,389 @@
+# Formatting Percent Values
+
+Status: **Proposed**
```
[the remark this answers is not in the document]
aphillips: I hope that we will choose a design before committing.

exploration/percent-format.md
aphillips: We don't necessarily _want_ MF2 to scale. There is a hot debate about whether scaling or non-scaling is preferred.

exploration/percent-format.md
aphillips: > There is no set of rules I've seen to dictate when or when not to add new functions. Apparently we are okay with :integer and :number being separate even though one is a superset of another. And we seem to have been okay with :date, :time, and :datetime being different. So I don't think "we don't want to grow the list of functions" is sufficient justification to not add this function.

In the call (and elsewhere), we discussed that `:integer` is a convenience function for `:number` (specifically a shorthand for `{$n :number maximumFractionDigits=0}`. The question is whether we are "splitters" or "lumpers" for functions, that is, whether we prefer a single function with lots of options or many functions with few(er) options. 

I called out in the call that there is no rhyme nor reason to our current policy.

> Apparently the author thinks that it is "important enough" to muddy :number by adding back the style=percent option, and that it is important enough to make :unit unit=percent be REQUIRED.

The author is recording what the WG wanted recorded. (In fact, the author does not agree with this design)

exploration/percent-format.md
eemeli: Probably, yes. I don't want to commit to the position prematurely, but it'd be my starting point to be a bit queasy about needing to think about whether `4.2e2` is really an "integer" or not.

exploration/percent-format.md
eemeli: > > this may require supporting something like style as an Intl.PluralRules option
> 
> I don't think this necessarily follows? The function handler would scale the value before using plural rules, just as the function handler calls `Intl.NumberFormat` to format the number later.

Quite recently in tc39/ecma402#989 we added `notation` as one of the `Intl.PluralRules` options. It would therefore be a bit surprising to effectively support `style=percent` for selection scaling in Intl.MessageFormat, but not for `Intl.PluralRules`.

> > My preference would be for us to not allow for selection with percent formatting at this time.
> 
> This would require a separate placeholder just to perform message selection on what is fundamentally a number.

Does your position generalise to preferring selection to work on all `:unit` values?

aphillips:
(chair hat **OFF**)

Adding `style=percent` to `:number` would not prevent us from having a convenience function `:percent` later (or now). This suggests that we prefer a "lumping" vs. "splitting" design for functions, with at least one function with "all the options", possibly surrounded by a host of convenience functions. This is to some degree what we have now, e.g.:
- `:number` with `:integer` and possibly `:percent` and other later functions as convenience
- `:datetime` with `:date`, `:time` (and as many as 16 more) functions as convenience

I'm not opposed to adding convenience functions. I think `:integer` was exactly the right thing to do for our users.

Alternatively we could regard `:percent` as a separate function handler (including with the special selection logic related to scaling) with targeted options. This suggests that we might add other specialized functions in the future, although it doesn't require it. We also have this as a model:
- `:number` has `:currency` as a (draft) friend. `:number` cannot format _currency_ values (it can format a number from a currency, but not as a currency)

I think `:unit unit=percent` is a red herring. It exists because carving out percent from CLDR units is janky (especially given that per-mille and per-myriad exist). My guess is that no one will use it when there's a big shiny function `:percent` available. Having _two_ baroque means (`{$p :number style=percent}` and `{$p :unit unit=percent}`) for formatting percentages is just weird: we have a common-enough use case and two different, equally-inconvenient means of formatting it?

My preferred solution is:
- add `:percent`
- do NOT add `:number style=percent`
- do NOT require `:unit unit=percent` (but permit it)

(chair hat **ON**)

I observe that we're not closing on a design. If we do not achieve consensus on a design in the next (2025-06-02) call, I will call for a ballot.

exploration/percent-format.md
aphillips: > Does your position generalise to preferring selection to work on all :unit values?

I'm not sure. I think it might, at least in terms of exact selection. Plural rules don't work well/aren't needed (what's actually wanted is more like inflection). But there are plenty of cases where one might want a specific value to produce a specific message, with `0` being the most common case. 

A lot of unit selection cases actually look like what `ChoiceFormat` is good at: separating messages above/below a given threshold, e.g. switching from kilometers to meters (or miles to feet) in driving directions at some cutoff.

eemeli:
I talked with @sffc while we met in person this past week, and it came up that ICU4X would almost certainly want unit formatting to be separated by category. As in, when formatting a value that includes its own unit (say, `kilometer`), the expression would need to define at least `length` as the category of supported units. The rationale here is to limit the data loading that would be required for the formatter before it can tell exactly which unit it'll be formatting.

If that is a requirement that we accept, then it suggests to me that we ought to include the category in the function name, so we'd have e.g. `:unit:length`, `:unit:volume` and so on, rather than a catch-all `:unit`. With such an approach, we ought to consider a dedicated `:unit:percent`, and if so, promote its use rather than adding a `style=percent` on `:number`.

aphillips:
@eemeli suggested:
> If that is a requirement that we accept, then it suggests to me that we ought to include the category in the function name, so we'd have e.g. :unit:length, :unit:volume and so on, rather than a catch-all :unit. With such an approach, we ought to consider a dedicated :unit:percent, and if so, promote its use rather than adding a style=percent on :number.

So `:unit` would be a namespace? We don't permit nested namespaces, so that would limit implementation-specific extension. Perhaps use an unreserved sigil as a separator instead, e.g. `:unit-length`, `:unit-volume`, `:unit-percent`. Only, once we do that, the `unit-` part starts to look superfluous. What's the difference between `:unit-percent` and `:percent`? Similarly, why type `:unit-length` instead of `:length`?

The "requirement" is really a dodge around creating separate functions for each unit or around creating a single function whose data loading depends on an option (or on the operand value). Most implementations bind the unit data late, but we should allow for ICU4X and its need/desire to bind the data early.

sffc:
The current design of `:unit` is not implementable with ICU4X's data design for reasons laid out in https://github.com/unicode-org/message-format-wg/issues/1006. We did not take too close of a look at `:unit` when writing that doc because `:unit` was not marked as being required, but it suffers from many of the same problems as, for example, `u:locale`. As a result, requiring "part" of `:unit` for percent formatting is not feasible.

@eemeli's suggestion of `:unit:percent` or `:unit:length` would mitigate this problem.

There are many quantities, but I would be happy enough splitting out the most important ones and throwing everything else into `:unit:other` or something.

exploration/percent-format.md
```diff
@@ -0,0 +1,389 @@
...
+
+## Proposed Design
+
+_Describe the proposed solution. Consider syntax, formatting, errors, registry, tooling, interchange._
+
+Support the formatting of percent values as follows:
+
+- REQUIRE the `:unit` function for all implementations
+  - Only specific `unit` option values are required, initially the unit `percent`.
+  - The function `:unit unit=percent` does not scale the operand, e.g. `{5 :unit unit=percent}` formats as `5%`.
```
sffc, line 142: Sorry but requiring `:unit` in its current form is a non-starter for ICU4X.

sffc:
> > The current design of :unit is not implementable with ICU4X's data design for reasons laid out in #1006.
> 
> I looked at #1006 and didn't find a discussion of unit.

We did not take too close of a look at `:unit` when writing that doc because `:unit` was not marked as being required, but it suffers from many of the same problems as, for example, `u:locale`.

sffc:
I'm spinning off the `:unit` discussion into #1079

eemeli (approved):
As discussed on today's call, I've moved the proposed solution to be one of the alternatives under consideration.

The intent with this change is to allow us to land this PR, and to iterate further from there. As discussed today, adding a separate `:percent` function might be the least worst option here, and it would match what we're doing with unit and currency formatting.

exploration/percent-format.md
sffc: ^ this comment is unresolved. At least the third bullet point needs to clarify that `scaling` _is_ an option that could uniquely apply to percentage formatting.

exploration/percent-format.md
```diff
@@ -0,0 +1,396 @@
...
+.local $n = {0.5 :math multiply=100}
+{{Completion: {$n :unit unit=percent}.}}
+```
+
+_Pros_
+- Can be used for other general purpose math
+
+_Cons_
+- Brings in multiplication
+
```
sffc, line 368:
```suggestion
_Cons_
- Increases implementation burden: multiplication must be handled on arbitrary numeric input types

```

sffc (commented):
I like landing this with all the alternatives listed as alternatives and not taking a position on which one we want to pick.

exploration/percent-format.md
```diff
@@ -0,0 +1,389 @@
...
+> [!NOTE]
+> @sffc suggested that we should consider other names for `:percent`.
+> The name shown here could be considered a placeholder pending other suggestions.
+
+_Pros_
+- Least verbose placeholder
+- Clear what the placeholder does; self-documenting
+- Consistent with separating specialized formats from `:number`/`:integer`
+  as was done with `:currency`
+
```
eemeli, line 228:
```suggestion
- Makes it possible to apply a `scaling` option to only percent formatting.

```

exploration/percent-format.md
```diff
@@ -0,0 +1,396 @@
...
+### Why not both?
+
+Rather than choosing only one option, choose multiple parallel solutions:
+
+- REQUIRE the `:unit` function for all implementations
+  - Only specific `unit` option values are required, initially the unit `percent`.
+  - The function `:unit unit=percent` does not scale the operand, e.g. `{5 :unit unit=percent}` formats as `5%`.
+- REQUIRE the `:number` and `:integer` functions to support `style=percent` as an option
+  - The functions `:number` and `:integer` scale the operand, e.g. `{5 :integer style=percent}` formats as `500%`.
```
eemeli, line 379: From https://github.com/unicode-org/message-format-wg/pull/1068#discussion_r2099629795:

```suggestion
- REQUIRE the `:number` function to support `style=percent` as an option
  - The function `:number`scales the operand, e.g. `{5 :number style=percent}` formats as `500%`.
```

exploration/percent-format.md
eemeli: Moved to https://github.com/unicode-org/message-format-wg/pull/1068#discussion_r2206721332

sffc (approved):

PR #1070, opened 2025-04-14
Allow clamping of digit size options

### LINKED ISSUE #960: [FEEDBACK] Limits of Digit Size Options
The Digit Size Options in registry.md say they use the following EBNF:

> digit-size-option = "0" / (("1"-"9") [DIGIT])

This appears to accept 0, 1, 2, ..., 9, 10, 11, ..., 98, 99.

However, ECMA-402 restricts the range of some of these integers to be less. For example, minimum and maximum significant digits and minimum integer digits are restricted to the range 1-21:

https://tc39.es/ecma402/#sec-setnfdigitoptions

aphillips:
Fixes #960 

Per the discussion in the 2025-04-14 telecon, allow digit size options to be to clamped to an implementation-specific value.

### DIFF
--- a/spec/functions/number.md
+++ b/spec/functions/number.md
@@ -631,8 +631,11 @@ digit-size-option = "0" / (("1"-"9") [DIGIT])
 
 If the value of a _digit size option_ does not evaluate as a non-negative integer,
 or if the value exceeds any implementation-defined and option-specific upper or lower limit,
-the implementation will emit a _Bad Option Error_ and
-either ignore the option or replace it with an implementation-defined value.
+the implementation will emit a _Bad Option Error_
+and ignore the _option_.
+An implementation MAY replace a _digit size option_
+that exceeds an implementation-defined or option-specific upper or lower limit
+with an implementation-defined value rather than ignoring the _option_.
 
 #### Number Selection
 
--- a/spec/functions/number.md
+++ b/spec/functions/number.md
@@ -633,7 +633,7 @@ If the value of a _digit size option_ does not evaluate as a non-negative intege
 or if the value exceeds any implementation-defined and option-specific upper or lower limit,
 a _Bad Option_ error is emitted.
 If the value exceeds an implementation-defined upper or lower limit, 
-the implementation MAY replace the offending value with an implementation-defined value.
+the implementation MAY replace it with an implementation-defined value.
 
 #### Number Selection
 
--- a/spec/functions/number.md
+++ b/spec/functions/number.md
@@ -632,6 +632,8 @@ digit-size-option = "0" / (("1"-"9") [DIGIT])
 If the value of a _digit size option_ does not evaluate as a non-negative integer,
 or if the value exceeds any implementation-defined and option-specific upper or lower limit,
 a _Bad Option_ error is emitted.
+If the value exceeds an implementation-defined upper or lower limit, 
+the implementation MAY replace the offending value with an implementation-defined value.
 
 #### Number Selection
 
--- a/spec/functions/number.md
+++ b/spec/functions/number.md
@@ -617,7 +617,7 @@ such as the number of fraction, integer, or significant digits.
 A **_<dfn>digit size option</dfn>_** is an _option_ 
 whose _option value_ is interpreted by the _function_
 as a small integer greater than or equal to zero.
-Implementations MAY define an upper limit on the _resolved value_
+Implementations MAY define upper and lower limits on the _resolved value_
 of a _digit size option_ consistent with that implementation's practical limits.
 
 In most cases, the value of a _digit size option_ will be a string that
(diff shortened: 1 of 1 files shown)


spec/functions/number.md
```diff
@@ -632,6 +632,8 @@ digit-size-option = "0" / (("1"-"9") [DIGIT])
 If the value of a _digit size option_ does not evaluate as a non-negative integer,
 or if the value exceeds any implementation-defined and option-specific upper or lower limit,
 a _Bad Option_ error is emitted.
+If the value exceeds an implementation-defined upper or lower limit, 
+the implementation MAY replace the offending value with an implementation-defined value.
```
eemeli, line 636: Nitpicky language suggestion, as "offending" seems a little surprising in this context.

```suggestion
If the value exceeds an implementation-defined upper or lower limit, 
the implementation MAY replace it with an implementation-defined value.
```

eemeli (approved):

spec/functions/number.md
```diff
@@ -632,6 +632,8 @@ digit-size-option = "0" / (("1"-"9") [DIGIT])
 If the value of a _digit size option_ does not evaluate as a non-negative integer,
 or if the value exceeds any implementation-defined and option-specific upper or lower limit,
 a _Bad Option_ error is emitted.
+If the value exceeds an implementation-defined upper or lower limit, 
+the implementation MAY replace it with an implementation-defined value.
```
sffc, line 636: Does this happen _instead of_ the Bad Option error?

Is there somewhere else in the spec that says what happens if the provided option value is not in-range, whether it is dropped, etc? (Seems like there would be)

aphillips: > Does this happen instead of the Bad Option error?

No. The spec (on the previous line!) says a Bad Option is emitted, so one is emitted. This proposed addition allows the implementation to do something other than dropping the option on the floor (which would presumably result in the default value for that implementation being used). It also would have an effect on the _resolved value_ of the option.

The spec in general does not define behavior like this because of the layering in the design. What happens in a function is the function's business. Other functions might have similar or different behavior, depending on their needs. (Some options, when in error, leave the function unable to proceed.)

There are two places in the spec where something else could happen (on another layer).

One is during [option resolution](https://www.unicode.org/reports/tr35/dev/tr35-messageFormat.html#option-resolution), but that only results in a map of options to resolved values. That is where the text says that option resolution cannot fail (produce a fatal error). @macchiati and others have proposed allowing function descriptors and to be used to validate values _before_ function resolution. Such evaluation might result in the Bad Option occurring sooner.

The other is during [function resolution](https://www.unicode.org/reports/tr35/dev/tr35-messageFormat.html#function-resolution) during formatting. Effectively, "function resolution" is when the function is actually called. The function itself decides whether a Bad Option is a _failure_ or not. If it is a failure, the fallback representation is the result of formatting. Presumably if one clamped the value of a Digit Size Option, one would not emit a fallback representation (there would be no point in clamping the value, since the result is something like `{$foo}` not a formatted value such as `0.123456789012345678901`) making the Bad Option just a non-fatal error.

sffc: ok, so maybe a better resolution here is to just say somewhere that when a Bad Option error occurs, implementations may _either_ drop the option value _or_ substitute an implementation-defined replacement?

eemeli, line 636: To resolve the ambiguity noted by @sffc in https://github.com/unicode-org/message-format-wg/pull/1070#discussion_r2043120996:
```suggestion
a _Bad Option_ error is emitted and
the implementation MAY ignore the option value or
replace it with an implementation-defined value.
```

eemeli: One part of the change that I'm proposing here is allowing an implementation to behave the same way for a digit size option ending up with either `-1` or `0` as values for an option that has an implementation-defined minimum of `1`. Is there a reason why this should not be allowed?

aphillips: There is no reason that should not be allowed. I think my proposal allows that (the implementation can ignore the option **_or_** replace it)?

eemeli (approved):

spec/functions/number.md
```diff
@@ -631,9 +631,8 @@ digit-size-option = "0" / (("1"-"9") [DIGIT])
 
 If the value of a _digit size option_ does not evaluate as a non-negative integer,
 or if the value exceeds any implementation-defined and option-specific upper or lower limit,
-a _Bad Option_ error is emitted.
-If the value exceeds an implementation-defined upper or lower limit, 
-the implementation MAY replace it with an implementation-defined value.
+the implementation will emit a _Bad Option Error_ and
```
catamorphism: Should the "will" here be a "MUST"?

mihnita (approved):

PR #1071, opened 2025-04-14
Currency and unit conformance

### LINKED ISSUE #1009: [FEEDBACK] Conformance for currency and unit should be more strict
Following up to my #977 audit:

> The value of the operand's currency MUST be either a string containing a well-formed [Unicode Currency Identifier](https://unicode.org/reports/tr35/tr35.html#UnicodeCurrencyIdentifier) or an implementation-defined currency type. Although currency codes are expected to be uppercase, implementations SHOULD treat them in a case-insensitive manner.

Why is this SHOULD and not MUST? Pick a behavior but don't let this be implementation-defined.

> The value of the operand's unit SHOULD be either a string containing a valid [Unit Identifier](https://www.unicode.org/reports/tr35/tr35-general.html#unit-identifiers) or an implementation-defined unit type.

Huh? So the operand is a TR35 Unit ID or an implementation-defined unit type, but hey, it's not necessarily either one due to the SHOULD?

I suggest:

1. The string MUST be a TR35 unit identifier
2. An implementation MAY allow other syntaxes (although I prefer to just be strict and require TR35 unit IDs)

aphillips:
Fixes #1009 

Makes currency codes case-insensitive more clearly. We used a squidgy "SHOULD" before in spite of the ABNF provided. 

Make units either CLDR identifiers or impl-defined types (SHOULD->MUST).

This is consistent with the stability policy (no new error is specified, no valid message becomes invalid, and `:unit` is DRAFT)

### DIFF
--- a/spec/functions/number.md
+++ b/spec/functions/number.md
@@ -257,8 +257,7 @@ Using this _option_ in such a case results in a _Bad Option_ error.
 The value of the _operand_'s `currency` MUST be either a string containing a
 well-formed [Unicode Currency Identifier](https://unicode.org/reports/tr35/tr35.html#UnicodeCurrencyIdentifier)
 or an implementation-defined currency type.
-Although currency codes are expected to be uppercase,
-implementations SHOULD treat them in a case-insensitive manner.
+Currency codes are case-insensitive.
 A well-formed Unicode Currency Identifier matches the production `currency_code` in this ABNF:
 
 ```abnf
@@ -431,7 +430,7 @@ each of which contains a numerical `value` plus a `unit`
 or it can be a [Number Operand](#number-operands), as long as the _option_
 `unit` is provided.
 
-The value of the _operand_'s `unit` SHOULD be either a string containing a
+The value of the _operand_'s `unit` MUST be either a string containing a
 valid [Unit Identifier](https://www.unicode.org/reports/tr35/tr35-general.html#unit-identifiers)
 or an implementation-defined unit type.
 


spec/functions/number.md
```diff
@@ -257,8 +257,7 @@ Using this _option_ in such a case results in a _Bad Option_ error.
 The value of the _operand_'s `currency` MUST be either a string containing a
 well-formed [Unicode Currency Identifier](https://unicode.org/reports/tr35/tr35.html#UnicodeCurrencyIdentifier)
 or an implementation-defined currency type.
-Although currency codes are expected to be uppercase,
-implementations SHOULD treat them in a case-insensitive manner.
+Currency codes are case-insensitive.
```
eemeli, line 260: This leaves it unclear what the resolved value is. Could we be more explicit?
```suggestion
The resolved `currency` value is normalized to upper-case.
```

eemeli (changes requested):

spec/functions/number.md
aphillips: As a reminder, even ASCII letters have locale-sensitive case folding exceptions (Turkic languages).

Do we need to specify that the resolved value is uppercased? Mostly we treat the operands as immutable. We don't have to require that implementations perform the case fold operation. It's a good idea to do it to save later processing, but we can't really test it at the MF level except to say that a value with `usd` is as functional as one with `USD`.

eemeli: I think we should specify the casing, so that it's clear in a message like
```
.input {$n :currency currency=Usd}
{{{$n :foo:bar}}}
```
what `currency` value the `:foo:bar` function handler will see.

```suggestion
The resolved `currency` value is locale-insensitively normalized to upper-case.
```

aphillips: I have a variety of quibbles with the suggestion, starting with not using the defined term _resolved value_. Also, we allow implementation-defined (currency) types in this option also. Since the literals must be ASCII letters, we can use Unicode's casefold algorithm. Perhaps:

```suggestion
Currency codes are case-insensitive.
If the value of the `currency` _option_ is a _literal_, the _resolved value_ MUST be case-folded to uppercase.
```

eemeli: Why limit the case-folding to literal values only? Why should the resolved value be different for `$a` vs `$b`?
```
.local $cc = {eur}
.input {$a :currency currency=eur}
.input {$b :currency currency=$cc}
...
```

aphillips: The value would have been resolved before reaching the function. So it is either resolved as a literal or as an implementation-defined type. We can't uppercase a `Currency` object, but we can on a literal. Note that your examples are in the MF processor space, but this directive is inside the function handler.

eemeli: We don't currently have any place in the spec that defines specific behaviour based on whether a value originates from a _literal_, and I would very much prefer not adding anything like that. The restriction on `:number select` makes sense and is only about the immediate _option value_ in part to avoid needing to track the origins of variable values like this.

Given that we say that
> The _resolved value_ of a _text_ or a _literal_ contains the character sequence of the _text_ or _literal_ after any character escape has been converted to the escaped character.

The definition here ought to limit its impact not on whether the option's resolved value came from a _literal_, but on whethe or not it contains a character sequence.

catamorphism (changes requested):

mihnita:
> the _resolved value_ MUST be case-folded to uppercase.

I don't think we should say this.

Most of the resolving work is done by the MF proper.
And at some point the function is invoked.

In order for the "MF engine" to resolve currencies by upper-casing them, it means that the engine should know about the `:currency` function, and about the rules governing it's options (in this case `currency`).

That is not clean design, and it is unnecessary.

The `:currency` function should recognize the currency codes regardless of case.

We can say that the parts returned by the `:currency` function (when we format to parts) are normalized to uppercase.
But not "resolving" in general.

spec/functions/number.md
sffc: Why can't we just require that the currency is upper-case, and then we don't need to worry about the case folding stuff.

sffc: Also, "case-folded to uppercase" doesn't make sense. Unicode Case Mapping defines algorithms for Uppercase, Lowercase, Titlecase, and Case Fold. You don't "case fold to uppercase".

catamorphism, line 260:
```suggestion
Currency codes are case-insensitive.

If the _resolved value_ of the `currency` option is a character sequence, then the character sequence MUST be treated within the function handler as if it had been case-folded to uppercase. The function handler MUST reproduce the original option value in the _resolved value_ that it returns (not the case-folded version). For example, the resolved value returned by the `currency` function handler for `{$a :currency currency=eur}` MUST have a `resolvedOptions()` method that returns an option mapping including a mapping from `currency` to the character sequence `"eur"` (not `"EUR"`.)
```

New suggestion based on today's discussion. A bit verbose, but feel free to bikeshed.

catamorphism: > We definitely do not want to specify a "resolvedOptions() method" because we have gone out of our way not to specify the specific APIs or data structures.

That's a good point, and is a reminder about a bigger issue: we need to decide whether to make the resolved-value interface normative.

It's hard for me to see how to re-word this text given that it's non-normative. "The _resolved value_ of the _option_ is not modified" isn't sufficient, IMO, because that's always true (option values are immutable). What we want to say is that in the data structure produced by the function handler, the value of the `currency` option is the same as the value that was specified in the function expression. And without committing to what that data structure looks like, I don't know how to express that.

I think this will come up with specifications of other functions, not just currency.

macchiati: This gets back to a fundamental issue. When we compose functions, what will users expect? Take the following example:
```
.local x = {$in :func1 opt1=oval1 opt2=oval2}
.local y = {$in :func2            opt2=oval2x opt3=oval3}
```
There is a tension between simplicity/predictability and capability. Sometimes making something more capable thereby reduces the simplicity and predictability. We can do that, but _only_ if the benefit outweighs the cost.

In my mind, the simplest and most predictable behavior would be that for all functions and options, the above behaves *precisely* like:
 ```
.local x = {$in :func1 opt1=oval1 opt2=oval2}
.local y = {$in :func2 opt1=oval1 opt2=oval2x opt3=oval3}
```
except that `:func2` must ignore the implicit `opt1=oval1` if it would be invalid for `:func2`. Invalidity could be caused by multiple reasons, of course.
1. `opt1` is not allowed by `:func` *, or 
2. `opt1=oval1` is not allowed by `:func2`, or 
3. `opt1=oval1` would be allowed by `:func2`, but conflicts with one or more of the options that are _explicitly_ set (`opt2=oval2x`, `opt3=oval3`).

\* We could  (optionally) allow functions to specify some _implicit_ function/option pairs that they would not allow explicitly.

Trivially, if `func2` = `:func1`, then opt1 is allowed unless it conflicts (3).

So then the question becomes: what are some powerful use cases that would justify exceptions to this policy?

Comment (aphillips) [reply spec/functions/number.md:260]: @catamorphism I think you just did express it :😉? I think we could say in the text about function resolution that the options passed are immutable by the handler. The resolved value might contain additional options. I'm wondering if it might also modify an option (consider DSOs in the other PR!) I could see not allowing function local mods to show. 

@macchiati we already allow functions to ignore options they don't care about. And we allow functions to disallow combos and values. It's okay for chained functions to result in Bad Option, even mid chain. That just makes sense. But the option isn't removed, just ignored (we already say that) 

(Doing this on a phone, so handicapped in my response)

**Comment by eemeli** [comment]
One aspect we have not discussed yet is prior art, and how similar APIs elsewhere behave. Do ICU APIs provide for the inspection of any option-like values in intermediate formattable values? If so, do they modify their input values at all?

In JS there's
```js
const nf = new Intl.NumberFormat(undefined, { style: 'currency', currency: 'eur' })
nf.resolvedOptions().currency === 'EUR'
```
where the option values are parsed, filtered, and normalized.

This seems quite relevant to me, as the authors of new messages are likely to be developers, and their presumptions about MF2 will draw on their experiences with related i18n and L10n APIs.

spec/functions/number.md
aphillips: > It's hard for me to see how to re-word this text given that it's non-normative.

This text _is_ normative, even if it does not use 2119 keywords. It would only be non-normative if it appears in a note or in a section otherwise marked as non-normative. We use 2119 keywords to indicate particular normative behavior for clarity. but bald statements behave like MUST.

@macchiati 
I think your example is mixing a bunch of things together. In particular, it implies that the first `.local` causes the variable `$in` to be annotated with `opt1` and `opt2`, but I don't think this is true. See [here](https://www.unicode.org/reports/tr35/dev/tr35-messageFormat.html#resolved-values) in the spec.

Your point is somewhat different, but we need to be careful about resolved value chaining. I think the example should be more like:

```
.input      {$in :func1 opt1=oval1 opt2=oval2}
.local $y = {$in :func2            opt2=oval2x opt3=oval3}
{{$y has RV: {$y :func2 opt1=oval1 opt2=oval2x opt3=oval3}
 $in has RV {$in :func1 opt1=oval1 opt2=oval2}.}}
```

Note that `$y`'s value might be different from that of `$in` because of the operation of `:func1`. 

For example, If `:func1` were `:uppercase` and the input of `$in` were the string `usd`, then `$in`'s value would be `usd` but `$y`'s would be `USD`. The placeholder `{$in}` would print as `USD` (because of the function) or, if it were a selector, would match the key `USD` and **not** match `usd`.

The question here, of course, is about option transitivity. We already partially show this in the linked section above, like so:
```
.input {$a :number minimumFractionDigits=3}
.local $b = {$a :integer useGrouping=never}
.match $a
0 {{The value is zero.}}
* {{Without grouping separators, the value {$a} is rendered as {$b}.}}
```

I think we mean these to behave as if:
```
.input {$a :number minimumFractionDigits=3}
.local $b = {$a :integer minimumFractionDigits=3 useGrouping=never}
```
In this case, I think `:integer` ignores `minimumFractionDigits`, an option it doesn't know about/support and the option might not be expressed in the _resolved value_ of `$b`, because `minimumFractionDigits` is not one of the options `:integer` resolves.

This is a long way of saying that I think we need to add to the core spec (rather than to `:currency`/`:unit`) clarity about whether functions pass options they do **not** know (or not).

aphillips:
Discussed in 2025-05-12 teleconference

PR #1072, opened 2025-04-21
Fix unpaired surrogate in test

### LINKED ISSUE #1062: Test for unpaired surrogates is rejected by some JSON parsers
The syntax-errors.json file contains a test (which should fail) for an unpaired surrogate in a literal:

https://github.com/unicode-org/message-format-wg/blob/main/test/tests/syntax-errors.json#L195

This JSON is rejected by the parser I'm using in ICU4C, https://github.com/nlohmann/json/ :

`  { "src": "{\ud800}" }`

because it's not valid UTF-8. (Note: I would think that the escaping with `\u` would make it valid, but that's the error that the JSON parser emits.)

I solved the problem by adding an ICU-specific patch to the JSON parser, but that's not ideal. In [this comment](https://github.com/unicode-org/icu/pull/3423#issuecomment-2759420719), @srl295 suggested using base64 encoding for the test instead.

aphillips:
Fixes #1062 per 2025-04-21 call

### DIFF
--- a/test/tests/syntax-errors.json
+++ b/test/tests/syntax-errors.json
@@ -192,7 +192,11 @@
     { "src": "{^.}" },
     { "src": "{^ .}" },
     { "src": "{&}" },
-    { "src": "{\ud800}" },
+//  { "src": "{\ud800}" }, Unpaired surrogate test.
+//                         This is not valid JSON and some JSON parsers will not parse this test.
+//                         Uncomment if your implementation supports unpaired surrogates
+//                         for text and wants to test invalid placeholders.
+//                         See: https://www.unicode.org/reports/tr35/dev/tr35-messageFormat.html#the-message
     { "src": "{\ufdd0}" },
     { "src": "{\ufffe}" },
     { "src": "{!.\\{}" },


test/tests/syntax-errors.json
```diff
@@ -192,7 +192,9 @@
     { "src": "{^.}" },
     { "src": "{^ .}" },
     { "src": "{&}" },
-    { "src": "{\ud800}" },
+    { "description": "Valid JSON cannot contain an unpaired surrogate. Unescape backslash to create a syntax error for implementations that support UTF-16.",
+      "src": "{\\ud800}"
+    },
```
eemeli, line 197:
```suggestion
```

eemeli (changes requested):
If we don't want to include an escaped unpaired surrogate to accommodate a possible understanding of the JSON spec, we should not include a double-escaped one either.

In other words, if this one particular feature is not representable in JSON, we should instead describe the required test in `test/README.md` or somewhere else, as it'll need special handling in any case.

test/README.md
```diff
@@ -1,12 +1,21 @@
...
 
-These test files are intended to be useful for testing multiple different message processors in different ways:
+These test files are intended to be useful for testing multiple different _message_ processors in different ways:
 
 - `syntax.json` — Test cases that do not depend on any registry definitions.
 
 - `syntax-errors.json` — Strings that should produce a Syntax Error when parsed.
 
+> [!NOTE]
+> `syntax-errors.json` does not include tests for the disallowed uses of unpaired surrogate code points
```
eemeli, line 10: It's not relevant where these tests are not.
```suggestion
> Tests for the disallowed uses of unpaired surrogate code points are not included
```

test/README.md
```diff
@@ -1,12 +1,21 @@
...
+> because JSON does not permit unpaired surrogate code points.
+> If your implementation uses UTF-16 based strings (such as JavaScript `String` or Java `java.lang.String`)
+> or otherwise allows unpaired surrogates in text or literals, you will need to implement tests equivalent
+> to the following for syntax errors:
+> ```json
+> { "src": "{\ud800}" }
```
eemeli, line 16: This should be a fully qualified test, rather than relying on the `syntax-errors.json` `"defaultTestProperties"`.
```suggestion
> {
>   "locale": "en-US",
>   "src": "{\ud800}",
>   "expErrors": [{ "type": "syntax-error" }]
> }
```

eemeli (changes requested):

eemeli (approved):

catamorphism (approved):

mihnita (approved):

PR #1073, opened 2025-05-12
Change `:math` to `:offset` and remove draft status

### LINKED ISSUE #959: [FEEDBACK] The `:math` function should be better-specified
As far as I can tell, in registry.md, the `:math` function allows for arbitrary addition and subtraction.

I think this is a problem: it can run into the floating-point math problems well known to most programmers, such as 0.1 + 0.2 != 0.3. For example, a message of the following form does not have well-defined behavior:

```
.input {$num :number}
.local $num_adjusted = {$num :math subtract=0.1}
.match $num_adjusted
0.1 {{...}}
0.2 {{...}}
0.3 {{...}}
*   {{...}}
```

Are there any use cases for this function other than adding or subtracting integers?

I suggest changing the function to restrict the domain to integers and defining that the math operates as if on decimal values. So, for example, `{1.234 :math subtract=1}` should be guaranteed to return `0.234`.

aphillips:
In the 2025-05-12 teleconference, the group agreed to make the name of this function less generic. A number of alternative names were proposed in the call. The name `:offset` was used here, partially as a placeholder while we seek consensus.

We agreed that the new function would be limited to adjusting the value of the operand by small integer values. _Digit Size Option_, used here, would make that from `0` to `999`. It has been proposed that this be restricted to `1` to `99`.

In this pull request, note that I added the function name to subsections. The current spec doesn't link properly to subsections because all functions use the same section ID. I did not change the other functions to match, pending discussion.

Fixes #959

### DIFF
--- a/spec/functions/number.md
+++ b/spec/functions/number.md
@@ -159,22 +159,19 @@ together with the resolved options' values.
 
 The _function_ `:integer` performs selection as described in [Number Selection](#number-selection) below.
 
-#### The `:math` function
+#### The `:offset` function
 
-> [!IMPORTANT]
-> The _function_ `:math` has a status of **Draft**.
-> It is proposed for inclusion in a future release of this specification and is not Stable. 
-
-The _function_ `:math` is proposed as a _selector_ and _formatter_ for matching or formatting
-numeric values to which a mathematical operation has been applied.
+The _function_ `:offset` is a _selector_ and _formatter_ for matching or formatting
+numeric values to which an offset has been applied.
+The "offset" is a small integer adjustment of the _operand_'s value.
 
 > This function is useful for selection and formatting of values that
 > differ from the input value by a specified amount.
-> For example, it can be used in a message such as this:
+> For example, it can be used in a _message_ such as this:
 >
 > ```
 > .input {$like_count :integer}
-> .local $others_count = {$like_count :math subtract=1}
+> .local $others_count = {$like_count :offset subtract=1}
 > .match $like_count $others_count
 > 0 *   {{Your post has no likes.}}
 > 1 *   {{{$name} liked your post.}}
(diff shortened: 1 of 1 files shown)


spec/functions/number.md
```diff
@@ -159,40 +159,41 @@ together with the resolved options' values.
...
 > 0 *   {{Your post has no likes.}}
 > 1 *   {{{$name} liked your post.}}
 > * one {{{$name} and {$others_count} other user liked your post.}}
 > * *   {{{$name} and {$others_count} other users liked your post.}}
 > ```
 
-##### Operands
+> [!NOTE]
+> The purpose of this _function_ is to supply compatibility with
+> ICU's `PluralFormat` and its `offset` feature, also found in ICU MessageFormat.
```
macchiati, line 184:
```suggestion
> ICU's `PluralFormat` and its `offset` feature, also found in ICU MessageFormat.
> It is needed for migration, because it is commonly used for messages
> that need to convey the first few items of a list plus the count of the remainder,
> such as in the example above.
```

spec/functions/number.md
```diff
@@ -159,40 +159,41 @@ together with the resolved options' values.
...
 
-##### Options
+The function `:offset` requires a [Number Operand](#number-operands) as its _operand_.
+
+##### `:offset` Options
 
-The _options_ on `:math` are exclusive with each other,
+The _options_ on `:offset` are exclusive with each other,
 and exactly one _option_ is always required.
 The _options_ do not have default values.
```
macchiati, line 194:
```suggestion
```
Delete, there is only one, and it is required below.

spec/functions/number.md
```diff
@@ -159,40 +159,41 @@ together with the resolved options' values.
...
 
-The following _options_ are REQUIRED to be available on the function `:math`:
+The following _options_ are REQUIRED to be available on the function `:offset`:
 
 - `add`
```
macchiati, line 198: It's actually a subtract, since it is the remaining count.

spec/functions/number.md
```diff
@@ -222,18 +223,18 @@ If the _operand_ of the _expression_ is an implementation-defined numeric type,
 such as the _resolved value_ of an _expression_ with a `:number` or `:integer` _annotation_,
 it can include option values.
 These are included in the resolved option values of the _expression_.
-The `:math` _options_ are not included in the resolved option values.
+The `:offset` _options_ are not included in the resolved option values.
```
macchiati, line 226: Why?

spec/functions/number.md
eemeli: I think we're inverting things a bit here. As I see it, the primary purpose of this function is the same as that of plural offset in MF1, to enable the expression of messages with plural variance that depends on an offset count. Supporting migration from MF1 arises from that; it ought not be presented as "the purpose" of the function as a whole.

spec/functions/number.md
eemeli: At least two reasons:
1. Clarity; the function processes and "consumes" the option, as its resolved value includes the offset. This also allows for an implementation to present the resolved value of an `:offset` and a `:number` using the same representation.
2. We've not identified any use cases that would benefit from a consumer of an `:offset` value to know that an offset was applied to the value, or what that offset was.

eemeli (approved):

PR #1076, opened 2025-06-02
Make expErrors use an array (and only an array)

mihnita:
Fixes issue #993

### DIFF
--- a/test/tests/fallback.json
+++ b/test/tests/fallback.json
@@ -18,7 +18,7 @@
       "description": "function with quoted literal operand",
       "src": "{|C:\\\\| :test:function fails=format}",
       "exp": "{|C:\\\\|}",
-      "expErrors": [{ "type": "bad-option" }]
+      "expErrors": [{ "type": "bad-operand" }]
     },
     {
       "description": "unannotated implicit input variable",


test/tests/functions/datetime.json
```diff
@@ -5,7 +5,7 @@
   "defaultTestProperties": {
     "bidiIsolation": "none",
     "locale": "en-US",
-    "expErrors": false
+    "expErrors": []
```
aphillips, line 8: In fallback.md you left out `expErrors`. Here you supply an empty array. I think we should probably omit this one?

test/tests/functions/time.json
```diff
@@ -5,7 +5,7 @@
   "defaultTestProperties": {
     "bidiIsolation": "none",
     "locale": "en-US",
-    "expErrors": false
+    "expErrors": []
```
aphillips, line 8: ibid

test/tests/fallback.json
```diff
@@ -4,8 +4,7 @@
   "description": "Test cases for fallback behaviour.",
   "defaultTestProperties": {
     "bidiIsolation": "none",
-    "locale": "en-US",
-    "expErrors": true
```
eemeli, line 8: Was dropping this intentional? All of the messages in this file are expected to experience fallback and emit at least one error.

test/tests/functions/datetime.json
mradbourne: @aphillips It's required in this file otherwise some tests would be left without an expectation (e.g. lines 44-55). This would also be the case in `time.json`.

As Eemeli mentioned in his review comment, I think we need to keep the `expErrors` in `fallback.md`.

mihnita: I think we are talking about `fallback.json`, not `fallback.md`?

I updated `fallback.json` and added errors for each individual case, instead of doing it in `defaultTestProperties`

mihnita: For `date.json`, `datetime.json` and `time.json` I added `"expErrors": []` to `defaultTestProperties` because the schema says that entries are required to have one of `exp`, `expParts` or `expErrors`

In these 3 files we had tests with `src` only.
So I had to add a `expErrors` (not  enough info for `exp`, `expParts`, exact date/time formatting not part of the spec)

test/tests/functions/time.json
mihnita: see above

aphillips (approved):

aphillips:
need to change bad operand to bad option (2025-06-16)

aphillips:
In 2025-06-30 call, we agreed to merge this once @eemeli has reviewed positively. Nudging him gently with this comment.

test/tests/fallback.json
```diff
@@ -4,52 +4,59 @@
...
+    "locale": "en-US"
   },
   "tests": [
     {
       "description": "function with unquoted literal operand",
       "src": "{42 :test:function fails=format}",
       "exp": "{|42|}",
-      "expParts": [{ "type": "fallback", "source": "|42|" }]
+      "expParts": [{ "type": "fallback", "source": "|42|" }],
+      "expErrors": [{ "type": "bad-operand" }]
```
eemeli, line 15:
```suggestion
      "expErrors": [{ "type": "bad-option" }]
```

eemeli (changes requested):
There's a test that also needs to be updated to use `bad-option`.

test/tests/fallback.json
aphillips: Hmm... I found this confusing until I remembered that we agreed to use `bad-option` to avoid creating a label for `message-function-error` in the **_not at all official_** definition under tests [here](https://github.com/unicode-org/message-format-wg/tree/main/test#error-codes). Non-WG members won't have the benefit of having been on the telecon when we discussed it 😉 

Unlike the other classes of error, MFE can have implementation-defined errors. We should make this clearer in the test suite text so that implementations do not merely ape our selection of `bad-option` when an internal exception/error occurs. I think it would be okay to commit this, but we need to put more effort into the text in the /tests page to make super clear that `bad-option` is not required by non-test functions.

eemeli: Agreed on all counts. I also noticed while implementing this change that we lack a generic error for failures in formatting or selection, which could be due to the operand, options, or the phase of the moon. As in, values for which resolution succeeds, but then the `MessageValue.formatToString()` or similar call throws an error. It's pretty likely to involve some bad option, but that's not actually guaranteed -- and beyond the scope of this PR.

eemeli (approved):

PR #1077, opened 2025-06-08
Include :datetime, :date, and :time with style options only

eemeli:
As discussed on the 2025-06-02 call, this PR removes the "Draft" status from the `:datetime`, `:date`, and `:time` functions, while leaving their _field options_ as "Draft". The intent here is to allow at least _some_ datetime formatting to be made available while discussions are #1067 are ongoing.

### DIFF
--- a/spec/functions/datetime.md
+++ b/spec/functions/datetime.md
@@ -10,7 +10,7 @@ This subsection describes the _functions_ and _options_ for date/time formatting
 
 #### The `:datetime` function
 
-The function `:datetime` is used to format combined date/time values.
+The function `:datetime` is used to format both the date and time of a date/time value.
 
 If no options are specified, this function defaults to the following:
 
@@ -51,7 +51,7 @@ If the _operand_ of the _expression_ is an implementation-defined date/time type
 it can include other option values.
 Any _date/time override options_ of the operand are included in the resolved option values of the _expression_,
 with _options_ on the _expression_ taking priority over any options of the _operand_.
-Any _operand_ options not matching the _Date/time override options_ are ignored.
+Any _operand_ options not matching the _date/time override options_ are ignored.
 
 ###### Field Options
 
@@ -155,7 +155,7 @@ If the _operand_ of the _expression_ is an implementation-defined date/time type
 it can include other option values.
 Any _date/time override options_ of the operand are included in the resolved option values of the _expression_,
 with _options_ on the _expression_ taking priority over any options of the _operand_.
-Any _operand_ options not matching the _Date/time override options_ are ignored.
+Any _operand_ options not matching the _date/time override options_ are ignored.
 
 ##### Resolved Value
 
@@ -196,7 +196,7 @@ If the _operand_ of the _expression_ is an implementation-defined date/time type
 it can include other option values.
 Any _date/time override options_ of the operand are included in the resolved option values of the _expression_,
 with _options_ on the _expression_ taking priority over any options of the _operand_.
-Any _operand_ options not matching the _Date/time override options_ are ignored.
+Any _operand_ options not matching the _date/time override options_ are ignored.
 
 ##### Resolved Value
 
--- a/spec/functions/datetime.md
+++ b/spec/functions/datetime.md
@@ -2,10 +2,6 @@
 
 This subsection describes the _functions_ and _options_ for date/time formatting.
 
-> [!IMPORTANT]
-> The _functions_ in this section have a status of **Draft**.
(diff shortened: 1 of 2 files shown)


spec/functions/datetime.md
```diff
@@ -14,15 +10,14 @@ This subsection describes the _functions_ and _options_ for date/time formatting
 
 #### The `:datetime` function
 
-The function `:datetime` is used to format date/time values, including
-the ability to compose user-specified combinations of fields.
+The function `:datetime` is used to format combined date/time values.
```
aphillips, line 13: Perhaps clearer as:

```suggestion
The function `:datetime` is used to format both the date and time of a date/time value.
```

spec/functions/datetime.md
```diff
@@ -35,52 +30,52 @@ All other _operand_ values produce a _Bad Operand_ error.
...
+- _Date/time override options_
+
+`dateStyle` and `timeStyle` are **_<dfn>style options</dfn>_**
+which pertain to the overall styling or appearance of the formatted output.
+
+If the _operand_ of the _expression_ is an implementation-defined date/time type,
+it can include other option values.
+Any _date/time override options_ of the operand are included in the resolved option values of the _expression_,
+with _options_ on the _expression_ taking priority over any options of the _operand_.
+Any _operand_ options not matching the _Date/time override options_ are ignored.
```
aphillips, line 54: formatting

```suggestion
Any _operand_ options not matching the _date/time override options_ are ignored.
```

spec/functions/datetime.md
```diff
@@ -158,8 +153,9 @@ The function `:date` has these _options_:
 
 If the _operand_ of the _expression_ is an implementation-defined date/time type,
 it can include other option values.
-Any _operand_ options matching the `:datetime` _style options_ or _field options_ are ignored,
-as is any `style` option.
+Any _date/time override options_ of the operand are included in the resolved option values of the _expression_,
+with _options_ on the _expression_ taking priority over any options of the _operand_.
+Any _operand_ options not matching the _Date/time override options_ are ignored.
```
aphillips, line 158: ibid

spec/functions/datetime.md
```diff
@@ -198,8 +194,9 @@ The function `:time` has these _options_:
...
```
aphillips, line 199: ibid

spec/functions/datetime.md
eemeli:
```suggestion
Any _operand_ options not matching the _date/time override options_ are ignored.
```

sffc:
I've been clear about my concern with `:date` and `:time`, stated in https://github.com/unicode-org/message-format-wg/pull/1067#issuecomment-2882873871. I apologize that I didn't attend the 2025-06-02 meeting, but my concerns were made clear well before then.

macchiati:
I tend to agree with Shane's statement there. I see no value in the 3 way split; just more cumbersome for callers. Here's what he said:
> I've stated previously that I don't understand the mental model for having :date, :time, and :datetime but not :datetimezone (or :zoneddatetime). Just as a date is fundamentally different from a time, and a datetime is fundamentally different from a date or a time, a zoned datetime is fundamentally different from a date, time, or datetime. If we split date, time, and datetime into individual functions, why would we not do the same for the zoned versions?

```
:time
:timeWithZone
:timeWithDayOfWeek
:timeWithZoneAndDayOfWeek
:date
:dateWithEra
...
```

sffc:
An additional ergonomic issue with `:date` and `:time` is that it temps people to write messages such as `{$input :date} at {$input :time}` instead of `{$input :datetime}`.

aphillips:
@sffc noted:
> An additional ergonomic issue with :date and :time is that it temps people to write messages such as {$input :date} at {$input :time} instead of {$input :datetime}.

That temptation is not always wrong. UX designers sometimes want to enrich the text by separating the fields.

@macchiati noted:
> I tend to agree with Shane's statement there. I see no value in the 3 way split; just more cumbersome for callers.

The functions `:date` and `:time` have different functionality in this iteration. Without them `:datetime` might require an additional value of `none` for `dateStyle`/`timeStyle` to allow for part suppression and users have additional typing to get formatting they want.

As noted in the intro, this is NOT the final solution to date/time formatting. It's a stopgap while we decide on additional functions. I'm not opposed to reducing it to just `:datetime`, so long as we introduce no "deprecated at birth" functions, options, or option values. If that's not possible, we should focus on the final design instead of giving ourselves this breathing room.

aphillips:
closed without merging. superseded by 1083

PR #1081, opened 2025-06-13
Clarifications to 'Resolved Values' section

catamorphism:
I was re-reading the `MessageValue` example and noticed that it's easy to skip over what the last three methods are for, so I added a reference to the earlier text.

I also thought that the `getValue()` method could have another name, since it's confusing to think about the "value" that it returns, vs. the enclosing `MessageValue`. The name I thought of was `getResult()`, but I'm open to other suggestions. I also thought it needed a little more explanation.

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -156,7 +156,7 @@ and different implementations MAY choose to perform different levels of resoluti
 > interface MessageValue {
 >   formatToString(): string
 >   formatToX(): X // where X is an implementation-defined type
->   getResult(): unknown
+>   unwrap(): unknown
 >   resolvedOptions(): { [key: string]: MessageValue }
 >   selectKeys(keys: string[]): string[]
 >   directionality(): 'LTR' | 'RTL' | 'unknown'
@@ -172,21 +172,23 @@ and different implementations MAY choose to perform different levels of resoluti
 > - A _variable_ could be used as a _selector_ if
 >   calling the `selectKeys(keys)` method of its _resolved value_
 >   did not emit an error.
-> - Using a _variable_, the _resolved value_ of an _expression_
+> - The _resolved value_ of an _expression_
 >   could be used as an _operand_ or _option value_ if
->   calling the `getResult()` method of its _resolved value_ did not emit an error.
+>   calling the `unwrap()` method of its _resolved value_ did not emit an error.
+>   (This requires an intermediate _variable_ _declaration_.)
 >   In this use case, the `resolvedOptions()` method could also
 >   provide a set of option values that could be taken into account by the called function.
->   - In some cases (such as in a number formatting function), `getResult()` would
->     return the _operand_ of the function that returned the `MessageValue`. In other
->     cases (such as in a function that extracts a field from a data structure),
->     `getResult()` would return a different result than the _operand_.
+>   - The relationship between the _operand_ and the result of the `unwrap()` method
+>     is specific to each function. For example, with the built-in function `:number`,
+>     the `unwrap()` method would return the numeric value of the _operand_.
+>     In other cases, such as in a function that extracts a field from a data structure,
+>     `unwrap()` would return a different result than the _operand_.
 > - The `directionality()`, `isolate()`, and `isLiteralOptionValue()` methods
 >   fulfill requirements and recommendations mentioned elsewhere in this specification.
 >
 > Extensions of the base `MessageValue` interface could be provided for different data types,
 > such as numbers or strings,
-> for which the `unknown` return type of `getResult()` and
+> for which the `unknown` return type of `unwrap()` and
 > the generic `MessageValue` type used in `resolvedOptions()`
 > could be narrowed appropriately.
 > An implementation could also allow `MessageValue` values to be passed in as input variables,
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -180,7 +180,7 @@ and different implementations MAY choose to perform different levels of resoluti
 >   provide a set of option values that could be taken into account by the called function.
 >   - The `unwrap()` method returns the _function_-specific result
 >     of the _function_'s operation.
->     For example, the handlers for the following functions might
+>     For example, the handlers for the following _functions_ might
 >     behave as follows:
 >     - The handler for the _default function_ `:number` returns a value
 >       whose `unwrap()` method returns
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -182,7 +182,7 @@ and different implementations MAY choose to perform different levels of resoluti
 >     cases (such as in a function that extracts a field from a data structure),
 >     `getResult()` would return a different result than the _operand_.
 > - The `directionality()`, `isolate()`, and `isLiteralOptionValue()` methods
->   fulfill the requirements and recommendations mentioned earlier in this section.
+>   fulfill requirements and recommendations mentioned elsewhere in this specification.
 >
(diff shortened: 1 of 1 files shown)


spec/formatting.md
```diff
@@ -173,13 +174,19 @@ and different implementations MAY choose to perform different levels of resoluti
 >   did not emit an error.
 > - Using a _variable_, the _resolved value_ of an _expression_
 >   could be used as an _operand_ or _option value_ if
->   calling the `getValue()` method of its _resolved value_ did not emit an error.
+>   calling the `getResult()` method of its _resolved value_ did not emit an error.
 >   In this use case, the `resolvedOptions()` method could also
 >   provide a set of option values that could be taken into account by the called function.
+>   - In some cases (such as in a number formatting function), `getResult()` would
+>     return the _operand_ of the function that returned the `MessageValue`. In other
```
eemeli, line 181: This is incorrect; for an expression like `{42 :number}` the function is expected to return the numerical value 42 rather than the literal string `'42'`, which is the operand here.

Also, please do use [semantic line breaks](https://sembr.org/) in spec text.

spec/formatting.md
```diff
@@ -173,13 +174,19 @@ and different implementations MAY choose to perform different levels of resoluti
...
+>     cases (such as in a function that extracts a field from a data structure),
+>     `getResult()` would return a different result than the _operand_.
+> - The `directionality()`, `isolate()`, and `isLiteralOptionValue()` methods
+>   fulfill the requirements and recommendations mentioned earlier in this section.
```
eemeli, line 185: "earlier" is incorrect, as at least isolation is only mentioned in bidi handling.

```suggestion
> - The `directionality()`, `isolate()`, and `isLiteralOptionValue()` methods
>   fulfill requirements and recommendations mentioned elsewhere in this specification.
```

eemeli (changes requested):
I don't mind changing the accessor name, but see inline comments for a few things that need fixing.

spec/formatting.md
aphillips: I think the intention of this added paragraph is correct, but I think it could be a lot clearer.

Something like:

> - The value returned by `getResult()` might simply be the _operand_
>   of the function that returned the `MessageValue`,
>   or it might be some other value.
>   For example, for a function that extracts field from a data structure,
>   the result would be the extracted value,
>   rather than the _operand_ it was extracted from.

aphillips (changes requested):
I think the name change could be clearer?

spec/formatting.md
```diff
@@ -155,7 +156,7 @@ and different implementations MAY choose to perform different levels of resoluti
 > interface MessageValue {
 >   formatToString(): string
 >   formatToX(): X // where X is an implementation-defined type
->   getValue(): unknown
+>   getResult(): unknown
```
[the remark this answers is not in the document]
catamorphism: The idea is that `getResult()` returns the return value of the function handler; the resolved value is that return value, wrapped in an object that provides methods.

`getResolvedValue()` doesn't seem right to me, since the resolved value is the thing implementing the `MessageValue` interface, no?

spec/formatting.md
catamorphism: I've edited this part; please take another look.

spec/formatting.md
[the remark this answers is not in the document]
catamorphism: Since I think of `MessageValue` as "wrapping" something, I thought of another name, `unwrap()`, that returns the thing being wrapped. Let me know how that looks.

spec/formatting.md
```diff
@@ -173,13 +174,19 @@ and different implementations MAY choose to perform different levels of resoluti
 >   did not emit an error.
 > - Using a _variable_, the _resolved value_ of an _expression_
 >   could be used as an _operand_ or _option value_ if
->   calling the `getValue()` method of its _resolved value_ did not emit an error.
+>   calling the `getResult()` method of its _resolved value_ did not emit an error.
```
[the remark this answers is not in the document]
catamorphism: I did a slightly different rewording; let me know what you think.

spec/formatting.md
```diff
@@ -171,15 +172,24 @@ and different implementations MAY choose to perform different levels of resoluti
...
-> - Using a _variable_, the _resolved value_ of an _expression_
+> - The _resolved value_ of an _expression_
 >   could be used as an _operand_ or _option value_ if
->   calling the `getValue()` method of its _resolved value_ did not emit an error.
+>   calling the `unwrap()` method of its _resolved value_ did not emit an error.
+>   (This requires an intermediate _variable_ _declaration_.)
 >   In this use case, the `resolvedOptions()` method could also
 >   provide a set of option values that could be taken into account by the called function.
+>   - The relationship between the _operand_ and the result of the `unwrap()` method
+>     is specific to each function. For example, with the built-in function `:number`,
```
aphillips, line 182: I feel like the is dancing around what we're trying to say. Perhaps:

```suggestion
>   - The `unwrap()` method does not merely return the _resolved value_
>     of the _operand_.
>     Instead, it returns the function-specific result of the function's operation.
>     For example, the default function `:number` produces the implementation-defined
>     numeric value of the operand,
>     even if the original _operand_'s value were a literal.
>     A different function might produce a transformed value instead.
>     For example, an `:uppercase` function might produce an uppercase string in place
>     of the original _operand_ value,
>     or a function that extracts a field from a data structure
>     might return the extracted value.
```

catamorphism: See what you think about [d70142f](https://github.com/unicode-org/message-format-wg/pull/1081/commits/d70142fcef1745a7162327dbac56f0b57bf842e5)? In particular, I want to make it clear that the function handler returns a `MessageValue` that has an `unwrap()` method that returns something. This makes the text a little repetitive, but I'm open to suggestions.

Also, I know @mihnita didn't like the name `unwrap()`, but in the WG meeting yesterday I wasn't hearing an alternative name suggestion.

aphillips: > ... I want to make it clear that the function handler returns a MessageValue that has an unwrap() method that returns something...

The `unwrap()` method/function returns something unless it emits an error, unless you presume that the ctor blows up on a bad value? A late-binding implementation might not bind the resolved value until required to do so, which might be very late indeed.

The more I think of it, the more I want to bikeshed the name towards `MessageValue.resolvedValue()`, not because that's how you'd want to name an interface or its method but because we are _consistent_ in calling the thing we're talking about a _resolved value_ and the point of the example is to illustrate that concept. We don't talk about "unwrapping" anywhere. I'd go further and suggest that `MessageValue` is wrong because its actually an `Expression` interface (perhaps `MessageFormatExpression` for clarity).

I think we do a good job not of illustrating that the method is not just required to resolve the operand, but that it might also apply local processing to it, which could include a change in type (for typed languages/environments).

I feel like the example should maybe be made less concrete, not merely start with "you're not required to do it like this" but to be more pseudocode-like, e.g.:

> For example, an implementation might require function handlers
> to provide an interface similar to:
>```
>MessageExpression {
>    getResolvedValue(), // unwrap if you prefer
>    ... // to show we might be omitting stuff
>    /* for functions that support selection 
>       note that I changed to match our new text */
>    Match(rv, key),
>    BetterThan(rv, key1, key2),
>    /* for functions that support formatting */
>    formatToString(),
>    formatToX(), // where X is an impl-def type, perhaps to support a formatToParts functionality
>    ... // other methods needed such as `isLiteral`, `isolate`, or `directionality`
>    }
>```

PR #1084, opened 2025-06-28
Fix contradiction in markup resolution

catamorphism:
Previously, this stated that markup resolution MUST always succeed. However, resolution of `u:` options can fail, so that wasn't true.

### DIFF
--- a/spec/formatting.md
+++ b/spec/formatting.md
@@ -193,47 +193,6 @@ _Expressions_ are used in _declarations_ and _patterns_.
 _Markup_ is only used in _patterns_.
 _Options_ are used in _expressions_ and _markup_.
 
-#### Option Resolution
-
-**_<dfn>Option resolution</dfn>_** is the process of computing the _options_
-for a given _expression_ or _markup_. 
-_Option resolution_ results in a mapping of string _identifiers_ to _resolved values_.
-The order of _options_ MUST NOT be significant.
-
-> For example, the following _message_ treats both both placeholders identically:
-> ```
-> {$x :ns:func option1=foo option2=bar} {$x :ns:func option2=bar option1=foo}
-> ```
-
-For each _option_:
-
-1. Let `res` be a new empty mapping.
-1. For each _option_:
-   1. Let `id` be the string value of the _identifier_ of the _option_.
-   1. Let `rv` be the _resolved value_ of the _option value_.
-   1. If `rv` is a _fallback value_:
-      1. Emit a _Bad Option_ error, if supported.
-   1. Else:
-      1. If the _option value_ consists of a _literal_:
-         1. Mark `rv` as a _literal_ _option value_.
-      1. Set `res[id]` to be `rv`.
-1. Return `res`.
-
-> [!NOTE]
-> If the _resolved value_ of an _option value_ is a _fallback value_,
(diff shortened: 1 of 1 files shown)


catamorphism:
I noticed this when looking at this test:

https://github.com/unicode-org/message-format-wg/blob/main/test/tests/u-options.json#L37

eemeli (changes requested):
While the resolution of individual options can fail, this still holds: https://github.com/unicode-org/message-format-wg/blob/944804044f47de11ee6db496330aaa9c203aae61/spec/formatting.md?plain=1#L410-L412

In other words, overall resolution of _markup_ will always succeed.

If there is some language in `u:` options or elsewhere that implies something different, that needs to be fixed rather than changing this.

eemeli (changes requested):
I think the more appropriate fix here would be for the "Option Resolution" section to be moved to be directly under "Expression and Markup Resolution", as it pertains to _markup resolution_ as well as _function resolution_, and for its first sentence to be corrected to note that.

Then we would not need to repeat the RFC2119 language that it already includes here.

catamorphism:
> I think the more appropriate fix here would be for the "Option Resolution" section to be moved to be directly under "Expression and Markup Resolution", as it pertains to _markup resolution_ as well as _function resolution_, and for its first sentence to be corrected to note that.
> 
> Then we would not need to repeat the RFC2119 language that it already includes here.

Done.

spec/formatting.md
```diff
@@ -191,6 +191,48 @@ and different implementations MAY choose to perform different levels of resoluti
 
 _Expressions_ are used in _declarations_ and _patterns_.
 _Markup_ is only used in _patterns_.
+_Options_ are used in _expressions_ and _markup_.
+
+#### Option Resolution
```
eemeli, line 196: The section should probably go either immediately before or after Markup Resolution, rather than being the first one. These sections have a top-down-ish order, and we start from "Expression and Markup Resolution".

spec/formatting.md
```diff
@@ -191,6 +191,48 @@ and different implementations MAY choose to perform different levels of resoluti
...
+
+**_<dfn>Option resolution</dfn>_** is the process of computing the _options_
+for a given _expression_ or _markup_ _placeholder_.
```
eemeli, line 199: It's just _markup_.
```suggestion
**_<dfn>Option resolution</dfn>_** is the process of computing the _options_
for a given _expression_ or _markup_. 
```

spec/formatting.md
eemeli: Your current proposal has the sections ordered as:
1. Option Resolution
2. Expression Resolution
3. Markup Resolution
4. Fallback Resolution

I think the order should be instead:
1. Expression Resolution
2. Markup Resolution
3. Option Resolution
4. Fallback Resolution

spec/formatting.md
```diff
@@ -434,6 +435,8 @@ supported by the implementation, process them as specified.
 Such `u:` options MAY be removed from the resolved mapping of _options_.
 
 The resolution of _markup_ MUST always succeed.
+(_Option resolution_ MAY emit errors,
+but such errors MUST NOT be fatal.)
```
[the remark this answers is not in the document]
eemeli: Yeah, that works.

eemeli (approved):

PR #1094, opened 2025-08-04
Add :percent

### LINKED ISSUE #956: [FEEDBACK] Please move style percent to its own function

eemeli:
Closes #956

As discussed on the preceding weekly calls, it looks like the ~least worst~ best option for us is to add a dedicated `:percent` function, to match what we're doing splitting `:currency` and `:unit` from `:number`.

This PR adds that function, initially as Draft. It also updates the related design doc to match this choice, and updates the test suite.

To make it into LDML 48, this PR will need to land by 2025-08-18 at latest.

### DIFF
--- a/spec/functions/number.md
+++ b/spec/functions/number.md
@@ -423,7 +423,7 @@ The function `:percent` is a selector and formatter for percent values.
 
 ##### `:percent` Operands
 
-The function `:percent` requires a [Number Operand](#number-operands) as its _operand_.
+The function `:percent` requires a _numeric operand_ as its _operand_.
 
 When either selecting or formatting the _expression_,
 the numeric value of the _operand_ is multiplied by 100.


spec/functions/number.md
```diff
@@ -413,6 +413,107 @@ contains an implementation-defined currency value
...
+  - `expand`
+  - `trunc`
+  - `halfCeil`
+  - `halfFloor`
+  - `halfExpand` (default)
+  - `halfTrunc`
+  - `halfEven`
+
+When formatting or selecting, each of the options is applied
+after the _resolved value_ of the _operand_ is multiplied by 100
```
aphillips, line 483: This seems misleading? The options apply to the formatting (or selection). They don't alter the value itself.

The example that follows illustrates the application of options to the formatting, but isn't really an example of what this text currently says. 

Perhaps:
```suggestion
The _resolved value_ of the _operand_ is multiplied by 100
prior to formatting or selection.
Each _option_ is applied to that value.
```

spec/functions/number.md
```diff
@@ -413,6 +413,107 @@ contains an implementation-defined currency value
...
+
+> For example, this _placeholder_:
+>
+> ```
+> {0.1234 :percent maximumFractionDigits=1}
+> ```
+>
+> would be formatted as "12.3%" in an English locale.
```
aphillips, line 491: would => might

spec/functions/number.md
```diff
@@ -413,6 +413,107 @@ contains an implementation-defined currency value
...
+with _options_ on the _expression_ taking priority over any options of the _operand_.
+Options with the following names are however discarded if included in the _operand_:
+
+- `minimumIntegerDigits`
+- `roundingIncrement`
+- `select`
+
+##### Resolved Value
+
+The _resolved value_ of an _expression_ with a `:percent` _function_
```
aphillips, line 506: Is the resolved value * 100 or the original value? This implies (but does not state) that it is the original value. We should be explicit here.

spec/functions/number.md
```diff
@@ -413,6 +413,107 @@ contains an implementation-defined currency value
...
+contains an implementation-defined numerical value
+of the _operand_ of the annotated _expression_,
+together with the resolved options' values.
+
+##### Selection
+
+The _function_ `:percent` performs selection as described in [Number Selection](#number-selection) below,
+with selection always using `plural` selection mode,
+and with the _resolved value_ of the _operand_ multiplied by 100.
```
aphillips, line 515: I would include an example, e.g.

> For example,
>```
> .input $pct = {0.1234 :percent maximumFractionDigits=0}
> .match {$pct}
> 12 {I match!}
> *  {Not so much}

spec/functions/number.md
eemeli:
```suggestion
> might be formatted as "12.3%" in an English locale.
```

spec/functions/number.md
eemeli: The resolved value is not multiplied by 100. This is necessary to allow `:percent` calls to be chained without ill effect:
```
.local $p = {1 :percent}
{{{$p :percent}}}
```
should still format as something like "100%" rather than "10,000%".

I'll add a clarification below to make this explicit.

spec/functions/number.md
```diff
@@ -413,6 +413,121 @@ contains an implementation-defined currency value
...
+- `roundingIncrement`
+- `select`
+
+##### Resolved Value
+
+The _resolved value_ of an _expression_ with a `:percent` _function_
+contains an implementation-defined numerical value
+of the _operand_ of the annotated _expression_
+(not multiplied by 100),
+together with the resolved options' values.
```
aphillips, line 512: This is fine, but I'd make it a separate sentence.

```suggestion
of the _operand_ of the annotated _expression_
together with the resolved options' values.
The _resolved value_ is not altered by `:percent`,
that is, it is not multiplied by 100.
```

spec/functions/number.md
aphillips: Love the example, but the text above the example contradicts what we just wrote above about the _resolved value_.

aphillips: Perhaps for the text...

> The _function_ `:percent` performs selection as described in Number Selection below.
> This selection always uses the `plural` selection mode,
> and is performed on the _resolved value_ of the _operand_
> multiplied by 100.
>
> [!NOTE]
> The _resolved value_ of the _operand_ is not altered by selection.

aphillips (changes requested):

spec/functions/number.md
```diff
@@ -413,6 +413,121 @@ contains an implementation-defined currency value
...
+
+##### Selection
+
+The _function_ `:percent` performs selection as described in [Number Selection](#number-selection) below,
+with selection always using `plural` selection mode,
+and with the _resolved value_ of the _operand_ multiplied by 100.
```
eemeli, line 518:
```suggestion
The _function_ `:percent` performs selection as described in [Number Selection](#number-selection) below.
This selection always uses the `plural` selection mode,
and is performed on the _resolved value_ of the _operand_
multiplied by 100.
```

exploration/percent-format.md
```diff
@@ -134,9 +135,10 @@ _What prior decisions and existing conditions limit the possible design?_
 
 ## Proposed Design
 
-_Describe the proposed solution. Consider syntax, formatting, errors, registry, tooling, interchange._
+Add a new, dedicated `:percent` function,
+which scales the _resolved value_ of its _operand_ by 100.
 
-TBD
+Separately, allow `:unit unit=percent` formatting, without scaling.
```
aphillips, line 141: Last material comment. I would make this suppositional and not part of the percent design.

```suggestion
This design could allow a function such as `:unit unit=percent` 
for formatting values without scaling
when the working group tackles units.
```

aphillips (approved):
One comment on the design doc, which is optional.

Thank you for wrapping up percent!!

spec/functions/number.md
```diff
@@ -413,6 +413,123 @@ contains an implementation-defined currency value
...
+> It is proposed for inclusion in a future release of this specification and is not Stable.
+
+The function `:percent` is a selector and formatter for percent values.
+
+##### Operands
+
+The function `:percent` requires a [Number Operand](#number-operands) as its _operand_.
+
+When either selecting or formatting the _expression_,
+the _resolved value_ of the _operand_ is multiplied by 100.
```
catamorphism, line 429: This is being a little fast and loose with the term "resolved value", right? If the non-normative example is followed, the resolved value is an object with a bunch of methods, and `unwrap()` would have to be called to get the underlying number.

I'm not sure how I would phrase it better unless the underlying language about resolved values was made more precise, though.

spec/functions/number.md
```diff
@@ -413,6 +413,123 @@ contains an implementation-defined currency value
...
+- `select`
+
+##### Resolved Value
+
+The _resolved value_ of an _expression_ with a `:percent` _function_
+contains an implementation-defined numerical value
+of the _operand_ of the annotated _expression_
+together with the resolved options' values.
+The _resolved value_ is not altered by `:percent`,
+that is, it is not multiplied by 100.
```
catamorphism, line 513: Same comment here (the previous sentence says it _contains_ a numerical value, but this sentence implies it _is_ a numerical value.)

macchiati:
I agree that the terminology is very confusing. For example, a resolved value is not a "container" as the following implies (my bolding).

> The _resolved value_ of an _expression_ with a `:string` _function_
> **contains the string value** of the _operand_ of the annotated _expression_,
> together with its resolved locale and directionality.
> None of the _options_ set on the _expression_ are part of the _resolved value_.

It is instead suggested that an RV is like the class definition given in https://github.com/unicode-org/message-format-wg/blob/1f87da0574e793972d84532b25dc950115734739/spec/formatting.md#resolved-values

> While this specification does not require it, a resolved value could be implemented by requiring each function handler to return a value matching the following interface:

```
interface MessageValue {
  formatToString(): string
  formatToX(): X // where X is an implementation-defined type
  unwrap(): unknown
  resolvedOptions(): { [key: string]: MessageValue }
  match(key: string): boolean
  betterThan(key1: string, key2: string): boolean
  directionality(): 'LTR' | 'RTL' | 'unknown'
  isolate(): boolean
  isLiteralOptionValue(): boolean
}
```

But that really doesn't sit very well at all with saying "contains the string value". Instead, we'd really have to say exactly what _each_ of the methods do. And we don't do that. 

If we are going down this road, we could be more specific, and replace "While this specification does not require it, a resolved value could be implemented by requiring each function handler to return a value matching the following interface:"
by the following (we use the "logically equivalent" language at many places in Unicode):

> A resolved value is logically equivalent to a value matching the following interface:

We can then just have a standard table for each method, with the 2nd column's cell describing what it does; that way we won't miss something important (like describing whether the unwrap() method produces a value (vs an error) and if so, what the value is defined to be.)

No implementation is required to use those names or functions internally, but that is a more concrete, uniform way to describe what a function does and does not do

BTW
1. I think 'unknown' above would be much better as 'Y // where Y is an implementation-defined type'.
2. The above also implies that resolvedOptions has exactly 1 key-value pair, whereas I think the intent is that it return a map with 0 or more key-value pairs.
3. I find the unwrap() name and explanation to be particularly obscure. 
4. The grammar of the following sentence is a bit fuzzy:
* The resolved value of an expression could be used as an operand or option value if calling the unwrap() method of its resolved value did not emit an error. (This requires an intermediate variable declaration.) In this use case, the resolvedOptions() method could also provide a set of option values that could be taken into account by the called function.
* I think it is intended to be something like:
* The resolved value of an expression can be used as an operand or option value, but **only if** calling the unwrap() method does not emit an error. In that case, then resolvedOptions() method provides a set of option values that can be taken into account by the called function. These option values may be a subset of those used to create the resolved value.

spec/functions/number.md
```diff
@@ -413,6 +413,123 @@ contains an implementation-defined currency value
...
+  - `expand`
+  - `trunc`
+  - `halfCeil`
+  - `halfFloor`
+  - `halfExpand` (default)
+  - `halfTrunc`
+  - `halfEven`
+
+The _resolved value_ of the _operand_ is multiplied by 100
+at the start of formatting or selection.
```
eemeli, line 483:
```suggestion
The numeric value of the _operand_ is multiplied by 100
at the start of formatting or selection.
```

eemeli:
@catamorphism Thank you for catching the loose use of "resolved value". I've now replaced the places where that's used with "numeric value", which is what we use in describing `:offset`.

@macchiati I've moved your comment to be its own new issue #1095, as it goes beyond the scope of this PR, and we should discuss and address it separately.

spec/functions/number.md
```diff
@@ -413,6 +413,123 @@ contains an implementation-defined currency value
...
+- `select`
+
+##### Resolved Value
+
+The _resolved value_ of an _expression_ with a `:percent` _function_
+contains an implementation-defined numerical value
+of the _operand_ of the annotated _expression_
+together with the resolved options' values.
+The _resolved value_ is not altered by `:percent`,
+that is, its numerical value is not multiplied by 100.
```
eemeli, line 513:
```suggestion
The numerical value of the _resolved value_ of the _expression_
is the same as the numerical value of its _operand_;
it is not multiplied by 100.
```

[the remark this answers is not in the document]
eemeli: @catamorphism I revised the text to match your suggestion; PTAL?

PR #1096, opened 2025-08-10
Preparations for LDML 48 release

eemeli:
- Editorial changes, most significantly making section titles unique and adding _numeric operand_ as a term
- Updated acknowledgements (list order is first by commit count, second by added lines)
- Add `publish/` with an instructional readme & publication script.


aphillips:
Some of this is done by the LDML build process? Did you talk to @srl295 about making these changes? I can walk you through the LDML tooling (which is pretty slick)

aphillips:
Ah, so you found the steps I was going to point to, which I automated on my desktop Linux box, but didn't check in 😛. The changes to make unique IDs in this PR are totally necessary. I wouldn't pre-generate the TOC the way your readme (and my notes) suggests. The tool is fine--if you've fixed it so the IDs are unique. Some of the `.md` boilerplate you have is more than I think you need to provide?? Steven would know, tho.

eemeli:
> Looks good. Did you run TR tool to get the output?

Didn't find it until now, at: https://github.com/unicode-org/cldr/tree/main/tools/scripts/tr-archive

The `fix-tocs` script does do the ToC generation, and appears to include a bit of TR35-internal link fixing. I'll update the script here to use it.

I'm rather tempted to set up a GitHub action to keep a rendered copy of our spec version continuously available via GitHub Pages; it should be pretty easy with the `archive` script.

srl295:
> Some of this is done by the LDML build process? Did you talk to @srl295 about making these changes? I can walk you through the LDML tooling (which is pretty slick)

So as merged, I don't see any major structural issues.    This is now better documented and supported by tools than the tr-archiver has been! Congrats I think. 

Anyway, I don't see any problem with pushing on here and we'll deal with any issues once we actually merge back into CLDR…

PR #1104, opened 2026-03-06
Specify well-formed values for timezones, calendars & units

### LINKED ISSUE #935: Well-formed vs valid
Added text 2024-11-24

----
I think we need to be careful about our usage of the terms 'well-formed' and 'valid'. The following is not fully fleshed out; it is more of a discussion of the issue and some ideas for the future.

We often reference other sources for identifiers, and want them to be interpreted according to that source. Sources that change over time should (and typically do) distinguish between well-formed and valid. For example, 'ge:manic' is not a **well-formed** locale identifier, and 'de-Flub' is not a **valid** locale identifier. However, 'de-Flub' could (conceivably) become valid in the future, if a script is given the code 'Flub'. Good sources also never remove identifiers, or make material changes in the meaning, but may deprecate them: those are still treated as valid.

When we reference such sources in message format, such as with option values, we have a few goals.
- Ideally, implementations could only accept well-formed and valid identifiers, and only interpret them according to the source semantics. For example, interpret 'de' as German and not as Dezfuli.
- However, we don't want to force implementations to break if they don't support all the identifiers, nor if they don't support the latest version, or if they support an identifier that has become deprecated.

----

This is also true for our own enums, . We have in [registry.md](https://github.com/unicode-org/message-format-wg/blob/main/spec/registry.md):

> Implementations MAY accept additional option values for options defined here. However, such values might become defined with a different meaning in the future, including with a different, incompatible name or using an incompatible value space. Supporting implementation-specific option values for standard or optional functions is NOT RECOMMENDED.

We also have BNF:

> option = identifier o "=" o (literal / variable)

The implications are that conformant implementation can interpret any of:

{$x :currency compactDisplay=short} 
{$x :currency compactDisplay=medium} 
{$x :currency compactDisplay=μικρός}
{$x :currency compactDisplay=|🐭|}
{$x :currency compactDisplay=$myDisplay}

It can also interpret:

{$x :currency currency=CAD} 
{$x :currency currency=MyCurrency} 
{$x :currency currency=δολάριοΚαναδά}
{$x :currency currency=|¥|}
{$x :currency currency=|🐭|}
{$x :currency currency=$myCurrency}

It could also interpret compactDisplay=short by formatting a long form, and compactDisplay=long by formatting a short form. Or a value of CAD as being GBP, etc.

**This level of freedom seems counterproductive for interoperability.**

----

So I propose that we have the general rule something like the following, where option values are defined according to a reference to an external source

- An implementation MUST ignore any option with a literal option value that is ill-formed according to its external source, and signal that error. This allows linters and message builders to catch ill-formed values early.
    - [It must ignore the option `locale=|ge:manic|`]
- An implementation MUST ignore any option with an option value that isn't valid according to any version of the external source. 
    - [At the time of this writing, must ignore `locale=|dab|`]
- An implementation SHOULD (but need not) ignore an option with an option value that is valid according to some version of the external source.
    - [An implementation might not support Dezfuli, and thus ignore `locale=|def|`; it may also ignore all deprecated language identifiers, and thus ignore `locale=|daf|`.]
- If an implementation doesn't ignore an option, then it MUST interpret its option value in accordance with some version of the source.
    - [It must not interpret 'de' as Dezfuli, or 'def' as German.]

_Ignore_ means that the expression is interpreted as if the option were not there.  (I won't talk here about what signals to the caller are associated with that.)

----
I think we could apply that to our standard enum option values, such as the following in https://github.com/unicode-org/message-format-wg/blob/main/spec/registry.md#options-1, so that |@!$| could be recognized as ill-formed.

- useGrouping
    - auto (default)
    - always
    - never
    - min2

That is, perhaps we can have a rule in the registry for our functions, something like: the default well-formedness criteria for standard function option values matches the constraints on function option identifiers in README.md. Thus |$abc| would be ill-formed for useGrouping. Any function option that had different criteria for well-formedness of its values would simply have have an explicit well-formedness statement.

----

### LINKED ISSUE #936: Add short timezone identifiers
I had made a comment about adding the short timezone identifiers to 

> timeZone (default is system default time zone or UTC)
> - valid identifier per [BCP175](https://www.rfc-editor.org/rfc/rfc6557)

That we should add the following:

> or per [Unicode Timezone Identifier](https://unicode.org/reports/tr35/tr35.html#UnicodeTimezoneIdentifier)

(That allows implementations to use the short ids, but shouldn't require them. See also https://github.com/unicode-org/message-format-wg/issues/935)

### LINKED ISSUE #962: [FEEDBACK] Be more specific about time zone IDs
registry.md says that time zone IDs are "A valid time zone identifier" with links to LDML and TZDB.

However, LDML and TZDB define two different standards for time zone identifiers: a BCP-47 ID and a longer form string ID. The longer form IANA time zone ID is generally more well understood.

If we keep this function (https://github.com/unicode-org/message-format-wg/issues/961), it should be restricted to one set of identifiers, probably the TZDB identifiers.

eemeli:
Closes #935
Closes #936
See also https://github.com/unicode-org/message-format-wg/issues/961#issuecomment-3109611039
Closes #962

As resolved on this week's call, the options for which we accept values defined elsewhere should all be spec'd as "well-formed" rather than "valid". As we do not mandate that all well-formed values are accepted without error, or that the errors emitted for "not well-formed" and "not valid" are different from each other, this allows for a linter/validator to detect not-well-formed values early, while not requiring a formatting implementation to differentiate well-formed-but-not-valid from not-well-formed values.

While the spec text does not make mention of Unicode short time zone identifiers, it should be noted that all such values are technically well-formed time zone identifiers according to RFC 9557. This means that an implementation _could_ support such values, but they would probably not work in all implementations. If we'd like to discourage this, I think including a non-normative note to that effect might be appropriate (suggestions for exact text are most welcome).

This PR also addresses https://github.com/unicode-org/message-format-wg/issues/961#issuecomment-3109611039 by defining the behaviour when formatting date/time values that do not include a time zone.

### DIFF
--- a/spec/functions/datetime.md
+++ b/spec/functions/datetime.md
@@ -256,12 +256,11 @@ The following _option_ is REQUIRED to be available on
 the functions `:datetime`, `:date`, and `:time`.
 
 - `timeZone`
-  - A valid time zone identifier
-    (see [TZDB](https://www.iana.org/time-zones)
-    and [LDML](https://www.unicode.org/reports/tr35/tr35-dates.html#Time_Zone_Names)
-    for information on identifiers)
   - `input`
-  - `UTC`
+  - A well-formed time zone identifier matching the `time-zone-name` rule of
+    [RFC 9557](https://www.rfc-editor.org/rfc/rfc9557#name-abnf).
+  - A well-formed UTC offset matching the `time-numoffset` rule of
+    [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339#section-5.6).
 
 The default value for `timeZone` is the default time zone provided by the _formatting context_.
 
@@ -269,12 +268,18 @@ The value `input` corresponds to the time zone of the _operand_.
 If it is used and the _resolved value_ of the _operand_ does not include a time zone or offset,
 a _Bad Operand_ error is emitted and the default time zone is used to format the _expression_.
 
+If the _resolved value_ of the _operand_ does not include a time zone or offset,
+the _resolved value_ of the `timeZone` _option_ is used as its time zone.
+
 If the _resolved value_ of the _operand_ includes a time zone or offset,
 and the _resolved value_ of the `timeZone` _option_ is different from that,
 an implementation SHOULD convert the _resolved value_ of the _operand_
 to the time zone indicated by the _resolved value_ of the `timeZone` _option_.
 If such conversion is not supported, an implementation MAY alternatively
 emit a _Bad Option_ error and use a _fallback value_ as the _resolved value_ of the _expression_.
+
+> [!NOTE]
+> A date/time type encapsulating a Unix time is considered to include UTC as its time zone.
 
 The following _option_ is REQUIRED to be available on
 the functions `:datetime` and `:time`:
@@ -287,4 +292,5 @@ The following _option_ is RECOMMENDED to be available on
 the functions `:datetime`, `:date`, and `:time`.
 
 - `calendar`
-  - valid [Unicode Calendar Identifier](https://unicode.org/reports/tr35/tr35.html#UnicodeCalendarIdentifier)
+  - A well-formed [Unicode Calendar Identifier](https://unicode.org/reports/tr35/tr35.html#UnicodeCalendarIdentifier),
+    i.e. a [uvalue](https://unicode.org/reports/tr35/tr35.html#uvalue).
--- a/spec/functions/datetime.md
+++ b/spec/functions/datetime.md
@@ -257,6 +257,7 @@ the functions `:datetime`, `:date`, and `:time`.
 
 - `timeZone`
   - `input`
+  - `UTC`
   - A well-formed time zone identifier matching the `time-zone-name` rule of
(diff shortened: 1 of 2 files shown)


spec/functions/datetime.md
```diff
@@ -256,26 +256,31 @@ The following _option_ is REQUIRED to be available on
 the functions `:datetime`, `:date`, and `:time`.
 
 - `timeZone`
-  - A valid time zone identifier
-    (see [TZDB](https://www.iana.org/time-zones)
-    and [LDML](https://www.unicode.org/reports/tr35/tr35-dates.html#Time_Zone_Names)
-    for information on identifiers)
   - `input`
-  - `UTC`
+  - A well-formed time zone identifier matching the `time-zone-name` rule of
```
aphillips: Although it is well-formed, we might want to specify the special value `UTC`, since that is a common time zone that users need. Specifying it explicitly would aid interop, since a very minimal implementation would support UTC plus whatever the contextual zone is.

aphillips (changes requested):
Minor comments. Generally looks very good. I support the direction this is taking.

gibson042 (approved):
I am also happy with this direction.

PR #1105, opened 2026-03-06
Use RFC 3339/9557 rules for :datetime, :date & :time operand values

### LINKED ISSUE #680: Restrict literals for `:date` and `:time`

eemeli:
Closes #680

Since the current regular expression definition of "Date and Time Operands" was added to the spec in #687, [RFC 9557](https://www.rfc-editor.org/rfc/rfc9557) has been published, the datetime functions got refactored in #1083, and we're now _this_ close to having Temporal advance to Stage 4 at TC39.

With all fo the above in mind, we ought to revisit the rules for the `:datetime`, `:date`, and `:time` operands, and define their string values separately, relying on RFC 9557 or RFC 3339 rules, as proposed here.

I'm not 100% sure about allowing the [`time`](https://www.rfc-editor.org/rfc/rfc3339#appendix-A) rule for `:time` operands, as I think we'd be best served by something like
```
HH:mm:ss.sss
```
with the fractional and integer seconds being optional, but I've not found a good reference for such a rule. I'm not completely convinced that we ought to accept a time with an offset but no date as a literal value, as `time` allows for.

### DIFF
--- a/spec/functions/datetime.md
+++ b/spec/functions/datetime.md
@@ -36,9 +36,30 @@ If no options are specified, this function defaults to the following:
 ##### `:datetime` Operands
 
 The _operand_ of the `:datetime` function is either
-an implementation-defined date/time type
-or a _date/time literal value_, as defined in [Date and Time Operand](#date-and-time-operands).
+an implementation-defined date/time type,
+or a string matching the `date-time-ext` rule of
+[RFC 9557](https://www.rfc-editor.org/rfc/rfc9557#name-abnf).
 All other _operand_ values produce a _Bad Operand_ error.
+
+If the _resolved value_ of the _operand_ does not include a time,
+`00:00:00` is used as the time.
+
+If the _resolved value_ of the _operand_ does not include a time zone or offset,
+the _resolved value_ of the `timeZone` _option_ is used as its time zone.
+
+> [!NOTE]
+> String values passed as variables in the _formatting context_'s
+> _input mapping_ can be formatted as datetime values as long as their
+> contents match the `date-time-ext` rule.
+>
+> For example, if the value of the variable `now` were the string
+> `2024-02-06T16:40:00Z`, it would behave identically to the local
+> variable in this example:
+>
+> ```
+> .local $example = {|2024-02-06T16:40:00Z| :datetime}
+> {{{$now :datetime} == {$example}}}
+> ```
 
 ##### `:datetime` Options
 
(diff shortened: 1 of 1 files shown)


spec/functions/datetime.md
```diff
@@ -36,10 +36,31 @@ If no options are specified, this function defaults to the following:
...
 The _operand_ of the `:datetime` function is either
-an implementation-defined date/time type
-or a _date/time literal value_, as defined in [Date and Time Operand](#date-and-time-operands).
+an implementation-defined date/time type,
+or a string matching the `date-time-ext` rule of
+[RFC 9557](https://www.rfc-editor.org/rfc/rfc9557#name-abnf).
 All other _operand_ values produce a _Bad Operand_ error.
 
+If the _resolved value_ of the _operand_ does not include a time,
+`00:00:00` is used as the time.
```
gibson042: Note that time 00:00 is not guaranteed to exist for every date in any particular time zone, although that is also true for e.g. 02:xx.

gibson042 (changes requested):

spec/functions/datetime.md
```diff
@@ -97,10 +118,33 @@ If no options are specified, this function defaults to the following:
...
 
 The _operand_ of the `:date` function is either
-an implementation-defined date/time type
-or a _date/time literal value_, as defined in [Date and Time Operand](#date-and-time-operands).
+an implementation-defined date/time type,
+or a string matching the `date-time-ext` rule of
+[RFC 9557](https://www.rfc-editor.org/rfc/rfc9557#name-abnf),
+or a string matching the `full-date` rule of
+[RFC 3339](https://www.rfc-editor.org/rfc/rfc3339#section-5.6).
 All other _operand_ values produce a _Bad Operand_ error.
```
rxaviers, line 126: > @eemeli said: For (time), the proposal specifically includes this: 
>  > If the _resolved value_ of the _operand_ does not include a time, 
 a _Bad Operand_ error is produced. 

@eemeli, for consistency, shall we include the analogous here?

```suggestion
All other _operand_ values produce a _Bad Operand_ error.

If the _resolved value_ of the _operand_ does not include a date, 
 a _Bad Operand_ error is produced. 
```

... to guarantee this edge case also produces _Bad Operand_

PR #1112, opened 2026-05-26
CLDR-19371 update license

srl295:
Spec is under different terms, rest is Unicode v3

See also https://github.com/unicode-org/cldr/pull/5762

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,8 +1,25 @@
-UNICODE LICENSE V3
-
 COPYRIGHT AND PERMISSION NOTICE
 
-Copyright © 2019-2023 Unicode, Inc.
+Copyright © 2019–2026 Unicode, Inc.
+
+Unicode Technical Standard #35, Unicode Locale Data Markup Language
+(LDML) Part 9 MessageFormat made available in this repository is
+protected by copyright, and permission must be obtained from Unicode,
+Inc. prior to any reproduction, modification, or other use not permitted
+by the Unicode Terms of Use <https://www.unicode.org/copyright.html>.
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -1,8 +1,25 @@
-UNICODE LICENSE V3
-
 COPYRIGHT AND PERMISSION NOTICE
 
-Copyright © 2019-2023 Unicode, Inc.
+Copyright © 2019–2026 Unicode, Inc.
```
eemeli, line 3: Could we leave out the year here, so that this does not need annual updates?
```suggestion
Copyright © Unicode, Inc.
```

eemeli (approved):

LICENSE
srl295: we've gone back and forth on this with legal and my understanding is that no, they want the latest year in there. The good news if any is that I have tools that are supposed to check the year and complain if it's wrong (i.e. almost automatic updating).

alerque: There is considerable legal precedent for just having the *first* year (not the latest). The automatic period of copyright from first publishing is ~70 years anyway, and anybody that says the copyright expired because somebody didn't bump the date is just wrong. If the source is active and the changes substantial that will count as renewing anyway and "date bumping" when there are *not* significant changes brings other problems.

https://liferay.dev/b/how-and-why-to-properly-write-copyright-statements-in-your-code#why-not-bump-the-year-on-change

srl295: @alerque I am, of course, not the legal counsel, but I will forward this on.

PR #1113, opened 2026-06-10
In specification for :test:function, use correct error

catamorphism:
There is no bad-input error; use bad-operand instead.


aphillips (approved):
LGTM

eemeli (approved):

